PR #17, opened 2016-11-08
install tsc during build

### LINKED ISSUE #13: Project doesn't build locally without TypeScript Compiler on path
After cloning the repo, a local build causes two errors (in ChatSample and SocketsSample) when you don't have tsc on your path.

anurse:
fixes #13 

/cc @davidfowl @BrennanConroy @moozzyk 
/fyi @MisterJames

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -1,0 +1,21 @@
+var VERSION='0.1'
+var FULL_VERSION='0.1'
+var AUTHORS='Microsoft'
+
+use-standard-lifecycle
+k-standard-goals
+
+default TYPESCRIPT_PROJECT_GLOB = "src/*/tsconfig.json"
+
+var nodeDirectory = '${Path.Combine(Directory.GetCurrentDirectory(), "bin", "nodejs")}'
+var nodeBin = '${ Path.Combine(nodeDirectory, "node_modules", ".bin") }'
+var localTsc = '${ Path.Combine(nodeBin, "tsc") }'
+var localTscExists = '${ File.Exists(localTsc) }'
+
+#install-tsc target='initialize'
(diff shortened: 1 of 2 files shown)


samples/SocketsSample/project.json
```diff
@@ -55,7 +55,7 @@
   },
 
   "scripts": {
-    "precompile": [ "tsc --project ../../src/Microsoft.AspNetCore.SignalR.Client.TS/ --out wwwroot/js/signalr-client.js" ],
+    "precompile": [ "../../bin/nodejs/node_modules/.bin/tsc --project ../../src/Microsoft.AspNetCore.SignalR.Client.TS/ --out wwwroot/js/signalr-client.js" ],
```
moozzyk: won't this break in VS though where tsc is installed globally?
Also with VS - not sure how to enforce tsc 2.0

anurse: It would break if you haven't run ./build.sh before. I tried to get a local TypeScript to work cleanly but I was having trouble with `PATH`.

Right now we need something to unblock the CI. Eventually I think we'll want the TypeScript code to actually be in a TypeScript Project so we can reference it from other places.

anurse: Alternatively we should just install TypeScript globally during build... which seems... ok, I guess :)

moozzyk: Only it may not work - see: https://github.com/Microsoft/TypeScript/issues/5995#issuecomment-254408809

As you said - let's get this in to unblock the CI and then try fix VS.

moozzyk:
:ship::it:

anurse:
:up: :date:

moozzyk:
Looks good. Thanks for adding typings for EventSource for now to make warnings go away.

Issue #3, opened 2016-11-03
TS-client outstanding work


moozzyk:
As per https://github.com/davidfowl/Sockets/pull/12

 - [x] add transport closed event (33c94c1)
 - [x] remove fallback as per discussion today (325c909dffcd3cd5e7b174869a29c78d00474564)
 - [x] figure out modules (2039a189710ddf7875d4a198348fa02f28926dfb)
 - [x] unify errors (33c94c1, 913354f)
 - [x] review capturing of `this` in callbacks/eventhandlers (b32d3ad4e18dd314aca32bbb53b928e298f3cf04)
 - [x] ~~logging~~ tracked now as: https://github.com/aspnet/SignalR/issues/485
 - [x] convert RpcConnection to HubConnection (e6b34fdce30f2da12dd01afe97a2eb29c21509aa)
 - [x] detach dispatching from transports/connection (70e4ec9a06b5c3080fdff7b1392f50ff347e9a2a)
 - [x] enable plugging in custom formatters (https://github.com/aspnet/SignalR/issues/12)
 - [x] split messages (https://github.com/aspnet/SignalR/issues/6)
 - [x] joined messages (especially important for longpolling - multiple messages in one poll response) (78dfd278c151af26ec7873acc6f6c62ee6fb0749)
 - [x] is there a better way of bringing typings for EventSource (currently we have them in the repo: https://github.com/aspnet/SignalR/commit/02c98137c7f12c61899edef32be6a1a22f0e8971)
 - [x] tests
 - [x] consider using `async`/`await` (requires typescript 2.1.1?) (3a01d6c)
 - [ ] handle cases where a method called from the server does not exist on the client
 - [x] hubConnection needs error event (2bbca5e)
 - [x] review usages of httpClient. It shouldn't be created each time it is needed and it should be possible to provide alternate implementations (e.g. for mocking) (a9c7e14)
 - [ ] try using IHttpClient for polling
 - [x] don't build the client multiple times when running build.cmd (after moving to MSBuild) (0162c19ccf682f270b274a77b87f2a9f54c164d3)
 - [x] create typescript VS projects so that they show up in VS (0162c19ccf682f270b274a77b87f2a9f54c164d3)
-  [x] https://github.com/aspnet/SignalR/commit/ccabbbc5167ef65ccf99d55939145d349f1de0de#commitcomment-20536735

valeriob:
[...]

PR #25, opened 2016-11-14
Removing transport fallback in the client

moozzyk:


davidfowl:
Time to get some async await!

moozzyk:
Yes, once we move from .NET 4 errrr ES6

davidfowl:
Typescript 2.1.1 allows down level compilation from async await to ES3

moozzyk:
I added this to the [list](https://github.com/aspnet/SignalR/issues/3) and will see if it feasible but it's not a priority ATM (need tests first)

Issue #29, opened 2016-11-15
Improve WebSocket testing


anurse:
Need functional tests (which needs #26) and unit tests (which need a little refactoring of the transport to take an existing socket rather than handshaking).

Also need handshaking tests.

anurse:
Sounds like @moozzyk and @mikaelm12 are on the functional testing aspects of this.

PR #33, opened 2016-11-15
fix #32 by handling connection header correctly

### LINKED ISSUE #32: Connection header parsing in Websockets doesn't handle "keep-alive, Upgrade"
@anurse 

Bring back Firefox!

anurse:
Fixes #32 

We'll need tests for this when we do #29 

/cc @BrennanConroy


BrennanConroy:
:shipit:

PR #34, opened 2016-11-16
Add prefix to redis channel names

BrennanConroy:
@anurse @davidfowl @mikaelm12 @moozzyk 

Fix issue where user and group have the same name

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -87,7 +87,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
                 Arguments = args
             };
 
-            return PublishAsync(typeof(THub).FullName + "." + groupName, message);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -87,7 +87,7 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
                 Arguments = args
             };
 
-            return PublishAsync(typeof(THub).FullName + "." + groupName, message);
+            return PublishAsync(typeof(THub).FullName + ".group." + groupName, message);
```
mikaelm12: Should Group and User be capitalized?

BrennanConroy:
Should I do the same for connectionId channel name? You could technically set the group name or user name to the connectionId

BrennanConroy:
Technically we get connectionId for free because the other 2 have a prefix.

mikaelm12:
Maybe for consistency then

PR #35, opened 2016-11-16
add tests to WebSockets transport

anurse:
also adds some error handling

/cc @BrennanConroy @moozzyk @mikaelm12 @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/WebSockets.cs
+++ b/src/Microsoft.AspNetCore.Sockets/WebSockets.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
@@ -22,6 +23,7 @@ namespace Microsoft.AspNetCore.Sockets
         private readonly WebSocketOpcode _opcode;
         private readonly ILogger _logger;
 
+        public WebSockets(Connection connection, Format format) : this(connection, format, loggerFactory: null) { }
         public WebSockets(Connection connection, Format format, ILoggerFactory loggerFactory)
         {
             _channel = (HttpConnection)connection.Channel;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Sockets/WebSockets.cs
```diff
@@ -22,6 +23,7 @@ public class WebSockets : IHttpTransport
         private readonly WebSocketOpcode _opcode;
         private readonly ILogger _logger;
 
+        public WebSockets(Connection connection, Format format) : this(connection, format, loggerFactory: null) { }
```
davidfowl: Get rid of NullLogger.Instance

anurse: Ok, wasn't sure what the best practice here was since I saw `NullLogger`. Will update.

src/Microsoft.AspNetCore.Sockets/WebSockets.cs
```diff
@@ -43,44 +45,61 @@ public WebSockets(Connection connection, Format format, ILoggerFactory loggerFac
...
+                // The application finished sending. We're not going to keep the connection open,
+                // so close it and wait for the client to ack the close
+                _channel.Input.CompleteWriter();
+                _logger.LogDebug(!failed ? "Application finished sending. Sending close frame." : "Application failed during sending. Sending InternalServerError close frame");
+                await socket.CloseAsync(!failed ? WebSocketCloseStatus.NormalClosure : WebSocketCloseStatus.InternalServerError);
+
+                _logger.LogDebug("Waiting for the client to close the socket");
+
+                // TODO: Timeout.
+                await receiving;
```
davidfowl: I need to think about this. I had this comment before:

https://github.com/aspnet/SignalR/blob/d00f1f93b2676706203f0ce65251a68bea9cfcbe/src/Microsoft.AspNetCore.Sockets/WebSockets.cs#L35-L38

src/Microsoft.AspNetCore.Sockets/WebSockets.cs
```diff
@@ -43,44 +45,61 @@ public WebSockets(Connection connection, Format format, ILoggerFactory loggerFac
...
-                    // so close it and wait for the client to ack the close
-                    _channel.Input.CompleteWriter();
-                    _logger.LogDebug("Application finished sending. Sending close frame.");
-                    await ws.CloseAsync(WebSocketCloseStatus.NormalClosure);
-
-                    _logger.LogDebug("Waiting for the client to close the socket");
-                    // TODO: Timeout.
-                    await receiving;
+                    // The receiver faulted or cancelled. This means the client is probably broken. Just propagate the exception and exit
+                    receiving.GetAwaiter().GetResult();
```
davidfowl: We don't handle errors properly.

anurse:
:up: :date: I changed things around a little to keep the input channel open until the client finishes, even if the output finishes. We'll need to add some functional tests to ensure we have the right behavior propagating up the pipeline (pun slightly intended) in case of disgraceful termination. In theory, the WebSocket keep alive should cause an exception to propagate up if the client fails to ack, but we should try to build a test for that (docker maybe?)

src/Microsoft.AspNetCore.Sockets/WebSockets.cs
anurse: Yeah, we'll need to do that.

davidfowl: Change this line to just:

```
await receiving;
```

Looks better than the blocking call.

src/Microsoft.AspNetCore.Sockets/WebSockets.cs
```diff
@@ -43,44 +51,64 @@ public WebSockets(Connection connection, Format format, ILoggerFactory loggerFac
...
+                _logger.LogDebug(!failed ? "Application finished sending. Sending close frame." : "Application failed during sending. Sending InternalServerError close frame");
+                await socket.CloseAsync(!failed ? WebSocketCloseStatus.NormalClosure : WebSocketCloseStatus.InternalServerError);
+
+                _logger.LogDebug("Waiting for the client to close the socket");
+
+                // Wait for the client to close.
+                // TODO: Consider timing out here and cancelling the receive loop.
+                await receiving;
+
+                // Now close the input, since the receiver
```
moozzyk: This is what I

PR #36, opened 2016-11-17
Introducing modules for the ts client

moozzyk:


davidfowl:
Is this the first pass? Is there any way to do this without using requirejs?

moozzyk:
Yes this is the first pass. I wanted to give it a spin and see how it works and iterate on this. This should also unblock adding node based tests. 
Re requirejs - you need a loader to be able to load amd modules. One idea that I have not yet full investigated is using es2015/es6 modules. Looks like it is not yet widely supported by browsers (only Edge supports it) so on other browsers you would still need to use a js library to use it. es2015 modules should be compatible with node but as I said I have not tried it yet. 
We probably could also just have just a js file but we would have to make sure it does not dump everything in the global scope.

davidfowl:
> We probably could also just have just a js file but we would have to make sure it does not dump everything in the global scope.

We need a single js file, we can't require a loader.

moozzyk:
:up::date:

samples/ChatSample/project.json
```diff
@@ -91,7 +91,11 @@
   },
 
   "scripts": {
-    "precompile": [ "dotnet bundle", "tsc --project ../../src/Microsoft.AspNetCore.SignalR.Client.TS/ --outDir wwwroot/lib/signalr-client/ --module amd" ],
+    "precompile": [ "dotnet bundle",
```
moozzyk: This is super ugly. I am going to try to move it to gulp entirely.

samples/ChatSample/project.json
```diff
@@ -91,7 +91,11 @@
   },
 
   "scripts": {
-    "precompile": [ "dotnet bundle", "tsc --project ../../src/Microsoft.AspNetCore.SignalR.Client.TS/ --out wwwroot/js/signalr-client.js" ],
+    "precompile": [ "dotnet bundle",
```
davidfowl: This looks kinda crazy

moozzyk:
:up::date: - moved building js client to gulp

test/Microsoft.AspNetCore.SignalR.Test.Server/project.json
```diff
@@ -39,7 +39,7 @@
   },
 
   "scripts": {
-    "precompile": [ "npm install", "gulp copy-jasmine"],
+    "precompile": [ "npm install", "npm run gulp -- --gulpfile %project:Directory%/gulpfile.js copy-jasmine"],
```
moozzyk: @pranavkm @davidfowl @muratg @BrennanConroy 
This fixes the CI.

.gitignore
```diff
@@ -35,7 +35,6 @@ node_modules/
 *.nuget.props
 *.nuget.targets
 autobahnreports/
-signalr-client.js
```
davidfowl: Why is this removed from ignore now?

davidfowl: Ah I see

gulpfile.js
```diff
@@ -0,0 +1,40 @@
+const gulp = require('gulp');
```
davidfowl: Why is this in the root?

davidfowl: Why is this in the root?

package.json
```diff
@@ -6,19 +6,27 @@
   "directories": {
```
davidfowl: Why is this in the root?

samples/SocketsSample/.gitignore
```diff
@@ -0,0 +1 @@
+wwwroot/js
```
davidfowl: Why not globally?

samples/SocketsSample/wwwroot/.gitignore
```diff
@@ -1 +1 @@
-js/
\ No newline at end of file
+lib/signalr-client
```
davidfowl: Why do we have gitignores per sample? We don't do this anywhere else

gulpfile.js
moozzyk: because it's common for multiple sample projects

package.json
moozzyk: because it is common to multiple projects. Also avoids having node_modules in each projects that uses node

gulpfile.js
davidfowl: Having it in the root looks untidy. Put it in the samples folder.

samples/SocketsSample/.gitignore
davidfowl: Can we be consistent about this? I don't like the per project git ignore. Lets just keep the global signalr-client.js include for now...

samples/SocketsSample/wwwroot/.gitignore
moozzyk: If I unified the projects and put signalr in the same relative path we could move it to the common .gitignore.

gulpfile.js
moozzyk: but it is not just samples - this is/will be part of the build process. This is how we build the library for browsers (i.e. as a single file).

moozzyk: If we wanted in the future we will be able to replace gulp with MSBuild...

moozzyk: Maybe I could move it to a build or tools folder then?

davidfowl: It really, shouldn't be in the root. It only applies to the project that produces the js file. We're not using gulp for project builds. Move it to the typescript client then.

samples/SocketsSample/.gitignore
moozzyk: OK. as per my comment below I can unify where we put signalr (e.g. wwwroot/lib/signalr) and ignore this in the main .gitignore.

moozzyk:
:up::date: Moved gulpfile to the Client.TS folder, unified signalr-client.js location. Removed local .gitignore files in samples.

PR #43, opened 2016-11-18
[WIP] Prototype .NET Client

anurse:
PROTOTYPE. For initial review only.

Only supports Long Polling (why not WebSockets? I don't know, I just picked a transport that seemed easy to bash out in an hour, I may have chosen poorly :P). Updated ClientSample to use it.

API is way early. I think there will be some major clean-up. This is also only a Sockets client, not a SignalR RPC/Hubs client. It just bridges the transports back over to a Pipeline and lets you read/write.

/cc @DamianEdwards @davidfowl @BrennanConroy @mikaelm12 @moozzyk


moozzyk:
I propose I merge my tests first and you use the same server for tests here.

src/Microsoft.AspNetCore.SignalR.Client/ConnectionFactory.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    // REVIEW: Split into Microsoft.AspNetCore.Sockets.Client?
+    public class ConnectionFactory
+    {
+        private readonly HttpClient _httpClient;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly PipelineFactory _pipelineFactory;
+        private readonly ILogger _logger;
+
+        public ConnectionFactory(HttpClient httpClient, PipelineFactory pipelineFactory, ILoggerFactory loggerFactory)
```
moozzyk: This can't take httpClient. You won't be able to mock http.

src/Microsoft.AspNetCore.SignalR.Client/ConnectionFactory.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using System.IO.Pipelines;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    // REVIEW: Split into Microsoft.AspNetCore.Sockets.Client?
+    public class ConnectionFactory
```
moozzyk: Why do we want the factory? Looking at the code now you now need a couple of factories and a transport to start a connection. Looks like a very complicated API comparing to what we had before.

src/Microsoft.AspNetCore.SignalR.Client/ConnectionFactory.cs
moozzyk: What's the point of taking PipelineFactory? Does not look like you can provide your own useful implementation - there is no interface and no virtual methods. 
@davidfowl - this PipeLineFactory makes code hard to test - should it be changed or we need to provide our own interface delegating the code to the default implementation?

src/Microsoft.AspNetCore.SignalR.Client/Connection.cs
```diff
@@ -0,0 +1,33 @@
...
+    public class Connection : IPipelineConnection
+    {
+        private IPipelineConnection _consumerPipe;
+        private ITransport _transport;
+        private readonly ILogger _logger;
+
+        public Uri Url { get; }
+
+        public Connection(Uri url, ITransport transport, IPipelineConnection consumerPipe, ILoggerFactory _loggerFactory)
+        {
```
moozzyk: validate parameters in public APIs
we should have a ctor that does not take LoggerFactory (and possibly transport - I think we should default to webSockets)

samples/ClientSample/Program.cs
```diff
@@ -2,68 +2,101 @@
...
-            RunWebSockets().GetAwaiter().GetResult();
-        }
+        public static void Main(string[] args) => MainAsync(args).Wait();
 
-        private static async Task RunWebSockets()
+        public static async Task MainAsync(string[] args)
         {
-            var ws = new ClientWebSocket();
-            await ws.ConnectAsync(new Uri("ws://localhost:5000/chat/ws"), CancellationToken.None);
+            var baseUrl = "http://localhost:5000/chat";
```
moozzyk: how does it work if you don't specify transport specific endpoint?

src/Microsoft.AspNetCore.SignalR.Client/Utils.cs
```diff
@@ -0,0 +1,32 @@
...
+            return builder.Uri;
+        }
+
+        internal static Uri AppendQueryString(Uri url, string qs)
+        {
+            var builder = new UriBuilder(url);
+            if(string.IsNullOrEmpty(builder.Query))
+            {
+                builder.Query += "?";
+            } else
```
moozzyk: nit: format

src/Microsoft.AspNetCore.SignalR.Client/Utils.cs
```diff
@@ -0,0 +1,32 @@
...
+            if (!builder.Path.EndsWith("/"))
+            {
+                builder.Path += "/";
+            }
+            builder.Path += path;
+            return builder.Uri;
+        }
+
+        internal static Uri AppendQueryString(Uri url, string qs)
+        {
```
moozzyk: nit: return on empty/null qs?

src/Microsoft.AspNetCore.SignalR.Client/LongPollingTransport.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public class LongPollingTransport : ITransport
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
```
moozzyk: preparing requests should be common. Even now this is not being sent when requesting a connection id.

src/Microsoft.AspNetCore.SignalR.Client/ConnectionState.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public enum ConnectionState
```
moozzyk: needed at all?

src/Microsoft.AspNetCore.SignalR.Client/Connection.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.IO.Pipelines;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public class Connection : IPipelineConnection
```
moozzyk: What is the benefit of implementing the IPipelineConnection here?

src/Microsoft.AspNetCore.SignalR.Client/project.json
```diff
@@ -0,0 +1,33 @@
...
+    "warningsAsErrors": true,
+    "keyFile": "../../tools/Key.snk",
+    "nowarn": [
+      "CS1591"
+    ],
+    "xmlDoc": true
+  },
+
+  "dependencies": {
+    "NETStandard.Library": "1.6.1-*",
```
moozzyk: is 1.6.1 because of ~~channels~~ pipelines?

samples/ClientSample/Program.cs
```diff
@@ -2,68 +2,101 @@
...
 
-            var sending = Task.Run(async () =>
+            using (var httpClient = new HttpClient(new LoggingMessageHandler(loggerFactory, new HttpClientHandler())))
+            using (var pipelineFactory = new PipelineFactory())
             {
-                string line;
-                while ((line = Console.ReadLine()) != null)
+                var connectionFactory = new ConnectionFactory(httpClient, pipelineFactory, loggerFactory);
+                logger.LogInformation("Connecting to {0}", baseUrl);
+                var transport = new LongPollingTransport(httpClient, loggerFactory);
```
moozzyk: I actually found the ws sample useful as well...

davidfowl: Yea, lets keep that around.

src/Microsoft.AspNetCore.SignalR.Client/ConnectionFactory.cs
davidfowl: Hmmmm, it's possible this should be an interface we're passing around. The problem is it gives you back concrete channels.

src/Microsoft.AspNetCore.SignalR.Client/PipelineConnection.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.IO.Pipelines;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    internal class PipelineConnection : IPipelineConnection
```
davidfowl: We should move something like this into pipelines directly.

src/Microsoft.AspNetCore.SignalR.Client/Utils.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    internal static class Utils
```
davidfowl: Urg, no. Also we might be able to use the types in WebUtilities for this.

/cc @Tratcher

src/Microsoft.AspNetCore.SignalR.Client/project.json
```diff
@@ -0,0 +1,33 @@
...
+    "System.IO.Pipelines": "0.1.0-*",
+    "Microsoft.Extensions.Logging.Abstractions": "1.2.0-*"
+  },
+
+  "frameworks": {
+    "netstandard1.6": {}
```
davidfowl: Should cross compile like our other stuff.

src/Microsoft.AspNetCore.SignalR.Client/Utils.cs
Tratcher: Meh, server side we don't use System.Uri for anything. You can keep these.

src/Microsoft.AspNetCore.SignalR.Client/ConnectionFactory.cs
anurse: The most common use of Pipeline Factory is to create a pair of `IPipelineConnection` objects, representing either end of a pipeline. I wonder if we should put that on an interface or something and use that instead. That's all we need here, a pair of pipelines representing input/output.

src/Microsoft.AspNetCore.SignalR.Client/Connection.cs
anurse: A PipelineConnection is an object that provides a reader for one pipeline and a writer for another pipeline. That's what our Connection object is, so we want to implement the interface to inherit any extension methods, etc. (though, to be fair, I don't think there are any extension methods on here yet :))

samples/ClientSample/Program.cs
anurse: Right now it just adds `/poll` and `/send` on the end. I expect we'll tidy this up, that's why this is extremely early prototype code :).

src/Microsoft.AspNetCore.SignalR.Client/LongPollingTransport.cs
anurse: Of course, I refactor by need though and this prototype was a single transport, so I hadn't reached that yet.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -73,7 +74,11 @@ private static string MakeNewConnectionId()
 
         private static void Scan(object state)
         {
-            ((ConnectionManager)state).Scan();
+            // Don't trim timed out connections if the debugger is attached
+            if (!Debugger.IsAttached)
```
anurse: I agree for now, but this is something that will mess with anyone trying to debug a SignalR Hub.

src/Microsoft.AspNetCore.SignalR.Client/LongPollingTransport.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            using (cancellationToken.Register(() => _pipeline.Input.Complete()))
+            {
+                try
+                {
+                    while (!cancellationToken.IsCancellationRequested)
+                    {
+                        var result = await _pipeline.Input.ReadAsync();
+                        var buffer = result.Buffer;
+                        if (buffer.Length == 0 || result.IsCompleted)
```
anurse: Yep, I even used it elsewhere, just hadn't done a pass through this to fix that up :)

anurse:
@moozzyk Well, this is a long way from merging anyway, as I said in the initial comment :).

src/Microsoft.AspNetCore.SignalR.Client/ConnectionFactory.cs
anurse: Re `HttpClient`, can we not use a custom handler to mock out Http? If that's been tried and doesn't work I'm certainly happy to change, but don't we use `HttpClient` in our own Test Host already?

moozzyk: Would we only write E2E/functional tests then? Emulating failure conditions can also be challenging...

samples/ClientSample/Program.cs
moozzyk: For whatever reason I could not find the place we append these but now I can.

src/Microsoft.AspNetCore.SignalR.Client/ConnectionFactory.cs
moozzyk: I misread your comment - if we can override the default behavior (i.e. cut/mock the network) then I don't see a problem.

src/Microsoft.AspNetCore.SignalR.Client/project.json
anurse: No, just because everything else is 1.6.1/1.2.0

anurse:
I got RPC working now. It's still ugly and needs a lot of API clean-up. I had to rework IInvocationAdapter, and I'm not 100% happy with how it went. I think we need to think about the formatting API and what it's going to look like from both sides (client and server) since we're going to need both parties to be able to receive all message types.

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,147 @@
...
+            _reader = ReceiveMessages(_readerCts.Token);
+        }
+
+        public Task<T> Invoke<T>(string methodName, params object[] args) => Invoke<T>(methodName, CancellationToken.None, args);
+        public async Task<T> Invoke<T>(string methodName, CancellationToken cancellationToken, params object[] args) => ((T)(await Invoke(methodName, typeof(T), cancellationToken, args)));
+
+        public Task<object> Invoke(string methodName, Type returnType, params object[] args) => Invoke(methodName, returnType, CancellationToken.None, args);
+        public async Task<object> Invoke(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
+        {
+            // Write a
```
moozzyk: Use single quotes for 'a'?

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,147 @@
...
+            var descriptor = new InvocationDescriptor()
+            {
+                Id = GetNextId(),
+                Method = methodName,
+                Arguments = args
+            };
+
+            // I just want an excuse to use 'irq' as a variable name...
+            var irq = new InvocationRequest(cancellationToken, returnType);
```
moozzyk: we need nmi as well then...

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.IO.Pipelines;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets.Client;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public class RpcConnection : IDisposable
```
moozzyk: Should this be called HubConnection? (I am planning to rename the RpcConnection in ts client to HubConnection because we decided not to expose RPC per se).

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,147 @@
...
+        private readonly Connection _connection;
+        private readonly Stream _stream;
+        private readonly Task _reader;
+
+        private readonly CancellationTokenSource _readerCts = new CancellationTokenSource();
+        private readonly ConcurrentDictionary<string, InvocationRequest> _pendingCalls = new ConcurrentDictionary<string, InvocationRequest>();
+
+        private int _nextId = 0;
+
+        private RpcConnection(Connection connection, IInvocationAdapter adapter)
```
moozzyk: It would be useful to be able to pass fake Connection here for unit testing

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        public Task<T> Invoke<T>(string methodName, params object[] args) => Invoke<T>(methodName, CancellationToken.None, args);
+        public async Task<T> Invoke<T>(string methodName, CancellationToken cancellationToken, params object[] args) => ((T)(await Invoke(methodName, typeof(T), cancellationToken, args)));
+
+        public Task<object> Invoke(string methodName, Type returnType, params object[] args) => Invoke(methodName, returnType, CancellationToken.None, args);
+        public async Task<object> Invoke(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
+        {
+            // Write a
+            var descriptor = new InvocationDescriptor()
```
moozzyk: nit: remove `()`

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,147 @@
...
+                        irq.Completion.TrySetResult(result.Result);
+                    }
+                }
+            }
+        }
+
+        private string GetNextId()
+        {
+            // Increment returns the incremented value, so we subtract one to get the previous value.
+            return (Interlocked.Increment(ref _nextId) - 1).ToString();
```
moozzyk: why can't you just use the returned value?

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,147 @@
...
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                // This is a little odd... we want to remove the InvocationRequest once and only once so we pull it out in the callback,
+                // and stash it here because we know the callback will have finished before the end of the await.
+                InvocationRequest irq = default(InvocationRequest);
+                var result = await _adapter.ReadInvocationResultDescriptorAsync(_stream, id =>
+                {
+                    if (!_pendingCalls.TryRemove(id, out irq))
+                    {
+                        throw new InvalidOperationException($"Unsolicited response received for invocation '{id}'");
```
moozzyk: this should be logged but probably should not throw

samples/ClientSample/Program.cs
```diff
@@ -7,6 +7,8 @@
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
```
moozzyk: I think having Socket sample side by side to Hub sample is useful

samples/ClientSample/Program.cs
```diff
@@ -45,56 +47,22 @@ public class Program
...
-            logger.LogInformation("Receive loop starting");
-            using (cancellationToken.Register(() => connection.Input.Complete()))
-            {
-                while (!cancellationToken.IsCancellationRequested)
-                {
-                    var result = await connection.Input.ReadAsync();
-                    var buffer = result.Buffer;
-                    try
+                    // Set up handler
+                    connection.On("Send", new[] { typeof(string) }, a =>
```
moozzyk: We would need to add [generic extensions](https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Client/HubProxyExtensions.cs#L122) later...

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -12,33 +13,47 @@
...
             _connection = connection;
             _stream = connection.GetStream();
             _adapter = adapter;
+            _logger = logger;
+
             _reader = ReceiveMessages(_readerCts.Token);
         }
 
+        // TODO: Client return values/tasks?
+        // TODO: Overloads that use type parameters (like On<T1>, On<T1, T2>, etc.)
```
moozzyk: I commented on this above actually... It was implemented as extensions before: https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Client/HubProxyExtensions.cs#L122

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -66,57 +90,86 @@ public void Dispose()
...
+            _logger.LogTrace("Beginning receive loop");
             while (!cancellationToken.IsCancellationRequested)
             {
                 // This is a little odd... we want to remove the InvocationRequest once and only once so we pull it out in the callback,
                 // and stash it here because we know the callback will have finished before the end of the await.
-                InvocationRequest irq = default(InvocationRequest);
-                var result = await _adapter.ReadInvocationResultDescriptorAsync(_stream, id =>
+                var message = await _adapter.ReadMessageAsync(_stream, this, cancellationToken);
+
+                var invocationDescriptor = message as InvocationDescriptor;
```
moozzyk: (When doing protobuf) I was wondering if we can have some kind of base class that would have the message type inside to avoid this kind of casts.

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -127,6 +180,38 @@ private string GetNextId()
             return (Interlocked.Increment(ref _nextId) - 1).ToString();
         }
 
+        Type IInvocationBinder.GetReturnType(string invocationId)
```
moozzyk: this explicit interface implementation is quite ugly - can we have a separate internal/nested class we pass as IInvocationBinder instead of RpcConnection?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -212,8 +206,8 @@ private void DiscoverHubMethods()
 
                         if (created)
                         {
-                            // Dispose the object if it's disposable and we created it
-                            (hub as IDisposable)?.Dispose();
+                        // Dispose the object if it's disposable and we created it
```
moozzyk: nit: formatting is a bit off

moozzyk:
This is probably separate but we need some lifecycle hooks - especially OnError/OnClosed. RpcConnection/HubConnection needs to complete all pending tcs'es if the connection is closed.

moozzyk:
We should also reject Send requests if the connection is not connected. Probably also kill the connection if Send throws.

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
anurse: Agreed, we may want an interface. Ideally we could just use a Pipeline, but I'm not sure if that fits in well with the necessary lifetime events... though we could probably model those with Complete on the Pipeline.

samples/ClientSample/Program.cs
anurse: Yeah, way ahead of you :). I'm bringing the socket sample back as a commented out chunk of code for now. I don't want to overload us with samples...

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
moozzyk: or you could initialize _nextId to -1. Starting at 1 is fine too - this is an opaque identifier

samples/ClientSample/Program.cs
```diff
@@ -2,67 +2,65 @@
...
 
-            var sending = Task.Run(async () =>
+            using (var httpClient = new HttpClient(new LoggingMessageHandler(loggerFactory, new HttpClientHandler())))
+            using (var pipelineFactory = new PipelineFactory())
             {
-                string line;
-                while ((line = Console.ReadLine()) != null)
+                logger.LogInformation("Connecting to {0}", baseUrl);
+                var transport = new LongPollingTransport(httpClient, loggerFactory);
+                using (var connection = await RpcConnection.ConnectAsync(new Uri(baseUrl), new JsonNetInvocationAdapter(), transport, httpClient, pipelineFactory, loggerFactory))
```
mikaelm12: It looks like you renamed this to HubConnection in the RpcConnection.cs file

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,245 @@
...
+            _stream = connection.GetStream();
+            _adapter = adapter;
+            _logger = logger;
+
+            _reader = ReceiveMessages(_readerCts.Token);
+        }
+
+        // TODO: Client return values/tasks?
+        // TODO: Overloads that use type parameters (like On<T1>, On<T1, T2>, etc.)
+        public void On(string methodName, Type[] parameterTypes, Action<object[]> handler)
```
davidfowl: Yea, we're going to talk about this pattern. I think we'll fundamentally move away from this.

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,245 @@
...
+                else
+                {
+                    var invocationResultDescriptor = message as InvocationResultDescriptor;
+                    if (invocationResultDescriptor != null)
+                    {
+                        DispatchInvocationResult(invocationResultDescriptor, cancellationToken);
+                    }
+                }
+            }
+            _logger.LogTrace("Ending receive loop");
```
davidfowl: Cancel all pending calls here?

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,245 @@
...
+                // Complete the request based on the result
+                if (!string.IsNullOrEmpty(result.Error))
+                {
+                    _logger.LogInformation("Completing Invocation #{0} with error: {1}", result.Id, result.Error);
+                    irq.Completion.TrySetException(new Exception(result.Error));
+                }
+                else
+                {
+                    _logger.LogInformation("Completing Invocation #{0} with result of type: {1}", result.Id, result.Result?.GetType()?.FullName ?? "<<void>>");
+                    irq.Completion.TrySetResult(result.Result);
```
davidfowl: We need to dispatch here.

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
anurse: Most, if not all, of the other patterns can be built on this primitive, so I figured it was a good start.

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -0,0 +1,245 @@
...
+        {
+            // Find the handler
+            InvocationHandler handler;
+            if (!_handlers.TryGetValue(invocationDescriptor.Method, out handler))
+            {
+                _logger.LogWarning("Failed to find handler for '{0}' method", invocationDescriptor.Method);
+            }
+
+            // TODO: Return values
+            handler.Handler(invocationDescriptor.Arguments);
```
anurse: Noted, but I'm not going to be able to get that in before handing off :).

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
anurse: Ditto.

anurse:
I've squashed and rebased on dev for @mikaelm12 to take over.

anurse:
Closing this PR now, but the branch remains open for work.

samples/ClientSample/HubSample.cs
```diff
@@ -0,0 +1,62 @@
...
+                    // Set up handler
+                    connection.On("Send", new[] { typeof(string) }, a =>
+                    {
+                        var message = (string)a[0];
+                        Console.WriteLine("RECEIVED: " + message);
+                    });
+
+                    while (!cts.Token.IsCancellationRequested)
+                    {
+                        var line = Console.ReadLine();
```
moozzyk: nit: does it mean you need to press enter after Ctrl+C to exit?

PR #44, opened 2016-11-21
Dispose Hubs from ActivatorUtilities

BrennanConroy:
We weren't disposing the hubs from `Activator.CreateInstance<THub>()` in a couple places

@moozzyk @anurse @davidfowl @mikaelm12 @muratg

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
@@ -63,9 +63,15 @@ namespace Microsoft.AspNetCore.SignalR
 
                 using (var scope = _serviceScopeFactory.CreateScope())
                 {
-                    var hub = scope.ServiceProvider.GetService<THub>() ?? Activator.CreateInstance<THub>();
-                    InitializeHub(connection, hub);
+                    bool created;
+                    var hub = CreateHub(scope.ServiceProvider, connection, out created);
+
                     await hub.OnConnectedAsync();
+
+                    if (created)
+                    {
+                        hub.Dispose();
+                    }
                 }
 
                 await DispatchMessagesAsync(connection);
@@ -74,9 +80,15 @@ namespace Microsoft.AspNetCore.SignalR
             {
                 using (var scope = _serviceScopeFactory.CreateScope())
                 {
-                    var hub = scope.ServiceProvider.GetService<THub>() ?? Activator.CreateInstance<THub>();
-                    InitializeHub(connection, hub);
+                    bool created;
+                    var hub = CreateHub(scope.ServiceProvider, connection, out created);
+
                     await hub.OnDisconnectedAsync();
+
+                    if (created)
+                    {
+                        hub.Dispose();
+                    }
                 }
 
                 await _lifetimeManager.OnDisconnectedAsync(connection);
@@ -132,11 +144,22 @@ namespace Microsoft.AspNetCore.SignalR
             }
         }
 
-        private void InitializeHub(Connection connection, THub hub)
+        private THub CreateHub(IServiceProvider provider, Connection connection, out bool created)
         {
+            var hub = provider.GetService<THub>();
+            created = false;
+
+            if (hub == null)
+            {
+                hub = ActivatorUtilities.CreateInstance<THub>(provider);
+                created = true;
+            }
+
             hub.Clients = _hubContext.Clients;
             hub.Context = new HubCallerContext(connection);
             hub.Groups = new GroupManager<THub>(connection, _lifetimeManager);
+
+            return hub;
         }
 
         private void DiscoverHubMethods()
@@ -169,17 +192,8 @@ namespace Microsoft.AspNetCore.SignalR
 
                     using (var scope = _serviceScopeFactory.CreateScope())
                     {
-                        var hub = scope.ServiceProvider.GetService<THub>();
-
-                        bool created = false;
-
-                        if (hub == null)
-                        {
-                            hub = Activator.CreateInstance<THub>();
-                            created = true;
-                        }
-
-                        InitializeHub(connection, hub);
+                        bool created;
+                        var hub = CreateHub(scope.ServiceProvider, connection, out created);
 
                         try
                         {
@@ -212,8 +226,7 @@ namespace Microsoft.AspNetCore.SignalR
 
                         if (created)
                         {
-                            // Dispose the object if it's disposable and we created it
-                            (hub as IDisposable)?.Dispose();
+                            hub.Dispose();
                         }
                     }
 
--- a/test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
@@ -1,0 +1,107 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO.Pipelines;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Internal;
+using Microsoft.AspNetCore.Sockets;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    public class HubEndpointTests
+    {
+        [Fact]
+        public async Task HubsAreDisposed()
+        {
+            var trackDispose = new TrackDispose();
+            var serviceProvider = CreateServiceProvider(s => s.AddSingleton(trackDispose));
+
+            var endPoint = serviceProvider.GetService<HubEndPoint<TestHub>>();
+
+            using (var connectionWrapper = new ConnectionWrapper())
+            {
+                var endPointTask = endPoint.OnConnectedAsync(connectionWrapper.Connection);
+
+                await connectionWrapper.HttpConnection.Input.ReadingStarted;
+
+                // kill the connection
+                connectionWrapper.Connection.Channel.Dispose();
+
+                await endPointTask;
+
+                Assert.Equal(2, trackDispose.DisposeCount);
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -132,6 +146,22 @@ public class HubEndPoint<THub, TClient> : EndPoint where THub : Hub<TClient>
...
+                hub = Activator.CreateInstance<THub>();
+                created = true;
+            }
+
+            InitializeHub(connection, hub);
+
+            return hub;
+        }
+
         private void InitializeHub(Connection connection, THub hub)
```
moozzyk: Do you still need this method? Why not inline it in CreateHub?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -63,9 +63,16 @@ public class HubEndPoint<THub, TClient> : EndPoint where THub : Hub<TClient>
 
                 using (var scope = _serviceScopeFactory.CreateScope())
                 {
-                    var hub = scope.ServiceProvider.GetService<THub>() ?? Activator.CreateInstance<THub>();
-                    InitializeHub(connection, hub);
+                    bool created;
```
muratg: `var created = false`? 

Also, I'd like someone to take a look at the name `created` in this context (and in `CreateHub` method below). It's set to true only if it's created by `Activator`, though even if it's coming through a `GetService` call, someone have "created" it before :)

moozzyk: `bool created` is fine. In general I find this pattern ugly and it is now across the entire file but honestly, it works.

davidfowl:
Add a test

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -63,9 +63,16 @@ public class HubEndPoint<THub, TClient> : EndPoint where THub : Hub<TClient>
 
                 using (var scope = _serviceScopeFactory.CreateScope())
```
davidfowl: Hubs are transient and a scope needs to be created whenever we make a hub. We'll need to see how bad the performance is for this.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: It's really a design flaw with the pattern. We need a HubActivator with a Create and Dispose like MVC's ControllerActivator (or is it the factory that disposes..)

BrennanConroy:
Added a test

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        public async Task HubsAreDisposed()
+        {
+            var serviceProvider = CreateServiceProvider();
+            var endPoint = serviceProvider.GetService<HubEndPoint<TestHub>>();
+
+            using (var connectionWrapper = new ConnectionWrapper())
+            {
+                var endPointTask = Task.Run(() => endPoint.OnConnectedAsync(connectionWrapper.Connection));
+
+                Thread.Sleep(100);
```
davidfowl: What?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                await endPointTask;
+
+                Assert.Equal(2, TestHub.DisposedCount);
+            }
+        }
+
+        private class TestHub : Hub
+        {
+            static public int DisposedCount = 0;
```
davidfowl: Why is this static?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -63,9 +63,16 @@ public class HubEndPoint<THub, TClient> : EndPoint where THub : Hub<TClient>
...
-                    InitializeHub(connection, hub);
+                    bool created;
+                    var hub = CreateHub(scope.ServiceProvider, connection, out created);
+
                     await hub.OnConnectedAsync();
+
+                    if (created)
+                    {
+                        // Dispose the object if it's disposable and we created it
+                        (hub as IDisposable)?.Dispose();
```
BrennanConroy: Odd, I thought I copied it from somewhere...

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
BrennanConroy: I wasn't sure what to do here, don't want to `await endPoint.OnConnectedAsync(...)` because it wont complete, and need to wait a small amount of time for `endPoint.OnConnectedAsync(...)` to actually run and get to `DispatchMessagesAsync`

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
BrennanConroy: So we can `Assert` that dispose was called on connect and on disconnect.

davidfowl: I think you should pass in an object that has the state and not use a static for this.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
davidfowl: Since you have a handle on the Pipeline that was created and passed into OnConnectedAsync, you can get access to the PipelineReaderWriter and await ReadingStarted.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            var trackDispose = new TrackDispose();
+            var serviceProvider = CreateServiceProvider((s) => s.AddSingleton(trackDispose));
+
+            var endPoint = serviceProvider.GetService<HubEndPoint<TestHub>>();
+
+            using (var connectionWrapper = new ConnectionWrapper())
+            {
+                var endPointTask = Task.Run(() => endPoint.OnConnectedAsync(connectionWrapper.Connection));
+
+                await ((HttpConnection)connectionWrapper.Connection.Channel).Input.ReadingStarted;
```
davidfowl: Expose the HttpConnection directly on connectionWrapper so you can avoid the cast

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                // kill the connection
+                connectionWrapper.Connection.Channel.Dispose();
+
+                await endPointTask;
+
+                Assert.Equal(2, trackDispose.DisposeCount);
```
davidfowl: Why is it 2?

BrennanConroy: `OnConnectedAsync` + `OnDisconnectedAsync`

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
```
mikaelm12: License and copyright header

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    public class HubEndpointTests
+    {
+        [Fact]
+        public async Task HubsAreDisposed()
+        {
+            var trackDispose = new TrackDispose();
+            var serviceProvider = CreateServiceProvider((s) => s.AddSingleton(trackDispose));
```
moozzyk: nit: parens not needed in `(s)`

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            var endPoint = serviceProvider.GetService<HubEndPoint<TestHub>>();
+
+            using (var connectionWrapper = new ConnectionWrapper())
+            {
+                var endPointTask = Task.Run(() => endPoint.OnConnectedAsync(connectionWrapper.Connection));
```
moozzyk: can't it be just var `endPointTask = endPoint.OnConnectedAsync(connectionWrapper.Connection);` ?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
davidfowl: Ah ok.

davidfowl:
Next, this logic needs to move into a hub activator like component.

test/Microsoft.AspNetCore.SignalR.Tests/Microsoft.AspNetCore.SignalR.Tests.xproj
```diff
@@ -0,0 +1,20 @@
...
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">.\obj</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">.\bin\</OutputPath>
+  </PropertyGroup>
+  <PropertyGroup>
+    <SchemaVersion>2.0</SchemaVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
+  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.targets" Condition="'$(VSToolsPath)' != ''" />
```
davidfowl: DNX? What version of VS do you have??

PR #41, opened 2016-11-18
Adding E2E test project

moozzyk:

PR #45, opened 2016-11-22
Removing checked in jasmine

moozzyk:


moozzyk:
This is to address: https://github.com/aspnet/SignalR/pull/41#discussion_r88588262

/cc @davidfowl @anurse

PR #46, opened 2016-11-23
Sync writes to pipeline output 

BrennanConroy:


samples/ChatSample/Hubs/Chat.cs
```diff
@@ -18,7 +18,7 @@ public override Task OnConnectedAsync()
         {
             if (!Context.User.Identity.IsAuthenticated)
             {
-                Context.Connection.Channel.Dispose();
+                //Context.Connection.Channel.Dispose();
```
davidfowl: ?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -274,6 +269,15 @@ public void Dispose()
             _redisServerConnection.Dispose();
         }
 
+        private async Task InvokeAsync(HubConnection connection, byte[] data)
+        {
+            // BAD
+            using (var ms = new MemoryStream(data))
```
davidfowl: get rid of the memory stream

samples/ChatSample/Hubs/Chat.cs
BrennanConroy: `Context.Connection` is now the `HubConnection` so you don't have direct access to the `Channel`

BrennanConroy: That's why I left the comment :)

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public Task WriteAsync(byte[] data)
+        {
+            return Enqueue(() => _connection.Channel.Output.WriteAsync(data));
+        }
+
+        public void Close()
+        {
+            _connection.Channel.Dispose();
```
moozzyk: what happens if there are incomplete tasks in the queue?

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        internal Task Enqueue(Func<Task> taskFactory)
+        {
+            lock (_lock)
+            {
+                return _taskQueue = _taskQueue.ContinueWith((t) => taskFactory());
```
moozzyk: do we need to observe exceptions?

BrennanConroy: I'll keep this in mind while writing tests

BrennanConroy:
Added a couple tests and throw from queued tasks if any previous one failed

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
BrennanConroy: Currently it runs them, but they don't do very much because the channel is closed.

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            _connection.Channel.Dispose();
+        }
+
+        internal Task Enqueue(Func<Task> taskFactory)
+        {
+            // Don't queue tasks if connection is closed
+            if (_connection.Channel.Output.Writing.IsCompleted)
+            {
+                return TaskCache.CompletedTask;
```
BrennanConroy: Should we do something else here?

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubConnection : IClientProxy
+    {
+        private object _lock = new object();
+        private Task _taskQueue = TaskCache.CompletedTask;
```
moozzyk: Does this mean that when not using HubConnection you have to do this on your own? Can't the queue be not coupled to the HubConnection and be reusable?

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            return Enqueue(() => _connection.Channel.Output.WriteAsync(data));
+        }
+
+        public void Close()
+        {
+            _connection.Channel.Dispose();
+        }
+
+        internal Task Enqueue(Func<Task> taskFactory)
```
moozzyk: This feels misplaced - you could enqueue anything here but this is not what the hubConnection is for...

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            // Don't queue tasks if connection is closed
+            if (_connection.Channel.Output.Writing.IsCompleted)
+            {
+                return TaskCache.CompletedTask;
+            }
+
+            lock (_lock)
+            {
+                return _taskQueue = _taskQueue.ContinueWith((t) =>
```
moozzyk: nit: brackets not needed.

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    if (t.IsFaulted)
```
moozzyk: what if task is cancelled?

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionTests.cs
```diff
@@ -0,0 +1,126 @@
...
+                    Assert.True(false);
+                    return TaskCache.CompletedTask;
+                });
+
+                var exceptions = 0;
+                try
+                {
+                    await throwTask;
+                }
+                catch (Exception)
```
moozzyk: you don't know what you catch really

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        public async Task QueuedTaskThrows_EndsOtherQueuedTasks()
+        {
+            using (var connectionWrapper = new TestHelpers.ConnectionWrapper())
+            {
+                var serviceProvider = TestHelpers.CreateServiceProvider();
+                var hubConnection = new HubConnection(connectionWrapper.Connection, serviceProvider.GetService<InvocationAdapterRegistry>());
+
+                var throwTask = hubConnection.Enqueue(() =>
+                {
+                    throw new Exception();
```
moozzyk: don't throw Exception

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+                var tasks = new List<Task>(5);
+                var events = new List<ManualResetEvent>(5);
+                for (var i = 0; i < 5; i++)
+                {
+                    events.Add(new ManualResetEvent(false));
+                    int captureIndex = i;
+                    tasks.Add(hubConnection.Enqueue(() =>
+                    {
+                        int index = captureIndex;
```
moozzyk: why don't you just append indexes to a list and then check if items in the list are the same as the index? Seems much simpler than using ManualResetEvent...

src/Microsoft.AspNetCore.SignalR/HubConnection.cs
```diff
@@ -0,0 +1,79 @@
...
+        public HubConnection(Connection connection, InvocationAdapterRegistry registry)
+        {
+            _connection = connection;
+            _registry = registry;
+            Stream = _connection.Channel.GetStream();
+        }
+
+        public Task InvokeAsync(string method, params object[] args)
+        {
+            var invocationAdapter = _registry.GetInvocationAdapter(_connection.Metadata.Get<string>("formatType"));
```
BrennanConroy: What if the format for the connection changes?

moozzyk: I thought we decided not to support changing formatTypes after the connection has been established.

davidfowl:
Closing In favor of https://github.com/aspnet/SignalR/pull/110/files

PR #47, opened 2016-11-24
DI enabled Microsoft.AspNetCore.Sockets

davidfowl:
- Added AddSockets extension method to IServiceCollection
- Inject IApplicationLifetime into ConnectionManager to handle
graceful shutdown and added test.
- Call AddSockets from AddSignalR

### DIFF
--- a/samples/SocketsSample/Startup.cs
+++ b/samples/SocketsSample/Startup.cs
@@ -21,13 +21,12 @@ namespace SocketsSample
 
             services.AddSockets();
 
-            services.AddSignalR()
-                    .AddSignalROptions(options =>
+            services.AddSignalR(options =>
                     {
                         options.RegisterInvocationAdapter<ProtobufInvocationAdapter>("protobuf");
                         options.RegisterInvocationAdapter<LineInvocationAdapter>("line");
                     });
-                 // .AddRedis();
+            // .AddRedis();
 
             services.AddSingleton<ChatEndPoint>();
             services.AddSingleton<ProtobufSerializer>();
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
```diff
@@ -8,10 +8,11 @@
 
 namespace Microsoft.Extensions.DependencyInjection
 {
-    public static class DependencyInjectionExtensions
+    public static class SignalRDependencyInjectionExtensions
     {
         public static ISignalRBuilder AddSignalR(this IServiceCollection services)
```
BrennanConroy: While you're here should we make this more like other projects and not have an `ISignalRBuilder` and make an overload that accepts `Action<SignalROptions>`

samples/SocketsSample/Startup.cs
```diff
@@ -16,18 +16,17 @@ public class Startup
...
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddRouting();
-
             services.AddSingleton<ProtobufInvocationAdapter>();
             services.AddSingleton<LineInvocationAdapter>();
 
-            services.AddSignalR()
-                    .AddSignalROptions(options =>
+            services.AddSockets();
```
moozzyk: `AddSignalR` adds sockets so this line seems unneeded.

davidfowl: I wanted to add it explicitly anyways.

moozzyk: Why? Do you expect people do this? Because they will if they see it in a sample...

davidfowl: Yes especially if they need to configure options for sockets (which there will inevitably be).

moozzyk: I see.

PR #52, opened 2016-11-28
Added Some Tests

mikaelm12:
Some initial testing for the .NET client
@anurse @BrennanConroy @moozzyk @davidfowl

PR #58, opened 2016-12-02
Generic param overloads

mikaelm12:
Added overloads so that users can use up to 7 parameters.
I'll rebase and add tests when https://github.com/aspnet/SignalR/pull/52 is merged into the net-client branch.
@anurse @moozzyk @davidfowl @BrennanConroy

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
@@ -47,6 +47,41 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
         }
 
+        public void On<T1>(string methodName, T1 parameter1, Action<object[]> handler)
+        {
+            On(methodName, new[] { typeof(T1) }, handler);
+        }
+
+        public void On<T1,T2>(string methodName, T1 parameter1, T2 parameter2, Action<object[]> handler)
+        {
+            On(methodName, new[] { typeof(T1), typeof(T2) }, handler);
+        }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -47,6 +47,41 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
             _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
         }
 
+        public void On<T1>(string methodName, T1 parameter1, Action<object[]> handler)
```
moozzyk: I don't think these should be methods on this class. Rather I'd prefer these to be extension methods.

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
```diff
@@ -47,6 +47,41 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
+        {
+            On(methodName, new[] { typeof(T1) }, handler);
+        }
+
+        public void On<T1,T2>(string methodName, T1 parameter1, T2 parameter2, Action<object[]> handler)
```
moozzyk: nit: space between T1 and T2

src/Microsoft.AspNetCore.SignalR.Client/RpcConnection.cs
moozzyk: Previously they were extensions on IHubProxy - https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Client/HubProxyExtensions.cs

davidfowl:
@mikaelm12 we shouldn't be doing PRs on top of non master branches. I'd like to get the basic client in with some tests then build on that.

mikaelm12:
Cool. Noted.

PR #61, opened 2016-12-06
Pawelka/e2etests

moozzyk:
Adding basic end-to-end tests for the ts/js client.

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
@@ -1,4 +1,7 @@
-﻿using Microsoft.AspNetCore.Builder;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
@@ -25,6 +28,10 @@ namespace Microsoft.AspNetCore.SignalR.Test.Server
 
             app.UseStaticFiles();
             app.UseSockets(options => options.MapEndpoint<EchoEndPoint>("/echo"));
+            app.UseSignalR(routes =>
+            {
+                routes.MapHub<TestHub>("/testhub");
+            });
         }
     }
 }
--- a/test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
@@ -9,7 +9,8 @@ namespace Microsoft.AspNetCore.SignalR.Test.Server
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddRouting();
+            services.AddSockets();
+            services.AddSignalR();
             services.AddSingleton<EchoEndPoint>();
         }
 
--- a/test/Microsoft.AspNetCore.SignalR.Test.Server/TestHub.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Test.Server/TestHub.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.SignalR.Test.Server
+{
+    public class TestHub : Hub
+    {
+        public string Echo(string message)
+        {
+            return message;
+        }
+
+        public void ThrowException(string message)
+        {
+            throw new InvalidOperationException(message);
+        }
+
+        public Task InvokeWithString(string message)
+        {
+            return Clients.Client(Context.Connection.ConnectionId).InvokeAsync("Message", message);
+        }
+    }
+}
--- a/test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/connectionTests.js
+++ b/test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/connectionTests.js
@@ -1,0 +1,31 @@
+
+describe('connection', () => {
+    eachTransport(transportName => {
+        it(`over ${transportName} can send and receive messages`, done => {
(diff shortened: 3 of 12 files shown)


test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
```diff
@@ -9,7 +9,8 @@ public class Startup
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddRouting();
```
mikaelm12: Why did you remove the AddRouting call?

davidfowl: It's part of AddSockets and AddSignalR

mikaelm12: Ah okay. Looks like it was added to AddSockets here https://github.com/aspnet/SignalR/commit/1b59fc6f804e4576e704e6c6f8f9aa90b79ab64d

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
```diff
@@ -0,0 +1,79 @@
...
+
+            hubConnection.start(transportName)
+                .then(() => {
+                    hubConnection.invoke('Microsoft.AspNetCore.SignalR.Test.Server.TestHub.ThrowException', errorMessage)
+                        .then(() => {
+                            // exception expected but none thrown
+                            expect(true).toBe(false);
+                        })
+                        .catch(e => {
+                            expect(e.message).toBe(errorMessage);
```
mikaelm12: Do we want to (Can we?) check the exception type here?

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
```diff
@@ -0,0 +1,79 @@
...
+        it(`over ${transportName} rethrows an exception from the server`, done => {
+            const errorMessage = "An error occurred.";
+            let hubConnection = new signalR.HubConnection(TESTHUBENDPOINT_URL, 'formatType=json&format=text');
+
+            hubConnection.start(transportName)
+                .then(() => {
+                    hubConnection.invoke('Microsoft.AspNetCore.SignalR.Test.Server.TestHub.ThrowException', errorMessage)
+                        .then(() => {
+                            // exception expected but none thrown
+                            expect(true).toBe(false);
```
mikaelm12: nit: Not sure if this was intentional/ has a different meaning but in the test failure case above you say `expect(false).toBe(true) ` and here it is reversed.

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
```diff
@@ -0,0 +1,79 @@
...
+                })
+        });
+
+        it(`over ${transportName} rethrows an exception from the server`, done => {
+            const errorMessage = "An error occurred.";
+            let hubConnection = new signalR.HubConnection(TESTHUBENDPOINT_URL, 'formatType=json&format=text');
+
+            hubConnection.start(transportName)
+                .then(() => {
+                    hubConnection.invoke('Microsoft.AspNetCore.SignalR.Test.Server.TestHub.ThrowException', errorMessage)
```
mikaelm12: Does this now just get changed to "ThrowException" since https://github.com/aspnet/SignalR/pull/63 was merged?

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
```diff
@@ -0,0 +1,79 @@
...
+
+            client.on("Message", msg => {
+                expect(msg).toBe(message);
+                client.stop();
+                done();
+            });
+
+            client.start(transportName)
+                .then(() => {
+                    client.invoke('Microsoft.AspNetCore.SignalR.Test.Server.TestHub.InvokeWithString', message)
```
mikaelm12: Same question here. Should the first param just become "InvokeWithString" ?

test/Microsoft.AspNetCore.SignalR.Test.Server/TestHub.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
mikaelm12: Copyright and license header here

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/connectionTests.js
```diff
@@ -0,0 +1,31 @@
+
```
mikaelm12: Remove leading newline

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
moozzyk: Yes, I had made the changes locally before #63 was merged to actually test #63. I will rebase and update.

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
moozzyk: jasmine does not have `fail` so for the purpose of testing both are equivalent as the mean to fail the test. I think I could just create a fail() function and replace all these.

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
moozzyk: Not really. Exception types from one language don't translate well to exception types in another language and trying to force it would be weird. So we just throw `Error` in javascript. We also don't even send any exception details at the moment (stacktrace and maybe even the exception type might be useful for debugging/tracing)

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
moozzyk: As per the comment above - will rebase on dev and update

PR #72, opened 2016-12-13
Adding more tests

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -59,6 +161,25 @@ private class TrackDispose
             public int DisposeCount = 0;
         }
 
+        public async Task SendRequest(HttpConnection connection, IInvocationAdapter writer, string method, params object[] args)
+        {
+            if (connection == null)
+            {
+                throw new Exception();
```
mikaelm12: nit: This should be an ArgumentNullException

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +38,106 @@ public class HubEndpointTests
...
+        private class MethodHub : Hub
+        {
+            public Task BroadcastMethod(string message)
+            {
+                return Clients.All.InvokeAsync("Broadcast", message);
+            }
+
+            public Task<int> TaskValueMethod()
+            {
+                return Task.FromResult(10);
```
mikaelm12: Boo magic numbers lol

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +38,106 @@ public class HubEndpointTests
             }
         }
 
+        [Fact]
+        public async Task CanReturnValuesFromMethodsOnHub()
```
mikaelm12: Should this be split up into multiple tests? One for each of the HubMethods you are checking

BrennanConroy:
@moozzyk @mikaelm12 @anurse @davidfowl

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -59,6 +354,32 @@ private class TrackDispose
...
         }
 
+        public async Task SendRequest(HttpConnection connection, IInvocationAdapter writer, string method, params object[] args)
+        {
+            if (connection == null)
+            {
+                throw new ArgumentNullException();
+            }
+
+            var stream = new MemoryStream();
```
moozzyk: Can't you just write directly to connection.Input.GetStream() ?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -59,6 +354,32 @@ private class TrackDispose
...
+                Arguments = args,
+                Method = method
+            }, stream);
+
+            var buffer = connection.Input.Alloc();
+            buffer.Write(stream.ToArray());
+            await buffer.FlushAsync();
+        }
+
+        public async Task SendRequest_IgnoreWrite(HttpConnection connection, IInvocationAdapter writer, string method, params object[] args)
```
moozzyk: IgnoreReceive?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+                // kill the connections
+                firstConnection.Connection.Channel.Dispose();
+                secondConnection.Connection.Channel.Dispose();
+
+                await firstEndPointTask;
+                await secondEndPointTask;
+            }
+        }
+
+        private async Task<T> ReadConnectionOutputAsync<T>(HttpConnection connection, JsonSerializer serializer)
```
moozzyk: nit: group with `SendRequest`?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+
+            using (var connectionWrapper = new ConnectionWrapper())
+            {
+                var endPointTask = endPoint.OnConnectedAsync(connectionWrapper.Connection);
+
+                await connectionWrapper.HttpConnection.Input.ReadingStarted;
+
+                var invocationAdapter = serviceProvider.GetService<InvocationAdapterRegistry>();
+                var writer = invocationAdapter.GetInvocationAdapter("json");
+                var serializer = new JsonSerializer();
```
moozzyk: Can't you create this serializer in the ReadConnectionOutputAsync? Each test is polluted with this serializer even though it it not being used outside of ReadConnectionOutputAsync. You can't also really pass anything else to ReadConnectionOutputAsync.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+
+                // return Task<int>
+                await SendRequest(connectionWrapper.HttpConnection, writer, "TaskValueMethod");
+                var res = await ReadConnectionOutputAsync<InvocationResultDescriptor>(connectionWrapper.HttpConnection, serializer);
+                Assert.Equal(10.ToString(), res.Result.ToString());
```
moozzyk: `ToString`? Compare actual values

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
moozzyk: ...especially that it should be 42

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+                connectionWrapper.Connection.Channel.Dispose();
+
+                await endPointTask;
+            }
+        }
+
+        [Fact]
+        public async Task HubMethodCanReturnValue()
+        {
+            var serviceProvider = CreateServiceProvider();
```
moozzyk: same comments as above

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+                Assert.Equal("Broadcast", res.Method);
+                Assert.Equal(1, res.Arguments.Length);
+                Assert.Equal("test", res.Arguments[0]);
+
+                // kill the connections
+                firstConnection.Connection.Channel.Dispose();
+                secondConnection.Connection.Channel.Dispose();
+
+                await firstEndPointTask;
+                await secondEndPointTask;
```
moozzyk: `Task.WhenAll` ?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+            var endPoint = serviceProvider.GetService<HubEndPoint<MethodHub>>();
+
+            using (var firstConnection = new ConnectionWrapper())
+            using (var secondConnection = new ConnectionWrapper())
+            {
+                var firstEndPointTask = endPoint.OnConnectedAsync(firstConnection.Connection);
+                var secondEndPointTask = endPoint.OnConnectedAsync(secondConnection.Connection);
+
+                await firstConnection.HttpConnection.Input.ReadingStarted;
+                await secondConnection.HttpConnection.Input.ReadingStarted;
```
moozzyk: Task.WhenAll?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+
+                var invocationAdapter = serviceProvider.GetService<InvocationAdapterRegistry>();
+                var writer = invocationAdapter.GetInvocationAdapter("json");
+                var serializer = new JsonSerializer();
+
+                await SendRequest(firstConnection.HttpConnection, writer, "BroadcastMethod", "test");
+
+                var res = await ReadConnectionOutputAsync<InvocationDescriptor>(firstConnection.HttpConnection, serializer);
```
moozzyk: Use Task.WhenAll and loop over the results. will have less duplicate code and will be more robust. Right now because you reuse `res` it can be easily broken and it will be hard to even notice because expected values are the same in both cases.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
moozzyk: Maybe just pass `serviceProvider`?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+
+                await SendRequest_IgnoreWrite(firstConnection.HttpConnection, writer, "GroupSendMethod", "testGroup", "test");
+                // check that 'secondConnection' hasn't received the group send
+                Assert.False(secondConnection.HttpConnection.Output.ReadAsync().IsCompleted);
+
+                await SendRequest_IgnoreWrite(secondConnection.HttpConnection, writer, "GroupAddMethod", "testGroup");
+
+                await SendRequest_IgnoreWrite(firstConnection.HttpConnection, writer, "GroupSendMethod", "testGroup", "test");
+                // check that 'secondConnection' has received the group send
+                var res = await ReadConnectionOutputAsync<InvocationDescriptor>(secondConnection.HttpConnection, serializer);
```
moozzyk: I think you should also check that the firstConnection did not receive the group message since it is not a group member

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+                firstConnection.Connection.Channel.Dispose();
+                secondConnection.Connection.Channel.Dispose();
+
+                await firstEndPointTask;
+                await secondEndPointTask;
+            }
+        }
+
+        [Fact]
+        public async Task RemoveFromGroupWhenNotInGroup()
```
moozzyk: RemoveFromGroupWhenNotInGroupDoesNotFail/Throw? (because there are no Asserts in the test)

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +39,298 @@ public class HubEndpointTests
...
+        private async Task<T> ReadConnectionOutputAsync<T>(HttpConnection connection, JsonSerializer serializer)
+        {
+            var methodResult = await connection.Output.ReadAsync();
+            var res = serializer.Deserialize<T>(new JsonTextReader(new StreamReader(new MemoryStream(methodResult.Buffer.ToArray()))));
+            connection.Output.AdvanceReader(methodResult.Buffer.End, methodResult.Buffer.End);
+
+            return res;
+        }
+
+        private class MethodHub : Hub
```
moozzyk: We need a few more cases here like:
- a `void` method
- methods with more than one parameter
- parameters of different types than just string 
- etc.

It would be good to look at what we did **not** support in the previous version and potentially start with the same list (this would be a separate PR though)

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
BrennanConroy: Nope, Input doesn't have a Stream

BrennanConroy: Yeah, but `connection.Input` is a `PipelineReaderWriter`

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -232,7 +232,7 @@ private void DiscoverHubMethods()
         private static bool IsHubMethod(MethodInfo m)
         {
             // TODO: Add more checks
-            return m.IsPublic;
+            return m.IsPublic && typeof(Hub<IClientProxy>) != m.GetBaseDefinition().DeclaringType;
```
davidfowl: This isn't right. You want to look for an open generic `Hub<>`, also, why do we care about this specifically? Did you look at what old signalr did?

BrennanConroy: We care about this so clients can't call methods like `OnConnectedAsync` and `OnDisconnectedAsync`

BrennanConroy: Old SignalR did it through the IHub interface https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Core/Hubs/ReflectionHelper.cs#L24

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +40,415 @@ public class HubEndpointTests
...
+
+                var invocationAdapter = serviceProvider.GetService<InvocationAdapterRegistry>();
+                var adapter = invocationAdapter.GetInvocationAdapter("json");
+
+                // return Task<int>
+                await SendRequest(connectionWrapper.HttpConnection, adapter, "TaskValueMethod");
+                var res = await ReadConnectionOutputAsync<InvocationResultDescriptor>(connectionWrapper.HttpConnection);
+                Assert.Equal(42L, res.Result);
+
+                // kill the connection
```
moozzyk: oh - it is here to unblock the `endPointTask` right?

BrennanConroy: Yes

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +40,415 @@ public class HubEndpointTests
...
+
+                var invocationAdapter = serviceProvider.GetService<InvocationAdapterRegistry>();
+                var adapter = invocationAdapter.GetInvocationAdapter("json");
+
+                // return int
+                await SendRequest(connectionWrapper.HttpConnection, adapter, "OnDisconnectedAsync");
+
+                await Task.Delay(1000);
+                // kill the connection
+                connectionWrapper.Connection.Channel.Dispose();
```
BrennanConroy: If we ever break the test I'm future proofing against it hanging.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,6 +40,415 @@ public class HubEndpointTests
...
+
+                try
+                {
+                    await endPointTask;
+                    Assert.True(false);
+                }
+                catch (Exception)
+                {
+                    // TODO
```
moozzyk: I will merge this today. I am just thinking about adding one more test.

BrennanConroy:
Rebased and 🆙 📅

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -233,7 +233,19 @@ private void DiscoverHubMethods()
         private static bool IsHubMethod(MethodInfo m)
         {
             // TODO: Add more checks
-            return m.IsPublic && !m.IsSpecialName;
+            if (!m.IsPublic || m.IsSpecialName)
```
mikaelm12: Can we give this var a better name

BrennanConroy:
Update and Rebase

Please Review @moozzyk @mikaelm12 @anurse

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -356,9 +351,9 @@ public override Task OnDisconnectedAsync(Exception exception)
                     await endPointTask;
```
moozzyk: nit: why not `Assert.ThrowsAsync`?

BrennanConroy: I want to check the message to make sure its a method not found exception

src/Microsoft.AspNetCore.Sockets/Internal/PipelineConnection.cs
```diff
@@ -28,4 +28,4 @@ public void Dispose()
             Output.CompleteWriter();
         }
     }
-}
+}
```
davidfowl: revert?

Issue #62, opened 2016-12-06
Catch exceptions from formatters


moozzyk:
Since formatters are pluggable they can fail in unexpected ways bringing the server down e.g:

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.SignalR.JsonNetInvocationAdapter.<>c__DisplayClass2_0.<ReadInvocationDescriptorAsync>b__0() i
n C:\source\SignalR-Core\src\Microsoft.AspNetCore.SignalR\JsonNetInvocationAdapter.cs:line 40
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Microsoft.AspNetCore.SignalR.HubEndPoint`2.<DispatchMessagesAsync>d__9.MoveNext() in C:\source\SignalR-Core\src\Mi
crosoft.AspNetCore.SignalR\HubEndPoint.cs:line 105
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Microsoft.AspNetCore.SignalR.HubEndPoint`2.<OnConnectedAsync>d__8.MoveNext() in C:\source\SignalR-Core\src\Microso
ft.AspNetCore.SignalR\HubEndPoint.cs:line 77
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.AspNetCore.SignalR.HubEndPoint`2.<OnConnectedAsync>d__8.MoveNext() in C:\source\SignalR-Core\src\Microso
ft.AspNetCore.SignalR\HubEndPoint.cs:line 95
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #77, opened 2016-12-21
Handling exceptions thrown on the server side

moozzyk:
Addresses: #62

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/Hub.cs
+++ b/src/Microsoft.AspNetCore.SignalR/Hub.cs
@@ -66,7 +66,7 @@ namespace Microsoft.AspNetCore.SignalR
             return TaskCache.CompletedTask;
         }
 
-        public virtual Task OnDisconnectedAsync()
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.SignalR/Hub.cs
```diff
@@ -66,7 +66,7 @@ public virtual Task OnConnectedAsync()
             return TaskCache.CompletedTask;
         }
 
-        public virtual Task OnDisconnectedAsync()
+        public virtual Task OnDisconnectedAsync(Exception exception)
```
davidfowl: Hmmm, what would somebody do with this exception besides log it?

moozzyk: This is to let them know that the connection was not closed cleanly.

davidfowl: Meh, I guess it's harmless.

moozzyk:
Added a test.

Issue #57, opened 2016-11-30
"No Channel" exception in long polling


moozzyk:
When running end to end tests I am working on I semi occasionally (25% of times) receive this:

```
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:5000/testhub/getid?formatType=json&format=text
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 0.2851ms 200
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 POST http://localhost:5000/testhub/send?id=055127ed-e4f0-421d-96f7-eaf32a251ced&055127ed-e4f0-421d-96f7-eaf32a251ced application/x-www-form-urlencoded 120
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:5000/testhub/poll?id=055127ed-e4f0-421d-96f7-eaf32a251ced&055127ed-e4f0-421d-96f7-eaf32a251ced
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HL0PDKEHGN48": An unhandled exception was thrown by the application.
System.InvalidOperationException: No channel
   at Microsoft.AspNetCore.Sockets.HttpConnectionDispatcher.ProcessSend(HttpContext context) in C:\source\SignalR-Core\src\Microsoft.AspNetCore.Sockets\HttpConnectionDispatcher.cs:line 214
   at Microsoft.AspNetCore.Sockets.HttpConnectionDispatcher.<ExecuteAsync>d__4`1.MoveNext() in C:\source\SignalR-Core\src\Microsoft.AspNetCore.Sockets\HttpConnectionDispatcher.cs:line 36
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Builder.RouterMiddleware.<Invoke>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #78, opened 2016-12-21
 #courage #prototype

moozzyk:
Fixing a bug where the server would throw if send request is received before the first poll or sse request (#57). We will now delay send until the first poll or sse is received or reject the send request if poll or sse not received within a timeout


davidfowl:
We should use the existing background timer in connection manager.

PR #79, opened 2016-12-21
Fixing Promise related compiler errors

moozzyk:
We don't expect any content for `ITransport.send` methods so the `.ITransportsend` method just returns `Promise<void>`. However LongPolling and ServerSentEvents transport use the standard HttpClient to send send requests whose return type is `Promise<string>`. This causes:

error TS2322: Type `Promise<string>` is not assignable to type `Promise<void>`.


moozzyk:
Obsoleted by 3a01d6cff1a89c933ad60eccd67aac2f2ebe0102

PR #80, opened 2016-12-22
Introducing async/await in TS client + minor clean up

moozzyk:


src/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -28,13 +28,13 @@ export class HubConnection {
             thisHubConnection.dataReceived(data);
         };
 
-        this.callbacks = new Map<string, (any) => void>();
+        this.callbacks = new Map<string, (InvocationResultDescriptor) => void>();
         this.methods = new Map<string, (...args:any[]) => void>();
```
BrennanConroy: nit: `...args: any`

moozzyk:
I keep forgetting it to add to my list. Another reason for reviewing http client is enabling passing your own implementation to allow mocking. 
Updated: https://github.com/aspnet/SignalR/issues/3

PR #82, opened 2016-12-27
Fixing send/poll race

moozzyk:
Fixing a bug where the server would throw if send request is received before the first poll or sse request (#57). We will now delay send until the first poll or sse is received or reject the send request if poll or sse not received within a timeout
Also making the connection not to be marked as Active upon reserving a connection. This enables cleaning up connections that were reserved but never used.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/InvocationMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/InvocationMessage.cs
@@ -1,11 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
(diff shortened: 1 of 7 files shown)


davidfowl:
Why not always add the channel? We don't need the new TCS since we're doing to go this route.

moozzyk:
I can try that as well

BrennanConroy:
One problem with creating it on `/getid` is that `formattype` won't be set if `/send` is called before `/poll`

BrennanConroy:
Looks like there is a race, if `/send` is waiting and `/poll` sets the `tcs` then `/send` can continue and do things before `/poll` sets the `formattype`

davidfowl:
@BrennanConroy maybe we need to allow setting the format type as early as needed.

moozzyk:
Because you don't know when it is it would just mean you always need to send it.

moozzyk:
This will be solved differently (i.e. by specifying the format during negotiation)

PR #83, opened 2016-12-27
Introducing HubActivator

moozzyk:


test/Microsoft.AspNetCore.SignalR.Tests/HubActivatorTests.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
```
mikaelm12: Copyright and license header here

src/Microsoft.AspNetCore.SignalR/IHubActivator.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public interface IHubActivator
+    {
+        HubWrapper<THub, TClient> Create<THub, TClient>(IServiceProvider provider)
```
mikaelm12: It might be a little confusing that `Create` here creates a HubWrapper and not a Hub directly. Maybe think about renaming.

src/Microsoft.AspNetCore.SignalR/HubWrapper.cs
```diff
@@ -0,0 +1,33 @@
...
+        private readonly THub _hub;
+        private readonly bool _shouldDispose;
+
+        public HubWrapper(THub hub, bool shouldDispose)
+        {
+            _hub = hub;
+            _shouldDispose = shouldDispose;
+        }
+
+        public static explicit operator THub(HubWrapper<THub, TClient> wrapper)
```
mikaelm12: Is this the best way to expose the Hub? Why use a cast here?

mikaelm12: Really asking for my own benefit 😄

moozzyk: This is actually the explicit conversion operator. It could be just a property as well...

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -62,16 +66,11 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
                 await _lifetimeManager.OnConnectedAsync(connection);
 
                 using (var scope = _serviceScopeFactory.CreateScope())
+                using (var hubWrapper = _hubActivator.Create<THub, TClient>(scope.ServiceProvider))
```
davidfowl: This doesn't look right. Why do we need to pass the service provider? We should be looking at MVC and copying their patterns where possible.

src/Microsoft.AspNetCore.SignalR/IHubActivator.cs
moozzyk: How would you track if the instance needs to be disposed?

moozzyk: Bringing back `IHub` would allow removing the wrapper.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
moozzyk: because it we created a new scope. I will check what they do in MVC

moozzyk: In MVC they either pass the serviceProvider: 
https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/TypeFilterAttribute.cs#L66-L80
https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/Internal/TypeActivatorCache.cs#L23-L37
or they use requestServices from HttpContext which are request scoped:
https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinder.cs#L51-L52

moozzyk:
:up::date: - added the `IHub` interface. It hides the wrapper/Dispose but the returned object can't be used to invoke a hub method directly so the interface needs to have the Invoke method. Not sure if this is much better (unless there will be other usages of `IHub` interface)

src/Microsoft.AspNetCore.SignalR/IHubActivator.cs
davidfowl: Lets do that then.

moozzyk: It's already done - I updated the PR

davidfowl:
This still doesn't look right. The `HubWrapper` and passing the service provider look incorrect. The `IHubActivator` should be resolved with the right scope. I'm going to take a deeper look at the layer before suggesting something.

src/Microsoft.AspNetCore.SignalR/DefaultHubActivator.cs
```diff
@@ -0,0 +1,51 @@
...
+        public void Release(THub hub)
+        {
+            if (hub == null)
+            {
+                throw new ArgumentNullException(nameof(hub));
+            }
+
+            Debug.Assert(_created.HasValue, "hubs must be released with the hub activator they were created");
+
+            if ((bool)_created)
```
davidfowl: _created.Value

moozzyk: Sure

BrennanConroy:
@moozzyk I think you're missing the `DefaultHubActivator` file

moozzyk:
@BrennanConroy - funnily @davidfowl [commented](https://github.com/aspnet/SignalR/pull/83#pullrequestreview-14725088) on it... (but yes it seems to be missing)

PR #84, opened 2016-12-30
Sorting out exceptions

moozzyk:
There were a few places where exceptions could be left not handled which could bring the server down. Also making sure that OnConnected/OnDisconnected events are invoked correctly (e.g. if invoking OnDisconnectedAsync on hub threw we would not call OnDisconnectedAsync on lifetime manager and therefore we would continue to use/track connections that were already closed)

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
@@ -56,11 +56,40 @@ namespace Microsoft.AspNetCore.SignalR
         {
             // TODO: Dispatch from the caller
             await Task.Yield();
-            Exception exception = null;
+
             try
             {
                 await _lifetimeManager.OnConnectedAsync(connection);
+                await RunHubAsync(connection);
+            }
+            finally
+            {
+                await _lifetimeManager.OnDisconnectedAsync(connection);
+            }
+        }
 
+        private async Task RunHubAsync(Connection connection)
+        {
+            await HubOnConnectedAsync(connection);
+
+            try
+            {
+                await DispatchMessagesAsync(connection);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(0, ex, "Error when processing requests.");
+                await HubOnDisconnectedAsync(connection, ex);
+                throw;
+            }
+
+            await HubOnDisconnectedAsync(connection, null);
+        }
+
+        private async Task HubOnConnectedAsync(Connection connection)
+        {
+            try
+            {
                 using (var scope = _serviceScopeFactory.CreateScope())
                 {
                     var hubActivator = scope.ServiceProvider.GetRequiredService<IHubActivator<THub, TClient>>();
@@ -75,17 +104,17 @@ namespace Microsoft.AspNetCore.SignalR
                         hubActivator.Release(hub);
                     }
                 }
-
-                await DispatchMessagesAsync(connection);
             }
             catch (Exception ex)
             {
-                _logger.LogError(0, ex, "Error when processing requests.");
-                exception = ex;
-                connection.Channel.Input.Complete(exception);
-                connection.Channel.Output.Complete(exception);
+                _logger.LogError(0, ex, "Error when invoking OnConnectedAsync on hub.");
+                throw;
             }
-            finally
+        }
+
+        private async Task HubOnDisconnectedAsync(Connection connection, Exception exception)
+        {
+            try
             {
                 using (var scope = _serviceScopeFactory.CreateScope())
                 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -75,17 +116,16 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
                         hubActivator.Release(hub);
                     }
                 }
-
-                await DispatchMessagesAsync(connection);
             }
-            catch (Exception ex)
+            catch(Exception ex)
```
moozzyk: Now I actually think it should terminate the connection.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -56,11 +56,52 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
+            await HubOnConnectedAsync(connection);
+
+            try
+            {
+                await DispatchMessagesAsync(connection);
+            }
+            catch(Exception ex)
+            {
+                _logger.LogError(0, ex, "Error when processing requests.");
+                connection.Channel.Input.Complete(ex);
```
davidfowl: I think we need to let errors bubble out to the middleware and handle closing the connection here

moozzyk: Then something is not right because when we let the errors bubble up they crashed the entire server - #62

davidfowl: It shouldn't if things are properly returning faulted tasks.

moozzyk:
:up::date: - exceptions will now propagate up the stack

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -56,11 +56,46 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
+
             try
             {
                 await _lifetimeManager.OnConnectedAsync(connection);
+                await RunHubAsync(connection);
+            }
+            catch (Exception ex)
+            {
+                connection.Channel.Input.Complete(ex);
+                connection.Channel.Output.Complete(ex);
```
davidfowl: Only close the input.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -56,11 +56,46 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
             // TODO: Dispatch from the caller
             await Task.Yield();
-            Exception exception = null;
+
             try
             {
                 await _lifetimeManager.OnConnectedAsync(connection);
+                await RunHubAsync(connection);
+            }
+            catch (Exception ex)
```
davidfowl: remove this

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -101,8 +136,11 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
                         hubActivator.Release(hub);
                     }
                 }
-
-                await _lifetimeManager.OnDisconnectedAsync(connection);
+            }
+            catch(Exception ex)
```
davidfowl: nit: spacing

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -64,18 +63,127 @@ public class HubEndpointTests
...
+                services.AddSingleton(mockLifetimeManager.Object);
+                services.AddSingleton(mockHubActivator.Object);
+            });
+
+            var endPoint = serviceProvider.GetService<HubEndPoint<Hub>>();
+
+            using (var connectionWrapper = new ConnectionWrapper())
+            {
+                Assert.Equal("Lifetime manager OnConnectedAsync failed.",
+                    (await Assert.ThrowsAsync<InvalidOperationException>(
```
davidfowl: this is super hard to read

moozzyk:
This needs addendum due to `async void` possibly [throwing](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs#L116) on a threadpool thread and crashing. (This is why appveyor is failing)

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Nope

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
moozzyk: separated to two asserts.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
moozzyk: Actually it is needed otherwise we will just close a pending poll with 204 in case of exceptions. As discussed I moved it to the endpoint.

moozzyk:
@davidfowl Is this good now?

PR #92, opened 2017-01-05
add "messaging" endpoints and transports

anurse:
* Need a separate set of primitives to handle messaging
* Using Channels (not Pipelines!) to provide the data flow for messaging
* All transports are now "message" based transports
* Added an adaptor to convert message-based transports to serve
streaming endpoints

Side things, required in order to reference Channels:
* Removed `net451` builds of libraries, we're `netstandard1.3` only (means .NET 4.6 is the minimum)
* Changed `net451` builds of apps/tests to `net46`
* Updated to `netcoreapp1.2`


global.json
```diff
@@ -2,8 +2,5 @@
   "projects": [
     "src",
     "test"
-  ],
-  "sdk": {
-    "version": "1.0.0-preview2-1-003180"
```
davidfowl: Fix this

anurse: You want it back? Doesn't it completely bork KoreBuild's behavior of updating to our latest version?

samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
```
mikaelm12: Copyright and license header

global.json
anurse: It looks like every other repo has it, so fine, and we're going to migrate this repo to MSBuild soon anyway :).

samples/SocketsSample/EndPoints/MessagesEndPoint.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
```
mikaelm12: Header needed here too

anurse: Running my tool to add license headers (https://github.com/anurse/aspnet-fmt) update coming.

anurse: Update pushed.

src/Microsoft.AspNetCore.SignalR.Redis/project.json
```diff
@@ -31,7 +31,6 @@
     "StackExchange.Redis.StrongName": "1.1.605"
   },
   "frameworks": {
-    "netstandard1.6": {},
-    "net451": {}
+    "netstandard1.6": {}
```
BrennanConroy: Add net46

samples/SocketsSample/EndPoints/MessagesEndPoint.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets;
+
+namespace SocketsSample.EndPoints
+{
+    public class MessagesEndPoint : MessagingEndPoint
+    {
+        public ConnectionList Connections { get; } = new ConnectionList();
```
anurse: Yeah, I don't disagree. I was wondering if there was ever a purpose for a heterogeneous collection of connections, but I don't know that there is.

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
```diff
@@ -140,43 +152,43 @@ private void LogFrame(string action, WebSocketFrame frame)
...
-                    {
-                        _channel.Output.Advance(buffer.End);
                     }
                 }
-            }
-            finally
-            {
-                // No longer reading from the channel
-                _channel.Output.CompleteReader();
+                catch (Exception ex) when (ex.GetType().IsNested && ex.GetType().DeclaringType == typeof(Channel))
```
anurse: https://github.com/dotnet/corefxlab/issues/1068

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -202,58 +227,54 @@ private Task ProcessSend(HttpContext context)
...
+                        buffer = strm.ToArray();
+                    }
 
-                return context.Request.Body.CopyToAsync(httpChannel.Input);
+                    var format =
+                        string.Equals(context.Request.Query["format"], "binary", StringComparison.OrdinalIgnoreCase)
+                            ? Format.Binary
+                            : Format.Text;
+                    var message = new Message(
+                        ReadableBuffer.Create(buffer).Preserve(),
```
anurse: I agree, but we don't have a Pipeline to alloc from, since we're dealing with Messaging transports. This leads to the desire to allow allocating PreservedBuffers directly out of the buffer pool.

src/Microsoft.AspNetCore.SignalR/HubLifetimeManager.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public abstract class HubLifetimeManager<THub>
```
anurse: Yes

samples/SocialWeather/project.json
```diff
@@ -17,7 +17,7 @@
     "Microsoft.AspNetCore.Server.IISIntegration.Tools": "1.0.0-preview2-final"
   },
   "frameworks": {
-    "netcoreapp1.1": {}
+    "netcoreapp1.2": { }
```
anurse: Here, we can downgrade, but in other projects (ChatSample) some of the packages are pulling the NetCore.App package up to 1.2. It seems odd to have the package 1.2 and the TFM `netcoreapp1.1` so I just updated all to be consistent.

test/Microsoft.AspNetCore.SignalR.Tests/project.json
```diff
@@ -14,15 +14,14 @@
...
+    "netcoreapp1.2": {
       "dependencies": {
         "Microsoft.NETCore.App": {
           "version": "1.2.0-*",
           "type": "platform"
         }
       }
-    },
-    "net451": { }
+    }
```
BrennanConroy: net46?

test/Microsoft.AspNetCore.Sockets.Tests/project.json
```diff
@@ -16,15 +16,14 @@
...
+    "netcoreapp1.2": {
       "dependencies": {
         "Microsoft.NETCore.App": {
           "version": "1.2.0-*",
           "type": "platform"
         }
       }
-    },
-    "net451": {}
+    }
```
BrennanConroy: net46

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
anurse: Fixed in https://github.com/dotnet/corefxlab/pull/1097, just waiting for it to propagate

anurse:
Blocked by Coherence-Signed#395 in Hosting. PR to Hosting pending.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/project.json
```diff
@@ -23,7 +23,8 @@
         }
       }
     },
-    "net451": {}
+    "net46": {
```
BrennanConroy: nit: format

test/Microsoft.AspNetCore.SignalR.Test.Server/project.json
```diff
@@ -15,7 +11,16 @@
...
-    "netcoreapp1.1": {}
+    "netcoreapp1.1": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "version": "1.2.0-*",
+          "type": "platform"
+        }
+      }
+    },
+    "net46": {
```
BrennanConroy: nit: format

src/Microsoft.AspNetCore.Sockets/EndPoint.cs
```diff
@@ -8,13 +8,22 @@ namespace Microsoft.AspNetCore.Sockets
...
     /// Represents an end point that multiple connections connect to. For HTTP, endpoints are URLs, for non HTTP it can be a TCP listener (or similar)
     /// </summary>
+    // REVIEW: This doesn't have any members any more... marker interface? Still even necessary?
     public abstract class EndPoint
     {
         /// <summary>
+        /// Gets the connection mode supported by this endpoint.
+        /// </summary>
+        /// <remarks>
+        /// This usually maps directly to which of <see cref="MessagingEndPoint"/> or <see cref="StreamingEndPoint"/> the end point subclasses.
```
moozzyk: nit: remove `usually` or add a note where it does not map.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -17,171 +20,193 @@ public class HttpConnectionDispatcher
...
+                {
+                    _logger.LogDebug("Resuming existing Long Polling connection: {0}", state.Connection.ConnectionId);
+                }
 
-                        await transportTask;
-                    }
+                var resultTask = await Task.WhenAny(endpointTask, transportTask);
+
+                // Observe any exception
+                resultTask.GetAwaiter().GetResult();
```
moozzyk: I think this is wrong. if the resultTask is a faulted endpointTask you will not release a pending poll. This is why we await endPointTask in the Close (which is actually broken too). 

I have a PR that sorts this out (https://github.com/aspnet/SignalR/pull/84)

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -69,15 +69,15 @@ private Task InvokeAllWhere(string methodName, object[] args, Func<Connection, b
...
+                tasks.Add(invocationAdapter.WriteMessageAsync(message, connection.Transport.GetStream()));
             }
 
             return Task.WhenAll(tasks);
         }
 
         public override Task InvokeConnectionAsync(string connectionId, string methodName, object[] args)
         {
-            var connection = _connections[connectionId];
+            var connection = (StreamingConnection)_connections[connectionId];
```
davidfowl: Why do you need a cast here?

samples/SocialWeather/Protobuf/WeatherReport.cs
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
BrennanConroy: Remove :P

src/Microsoft.AspNetCore.Sockets/Internal/MessagingConnectionState.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.AspNetCore.Sockets.Internal
+{
+    public class MessagingConnectionState : ConnectionState
+    {
+        public new MessagingConnection Connection => (MessagingConnection)base.Connection;
+        public IChannelConnection<Message> Application { get; }
+
+        public MessagingConnectionState(MessagingConnection connection, IChannelConnection<Message> application) : base(connection)
+        {
```
moozzyk: Ignore - I missed it was in the Internal namespace

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -60,7 +60,7 @@ private Task InvokeAllWhere(string methodName, object[] args, Func<Connection, b
             };
 
             // TODO: serialize once per format by providing a different stream?
-            foreach (var connection in _connections)
+            foreach (var connection in _connections.Cast<StreamingConnection>())
```
mikaelm12: This cast can be removed now

src/Microsoft.AspNetCore.Sockets/EndPoint.cs
```diff
@@ -8,13 +8,22 @@ namespace Microsoft.AspNetCore.Sockets
...
+        public abstract ConnectionMode Mode { get; }
+
+        /// <summary>
         /// Called when a new connection is accepted to the endpoint
         /// </summary>
-        /// <param name="connection">The new <see cref="Connection"/></param>
+        /// <param name="connection">The new <see cref="MessagingConnection"/></param>
         /// <returns>A <see cref="Task"/> that represents the connection lifetime. When the task completes, the connection is complete.</returns>
-        public abstract Task OnConnectedAsync(Connection connection);
+        internal protected abstract Task OnConnectedAsync(Connection connection);
```
anurse: It was originally to try and guide users to override the correct one (the one provided by their sub-class, which uses the specific connection type they expect), but it doesn't really prevent that. I'm going to review some options, but for now I'm making it public.

src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public ChannelConnection(IReadableChannel<T> input, IWritableChannel<T> output)
+        {
+            Input = input;
+            Output = output;
+        }
+
+        public void Dispose()
+        {
+            Output.Complete();
```
anurse: There is no `Complete` for `IReadableChannel<T>`

anurse: Unlike Pipelines, Channels don't have a way for a reader to signal they have stopped reading.

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -0,0 +1,66 @@
...
+            {
+                using (var message = await _connection.ReadAsync(_cancellationSource.Token))
+                {
+                    _logger.LogDebug("Writing {0} byte message to response", message.Payload.Buffer.Length);
+                    context.Response.ContentLength = message.Payload.Buffer.Length;
+                    await message.Payload.Buffer.CopyToAsync(context.Response.Body);
+                }
+            }
+            catch (Exception ex) when (ex.GetType().IsNested && ex.GetType().DeclaringType == typeof(Channel))
+            {
```
anurse: It's thrown when there is no message to read because the channel is closed. The exception type was private and nested, but has been made public. This is here to unblock until that change propagates. I'll have an issue filed if I need to merge this without updating to use the public type (due to propagation lag)

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
anurse: I've adapted the code from your PR. Take a look when the update is posted and see if I've got the idea right :).

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -125,12 +125,13 @@ public HttpConnectionDispatcher(ConnectionManager manager, PipelineFactory facto
 
                 var resultTask = await Task.WhenAny(endpointTask, transportTask);
 
-                // Observe any exception
-                resultTask.GetAwaiter().GetResult();
-
                 if (resultTask == endpointTask)
                 {
                     // Notify the long polling transport to end
+                    if (endpointTask.IsFaulted)
```
moozzyk: This looks good to me. 

You may also want to copy bits from the .Close Action. It is basically `async void` and it is being run as a timer action (i.e. on a ThreadPool thread). We await the endPointTask there and if the task is faulted it crashes the process (or leave it and I will do it in my PR)

src/Microsoft.AspNetCore.Sockets/EndPoint.cs
```diff
@@ -15,15 +15,15 @@ public abstract class EndPoint
         /// Gets the connection mode supported by this endpoint.
         /// </summary>
         /// <remarks>
-        /// This usually maps directly to which of <see cref="MessagingEndPoint"/> or <see cref="StreamingEndPoint"/> the end point subclasses.
+        /// This maps directly to which of <see cref="MessagingEndPoint"/> or <see cref="StreamingEndPoint"/> the end point subclasses.
```
moozzyk: nit: I think I saw one more place like this.

anurse: I don't see anywhere else.

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
davidfowl: Is it public yet

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.FunctionalTests
     public class HubConnectionTests : IDisposable
     {
         private readonly TestServer _testServer;
+        private static readonly bool _verbose = string.Equals(Environment.GetEnvironmentVariable("SIGNALR_TEST_VERBOSE"), "1");
```
moozzyk: /cc @mikaelm12 - you are adding something similar in your change

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -220,9 +220,9 @@ public override Task OnDisconnectedAsync(Connection connection)
                     await previousTask;
 
                     var tasks = new List<Task>(group.Connections.Count);
-                    foreach (var groupConnection in group.Connections)
+                    foreach (var groupConnection in group.Connections.Cast<StreamingConnection>())
```
davidfowl: Should't be.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -69,6 +79,8 @@ public HubConnectionTests()
                 {
                     //TODO: Get rid of this. This is to prevent "No channel" failures due to sends occuring before the first poll.
                     await Task.Delay(500);
+                    EnsureConnectionEstablished(connection);
```
moozzyk: why do we need this? awaiting ConnectAsync should be enough to ensure the connection was started

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -25,11 +26,18 @@ public class HubConnection : IDisposable
...
+
+        // We need to ensure pending calls added after a connection failure don't hang. Right now the easiest thing to do is lock.
+        private readonly object _pendingCallsLock = new object();
+        private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>();
+
         private readonly ConcurrentDictionary<string, InvocationHandler> _handlers = new ConcurrentDictionary<string, InvocationHandler>();
 
         private int _nextId = 0;
 
+        public Task Completion { get; }
```
moozzyk: I have doubts about exposing this. We probably should have OnConnectionClosed event or something to that effect. (Not necessarily in this PR but if you remove EnsureConnectionEstablished you can remove this as well for now)

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -117,45 +130,69 @@ public void Dispose()
...
+                    else
                     {
-                        DispatchInvocationResult(invocationResultDescriptor, cancellationToken);
+                        var invocationResultDescriptor = message as InvocationResultDescriptor;
+                        if (invocationResultDescriptor != null)
+                        {
+                            InvocationRequest irq;
+                            lock (_pendingCallsLock)
+                            {
+                                _connectionActive.Token.ThrowIfCancellationRequested();
```
moozzyk: what happens to this exception after it is thrown? because I don't see anyone awaiting the _reader task... I wonder if it shouldn't be just `break;`

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -117,45 +130,69 @@ public void Dispose()
...
+                                irq = _pendingCalls[invocationResultDescriptor.Id];
+                                _pendingCalls.Remove(invocationResultDescriptor.Id);
+                            }
+                            DispatchInvocationResult(invocationResultDescriptor, irq, cancellationToken);
+                        }
                     }
                 }
             }
-            _logger.LogTrace("Ending receive loop");
+            finally
```
moozzyk: Do we need to handle exceptions here and shutdown the connection? (not necessarily a part of this PR if we do)

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
anurse: Because the first poll could fail, which would trigger the completion, but that would happen after ConnectAsync.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: Yeah, this is a pattern that is developing in other places, we need to decide how we want to express it. You can easily implement an OnConnectionClosed event handler with ContinueWith.

anurse: This is basically how Channels and Pipelines work today.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: We could do that, sure. We'd still want the finally though in case a random exception brings the reader down.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: We should review and improve exception handling for sure. It's only a problem if an exception occurs that did **not** come from the pipeline, since we have the ContinueWith. I don't disagree that we should be better at protecting against exceptions here though. I'll file an issue for improved exception handling and testing.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: Agreed - we need finally and need to handle these random exceptions (as commented below)

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: :+1: internally we would keep this task and we would invoke ConnectionClosed event in a continuation.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: #99 and #98

anurse:
Merging when the builds complete unless someone has any complaints :). @davidfowl has verbally cleared his "requested changes" status

moozzyk:
:clap:

PR #94, opened 2017-01-06
WebSockets Testing

mikaelm12:
Issue tracked here: https://github.com/aspnet/SignalR/issues/29
Test is based on the old tests from the SignalR-Server repo [(Server Fixture)](https://github.com/aspnet/SignalR-Server/blob/dev/test/Microsoft.AspNetCore.SignalR.Testing.Common/ServerFixture.cs) 

@BrennanConroy @anurse @moozzyk @davidfowl

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
@@ -5,6 +5,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using System.Text;
 
 namespace Microsoft.AspNetCore.SignalR.Test.Server
 {
@@ -31,6 +32,12 @@ namespace Microsoft.AspNetCore.SignalR.Test.Server
             app.UseSignalR(routes =>
             {
                 routes.MapHub<TestHub>("/testhub");
+            });
+
+            var data = Encoding.UTF8.GetBytes("Server online");
+            app.Use(async (context, next) =>
+            {
+                await context.Response.Body.WriteAsync(data, 0, data.Length);
             });
         }
     }
--- a/test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
@@ -1,0 +1,113 @@
+﻿using System;
+using System.IO;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.IntegrationTesting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.PlatformAbstractions;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    public class ServerFixture : IDisposable
+    {
+        private readonly bool _verbose;
+
+        private Lazy<string> _baseUrl;
+        private object _lock = new object();
+        private ILoggerFactory _loggerFactory;
+        private IApplicationDeployer _deployer;
+
+        public string BaseUrl => _baseUrl.Value;
+
+        public string ServerProjectName
+        {
+            get { return "Microsoft.AspNetCore.SignalR.Test.Server"; }
+        }
+
+        public ServerFixture()
+        {
+            _loggerFactory = new LoggerFactory();
+
+            _verbose = string.Equals(Environment.GetEnvironmentVariable("SIGNALR_TESTS_VERBOSE"), "1");
+            if (_verbose)
+            {
+                _loggerFactory.AddConsole();
+            }
+
+            _baseUrl = new Lazy<string>(() => Deploy().Result);
+        }
+
+        private async Task<string> Deploy()
+        {
+            var url = Environment.GetEnvironmentVariable("SIGNALR_TESTS_URL");
+            if (!string.IsNullOrEmpty(url))
+            {
+                return url;
+            }
+
+            Console.WriteLine("Deploying test server...");
+
+            var parameters = new DeploymentParameters(
+                applicationPath: GetApplicationPath(ServerProjectName),
+                serverType: ServerType.Kestrel,
+                runtimeFlavor: RuntimeFlavor.CoreClr,
+                runtimeArchitecture: RuntimeArchitecture.x64);
+            _deployer = ApplicationDeployerFactory.Create(parameters, _loggerFactory.CreateLogger("Deployment"));
+            var result = _deployer.Deploy();
+
+            // Ensure it's working
+            var client = new HttpClient();
+            var logger = _loggerFactory.CreateLogger("Connection");
+            var resp = await RetryHelper.RetryRequest(
+                () => client.GetAsync(result.ApplicationBaseUri), logger, result.HostShutdownToken).ConfigureAwait(false);
+            resp.EnsureSuccessStatusCode();
+
+            Console.WriteLine("Test server ready. Running tests...");
+            return result.ApplicationBaseUri;
+        }
+
+        private static string GetApplicationPath(string projectName)
+        {
+            return Path.Combine(GetSolutionDir(), "test", projectName);
+        }
+
+        public void Dispose()
+        {
+            if (_deployer != null)
+            {
+                _deployer.Dispose();
+                _deployer = null;
+            }
+        }
+
+        public static string GetSolutionDir(string slnFileName = null)
+        {
+            var applicationBasePath = PlatformServices.Default.Application.ApplicationBasePath;
+
+            var directoryInfo = new DirectoryInfo(applicationBasePath);
+            do
+            {
+                if (string.IsNullOrEmpty(slnFileName))
+                {
+                    if (directoryInfo.EnumerateFiles("*.sln").Any())
+                    {
+                        return directoryInfo.FullName;
+                    }
+                }
+                else
+                {
+                    if (File.Exists(Path.Combine(directoryInfo.FullName, slnFileName)))
+                    {
+                        return directoryInfo.FullName;
+                    }
+                }
+
+                directoryInfo = directoryInfo.Parent;
+            }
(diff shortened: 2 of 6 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
```diff
@@ -0,0 +1,113 @@
+﻿using System;
+using System.IO;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.IntegrationTesting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.PlatformAbstractions;
+using System.Linq;
```
BrennanConroy: Sort

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            const string message = "Hello, World!";
+
+            using (var ws = new ClientWebSocket())
+            {
+                var url = _serverFixture.BaseUrl.Replace("http", "ws") + "echo/ws";
+
+                await ws.ConnectAsync(new Uri(url), CancellationToken.None);
+                var bytes = Encoding.UTF8.GetBytes(message);
+                await ws.SendAsync(new ArraySegment<byte>(bytes), WebSocketMessageType.Binary, true, CancellationToken.None);
```
BrennanConroy: Expose the `HostShutdownToken` in `_serverFixture` and pass it here?

test/Microsoft.AspNetCore.SignalR.Tests/project.json
```diff
@@ -19,7 +21,8 @@
         "Microsoft.NETCore.App": {
           "version": "1.2.0-*",
           "type": "platform"
-        }
+        },
+        "System.Net.WebSockets.Client": "4.4.0-beta-24829-01"
```
BrennanConroy: Star

test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
```diff
@@ -0,0 +1,113 @@
+﻿using System;
```
BrennanConroy: license

test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Linq;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    public class ServerFixture : IDisposable
+    {
+        private readonly bool _verbose;
+
+        private Lazy<string> _baseUrl;
+        private object _lock = new object();
```
BrennanConroy: unused

test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
```diff
@@ -0,0 +1,113 @@
...
+        public string ServerProjectName
+        {
+            get { return "Microsoft.AspNetCore.SignalR.Test.Server"; }
+        }
+
+        public ServerFixture()
+        {
+            _loggerFactory = new LoggerFactory();
+
+            _verbose = string.Equals(Environment.GetEnvironmentVariable("SIGNALR_TESTS_VERBOSE"), "1");
```
BrennanConroy: This is the only use of `_verbose` probably don't need to make it a class member.

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    [CollectionDefinition(Name)]
+    public class FunctionalTestsCollection : ICollectionFixture<ServerFixture>
+    {
+        public const string Name = "FunctionalTests";
```
moozzyk: rename this to "EndToEndTests"?

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    [CollectionDefinition(Name)]
+    public class FunctionalTestsCollection : ICollectionFixture<ServerFixture>
```
moozzyk: Rename to EndToEndTestsCollection and move to a separate file. (FunctionalTests is confusing since there is a separate called FunctionalTests). 

Side note: we need to clean the mess with the test projects we have many projects but most of them don't have many tests (or at all)

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTests.cs
moozzyk: What is the scenario for this? What would it prevent? If the server goes down the websocket will be closed..

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            using (var ws = new ClientWebSocket())
+            {
+                var url = _serverFixture.BaseUrl.Replace("http", "ws") + "echo/ws";
+
+                await ws.ConnectAsync(new Uri(url), CancellationToken.None);
+                var bytes = Encoding.UTF8.GetBytes(message);
+                await ws.SendAsync(new ArraySegment<byte>(bytes), WebSocketMessageType.Binary, true, CancellationToken.None);
+
+                var buffer = new ArraySegment<byte>(new byte[1024]);
+                var result = await ws.ReceiveAsync(buffer, CancellationToken.None);
```
moozzyk: Should we have a timoeut for this?

test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
```diff
@@ -5,6 +5,7 @@
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using System.Text;
```
BrennanConroy: Sort

test/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
```diff
@@ -32,6 +33,12 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
             {
                 routes.MapHub<TestHub>("/testhub");
             });
+
+            var data = Encoding.UTF8.GetBytes("Server online");
```
davidfowl: You don't need to get the bytes, you can just call WriteAsync("Server online")

test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.Extensions.PlatformAbstractions;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    public class ServerFixture : IDisposable
+    {
+        private readonly bool _verbose;
+
+        private Lazy<string> _baseUrl;
```
davidfowl: Why does this need to be lazy?

davidfowl:
Why are we doing this level of testing? I thought we were going to do Kestrel testing inplace of the TestServer?

moozzyk:
@davidfowl It [is using](https://github.com/aspnet/SignalR/pull/94/files#diff-ff7d640564e6cda828ade24c8c4d840eR53) Kestrel. If you referring to Microsoft.AspNetCore.SignalR.Test.Server as the TestServer it is just the application we run. As for the test itself - we do want to support using bare websockets with SignalR server. This basic smoke test ensures we don't break this scenario. (It was also the simplest test to test the infrastructure for end-to-end tests added in this PR)

davidfowl:
> @davidfowl It is using Kestrel. If you referring to Microsoft.AspNetCore.SignalR.Test.Server as the TestServer it is just the application we run. As for the test itself - we do want to support using bare websockets with SignalR server. This basic smoke test ensures we don't break this scenario. (It was also the simplest test to test the infrastructure for end-to-end tests added in this PR)

I'm seeing deployers which is a certain kind of test but not the one I was envisioning for this. I want an "everything should run in a single process" functional test.

test/Microsoft.AspNetCore.SignalR.Tests/ServerFixture.cs
moozzyk: To delay deployment - if you have some tests in the same test class that are not using the server you don't want spend time on deployment since it is not needed (e.g. when you try to debug this test)

moozzyk:
We do have test(s) like that (https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs) but we concluded that it won't work with WebSockets (@anurse knows exact reasons) and hence we needed to use deployers to test that. (One thing that is not great is that unless we fix issues with websockets we will have to use this pattern to write tests where the webocket transport is used). 
A side note: we also need to clean up test projects and decides what goes where - currently we seem to have more test projects that tests.

davidfowl:
@moozzyk it should work fine with websockets. Those tests you point to aren't what I was thinking we'd do. We'd just create a `WebHostBuilder` using Kestrel and SignalR in process and get the URL and hit it. No deployers, or file copying or IIS. WebSockets works fine with that approach.

moozzyk:
OK. We will try that. I agree that not having deployers and copying would be much better.

anurse:
In-proc functional tests should work with WebSockets. It's **TestHost** that won't work with WebSockets because it omits the actual network pipeline.

mikaelm12:
Closing this and now working in https://github.com/aspnet/SignalR/pull/100

PR #101, opened 2017-01-10
Fix endpoint and transport tear down

davidfowl:
Various fixes to make closing connections work well.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
@@ -11,10 +11,14 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public class ChannelConnection<T> : IChannelConnection<T>
     {
-        public IReadableChannel<T> Input { get; }
-        public IWritableChannel<T> Output { get; }
+        public IChannel<T> Input { get; }
+        public IChannel<T> Output { get; }
 
-        public ChannelConnection(IReadableChannel<T> input, IWritableChannel<T> output)
+        IReadableChannel<T> IChannelConnection<T>.Input => Input;
+
+        IWritableChannel<T> IChannelConnection<T>.Output => Output;
+
+        public ChannelConnection(IChannel<T> input, IChannel<T> output)
         {
             Input = input;
             Output = output;
@@ -22,9 +26,8 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 
         public void Dispose()
         {
-            Output.Complete();
-            (Input as IDisposable)?.Dispose();
-            (Output as IDisposable)?.Dispose();
+            Output.TryComplete();
+            Input.TryComplete();
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Sockets/Internal/FramingChannel.cs
+++ b/src/Microsoft.AspNetCore.Sockets/Internal/FramingChannel.cs
@@ -43,6 +43,15 @@ namespace Microsoft.AspNetCore.Sockets.Internal
             }
         }
 
+        private void CancelRead()
+        {
+            // We need to fake cancellation support until we get a newer build of pipelines that has CancelPendingRead()
+
+            // HACK: from hell, we attempt to cast the input to a pipeline writer and write 0 bytes so it so that we can
+            // force yielding the awaiter, this is buggy because overlapping writes can be a problem.
+            (_connection.Input as IPipelineWriter)?.WriteAsync(Span<byte>.Empty);
+        }
+
         bool IReadableChannel<Message>.TryRead(out Message item)
         {
             // We need to think about how we do this. There's no way to check if there is data available in a Pipeline... though maybe there should be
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -35,7 +34,10 @@ public LongPollingTransport(IReadableChannel<Message> connection, ILoggerFactory
 
             try
             {
-                using (var message = await _connection.ReadAsync(_cancellationSource.Token))
+                // TODO: We need the ability to yield the connection without completing the channel.
```
davidfowl: We won't be able to use this to indicate poll but don't complete the channel. We'll likely end up having to send a null message or something similar.

src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
```diff
@@ -11,20 +11,23 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public class ChannelConnection<T> : IChannelConnection<T>
     {
-        public IReadableChannel<T> Input { get; }
-        public IWritableChannel<T> Output { get; }
+        public IChannel<T> Input { get; }
```
anurse: Why do we need to expose both sides of both channels here? I've been trying to avoid that as much as possible...

src/Microsoft.AspNetCore.Sockets/Internal/FramingChannel.cs
```diff
@@ -81,14 +90,18 @@ bool IWritableChannel<Message>.TryWrite(Message item)
...
+            _connection.Input.Complete(error);
             return true;
         }
 
         private async Task<Message> AwaitReadAsync(ReadableBufferAwaitable awaiter, CancellationToken cancellationToken)
         {
-            // Just await and then call ReadSync
-            var result = await awaiter;
-            return ReadSync(result, cancellationToken);
+            using (cancellationToken.Register(state => ((FramingChannel)state).CancelRead(), this))
```
BrennanConroy: Can we register this in the constructor instead of every `AwaitReadAsync`?

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
moozzyk: I assume this will be used to timeout poll requests for which there was no data to send within a timeout so that the poll request are not killed by IIS/routers as stale requests. 

Do we want to open a bug for this?

src/Microsoft.AspNetCore.Sockets/Internal/PipelineConnection.cs
```diff
@@ -1,25 +1,31 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
moozzyk: nit: needed?

src/Microsoft.AspNetCore.Sockets/Internal/FramingChannel.cs
```diff
@@ -43,6 +43,15 @@ ValueTask<Message> IReadableChannel<Message>.ReadAsync(CancellationToken cancell
             }
         }
 
+        private void CancelRead()
+        {
+            // We need to fake cancellation support until we get a newer build of pipelines that has CancelPendingRead()
```
moozzyk: Will we be able to revert changes to PipelineConnection.cs then?

src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
davidfowl: It can be private, it just needs to be typed as `IChannel<T>` in this class so we can complete it.

src/Microsoft.AspNetCore.Sockets/Internal/FramingChannel.cs
davidfowl: The token is passed into ReadAsync.

BrennanConroy: Oh, my bad

src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
anurse: I'm OK with that. I just don't want user code getting access to the full Channel :).

src/Microsoft.AspNetCore.Sockets/Internal/FramingChannel.cs
davidfowl: Yes there's a newer API but @pranavkm wrote some dubious logic and we're not getting updated pipeline packages.

davidfowl: Yep

PR #108, opened 2017-01-10
Remove delays in HubConnectionTests

mikaelm12:
We can remove these delays now that https://github.com/aspnet/SignalR/pull/92 has been merged.
@anurse @moozzyk @BrennanConroy


moozzyk:
:clap:

PR #110, opened 2017-01-11
Remove streaming transport as a top level API

davidfowl:
- Remove Streaming* classes from Sockets. The main
API will be channels based and streaming transports
will use the PipelineChannel (formerly FramingChannel) to
access messages.
- Added WriteAsync and ReadAsync to Connection and hid
the IChannelConnection from public API.
- Also fixed the fact that unknown methods caused server side
exceptions.
- Changed the consumption pattern to WaitToReadAsync/TryRead to avoid
exceptions.
- React to API changes

PS: I'm leaving this here but I'm going to merge the PR since it's so disruptive and it also fixes the API changes made to channels and pipelines.


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -220,9 +220,9 @@ public override Task OnDisconnectedAsync(StreamingConnection connection)
                     await previousTask;
 
                     var tasks = new List<Task>(group.Connections.Count);
-                    foreach (var groupConnection in group.Connections.Cast<StreamingConnection>())
+                    foreach (var groupConnection in group.Connections.Cast<Connection>())
```
anurse: Cast?

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -37,18 +37,19 @@ public LongPollingTransport(IReadableChannel<Message> connection, ILoggerFactory
                 // TODO: We need the ability to yield the connection without completing the channel.
                 // This is to force ReadAsync to yield without data to end to poll but not the entire connection.
                 // This is for cases when the client reconnects see issue #27
-                using (var message = await _connection.ReadAsync(context.RequestAborted))
+                await _application.WaitToReadAsync(context.RequestAborted);
```
anurse: Do we care about the return value? If there are no more messages and the channel is closed during this await the request will fall through the below and end up... 404ing?

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
```diff
@@ -152,12 +152,13 @@ private void LogFrame(string action, WebSocketFrame frame)
 
         private async Task StartSending(IWebSocketConnection ws)
         {
-            while (true)
+            while (await _application.Input.WaitToReadAsync())
             {
                 // Get a frame from the application
-                try
+                Message message;
+                if (_application.Input.TryRead(out message))
```
anurse: We should update this pattern to use a loop here. `WaitToReadAsync` returns `Task<T>` so we should avoid allocating a new Task until we actually have a high likelihood of blocking. We should try to synchronously gather all the messages we can until we hit the end.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
davidfowl: Old codez, will fix

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
davidfowl: Why would it 404? TryRead would return false no? It would return 200 and the next request would get the 204 because the channel is complete. We could optimize it by returning the 204 immediately if WaitToReadAsync returns false.

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
davidfowl: Sure, I was initially more concerned with correctness. We can do a perf pass.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -138,15 +139,26 @@ public class HubEndPoint<THub, TClient> : StreamingEndPoint, IInvocationBinder w
...
 
-            while (true)
+            while (await connection.Transport.Input.WaitToReadAsync())
             {
-                // TODO: Handle receiving InvocationResultDescriptor
-                var invocationDescriptor = await invocationAdapter.ReadMessageAsync(stream, this) as InvocationDescriptor;
+                Message message;
+                if (!connection.Transport.Input.TryRead(out message))
+                {
+                    continue;
```
moozzyk: what does that mean (i.e. TryRead returns false) and why it is OK to continue?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -177,11 +189,19 @@ public class HubEndPoint<THub, TClient> : StreamingEndPoint, IInvocationBinder w
                     _logger.LogError("Unknown hub method '{method}'", invocationDescriptor.Method);
                 }
 
-                await invocationAdapter.WriteMessageAsync(result, stream);
+                // TODO: Pool memory
+                var outStream = new MemoryStream();
+                await invocationAdapter.WriteMessageAsync(result, outStream);
+
+                var buffer = ReadableBuffer.Create(outStream.ToArray()).Preserve();
+                if (await connection.Transport.Output.WaitToWriteAsync())
```
moozzyk: What happens if WaitToWriteAsync returns false? Does this mean that the channel is being closed and WaitToReadAsync() is guranteed to return false so we exit the loop? Otherwise we will never send the message....

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -290,7 +310,7 @@ Type[] IInvocationBinder.GetParameterTypes(string methodName)
             Type[] types;
             if (!_paramTypes.TryGetValue(methodName, out types))
             {
-                throw new InvalidOperationException($"The hub method '{methodName}' could not be resolved.");
+                return Type.EmptyTypes;
```
moozzyk: This does not seem right. If you are using non-self descriptive format (like protobuf) you don't know what to read and you will mess up the stream.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -28,8 +24,23 @@ public bool TryGetConnection(string id, out ConnectionState state)
...
+            var applicationToTransport = Channel.CreateUnbounded<Message>();
+
+            var transportSide = new ChannelConnection<Message>(applicationToTransport, transportToApplication);
+            var applicationSide = new ChannelConnection<Message>(transportToApplication, applicationToTransport);
+
+            var state = new ConnectionState(
+                new Connection(id, applicationSide),
+                transportSide);
+
+            _connections.TryAdd(id, state);
```
moozzyk: can this ever fail?

samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
```diff
@@ -35,7 +36,10 @@ public void OnDisconnectedAsync(StreamingConnection connection)
             foreach (var connection in _connectionList)
             {
                 var formatter = _formatterResolver.GetFormatter<T>(connection.Metadata.Get<string>("formatType"));
-                await formatter.WriteAsync(data, connection.Transport.GetStream());
+                var ms = new MemoryStream();
+                await formatter.WriteAsync(data, ms);
+                var buffer = ReadableBuffer.Create(ms.ToArray()).Preserve();
+                await connection.Transport.Output.WriteAsync(new Message(buffer, Format.Binary, endOfMessage: true));
```
moozzyk: Should we remove format?

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
anurse: Yeah, it just means we'd get a 200 with no data.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: TryRead means, get the next item from the buffer if there's any. It returning false means there's nothing available right now so wait for more data.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Yes

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
moozzyk: I was thinking about 204 too but the client MUST be able to handle 200 with no data since the server should timeout a pending poll if there was no messages for the corresponding connection within a timeout (we used 110 secs by default before because IIS would kill requests after 120 secs)

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
davidfowl: Nope

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
moozzyk: but why wouldn't there be anything to Read if `WaitToReadAsync` unblocked?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Ending the connection for an unknown method is in all cases in incorrect. We need to expand the interface so that we know we *have* to end the connection if the format isn't self describing.

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
davidfowl: I sent a PR to handle that.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
moozzyk: Yeah, the concern I had about this was that because we allow mixing formats different connections will behave differently for the same hub method and parameters...

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Right, this should never return false (in theory). The code is cleaner this way though as I don't have to nest the entire code block in an if. This should be changed to a loop anyways. I'll update the other PR to loop.

anurse: It can return false due to a race between readers. We're a single-consumer system so it's not really a problem, but that's the theoretical case where it can be false even when WaitToReadAsync unblocks

moozzyk: If we want to check the result then yes - keeping it is cleaner.

PR #112, opened 2017-01-11
Use connections format

BrennanConroy:
@davidfowl @anurse @moozzyk @mikaelm12

### DIFF
--- a/samples/SocketsSample/Hubs/Chat.cs
+++ b/samples/SocketsSample/Hubs/Chat.cs
@@ -11,12 +11,12 @@ namespace SocketsSample.Hubs
     {
         public override async Task OnConnectedAsync()
         {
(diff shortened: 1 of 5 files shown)


samples/SocketsSample/Hubs/Chat.cs
```diff
@@ -40,7 +40,12 @@ public Task SendToGroup(string groupName, string message)
         {
             await Groups.RemoveAsync(groupName);
 
-            await Clients.Group(groupName).InvokeAsync("Send", $"{Context.Connection.ConnectionId} left {groupName}");
+            await Clients.Group(groupName).InvokeAsync("Send", $"{Context.ConnectionId} left {groupName}");
+        }
+
+        public Task Self(string message)
```
davidfowl: Self? Use a verb. Echo maybe?

PR #113, opened 2017-01-11
Use TryRead and TryWrite

davidfowl:
- Use TryWrite to avoid errors on channel close for /send requests
- Use TryRead until it returns false for all transports but long polling

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
@@ -226,9 +226,12 @@ namespace Microsoft.AspNetCore.Sockets
                     endOfMessage: true);
 
                 // REVIEW: Do we want to return a specific status code here if the connection has ended?
-                if (await state.Application.Output.WaitToWriteAsync())
+                while (await state.Application.Output.WaitToWriteAsync())
                 {
-                    state.Application.Output.TryWrite(message);
+                    if (state.Application.Output.TryWrite(message))
+                    {
+                        break;
+                    }
                 }
             }
             else
--- a/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
@@ -225,8 +225,11 @@ namespace Microsoft.AspNetCore.Sockets
                     format,
                     endOfMessage: true);
 
-                await state.Application.Output.WriteAsync(message);
-
+                // REVIEW: Do we want to return a specific status code here if the connection has ended?
+                if (await state.Application.Output.WaitToWriteAsync())
+                {
+                    state.Application.Output.TryWrite(message);
+                }
             }
             else
             {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -225,8 +225,11 @@ private Task ProcessGetId(HttpContext context)
                     format,
                     endOfMessage: true);
 
-                await state.Application.Output.WriteAsync(message);
-
+                // REVIEW: Do we want to return a specific status code here if the connection has ended?
+                if (await state.Application.Output.WaitToWriteAsync())
```
BrennanConroy: We need a loop here and if TryWrite returns true then break

davidfowl: A loop where?

BrennanConroy:
```
while (WaitToWriteAsync())
{
    if (TryWrite())
    {
        break;
    }
}
```

davidfowl: Not sure that's needed for an unbounded channel.

moozzyk:
:ship::it:

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
BrennanConroy: True, though we'll want to look at that later

anurse: I'd rather not rely on the implementation details to make the code correct. If we change to a bounded channel, this could break, so let's be defensive.

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
```diff
@@ -33,7 +33,7 @@ public ServerSentEventsTransport(IReadableChannel<Message> application, ILoggerF
                 while (await _application.WaitToReadAsync(context.RequestAborted))
                 {
                     Message message;
-                    if (_application.TryRead(out message))
+                    while (_application.TryRead(out message))
```
anurse: It does strike me that WaitToReadAsync could still avoid heap allocation when a wait isn't needed since the return is a bool so it could cache the completed `true` and `false` tasks. Still, better for us to avoid depending on that and do the loop ourselves as well.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
davidfowl: @stephentoub whats the correct usage of this thing.

stephentoub: If you know the channel is unbounded, then you also know that writes will always succeed (standard caveats of OOMs and the like) unless the channel has been closed. Thus, as long as the channel hasn't been closed, TryWrite will aways return true, and WaitToWriteAsync will always complete synchronously with a true result. This would be documented, and is inherent to the nature of it being unbounded.
If you don't know whether the channel has room, e.g. it's bounded, then you would need to do something like a loop with WaitToWriteAsync/TryWrite or use WriteAsync to wait until there's room to write.

PR #114, opened 2017-01-11
fix autobahn test app to be an app

anurse:
also remove net46 version of autobahn conformance test, the autobahn
doesn't need to be run twice...


anurse:
I'm going to add code to appveyor and travis to install autobahn so they can run those tests

PR #120, opened 2017-01-12
Remove format from connect and only use it in /send

BrennanConroy:


davidfowl:
This PR brings up an interesting scenario. What happens when you are sending from the HubLifetimeManager. What format do you use? That needs to flow from the callsite somehow.

moozzyk:
We came to the same conclusion with @BrennanConroy yesterday. Basically the format of incoming messages is not coupled to the format of outgoing messages.

Issue #118, opened 2017-01-11
Inheriting DefaultHubLifetimeManager


harisbotic:
I would like to have a method which invokes all users through user id (Guid), which I would extract from claims.

Would it make sense if you change the method '**InvokeAllWhere**' from private to **protected** 
[Link to InvokeAllWhere method](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs#L58)

In that case I could have my own HubLifetimeManager where I would change only what I need and the rest of functionality would stay default

davidfowl:
Send a pull request 😄

davidfowl:
@harisbotic I do expect us to support this out of the box with a service like we had in older versions of SignalR. See https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Core/IUserIdProvider.cs. The interface would likely be `string GetUserId(ClaimsPrincipal)`

harisbotic:
@davidfowl I might be on yours level, but I have another suggestion for this.

Is it better to just change `string userId` to `ClaimsPrincipal claims`, everywhere in library.

For ex.

```csharp
public override Task InvokeUserAsync(ClaimsPrincipal claims, string methodName, object[] args)
        {
            return InvokeAllWhere(methodName, args, connection =>
            {
                return claims.Equals(connection.User);
            });
        }
```

In this case userId could be both string and Guid, or int,  and it would work for anyone out of the box.

While I don't think that this would be the case with `IUserIdProvider` solution

harisbotic:
Actually, its not that good. It would work in my case, but it would require you to have original reference to user claims.

`IUserIdProvider` is the solution I think, but unfortunately I don't have time atm to make PR for this, but after February if it doesn't get solved, I will try to do it myself.

grahamehorner:
thoughts on connection metadata

if the HubLifetimeManager had and injectable IHubLifetimeManagerMetadataIndex that stored the connection id along with other meta information i.e userName, groupName etc. this would allow developers to perform a queries that can used to identify the connection with the need to interate over all connections, it could also be to implement a persistant/distributed cache repository where needed.

davidfowl:
Tracked by https://github.com/aspnet/SignalR/issues/604

PR #121, opened 2017-01-12
Change DefaultHubLifetimeManager members to protected

harisbotic:
Just changed all members which were private to protected

Issue #118


harisbotic:
This needs better solution, its discussed here #118

PR #122, opened 2017-01-12
Use WaitToWriteAsync(), TryWrite() in HubLifetimeManager(s)

davidfowl:
Avoids exceptions when connections are dead but still in the list.


BrennanConroy:
👍 
🐑 :it:

Issue #119, opened 2017-01-12
Transport Protocols


anurse:
We need to build some small protocols for transports in order to make them consistent with each other. This issue tracks the requirements and the proposed solutions. I'll start some PRs for simple "specs" for the necessary protocols to discuss those. Then I'll open issues for the actual work to get these implemented.

Requirements:

1. Text/Binary frame distinction
2. Binary-safe protocol (able to transmit arbitrary binary data)
3. Efficient round-trip time for sending messages to clients
4. Duplex (server can send to client and vice versa)

## WebSockets

Nothing, it's perfect :) (For our purposes). It has Text/Binary frame distinction, it's Binary-safe, and messages can be sent directly from server to client

## Server Sent Events

This is almost good enough, but has a few limitations.

It has no built in way to distinguish between binary and text messages, but it does have a `type` field, which we can use to transmit this.

It is not binary-safe because the payload is stored in a single line, with the line-terminator is the delimiter, and there is no way to escape it in a stable way. You can write multi-line messages, but `\r\n` terminators are converted to `\n`, which would break binary data. To solve this, we base64-encode binary messages, and properly escape line breaks in text messages.

It has an efficient round-trip time for sending to clients because the server can send unsolicited messages and the connection stays open.

Finally, it's not duplex, it's server-to-client only, so we need the `/send` endpoint to send from client to server

So, some sample encodings:

### Example 1

Text Message: `FooBar`
Encoding:

```
type: text\r\n
data: FooBar\r\n
\r\n
```

Reconstruction: `FooBar`

### Example 2

Text Message: `Foo\nBar`
Encoding:

```
type: text\r\n
data: Foo\r\n
data: Bar\r\n
\r\n
```

Reconstruction: `Foo\nBar`

### Example 3

Text Message: `Foo\r\nBar`
Encoding:

```
type: text\r\n
data: Foo\r\n
data: Bar\r\n
\r\n
```

Reconstruction: `Foo\nBar` * Note, loss of `\r\n`, but for text frames we allow that.

### Example 4

Binary Message: UTF-8 encoding of `A\r\nB`
Encoding:

```
type: binary\r\n
data: QQ0KQg==\r\n
\r\n
```

Reconstruction: UTF-8 encoding of `A\r\nB`

## Long Polling

The first major problem is that the round-trip time is bad since the client needs to start a new poll for each individual message from server to client. So to solve this, we batch messages and send multiple messages in a single HTTP response.

[...]

Issue #116, opened 2017-01-11
Review ServerSentEvents transport for raw binary data


anurse:
The ServerSentEvents transport is currently only able to send simple text messages with no new lines (why? details below). We need to make some changes to support multi-line messages and binary messages

1. Split text messages on newlines and write them as individual `data: [line]` lines, in order to ensure they are properly reconstructed (see https://html.spec.whatwg.org/multipage/comms.html#event-stream-interpretation)

1. Base64-encode binary messages

In general we just need to review how we are intending to encode messages on transports. We can't just blindly write binary, or even text (if it has newlines), data to the transport because if the data includes bytes `0x0A` (`\n`) or `0x0D` (`\r`) it will break the protocol. Something to consider as we build the protocol for Long Polling to batch messages.

/cc @davidfowl

anurse:
See #125

muratg:
@anurse is this good to close?

anurse:
Yes.

PR #125, opened 2017-01-12
Describe transport protocols

anurse:
Completes #119 and #116 

This describes the proposed protocols for Long Polling, Server Sent Events, and HTTP Post (aka `/send`). It also describes WebSockets briefly and the requirements imposed on ASP.NET Endpoint Transports.

/cc @moozzyk @BrennanConroy @mikaelm12 @DamianEdwards @muratg

### DIFF
--- a/specs/TransportProtocols.md
+++ b/specs/TransportProtocols.md
@@ -8,11 +8,13 @@ A transport is required to have the following attributes:
 
 1. Duplex - Able to send messages from Server to Client and from Client to Server
 1. Frame-based - Messages have fixed length (as opposed to streaming, where data is just pushed throughout the connection)
-1. Frame Metadata - Able to encode two pieces of Frame Metadata
+1. Frame Metadata - Able to encode one piece of Frame Metadata
     * `Type` - Either `Binary` or `Text`.
-    * `EndOfMessage` - Boolean indicating if this frame represents the final frame of a complete message.
+1. Able to handle partial messages from the Server application (i.e. where `EndOfMessage` is `false`). The transport can opt to implement this by buffering, since only WebSockets natively supports this functionality.
 1. Binary-safe - Able to transmit arbitrary binary data, regardless of content
-1. Text-safe - Able to transmit arbitrary text data, preserving the content. Line-endings must be preserved **but may be converted to a different format**. For example `\r\n` may be converted to `\n`. This is due to quirks in some transports (Server Sent Events)
+1. Text-safe - Able to transmit arbitrary text data, preserving the content. Line-endings must be preserved **but may be converted to a different format**. For example `\r\n` may be converted to `\n`. This is due to quirks in some transports (Server Sent Events). If the exact line-ending needs to be preserved, the data should be sent as a `Binary` message.
+
+Multi-frame messages (where a message is split into multiple frames) may not overlap and all frames **must** have the same `Type` value (`Text` or `Binary`). It is an error to change the `Type` flag mid-message and the endpoint may terminate the connection if that occurs.
 
 The only transport which fully implements the duplex requirement is WebSockets, the others are "half-transports" which implement one end of the duplex connection. They are used in combination to achieve a duplex connection.
 
@@ -35,6 +37,8 @@ This transport uses HTTP Post requests to a specific end point to encode message
 This transport requires that a connection be established using the `[endpoint-base]/negotiate` end point.
 
 The HTTP POST request is made to the URL `[endpoint-base]/[connection-id]/send`. The content consists of frames in the same format as the Long Polling transport. It is up to the client which of the Text or Binary protocol they use, and the server is able to detect which they use either via the `Content-Type` (see the Long Polling transport section) or via the first byte (`T` for the Text-based protocol, `B` for the binary protocol). Upon receipt of the **entire** request, the server will process and deliver all the messages, responding with `202 Accepted` if all the messages are successfully processed. If a client makes another request to `/send` while an existing one is outstanding, the existing request is terminated by the server with the `409 Conflict` status code.
+
+This transport will always produce single-frame messages, since the Long-Polling protocol below does not support encoding an end-of-message flag.
 
 ## Server-Sent Events (Server-to-Client only)
 
@@ -59,34 +63,31 @@ In the first event, the value of `baz` would be `boz\nbiz\nflarg`, due to the co
 In this transport, the client establishes an SSE connection to `[endpoint-base]/sse`, and the server responds with an HTTP response with a `Content-Type` of `text/event-stream`. Each SSE event represents a single frame from client to server. The transport uses unnamed events, which means only the `data` field is available. Thus we use the first line of the `data` field for frame metadata. The frame body starts on the **second** line of the `data` field value. The first line has the following format (Identifiers in square brackets `[]` indicate fields defined below):
 
 ```
-[Type],[Fin]
+[Type]
 ```
 
 * `[Type]` is a single UTF-8 character representing the type of the frame; `T` indicates Text, and `B` indicates Binary.
-* `[Fin]` is a single UTF-8 character indicating if the frame is the last frame of the message; `T` indicates `true`, `F` indicates `false`
 
 If the `[Type]` field is `T`, the remaining lines of the `data` field contain the value, in UTF-8 text. If the `[Type]` field is `B`, the remaining lines of the `data` field contain Base64-encoded binary data. Any `\n` characters in Binary frames are removed before Base64-decoding. However, servers should avoid line breaks in the Base64-encoded data.
 
 For example, when sending the following frames (`\n` indicates the actual Line Feed character, not an escape sequence):
 
-* Type=`Text`, Fin=`true`, "Hello\nWorld"
-* Type=`Binary`, Fin=`false`, `0x01 0x02`
-* Type=`Binary`, Fin=`true`, `0x03 0x04`
+* Type=`Text`, "Hello\nWorld"
+* Type=`Binary`, `0x01 0x02`
 
 The encoding will be as follows
 
 ```
-data: T,T
+data: T
 data: Hello
 data: World
 
-data: B,F
+data: B
 data: AQI=
 
-data: B,T
-data: AwQ=
-
 ```
+
+This transport will buffer incomplete frames sent by the server until the full message is available and then send the message in a single frame.
 
 ## Long Polling (Server-to-Client only)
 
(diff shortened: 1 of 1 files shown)


specs/TransportProtocols.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Transport Requirements
+
+A transport is required to have the following attributes:
+
+1. Duplex - Able to send messages from Server to Client and from Client to Server
+1. Frame-based - Messages have fixed length (as opposed to streaming, where data is just pushed throughout the connection)
+1. Frame Metadata - Able to encode two pieces of Frame Metadata
+    * `Type` - Either `Binary` or `Text`.
+    * `EndOfMessage` - Boolean indicating if this frame represents the final frame of a complete message.
```
anurse: How much do we care about "EndOfMessage" here? We want to represent this data over a transport like WebSockets which provides it, but SignalR certainly doesn't care about it. It seems reasonable to me that we could just ignore it in the other transports. But the issue does arise when someone puts a partial message in the Channel, and the Long-Polling or SSE transport is attached. If we don't represent this data, are we misrepresenting the user's intent?

Despite describing how to represent it here, I actually think I fall in the camp of only having WebSockets support this. If the other transports receive a "partial" message, they just render it and ignore the EndOfMessage flag. However, we definitely need to propagate this flag up from/to transports that do support it (i.e. WebSockets) in order to allow ASP.NET Endpoints to be able to serve any kind of WebSockets-based protocol.

specs/TransportProtocols.md
```diff
@@ -0,0 +1,176 @@
...
+
+Throughout this document, the term `[endpoint-base]` is used to refer to the route assigned to a particular end point
+
+## WebSockets (Full Duplex)
+
+The WebSockets transport is unique in that it is full duplex, and a persistent connection that can be established in a single operation. As a result, the client is not required to use the `[endpoint-base]/negotiate` endpoint to establish a connection in advance. It also includes all the necessary metadata in it's own frame metadata.
+
+The WebSocket transport is activated by making a WebSocket connection to `[endpoint-base]/ws`. Upon doing so, the connection is fully established and immediately ready for frames to be sent/received. The WebSocket OpCode field is used to indicate the type of the frame (Text or Binary) and the WebSocket "FIN" flag is used to indicate the end of a message.
+
+Establishing a second WebSocket connection when there is already a WebSocket connection associated with the Endpoints connection is not permitted and will fail with a `400` (Bad Request) status code.
```
davidfowl: nit: 409

specs/TransportProtocols.md
```diff
@@ -0,0 +1,176 @@
...
+
+## HTTP Post (Client-to-Server only)
+
+HTTP Post is a half-transport, it is only able to send messages from the Client to the Server, as such it is always used with one of the other half-transports which can send from Server to Client (Server Sent Events and Long Polling).
+
+This transport uses HTTP Post requests to a specific end point to encode messages. Each HTTP Request represents a single message (this achieves the Frame-based requirement), contains raw binary data (Binary-safe) and uses the query string to encode the metadata.
+
+This transport requires that a connection be established using the `[endpoint-base]/negotiate` end point.
+
+The HTTP request is made to the URL `[endpoint-base]/send` with the following query string parameters
```
davidfowl: I always wondered if we should make this more RESTy (optional args gets harder though).

specs/TransportProtocols.md
```diff
@@ -0,0 +1,176 @@
...
+
+## WebSockets (Full Duplex)
+
+The WebSockets transport is unique in that it is full duplex, and a persistent connection that can be established in a single operation. As a result, the client is not required to use the `[endpoint-base]/negotiate` endpoint to establish a connection in advance. It also includes all the necessary metadata in it's own frame metadata.
+
+The WebSocket transport is activated by making a WebSocket connection to `[endpoint-base]/ws`. Upon doing so, the connection is fully established and immediately ready for frames to be sent/received. The WebSocket OpCode field is used to indicate the type of the frame (Text or Binary) and the WebSocket "FIN" flag is used to indicate the end of a message.
+
+Establishing a second WebSocket connection when there is already a WebSocket connection associated with the Endpoints connection is not permitted and will fail with a `400` (Bad Request) status code.
+
+## HTTP Post (Client-to-Server only)
```
davidfowl: A while back I looked at a chunked upload send end point for SignalR. Basically reverse SSE. I wonder if we should explore that again. It would be for non browser clients of course...

specs/TransportProtocols.md
```diff
@@ -0,0 +1,176 @@
...
+In the first event, the value of `baz` would be `boz\nbiz\nflarg`, due to the concatenation behavior above. Full details can be found in the spec linked above.
+
+In this transport, the client establishes an SSE connection to `[endpoint-base]/sse`, and the server responds with an HTTP response with a `Content-Type` of `text/event-stream`. Each SSE event represents a single frame from client to server. The transport uses unnamed events, which means only the `data` field is available. Thus we use the first line of the `data` field for frame metadata. The frame body starts on the **second** line of the `data` field value. The first line has the following format (Identifiers in square brackets `[]` indicate fields defined below):
+
+```
+[Type],[Fin]
+```
+
+* `[Type]` is a single UTF-8 character representing the type of the frame; `T` indicates Text, and `B` indicates Binary.
+* `[Fin]` is a single UTF-8 character indicating if the frame is the last frame of the message; `T` indicates `true`, `F` indicates `false`
```
davidfowl: Maybe this should be 1,0 instead of T,F.

specs/TransportProtocols.md
```diff
@@ -0,0 +1,176 @@
...
+data: AQI=
+
+data: B,T
+data: AwQ=
+
+```
+
+## Long Polling (Server-to-Client only)
+
+Long Polling is a server-to-client half-transport, so it is always paired with HTTP Post. It requires a connection already be established using the `[endpoint-base]/negotiate` endpoint.
```
davidfowl: You don't mention the negotiate requirement for SSE.

specs/TransportProtocols.md
```diff
@@ -0,0 +1,176 @@
...
+
+Long Polling requires that the client poll the server for new messages. Unlike traditional polling, if there are no messages available, the server will simply block the request waiting for messages to be dispatched. At some point, the server, client or an upstream proxy will likely terminate the connection, at which point the client should immediately re-send the request. Long Polling is the only transport that allows a "reconnection" where a new request can be received while the server believes an existing request is in process. This can happen because of a time out. When this happens, the existing request is immediately terminated with an empty result (even if messages are available) and the new request replaces it as the active poll request.
```
davidfowl: > the server will simply block the request waiting for messages to be dispatched

Do we want to mention timeouts here? Long polling transports will have to timeout and re-poll even if there's no data to deal with proxies that kill requests.

specs/TransportProtocols.md
```diff
@@ -0,0 +1,176 @@
...
+* Type=`Text`, Fin=`true`, "Hello\nWorld"
+* Type=`Binary`, Fin=`false`, `0x01 0x02`
+* Type=`Binary`, Fin=`true`, `0x03 0x04`
+
+The encoding will be as follows, as a list of binary digits in hex (text in parentheses `()` are comments). Whitespace and newlines are irrelevant and for illustration only.
+```
+0x66                                                   (ASCII 'B')
+0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x0B                (start of frame; 64-bit integer value: 11)
+0x80                                                   (Type = Text, Fin = True)
+0x68 0x65 0x6C 0x6C 0x6F 0x0A 0x77 0x6F 0x72 0x6C 0x64 (UTF-8 encoding of 'Hello\nWorld')
```
davidfowl: Are we going to have to write a utf8 decoder over typed arrays ?

specs/TransportProtocols.md
anurse: RESTy just meaning using the path instead of query-string? Because that's not actually how REST works... Just moving things to the path does not a REST API make.

anurse: Also, this should use POST form parameters rather than query string. I'll update that.

anurse: We could put the connection ID in the path though, that one actually makes sense.

anurse: If we do that, we should do it for all of the transports.

specs/TransportProtocols.md
anurse: If we force HTTP post to use a protocol like Long Polling does that may work. I believe the LP protocol will work as-is, since we'll just be reading the stream, waiting for a complete frame to arrive before emitting one.

specs/TransportProtocols.md
anurse: Yeah... `T`/`B` was to make Text/Binary clearer, but `T`/`F` for `Fin` seems weird. If we decide we don't want to do `Fin` at all, we could stick with `T`/`B` for the type field.

specs/TransportProtocols.md
anurse: Depends on how we want to support strings. Looks like JavaScript uses UTF-16. If we wanted to make our protocol use UTF-16 for strings that would be fine with me.

specs/TransportProtocols.md
anurse: Ping @davidfowl - I'm leaning really close to just saying that only WebSockets supports "partial" messages. The big trick is what do we do with EndPoints that provide Message objects with EndOfMessage set to false, when the transport doesn't support them. Buffer?

specs/TransportProtocols.md
anurse: Fixed

specs/TransportProtocols.md
anurse: The big problem, it seems, is that there isn't really a process to allow the server to terminate the open request gracefully. Only the client can gracefully terminate the connection because they need to write a chunk terminator. We could try just sending the response while the client is still writing chunks, but I'm not confident that would trigger the desired behavior. Obviously we can disgracefully terminate, but then it's very difficult for the client to know how much the server has read.

We can do some testing an examining, but in the mean time, I'm kinda liking the idea of the `/send` endpoint using the same protocol as the long-polling responses. For now, let's just specify that the server is not expected to deliver any of the messages until the entire response is received (i.e. streaming messages won't work, yet).

specs/TransportProtocols.md
davidfowl: Yes buffer

anurse: Ok, we'll definitely need to make EndOfMessage default to `true` so people don't accidentally forget to set it and end up never actually transmitting a frame :).

Also, WebSockets requires that all frames of a multi-frame message have the same "Opcode" (Text/Binary). We will enforce the same requirement.

specs/TransportProtocols.md
anurse: So the answer is: Yes, as currently written. However, I don't believe it'll be difficult. It's just byte manipulation, there's no crazy table lookup or anything.

anurse: There's also the [TextDecoder](http://caniuse.com/#feat=textencoder) API which converts from UInt8Array to String using UTF-8. We can make support for that a condition for using the Binary protocol in our clients.

anurse:
I'm going to merge this tomorrow morning unless I hear from @davidfowl. It's not set in stone by any means when I've merged it though. It just means I'll start implementing it :).

specs/TransportProtocols.md
davidfowl: I looked into this actually and it's not that hard to do. TextDecoder is experimental but at least there's something the browser vendors are looking at. Let's stick with utf8.

anurse: UTF-8 is the One True Text Format after all ;)

PR #128, opened 2017-01-13
React to Channel API changes

BrennanConroy:
@anurse @moozzyk @davidfowl @mikaelm12

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
@@ -6,15 +6,23 @@ using System.Threading.Tasks.Channels;
 
 namespace Microsoft.AspNetCore.Sockets.Internal
 {
+    public static class ChannelConnection
+    {
+        public static ChannelConnection<T> Create<T>(Channel<T> input, Channel<T> output)
+        {
+            return new ChannelConnection<T>(input, output);
+        }
+    }
+
     public class ChannelConnection<T> : IChannelConnection<T>
     {
-        public IChannel<T> Input { get; }
-        public IChannel<T> Output { get; }
+        public Channel<T> Input { get; }
+        public Channel<T> Output { get; }
 
-        IReadableChannel<T> IChannelConnection<T>.Input => Input;
-        IWritableChannel<T> IChannelConnection<T>.Output => Output;
+        ReadableChannel<T> IChannelConnection<T>.Input => Input;
+        WritableChannel<T> IChannelConnection<T>.Output => Output;
 
-        public ChannelConnection(IChannel<T> input, IChannel<T> output)
+        public ChannelConnection(Channel<T> input, Channel<T> output)
         {
             Input = input;
             Output = output;
@@ -22,8 +30,8 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 
         public void Dispose()
         {
-            Output.TryComplete();
-            Input.TryComplete();
+            Output.Out.TryComplete();
+            Input.Out.TryComplete();
         }
     }
 }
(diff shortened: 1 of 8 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -490,21 +491,21 @@ private static Type GetGenericType(Type genericType, Type hubType)
             stream);
 
             var buffer = ReadableBuffer.Create(stream.ToArray()).Preserve();
-            await connection.Transport.Input.WriteAsync(new Message(buffer, Format.Binary, endOfMessage: true));
+            await connection.Transport.Input.Out.WriteAsync(new Message(buffer, Format.Binary, endOfMessage: true));
```
moozzyk: the number of dots is getting ridiculous...

BrennanConroy: ...

moozzyk:
Merge it once CI is green... :trollface:

src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
```diff
@@ -6,24 +6,32 @@
 
 namespace Microsoft.AspNetCore.Sockets.Internal
 {
+    public static class ChannelConnection
```
davidfowl: Why not put this on ChannelConnection<T>.Create?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -629,89 +606,28 @@ private class TrackDispose
...
-
-                Transport = new ChannelConnection<Message>(_input, applicationToTransport);
+                Application = ChannelConnection.Create(input: applicationToTransport, output: transportToApplication);
+                var transport = ChannelConnection.Create(input: transportToApplication, output: applicationToTransport);
 
-                Connection = new Connection(Guid.NewGuid().ToString(), Transport);
+                Connection = new Connection(Guid.NewGuid().ToString(), transport);
                 Connection.Metadata["formatType"] = format;
-                Connection.User = new ClaimsPrincipal(new ClaimsIdentity(new[] { new Claim(ClaimTypes.Name, Interlocked.Increment(ref _id).ToString()) }));
+                Connection.User = new ClaimsPrincipal(new GenericIdentity(Interlocked.Increment(ref _id).ToString()));
```
davidfowl: What's the deal with this change?

BrennanConroy: Unrelated to the Channel changes, it was from some local testing with Redis and needing the User to be authenticated

src/Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs
anurse: Because it doesn't infer. The whole point of this is you can infer the `T`. But inference only works with method type parameters. Hence [`Task.FromResult`](https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,11a386e7d7cae64a), [`Tuple.Create`](https://referencesource.microsoft.com/#mscorlib/system/tuple.cs,9124c4bea9ab0199), etc.

If we didn't care about inference, we'd just keep using the constructor.

davidfowl:
@BrennanConroy what's the deal with this?

BrennanConroy:
Waiting for CoreCLR to pass, fixed an issue with Xunit versions yesterday, now just waiting on an EF change that should go in today.

PR #129, opened 2017-01-15
Adding infrastructure to write node (a.k.a. unit) tests for ts client

moozzyk:
Github closed this PR when the base branch was deleted.


test/Microsoft.AspNetCore.Client.SignalR.TS.Tests/Connection.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+                    return Promise.resolve("");
+                }
+            }
+        } as ISignalROptions;
+
+        let connection = new Connection("http://tempuri.org", undefined, options);
+
+        try {
+            await connection.start();
+            expect(false).toBe(true);
```
davidfowl: ?

davidfowl: I prefer the fail helper

moozzyk: I will look into this.

moozzyk:
It does not slow down the build at the moment. There is just one new node module. The tests were not plugged into build system yet (next on my list) but when running manually they run fast. 
One thing that I want to look into is how to prevent building the ts-client multiple times when running build.cmd. Currently it builds a few times (most of the sample projects build on their own) and each build takes 10-12 secs. Fixing this will speed the build by at least 30 secs.

moozzyk:
I was able to remove dependency on jasmine-node.

anurse:
Have you tried creating VS TypeScript projects for these so they show up when opening the solution in VS? I often forget they exist :). Also, as we move towards MSBuild, it'll help with integrating the output of their build into the C# projects.

moozzyk:
I haven't looked into creating VS projects - IMO VS Code works much better with TypeScript than VS. I will add this to my list.

PR #138, opened 2017-01-19
some clean-up and refactoring

anurse:
* treat `/send` like a transport and run it closer to when the transports run
* refactor code to allow us to share one code path when looking up connection state, without funky behavior like having the `GetConnectionAsync` method write to the response
* refactor tests and add additional disconnect/error test cases.

No functional changes here, just some refactoring.

/cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
@@ -84,9 +84,8 @@ namespace Microsoft.AspNetCore.Sockets
             else if (context.Request.Path.StartsWithSegments(path + "/sse"))
             {
                 // We only need to provide the Input channel since writing to the application is handled through /send.
-                var sse = new ServerSentEventsTransport(connectionState.Application.Input, _loggerFactory);
-                await DoPersistentConnection(endpoint, sse, context, connectionState);
-                _manager.RemoveConnection(connectionState.Connection.ConnectionId);
+                var transport = new ServerSentEventsTransport(connectionState.Application.Input, _loggerFactory);
+                await ExecuteTransportAsync(context, endpoint, connectionState, transport);
             }
             else if (context.Request.Path.StartsWithSegments(path + "/poll"))
             {
@@ -104,7 +103,7 @@ namespace Microsoft.AspNetCore.Sockets
             if (!EnsureConnectionState(connectionState, context, transport.Name))
             {
                 // Changed transports mid-connection. Bad client!
-                context.Response.StatusCode = 400;
+                context.Response.StatusCode = StatusCodes.Status400BadRequest;
                 await context.Response.WriteAsync("Cannot change transports mid-connection");
                 return;
             }
@@ -124,7 +123,7 @@ namespace Microsoft.AspNetCore.Sockets
             if (!EnsureConnectionState(connectionState, context, longPolling.Name))
             {
                 // Changed transports mid-connection. Bad client!
-                context.Response.StatusCode = 400;
+                context.Response.StatusCode = StatusCodes.Status400BadRequest;
                 await context.Response.WriteAsync("Cannot change transports mid-connection");
                 return;
             }
@@ -238,7 +237,7 @@ namespace Microsoft.AspNetCore.Sockets
             if (string.Equals(connectionState.Connection.Metadata.Get<string>("transport"), WebSocketsTransport.TransportName))
             {
                 context.Response.StatusCode = StatusCodes.Status400BadRequest;
-                await context.Response.WriteAsync("Cannot send to a WebSockets connection");
+                await context.Response.WriteAsync("Cannot use '/send' to send messages to a WebSockets connection");
                 return;
             }
 
--- a/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
@@ -37,10 +37,6 @@ namespace Microsoft.AspNetCore.Sockets
             {
                 await ProcessNegotiate(context);
             }
-            else if (context.Request.Path.StartsWithSegments(path + "/send"))
-            {
-                await ProcessSend(context);
-            }
             else
             {
                 await ExecuteEndpointAsync(path, context, endpoint);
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -49,128 +45,141 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
 
-                // Mark the connection as active
-                state.Active = true;
+        private async Task ExecuteTransportAsync(HttpContext context, EndPoint endpoint, ConnectionState connectionState, IHttpTransport transport)
+        {
+            // Set up connection state
+            if (!EnsureConnectionState(connectionState, context, transport.Name))
+            {
+                // Changed transports mid-connection. Bad client!
+                context.Response.StatusCode = 400;
```
BrennanConroy: No `StatusCodes` value for 400?

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -49,128 +45,141 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
+            if (isWebSockets)
+            {
+                var transport = new WebSocketsTransport(connectionState.Application, _loggerFactory);
+                await ExecuteTransportAsync(context, endpoint, connectionState, transport);
+            }
+            else if (context.Request.Path.StartsWithSegments(path + "/sse"))
+            {
+                // We only need to provide the Input channel since writing to the application is handled through /send.
+                var sse = new ServerSentEventsTransport(connectionState.Application.Input, _loggerFactory);
+                await DoPersistentConnection(endpoint, sse, context, connectionState);
```
BrennanConroy: Call `ExecuteTransportAsync` here?

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -49,128 +45,141 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
-                    _logger.LogDebug("Establishing new Long Polling connection: {0}", state.Connection.ConnectionId);
+            var longPolling = new LongPollingTransport(connectionState.Application.Input, _loggerFactory);
 
-                    // This will re-initialize formatType metadata, but meh...
-                    state.Connection.Metadata["transport"] = LongPollingTransport.Name;
+            // Set up connection state
+            if (!EnsureConnectionState(connectionState, context, longPolling.Name))
+            {
+                // Changed transports mid-connection. Bad client!
+                context.Response.StatusCode = 400;
```
BrennanConroy: Same

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Sockets.Transports
 {
     public class WebSocketsTransport : IHttpTransport
     {
-        public static readonly string Name = "webSockets";
+        public static readonly string TransportName = "webSockets";
```
BrennanConroy: Why does websockets get to keep its static readonly string and the others don't

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -77,31 +44,20 @@ public class HttpConnectionDispatcherTests
...
-        public override Task OnConnectedAsync(Connection connection)
+        [Fact]
+        public async Task CannotUseSendEndpointWithWebSockets()
         {
-            throw new NotImplementedException();
+            var host = SocketsTestHost.CreateWithDefaultEndPoint();
+            var connectionState = host.CreateConnection(transportName: WebSocketsTransport.TransportName);
+            var result = await host.ExecuteRequestAsync("/send", queryString: $"?id={connectionState.Connection.ConnectionId}");
+            Assert.Equal(StatusCodes.Status400BadRequest, result.HttpContext.Response.StatusCode);
+            Assert.Equal("Cannot send to a WebSockets connection", Encoding.UTF8.GetString(result.ResponseBody));
```
BrennanConroy: Could we change this to something like `/send is not supported on WebSocket connections`

test/Microsoft.AspNetCore.Sockets.Tests/TestTaskExtensions.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Sockets.Tests
+{
+    public static class TestTaskExtensions
```
davidfowl: Is this just incase we write awful code that hangs forever?

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
```diff
@@ -24,27 +25,43 @@ public ServerSentEventsTransport(ReadableChannel<Message> application, ILoggerFa
...
                         {
                             await Send(context, message);
                         }
                     }
+
+                    _logger.LogTrace("Sent batch of {0} frames", counter);
                 }
+
+                // If the completion is faulted or cancelled, we want to manifest that error
+                _application.Completion.GetAwaiter().GetResult();
```
davidfowl: Same as above, I need to understand why this is important. What pieces of the system need to be notified of the exception and how does that flow.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -49,128 +45,141 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
 
         private async Task ExecuteEndpointAsync(string path, HttpContext context, EndPoint endpoint)
         {
-            // Server sent events transport
-            if (context.Request.Path.StartsWithSegments(path + "/sse"))
```
davidfowl: I think FXCop would complain about the cyclomatic complexity of this method... I honestly don't know if this refactoring makes it better or worse yet...

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -49,128 +45,141 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
-                var endpointTask = state.Connection.Metadata.Get<Task>("endpoint");
-                if (endpointTask == null)
-                {
-                    _logger.LogDebug("Establishing new Long Polling connection: {0}", state.Connection.ConnectionId);
+            var longPolling = new LongPollingTransport(connectionState.Application.Input, _loggerFactory);
 
-                    // This will re-initialize formatType metadata, but meh...
-                    state.Connection.Metadata["transport"] = LongPollingTransport.Name;
+            // Set up connection state
+            if (!EnsureConnectionState(connectionState, context, longPolling.Name))
```
davidfowl: Seems like this logic for ensuring the transport doesn't change mid connection can be moved up?

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
anurse: Just missed it, updating now

test/Microsoft.AspNetCore.Sockets.Tests/TestTaskExtensions.cs
anurse: Yes, which is not exactly unprecedented. I don't want any tests waiting indefinitely for things.

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -32,6 +32,9 @@ public LongPollingTransport(ReadableChannel<Message> application, ILoggerFactory
                 // This is for cases when the client reconnects see issue #27
                 if (!await _application.WaitToReadAsync(context.RequestAborted))
                 {
+                    // If the completion is faulted or cancelled, we want to manifest that error
+                    _application.Completion.GetAwaiter().GetResult();
```
anurse: Right now, if the app completes the channel with an error, that error vanishes into the ether. I can log it or rethrow it. Here, I chose to rethrow it. If you think logging would be better, then we can do that instead.

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
anurse: Because so far it's the only one we need to check in a situation where we don't have an instance. See HttpConnectionDispatcher.cs:L27

anurse: But I can add it to the others for consistency.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
anurse: The biggest problem is that there's no consistent transport-agnostic way to look up the connection because WebSockets allows lazy initialization of the connection. That's why refactoring to reduce duplication didn't really reduce complexity. It certainly didn't increase it based on my measurement.

anurse: Also Long Polling having such a complicated execution model. I'm not sure I even fully understand it yet, it's pretty sketchy right now.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
anurse: Except that it needs to be in each separate branch because we need to know what transport the connection is using before we can do it.

anurse: If we could consolidate `ExecuteLongPollingAsync` and `ExecuteTransportAsync` maybe.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -49,128 +45,140 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
-                {
-                    // Bad connection state. It's already set the response status code.
-                    return;
-                }
+                await ExecuteLongPollingAsync(context, endpoint, connectionState);
+            }
+            else if (context.Request.Path.StartsWithSegments(path + "/send"))
+            {
+                await ExecuteSendAsync(context, connectionState);
+            }
```
moozzyk: what happens if the url did not have any of the above?

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -32,6 +34,11 @@ public LongPollingTransport(ReadableChannel<Message> application, ILoggerFactory
                 // This is for cases when the client reconnects see issue #27
                 if (!await _application.WaitToReadAsync(context.RequestAborted))
                 {
+                    if(_application.Completion.IsFaulted)
+                    {
```
moozzyk: Should it be really 204 in this case?

test/Microsoft.AspNetCore.Sockets.Tests/TestTaskExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+            return self.GetAwaiter().GetResult();
+        }
+
+        public static Task WithTimeout(this Task self) => WithTimeout(self, TimeSpan.FromSeconds(5));
+
+        public static Task WithTimeout(this Task self, int millisecondsTimeout) => WithTimeout(self, TimeSpan.FromMilliseconds(millisecondsTimeout));
+
+        public static async Task WithTimeout(this Task self, TimeSpan timeout)
+        {
+            var cts = new CancellationTokenSource();
```
moozzyk: Why not just use `if (self != await Task.WhenAny(self, Task.Delay(timespan)) { throw new TimeoutException())} await self` ?

test/Microsoft.AspNetCore.Sockets.Tests/SocketsTestHost.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Sockets.Internal;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Sockets.Tests
+{
+    public static class SocketsTestHost
```
moozzyk: Should this be in `HttpConnectionDispatcherTests.cs`? It seems to be used only there.

test/Microsoft.AspNetCore.Sockets.Tests/SocketsRequestResult.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.AspNetCore.Http;
```
mikaelm12: Header

test/Microsoft.AspNetCore.Sockets.Tests/SocketsTestHost.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
```
mikaelm12: Header

test/Microsoft.AspNetCore.Sockets.Tests/TestTaskExtensions.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
```
mikaelm12: Header

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
davidfowl: The LongPolling logic is insane, but it totally makes sense 😄

test/Microsoft.AspNetCore.Sockets.Tests/TestTaskExtensions.cs
davidfowl: This is something we want to do for all tests though. I think @mikeharder added something to our xunit test infrastructure to allow timing out tests.

mikeharder: I started working on modifying xunit to support timeouts, but it was too difficult.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
anurse: Same thing as before, we fall through.

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
anurse: Probably not

anurse:
I'm gonna rebase on @davidfowl 's recent change and adjust. I think it'll simplify a bunch.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
moozzyk: but is this the correct behavior?

anurse:
very out of date

Issue #51, opened 2016-11-28
.NET Client To Do List


mikaelm12:
- [x] Add functional tests
- [x] ~Add generic overloads for the On method of the HubConnection~ - #311
- [x] Rename RpcConnection file to match HubConnection
- [x] Make sure we aren't getting the split messages issue as seen in https://github.com/aspnet/SignalR/issues/6
- [x] ~Clean up the public api~ - #311
- [x] Support WebSockets transport
- [x] ~Support SSE transport~ #309
- [x] ~handle cases where a method called from the server does not exist on the client~ - #312
- [x] ~client needs invoke overload for void methods~ - #311
- [x] ~re-usability of transports - currently the (longpolling transport) can be reused after it has been Disposed()~ #310
- [x] ~figure out ownership of transport - currently transports are passed to StartAsync and then Disposed in the Connection.Dispose() (maybe there should be a Stop() method?)~ -#310
- [x] re-entrancy - StartAsync on a transport should throw if the transport is already running
- [x] ~do we need to enable to close both Input/Output channels forcibly (example - right now if the user calls connection.Output.TryComplete(), connection.Input.Completion task will not be completed until the channel is drained) ?~ 
- [x] do we need Send method and Receive callback on Connection instead of exposing Input/Output - writing data/receiving data is really gross now with having to create messages and `PreservedBuffer`) - or make it easier to create/read from Message
- [x] do we need to track the state of the client? It is impossible to tell what state the client is in if the client is disconnected but the channel is not drained (this in turn makes it impossible to create the Closed event hanging on the Completion task)
- [x] do not expose channels (https://github.com/aspnet/SignalR/pull/141)
- [x] remove ContinueWith() (https://github.com/aspnet/SignalR/pull/141)
- [ ] consider removing the `Running` task from Transport - it's not part of interface and is a bit weird https://github.com/aspnet/SignalR/pull/141 - not doing at the moment - it is useful for testing.

moozzyk:
Created individual issues - closing.

PR #141, opened 2017-01-20
Pawelka/channels client

moozzyk:


src/Microsoft.AspNetCore.Sockets.Client/project.json
```diff
@@ -18,12 +18,20 @@
     "nowarn": [
       "CS1591"
     ],
-    "xmlDoc": true
+    "xmlDoc": true,
+    "compile": {
+      "include": [
```
moozzyk: We could create a separate common project to share this code but it felt too much for these 3 file (one of which we hope will make it to the channels)

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -2,56 +2,45 @@
...
 using System.Net.Http;
 using System.Threading.Tasks;
+using System.Threading.Tasks.Channels;
+using Microsoft.AspNetCore.Sockets.Internal;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
-    public class Connection : IPipelineConnection
+    public class Connection : IChannelConnection<Message>
```
moozzyk: I have mixed feelings about Connection being IChannelConnection
- exposing Message
- stopping the connection is ugly and has loose ends (especially you need to drain the channel in order to make the Completion task complete)

I updated issue https://github.com/aspnet/SignalR/issues/51 but we need more to do more thinking here (or maybe it is fine at the low level?)

src/Microsoft.AspNetCore.Sockets.Client/Message.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO.Pipelines;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public struct Message : IDisposable
+    {
+        public Format MessageFormat { get; }
```
moozzyk: If we want the connection to be IChannel<Message> then we should add make using this struct more convenient (e.g. add Message(string data)/Message(byte[] data) ctors etc)

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: Related/rephrased - it is impossible to tell currently what state the client is in when it is disconnected but the channel is not drained

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -98,9 +96,21 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
             }
 
-            // Write the invocation to the stream
+            var ms = new MemoryStream();
+            await _adapter.WriteMessageAsync(descriptor, ms, cancellationToken);
+
             _logger.LogInformation("Sending Invocation #{0}", descriptor.Id);
-            await _adapter.WriteMessageAsync(descriptor, _stream, cancellationToken);
+
+            // TODO: Format.Text - who, where and when decides about the format of outgoing messages
```
BrennanConroy: Issue 132

moozzyk:
Travis fails since apparently an older version of node that the one required by typescript 2.5 is installed. Will look into it.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -98,9 +96,21 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
+            var message = new Message(ReadableBuffer.Create(ms.ToArray()).Preserve(), Format.Text);
```
moozzyk: as pointed by @BrennanConroy Message needs to be disposed.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -39,104 +41,98 @@ public void Dispose()
...
+        private async Task SendMessages(Uri sendUrl, CancellationToken sendCancellationToken)
         {
-            using (cancellationToken.Register(() => _pipeline.Input.Complete()))
+            try
             {
-                try
+                while (await _toFromConnection.Input.WaitToReadAsync(sendCancellationToken))
                 {
-                    while (!cancellationToken.IsCancellationRequested)
+                    Message message;
```
BrennanConroy: Same here, we should dispose?

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using Microsoft.Extensions.Logging;
+using Moq;
+using Moq.Protected;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Sockets.Client.Tests
+{
+    public class ConnectionTests
+    {
+
```
BrennanConroy: nit:

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: Yes

moozzyk: Updated everywhere we Read messages.

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            using (var longPollingTransport = new LongPollingTransport(httpClient, new LoggerFactory()))
+            using (var connection = await Connection.ConnectAsync(new Uri("http://fakeuri.org/"), longPollingTransport, httpClient))
+            {
+                Assert.False(connection.Input.Completion.IsCompleted);
+
+                connection.Output.TryWrite(
+                    new Message(ReadableBuffer.Create(new byte[] { 1, 1, 2, 3, 5, 8 }).Preserve(), Format.Binary));
+
+                Assert.Equal(sendTcs.Task, await Task.WhenAny(Task.Delay(10000), sendTcs.Task));
+                Assert.Equal(new byte[] { 1, 1, 2, 3, 5, 8 }, sendTcs.Task.Result);
```
BrennanConroy: Extract to `var data` or something

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            using (var httpClient = new HttpClient(mockHttpHandler.Object))
+            using (var longPollingTransport = new LongPollingTransport(httpClient, new LoggerFactory()))
+            using (var connection = await Connection.ConnectAsync(new Uri("http://fakeuri.org/"), longPollingTransport, httpClient))
+            {
+                Assert.False(connection.Input.Completion.IsCompleted);
+
+                connection.Output.TryWrite(
+                    new Message(ReadableBuffer.Create(new byte[] { 1, 1, 2, 3, 5, 8 }).Preserve(), Format.Binary));
+
+                Assert.Equal(sendTcs.Task, await Task.WhenAny(Task.Delay(10000), sendTcs.Task));
```
BrennanConroy: 10 seconds seems a bit much

moozzyk: Will fix. I set it to 10 secs when debugging.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -23,7 +24,8 @@ public class LongPollingTransport : ITransport
         private readonly CancellationTokenSource _senderCts = new CancellationTokenSource();
         private readonly CancellationTokenSource _pollCts = new CancellationTokenSource();
 
-        private IPipelineConnection _pipeline;
+        private IChannelConnection<Message> _toFromConnection;
+
```
mikaelm12: nit: remove new line

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -0,0 +1,167 @@
...
+                });
+
+            using (var httpClient = new HttpClient(mockHttpHandler.Object))
+            using (var longPollingTransport = new LongPollingTransport(httpClient, new LoggerFactory()))
+            using (var connection = await Connection.ConnectAsync(new Uri("http://fakeuri.org/"), longPollingTransport, httpClient))
+            {
+                Assert.False(connection.Input.Completion.IsCompleted);
+                connection.Output.TryComplete();
+
+                var completionTask = Task.WhenAny(Task.Delay(1000), connection.Input.Completion);
```
BrennanConroy: This seems odd, if we need to read messages to complete the task then `Task.Delay` will always be returned?

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -39,104 +41,101 @@ public void Dispose()
...
                     else
                     {
-                        // Write the data to the output
-                        var buffer = _pipeline.Output.Alloc();
-                        var stream = new WriteableBufferStream(buffer);
-                        await response.Content.CopyToAsync(stream);
-                        await buffer.FlushAsync();
+                        var ms = new MemoryStream();
+                        await response.Content.CopyToAsync(ms);
+                        var message = new Message(ReadableBuffer.Create(ms.ToArray()).Preserve(), Format.Text);
```
mikaelm12: Should this message be disposed as well?

moozzyk: No. We are sending it and the receiver is responsible for disposing it. Otherwise we could dispose something they are using.

test/Microsoft.AspNetCore.Sockets.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -0,0 +1,167 @@
...
+using Xunit;
+
+namespace Microsoft.AspNetCore.Sockets.Client.Tests
+{
+    public class LongPollingTransportTests
+    {
+        [Fact]
+        public async Task LongPollingTransportStopsPollAndSendLoopsWhenTransportDisposed()
+        {
+            var mockHttpHandler = new Mock<HttpMessageHandler>();
```
mikaelm12: This seems to be repeated. Can we move some of this set up into a helper method?

moozzyk: There is no point in just moving this single line and the configuration varies (the returned response) across the tests so I decided to leave it as is.

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
moozzyk: Notice we don't await the task here but below, after we have drained the channel. Perhaps I should call it `whenAny` task and not the `completionTask`.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -132,11 +142,20 @@ public void Dispose()
...
             try
             {
-                while (!cancellationToken.IsCancellationRequested)
+                while (await _connection.Input.WaitToReadAsync(cancellationToken))
                 {
-                    // This is a little odd... we want to remove the InvocationRequest once and only once so we pull it out in the callback,
-                    // and stash it here because we know the callback will have finished before the end of the await.
-                    var message = await _adapter.ReadMessageAsync(_stream, _binder, cancellationToken);
+                    Message incomingMessage;
+                    if (!_connection.Input.TryRead(out incomingMessage))
```
davidfowl: This should be in a while loop

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
davidfowl: I've been thinking about this a bit and I wonder if there's an upside in exposing the channel interfaces on the client side. The benefit in theory is the ability to share code between client and server but we aren't really doing that. We should make sure the design is right here before going too far with it.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -39,104 +41,101 @@ public void Dispose()
...
-            {
-                _pollCts.Cancel();
-                return TaskCache.CompletedTask;
-            });
+            _toFromConnection = toFromConnection;
 
             // Start sending and polling
-            _poller = Poll(Utils.AppendPath(url, "poll"), _pollCts.Token);
-            _sender = SendMessages(Utils.AppendPath(url, "send"), _senderCts.Token);
+            _poller = Poll(Utils.AppendPath(url, "poll"), _pollCts.Token).ContinueWith(_ => _senderCts.Cancel());
```
davidfowl: Why do we need to cancel the CTS? Who is it for?

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -39,104 +41,101 @@ public void Dispose()
         {
             _senderCts.Cancel();
             _pollCts.Cancel();
-            _pipeline?.Dispose();
         }
 
-        public Task StartAsync(Uri url, IPipelineConnection pipeline)
+        public Task StartAsync(Uri url, IChannelConnection<Message> toFromConnection)
```
davidfowl: This naming is throwing me off. Can we name things similar to the server? Transport and Application?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: Why? No one else is reading so it should always succeed. If it fails (what would be the reason?) is there a guarantee that a subsequent call will succeed because if there isn't then we are in tight infinite loop that can't be completed (unless we also check the cancellation token inside the loop).

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: Yes. We need to talk about this but this would be separate from this commit.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: It actually helped me (especially at the beginning) but can change it.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: It is to exit the sending loop. They run independently and if polling fails/completes we need to stop the loop that is waiting on the user's requests.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
davidfowl: Multiple messages can come in. It's not about failure.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
davidfowl: Why not just complete the channel from the outside? Why the cancellation token?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: I see. You basically want to save a call to `await _connection.Input.WaitToReadAsync`? Because even now multiple messages will work only we will have to call WaitToReadAsync for each message.

davidfowl: Right, it's the same thing we do on the server.

moozzyk: Will do.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: The problem is that even if you TryComplete from outside the send loop will continue to run until it drains the channel. This means that until the channel is drained we will keep sending requests to the server even though the connection is already disconnected. Cancellation token prevents that. 
(Note: the channel from application to transport is not currently drained - I think it should be done by the side writing to the channel i.e. HubConnection in case of hubs or the user himself when using Sockets but this is one of the issues I wanted to talk about)
I also think that we actually could use just one cancellation tokens for both send and poll loops.

test/Microsoft.AspNetCore.Sockets.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            using (var longPollingTransport = new LongPollingTransport(httpClient, new LoggerFactory()))
+            {
+                var connectionToTransport = Channel.CreateUnbounded<Message>();
+                var transportToConnection = Channel.CreateUnbounded<Message>();
+                var channelConnection = new ChannelConnection<Message>(connectionToTransport, transportToConnection);
+                await longPollingTransport.StartAsync(new Uri("http://fakeuri.org"), channelConnection);
+
+                connectionToTransport.Out.Complete();
+
+                Assert.Equal(longPollingTransport.Running, await Task.WhenAny(Task.Delay(1000), longPollingTransport.Running));
```
davidfowl: Are you going to update it to use the new helpers?

moozzyk: I thought about this. I was thinking about creating a Testing.Common project that contains this kind of things since we would want to use it across different test projects in this repo.

moozzyk:
🆙 :date:

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -20,12 +21,10 @@ public class LongPollingTransport : ITransport
...
-        private readonly CancellationTokenSource _senderCts = new CancellationTokenSource();
-        private readonly CancellationTokenSource _pollCts = new CancellationTokenSource();
-
-        private IPipelineConnection _pipeline;
+        private IChannelConnection<Message> _application;
         private Task _sender;
         private Task _poller;
+        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
 
         public Task Running { get; private set; }
```
davidfowl: Who is this task for? It's not part of the interface right?

moozzyk: No, it isn't. It was there and I left it because it is useful for testing but it's a bit weird.

moozzyk:
@davidfowl - This change was to convert from pipelines to channels but I agree we need to redesign the client a little bit and remove ContinueWith() usages.

moozzyk:
I updated https://github.com/aspnet/SignalR/issues/51 with follow up items from this PR.

moozzyk:
Fixed in c997ea81650420ed42fca3d1bdedb03527ad711f

PR #142, opened 2017-01-22
Add transport type constants and factory to extract creation responsability and magic strings from Connection.ts class

CarlosLanderas:
Add transport type constants and factory to extract creation responsability and magic strings from Connection.ts class

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
+++ b/src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
@@ -18,6 +18,7 @@ export class Connection {
     private connectionId: string;
     private httpClient: IHttpClient;
     private transport: ITransport;
+    private transportFactory: TransportFactory;
     private dataReceivedCallback: DataReceived = (data: any) => { };
     private connectionClosedCallback: ConnectionClosed = (error?: any) => { };
 
@@ -26,6 +27,7 @@ export class Connection {
         this.queryString = queryString;
         this.httpClient = options.httpClient || new HttpClient();
         this.connectionState = ConnectionState.Disconnected;
+        this.transportFactory = new TransportFactory(this.httpClient);
     }
 
     async start(transportName: string = TransportType.webSockets): Promise<void> {
@@ -33,7 +35,7 @@ export class Connection {
             throw new Error("Cannot start a connection that is not in the 'Disconnected' state");
         }
 
-        this.transport = new TransportFactory(this.httpClient).create(transportName);        
+        this.transport = this.transportFactory.create(transportName);        
         this.transport.onDataReceived = this.dataReceivedCallback;
         this.transport.onError = e => this.stopConnection(e);
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
+++ b/src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
@@ -3,6 +3,7 @@ import { ITransport, WebSocketTransport, ServerSentEventsTransport, LongPollingT
 import { IHttpClient, HttpClient } from "./HttpClient"
 import { ISignalROptions } from "./ISignalROptions"
 import { TransportType } from './TransportType';
+import { TransportFactory } from './TransportFactory';
 
 enum ConnectionState {
     Disconnected,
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
```diff
@@ -26,12 +28,12 @@ export class Connection {
...
     }
 
-    async start(transportName: string = 'webSockets'): Promise<void> {
+    async start(transportName: string = TransportType.webSockets): Promise<void> {
         if (this.connectionState != ConnectionState.Disconnected) {
             throw new Error("Cannot start a connection that is not in the 'Disconnected' state");
         }
 
-        this.transport = this.createTransport(transportName);
+        this.transport = new TransportFactory(this.httpClient).create(transportName);
```
davidfowl: Shouldn't the transport factory be passed into the connection object? Or created in the ctor (since we don't quite have pluggability here as yet).

CarlosLanderas: Since httpClient is created considering the ISignalROptions in Connection class, I've changed it to create the instance in the Ctor.

src/Microsoft.AspNetCore.SignalR.Client.TS/TransportFactory.ts
```diff
@@ -0,0 +1,23 @@
+import { ITransport, WebSocketTransport, ServerSentEventsTransport, LongPollingTransport } from "./Transports"
```
moozzyk: I don't like this being exposed as a separate module. I don't think it is of any use outside the connection (that's why createTransport was private)

src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
```diff
@@ -16,6 +18,7 @@ export class Connection {
     private connectionId: string;
     private httpClient: IHttpClient;
     private transport: ITransport;
+    private transportFactory: TransportFactory;
```
moozzyk: Not sure why this is a class variable. We create transport only in one place...

src/Microsoft.AspNetCore.SignalR.Client.TS/TransportType.ts
```diff
@@ -0,0 +1,5 @@
+export const TransportType =  {
+    webSockets : 'webSockets',
+    serverSentEvents : 'serverSentEvents',
+    longPolling : 'longPolling'
+};
```
moozzyk: nit: add new line

src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
```diff
@@ -24,14 +27,15 @@ export class Connection {
...
     }
 
-    async start(transportName: string = 'webSockets'): Promise<void> {
+    async start(transportName: string = TransportType.webSockets): Promise<void> {
         if (this.connectionState != ConnectionState.Disconnected) {
             throw new Error("Cannot start a connection that is not in the 'Disconnected' state");
         }
 
-        this.transport = this.createTransport(transportName);
+        this.transport = this.transportFactory.create(transportName);
```
moozzyk: nit: remove trailing spaces

src/Microsoft.AspNetCore.SignalR.Client.TS/Connection.ts
moozzyk: I have some doubts here 
 - we create transports only in one place - why do we need the factory (which is public module)? What's the advantage over the factory function that did just that, was "private" and did not add more to the object state?
 - because the factory is now created in the ctor we need to keep track of it in yet another class variable (referred to as "object state" above)
 - I am also sure if the factory will help in case of custom transports. I would imagine that it would be much easier for the user just to create a custom transport and pass it (when supported) instead of creating a factory that creates a custom transport especially if you think about the dependencies custom transport will require

CarlosLanderas:
Thanks for the review moozzyk, maybe the best idea is closing the pull request, because the only thing that would be keeped is the module with constants. I will keep an eye in the evolution of your code design next months.

moozzyk:
@CarlosLanderas - thank you. I do appreciate your effort. I think this code will evolve when we start adding support for custom transports and formatters when we may bring some of your ideas back to life.

PR #143, opened 2017-01-23
Allow processing of incoming invocations in parallel

davidfowl:
- Don't wait on the response of an invocation to pick up
the next message from the channel.

PS: I need to write a test

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
@@ -168,38 +168,45 @@ namespace Microsoft.AspNetCore.SignalR
                         _logger.LogDebug("Received hub invocation: {invocation}", invocationDescriptor);
                     }
 
-                    InvocationResultDescriptor result;
-                    Func<Connection, InvocationDescriptor, Task<InvocationResultDescriptor>> callback;
-                    if (_callbacks.TryGetValue(invocationDescriptor.Method, out callback))
-                    {
-                        result = await callback(connection, invocationDescriptor);
-                    }
-                    else
-                    {
-                        // If there's no method then return a failed response for this request
-                        result = new InvocationResultDescriptor
-                        {
-                            Id = invocationDescriptor.Id,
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -168,38 +168,45 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
-                    {
-                        // If there's no method then return a failed response for this request
-                        result = new InvocationResultDescriptor
-                        {
-                            Id = invocationDescriptor.Id,
-                            Error = $"Unknown hub method '{invocationDescriptor.Method}'"
-                        };
+                    // Don't wait on the result of execution, continue processing other
+                    // incoming messages on this connection.
+                    var ignore = Execute(connection, invocationAdapter, invocationDescriptor);
```
moozzyk: what happens to exceptions?

(side note: I am not even sure if/how this compiles due to the unused variable)

davidfowl: > what happens to exceptions?

Which ones?

> (side note: I am not even sure if/how this compiles due to the unused variable)

Compiles fine

davidfowl: We could say that unexpected exceptions end the connection.

moozzyk: The most prominent would be when the `callback` throws.

moozzyk: unexpected exceptions should end the connection indeed. but because we are not awaiting the task returned by `Execute` it will be faulted and the exception will not propagate right?

davidfowl: Right. Callbacks don't throw in normal operation.

davidfowl:
🆙 📅

moozzyk:
:ship::it:

Issue #71, opened 2016-12-11
Longer messages get split into chunks of 1024 characters


charlieamer:
I was trying to call a function from server to a JS client with parameter string which is around 5000 characters long. The message, however, was split into chunks of 1024 bytes and sent multiple times (that is, firstly, chunk 0-1024 is sent, then 1024-2048, etc). This resulted in a corrupt JSON, as you can imagine, library is trying to read message by message, and it gets chunks instead. This happens on WebSocket transport and ServerSideEvents transport (long polling transport works).
Is there a way to change the max length of text that I can send, or will you upgrade the client that he knows how to receive these multipart messages.

davidfowl:
This library is still under heavy development and isn't ready for real usage yet.

moozzyk:
https://github.com/aspnet/SignalR/issues/6

PR #147, opened 2017-01-24
Fixing split messages error if message length are too big

### LINKED ISSUE #6: Occasionally getting split messages in ChatSample
Sending a bunch of requests will occasionally result in a split message

tsclient
@anurse

grandchamp:
This fixes #6 #71


moozzyk:
Do you still see this? This should be fixed on the server side now...

Issue #123, opened 2017-01-12
Allow the endpoint to set transport state


davidfowl:
We need to execute the EndPoint before executing the transport so that user code can affect the transport.

davidfowl:
Done!

PR #148, opened 2017-01-24
Clean up disposal of connection state

davidfowl:
- Removed IDisposable and added a DisposeAsync method to ConnectionState
- Added ApplicationTask and TransportTask to ConnectionState as first class
properties so that it is easy to see (in a process dump or debugger) the
outstanding tasks that Sockets is keeping track of on a per connection basis.
- Swapped transport and application initialization to enable https://github.com/aspnet/SignalR/issues/123 

Sorry @anurse 😄

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
+++ b/src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
@@ -2,16 +2,18 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.Sockets.Internal
 {
-    public class ConnectionState : IDisposable
+    public class ConnectionState
     {
         public Connection Connection { get; set; }
         public IChannelConnection<Message> Application { get; }
 
-        // These are used for long polling mostly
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
```diff
@@ -22,10 +24,25 @@ public ConnectionState(Connection connection, IChannelConnection<Message> applic
...
+            if (TransportTask.IsFaulted)
+            {
+                Application.Output.TryComplete(TransportTask.Exception.InnerException);
+            }
+
             Connection.Dispose();
             Application.Dispose();
+
+            // REVIEW: Add a timeout so we don't wait forever
+            await Task.WhenAll(ApplicationTask, TransportTask);
```
moozzyk: I think this might be causing test hangs in my branch... if a channel is not drained the task will never complete. I am not sure if timeout is good enough.

davidfowl: The channels are both disposed before waiting. The only way it would hang forever is if the code isn't awaiting the channel.

moozzyk: oops - Dispose was not marked green...

PR #149, opened 2017-01-24
Fixing test hangs, adding timeouts.

moozzyk:


moozzyk:
Note: there isn't a good general way to signal that a connection was connected and is tracked by the lifetime manager. I am not sure though if one is needed besides our tests though. For now this should fix the hangs and we can take design a generic solution if needed separately.

/cc @BrennanConroy

Issue #27, opened 2016-11-15
Reconnects


davidfowl:
If a request comes in with the same connection id for long polling (only), close the previous poll with a zero byte read. We have to await the previous transport task so make sure it unwinds properly. 

For other transport, reject connections with a 409 for an existing connection id.

davidfowl:
Fixed

Issue #4, opened 2016-11-03
Timed out poll connections kill the connection


moozzyk:
If a poll connection times out the server throws: 

```
System.InvalidOperationException occurred
  HResult=-2146233079
  Message=Concurrent reads are not supported.
  Source=System.Private.CoreLib
  StackTrace:
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at Channels.Channel.Channels.IReadableBufferAwaiter.GetResult()
       at Microsoft.AspNetCore.Sockets.LongPolling.<ProcessRequestAsync>d__3.MoveNext() in C:\source\NetworkingProtoype\src\Microsoft.AspNetCore.Sockets\LongPolling.cs:line 21
  InnerException: 
```

and kills the connection. The client restarts the poll only to get: 

GET http://localhost:5000/hubs/poll?id=87085e15-6bb5-492e-850b-f7965c7e9bbc&87085e15-6bb5-492e-850b-f7965c7e9bbc 500 (Internal Server Error)

davidfowl:
Fixed

PR #151, opened 2017-01-25
Various fixes in HttpConnectionDispatcher

davidfowl:
- The connection state object is manipulated by multiple parties in a non thread safe way. This change introduces a semaphore that should be used by anyone updating or reading the connection state. 
- Handle cases where there's an active request for a connection id and another incoming request for the same connection id, sse and websockets 409 and long polling kicks out the previous connection (https://github.com/aspnet/SignalR/issues/27 and https://github.com/aspnet/SignalR/issues/4)
- Handle requests being processed for disposed connections. There was a race where the background thread could remove and clean up the connection while it was about to be processed.
- Synchronize between the background scanning thread and the request threads when updating the connection state.
- Added `DisposeAndRemoveAsync` to the connection manager that handles`DisposeAsync` throwing and properly removes connections from connection tracking.
- Added Start to ConnectionManager so that testing is easier (background timer doesn't kick in unless start is called).
- Added RequestId to connection state for easier debugging and correlation (can easily see which request is currently processing the logical connection).
- Added tests

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
+++ b/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
@@ -16,7 +16,6 @@ namespace Microsoft.AspNetCore.Sockets
     {
         private readonly ConcurrentDictionary<string, ConnectionState> _connections = new ConcurrentDictionary<string, ConnectionState>();
         private Timer _timer;
-        private volatile bool _running;
         private readonly ILogger<ConnectionManager> _logger;
 
         public ConnectionManager(ILogger<ConnectionManager> logger)
@@ -28,7 +27,7 @@ namespace Microsoft.AspNetCore.Sockets
         {
             if (_timer == null)
             {
-                _timer = new Timer(Scan, this, 0, 1000);
+                _timer = new Timer(Scan, this, TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1));
             }
         }
 
@@ -78,15 +77,11 @@ namespace Microsoft.AspNetCore.Sockets
 
         private void Scan()
         {
-            if (_running)
-            {
-                return;
-            }
+            // Pause the timer while we're running
+            _timer.Change(Timeout.Infinite, Timeout.Infinite);
 
             try
             {
-                _running = true;
-
                 // Scan the registered connections looking for ones that have timed out
                 foreach (var c in _connections)
                 {
@@ -116,7 +111,8 @@ namespace Microsoft.AspNetCore.Sockets
             }
             finally
             {
-                _running = false;
+                // Resume once we finished processing all connections
+                _timer.Change(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1));
             }
         }
 
--- a/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
+++ b/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
@@ -80,8 +80,6 @@ namespace Microsoft.AspNetCore.Sockets
                             // REVIEW: Should we keep firing and forgetting this?
                             var ignore = s.DisposeAsync();
                         }
-
-                        c.Value.Status = ConnectionState.State.Disposed;
                     }
                     finally
                     {
--- a/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
+++ b/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
@@ -8,6 +8,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Channels;
 using Microsoft.AspNetCore.Sockets.Internal;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Sockets
 {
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -65,38 +78,79 @@ private static void Scan(object state)
...
-            foreach (var c in _connections)
+            if (_running)
             {
-                if (!c.Value.Active && (DateTimeOffset.UtcNow - c.Value.LastSeenUtc).TotalSeconds > 5)
+                return;
+            }
+
+            try
+            {
+                _running = true;
```
anurse: If this can actually happen simultaneously, you need `cmpexchg`. There's a race here otherwise. But I'm not sure how it can happen simultaneously, is it just in situations where the `Scan` takes more than one Timer interval? If so, suspend the timer when you start or something.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -65,38 +78,79 @@ private static void Scan(object state)
...
                     }
                 }
             }
+            finally
+            {
+                _running = false;
+            }
         }
 
         public void CloseConnections()
```
anurse: y u no async?

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -65,38 +78,79 @@ private static void Scan(object state)
...
             {
-                ConnectionState s;
-                if (_connections.TryRemove(c.Key, out s))
-                {
-                    tasks.Add(s.DisposeAsync());
-                }
+                tasks.Add(DisposeAndRemoveAsync(c.Value));
             }
 
             Task.WaitAll(tasks.ToArray(), TimeSpan.FromSeconds(5));
```
anurse: `WaitAll`. My eyes.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -65,38 +78,79 @@ private static void Scan(object state)
...
+                    }
+                    finally
                     {
-                        // REVIEW: Should we keep firing and forgetting this?
-                        var ignore = s.DisposeAsync();
+                        c.Value.Lock.Release();
+                    }
+
+                    // Once the decision has been made to to dispose we don't check the status again
+                    if (status == ConnectionState.ConnectionStatus.Inactive && (DateTimeOffset.UtcNow - lastSeenUtc).TotalSeconds > 5)
```
anurse: There's a race here, but I suppose we don't have to get too worked up about it... It just feels odd to be adding a race just to save a few microseconds inside the lock. We can make the determination that we're going to dispose and change the state to "Disposing" or something inside the lock and still do the actual Dispose outside the lock. That prevents the race.

src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
```diff
@@ -26,23 +34,54 @@ public ConnectionState(Connection connection, IChannelConnection<Message> applic
...
+            finally
             {
-                Application.Output.TryComplete(TransportTask.Exception.InnerException);
+                Lock.Release();
             }
 
-            Connection.Dispose();
-            Application.Dispose();
-
             // REVIEW: Add a timeout so we don't wait forever
```
anurse: Just add it now :). All it takes is a Task.Delay (and maybe a check to fire TimeoutException if the delay task is the one that fires first).

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -50,14 +47,17 @@ public LongPollingTransport(ReadableChannel<Message> application, ILoggerFactory
             }
             catch (OperationCanceledException)
             {
-                // Suppress the exception
+                if (!context.RequestAborted.IsCancellationRequested)
+                {
+                    _logger.LogInformation("Terminating Long Polling connection by sending 204 response.");
+                    context.Response.StatusCode = 204;
```
anurse: `StatusCodes` constant?

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
davidfowl: Dispose is synchronous brah. At least there's a timeout 😄

src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
davidfowl: nooooo, later.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
davidfowl: Will change to do a compare exchange https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Core/Transports/TransportHeartBeat.cs#L171. 

> is it just in situations where the Scan takes more than one Timer interval? If so, suspend the timer when you start or something.

Yes, you mean `timer.Change(Timeout.Infinite, Timeout.Infinite);`?

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
davidfowl: I originally had that but I wanted the state transition to happen in `DisposeAsync`. We don't care about the race here

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
davidfowl: Happens during shutdown of the application, disposing is always synchronous.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
anurse: Basically, yeah. But the compare-exchange is fine.

src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
anurse: lazy

PR #153, opened 2017-01-25
Reduced constructor boilerplate assignment

herecydev:


moozzyk:
I am not in favor of this change. The classes are not simple data holder classes and I find it more difficult to read if I want to quickly see what private members the class has. Granted, I am not used to it so can be convinced.

herecydev:
I think that's a fair assessment, it's definitely more beneficial when all of the params conform. I prefer less boilerplate to read through, but more than happy with whatever.

PR #159, opened 2017-01-25
Handle misbehaving user code

davidfowl:
- Execute EndPoint logic on a threadpool thread
- Turn synchronous exceptions into async ones to unify the
error handling
- Added tests

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
@@ -290,7 +290,7 @@ namespace Microsoft.AspNetCore.Sockets
             // connection and transport
             await AwaitableThreadPool.Yield();
 
-            // Running this in ann async method turns sync exceptions into async ones
+            // Running this in an async method turns sync exceptions into async ones
             await endpoint.OnConnectedAsync(connection);
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -284,6 +284,16 @@ private ConnectionState CreateConnection(HttpContext context)
             await _manager.DisposeAndRemoveAsync(state);
         }
 
+        private async Task ExecuteApplication(EndPoint endpoint, Connection connection)
+        {
+            // Jump onto the thread pool thread so blocking user code doesn't block the setup of the 
+            // connection and transport
+            await AwaitableThreadPool.Yield();
+
+            // Running this in ann async method turns sync exceptions into async ones
```
anurse: "an" async method

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -284,6 +284,16 @@ private ConnectionState CreateConnection(HttpContext context)
             await _manager.DisposeAndRemoveAsync(state);
         }
 
+        private async Task ExecuteApplication(EndPoint endpoint, Connection connection)
+        {
+            // Jump onto the thread pool thread so blocking user code doesn't block the setup of the 
+            // connection and transport
+            await AwaitableThreadPool.Yield();
```
anurse: Is this really clearer than using `Task.Run` or are you just trying to be clever? It seems like unnecessary cleverness hiding the actual code to me.

davidfowl: Yes, it is absolutely 100% clearer. See https://github.com/dotnet/corefx/issues/15490.

Issue #152, opened 2017-01-25
'Parameter count mismatch' after invoking hub from typescript client


JonasMH:
I'm reporting what i think is a bug, where after invoking a hub the server responds with 'Parameter count mismatch'.

The hub:

```
public class NotificationHub : Hub, INotificationHub
{
	private readonly IHubNotificationService _service;

	public NotificationHub(IHubNotificationService service)
	{
		_service = service;
	}

	public override Task OnConnectedAsync()
	{
		return Task.CompletedTask;
	}

	public override Task OnDisconnectedAsync(Exception exception)
	{
		_service.ClientDisconnected(Context.ConnectionId);
		return Task.CompletedTask;
	}

	public Task Subscribe(string topic)
	{
		string id = _service.ClientSubscribed(new SubscriptionUpdate(
			topic,
			Context.User.Identity.Name,
			Context.ConnectionId));

		return Task.CompletedTask;
	}
}
```

Client side (angular2, typescript): 

```
public subscribe(topic: string): void {
  this.hubConnection.invoke("subscribe", {topic: topic});
}
```

The hub's Constructor, OnConnectedAsync and OnDisconnectedAsync gets called as expected when a server connects and disconnects, but as soon as i try to invoke it, the following happens (websocket frames):
Request: `{"Id":"0","Method":"subscribe","Arguments":[{"topic":"/someTopic/"}]}`
Response: `{"Result":null,"Error":"Parameter count mismatch.","Id":"0"}`

The connection does not get shut down or anything, but i get the 'Parameter count mismatch' and the subscribe method on the hub is never called. Any suggestions?

davidfowl:
Try changing the code to do:

```C#
this.hubConnection.invoke("subscribe", topic);
```

You're not passing an object on the server side, you're passing a string.

moozzyk:
@JonasMH - can you provide a stack trace from the server. While what @davidfowl says is correct - you have a type mismatch - something else seems going on here. I tried to repro this and the server threw a different exception and close the connection. The exception I received was a parse error because the expected parameter type is String so SignalR tries to read the value as string but it encountered a json object.

JonasMH:
@moozzyk On the server nothing gets thrown, it's just running as normal
@davidfowl Just tried and nothing changed, still the same response as before.

Here's some more information that may be helpful:"

Microsoft.AspNetCore.SignalR version 1.0.0-preview1-23180
netcoreapp1.0
win10-x64
[...]

PR #162, opened 2017-01-26
Fixed parameter count mismatch when invoking methods with wrong case

davidfowl:
- Hub methods were being tracked with 2 dictionaries, one for parameter names
the other for callbacks. This change introduces a single dictionary that stores
the hub name to a HubMethodDescriptor. That descriptor stores the parameter types
and method info for the bound hub method.
- The callback is now just an invoke method on the HubEndPoint itself.
- Added tests for case sensitivity in hub method names

https://github.com/aspnet/SignalR/issues/152

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
@@ -246,7 +246,7 @@ namespace Microsoft.AspNetCore.SignalR
 
         private async Task<InvocationResultDescriptor> Invoke(HubMethodDescriptor descriptor, Connection connection, InvocationDescriptor invocationDescriptor)
         {
-            var invocationResult = new InvocationResultDescriptor()
+            var invocationResult = new InvocationResultDescriptor
             {
                 Id = invocationDescriptor.Id
             };
--- a/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
@@ -29,9 +29,7 @@ namespace Microsoft.AspNetCore.SignalR
 
     public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub : Hub<TClient>
     {
-        private readonly Dictionary<string, Func<Connection, InvocationDescriptor, Task<InvocationResultDescriptor>>> _callbacks
-            = new Dictionary<string, Func<Connection, InvocationDescriptor, Task<InvocationResultDescriptor>>>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, Type[]> _paramTypes = new Dictionary<string, Type[]>();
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -29,9 +29,7 @@ public class HubEndPoint<THub> : HubEndPoint<THub, IClientProxy> where THub : Hu
 
     public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub : Hub<TClient>
     {
-        private readonly Dictionary<string, Func<Connection, InvocationDescriptor, Task<InvocationResultDescriptor>>> _callbacks
-            = new Dictionary<string, Func<Connection, InvocationDescriptor, Task<InvocationResultDescriptor>>>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, Type[]> _paramTypes = new Dictionary<string, Type[]>();
+        private readonly Dictionary<string, HubMethodDescriptor> _methods = new Dictionary<string, HubMethodDescriptor>(StringComparer.OrdinalIgnoreCase);
```
moozzyk: Why do we want method names to be case insensitive?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -246,6 +244,59 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
             }
         }
 
+        private async Task<InvocationResultDescriptor> Invoke(HubMethodDescriptor descriptor, Connection connection, InvocationDescriptor invocationDescriptor)
+        {
+            var invocationResult = new InvocationResultDescriptor()
```
moozzyk: nit: remove `()`

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: They've always been in the past. Also, most things we do in ASP.NET Core are case insensitive in general (except the file system). What's the downside?

BrennanConroy: I would like it to be case insensitive, but if we care about correctness if would need to be case sensitive. e.g. Hub with 2 methods `Echo` and `echo`

davidfowl: I think case insensitivity makes the most sense.

moozzyk: I am weighing what @BrennanConroy said with "having two methods with different casing is not a good practice", not having to care about casing makes it easier to use, VB.NET is case insensitive (LOL). I think case insensitive is ultimately better - don't remember anyone complaining about this with SignalR 2.

davidfowl: I was going to bring up VB 😄.

BrennanConroy:
🚢 :it:

PR #167, opened 2017-01-29
Removed explicit capturing of "this" for lamdas

### LINKED ISSUE #166: Capturing "this" context
In the transport and connection typescript classes, you are capturing "this" explicitly but only using lamdas which will capture this for you. See [LongPollingTransport](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts#L160). Is there any reason for that? I've [committed something](https://github.com/herecydev/SignalR/commit/98bc22aed75584fff7f41e2ea2d9c6640bdec6c1) if you would like a PR?

herecydev:
Resolves #166

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
+++ b/src/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
@@ -24,9 +24,8 @@ export class HubConnection {
     constructor(url: string, queryString?: string) {
         this.connection = new Connection(url, queryString);
 
-        let thisHubConnection = this;
         this.connection.dataReceived = data => {
-            thisHubConnection.dataReceived(data);
+            this.dataReceived(data);
         };
 
         this.callbacks = new Map<string, (invocationDescriptor: InvocationResultDescriptor) => void>();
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -68,18 +67,16 @@ export class HubConnection {
     }
 
     invoke(methodName: string, ...args: any[]): Promise<any> {
-
-        let id = this.id;
```
davidfowl: not sure about this change, won't the promise potentially capture the wrong value for id at https://github.com/aspnet/SignalR/pull/167/files#diff-98c1bfb809ef38f16f965a508fe52526R79

herecydev: Yep, a little overzealous on the removal.

moozzyk:
Is it correct though? My understanding is that in case of event handlers `this` will be set to `event.currentTarget` and this is the reason why in all event handlers I uses the `that = this` pattern.
image

herecydev:
I'm pretty tired so forgive me if this is nonsense, but I've put together [this gist](https://gist.github.com/herecydev/294f1aedcf1bc3a58fc80652f8e68b79).

Either using this or explicitly capturing it results in the correct "this" being capture:

```js
var _this = this;
var thisHubConnection = this;
```

moozzyk:
Just noticed my mistake - in the picture above I marked `this` but this was in the Local scope and not in the closure. I need to confirm that indeed `this` works as expected in arrow functions used as event handlers and if this is the case I will merge this PR. 
(btw. I ran end-to-end TS client tests with your change and all tests passed so most likely this change is good).

moozzyk:
I think I was wrong and this should be captured just fine in event handlers.

PR #168, opened 2017-01-29
Features/transport enum

herecydev:
Allows the transports to be strongly typed, any other value will fail at compile-time:

image


moozzyk:
I was actually planning to move away from specifying transport as enums/strings/etc. In general users should be able to provide their own transports in which case the current API won't work. The idea is that instead of passing strings/enums the connection will take an object implementing the `ITransport` interface. The argument will be optional and if not passed WebSocketTransport would be used.

herecydev:
Sure that's fine, I guess they have to instantiate the httpClient also then? Is that intuitive?

moozzyk:
Not necessarily - if a transport needs an http client there is a default implementation that can be used if the user does not provide their own.

herecydev:
Yer I saw the implementation there, just wondering if it's immediately obvious. Guess it's best to let it bake for a while and see how it feels.

moozzyk:
I am closing this for now but we may want to resurrect it.

PR #172, opened 2017-01-30
Add TestClient to clean up tests

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
```
mikaelm12: Header

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        private static int _id;
+        private IInvocationAdapter _adapter;
+        private CancellationTokenSource _cts;
+
+        public Connection Connection;
+        public IChannelConnection<Message> Application { get; }
+        public Task Connected => Connection.Metadata.Get<TaskCompletionSource<bool>>("ConnectedTask").Task;
+
+        public TestClient(IServiceProvider serviceProvider, string format = "json")
```
mikaelm12: Don't we generally like to avoid default params in constructors?

moozzyk: Yes, especially in public API. Should be fine in tests though.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -0,0 +1,107 @@
...
+                    value = serializer.Deserialize<T>(new JsonTextReader(new StreamReader(new MemoryStream(message.Payload.Buffer.ToArray()))));
+                    return true;
+                }
+            }
+
+            value = default(T);
+            return false;
+        }
+
+        public void Close()
```
davidfowl: Why do we have Close and Dispose

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -0,0 +1,107 @@
...
+                Arguments = args,
+                Method = methodName
+            },
+            stream);
+
+            var buffer = ReadableBuffer.Create(stream.ToArray()).Preserve();
+            await Application.Output.WriteAsync(new Message(buffer, Format.Binary, endOfMessage: true));
+        }
+
+        public async Task<T> Read<T>()
```
davidfowl: Why do we still need this? Is this the lower level API?

BrennanConroy: e.g. Client1 broadcasts to all clients, how else do we let Client2 see the message?

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        public bool TryRead<T>(out T value)
+        {
+            Message message;
+            if (Application.Input.TryRead(out message))
+            {
+                using (message)
+                {
+                    var serializer = new JsonSerializer();
+                    value = serializer.Deserialize<T>(new JsonTextReader(new StreamReader(new MemoryStream(message.Payload.Buffer.ToArray()))));
```
davidfowl: Shouldn't this use the the adapter?

BrennanConroy: Ideally :/
It adds a lot of complexity now.

BrennanConroy:
:up: 📅

BrennanConroy:
@anurse Happy

moozzyk:
Looks good to me.

PR #173, opened 2017-01-30
Adding connection level tests

mikaelm12:
I had to remove the includes and make copies of the definitions of Format, IChannelConnection and ChannelConnection because there were issues with the conflicting namespaces.


src/Microsoft.AspNetCore.Sockets.Client/project.json
```diff
@@ -20,11 +20,7 @@
     ],
     "xmlDoc": true,
     "compile": {
-      "include": [
-        "../Microsoft.AspNetCore.Sockets/IChannelConnection.cs",
-        "../Microsoft.AspNetCore.Sockets/Internal/ChannelConnection.cs",
-        "../Microsoft.AspNetCore.Sockets/Format.cs"
-      ]
+      "include": [ ]
```
moozzyk: remove

src/Microsoft.AspNetCore.Sockets.Client/IChannelConnection.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: @davidfowl - is this moving to Channels?

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -5,9 +5,10 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Channels;
-using Microsoft.AspNetCore.Sockets.Internal;
+using Microsoft.AspNetCore.Sockets.Client.Internal;
 using Microsoft.Extensions.Logging;
 
+
```
moozzyk: nit: remove extraneous line

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -45,5 +49,42 @@ public EndToEndTests(ServerFixture serverFixture)
...
+        }
+
+
+        private static async Task<string> ReceiveMessage(Connection connection)
+        {
+            Message message;
+            while (await connection.Input.WaitToReadAsync())
+            {
+                if (connection.Input.TryRead(out message))
+                {
```
moozzyk: I think the message should be disposed.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -45,5 +49,42 @@ public EndToEndTests(ServerFixture serverFixture)
...
+                        ReadableBuffer.Create(Encoding.UTF8.GetBytes(message)).Preserve(),
+                        Format.Text));
+
+                    var received = await ReceiveMessage(connection).OrTimeout();
+                    Assert.Equal(message, received);
+                }
+            }
+        }
+
+
```
moozzyk: nit: extra line

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -45,5 +49,42 @@ public EndToEndTests(ServerFixture serverFixture)
                 await ws.CloseAsync(WebSocketCloseStatus.Empty, "", CancellationToken.None);
             }
         }
+
+        [Fact]
+        public async Task CheckConnection()
```
moozzyk: rename to something more descriptive (e.g. ConnectionCanSendReceiveMessages)

src/Microsoft.AspNetCore.Sockets.Client/Format.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public enum Format
```
anurse: Hmm. We have a SignalR.Common we use for sharing types like this. Maybe we need a Sockets.Common?

src/Microsoft.AspNetCore.Sockets.Client/IChannelConnection.cs
anurse: If we did Sockets.Common we could at least move this there until it is accepted into Channels. We should probably just submit a PR to corefxlab

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
anurse: Indeed it should, you'll need a `using` block inside the `if`. Technically it doesn't need to be disposed in this scenario, but we want to ensure we're using the right pattern for when we introduce pooling (disposing would return message buffers to the pool)

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
BrennanConroy: nit

mikaelm12:
I created Sockets.Common. It contains `IChannelConnection`, `ChannelConnection`,`Format`, and as per @davidfowl's suggestion `Message` .

davidfowl:
@mikaelm12 CI is failing

mikaelm12:
Updated

PR #179, opened 2017-02-02
Pawelka/api

moozzyk:
Preparation to changing to Connection to be event based
- Separating creating connection from starting connection (needed to be able to subscribe to DataReceived, Closed events before the connection is started)
- Enabling stopping connection without disposing connection
- Enabling stopping transports without disposing transports
- Fixing ownership issues 
   - LongPollingTransport will not dispose httpClient if it did not create one
   - Connection will not dispose transport if it did not create one
   - TODO: connection should dispose httpClient if it created one (easy to fix once LongPollingTransport is not the default transport)

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
@@ -7,11 +7,20 @@ using System.Threading.Tasks;
 using System.Threading.Tasks.Channels;
 using Microsoft.Extensions.Logging;
 using Microsoft.AspNetCore.Sockets.Internal;
+using System.Threading;
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
     public class Connection : IChannelConnection<Message>
     {
+        private class ConnectionState
+        {
+            public const int Disconnected = 0;
+            public const int Connecting = 1;
+            public const int Connected = 2;
+        }
+
+        private int _connectionState;
         private IChannelConnection<Message> _transportChannel;
         private ITransport _transport;
         private readonly ILogger _logger;
@@ -50,7 +59,12 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public async Task StartAsync(ITransport transport, HttpClient httpClient)
         {
-            // TODO: prevent from starting the connection if it has finished or is already running
+            if (Interlocked.CompareExchange(ref _connectionState, ConnectionState.Connecting, ConnectionState.Disconnected)
+                != ConnectionState.Disconnected)
+            {
+                throw new InvalidOperationException("Cannot start an already running connection.");
+            }
+
             if (httpClient == null)
             {
                 httpClient = new HttpClient();
@@ -70,6 +84,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
             catch (Exception ex)
             {
+                Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected);
                 _logger.LogError("Failed to start connection. Error getting connection id from '{0}': {1}", negotiateUrl, ex);
                 throw;
             }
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -7,99 +7,136 @@
 using System.Threading.Tasks.Channels;
 using Microsoft.Extensions.Logging;
 using Microsoft.AspNetCore.Sockets.Internal;
+using System.Threading;
```
mikaelm12: nit:Sort

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -20,27 +21,36 @@ public class LongPollingTransport : ITransport
...
+        public LongPollingTransport()
+            : this(null)
+        { }
 
-        public void Dispose()
+        public LongPollingTransport(ILoggerFactory loggerFactory)
+            : this(loggerFactory, null)
+        { }
+
+        public LongPollingTransport(ILoggerFactory loggerFactory, HttpClient httpClient)
```
mikaelm12: Should the HttpClient come first here? If someone wants to pass in their own httpclient they also have to provide a loggerFactory which seems a little odd

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -164,5 +173,162 @@ public class ConnectionTests
                 Assert.Equal(connection.Input.Completion, await whenAnyTask);
             }
         }
+
+        [Fact]
+        public async Task CannotStartNonDisconnectedConnection()
```
mikaelm12: Maybe change to CannotStartRunningConnection

moozzyk: It can also be connecting (which is actually much more likely)

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: I switched the order because of Intellisense. It looked really weird when switching between these overloads. An alternative is to add an overload that only takes HttpClient but I thought it was too much. Note that loggerFactory can be null so you don't technically have to provide it.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -20,27 +21,36 @@ public class LongPollingTransport : ITransport
...
-            _transportCts.Cancel();
+            _httpClient = httpClient ?? new HttpClient();
+            _ownsHttpClient = httpClient == null;
+            _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<LongPollingTransport>();
+            Running = Task.FromResult<object>(null);
         }
 
         public Task StartAsync(Uri url, IChannelConnection<Message> application)
         {
+            Debug.Assert(_transportCts == null || _transportCts.Token.IsCancellationRequested, "transport is still running");
```
davidfowl: We're assuming bad usage of the transport will be guarded by the connection?

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -7,99 +7,136 @@
...
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
     public class Connection : IChannelConnection<Message>
     {
+        private class ConnectionState
```
davidfowl: Private classes go last

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -7,99 +7,136 @@
...
-            if (transport == null)
+            if (httpClient == null)
             {
-                throw new ArgumentNullException(nameof(transport));
+                // TODO: httpClient needs to be disposed properly (after long polling is not the default transport)
+                httpClient = new HttpClient();
             }
 
-            if (httpClient == null)
+            var connectionId = await Negotiate(httpClient);
```
davidfowl: We're going to have to go through and add `ConfigureAwait(false)` on a bunch of these calls

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: I was not sure how far we want to go there. I will play with the usage and see.

PR #180, opened 2017-02-06
update MSBuild settings to make samples runnable

anurse:
/cc @davidfowl @natemcmaster


samples/SocialWeather/SocialWeather.csproj
```diff
@@ -1,12 +1,20 @@
...
   <PropertyGroup>
     <TargetFramework>netcoreapp1.1</TargetFramework>
     <PreserveCompilationContext>true</PreserveCompilationContext>
     <OutputType>Exe</OutputType>
+    <RuntimeFrameworkVersion>1.2.0-*</RuntimeFrameworkVersion>
+    <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <ItemGroup>
+    <Content Include="wwwroot\index.html" />
```
anurse: Uhh... weird. Cleaning this up. I renamed from VS, so there might be a bug here.

samples/ChatSample/ChatSample.csproj
```diff
@@ -1,11 +1,15 @@
...
+
+  <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
     <TargetFramework>netcoreapp1.1</TargetFramework>
     <PreserveCompilationContext>true</PreserveCompilationContext>
     <OutputType>Exe</OutputType>
     <UserSecretsId>aspnet-ChatSample-f11cf018-e0a8-49fa-b749-4c0eb5c9150b</UserSecretsId>
     <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8</PackageTargetFallback>
+    <RuntimeFrameworkVersion>1.2.0-*</RuntimeFrameworkVersion>
```
davidfowl: This is so that things work once the feature is fixed?

samples/SocialWeather/Startup.cs
```diff
@@ -35,7 +35,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
             }
 
             app.UseSockets(o => { o.MapEndpoint<SocialWeatherEndPoint>("/weather"); });
-            app.UseStaticFiles();
+            app.UseFileServer();
```
anurse: This just makes the sample a little more runnable. Right now it 404s when you Ctrl-F5 it. I could have changed the launch URL but this seemed smoother.

samples/ChatSample/ChatSample.csproj
```diff
@@ -1,11 +1,15 @@
...
+    <IsPackable>false</IsPackable>
```
davidfowl: @natemcmaster why wouldn't they just infer this for `<OutputType>Exe</OutputType>`?

samples/SocialWeather/Startup.cs
```diff
@@ -35,7 +35,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
             }
 
             app.UseSockets(o => { o.MapEndpoint<SocialWeatherEndPoint>("/weather"); });
-            app.UseStaticFiles();
```
moozzyk: why is this needed?

samples/SocialWeather/SocialWeather.csproj
anurse: Indeed. And @natemcmaster we still need to set it, or something. Either way, it seems appropriate to have it.

moozzyk:
Microsoft.AspNetCore.SignalR.Test.Server is broken too

samples/SocialWeather/Startup.cs
anurse: See below comment. It just makes Ctrl-F5 smoother. `UseFileServer` activates default document, so `index.html` works.

samples/ChatSample/ChatSample.csproj
moozzyk: This is even more confusing - if you target both netcoreapp1.1 and net46 what does actually Output type means?

anurse:
Shouldn't the build have caught that? We use the server in our tests, right?

samples/SocialWeather/Startup.cs
moozzyk: I thought 404 was because there was no index.html. I saw you changed that but I don't know why it requires FileServer...

samples/SocialWeather/Startup.cs
moozzyk: I did not know that so 👍

moozzyk:
Test server currently is used only for end-to-end ts client tests which are not run by default since they need to be run from browser.

samples/ChatSample/ChatSample.csproj
natemcmaster: @davidfowl IsPackable is true by default on all project types. Not sure why NuGet decided to make it that way. 

Setting this to false means calling "dotnet pack" on the .sln file won't try to pack this project.

natemcmaster: Also, there are valid scenarios for packing OutputType=Exe. e.g. dotnet-cli tools packages.

samples/ChatSample/ChatSample.csproj
natemcmaster: To be clear until https://github.com/dotnet/sdk/pull/820 is in, this will break apps if you didn't also reference Internal.AspNetCore.Sdk.

samples/ChatSample/ChatSample.csproj
moozzyk: Sure but in this case (which I think is more common) it probably means nothing but introduces confusion.

natemcmaster: > introduces confusion

Ok. Add a comment.

test/Microsoft.Extensions.WebSockets.Internal.Tests/WebSocketConnectionTests.PingPong.cs
```diff
@@ -29,7 +29,7 @@ public class PingPongBehavior
 
                     // Act
                     // Wait for pings to be sent
-                    await Task.Delay(200);
+                    await Task.Delay(500);
```
mikaelm12: 🙁

PR #183, opened 2017-02-07
Hiding Channels

moozzyk:
Supersedes #179


src/Microsoft.AspNetCore.Sockets.Client/ReceiveData.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
mikaelm12: nit: Remove line

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -7,18 +7,19 @@
 using System.Threading.Tasks.Channels;
 using Microsoft.Extensions.Logging;
 using Microsoft.AspNetCore.Sockets.Internal;
+using System.Threading;
```
mikaelm12: nit: Sort

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -142,82 +131,74 @@ public void Dispose()
...
+                    InvocationMessage message 
+                        = await _adapter.ReadMessageAsync(new MemoryStream(receiveData.Data), _binder, cancellationToken);
 
-                        var invocationDescriptor = message as InvocationDescriptor;
-                        if (invocationDescriptor != null)
-                        {
-                            DispatchInvocation(invocationDescriptor, cancellationToken);
-                        }
-                        else
+                    if (message is InvocationDescriptor invocationDescriptor)
```
mikaelm12: I might be wrong on this but I believe that the performance of using the `as` keyword doing a null check afterward is slightly better.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -28,58 +29,110 @@ private Connection(Uri url, ITransport transport, IChannelConnection<Message> tr
             _transportChannel = transportChannel;
         }
 
-        public ReadableChannel<Message> Input => _transportChannel.Input;
-        public WritableChannel<Message> Output => _transportChannel.Output;
-
-        public void Dispose()
+        private ReadableChannel<Message> Input => _transportChannel.Input;
+        private WritableChannel<Message> Output => _transportChannel.Output;
+
```
mikaelm12: nit: White space

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: as + null is to avoid double cast. This is C#7 and there is no double cast here.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: I really need to update the [Trailing Whitespace Flagger](https://marketplace.visualstudio.com/items?itemName=PawelKadluczka-MSFT.TrailingSpaceFlagger) to VS2017

mikaelm12: Yes! 👍

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
davidfowl: C# 7 ftfw

davidfowl: Switch case might look better here

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -28,58 +29,110 @@ private Connection(Uri url, ITransport transport, IChannelConnection<Message> tr
...
-                throw new ArgumentNullException(nameof(loggerFactory));
+                if (Output.TryWrite(message))
+                {
+                    return;
+                }
             }
 
-            var logger = loggerFactory.CreateLogger<Connection>();
-            var negotiateUrl = Utils.AppendPath(url, "negotiate");
+            throw new InvalidOperationException("Cannot send messages when the connection is stopped.");
```
davidfowl: Lets not throw here. This is one of those things that ends up being impossible to avoid.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -28,58 +29,110 @@ private Connection(Uri url, ITransport transport, IChannelConnection<Message> tr
...
+        }
 
-            string connectionId;
-            try
+        public async Task StopAsync()
+        {
+            Output.TryComplete();
+            await _transport.StopAsync();
```
davidfowl: What happens if Dispose is called then Stop?

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -120,48 +140,123 @@ public class ConnectionTests
...
+            var allowPollTcs = new TaskCompletionSource<object>();
+            var mockHttpHandler = new Mock<HttpMessageHandler>();
+            mockHttpHandler.Protected()
+                .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.IsAny<HttpRequestMessage>(), ItExpr.IsAny<CancellationToken>())
+                .Returns<HttpRequestMessage, CancellationToken>(async (request, cancellationToken) =>
                 {
-                    connection.Input.TryRead(out message);
-                    message.Dispose();
-                }
+                    await Task.Yield();
```
davidfowl: This is no good with xunit if you're trying to thread hop

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: Should be a no op now. Once we have state we could throw ObjectDisposedException if you call a public method after Dispose() (or Stop)

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
moozzyk: What is the alternative here? This pattern has been used in the tests for a while and did not cause issues.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: OK. Should we return a flag indicating that the channel was closed (i.e. sending data was not even attempted)?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: Let me try that.

PR #185, opened 2017-02-08
WebSockets Transport

mikaelm12:
Still a work in progress so not polished, but wanted to get some feedback.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/Microsoft.AspNetCore.Sockets.Client.csproj
+++ b/src/Microsoft.AspNetCore.Sockets.Client/Microsoft.AspNetCore.Sockets.Client.csproj
@@ -13,8 +13,6 @@
     <PackageReference Include="System.Threading.Tasks.Channels" Version="0.1.0-*" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.2.0-*" />
     <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
-  </ItemGroup>
-  <ItemGroup>
     <PackageReference Include="System.Net.WebSockets.Client" Version="4.4.0-*" />
   </ItemGroup>
 </Project>
--- a/src/Microsoft.AspNetCore.Sockets.Client/Microsoft.AspNetCore.Sockets.Client.csproj
+++ b/src/Microsoft.AspNetCore.Sockets.Client/Microsoft.AspNetCore.Sockets.Client.csproj
@@ -1,7 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <Import Project="..\..\build\common.props" />
-
   <PropertyGroup>
     <Description>Client for ASP.NET Core SignalR</Description>
     <TargetFramework>netstandard1.3</TargetFramework>
@@ -9,13 +7,14 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;signalr</PackageTags>
   </PropertyGroup>
-
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Sockets.Common\Microsoft.AspNetCore.Sockets.Common.csproj" />
     <PackageReference Include="System.IO.Pipelines" Version="0.1.0-*" />
     <PackageReference Include="System.Threading.Tasks.Channels" Version="0.1.0-*" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.2.0-*" />
-    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
   </ItemGroup>
-
+  <ItemGroup>
+    <PackageReference Include="System.Net.WebSockets.Client" Version="4.4.0-*" />
+  </ItemGroup>
 </Project>
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -15,21 +15,25 @@ namespace Microsoft.AspNetCore.Sockets.Client
     {
         private ClientWebSocket _webSocket = new ClientWebSocket();
         private IChannelConnection<Message> _application;
-        private CancellationToken token = new CancellationToken();
+        private CancellationToken _cancellationToken = new CancellationToken();
 
         public Task Running { get; private set; }
-
-        public WebSocketsTransport()
-        {
-        }
                                                 
         public async Task StartAsync(Uri url, IChannelConnection<Message> application)  
         {
+            if (url == null)
+            {
+                throw new ArgumentNullException(nameof(url));
+            }
+            if (application == null)
+            {
+                throw new ArgumentNullException(nameof(application));
+            }
             _application = application;
             var webSocketUrl = url.ToString();
             await Connect(webSocketUrl);
-            var sendTask = SendMessages(url, token);
-            var receiveTask = ReceiveMessages(url, token);
+            var sendTask = SendMessages(url, _cancellationToken);
+            var receiveTask = ReceiveMessages(url, _cancellationToken);
 
             Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t => {
                 _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
@@ -39,21 +43,21 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
         {
+            var totalBytes = 0;
             var incomingMessage = new List<byte[]>();
             while (!cancellationToken.IsCancellationRequested)
             {
-                var buffer = new ArraySegment<byte>(new byte[1024]);
-                var receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
-                var totalBytes = 0;
-                var completeMessage = receiveResult.EndOfMessage;
-                while (!completeMessage)
+                bool completedMessage;
+                WebSocketReceiveResult receiveResult;
+                ArraySegment<byte> buffer;
+                do
                 {
-                    completeMessage = receiveResult.EndOfMessage;
+                    buffer = new ArraySegment<byte>(new byte[1024]);
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                    completedMessage = receiveResult.EndOfMessage;
                     incomingMessage.Add(buffer.Array);
                     totalBytes += buffer.Count;
-                    buffer = new ArraySegment<byte>(new byte[2]);
-                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
-                }
+                } while (!completedMessage);
 
                 Message message;
                 if (incomingMessage.Count > 1)
@@ -108,12 +112,12 @@ namespace Microsoft.AspNetCore.Sockets.Client
         public async Task Connect(string url, string queryString)
         {
             var wsUrl = url.Replace("http", "ws").Replace("https", "ws");
-            await _webSocket.ConnectAsync(new Uri(wsUrl + queryString), token);
+            await _webSocket.ConnectAsync(new Uri(wsUrl + queryString), _cancellationToken);
         }
 
         public void Dispose()
         {
-            _webSocket.CloseAsync(WebSocketCloseStatus.Empty, null, token);
+            _webSocket.CloseAsync(WebSocketCloseStatus.Empty, null, _cancellationToken);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -6,7 +6,6 @@ using System.IO.Pipelines;
 using System.Net.WebSockets;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Logging;
 using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Sockets.Client
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -141,5 +141,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
         {
             _webSocket.Dispose();
         }
+
+        public async Task StopAsync()
+        {
+            await _webSocket.CloseAsync(WebSocketCloseStatus.Empty, null, _cancellationToken);
+        }
     }
 }
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -7,6 +7,7 @@ using System.Net.WebSockets;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
@@ -15,7 +16,15 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private ClientWebSocket _webSocket = new ClientWebSocket();
         private IChannelConnection<Message> _application;
         private CancellationToken _cancellationToken = new CancellationToken();
+        private readonly ILogger _logger;
 
+        public WebSocketsTransport()
+        {
+        }
+        public WebSocketsTransport(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<WebSocketsTransport>();
+        }
         public Task Running { get; private set; }
                                                 
         public async Task StartAsync(Uri url, IChannelConnection<Message> application)  
@@ -29,8 +38,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 throw new ArgumentNullException(nameof(application));
             }
             _application = application;
-            var webSocketUrl = url.ToString();
-            await Connect(webSocketUrl);
+            await Connect(url);
             var sendTask = SendMessages(url, _cancellationToken);
             var receiveTask = ReceiveMessages(url, _cancellationToken);
 
@@ -40,46 +48,45 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }).Unwrap();
         }
 
-        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
+        private async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
         {
-            var totalBytes = 0;
-            var incomingMessage = new List<byte[]>();
             while (!cancellationToken.IsCancellationRequested)
             {
-                bool completedMessage;
+                const int bufferSize = 1024;
+                var totalBytes = 0;
+                var incomingMessage = new List<byte[]>();
                 WebSocketReceiveResult receiveResult;
-                ArraySegment<byte> buffer;
                 do
                 {
-                    buffer = new ArraySegment<byte>(new byte[1024]);
+                    var buffer = new ArraySegment<byte>(new byte[bufferSize]);
                     receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
-                    completedMessage = receiveResult.EndOfMessage;
                     incomingMessage.Add(buffer.Array);
-                    totalBytes += buffer.Count;
-                } while (!completedMessage);
+                    totalBytes += receiveResult.Count;
+                } while (!receiveResult.EndOfMessage);
 
                 Message message;
                 if (incomingMessage.Count > 1)
                 {
                     var totalBuffer = new byte[totalBytes];
                     var offset = 0;
-                    for (int i = 0 ; i < incomingMessage.Count; i++)
+                    for (var i = 0 ; i < incomingMessage.Count; i++)
                     {
-                        System.Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
+                        Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
                         offset += incomingMessage[i].Length;
                     }
-
-                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), Format.Text, receiveResult.EndOfMessage);
+                    
+                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
                 }
                 else
                 {
-                    message = new Message(ReadableBuffer.Create(buffer.Array).Preserve(), Format.Text, receiveResult.EndOfMessage);
+                    message = new Message(ReadableBuffer.Create(incomingMessage[0]).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
                 }
 
                 while (await _application.Output.WaitToWriteAsync(cancellationToken))
                 {
                     if (_application.Output.TryWrite(message))
                     {
+                        incomingMessage.Clear();
                         break;
                     }
                 }
@@ -95,23 +102,33 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 {
                     using (message)
                     {
-                        await _webSocket.SendAsync(new ArraySegment<byte>(message.Payload.Buffer.ToArray()),
+                        try
+                        {
+                            await _webSocket.SendAsync(new ArraySegment<byte>(message.Payload.Buffer.ToArray()),
                             message.MessageFormat == Format.Text ? WebSocketMessageType.Text : WebSocketMessageType.Binary, true,
                             cancellationToken);
+                        } catch(OperationCanceledException ex)
+                        {
+                            _logger?.LogError("The WebSocket operation was cancelled");
+                        }
                     }
                 }
             }
         }
 
-        public async Task Connect(string url)
+        private async Task Connect(Uri url)
         {
-            await Connect(url, "");
-        }
+            var uriBuilder = new UriBuilder(url);
+            if (url.Scheme == "http")
+            {
+                uriBuilder.Scheme = "ws";
+            }
+            else if (url.Scheme == "https")
+            {
+                uriBuilder.Scheme = "wss";
+            }
 
-        public async Task Connect(string url, string queryString)
-        {
-            var wsUrl = url.Replace("http", "ws").Replace("https", "ws");
-            await _webSocket.ConnectAsync(new Uri(wsUrl + queryString), _cancellationToken);
+            await _webSocket.ConnectAsync(uriBuilder.Uri, _cancellationToken);
         }
 
         public void Dispose()
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -67,15 +67,15 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 Message message;
                 if (incomingMessage.Count > 1)
                 {
-                    var totalBuffer = new byte[totalBytes];
+                    var messageBuffer = new byte[totalBytes];
                     var offset = 0;
                     for (var i = 0 ; i < incomingMessage.Count; i++)
                     {
-                        Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
+                        Buffer.BlockCopy(incomingMessage[i], 0, messageBuffer, offset, incomingMessage[i].Length);
                         offset += incomingMessage[i].Length;
                     }
                     
-                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
+                    message = new Message(ReadableBuffer.Create(messageBuffer).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
                 }
                 else
                 {
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -20,7 +20,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public WebSocketsTransport()
         {
+            _logger = NullLoggerFactory.Instance.CreateLogger("WebSocketsTransport");
         }
+
         public WebSocketsTransport(ILoggerFactory loggerFactory)
         {
             _logger = loggerFactory.CreateLogger<WebSocketsTransport>();
@@ -33,10 +35,12 @@ namespace Microsoft.AspNetCore.Sockets.Client
             {
                 throw new ArgumentNullException(nameof(url));
             }
+
             if (application == null)
             {
                 throw new ArgumentNullException(nameof(application));
             }
+
             _application = application;
             await Connect(url);
             var sendTask = SendMessages(url, _cancellationToken);
@@ -54,13 +58,13 @@ namespace Microsoft.AspNetCore.Sockets.Client
             {
                 const int bufferSize = 1024;
                 var totalBytes = 0;
-                var incomingMessage = new List<byte[]>();
+                var incomingMessage = new List<ArraySegment<byte>>();
                 WebSocketReceiveResult receiveResult;
                 do
                 {
                     var buffer = new ArraySegment<byte>(new byte[bufferSize]);
                     receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
-                    incomingMessage.Add(buffer.Array);
+                    incomingMessage.Add(buffer);
                     totalBytes += receiveResult.Count;
                 } while (!receiveResult.EndOfMessage);
 
@@ -71,15 +75,15 @@ namespace Microsoft.AspNetCore.Sockets.Client
                     var offset = 0;
                     for (var i = 0 ; i < incomingMessage.Count; i++)
                     {
-                        Buffer.BlockCopy(incomingMessage[i], 0, messageBuffer, offset, incomingMessage[i].Length);
-                        offset += incomingMessage[i].Length;
+                        Buffer.BlockCopy(incomingMessage[i].Array, 0, messageBuffer, offset, incomingMessage[i].Count);
+                        offset += incomingMessage[i].Count;
                     }
                     
                     message = new Message(ReadableBuffer.Create(messageBuffer).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
                 }
                 else
                 {
-                    message = new Message(ReadableBuffer.Create(incomingMessage[0]).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
+                    message = new Message(ReadableBuffer.Create(incomingMessage[0].Array).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
                 }
 
                 while (await _application.Output.WaitToWriteAsync(cancellationToken))
@@ -98,7 +102,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             while (await _application.Input.WaitToReadAsync(cancellationToken))
             {
                 Message message;
-                while (!cancellationToken.IsCancellationRequested && _application.Input.TryRead(out message))
+                while (_application.Input.TryRead(out message))
                 {
                     using (message)
                     {
@@ -109,7 +113,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
                             cancellationToken);
                         } catch(OperationCanceledException ex)
                         {
-                            _logger?.LogError("The WebSocket operation was cancelled");
+                            _logger?.LogError(ex.Message);
+                            await _webSocket.CloseAsync(WebSocketCloseStatus.Empty, null, _cancellationToken);
+                            break;
                         }
                     }
                 }
@@ -133,7 +139,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public void Dispose()
         {
-            _webSocket.CloseAsync(WebSocketCloseStatus.Empty, null, _cancellationToken);
+            _webSocket.Dispose();
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -1,0 +1,97 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.IO.Pipelines;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public class WebSocketsTransport : ITransport
+    {
+        private ClientWebSocket _webSocket = new ClientWebSocket();
+        private IChannelConnection<Message> _application;
+        private ILogger _logger;
+        private CancellationToken token = new CancellationToken();
+
+        public Task Running { get; private set; }
+
+        public WebSocketsTransport()
+        {
+        }
+
+        public async Task StartAsync(Uri url, IChannelConnection<Message> application)
+        {
+            _application = application;
+            var webSocketUrl = url.ToString();
+            await Connect(webSocketUrl);
+            var sendTask = SendMessages(url, token);
+            var receiveTask = ReceiveMessages(url, token);
+
+            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t => {
+                _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
+                return t;
+            }).Unwrap();
+        }
+
+        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
+        {
+
+            while (!cancellationToken.IsCancellationRequested )
+            {
+                var buffer = new ArraySegment<byte>(new byte[1]);
+                var receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                var ms = new MemoryStream();
+                var message = new Message(ReadableBuffer.Create(buffer.Array).Preserve(), Format.Text, receiveResult.EndOfMessage);
+
+                while (await _application.Output.WaitToWriteAsync(cancellationToken))
+                {
+                    if (_application.Output.TryWrite(message))
+                    {
+                        break;
+                    }
+                }
+                    
+            }
+        }
+
+        private async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
+        {
+            while (await _application.Input.WaitToReadAsync(cancellationToken))
+            {
+                Message message;
+                while (!cancellationToken.IsCancellationRequested && _application.Input.TryRead(out message))
+                {
+                    using (message)
+                    {
+                        await _webSocket.SendAsync(new ArraySegment<byte>(message.Payload.Buffer.ToArray()),
+                            message.MessageFormat == Format.Text ? WebSocketMessageType.Text : WebSocketMessageType.Binary, true,
+                            cancellationToken);
+                    }
+                }
+            }
+        }
+
+        public async Task Connect(string url)
+        {
+            await Connect(url, "");
+        }
+
+        public async Task Connect(string url, string queryString)
+        {
+            var wsUrl = url.Replace("http", "ws").Replace("https", "ws");
+            await _webSocket.ConnectAsync(new Uri(wsUrl + queryString), token);
+        }
+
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+    public class WebSocketsTransport : ITransport
+    {
+        private ClientWebSocket _webSocket = new ClientWebSocket();
+        private IChannelConnection<Message> _application;
+        private CancellationToken token = new CancellationToken();
+
+        public Task Running { get; private set; }
+
+        public WebSocketsTransport()
```
BrennanConroy: remove

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+        }
+                                                
+        public async Task StartAsync(Uri url, IChannelConnection<Message> application)
```
BrennanConroy: null check

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public class WebSocketsTransport : ITransport
+    {
+        private ClientWebSocket _webSocket = new ClientWebSocket();
+        private IChannelConnection<Message> _application;
+        private CancellationToken token = new CancellationToken();
```
BrennanConroy: Fix variable names

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                var buffer = new ArraySegment<byte>(new byte[1024]);
+                var receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                var totalBytes = 0;
+                var completeMessage = receiveResult.EndOfMessage;
+                while (!completeMessage)
+                {
+                    completeMessage = receiveResult.EndOfMessage;
+                    incomingMessage.Add(buffer.Array);
+                    totalBytes += buffer.Count;
```
BrennanConroy: unused

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
+        {
+            var incomingMessage = new List<byte[]>();
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                var buffer = new ArraySegment<byte>(new byte[1024]);
+                var receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                var totalBytes = 0;
+                var completeMessage = receiveResult.EndOfMessage;
+                while (!completeMessage)
```
BrennanConroy: This should be convertible to `do while`

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    completeMessage = receiveResult.EndOfMessage;
+                    incomingMessage.Add(buffer.Array);
+                    totalBytes += buffer.Count;
+                    buffer = new ArraySegment<byte>(new byte[2]);
```
BrennanConroy: ?

mikaelm12: Yeah, lets ignore that

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,119 @@
...
+                return t;
+            }).Unwrap();
+        }
+
+        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
+        {
+            var incomingMessage = new List<byte[]>();
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                var buffer = new ArraySegment<byte>(new byte[1024]);
```
BrennanConroy: Didn't we want to make this number an internal variable for testing purposes?

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+            var sendTask = SendMessages(url, _cancellationToken);
+            var receiveTask = ReceiveMessages(url, _cancellationToken);
+
+            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t => {
+                _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
+                return t;
+            }).Unwrap();
+        }
+
+        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
```
BrennanConroy: private?

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -91,5 +91,26 @@ public EndToEndTests(ServerFixture serverFixture)
...
+            var transport = new WebSocketsTransport();
+            using (var connection = await ClientConnection.ConnectAsync(new Uri(baseUrl + "/echo/ws"), transport, loggerFactory))
+            {
+                await connection.Output.WriteAsync(new Message(
+                    ReadableBuffer.Create(Encoding.UTF8.GetBytes(message)).Preserve(),
+                    Format.Text));
+
+                var received = await ReceiveMessage(connection);
+                
+                Assert.Equal(message, received.Substring(0,message.Length));
```
BrennanConroy: space

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+using System;
+using System.IO.Pipelines;
+using System.Net.WebSockets;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public class WebSocketsTransport : ITransport
```
BrennanConroy: Make only the interface methods public, the rest private

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                WebSocketReceiveResult receiveResult;
+                ArraySegment<byte> buffer;
+                do
+                {
+                    buffer = new ArraySegment<byte>(new byte[1024]);
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                    completedMessage = receiveResult.EndOfMessage;
+                    incomingMessage.Add(buffer.Array);
+                    totalBytes += buffer.Count;
+                } while (!completedMessage);
```
anurse: Just use `receiveResult.EndOfMessage` here.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+            if (url == null)
+            {
+                throw new ArgumentNullException(nameof(url));
+            }
+            if (application == null)
+            {
+                throw new ArgumentNullException(nameof(application));
+            }
+            _application = application;
+            var webSocketUrl = url.ToString();
```
moozzyk: Consider not converting to String. WebSocketClient takes Uri. (I also am not convinced that `.ToString()` is the right way to get a string version of the Uri)

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+            var totalBytes = 0;
+            var incomingMessage = new List<byte[]>();
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                bool completedMessage;
+                WebSocketReceiveResult receiveResult;
+                ArraySegment<byte> buffer;
+                do
+                {
+                    buffer = new ArraySegment<byte>(new byte[1024]);
```
anurse: Fowler's not going to be a fan of this allocation ;). We may need to introduce a buffer pool or something here.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                throw new ArgumentNullException(nameof(url));
+            }
+            if (application == null)
+            {
+                throw new ArgumentNullException(nameof(application));
+            }
+            _application = application;
+            var webSocketUrl = url.ToString();
+            await Connect(webSocketUrl);
```
moozzyk: I would roll in the connect method in

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
+        {
+            var totalBytes = 0;
+            var incomingMessage = new List<byte[]>();
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                bool completedMessage;
+                WebSocketReceiveResult receiveResult;
+                ArraySegment<byte> buffer;
```
anurse: This should be moved into the `do`. See below.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                    {
+                        System.Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
+                        offset += incomingMessage[i].Length;
+                    }
+
+                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), Format.Text, receiveResult.EndOfMessage);
+                }
+                else
+                {
+                    message = new Message(ReadableBuffer.Create(buffer.Array).Preserve(), Format.Text, receiveResult.EndOfMessage);
```
anurse: I'd rather keep `buffer` scoped to the loop above. You can just use `incomingMessage[0]` if you really want to optimize the "one-buffer message" scenario.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+            }).Unwrap();
+        }
+
+        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
+        {
+            var totalBytes = 0;
+            var incomingMessage = new List<byte[]>();
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                bool completedMessage;
```
anurse: Not needed, see below.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                WebSocketReceiveResult receiveResult;
+                ArraySegment<byte> buffer;
+                do
+                {
+                    buffer = new ArraySegment<byte>(new byte[1024]);
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
```
moozzyk: I think we need to handle the OperationCancelledException from here.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                } while (!completedMessage);
+
+                Message message;
+                if (incomingMessage.Count > 1)
+                {
+                    var totalBuffer = new byte[totalBytes];
+                    var offset = 0;
+                    for (int i = 0 ; i < incomingMessage.Count; i++)
+                    {
+                        System.Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
```
anurse: Remove `System.`

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: consider making buffer size a const

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                bool completedMessage;
+                WebSocketReceiveResult receiveResult;
+                ArraySegment<byte> buffer;
+                do
+                {
+                    buffer = new ArraySegment<byte>(new byte[1024]);
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                    completedMessage = receiveResult.EndOfMessage;
+                    incomingMessage.Add(buffer.Array);
+                    totalBytes += buffer.Count;
```
anurse: This is the `Count` from the `ArraySegment` you passed in to `ReceiveAsync`, which will always be `1024` since you created it from the array above (`ReceiveAsync` doesn't modify it, since it's passed in by-value). The actual number of bytes read into the buffer is available in `receiveResult`.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: I don't think you need this flag. You should be able to check receiveResult.EndOfMessage

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                bool completedMessage;
+                WebSocketReceiveResult receiveResult;
+                ArraySegment<byte> buffer;
+                do
+                {
+                    buffer = new ArraySegment<byte>(new byte[1024]);
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                    completedMessage = receiveResult.EndOfMessage;
+                    incomingMessage.Add(buffer.Array);
```
anurse: You can't add the whole array to the message because the message might have been smaller than `1024` bytes (even if the message is "incomplete", since you may be getting a partial frame), you need to build a **new** `ArraySegment` from the array, using the `Count` from `receiveResult`. Then you need to store that in the `incomingMessage` list.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: I think this can and should be moved into the loop. You don't use it outside.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
anurse: As per comments below, it might be better to create the `ArraySegment` inline in the `ReceiveAsync` call below, since it's only usable in that call.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t => {
+                _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
+                return t;
+            }).Unwrap();
+        }
+
+        public async Task ReceiveMessages(Uri pollUrl, CancellationToken cancellationToken)
+        {
+            var totalBytes = 0;
+            var incomingMessage = new List<byte[]>();
```
moozzyk: This should be inside the while loop. You don't seem to clear it so you seem to be sending previous messages.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                    incomingMessage.Add(buffer.Array);
+                    totalBytes += buffer.Count;
+                } while (!completedMessage);
+
+                Message message;
+                if (incomingMessage.Count > 1)
+                {
+                    var totalBuffer = new byte[totalBytes];
+                    var offset = 0;
+                    for (int i = 0 ; i < incomingMessage.Count; i++)
```
moozzyk: var

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: `using System` instead of `System.`

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                    completedMessage = receiveResult.EndOfMessage;
+                    incomingMessage.Add(buffer.Array);
+                    totalBytes += buffer.Count;
+                } while (!completedMessage);
+
+                Message message;
+                if (incomingMessage.Count > 1)
+                {
+                    var totalBuffer = new byte[totalBytes];
```
moozzyk: consider calling it messageBuffer - I though it was size/int

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                    var offset = 0;
+                    for (int i = 0 ; i < incomingMessage.Count; i++)
+                    {
+                        System.Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
+                        offset += incomingMessage[i].Length;
+                    }
+
+                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), Format.Text, receiveResult.EndOfMessage);
```
moozzyk: Shouldn't format be taken from the receive result?

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+        private async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
+        {
+            while (await _application.Input.WaitToReadAsync(cancellationToken))
+            {
+                Message message;
+                while (!cancellationToken.IsCancellationRequested && _application.Input.TryRead(out message))
+                {
+                    using (message)
+                    {
+                        await _webSocket.SendAsync(new ArraySegment<byte>(message.Payload.Buffer.ToArray()),
```
moozzyk: handle OperationCancelledException

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+                            message.MessageFormat == Format.Text ? WebSocketMessageType.Text : WebSocketMessageType.Binary, true,
+                            cancellationToken);
+                    }
+                }
+            }
+        }
+
+        public async Task Connect(string url)
```
moozzyk: Inline

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            await Connect(url, "");
+        }
+
+        public async Task Connect(string url, string queryString)
```
moozzyk: Is queryString even a thing?

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -91,5 +91,26 @@ public EndToEndTests(ServerFixture serverFixture)
 
             return null;
         }
+
+        [Fact]
+        public async Task ConnectionCanSendAndReceiveMessagesWebSocketsTransport()
```
moozzyk: We need transport specific tests. Look at what we have in LongpollingTransportTests.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
mikaelm12: Good catch

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: Pooling however can be done separately. I would prefer to have this transport in and than add pooling. Than block on adding pooling. It is important though.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
moozzyk: Actually all EndToEnd tests should use each transport - make it a Theory

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            var wsUrl = url.Replace("http", "ws").Replace("https", "ws");
+            await _webSocket.ConnectAsync(new Uri(wsUrl + queryString), _cancellationToken);
+        }
+
+        public void Dispose()
```
moozzyk: You will need to add StopAsync() - I am going to merge my PR soon. https://github.com/aspnet/SignalR/pull/183/files#diff-2e320ab431f1fe2e1ddea4ab7b1aebfbR12

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+    public class WebSocketsTransport : ITransport
+    {
+        private ClientWebSocket _webSocket = new ClientWebSocket();
+        private IChannelConnection<Message> _application;
+        private CancellationToken _cancellationToken = new CancellationToken();
+        private readonly ILogger _logger;
+
+        public WebSocketsTransport()
+        {
+        }
```
anurse: nit: add a blank line separating constructors.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -15,7 +16,15 @@ public class WebSocketsTransport : ITransport
...
         private IChannelConnection<Message> _application;
         private CancellationToken _cancellationToken = new CancellationToken();
+        private readonly ILogger _logger;
 
+        public WebSocketsTransport()
+        {
+        }
+        public WebSocketsTransport(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<WebSocketsTransport>();
```
moozzyk: Consider using NullLoggerFactory if nothing is passed. This way we will not run into NREs because someone forgot to check for null.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -15,7 +16,15 @@ public class WebSocketsTransport : ITransport
         private ClientWebSocket _webSocket = new ClientWebSocket();
         private IChannelConnection<Message> _application;
         private CancellationToken _cancellationToken = new CancellationToken();
+        private readonly ILogger _logger;
 
+        public WebSocketsTransport()
```
moozzyk: Call the other ctor - if we add more initialization code it we will not have to do this in two places (also see my comment about loggerfactory below)

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -40,46 +48,45 @@ public class WebSocketsTransport : ITransport
...
-                    for (int i = 0 ; i < incomingMessage.Count; i++)
+                    for (var i = 0 ; i < incomingMessage.Count; i++)
                     {
-                        System.Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
+                        Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
                         offset += incomingMessage[i].Length;
                     }
-
-                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), Format.Text, receiveResult.EndOfMessage);
+
```
moozzyk: whitespaces

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -40,46 +48,45 @@ public class WebSocketsTransport : ITransport
...
+                    for (var i = 0 ; i < incomingMessage.Count; i++)
                     {
-                        System.Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
+                        Buffer.BlockCopy(incomingMessage[i], 0, totalBuffer, offset, incomingMessage[i].Length);
                         offset += incomingMessage[i].Length;
                     }
-
-                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), Format.Text, receiveResult.EndOfMessage);
+                    
+                    message = new Message(ReadableBuffer.Create(totalBuffer).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
```
moozzyk: Use WebSocketMessageType enum instead of 0. I think you are missing handling a case where the messge type is Close.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -40,46 +48,45 @@ public class WebSocketsTransport : ITransport
...
                 }
                 else
                 {
-                    message = new Message(ReadableBuffer.Create(buffer.Array).Preserve(), Format.Text, receiveResult.EndOfMessage);
+                    message = new Message(ReadableBuffer.Create(incomingMessage[0]).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
```
moozzyk: ditto

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -95,23 +102,33 @@ public class WebSocketsTransport : ITransport
                 {
                     using (message)
                     {
-                        await _webSocket.SendAsync(new ArraySegment<byte>(message.Payload.Buffer.ToArray()),
+                        try
+                        {
+                            await _webSocket.SendAsync(new ArraySegment<byte>(message.Payload.Buffer.ToArray()),
                             message.MessageFormat == Format.Text ? WebSocketMessageType.Text : WebSocketMessageType.Binary, true,
                             cancellationToken);
+                        } catch(OperationCanceledException ex)
```
moozzyk: nit: format
also remove `ex` because you are not using it (you actually should get a warning somewhere)

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -95,23 +102,33 @@ public class WebSocketsTransport : ITransport
...
+                        {
+                            _logger?.LogError("The WebSocket operation was cancelled");
```
moozzyk: Is cancellationToken is guaranteed to be canceled when you get this exception? If not, you should break from the loop.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -95,23 +102,33 @@ public class WebSocketsTransport : ITransport
...
             }
         }
 
-        public async Task Connect(string url)
+        private async Task Connect(Uri url)
         {
-            await Connect(url, "");
-        }
+            var uriBuilder = new UriBuilder(url);
+            if (url.Scheme == "http")
```
moozzyk: Aren't there enums for schemes?

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -95,23 +102,33 @@ public class WebSocketsTransport : ITransport
...
+                        {
+                            _logger?.LogError("The WebSocket operation was cancelled");
+                        }
                     }
                 }
             }
         }
 
-        public async Task Connect(string url)
+        private async Task Connect(Uri url)
```
moozzyk: This is not `Connect` but `prepare url and connect`. I would recommend renaming this method to PrepareUrl/ChangeScheme (or something like this) and just do replacement here and move connect back. Also moving it closer to where it is used will probably make it easier to read.

src/Microsoft.AspNetCore.Sockets.Client/Microsoft.AspNetCore.Sockets.Client.csproj
```diff
@@ -1,21 +1,20 @@
...
     <PackageReference Include="System.IO.Pipelines" Version="0.1.0-*" />
     <PackageReference Include="System.Threading.Tasks.Channels" Version="0.1.0-*" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.2.0-*" />
-    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
   </ItemGroup>
-
-</Project>
+  <ItemGroup>
+    <PackageReference Include="System.Net.WebSockets.Client" Version="4.4.0-*" />
```
davidfowl: Move this to the other item group, it looks lonely over here.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+            if (application == null)
+            {
+                throw new ArgumentNullException(nameof(application));
+            }
+            _application = application;
+            await Connect(url);
+            var sendTask = SendMessages(url, _cancellationToken);
+            var receiveTask = ReceiveMessages(url, _cancellationToken);
+
+            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t => {
```
davidfowl: Maybe wait on @moozzyk 's changes since this will all go away.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                const int bufferSize = 1024;
+                var totalBytes = 0;
+                var incomingMessage = new List<byte[]>();
+                WebSocketReceiveResult receiveResult;
+                do
+                {
+                    var buffer = new ArraySegment<byte>(new byte[bufferSize]);
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                    incomingMessage.Add(buffer.Array);
```
davidfowl: We can't have a List of `Span<byte>`. We'll need to use a list of array segments.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+                }
+            }
+        }
+
+        private async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
+        {
+            while (await _application.Input.WaitToReadAsync(cancellationToken))
+            {
+                Message message;
+                while (!cancellationToken.IsCancellationRequested && _application.Input.TryRead(out message))
```
davidfowl: I don't like this extra condition, let the thing throw an operationcancelledexception

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+                    using (message)
+                    {
+                        try
+                        {
+                            await _webSocket.SendAsync(new ArraySegment<byte>(message.Payload.Buffer.ToArray()),
+                            message.MessageFormat == Format.Text ? WebSocketMessageType.Text : WebSocketMessageType.Binary, true,
+                            cancellationToken);
+                        } catch(OperationCanceledException ex)
+                        {
+                            _logger?.LogError("The WebSocket operation was cancelled");
```
davidfowl: break? Call close?

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                uriBuilder.Scheme = "wss";
+            }
+
+            await _webSocket.ConnectAsync(uriBuilder.Uri, _cancellationToken);
+        }
+
+        public void Dispose()
+        {
+            _webSocket.CloseAsync(WebSocketCloseStatus.Empty, null, _cancellationToken);
```
davidfowl: This is async and needs to block.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+            else if (url.Scheme == "https")
+            {
+                uriBuilder.Scheme = "wss";
+            }
+
+            await _webSocket.ConnectAsync(uriBuilder.Uri, _cancellationToken);
+        }
+
+        public void Dispose()
```
davidfowl: @moozzyk we should remove dispose from the transport and instead just have StopAsync

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: maybe just call `_webSocket.Dispose()` ?

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: @davidfowl - not sure about this. What if the user does not call StopAsync? Each transport has potentially something to dispose - either ClientWebSocket or HttpClient. We should at least try clean up things.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: I would actually say iterate. Converting this to the new pattern is not a lot of work so it does not make sense to block on this. I am planning to send a PR today but it is quite a big change so may have a lot of feedback to address.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: @davidfowl - why?

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
davidfowl: The user doesn't call StopAsync on the transport, we do from the connection. Lets just remove IDisposable from the transport interface itself and leave the graceful unwinding to the  connection.

moozzyk: Only you can't gracefully unwind in Dispose because Dispose is not async. So the Dispose in connection would have to call StopAsync on the transport and ignore the returned task which - as per your comment above https://github.com/aspnet/SignalR/pull/185#discussion_r100467178 we shouldn't do. And indeed calling Dispose on _webSocket is much better in this case. It would also mean that StopAsync is really DisposeAsync because we would have to Dispose all disposable members there.
What is the problem with having Dispose()?

davidfowl: Dispose has to block, that's the problem. I'd rather no block at every layer (connection and transport). Instead, we shutdown and wait on the stop task in Dispose.

moozzyk: Let's discuss this later today. 
(Also, I just realized that there is a race in case someone calls StartAsync on the connection without awaiting and Dispose the connection).

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
anurse: `Span` can't go on the heap. We could use `Memory` but that's not really what it's designed for. So @davidfowl 's right, we need to use `ArraySegment`

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
mikaelm12: Does this belong in this PR? If we decide to do this I can open another PR for it

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,153 @@
...
+                const int bufferSize = 1024;
+                var totalBytes = 0;
+                var incomingMessage = new List<ArraySegment<byte>>();
+                WebSocketReceiveResult receiveResult;
+                do
+                {
+                    var buffer = new ArraySegment<byte>(new byte[bufferSize]);
+                    receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
+                    var truncBuffer = new ArraySegment<byte>(new byte[receiveResult.Count]);
+                    Buffer.BlockCopy(buffer.Array, 0, truncBuffer.Array, 0, receiveResult.Count);
```
anurse: Actually, you don't need to copy, you can create an ArraySegment with the existing buffer, just truncated to the length that was actually read. Something like this: `var truncBuffer = new ArraySegment<byte>(buffer.Array, 0, receiveResult.Count)`. Then you can skip the `BlockCopy` and copying the actual data.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,153 @@
...
+                    {
+                        Buffer.BlockCopy(incomingMessage[i].Array, 0, messageBuffer, offset, incomingMessage[i].Count);
+                        offset += incomingMessage[i].Count;
+                    }
+                    
+                    message = new Message(ReadableBuffer.Create(messageBuffer).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
+                }
+                else
+                {
+                    message = new Message(ReadableBuffer.Create(incomingMessage[0].Array).Preserve(), receiveResult.MessageType == 0 ? Format.Text : Format.Binary, receiveResult.EndOfMessage);
```
anurse: Use the overload of `Create` which takes an offset and a length and use the `Offset` and `Count` properties from `incomingMessage[0]`. `ArraySegment.Array` gives you the **entire** array, not just the specific slice that ended up having data, so we need to make sure to pass the offset and count along everywhere we use it.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -56,21 +57,26 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
...
                 do
                 {
                     var buffer = new ArraySegment<byte>(new byte[bufferSize]);
                     receiveResult = await _webSocket.ReceiveAsync(buffer, cancellationToken);
-                    var truncBuffer = new ArraySegment<byte>(new byte[receiveResult.Count]);
-                    Buffer.BlockCopy(buffer.Array, 0, truncBuffer.Array, 0, receiveResult.Count);
-
+                    if(receiveResult.MessageType == WebSocketMessageType.Close)
+                    {
+                        _application.Output.Complete();
```
moozzyk: return

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -56,21 +57,26 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
         {
             while (!cancellationToken.IsCancellationRequested)
             {
-                const int bufferSize = 1024;
+                const int bufferSize = 1;
```
moozzyk: I guess it's a leftover... 

I suggest you add a test where you send small message (e.g. 17 bytes), medium message (e.g. 11Kb) and a large-ish message (e.g. 7 MB)

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -56,21 +57,26 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
...
+                        _application.Output.Complete();
+                    }
+                    var truncBuffer = new ArraySegment<byte>(buffer.Array, 0, receiveResult.Count);
                     incomingMessage.Add(truncBuffer);
                     totalBytes += receiveResult.Count;
                 } while (!receiveResult.EndOfMessage);
 
+                //Making sure the message type is either text or binary
+                Debug.Assert((receiveResult.MessageType == WebSocketMessageType.Binary || receiveResult.MessageType == WebSocketMessageType.Text ), "Unexpected message type");
+
```
moozzyk: Consider pre-calculating messageType here to a variable and just use it below.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -98,13 +99,14 @@ public static IEnumerable<object[]> MessageSizesData
         {
             get
             {
-                yield return new object[] { new String('A', 5)};
+                yield return new object[] { new string('A', 5)};
```
moozzyk: at least one of the larger messages should not end at the buffer boundary. Otherwise you are not testing a condition where the last message part does not fill the buffer completely.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -68,6 +68,7 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
                     if(receiveResult.MessageType == WebSocketMessageType.Close)
```
moozzyk: One thing we talked about that has not been implemented is handling exceptions thrown ReceiveAsync.

anurse:
Rebasing on `dev` should clear the travis failure.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,166 @@
...
+    public class WebSocketsTransport : ITransport
+    {
+        private ClientWebSocket _webSocket = new ClientWebSocket();
+        private IChannelConnection<Message> _application;
+        private CancellationToken _cancellationToken = new CancellationToken();
+        private readonly ILogger _logger;
+
+        public WebSocketsTransport()
+        {
+            _logger = NullLoggerFactory.Instance.CreateLogger("WebSocketsTransport");
```
moozzyk: Call the other ctor with null and:

`_logger = (loggerFactory ?? NullLoggerFactory.Instance).Instance.CreateLogger("WebSocketsTransport");`

Otherwise there is NRE when you call `new WebSocketTransport(null);`

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        public WebSocketsTransport(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<WebSocketsTransport>();
+        }
```
moozzyk: add empty line

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
moozzyk: This will be done separately

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -75,5 +75,55 @@ public EndToEndTests(ServerFixture serverFixture)
                 }
             }
         }
+
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win2008R2, SkipReason = "No WebSockets Client for this platform")]
+        public async Task ConnectionCanSendAndReceiveSmallMessagesWebSocketsTransport()
```
moozzyk: Do we need this test?

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -75,5 +75,55 @@ public EndToEndTests(ServerFixture serverFixture)
...
+
+            var transport = new WebSocketsTransport();
+            using (var connection = await ClientConnection.ConnectAsync(new Uri(baseUrl + "/echo/ws"), transport, loggerFactory))
+            {
+                await connection.SendAsync(Encoding.UTF8.GetBytes(message), Format.Text);
+
+                var receiveData = new ReceiveData();
+
+                Assert.True(await connection.ReceiveAsync(receiveData).OrTimeout());
+                Assert.Equal(message.Length, Encoding.UTF8.GetString(receiveData.Data).Length);
```
moozzyk: I think it would be good to compare the message contents as well and not only the size.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -0,0 +1,167 @@
...
+        private readonly ILogger _logger;
+
+        public WebSocketsTransport()
+            : this(null)
+        {
+        }
+
+        public WebSocketsTransport(ILoggerFactory loggerFactory)
+        {
+            _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger("WebSocketsTransport");
```
BrennanConroy: nameof!

Issue #174, opened 2017-01-31
Implement Transport Protocol spec


anurse:
* [ ] Implement Text-based Long-Polling/HTTP Post format (https://github.com/aspnet/SignalR/pull/187)
* [ ] Implement Binary Long-Polling/HTTP Post format
* [ ] Add proper newline handling to SSE
* [ ] Add frame type syntax to SSE
* [ ] Update .NET and TS Clients to match changes
* [ ] Ensure HTTP status codes are correct
* [ ] Wrap exceptions in the application and produce correct error frames in all transports.

/cc @davidfowl

anurse:
Closing this, individual issues have been filed.

PR #187, opened 2017-02-09
Text Protocol Formatter

anurse:
Part of #174 

Provides the code for reading/writing message batches using the "Text" protocol defined in the spec. It's not wired up to anything, just wanted to get it in and have tests. Next I'll do binary.


src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+            cursor++;
+
+            // Separator
+            target[cursor] = (byte)':';
+            cursor++;
+
+            // Payload
+            if(message.Type == MessageType.Binary)
+            {
+                var payload = Encoding.UTF8.GetBytes(Convert.ToBase64String(message.Payload.Buffer.ToArray()));
```
davidfowl: Use the base64 encoder library in corefxlab

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+                    throw new FormatException("Invalid length.");
+                }
+                length = (length * 10) + (data[i] - '0');
+            }
+
+            return length;
+        }
+
+        // There isn't a Span.IndexOf that takes a start point :(.
+        private static int IndexOf(byte c, ReadOnlySpan<byte> data, int start)
```
davidfowl: Slice? /cc @KrzysztofCwalina

anurse: You had previously indicated that Slice was still slow and that in general we should use indexing. Happy to use slice. It'll be cleaner and easier :)

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
anurse: Was going to, but we can't get it mirrored right now: https://github.com/aspnet/CoreCLR/pull/150

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Attempts to write the specified messages to the target buffer
+        /// </summary>
+        /// <param name="target">The buffer in which to encode the messages</param>
+        /// <param name="messages">The messages to encode</param>
+        /// <returns>A boolean indicating if there was enough space in the buffer to encode the message.</returns>
+        public static bool WriteMessages(Span<byte> target, IReadOnlyList<Message> messages)
```
mikaelm12: Should this maybe be called TryWriteMessages since it might not actually write?

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
+﻿using System;
```
mikaelm12: nit: Header

SignalR.sln
```diff
@@ -58,100 +57,218 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.Signal
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.Sockets.Common", "src\Microsoft.AspNetCore.Sockets.Common\Microsoft.AspNetCore.Sockets.Common.csproj", "{F3EFFD9F-DD85-48A2-9B11-83A133ECC099}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.Sockets.Common.Tests", "test\Microsoft.AspNetCore.Sockets.Common.Tests\Microsoft.AspNetCore.Sockets.Common.Tests.csproj", "{B0D32729-48AA-4841-B52A-2A61B60EED61}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
moozzyk: do we need all these configurations?

test/Microsoft.AspNetCore.Sockets.Common.Tests/TextMessageBatchFormatterTests.cs
```diff
@@ -0,0 +1,144 @@
+﻿using System;
```
mikaelm12: Here too 😄

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+            // Add enough character to represent the length in digits
+            // Log10(0) == Infinity, so don't allow that to happen.
+            var digits = len == 0 ? 1 : (int)Math.Floor(Math.Log10(len)) + 1;
+            size += digits;
+            return size;
+        }
+
+        private static PreservedBuffer ParsePayload(ReadOnlySpan<byte> data, int length, MessageType messageType, ref int cursor)
+        {
+            int start = cursor;
```
moozzyk: nit: var (in all places)

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
KrzysztofCwalina: Also, should it not return how many bytes it has written? In general, why not follow signature pattern similar to primitive formatters: https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.cs

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        /// <summary>
+        /// Attempts to write the specified messages to the target buffer
+        /// </summary>
+        /// <param name="target">The buffer in which to encode the messages</param>
+        /// <param name="messages">The messages to encode</param>
+        /// <returns>A boolean indicating if there was enough space in the buffer to encode the message.</returns>
+        public static bool WriteMessages(Span<byte> target, IReadOnlyList<Message> messages)
+        {
+            var size = MeasureMessages(messages);
```
KrzysztofCwalina: You might want to implement IBufferFormattable on message: https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Formatting/IBufferFormattable.cs

And then just lop to write each of the messages. If a write fails, you return false from this routine, i.e. no need to measure.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            if (cursor - start != 1)
+            {
+                throw new FormatException("Type field must be exactly one byte long.");
+            }
+
+            switch (data[cursor - 1])
+            {
+                case (byte)'T': return MessageType.Text;
```
moozzyk: Consider creating consts for all these `(byte)'X'`

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
anurse: Nice, that's exactly the primitive @davidfowl and I have been talking about adding. I'll have to see if we already mirror that package, if not we may have to wait until we resolve some of the dependency conflict issues.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+                WriteMessage(target, message, ref cursor);
+            }
+            return true;
+        }
+
+        private static void WriteMessage(Span<byte> target, Message message, ref int cursor)
+        {
+            // Write the length as a UTF-8 string
+            var len = message.Payload.Buffer.Length.ToString();
+            var buf = Encoding.UTF8.GetBytes(len);
```
KrzysztofCwalina: you should do:
PrimitiveFormatter.TryFormat(message.Payload.Buffer.Length, target, out var written, default(TextFormat));

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
anurse: The original intent was that you'd measure and pass in a Span that is exactly big enough. Now that I've got an idea of how the formatter APIs work, I'll refactor this to fix those better.

SignalR.sln
anurse: No idea. VS added them and I figured it had a reason and would keep trying to do it if I undid it.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
anurse: Hmm... If I do that though, it'll be a little tricky to support the two different formats we want to write Messages to. IBufferFormattable seems to be designed for types that have only one wire format, but in our case we're going to be having two different formats (a Text-based one and a Binary one) in order to support various levels of browser compatibility.

SignalR.sln
moozzyk: Meh.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
KrzysztofCwalina: You could use TextFormat parameter to control that. If you don't like the name "text", we can (@shiftylogic, should?) rename that.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
KrzysztofCwalina: Why do you need it to be mirrored? Why not simply reference the package?

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            // If the message is binary, we need to convert from Base64
+            if (messageType == MessageType.Binary)
+            {
+                // TODO: Use System.Binary.Base64 to handle this with less allocation
+
+                // Parse the data as Base64
+                var str = Encoding.UTF8.GetString(buffer);
+                buffer = Convert.FromBase64String(str);
+            }
```
KrzysztofCwalina: You should not be copying out of data. You should use PrimitiveBarser to parse directly, and Base64 (decoder) to decode directly from the span.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            // Scan to a ':'
+            cursor = IndexOf((byte)':', data, cursor);
+            if (cursor >= data.Length)
+            {
+                throw new FormatException("Unexpected end-of-message while reading Length field.");
+            }
+
+            // Parse the length
+            int length = 0;
```
KrzysztofCwalina: You should use PrimitiveParser.TryParseInt32

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
anurse: Ok, I'll do some digging in that. I definitely agree that some of the formatting APIs appear (by naming) to be tied to "text", even though they operate on bytes.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageBatchFormatter.cs
anurse: We have dependency conflict issues with the latest versions of packages in corefxlab. @davidfowl would have more information.

SignalR.sln
davidfowl: nooooo

anurse:
OK, I've reworked it to use the new Formatting/Parsing APIs. I didn't end up using `TextFormat` because it seems somewhat difficult to adapt to this purpose, when all I really need is an indicator of two different types. I also don't like data types like `Message` having their formatting/parsing logic inside the type, I'd rather that logic be extracted elsewhere.

The `MessageBatchFormatter` is the primary entry point into the parse/format APIs.

src/Microsoft.AspNetCore.Sockets.Common/MessageBatchFormatter.cs
```diff
@@ -0,0 +1,134 @@
...
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    public static class MessageBatchFormatter
+    {
+        private const byte BatchTerminator = (byte)';';
+        private const byte TextFormatIndicator = (byte)'T';
+        private const byte BinaryFormatIndicator = (byte)'B';
+
+        public static bool TryFormatMessages(IEnumerable<Message> messages, Span<byte> buffer, MessageFormat format, out int bytesWritten)
```
KrzysztofCwalina: The routine either completely succeeds or fails. Would it be better if it formatted as many messages as it fit in the buffer and then returned the index (or enumerator) of the last one that fit?

src/Microsoft.AspNetCore.Sockets.Common/MessageBatchFormatter.cs
```diff
@@ -0,0 +1,134 @@
...
+                return false;
+            }
+
+            bytesConsumed = consumedSoFar + 1;
+            messages = readMessages;
+            return true;
+        }
+
+        private static bool TryParseMessageFormat(ReadOnlySpan<byte> buffer, out MessageFormat format, out int bytesConsumed)
+        {
```
KrzysztofCwalina: Depending how fast you want the code to be, this routine might be a bit of an overkill, i.e. you know you will consume 1 byte, etc. Maybe it should be a simple check inside Format/Parse messages.

src/Microsoft.AspNetCore.Sockets.Common/TextMessageFormatter.cs
```diff
@@ -0,0 +1,235 @@
...
+            {
+                bytesWritten = 0;
+                return false;
+            }
+            buffer[2] = FieldDelimiter;
+            buffer = buffer.Slice(3);
+            written += 3;
+
+            // Payload
+            if (message.Type == MessageType.Binary)
```
KrzysztofCwalina: Hmmm, higher up you throw NYI when message type is binary. Is that intentional?

anurse: Yes, the full implementation of the binary format is coming in a separate PR later. This was a trivial thing to add since the prefix is almost the same as for the text format.

src/Microsoft.AspNetCore.Sockets.Common/MessageBatchFormatter.cs
anurse: Yeah, I was trying to figure out the right pattern for this and probably went overkill on this one :)

src/Microsoft.AspNetCore.Sockets.Common/MessageBatchFormatter.cs
anurse: Well, in general the desired goal is to write the entire batch in a single HTTP response when long-polling (to avoid round-trips to get more messages), so it would still have to make clear that it failed to write so that the caller could get more space for it to continue writing. Also, there'd need to be state-tracking to ensure that when resuming writing it doesn't write the prefix again.

I'm going to leave it this way for today and do some reviewing of how we're going to use it over the weekend. I definitely felt, as I was writing it, that it seemed odd to get 99% of the way through writing only to fail and discard it all.

anurse: Actually, moving the "format-discriminator" writing/parsing code out of this might make this much cleaner, since that's only ever needed at the very start of the entire stream. I'm going to refactor a bit to make this work.

anurse:
Updated. I removed the batch formatting logic since it's not really a pure formatting/parsing concern, there are things like incomplete messages (a message with `endOfMessage` being false, which must be buffered and assembled before transmission because unlike WebSockets we don't support that in this protocol, by design), or insufficient buffer space and partial writes/reads to be considered.

anurse:
Investigating the travis failure, it appears to be unrelated (it'll need a fix in `dev` obvs.)

anurse:
@davidfowl I'd like to close this out today if possible. Already working on the binary one and nearly ready to submit that one :).

moozzyk:
:clap:

PR #188, opened 2017-02-09
add a terminator to message batches

anurse:
This a) helps if we ever need to support this protocol on a streaming transport (say TCP?) and b) helps the protocol fit better into the Formatter APIs coming in corefxlab

### DIFF
--- a/specs/TransportProtocols.md
+++ b/specs/TransportProtocols.md
@@ -127,12 +127,14 @@ If the `connectionId` parameter is missing, a `400 Bad Request` response is retu
 The body will be formatted as below and encoded in UTF-8. The `Content-Type` response header is set to `application/vnd.microsoft.aspnet.endpoint-messages.v1+text`. Identifiers in square brackets `[]` indicate fields defined below, and parenthesis `()` indicate grouping.
 
 ```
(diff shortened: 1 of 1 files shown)


specs/TransportProtocols.md
```diff
@@ -127,13 +127,15 @@ If the `connectionId` parameter is missing, a `400 Bad Request` response is retu
 The body will be formatted as below and encoded in UTF-8. The `Content-Type` response header is set to `application/vnd.microsoft.aspnet.endpoint-messages.v1+text`. Identifiers in square brackets `[]` indicate fields defined below, and parenthesis `()` indicate grouping.
 
 ```
-T([Length]:[Type]:[Body];)([Length]:[Type]:[Body];)... continues until end of the response body ...
+T([Length]:[Type]:[Body];)([Length]:[Type]:[Body];)... more messages ...([Length]:[Type]:[Body];);
```
moozzyk: This spec SHOULD use BNF :trollface:

anurse:
```
<Pawel> ::= <Pedantic><Nerd>
```

😛

anurse:
Actually, I don't think this is necessary any more. Closing

PR #191, opened 2017-02-10
Pawelka/events

moozzyk:
- changing static ConnectAsync to instance StartAsync 
- adding Connected, Received, Closed events

There is a few items that still need attention but are not blocking. I don't want to make this change any bigger. I will create a comment with a running issues to address. Feedback will be addressed in this PR or, if bigger/not-critical added to this list to be addressed separately.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
@@ -25,6 +25,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public Uri Url { get; }
 
+        public event Action Connected;
+        public event Action<byte[], Format> Received;
+        public event Action<Exception> Closed;
+
         public Connection(Uri url)
             : this(url, null)
         { }
@@ -64,6 +68,13 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
 
             Interlocked.Exchange(ref _connectionState, ConnectionState.Connected);
+
+            // Do not "simplify" - events can be removed from a different thread
+            var connectedEventHandler = Connected;
+            if (connectedEventHandler != null)
+            {
+                connectedEventHandler();
+            }
         }
 
         private static async Task<Uri> GetConnectUrl(Uri url, HttpClient httpClient, ILogger logger)
@@ -110,7 +121,17 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             _transportChannel = new ChannelConnection<Message>(applicationToTransport, transportToApplication);
 #pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
-            Input.Completion.ContinueWith(t => Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected));
+            Input.Completion.ContinueWith(t =>
+            {
+                Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected);
+
+                // Do not "simplify" - events can be removed from a different thread
+                var closedEventHandler = Closed;
+                if (closedEventHandler != null)
+                {
+                    closedEventHandler(t.IsFaulted ? t.Exception.InnerException : null);
+                }
+            });
 #pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
 
             // Start the transport, giving it one end of the pipeline
--- a/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
@@ -14,23 +14,92 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public class Connection : IDisposable
     {
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly ILogger _logger;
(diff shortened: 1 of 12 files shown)


moozzyk:
Running list of issues:
- [ ] consider adding Disposed state
- [ ] throw exception when sending and the connection is before Connected or in Disposed state
- [ ] fix a possible race when StartAsync is not awaited and StopAsync/Dispose is called
- [ ] HubConnection.Invoke should check the value returned by Connection.SendAsync
- [ ] more unit tests for HubConnection (mostly around error conditions) - see IConnection
- [ ] consider creating IConnection interface and enable passign IConnection to HubConnection. Would simplify testing HubConnection and will enable using HubConnection with custom connections (probably not common)
- [ ] event handlers should not run on main SignalR thread
- [ ] simplify HubConnection ctors - ILoggerFactory should not be required, IInvocationAdapter?
- [ ] Received event should not be raised after the connection is disposed (what about stopped)

moozzyk:
:watch:

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -17,7 +17,30 @@ namespace Microsoft.AspNetCore.Sockets.Client.Tests
...
+        {
+            var exception = Assert.Throws<ArgumentNullException>(() => new Connection(null));
+            Assert.Equal("url", exception.ParamName);
+        }
+
+        [Fact]
+        public void ConnectionReturnsUrlUsedToStartTheConnection()
+        {
+            var connectionUrl = new Uri("http://fakeuri.org/");
+            using (var connection = new Connection(connectionUrl))
```
BrennanConroy: This connection isn't being used

test/Microsoft.AspNetCore.Sockets.Client.Tests/HubConnectionTests.cs
```diff
@@ -0,0 +1,188 @@
+﻿using Microsoft.AspNetCore.Sockets;
```
BrennanConroy: Sort and header

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -14,70 +14,174 @@ namespace Microsoft.AspNetCore.Sockets.Client
...
-            _logger = logger;
-            _transport = transport;
-            _transportChannel = transportChannel;
+            _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
+            _logger = _loggerFactory.CreateLogger<Connection>();
         }
 
-        private ReadableChannel<Message> Input => _transportChannel.Input;
-        private WritableChannel<Message> Output => _transportChannel.Output;
+        public Task StartAsync() => StartAsync(null, null);
```
davidfowl: optional parameter syntax for null and bool args like this:

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -14,70 +14,174 @@ namespace Microsoft.AspNetCore.Sockets.Client
...
+        }
+
+        private async Task StartTransport(Uri connectUrl)
+        {
+            var applicationToTransport = Channel.CreateUnbounded<Message>();
+            var transportToApplication = Channel.CreateUnbounded<Message>();
+            var applicationSide = new ChannelConnection<Message>(transportToApplication, applicationToTransport);
+
+            _transportChannel = new ChannelConnection<Message>(applicationToTransport, transportToApplication);
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
```
davidfowl: Get rid of this and just use the pattern we use in other places. ignore = ....

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: I prefer when Intellisense basically Shows `StartAsync()` instead of overwhelming you with `StartAsync(ITransport transport = null, HttpClient httpClient = null)` - especially for most common functions.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: OK. (although I still prefer `_` for that 😄 )

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
davidfowl: I'm just talking about consumption syntax not removing overloads.

moozzyk: Oh, named optional parameters. :+1:

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
anurse: You can even do `_ = [expression]` in C# now, you don't need to declare a variable.

davidfowl: no _

moozzyk: :(

PR #193, opened 2017-02-11
Making `OrTimeout` common

moozzyk:
For now I just use linked source. If there is more common utils to share we can create a common test project.

### DIFF
--- a/test/Common/TaskExtensions.cs
+++ b/test/Common/TaskExtensions.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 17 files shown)


moozzyk:
@anurse - we also have a copy of `OrTimeout` in WebSockets tests. Do you want me to get rid of it and use the one I moved to `Common`

anurse:
Yeah, yours is much simpler... I'm... not sure what I was thinking when I wrote mine :)

moozzyk:
:up::date:

anurse:
Ah, I remember why I wrote mine that way. I wrote the cancellation token version first for other purposes and just kinda kept using that technique. I still like yours way better ;).

PR #198, opened 2017-02-13
Fix travis failure

anurse:
There was a conflict between a globally-installed python package and a package autobahn needs, so we remove `sudo` from the installation call and it doesn't try to mess with those packages.


BrennanConroy:
Still failures :/

anurse:
Yeah, it's something else though, and it's been happening for a while undetected. Looking into it now

anurse:
Looks like the MissingMethodException was transient.

Issue #144, opened 2017-01-23
Port ObjectMethodExecutor to HubEndPoint


davidfowl:
https://github.com/aspnet/Mvc/blob/13b32adeae0bd13f210a166c7d1a0dcb2a917611/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs

Use the same method invocation execution pipeline as MVC.

BrennanConroy:
#199

PR #199, opened 2017-02-14
Use ObjectMethodExecutor for hub method invocations

BrennanConroy:
#144

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR.Internal;
 using Microsoft.AspNetCore.Sockets;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
@@ -251,7 +252,7 @@ namespace Microsoft.AspNetCore.SignalR
                 Id = invocationDescriptor.Id
             };
 
-            var methodInfo = descriptor.MethodInfo;
+            var methodExecutor = descriptor.MethodExecutor;
 
             using (var scope = _serviceScopeFactory.CreateScope())
             {
@@ -262,21 +263,24 @@ namespace Microsoft.AspNetCore.SignalR
                 {
                     InitializeHub(hub, connection);
 
-                    var result = methodInfo.Invoke(hub, invocationDescriptor.Arguments);
-                    var resultTask = result as Task;
-                    if (resultTask != null)
+                    object result = null;
+                    if (methodExecutor.IsMethodAsync)
                     {
-                        await resultTask;
-                        if (methodInfo.ReturnType.GetTypeInfo().IsGenericType)
+                        if (methodExecutor.TaskGenericType == null)
                         {
-                            var property = resultTask.GetType().GetProperty("Result");
-                            invocationResult.Result = property?.GetValue(resultTask);
+                            await (Task)methodExecutor.Execute(hub, invocationDescriptor.Arguments);
+                        }
+                        else
+                        {
+                            result = await methodExecutor.ExecuteAsync(hub, invocationDescriptor.Arguments);
                         }
                     }
                     else
                     {
-                        invocationResult.Result = result;
+                        result = methodExecutor.Execute(hub, invocationDescriptor.Arguments);
                     }
+
+                    invocationResult.Result = result;
                 }
                 catch (TargetInvocationException ex)
                 {
@@ -306,9 +310,9 @@ namespace Microsoft.AspNetCore.SignalR
 
         private void DiscoverHubMethods()
         {
-            var type = typeof(THub);
+            var typeInfo = typeof(THub).GetTypeInfo();
 
-            foreach (var methodInfo in type.GetTypeInfo().DeclaredMethods.Where(m => IsHubMethod(m)))
+            foreach (var methodInfo in typeInfo.DeclaredMethods.Where(m => IsHubMethod(m)))
             {
                 var methodName = methodInfo.Name;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.SignalR/Internal/ObjectMethodExecutor.cs
```diff
@@ -0,0 +1,264 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    class ObjectMethodExecutor
```
davidfowl: internal

src/Microsoft.AspNetCore.SignalR/Internal/ObjectMethodExecutor.cs
```diff
@@ -0,0 +1,262 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    internal class ObjectMethodExecutor
```
moozzyk: Can we add tests for this class? I don't think our tests cover all the cases like static vs. instance methods, Task vs Task<T> vs, void return types, default parameter values etc.

BrennanConroy: Here's the original tests https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ObjectMethodExecutorTest.cs
Handles most scenarios, do you want to copy them?

moozzyk: I guess so - we will rev this code independently.

src/Microsoft.AspNetCore.SignalR/Internal/ObjectMethodExecutor.cs
```diff
@@ -0,0 +1,262 @@
...
+                var valueCast = Expression.Convert(valueObj, paramInfo.ParameterType);
+
+                // valueCast is "(Ti) parameters[i]"
+                parameters.Add(valueCast);
+            }
+
+            // Call method
+            MethodCallExpression methodCall;
+
+            if (!methodInfo.IsStatic)
```
BrennanConroy: FYI: This is the stuff I added to the class to allow for static methods

moozzyk: The very reason why we need to have our own tests

src/Microsoft.AspNetCore.SignalR/Internal/ObjectMethodExecutor.cs
anurse: Unless we're going to actually make this a common library/shared package with tests, we need to test it on our end as well.

src/Microsoft.AspNetCore.SignalR/Internal/ObjectMethodExecutor.cs
```diff
@@ -0,0 +1,262 @@
...
+
+        private static readonly MethodInfo _convertOfTMethod =
+            typeof(ObjectMethodExecutor).GetRuntimeMethods().Single(methodInfo => methodInfo.Name == nameof(ObjectMethodExecutor.Convert));
+
+        private ObjectMethodExecutor(MethodInfo methodInfo, TypeInfo targetTypeInfo)
+        {
+            if (methodInfo == null)
+            {
+                throw new ArgumentNullException(nameof(methodInfo));
+            }
```
mikaelm12: nit: new line here

src/Microsoft.AspNetCore.SignalR/Internal/ObjectMethodExecutor.cs
davidfowl: @rynowak should we make this a shared component?

rynowak:
Yeah it would be dope to make this shared. For instance we're going to be adding F# support to it soon. Y'all will want that as well.

anurse:
I think a shared-source package would be all we need. It's not quite worthy of it's own assembly.

BrennanConroy:
Couple minor things, they reference IActionResult and we need static method support and they don't

src/Microsoft.AspNetCore.SignalR/Internal/ObjectMethodExecutor.cs
davidfowl: It's possible we want to disallow this.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -262,21 +263,24 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
+                    object result = null;
+                    if (methodExecutor.IsMethodAsync)
                     {
-                        await resultTask;
-                        if (methodInfo.ReturnType.GetTypeInfo().IsGenericType)
+                        if (methodExecutor.TaskGenericType == null)
                         {
-                            var property = resultTask.GetType().GetProperty("Result");
-                            invocationResult.Result = property?.GetValue(resultTask);
+                            await (Task)methodExecutor.Execute(hub, invocationDescriptor.Arguments);
```
davidfowl: Is this pretty weird no? Why doesn't ExecuteAsync work here? If not that seems like worthy refactoring after this is merged.

BrennanConroy: ExecuteAsync calls TaskOfTActionExecutorAsync which assumes you have a Task<T> return type. I would be okay with refactoring it.

PR #200, opened 2017-02-14
Creating npm package for ts client

moozzyk:


anurse:
Could we put the `package.json` in the `src/Microsoft.AspNetCore.SignalR.Client.TS` folder? Seems awkward to have it at the repo root.

moozzyk:
Not sure. 2 reasons: 
- Samples and tests depend on that because they use browserify and gulp. If this is moved to a subfolder we will have to run npm install multiple time (slow) on multiple levels (I don't even think is supported currently by the build system) 
- in npm repo is a package if I moved it to a subfolder I would have to place distributable files to a subfolder of src/Microsoft.AspNetCore.SignalR.Client.TS to be able to publish the package. The src/Microsoft.AspNetCore.SignalR.Client.TS would have to have the .npmignore file and things are getting even messier. In the other PR I copy files from the dist folder which is now in the root - if this is moved they would have to search for the build artifacts in the src/Microsoft.AspNetCore.SignalR.Client.TS

As I said today during the meeting - it almost feels like the TS client should be in a separate repo (or we could create a repo just for building the npm package and consume JS client as an npm package here but we would always be 1 commit behind)

moozzyk:
⌚️

PR #203, opened 2017-02-15
finish binary protocol formatter/parser

### LINKED ISSUE #202: Implement Binary Protocol Formatter/Parser
Just the formatter/parser, not wired up to anything yet. Part of #137

anurse:
fixes #202

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
@@ -1,0 +1,127 @@
+﻿using System;
+using System.Binary;
+using System.IO.Pipelines;
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    internal static class BinaryMessageFormatter
+    {
+        private const byte TextTypeFlag = 0x00;
+        private const byte BinaryTypeFlag = 0x01;
+        private const byte ErrorTypeFlag = 0x02;
+        private const byte CloseTypeFlag = 0x03;
+
+        internal static bool TryFormatMessage(Message message, Span<byte> buffer, out int bytesWritten)
+        {
+            // We can check the size needed right up front!
+            var sizeNeeded = sizeof(long) + 1 + message.Payload.Buffer.Length;
+            if(buffer.Length < sizeNeeded)
+            {
+                bytesWritten = 0;
+                return false;
+            }
+
+            buffer.WriteBigEndian((long)message.Payload.Buffer.Length);
+            if(!TryFormatType(message.Type, buffer.Slice(sizeof(long), 1)))
+            {
+                bytesWritten = 0;
+                return false;
+            }
+
+            buffer = buffer.Slice(sizeof(long) + 1);
+
+            message.Payload.Buffer.CopyTo(buffer);
+            bytesWritten = sizeNeeded;
+            return true;
+        }
+
+        internal static bool TryParseMessage(ReadOnlySpan<byte> buffer, out Message message, out int bytesConsumed)
+        {
+            // Check if we have enough to read the size and type flag
+            if(buffer.Length < sizeof(long) + 1)
+            {
+                message = default(Message);
+                bytesConsumed = 0;
+                return false;
+            }
+
+            // REVIEW: The spec calls for 64-bit length but I'm thinking that's a little ridiculous.
+            // REVIEW: We don't really have a primitive for storing that much data. For now, I'm using it
+            // REVIEW: but throwing if the size is over 2GB.
+            var longLength = buffer.ReadBigEndian<long>();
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
```diff
@@ -0,0 +1,130 @@
...
+                    messageType = MessageType.Binary;
+                    return true;
+                case CloseTypeFlag:
+                    messageType = MessageType.Close;
+                    return true;
+                case ErrorTypeFlag:
+                    messageType = MessageType.Error;
+                    return true;
+                default:
+                    messageType = MessageType.Text;
```
BrennanConroy: default(...)

src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
```diff
@@ -0,0 +1,130 @@
...
+                return false;
+            }
+
+            // Copy the payload into the buffer
+            // REVIEW: Copy! Noooooooooo! But how can we capture a segment of the span as an "Owned" reference?
+            // REVIEW: If we do have to copy, we should at least use a pooled buffer
+            var buf = new byte[length];
+            buffer.Slice(sizeof(long) + 1, length).CopyTo(buf);
+
+            message = new Message(ReadableBuffer.Create(buf).Preserve(), messageType, endOfMessage: true);
```
davidfowl: Did we decide to keep endOfMessage ?

src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            // We can check the size needed right up front!
+            var sizeNeeded = sizeof(long) + 1 + message.Payload.Buffer.Length;
+            if (buffer.Length < sizeNeeded)
+            {
+                bytesWritten = 0;
+                return false;
+            }
+
+            buffer.WriteBigEndian((long)message.Payload.Buffer.Length);
```
davidfowl: These APIs suck. We need to change then to WriteLong etc.

src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            // Check if we have enough to read the size and type flag
+            if (buffer.Length < sizeof(long) + 1)
+            {
+                message = default(Message);
+                bytesConsumed = 0;
+                return false;
+            }
+
+            // REVIEW: The spec calls for 64-bit length but I'm thinking that's a little ridiculous.
```
davidfowl: I think I agree but there's going to be that dude that needs this 😄

test/Microsoft.AspNetCore.Sockets.Common.Tests/MessageFormatterTests.Text.cs
```diff
@@ -9,10 +9,10 @@
 
 namespace Microsoft.AspNetCore.Sockets.Tests
 {
-    public class MessageFormatterTests
+    public partial class MessageFormatterTests
```
moozzyk: Not a big fan. I think it would be better to have TextFormatMessageFormatterTests and BinaryFormatMessageFormatterTests. I would not change the test names though - the prefix will make it much easier to filter tests in the test explorer (when it happens to work)

moozzyk:
LGTM

src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
anurse: My understanding was that we kept it on the type but not in the non-WebSockets transports. That way we preserve fidelity when writing a WebSocket-based server.

test/Microsoft.AspNetCore.Sockets.Common.Tests/MessageFormatterTests.Text.cs
anurse: Sure, I don't really have a good clear pattern for this so I'm happy to adjust.

src/Microsoft.AspNetCore.Sockets.Common/BinaryMessageFormatter.cs
anurse: The biggest thing is that absolutely **nothing** in our stack can handle a 64-bit length buffer. We'll need to build a bunch of new primitives (buffer chains, etc.) to handle it. And a 2GB-long **single message** is really ridiculous. If someone wants to upload files over 2GB via this, they just have to split it into separate messages, which they **really should do already**.

PR #210, opened 2017-02-15
implemented server-sent events formatter

### LINKED ISSUE #204: Implement Server-Sent Events Protocol Formatter/Parser
Similar to long-polling, we have a "protocol" for Server Sent events (basically the SSE protocol plus a few specific details). We need to build a formatter/parser for that.

Part of #174

anurse:
fixes #204 

Another formatter/parser and tests. Still not wired up :).

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
@@ -1,0 +1,186 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO.Pipelines;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Sockets.Formatters
+{
+    public static class ServerSentEventsMessageFormatter
+    {
+        private static readonly Span<byte> DataPrefix = new byte[] { (byte)'d', (byte)'a', (byte)'t', (byte)'a', (byte)':', (byte)' ' };
+        private static readonly Span<byte> Newline = new byte[] { (byte)'\r', (byte)'\n' };
+
+        private const byte LineFeed = (byte)'\n';
+        private const byte TextTypeFlag = (byte)'T';
+        private const byte BinaryTypeFlag = (byte)'B';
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
```diff
@@ -0,0 +1,186 @@
...
+                Newline.CopyTo(buffer);
+                writtenSoFar += encodedSize;
+                buffer.Slice(Newline.Length);
+            }
+            else
+            {
+                while (true)
+                {
+                    // Seek to the end of buffer or newline
+                    var sliced = payload.TrySliceTo(LineFeed, out var slice, out var cursor);
```
davidfowl: If we're going to move away from `PreservedBuffer` you'll want to change this.

src/Microsoft.AspNetCore.Sockets.Common/Microsoft.AspNetCore.Sockets.Common.csproj
```diff
@@ -8,6 +8,7 @@
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;signalr</PackageTags>
+    <RootNamespace>Microsoft.AspNetCore.Sockets</RootNamespace>
```
davidfowl: ?

src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
anurse: Yeah, I'm going to do that separately since I had already written this code and the tests when we decided that.

src/Microsoft.AspNetCore.Sockets.Common/Microsoft.AspNetCore.Sockets.Common.csproj
anurse: We don't usually put things in the actual `.Common` namespace when we make a `.Common` project.

anurse: Or if we do, we probably shouldn't, it's weird

PR #218, opened 2017-02-16
Fix #215 and restore tests

### LINKED ISSUE #215: Fix code broken by new version of Pipelines
After updating to the new version of Pipelines tests are failing with:

```
   System.InvalidOperationException : Can't complete reader while consuming operation is ongoing.
  Stack Trace:
     at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource, String location)
     at System.IO.Pipelines.Pipe.System.IO.Pipelines.IPipeReader.Complete(Exception exception)
     at Microsoft.Extensions.WebSockets.Internal.WebSocketConnection.Dispose() in C:\Source\SignalR-Core\src\Microsoft.Extensions.WebSockets.Internal\WebSocketConnection.cs:line 135
     at Microsoft.Extensions.WebSockets.Internal.Tests.WebSocketPair.Dispose() in C:\Source\SignalR-Core\test\Microsoft.Extensions.WebSockets.Internal.Tests\WebSocketPair.cs:line 53
     at Microsoft.AspNetCore.Sockets.Tests.WebSocketsTests.<TransportFailsWhenClientDisconnectsAbnormally>d__5.MoveNext() in C:\Source\SignalR-Core\test\Microsoft.AspNetCore.Sockets.Tests\WebSocketsTests.cs:line 286
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

Fix the issues an revert: https://github.com/aspnet/SignalR/commit/ce4a8f027608068df8dc693c66f2acd714398d4b

### LINKED ISSUE #251: Integrate TS client into SignalR build and solution

anurse:
This should properly clean up the Pipes.

Fixes #215 

/cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -75,7 +75,7 @@ namespace Microsoft.AspNetCore.Sockets.Transports
             {
                 if (receiving.IsCanceled || receiving.IsFaulted)
                 {
-                    // The receiver faulted or cancelled. This means the client is probably broken. Just propagate the exception and exit
+                    // The receiver faulted or cancelled. This means the socket is probably broken. Abort the socket and propagate the exception
                     receiving.GetAwaiter().GetResult();
 
                     // Should never get here because GetResult above will throw
--- a/src/Microsoft.Extensions.WebSockets.Internal/Microsoft.Extensions.WebSockets.Internal.csproj
+++ b/src/Microsoft.Extensions.WebSockets.Internal/Microsoft.Extensions.WebSockets.Internal.csproj
@@ -13,10 +13,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.ValueTuple" Version="$(CoreFxVersion)" />
     <PackageReference Include="System.IO.Pipelines" Version="$(CoreFxLabsVersion)" />
     <PackageReference Include="System.IO.Pipelines.Text.Primitives" Version="$(CoreFxLabsVersion)" />
     <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
-    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.Extensions.WebSockets.Internal/Microsoft.Extensions.WebSockets.Internal.csproj
+++ b/src/Microsoft.Extensions.WebSockets.Internal/Microsoft.Extensions.WebSockets.Internal.csproj
@@ -15,7 +15,8 @@
   <ItemGroup>
     <PackageReference Include="System.IO.Pipelines" Version="$(CoreFxLabsVersion)" />
     <PackageReference Include="System.IO.Pipelines.Text.Primitives" Version="$(CoreFxLabsVersion)" />
-    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.Extensions.WebSockets.Internal/PipeReaderExtensions.cs
+++ b/src/Microsoft.Extensions.WebSockets.Internal/PipeReaderExtensions.cs
@@ -1,31 +1,29 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Threading;
-using System.Threading.Tasks;
 using System.IO.Pipelines;
+using System.Threading.Tasks;
 
 namespace Microsoft.Extensions.WebSockets.Internal
 {
     public static class PipeReaderExtensions
     {
-        public static ValueTask<ReadResult> ReadAtLeastAsync(this IPipeReader input, int minimumRequiredBytes) => ReadAtLeastAsync(input, minimumRequiredBytes, CancellationToken.None);
-
         // TODO: Pull this up to Channels. We should be able to do it there without allocating a Task<T> in any case (rather than here where we can avoid allocation
         // only if the buffer is already ready and has enough data)
-        public static ValueTask<ReadResult> ReadAtLeastAsync(this IPipeReader input, int minimumRequiredBytes, CancellationToken cancellationToken)
+        public static async ValueTask<ReadResult> ReadAtLeastAsync(this IPipeReader input, int minimumRequiredBytes)
         {
             var awaiter = input.ReadAsync(/* cancellationToken */);
 
             // Short-cut path!
+            ReadResult result;
             if (awaiter.IsCompleted)
             {
                 // We have a buffer, is it big enough?
-                var result = awaiter.GetResult();
+                result = awaiter.GetResult();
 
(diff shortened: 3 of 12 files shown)


src/Microsoft.Extensions.WebSockets.Internal/PipeReaderExtensions.cs
```diff
@@ -41,7 +41,10 @@ public static ValueTask<ReadResult> ReadAtLeastAsync(this IPipeReader input, int
             var result = await awaitable;
             while (!result.IsCompleted && result.Buffer.Length < minimumRequiredBytes)
             {
-                cancellationToken.ThrowIfCancellationRequested();
+                if(cancellationToken.IsCancellationRequested)
```
moozzyk: nit: space

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
```diff
@@ -113,16 +113,14 @@ private static void Pinger(object state)
             {
                 // We don't need to wait for this task to complete, we're "tail calling" and
                 // we are in a Timer thread-pool thread.
-#pragma warning disable 4014
-                connection.SendCoreLockAcquiredAsync(
+                _ = connection.SendCoreLockAcquiredAsync(
```
moozzyk: I am **NOT** commenting on this but I think someone will.

anurse: C# 7! Wildcard variable! Yaaaay! It's exactly like the `var ignore = [thing that returns a task]` @davidfowl has used in other places, just without the `var`, thanks to C# 7.

davidfowl: var ignore =

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
```diff
@@ -75,7 +75,8 @@ public WebSocketsTransport(IChannelConnection<Message> application, ILoggerFacto
             {
                 if (receiving.IsCanceled || receiving.IsFaulted)
                 {
-                    // The receiver faulted or cancelled. This means the client is probably broken. Just propagate the exception and exit
+                    // The receiver faulted or cancelled. This means the socket is probably broken. Abort the socket and propagate the exception
+                    socket.Abort();
```
davidfowl: Why are we aborting here?

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
```diff
@@ -249,8 +247,18 @@ private void WriteMaskingKey(Span<byte> buffer)
             _maskingKeyBuffer.CopyTo(buffer);
         }
 
+        /// <summary>
+        /// Terminates the socket abruptly.
+        /// </summary>
+        public void Abort()
+        {
+            _terminateReceiveCts.Cancel();
```
davidfowl: Why do we need to fire both?

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
anurse: Because we had a network error receiving from the socket. If we don't abort, Dispose will fail because of the (possible) active consumer. Although, after some changes I made in `ExecuteAsync`, we may be able to do this in a `finally` block in there rather than forcing the consumer to do it.

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
anurse: `CancelPendingRead` just releases the `ReadAsync` call, we need to cover the cancellation case where we aren't in a `ReadAsync` (L392 below)

anurse:
The state machine for this thing is getting a little crazy. I may try to spend some time tomorrow simplifying it as much as possible. It's making it a bit difficult to follow all the failure cases.

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
davidfowl: Cancellation tokens are so messy. CancelPendingRead should really be enough. Just call CancelPendingRead when the token trips. Then change all of the logic to just look at IsCancelled.

davidfowl: If you aren't currently reading, the next read will be completed synchronously with a cancellation.

test/Microsoft.Extensions.WebSockets.Internal.Tests/WebSocketConnectionExtensions.cs
```diff
@@ -8,11 +8,14 @@ namespace Microsoft.Extensions.WebSockets.Internal.Tests
 {
     public static class WebSocketConnectionExtensions
     {
-        public static async Task<WebSocketConnectionSummary> ExecuteAndCaptureFramesAsync(this IWebSocketConnection connection)
+        public static Task<WebSocketConnectionSummary> ExecuteAndCaptureFramesAsync(this IWebSocketConnection connection)
         {
-            var frames = new List<WebSocketFrame>();
-            var closeResult = await connection.ExecuteAsync(frame => frames.Add(frame.Copy()));
-            return new WebSocketConnectionSummary(frames, closeResult);
+            return Task.Run(async () =>
```
davidfowl: What are you doing Task.Run here? To get off the xunit sync context?

anurse: Partially it was to help debugging. Without the Task.Run, Parallel Stacks wasn't showing anything really helpful and I was debugging issues between two different Tasks. Once I added the Task.Run, it was much easier to see where one Task was blocked when a particular action occurred (and from what I could tell it wasn't because the timing shifted). I expect I'll remove it before merging. The current head of this PR is somewhat in-progress.

src/Microsoft.Extensions.WebSockets.Internal/Microsoft.Extensions.WebSockets.Internal.csproj
```diff
@@ -15,7 +15,8 @@
   <ItemGroup>
     <PackageReference Include="System.IO.Pipelines" Version="$(CoreFxLabsVersion)" />
     <PackageReference Include="System.IO.Pipelines.Text.Primitives" Version="$(CoreFxLabsVersion)" />
-    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="1.2.0-*" PrivateAssets="All" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
```
BrennanConroy: Use the $(CoreFxLabsVersion) variable

anurse: Nope, we want 4.3.0 explicitly here. I originally had it as $(CoreFxVersion) and things broke. I'll double check why and make sure bugs are filed if we still need it

anurse: Never mind, I was confusing CoreFx with CoreFxLabs :)

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
```diff
@@ -249,269 +245,322 @@ private void WriteMaskingKey(Span<byte> buffer)
...
-
-                    payload = buffer.Slice(0, payloadLen);
-
-                    if (masked)
-                    {
-                        // Unmask
-                        MaskingUtilities.ApplyMask(ref payload, maskingKey);
-                    }
+                    var maskingKeyStart = buffer.Start;
+                    maskingKey = buffer.Slice(0, 4).ReadBigEndian<uint>();
```
BrennanConroy: sizeof

BrennanConroy:
:shipit:

PR #219, opened 2017-02-17
Remove makefile.shade and add npm install to repo.targets

natemcmaster:
Also, add (commented-out) conditional cross compiling for tests on xplat


build/repo.targets
```diff
@@ -0,0 +1,10 @@
+<Project>
+  <Target Name="CleanNpm">
```
anurse: Do we have a repo-level clean target? It might be good to have one and wire this in to that. If not, we can add that in later.

natemcmaster: I haven't wired it up yet. Will make a note to do that.

PR #220, opened 2017-02-17
Removing Dispose from Transport, Connection and HubConnection

moozzyk:
Converting StopAsync to DisposeAsync

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -42,8 +42,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
             {
                 throw new ArgumentNullException(nameof(application));
             }
-
+            
             _application = application;
+
             await Connect(url);
             var sendTask = SendMessages(url, _cancellationToken);
             var receiveTask = ReceiveMessages(url, _cancellationToken);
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -38,12 +38,8 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
                 throw new ArgumentNullException(nameof(url));
             }
 
-            if (application == null)
-            {
-                throw new ArgumentNullException(nameof(application));
-            }
+            _application = application ?? throw new ArgumentNullException(nameof(application));
```
mikaelm12: Do we want to move to this pattern for param checking?
It's not consistent with how we do it right above

moozzyk: You can't do this above because url is not being assigned to anything.

davidfowl: We're not using throw expressions. Get rid of the,.

moozzyk: Why? What's wrong with them?

moozzyk:
⌚️

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
BrennanConroy: if (... == null) looks better IMO

BrennanConroy:
Looks good other than the `?? throw`

anurse:
Woo! `DisposeAsync`. Now we just need a syntax feature ;)

PR #221, opened 2017-02-17
Introducing IConnection interface

moozzyk:


src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -46,15 +46,20 @@ public class HubConnection
...
+
+        public HubConnection(IConnection connection, IInvocationAdapter adapter, ILoggerFactory loggerFactory)
         {
             // TODO: loggerFactory shouldn't be required
             if (loggerFactory == null)
             {
                 throw new ArgumentNullException(nameof(loggerFactory));
             }
 
+            _connection = connection ?? throw new ArgumentNullException(nameof(connection));
```
davidfowl: No throw expressions. I hate these things.

test/Microsoft.AspNetCore.Sockets.Client.Tests/HubConnectionTests.cs
```diff
@@ -167,5 +165,62 @@ public void CannotCreateHubConnectionWithNullUrl()
...
+            var mockConnection = new Mock<IConnection>();
+            mockConnection
+                .Setup(m => m.DisposeAsync())
+                .Callback(() => mockConnection.Raise(c => c.Closed += null, (Exception)null))
+                .Returns(Task.FromResult<object>(null));
+
+            var hubConnection = new HubConnection(mockConnection.Object, Mock.Of<IInvocationAdapter>(), new LoggerFactory());
+
+            await hubConnection.StartAsync(Mock.Of<ITransport>());
+            var invokeTask = hubConnection.Invoke("testMethod", typeof(int));
```
BrennanConroy: Is this racey? Invoke might finish before DisposeAsync?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse:
```
if(((_connection = connection) == null) == true) {
    string paramName = default(string);
    paramName = nameof(connection);
    throw new ArgumentNullException(paramName);
}
```

test/Microsoft.AspNetCore.Sockets.Client.Tests/HubConnectionTests.cs
moozzyk: Good catch. Will fix that.

moozzyk: Actually, no. There is no server so no one can complete this invocation/invokeTask.

PR #222, opened 2017-02-18
Prototype: Enabling feedback to SendAsync via asymmetric channels

moozzyk:
Returning errors to the user if send failed.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
@@ -17,12 +17,14 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
         private int _connectionState = ConnectionState.Initial;
-        private IChannelConnection<Message> _transportChannel;
+        // private IChannelConnection<Message> _transportChannel;
         private ITransport _transport;
         private Task _receiveLoopTask;
 
+        private IChannelConnection<Message, SendMessage> _transportChannel;
+
         private ReadableChannel<Message> Input => _transportChannel.Input;
-        private WritableChannel<Message> Output => _transportChannel.Output;
+        private WritableChannel<SendMessage> Output => _transportChannel.Output;
 
         public Uri Url { get; }
 
@@ -120,11 +122,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         private async Task StartTransport(Uri connectUrl)
         {
-            var applicationToTransport = Channel.CreateUnbounded<Message>();
+            var applicationToTransport = Channel.CreateUnbounded<SendMessage>();
             var transportToApplication = Channel.CreateUnbounded<Message>();
-            var applicationSide = new ChannelConnection<Message>(transportToApplication, applicationToTransport);
+            var applicationSide = new ChannelConnection<SendMessage, Message>(applicationToTransport, transportToApplication);
 
-            _transportChannel = new ChannelConnection<Message>(applicationToTransport, transportToApplication);
+            _transportChannel = new ChannelConnection<Message, SendMessage>(transportToApplication, applicationToTransport);
 
             var ignore = Input.Completion.ContinueWith(t =>
             {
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -183,34 +185,37 @@ public Connection(Uri url, ILoggerFactory loggerFactory)
             _logger.LogTrace("Ending receive loop");
         }
 
-        public Task<bool> SendAsync(byte[] data, MessageType type)
+        public Task<Exception> SendAsync(byte[] data, MessageType type)
```
davidfowl: Task already has a way to represent being faulted...

moozzyk: We don't want to return a faulted task because we don't want to throw from SendAsync (if the call is awaited). You can think of this Exception as a more sophisticated success/failure flag which you could rethrow. Or we can stick to bool as we had before because the user can get the exception that caused the failure in the Closed event (which is inevitable if SendAsync fails)

davidfowl: `Task<Exception>` is too gross of an API to expose to humans.

moozzyk: Agreed.

moozzyk:
:up::date:

moozzyk:
⌚️

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -188,7 +190,7 @@ public Task<bool> SendAsync(byte[] data, MessageType type)
             return SendAsync(data, type, CancellationToken.None);
         }
 
-        public async Task<bool> SendAsync(byte[] data, MessageType type, CancellationToken cancellationToken)
+        public async  Task<bool> SendAsync(byte[] data, MessageType type, CancellationToken cancellationToken)
```
mikaelm12: nit: Extra space before Task

src/Microsoft.AspNetCore.Sockets.Client/SendMessage.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
```
mikaelm12: License header

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -456,7 +455,7 @@ public void ConnectionReturnsUrlUsedToStartTheConnection()
 
                     Assert.False(await connection.SendAsync(new byte[] { 1, 1, 3, 5, 8 }, MessageType.Binary));
                 }
-                finally
+  finally
```
mikaelm12: nit: Fix spacing

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -17,12 +17,14 @@ public class Connection: IConnection
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
         private int _connectionState = ConnectionState.Initial;
-        private IChannelConnection<Message> _transportChannel;
+        // private IChannelConnection<Message> _transportChannel;
```
mikaelm12: remove?

src/Microsoft.AspNetCore.Sockets.Common/Internal/ChannelConnection.cs
```diff
@@ -7,21 +7,33 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public static class ChannelConnection
     {
+        public static ChannelConnection<T, U> Create<T, U>(Channel<T> input, Channel<U> output)
```
anurse: `TIn` and `TOut`

src/Microsoft.AspNetCore.Sockets.Common/Internal/ChannelConnection.cs
```diff
@@ -7,21 +7,33 @@ namespace Microsoft.AspNetCore.Sockets.Internal
...
 
-    public class ChannelConnection<T> : IChannelConnection<T>
+    public class ChannelConnection<T> : ChannelConnection<T, T>, IChannelConnection<T>
+    {
+        public ChannelConnection(Channel<T> input, Channel<T> output)
+            : base(input, output)
+        { }
+    }
+
+    public class ChannelConnection<T, U> : IChannelConnection<T, U>
```
anurse: `TIn` and `TOut`

src/Microsoft.AspNetCore.Sockets.Common/Internal/ChannelConnection.cs
```diff
@@ -7,21 +7,33 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public static class ChannelConnection
     {
+        public static ChannelConnection<T, U> Create<T, U>(Channel<T> input, Channel<U> output)
+        {
+            return new ChannelConnection<T, U>(input, output);
+        }
+
         public static ChannelConnection<T> Create<T>(Channel<T> input, Channel<T> output)
```
anurse: You can remove this. If the type is the same, the overload above will work and inference will just fill in both parameters with the same type. That would also mostly eliminate the need for `ChannelConnection<T>`.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -30,8 +30,8 @@ public TestClient(IServiceProvider serviceProvider, string format = "json")
             var transportToApplication = Channel.CreateUnbounded<Message>();
             var applicationToTransport = Channel.CreateUnbounded<Message>();
 
-            Application = ChannelConnection.Create(input: applicationToTransport, output: transportToApplication);
-            var transport = ChannelConnection.Create(input: transportToApplication, output: applicationToTransport);
+            Application = ChannelConnection.Create<Message>(input: applicationToTransport, output: transportToApplication);
```
anurse: Removing `Create<T>` would also solve this conflict.

PR #227, opened 2017-02-21
one last try to get this test de-flaked

anurse:
My hunch is that the lack of a Task.Run means things aren't yielding properly. If this doesn't de-flake things, I'm just gonna skip the dang test :).

/cc @moozzyk @mikaelm12

### DIFF
--- a/test/Microsoft.Extensions.WebSockets.Internal.Tests/WebSocketConnectionTests.PingPong.cs
+++ b/test/Microsoft.Extensions.WebSockets.Internal.Tests/WebSocketConnectionTests.PingPong.cs
@@ -22,11 +22,11 @@ namespace Microsoft.Extensions.WebSockets.Internal.Tests
                 var startTime = DateTime.UtcNow;
                 // Arrange
                 using (var pair = WebSocketPair.Create(
-                    serverOptions: new WebSocketOptions().WithAllFramesPassedThrough().WithPingInterval(TimeSpan.FromMilliseconds(50)),
+                    serverOptions: new WebSocketOptions().WithAllFramesPassedThrough().WithPingInterval(TimeSpan.FromMilliseconds(10)),
                     clientOptions: new WebSocketOptions().WithAllFramesPassedThrough()))
                 {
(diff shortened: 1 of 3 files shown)


test/Microsoft.Extensions.WebSockets.Internal.Tests/WebSocketConnectionTests.PingPong.cs
```diff
@@ -22,11 +22,11 @@ public class PingPongBehavior
                 var startTime = DateTime.UtcNow;
                 // Arrange
                 using (var pair = WebSocketPair.Create(
-                    serverOptions: new WebSocketOptions().WithAllFramesPassedThrough().WithPingInterval(TimeSpan.FromMilliseconds(100)),
+                    serverOptions: new WebSocketOptions().WithAllFramesPassedThrough().WithPingInterval(TimeSpan.FromMilliseconds(50)),
```
mikaelm12: Why the time change?

anurse: Just more ways to try an de-flake the test. More frequent pings :).

moozzyk:
The other one that fails frequently (even though the tests pass):
```
  info: fail: Microsoft.AspNetCore.Sockets.ConnectionManager[0]
  info:       Failed disposing connection 8655f364-c114-416f-8ce0-4de234759fe1
  info: System.ObjectDisposedException: Cannot access a disposed object.
  info: Object name: 'WebSocketConnection'.
``` 

https://travis-ci.org/aspnet/SignalR/jobs/203977072 has both failures.

anurse:
I believe my other PR will fix that (#218)

anurse:
Actually, there is still a possibility for it to occur due to a race between someone closing the socket and a protocol error, so I'm adding some code there to avoid the exception. We don't care if the race occurs so the exception is unnecessary.

BrennanConroy:
🐑 :it:

PR #229, opened 2017-02-22
fix #209 by converting to byte[]

### LINKED ISSUE #209: Change Message to use byte[] for Payload instead of PreservedBuffer
It's just getting too unwieldy. And we rarely actually take it off a Pipeline any more.

anurse:
As per discussion with @davidfowl, we're going to switch to `byte[]` and copy in `Message` for now. It's not ideal, but `PreservedBuffer` is just really not designed for use this way. And `Memory<byte>` doesn't (yet) offer us a better solution since we wouldn't be able to just swap it out for pooled memory anyway since it doesn't really give us a way to track when someone finishes with the memory (so we can return it).

A perk of this is it removes the need for `Message` to be `IDisposable`.

Fixes #209


src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
```diff
@@ -98,41 +98,68 @@ private static bool TryFormatPayload(ReadableBuffer payload, MessageType type, S
...
+                // to use a condition that ensure the only time we stop writing is when we write the slice after the final
+                // newline.
+                var keepWriting = true;
+                while (keepWriting)
                 {
                     // Seek to the end of buffer or newline
-                    var sliced = payload.TrySliceTo(LineFeed, out var slice, out var cursor);
+                    var sliceEnd = payload.IndexOf(LineFeed);
+                    var nextSliceStart = sliceEnd + 1;
+                    if(sliceEnd < 0)
```
BrennanConroy: format

src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
```diff
@@ -98,41 +98,68 @@ private static bool TryFormatPayload(ReadableBuffer payload, MessageType type, S
...
+                // It needs to be written as:
+                //   data: Hello\r\n
+                //   data: \r\n
+                //   \r\n
+                // Since we slice past the newline when we find it, after writing "Hello" in the previous example, we'll
+                // end up with an empty payload buffer, BUT we need to write it as an empty 'data:' line, so we need
+                // to use a condition that ensure the only time we stop writing is when we write the slice after the final
+                // newline.
+                var keepWriting = true;
+                while (keepWriting)
```
BrennanConroy: do while!

PR #231, opened 2017-02-22
use new Base64 codec in Text and SSE Formatters

### LINKED ISSUE #192: Update parsing/formatting code to use corefxlab formatters when available
Our CoreCLR mirroring tool isn't able to mirror the latest stuff from corefxlab until @pranavkm updates our repos to netstandard2.0. Once that's done we'll be able to update our packages and can convert some of the Message parsing/formatting code to use the Base64 and UTF-8 encoder/decoder code in corefx lab. This issue is to track that so we can merge stuff before the netstandard2.0 migration is finished.

anurse:
fix #192 

/cc @KrzysztofCwalina - Btw it's kinda odd that `Base64` doesn't have `TryDecode`/`TryEncode` APIs.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
@@ -76,7 +76,6 @@ namespace Microsoft.AspNetCore.Sockets.Formatters
             var writtenSoFar = 0;
             if (type == MessageType.Binary)
             {
-                // TODO: We're going to need to fix this as part of https://github.com/aspnet/SignalR/issues/192
                 var encodedSize = DataPrefix.Length + Base64.ComputeEncodedLength(payload.Length) + Newline.Length;
                 if (buffer.Length < encodedSize)
                 {
--- a/src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Text;
+using System.Binary;
 
 namespace Microsoft.AspNetCore.Sockets.Formatters
 {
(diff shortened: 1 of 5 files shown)


anurse:
Having a `TryDecode` that could return `false` when there's insufficient space in the buffer would remove the need to do the decoded length math I do which is a little sketchy ;)

src/Microsoft.AspNetCore.Sockets.Common/Formatters/ServerSentEventsMessageFormatter.cs
```diff
@@ -77,8 +77,7 @@ private static bool TryFormatPayload(ReadOnlySpan<byte> payload, MessageType typ
             if (type == MessageType.Binary)
             {
                 // TODO: We're going to need to fix this as part of https://github.com/aspnet/SignalR/issues/192
```
BrennanConroy: Remove comment

src/Microsoft.AspNetCore.Sockets.Common/Formatters/TextMessageFormatter.cs
```diff
@@ -149,39 +148,38 @@ internal static bool TryParseMessage(ReadOnlySpan<byte> buffer, out Message mess
...
-            var payloadBuffer = buffer.Slice(0, length);
+            var payload = buffer.Slice(0, length);
             buffer = buffer.Slice(length);
             consumedSoFar += length;
 
             // Parse the payload. For now, we make it an array and use the old-fashioned types.
             // I've filed https://github.com/aspnet/SignalR/issues/192 to update this. -anurse
-            var payload = payloadBuffer.ToArray();
-            if (messageType == MessageType.Binary)
+            if (messageType == MessageType.Binary && payload.Length > 0)
```
BrennanConroy: Don't think the `payload.Length > 0` check is needed

BrennanConroy: I guess we allow 0 for length, nvm

KrzysztofCwalina:
@anurse, the reason for not having Try APIs is that computing the lengths is O(1) and very deterministic. But if you think Try APIs would work better, I can add/change those. 

Also, I noticed I have ComputeEncodedLength but not ComputeDecodedLength. Do you want me to add it?

anurse:
That makes some sense. I do wonder if we need a clean way for these formatters to express how much additional space is needed. For most of our formatting APIs, after writing the first few bytes we end up knowing the final exact size (because the only variable-length thing is a UTF-8 encoded length prefix), but callers have to just allocate "more" memory and try again.

`ComputeDecodedLength` would definitely be useful and would save some code on our side.

BrennanConroy:
:shipit:

Issue #205, opened 2017-02-15
Wire up text-based transport protocols to Long Polling and SSE


anurse:
Once the transport protocol formatter/parser components are in and tested, we need to wire them up.

After #202, #187, #204 are done, we can do this.

PR #234, opened 2017-02-23
convert the samples to use the official client

anurse:
This will really help with #205 since it means I won't have to add the parsing logic to two places :).


samples/SocketsSample/wwwroot/sockets.html
```diff
@@ -0,0 +1,47 @@
...
+    <ul id="messages"></ul>
+    <script src="lib/signalr-client/signalr-client.js"></script>
+    <script src="utils.js"></script>
+    <script>
+        document.addEventListener('DOMContentLoaded', () => {
+            let transport = getParameterByName('transport') || 'webSockets';
+
+            document.getElementById('transportName').innerHTML = transport;
+
+            var url = `http://${document.location.host}/chat`
```
moozzyk: nit: all these `var`s could be `let`s

moozzyk:
Looks like the ping pong test is still failing https://travis-ci.org/aspnet/SignalR/jobs/204424474

anurse:
I'm going to skip the test.

moozzyk:
Yeah...

samples/SocketsSample/wwwroot/sockets.html
anurse: Yeah, since we're already using other ES6 features, and this is an internal sample, I suppose that's fine :)

anurse:
Maybe just on travis, since I've never really seen the error on TeamCity or AppVeyor

davidfowl:
We should keep a pure web socket page around.

anurse:
Oh yeah, I meant to do that. I'll restore that file.

PR #237, opened 2017-02-25
MSBuildifying gulpified node

moozzyk:
- Creating an MSBuild project for the TS client
- Adding project references to the TS client project from projects that need the client - (ensures the correct targets dependency graph and prevents building the client multiple times and related races)
- Removing gulp tasks from individual projects (allows containing npm only in the TS client source and node tests)
- Using incremental compilation to build the TS client only when inputs change (prevents building the client multiple times or when not needed at all)
- Removing `npm install` from all the projects (takes up to 10 seconds even if there is nothing to restore) - npm packages will still be installed when running full build (if needed) or need to be installed manually

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -1,8 +1,4 @@
 <Project>
-  <Target Name="CleanNpm" AfterTargets="Clean">
-    <RemoveDir Directories="$(RepositoryRoot)node_modules" />
-  </Target>
-
   <Target Name="RestoreNpm" AfterTargets="Restore">
     <Message Text="Restoring NPM modules" Importance="high" />
     <Exec Command="npm install" WorkingDirectory="$(RepositoryRoot)" />
--- a/samples/ChatSample/ChatSample.csproj
+++ b/samples/ChatSample/ChatSample.csproj
@@ -12,6 +12,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR\Microsoft.AspNetCore.SignalR.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Client.TS\Microsoft.AspNetCore.SignalR.Client.TS.csproj" />
   </ItemGroup>
 
   <ItemGroup>
@@ -36,16 +37,13 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
 
-  <Target Name="BuildTSClient" BeforeTargets="BeforeBuild">
+  <Target Name="DotnetBundle" BeforeTargets="BeforeBuild">
     <Exec Command="dotnet bundle" />
-    <PropertyGroup>
-      <SignalRClientDistPath>..\..\dist\browser\signalr-client.js</SignalRClientDistPath>
-      <SignalRClientTargetFolder>$(MSBuildProjectDirectory)/wwwroot/lib/signalr-client/</SignalRClientTargetFolder>
-      <SignalRClientTargetPath>$(MSBuildProjectDirectory)/wwwroot/lib/signalr-client/signalr-client.js</SignalRClientTargetPath>
-    </PropertyGroup>
-    <Copy Condition="Exists('$(SignalRClientDistPath)')" SourceFiles="$(SignalRClientDistPath)" DestinationFolder="$(SignalRClientTargetFolder)"/>
-    <Exec Condition="!Exists('$(SignalRClientDistPath)')" Command="npm install" />
-    <Exec Condition="!Exists('$(SignalRClientDistPath)')" Command="npm run gulp -- --gulpfile $(MSBuildProjectDirectory)/../../src/Microsoft.AspNetCore.SignalR.Client.TS/gulpfile.js bundle-client --bundleOutDir $(SignalRClientTargetFolder)" />
+  </Target>
+
+  <Target Name="CopyTSClient" BeforeTargets="AfterBuild">
+    <Copy SourceFiles="$(MSBuildThisFileDirectory)..\..\dist\browser\signalr-client.js"
+          DestinationFolder="$(MSBuildThisFileDirectory)wwwroot\lib\signalr-client" />
   </Target>
 
   <Target Name="BowerInstall" BeforeTargets="PrepareForPublish">
--- a/src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
+++ b/src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netstandard1.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <SignalRClientDistFolder>$(MSBuildThisFileDirectory)..\..\dist\</SignalRClientDistFolder>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Inputs Include="*.ts;" />
+    <Outputs Include="@(Inputs -> '$(SignalRClientDistFolder)src\%(FileName).d.ts')" />
+    <Outputs Include="@(Inputs -> '$(SignalRClientDistFolder)src\%(FileName).js')" />
+    <Outputs Include="$(SignalRClientDistFolder)browser\signalr-client.js" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="tsconfig.json" />
+  </ItemGroup>
+
+  <!-- this target relies on npm packages being restored manually or when running full build -->
+  <Target Name="BuildTSClient" Inputs="@(Inputs)" Outputs="@(Outputs)" BeforeTargets="Build">
+    <Exec Command="npm run gulp -- --gulpfile $(MSBuildThisFileDirectory)gulpfile.js build-ts-client" />
+  </Target>
+
+  <Target Name="CleanTSClient" AfterTargets="Clean">
+    <RemoveDir Directories="$(SignalRClientDistFolder)" />
+  </Target>
+</Project>
--- a/test/Microsoft.AspNetCore.SignalR.Test.Server/Microsoft.AspNetCore.SignalR.Test.Server.csproj
+++ b/test/Microsoft.AspNetCore.SignalR.Test.Server/Microsoft.AspNetCore.SignalR.Test.Server.csproj
@@ -3,11 +3,12 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR\Microsoft.AspNetCore.SignalR.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Client.TS\Microsoft.AspNetCore.SignalR.Client.TS.csproj" />
     <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Server.IISIntegration" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="$(AspNetCoreVersion)" />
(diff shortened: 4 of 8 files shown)


moozzyk:
VS does not like this change...

moozzyk:
VS should be fixed now.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -0,0 +1,34 @@
...
+  <!-- this target relies on npm packages being restored manually or when running full build -->
+  <Target Name="BuildTSClient" Inputs="@(Inputs)" Outputs="@(Outputs)" BeforeTargets="Build">
+    <Exec Command="npm run gulp -- --gulpfile $(MSBuildThisFileDirectory)gulpfile.js build-ts-client" />
+  </Target>
+
+  <Target Name="Clean">
+    <RemoveDir Directories="$(SignalRClientDistFolder)" />
+  </Target>
+
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\build\repo.targets" />
```
moozzyk: This import is no longer needed.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+  <PropertyGroup>
+    <TargetFramework>netstandard1.0</TargetFramework>
```
moozzyk: We probably need: `<IsPackable>false</IsPackable>`

davidfowl: Why does a typescript project need a target framework?

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
```
davidfowl: Is this even a typescript project?

davidfowl:
Does this still *work* with VS code?

build/repo.targets
```diff
@@ -1,9 +1,5 @@
 <Project>
-  <Target Name="CleanNpm" AfterTargets="Clean">
-    <RemoveDir Directories="$(RepositoryRoot)node_modules" />
-  </Target>
-
-  <Target Name="RestoreNpm" AfterTargets="Restore">
+  <Target Name="RestoreNpm">
```
natemcmaster: Why did you remove this from the lifecycle?

samples/ChatSample/ChatSample.csproj
```diff
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR\Microsoft.AspNetCore.SignalR.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Client.TS\Microsoft.AspNetCore.SignalR.Client.TS.csproj" />
```
natemcmaster: Try adding `ReferenceOutputAssembly="false"` to all ProjectReference to the TS project. This will suppress NuGet attempting to use the project reference as a compile reference.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
natemcmaster: Yes, remove this.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
natemcmaster: This probably don't need to reference Microsoft.NET.Sdk.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -0,0 +1,34 @@
...
+  <PropertyGroup>
+    <TargetFramework>netstandard1.0</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <SignalRClientDistFolder>$(MSBuildThisFileDirectory)..\..\dist\</SignalRClientDistFolder>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Inputs Include="*.ts;" />
```
natemcmaster: Not sure how well VS does with this project. You might try using `None` instead so VS recognizes the files.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
```
natemcmaster: Nit: suggest changing extension to `.proj`. This isn't a C# project.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
moozzyk: I needed to pretend it's a csproj otherwise project references did not work.

moozzyk: And it could not be open in VS

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
moozzyk: Similar to above - it makes project references work.

samples/ChatSample/ChatSample.csproj
moozzyk: OK.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
moozzyk: I started with .proj but VS refused to add it to the solution. Hence changing to project to pretend it's a .csproj

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
moozzyk: VS showed the files just fine.

moozzyk: I was also able to build the ts client.

moozzyk:
@davidfowl - VS Code should work. I plugged in things to MSBuild but under the cover it sill invokes npm/gulp to build things. This has not changed.

natemcmaster:
My only hesitation to this is that we're abusing what ProjectReference and Microsoft.NET.Sdk are meant to do, and this may be fragile. There are other ways to enforce build dependencies that may be cleaner, such as adding a `BeforeBuild` target to invoke MSBuild on the TS project, and using [Project Dependencies](https://msdn.microsoft.com/en-us/library/et61xzb3.aspx) in the solution to ensure build order.

moozzyk:
@natemcmaster - If this needs to be a .csproj because VS otherwise won't open it why using `BeforeBuild` (which I believe adds magic and requires creating additional targets in each project to have it under control) is better than a project dependency?
A bit more context: TypeScript tools are not available for VS2017 and in VS2015 a typescript project is... a .csproj (to be fair, it creates a host which is the cs part as far as I understand but apparently there isn't really a TypeScript project in the ecosystem AFAICT)

build/repo.targets
moozzyk: It's a mistake. I originally depended on this target directly but I decided not to do this anymore because it adds at least a few secs to the build even if there is nothing to install. Will revert this change.

samples/ChatSample/ChatSample.csproj
moozzyk: Turned out that I can't do that. With this settings I can build from command line but when I build with VS it fails to build the project due to: 
Cannot find project info for 'C:\Source\SignalR-Core\src\Microsoft.AspNetCore.SignalR.Client.TS\Microsoft.AspNetCore.SignalR.Client.TS.csproj'. This can indicate a missing project reference.

moozzyk:
:up::date:

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -0,0 +1,32 @@
...
+  <ItemGroup>
+    <None Include="tsconfig.json" />
+  </ItemGroup>
+
+  <!-- this target relies on npm packages being restored manually or when running full build -->
+  <Target Name="BuildTSClient" Inputs="@(Inputs)" Outputs="@(Outputs)" BeforeTargets="Build">
+    <Exec Command="npm run gulp -- --gulpfile $(MSBuildThisFileDirectory)gulpfile.js build-ts-client" />
+  </Target>
+
+  <Target Name="Clean">
```
natemcmaster: This target will be overridden by Microsoft.NET.Sdk because this project has implicit top/bottom imports. If you want to override this target, use explicit imports instead. Example: 
```xml
<Project>
  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />

   <!-- regular project stuff here -->

  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />

  <!-- Override default SDK targets-->
  <Target Name="Clean" />
  <Target Name="Build" />

</Project>
```

test/Microsoft.AspNetCore.SignalR.Test.Server/Microsoft.AspNetCore.SignalR.Test.Server.csproj
```diff
@@ -3,38 +3,31 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1</TargetFrameworks>
+    <PreserveCompilationContext>true</PreserveCompilationContext>
+    <OutputType>Exe</OutputType>
```
natemcmaster: PreserveCompilationContext and OutputType are not required. These are the default settings for `Microsoft.NET.Sdk.Web` projects.

test/Microsoft.AspNetCore.SignalR.Test.Server/Microsoft.AspNetCore.SignalR.Test.Server.csproj
```diff
@@ -3,38 +3,31 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1</TargetFrameworks>
```
natemcmaster: Nit: make this singular `TargetFramework`

test/Microsoft.AspNetCore.SignalR.Test.Server/Microsoft.AspNetCore.SignalR.Test.Server.csproj
moozzyk: OK

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
moozzyk: Will do.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
moozzyk: To clarify - currently there is [no Typescript support/tooling for VS2017](https://blogs.msdn.microsoft.com/typescript/2017/02/22/announcing-typescript-2-2/). I wanted to see what they did in previous versions and if you create a "Typescript project" in VS2015 it will be a *.csproj. I tried using their MSBuild tasks from VS2015 but they are compiled against .NET Framework and "our" MSBuild could not load them. When they have tools for VS2017 and netcore MSBuild tasks we need to make sure that there is an easy way to use the latest version of the compiler (i.e. you don't have to reinstall typescript MSBuild/tools if you want to move to a new compiler version)

natemcmaster: cref https://github.com/Microsoft/TypeScript/issues/11

anurse:
Given that there aren't TypeScript tools for 2017 yet, I'm OK with using a `csproj` here. Let's be prepared to move over to a `tsproj`-based system when it releases though.

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
natemcmaster: See also https://github.com/dotnet/cli/issues/4785#issuecomment-272315965 cc @mlorbetske

src/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
moozzyk: Adding this results in the following build warnings: 
warning MSB4011: "C:\Users\XXXX\AppData\Local\Microsoft\dotnet\sdk\1.0.0-rc4-004913\Sdks\Microsoft.NET.Sdk\Sdk\Sdk.targets" cannot be imported again. It was already imported at "C:\source\SignalR-Core\src\Microsoft.AspNetCore.SignalR.Client.TS\Microsoft.AspNetCore.SignalR.Client.TS.csproj (30,3)". This is most likely a build authoring error. This subsequent import will be ignored.

natemcmaster: Does your file say `<Project>` or `<Project Sdk="microsoft.net.sdk">`?

moozzyk: Without Sdk attribute on Project it does work indeed. I decided however to use AfterTargets. The equivalence between the Sdk attribute and Imports is magic and we cannot override the Build target anyways.

Issue #235, opened 2017-02-23
Clean up HubConnection API


mikaelm12:
Right now logger factories are required to create a HubConnection. This shouldn't be the case

mikaelm12:
Added more ctors in https://github.com/aspnet/SignalR/pull/238

PR #238, opened 2017-02-25
Remove logger factory requirement

mikaelm12:
Cleaning things up a little bit. We shouldn't require logger factories to do things.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
@@ -44,6 +44,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             add { _connection.Closed += value; }
             remove { _connection.Closed -= value; }
         }
+
         public HubConnection(Uri url, IInvocationAdapter adapter)
             : this(new Connection(url), adapter, null)
         { }
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
@@ -44,30 +44,29 @@ namespace Microsoft.AspNetCore.SignalR.Client
             add { _connection.Closed += value; }
             remove { _connection.Closed -= value; }
         }
+        public HubConnection(Uri url, IInvocationAdapter adapter)
+            : this(new Connection(url), adapter, null)
+        { }
 
         public HubConnection(Uri url, IInvocationAdapter adapter, ILoggerFactory loggerFactory)
             : this(new Connection(url, loggerFactory), adapter, loggerFactory)
         { }
 
+        public HubConnection(IConnection connection, IInvocationAdapter adapter)
+            : this(connection, adapter, null)
+        { }
+
         public HubConnection(IConnection connection, IInvocationAdapter adapter, ILoggerFactory loggerFactory)
         {
-            // TODO: loggerFactory shouldn't be required
-            if (loggerFactory == null)
-            {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Sockets.Client/NullLoggerFactory.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
-    internal class NullLoggerFactory : ILoggerFactory
```
moozzyk: let's put it into the Internal namespace (and subfolder)

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -45,29 +46,29 @@ public class HubConnection
             remove { _connection.Closed -= value; }
         }
 
+        public HubConnection(Uri url, IInvocationAdapter adapter)
```
davidfowl: How did we end up in overload hell so fast. Can we re-evaluate these constructors? Some are useful for testing, and some are useful for humans. What we have right now is not very well thought out.

moozzyk: There will be one more - there should be a default `IInvocationAdapter` (presumably JSon?) - the user should not have to provide one. `IConnection` adds to the crud but this one is useful for testing.
On the other - once we have the one with the default invocation adapter - we can remove this one? In that case you would have to provide loggerfactory if you provide invocation adapter but it can be null

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -45,29 +46,29 @@ public class HubConnection
...
+            : this(new Connection(url), adapter, null)
+        { }
+
         public HubConnection(Uri url, IInvocationAdapter adapter, ILoggerFactory loggerFactory)
             : this(new Connection(url, loggerFactory), adapter, loggerFactory)
         { }
 
+        public HubConnection(IConnection connection, IInvocationAdapter adapter)
```
moozzyk: We can remove this ctor. `IConnection` is mostly for testing (or some advanced scenarios not many people will probably try)

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -28,10 +29,14 @@ public class LongPollingTransport : ITransport
 
         public Task Running { get; private set; }
 
+        public LongPollingTransport(HttpClient httpClient)
```
moozzyk: not sure if we need these overloads for transports. Connection will always pass a factory and the users are not really expected to create these.

mikaelm12: Its nice for testing though so we don't have to create a logger factory each time we want to use Long Polling. I'll remove it though

mikaelm12:
Is this good to merge?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
davidfowl: We need to fix it. This isn't maintainable and these scenarios don't all have equal importance.

moozzyk: It was being tracked separately and yesterday I created individual issue for this - #235. We could fix it as part of this or separately.

PR #240, opened 2017-02-26
Add EndPoint Options

BrennanConroy:
Not a huge fan of `IOptions<EndPointOptions<HubEndPoint<THub, TClient>>> endPointOptions`
Suggestions welcome!

@moozzyk @davidfowl @anurse @mikaelm12 

(Also, currently `AddEndPoint` takes an `EndPoint` like `HubEndPoint<Chat>` but resolves `EndPointOptions<HubEndPoint<Chat, IClientProxy>>` in the constructor which wont contain the values you set.)

### DIFF
--- a/samples/SocketsSample/Startup.cs
+++ b/samples/SocketsSample/Startup.cs
@@ -29,7 +29,7 @@ namespace SocketsSample
                     });
             // .AddRedis();
 
-            services.AddSingleton<MessagesEndPoint>();
+            services.AddEndPoint<MessagesEndPoint>();
             services.AddSingleton<ProtobufSerializer>();
         }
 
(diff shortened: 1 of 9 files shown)


BrennanConroy:
Ping
Could someone take a look, need this for the Auth PR

BrennanConroy:
And an issue for changing Policy to name

samples/SocketsSample/Startup.cs
```diff
@@ -29,7 +29,7 @@ public void ConfigureServices(IServiceCollection services)
                     });
             // .AddRedis();
 
-            services.AddSingleton<MessagesEndPoint>();
+            services.AddEndPoint<MessagesEndPoint>();
```
moozzyk: Should we change also our test servers to use `.AddEndPoint`?

PR #241, opened 2017-02-26
Adding Auth to endpoints

BrennanConroy:
This PR is mainly to see what the current Auth progress is and if it is heading in the right direction and also to get input on design. Built on top of #240. Try to ignore most of the weird stuff in the samples.

@moozzyk @davidfowl @anurse @mikaelm12


src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Authorization;
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    public class EndPointOptions<TEndPoint> where TEndPoint : EndPoint
+    {
+        public AuthorizationPolicy Policy { get; set; }
```
HaoK: You probably want to take the name, as that would enable scenarios where the policy is resolved dynamically by name which would be blocked if you took the policy instance always.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -48,6 +58,59 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
             }
         }
 
+        private async Task<bool> AuthorizeAsync<TEndPoint>(HttpContext context) where TEndPoint : EndPoint
```
HaoK: Is this pretty much an exact copy of the AuthorizeFilter logic?  As apart of Auth 2.0 I'll try to consolidate this logic somewhere that MVC + SignalR can consume rather than duplicate...

BrennanConroy: Yeah its from AuthorizeFilter, a common spot for this would be great!

davidfowl: /cc @javiercn

davidfowl:
@BrennanConroy What's the deal with this PR?

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
javiercn: Agree. We should have live somewhere where both MVC and Signal R can consume

HaoK: So currently AuthZ doesn't depend on HttpAbstractions, assuming we want to keep that clean, we will need to add a new package for this helper in Security... maybe Microsoft.AspNetCore.Authorization.Http?

HaoK: Or is this more appropriate as a shared common code i.e. AuthorizationHelper similar to SecurityHelper (or we could just add this as a new method in SecurityHelper too)...

javiercn: I would make a package like Microsoft.AspNetCore.Authentication.Policy and I would offer a service
IPolicyAuthenticationProvider {
    Task<ClaimsPrincipal> AuthenticateAsync(HttpContext context, string policyName);
}

It's implementation will do what the AuthorizeFilter does before setting the principal into the context. This has two benefits:
* You can reuse this logic across implementations (MVC, SignalR, any other middleware).
* You can get access to a principal that matches exactly what that policy defines at any point where you have access to the HttpContext.

HaoK: Okay I'll start a prototype/design PR for the rough shape of something this week

HaoK: Something like https://github.com/aspnet/Security/pull/1148 look reasonable?  It basically splits things into the authentication method, and authorization method (eventually this might have logic to distinguish between challenge/forbidden)

davidfowl:
@BrennanConroy rebsae

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -51,6 +60,58 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
```
davidfowl: Move this code to a helper

BrennanConroy:
:up: 📅

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -53,6 +58,14 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
             }
         }
 
+        private Task<bool> AuthorizeAsync<TEndPoint>(HttpContext context) where TEndPoint : EndPoint
+        {
+            // TODO: Authorize attribute on EndPoint
+            var options = context.RequestServices.GetRequiredService<IOptions<EndPointOptions<TEndPoint>>>();
```
davidfowl: Why are you resolving options twice?

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -35,9 +35,12 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
 
         public async Task ExecuteAsync<TEndPoint>(string path, HttpContext context) where TEndPoint : EndPoint
         {
-            // Get the end point mapped to this http connection
-            var endpoint = (EndPoint)context.RequestServices.GetRequiredService<TEndPoint>();
             var options = context.RequestServices.GetRequiredService<IOptions<EndPointOptions<TEndPoint>>>().Value;
+            // TODO: Authorize attribute on EndPoint
```
anurse: Is there a bug for this?

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Sockets.Internal
+{
+    public static class AuthorizeHelper
+    {
+        public static async Task<bool> AuthorizeAsync(HttpContext context, AuthorizationPolicy policy)
```
anurse: Given that we complete synchronously if `policy` is `null`, perhaps this is a place for `ValueTask<bool>`? C# 7 lets you drop it in place and continue using `async`. It's not a super high-performance code path, but it seems useful for the relatively common case of not having an auth policy.

@davidfowl thoughts on more `ValueTask`s? Premature optimization? Other unknown costs?

test/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/lib/jasmine/boot.js
```diff
@@ -0,0 +1,152 @@
+/*
```
anurse: Is this supposed to be checked in? If not, can we gitignore it?

davidfowl: It totally confused me.

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
davidfowl: Do it in a follow up commit.

anurse: If there's going to be a new commit to remove the jasmine crap, may as well throw it in here, it's a small change (unless ValueTask isn't available in our current package set, then I agree, just get it in).

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
BrennanConroy: https://github.com/aspnet/SignalR/issues/256

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
anurse: Never mind, `Task<bool>` is cached: https://referencesource.microsoft.com/#mscorlib/system/runtime/compilerservices/AsyncMethodBuilder.cs,a431f9d7dd423518

Issue #244, opened 2017-02-27
NullReferenceException when disposing connection


moozzyk:
Not sure what the conditions are but I saw this in my console:
```
Now listening on: http://localhost:5000
Application started. Press Ctrl+C to shut down.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:5000/echo/negotiate
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 138.4393ms 200
fail: Microsoft.AspNetCore.Sockets.ConnectionManager[0]
      Failed disposing connection 150496f4-7486-4f7d-91b9-1322f927e24d
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Sockets.Internal.ConnectionState.<DisposeAsync>d__35.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Sockets.ConnectionManager.<DisposeAndRemoveAsync>d__14.MoveNext()
```

BrennanConroy:
Oh yeah, I saw it too when debugging in VS https://github.com/aspnet/SignalR/pull/241/files#diff-56f70a1ad7b0f822fd88bf4ff8fee123R54

davidfowl:
Hmm

davidfowl:
@BrennanConroy can you look into this?

davidfowl:
Seems like it's possible for DisposeAsync to be called without having ever set up the application or transport.

davidfowl:
Yes, that must be it.

BrennanConroy:
Fixed in https://github.com/aspnet/SignalR/pull/252

PR #252, opened 2017-02-28
Fix null ref in DisposeAsync on ConnectionState

BrennanConroy:
#244 

@moozzyk @davidfowl 

Alternatively, we could check `if (Status == Active)` and branch the code

### DIFF
--- a/test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
+++ b/test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Sockets.Internal;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


davidfowl:
Where's the test?

BrennanConroy:
Right... thought about it earlier then forgot

test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
```diff
@@ -101,6 +101,31 @@ public void RemoveConnection()
             await state.DisposeAsync();
         }
 
+        [Fact]
+        public async Task DisposeAndRemoveAsyncWithInactiveConnection()
```
moozzyk: Why don't you just create a connection and call DisposeAsync without setting anything?

BrennanConroy: Because that makes too much sense! :/

test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
```diff
@@ -101,6 +100,22 @@ public void RemoveConnection()
             await state.DisposeAsync();
         }
 
+        [Fact]
+        public async Task DisposeInactiveConnection()
+        {
+            var connectionManager = CreateConnectionManager();
+            var state = connectionManager.CreateConnection();
+
+            var transport = state.Connection.Transport;
```
moozzyk: nit: inline like for the other Asserts?

PR #257, opened 2017-03-01
Integrating new transport protocols

anurse:
Adds support for the Text message protocol and Server-Sent Event formatter to the Server, C# Client and JavaScript Client.

Note: JavaScript does not support Binary **Messages** yet, I'll file an issue for that. It requires custom Base64 and ArrayBuffer support (see TODOs).

Neither the Server nor the Clients support the Binary Message **Encoding** either, but that can be done without separately and doesn't require synchronous Client/Server changes, since the Server can add support but the Client can avoid requesting it until it supports it.


samples/ClientSample/RawSample.cs
```diff
@@ -9,13 +9,23 @@
 using Microsoft.AspNetCore.Sockets.Client;
 using Microsoft.Extensions.Logging;
 using Microsoft.AspNetCore.Sockets;
+using System.Linq;
```
mikaelm12: Sort

samples/SocketsSample/wwwroot/index.html
```diff
@@ -9,7 +9,7 @@
     <ul>
         <li><a href="sockets.html?transport=serverSentEvents">Server Sent Events</a></li>
         <li><a href="sockets.html?transport=longPolling">Long polling</a></li>
-        <li><a href="sockets.html?transport=webSocket">Web Sockets</a></li>
+        <li><a href="sockets.html?transport=webSockets">Web Sockets</a></li>
```
mikaelm12: Should this list be sorted?

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -114,6 +118,27 @@ public Task StartAsync(Uri url, IChannelConnection<Message> application)
             }
         }
 
+        private IEnumerable<Message> ReadMessages(ReadOnlySpan<byte> payload)
+        {
+            if(payload.Length == 0)
```
mikaelm12: nit: Space here

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -114,6 +118,27 @@ public Task StartAsync(Uri url, IChannelConnection<Message> application)
...
+            {
+                yield break;
+            }
+
+            var messageFormat = MessageFormatter.GetFormat(payload[0]);
+            payload = payload.Slice(1);
+
+            while(payload.Length > 0)
```
mikaelm12: nit: Space

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -114,6 +118,27 @@ public Task StartAsync(Uri url, IChannelConnection<Message> application)
...
+            {
+                if(!MessageFormatter.TryParseMessage(payload, messageFormat, out var message, out var consumed))
```
mikaelm12: nit: Space

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -114,6 +118,27 @@ public Task StartAsync(Uri url, IChannelConnection<Message> application)
...
+                {
+                    throw new InvalidDataException("Invalid message payload from server");
+                }
+                payload = payload.Slice(consumed);
```
mikaelm12: nit: New line

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
```diff
@@ -7,14 +7,20 @@ namespace Microsoft.AspNetCore.Sockets.Formatters
...
+        public static readonly byte BinaryFormatIndicator = (byte)'B';
+
+        public static readonly string TextContentType = "application/vnd.microsoft.aspnet.endpoint-messages.v1+text";
+        public static readonly string BinaryContentType = "application/vnd.microsoft.aspnet.endpoint-messages.v1+binary";
+
         public static bool TryFormatMessage(Message message, Span<byte> buffer, MessageFormat format, out int bytesWritten)
         {
             if (!message.EndOfMessage)
             {
                 // This is a truely exceptional condition since we EXPECT callers to have already
```
mikaelm12: nit: Should be "truly" 😄

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -34,13 +38,44 @@ public LongPollingTransport(ReadableChannel<Message> application, ILoggerFactory
...
+                while (_application.TryRead(out var message))
                 {
+                    var buffer = alloc.Memory.Span;
+
                     _logger.LogDebug("Writing {0} byte message to response", message.Payload.Length);
-                    context.Response.ContentLength = message.Payload.Length;
-                    await context.Response.Body.WriteAsync(message.Payload, 0, message.Payload.Length);
+
+                    // Try to format the message
+                    if(!MessageFormatter.TryFormatMessage(message, buffer, messageFormat, out var written))
```
mikaelm12: nit: space

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -34,13 +38,44 @@ public LongPollingTransport(ReadableChannel<Message> application, ILoggerFactory
...
+                    {
+                        // We need to expand the buffer
+                        // REVIEW: I'm not sure I fully understand the "right" pattern here...
+                        alloc.Ensure(MaxBufferSize);
+                        buffer = alloc.Memory.Span;
+
+                        // Try one more time
+                        if(!MessageFormatter.TryFormatMessage(message, buffer, messageFormat, out written))
```
mikaelm12: nit: space

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
```diff
@@ -27,43 +29,43 @@ public ServerSentEventsTransport(ReadableChannel<Message> application, ILoggerFa
...
                     {
-                        await Send(context, message);
+                        if (!ServerSentEventsMessageFormatter.TryFormatMessage(message, buffer.Memory.Span, out var written))
+                        {
+                            // We need to expand the buffer
+                            // REVIEW: I'm not sure I fully understand the "right" pattern here...
+                            buffer.Ensure(LongPollingTransport.MaxBufferSize);
+
+                            // Try one more time
+                            if (!ServerSentEventsMessageFormatter.TryFormatMessage(message, buffer.Memory.Span, out written))
```
mikaelm12: nit: out var

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
```diff
@@ -7,14 +7,20 @@ namespace Microsoft.AspNetCore.Sockets.Formatters
 {
     public static class MessageFormatter
     {
+        public static readonly byte TextFormatIndicator = (byte)'T';
+        public static readonly byte BinaryFormatIndicator = (byte)'B';
+
+        public static readonly string TextContentType = "application/vnd.microsoft.aspnet.endpoint-messages.v1+text";
```
BrennanConroy: aspnetcore!

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
```diff
@@ -27,43 +29,43 @@ public ServerSentEventsTransport(ReadableChannel<Message> application, ILoggerFa
...
                 {
+                    var buffer = pipe.Alloc();
                     while (_application.TryRead(out var message))
                     {
-                        await Send(context, message);
+                        if (!ServerSentEventsMessageFormatter.TryFormatMessage(message, buffer.Memory.Span, out var written))
+                        {
+                            // We need to expand the buffer
+                            // REVIEW: I'm not sure I fully understand the "right" pattern here...
+                            buffer.Ensure(LongPollingTransport.MaxBufferSize);
```
BrennanConroy: Ewwwww

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
```diff
@@ -27,43 +29,43 @@ public ServerSentEventsTransport(ReadableChannel<Message> application, ILoggerFa
             context.Response.ContentType = "text/event-stream";
             context.Response.Headers["Cache-Control"] = "no-cache";
             context.Response.Headers["Content-Encoding"] = "identity";
-            await context.Response.Body.FlushAsync();
+
+            var pipe = context.Response.Body.AsPipelineWriter();
+
+            await pipe.Alloc().FlushAsync();
```
davidfowl: wtf is this?

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
anurse: I have recieved your feedbak and whil applie it approrpriately

mikaelm12: What a troll lol

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
anurse: We're using an existing variable (`written`) that was declared above so we don't need `out var`.

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
anurse: Without it, we don't write the headers until the first message arrives.

anurse: Just like we used to do with the stream. I suppose I could do it with the Stream before adapting

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
anurse: Yeah, I agree. I need to work with @davidfowl to figure out the best way to handle this. The APIs for formatting messages don't return any kind of hint as to how much more data is needed.

samples/ClientSample/RawSample.cs
```diff
@@ -2,20 +2,30 @@
...
     internal class RawSample
     {
         public static async Task MainAsync(string[] args)
         {
+            if(args.Any(a => a == "--debug"))
+            {
+                Console.WriteLine($"Ready for debugger to attach. Process ID: {Process.GetCurrentProcess().Id}");
+                Console.Write("Press ENTER to Continue");
+                Console.ReadLine();
+                args = args.Where(a => a != "--debug").ToArray();
```
moozzyk: nit: could use Except

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -114,6 +117,28 @@ public Task StartAsync(Uri url, IChannelConnection<Message> application)
...
+        private IEnumerable<Message> ReadMessages(ReadOnlySpan<byte> payload)
+        {
+            if (payload.Length == 0)
+            {
+                yield break;
+            }
+
+            var messageFormat = MessageFormatter.GetFormat(payload[0]);
+            payload = payload.Slice(1);
+
```
moozzyk: Shouldn't we throw here if payload is 0? 0 would mean that we received only the messageFromat but not other data...

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
```diff
@@ -7,14 +7,20 @@ namespace Microsoft.AspNetCore.Sockets.Formatters
...
         public static bool TryFormatMessage(Message message, Span<byte> buffer, MessageFormat format, out int bytesWritten)
         {
             if (!message.EndOfMessage)
             {
-                // This is a truely exceptional condition since we EXPECT callers to have already
+                // This is truly an exceptional condition since we EXPECT callers to have already
                 // buffered incomplete messages and synthesized the correct, complete message before
                 // giving it to us. Hence we throw, instead of returning false.
-                throw new InvalidOperationException("Cannot format message where endOfMessage is false using this format");
+                throw new ArgumentException("Cannot format message where endOfMessage is false using this format", nameof(message));
```
moozzyk: I liked `InvalidOperation` a bit better but meh...

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
```diff
@@ -28,5 +34,44 @@ public static bool TryParseMessage(ReadOnlySpan<byte> buffer, MessageFormat form
...
+        }
+
+        public static MessageFormat GetFormat(byte formatIndicator)
+        {
+            // Can't use switch because our "constants" are not consts, they're "static readonly" (which is good, because they are public)
+            if (formatIndicator == TextFormatIndicator)
+            {
+                return MessageFormat.Text;
+            }
+            else if (formatIndicator == BinaryFormatIndicator)
```
moozzyk: nit: drop `else`

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
```diff
@@ -28,5 +34,44 @@ public static bool TryParseMessage(ReadOnlySpan<byte> buffer, MessageFormat form
...
+            {
+                return MessageFormat.Binary;
+            }
+            throw new ArgumentException($"Invalid message format: 0x{formatIndicator:X}", nameof(formatIndicator));
+        }
+
+        public static bool TryParseMessage(ReadOnlySpan<byte> payload, object messageFormat)
```
moozzyk: Where is this going to be used from?

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
```diff
@@ -34,13 +38,44 @@ public LongPollingTransport(ReadableChannel<Message> application, ILoggerFactory
...
+                        // We need to expand the buffer
+                        // REVIEW: I'm not sure I fully understand the "right" pattern here...
+                        alloc.Ensure(MaxBufferSize);
+                        buffer = alloc.Memory.Span;
+
+                        // Try one more time
+                        if (!MessageFormatter.TryFormatMessage(message, buffer, messageFormat, out written))
+                        {
+                            // Message too large
+                            throw new InvalidOperationException($"Message is too large to write. Maximum allowed message size is: {MaxBufferSize}");
```
moozzyk: 4KB is too much? Because people complain about 64KB limit on webSocket messages we have in the old SignalR. AFAIR we did not have message limits for other transports.

moozzyk: Also, it this buffer **after** serializing? Because the serialized messages can be bigger than original messages in which case the user can hit a case where a message they have is actually less than the size returned in the exception message.

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
anurse: I think I'm going to add back in the "MeasureMessage" call I had earlier which allowed us to determine exactly how much space we need. We can calculate it fully in advance so it should be able to improve things here.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
anurse: We could throw, but it's technically valid since the format doesn't actually require that a payload contain any frames. If we get a single `T` as a response, this code just treats that as a payload with no frames in it.

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
anurse: I changed it because the exception is due to an invalid argument, so it felt right.

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
anurse: I guess... I kinda prefer it but I don't really care :)

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
anurse: C# Client

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
anurse: The big problem is that we're only guaranteed to up to 4KB in a single chunk from the Pipeline. Anything bigger will be a chain of non-contiguous chunks. So I'm going to have to rework some of the formatting logic to solve this.

moozzyk: OK. Is this going to be a separate change?

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -114,6 +118,27 @@ public Task StartAsync(Uri url, IChannelConnection<Message> application)
             }
         }
 
+        private IEnumerable<Message> ReadMessages(ReadOnlySpan<byte> payload)
+        {
+            if (payload.Length == 0)
```
anurse: I don't think so, but I can check.

src/Microsoft.AspNetCore.Sockets/Transports/LongPollingTransport.cs
anurse: I'm not sure yet, but I think it will be separate.

anurse:
:up: :date: - This is now ready for prime-time :). JS client code is in, .NET client code is in, tests are passing (on my box, we'll see how the CIs go)

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
anurse: It looks like when poll times out, we get an `OperationCancelledException` and the whole transport shuts down...

anurse: But that's a separate unrelated bug

anurse: Filed https://github.com/aspnet/SignalR/issues/265

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+function splitAt(input: string, searchString: string, position: number): [string, number] {
+    let index = input.indexOf(searchString, position);
+    if (index < 0) {
+        return [input.substr(position), input.length];
+    }
+    let left = input.substring(position, index);
+    return [left, index + searchString.length];
+}
+
+export namespace SSE {
```
moozzyk: Why using internal modules instead of classes?

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+    export function parseMessage(input: string): [string, Message] {
```
moozzyk: I am not a fan of this return type. When I looked at the signature I did not know the string was actually an Error. I would prefer to use exceptions for errors.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+            // characters are found.
+            throw "TODO: Support for binary messages";
+        }
+
+        // Create the message
+        return [null, new Message(type, payload)];
+    }
+}
+
+export namespace Text {
```
moozzyk: same as above - I think we should use classes instead of internal modules.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+    const InvalidPayloadError = new Error("Invalid text message payload");
+    const LengthRegex = /^[0-9]+$/;
+
+    function hasSpace(input: string, offset: number, length: number): boolean {
+        let requiredLength = offset + length;
+        return input.length >= requiredLength;
+    }
+
+    function parseMessage(input: string, position: number, log?: (message: string) => void): [string, number, Message] {
```
moozzyk: `log` does not seem to be used

Logging is not fleshed out yet - I was thinking about creating ILogger interface and allowing people to pass it. We could have a default implementation that uses `console.log`

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+        let [header, offset] = splitAt(input, "\n", 0);
+        let payload = input.substring(offset);
+
+        // Just in case the header used CRLF as the line separator, carve it off
+        if (header.endsWith('\r')) {
+            header = header.substr(0, header.length - 1);
+        }
+
+        // Parse the header
+        var type = knownTypes[header];
```
moozzyk: nit: consider calling the variable `messageType`

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+        if (type === undefined) {
+            return ["Unknown type value: '" + header + "'", null];
+        }
+
+        if (type == MessageType.Binary) {
+            // We need to decode and put in an ArrayBuffer. Throw for now
+            // This will require our own Base64-decoder because the browser
+            // built-in one only decodes to strings and throws if invalid UTF-8
+            // characters are found.
+            throw "TODO: Support for binary messages";
```
moozzyk: because we are returning errors everywhere I don't think this exception is handled in any way

src/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -160,7 +173,21 @@ export class LongPollingTransport implements ITransport {
         pollXhr.onload = () => {
             if (pollXhr.status == 200) {
                 if (this.onDataReceived) {
-                    this.onDataReceived(pollXhr.response);
+                    // Parse the messages
+                    let [error, messages] = Formatters.parseMessages(pollXhr.response);
+                    if (error) {
+                        if (this.onError) {
+                            // TODO: Is this how we should emit this?
```
moozzyk: This is OK for now. There is a [_unify errors_](https://github.com/aspnet/SignalR/issues/3) item on my list which is about making sure that we actually use Error wherever possible/practical.

src/Microsoft.AspNetCore.Sockets.Common/Formatters/MessageFormatter.cs
```diff
@@ -7,14 +7,20 @@ namespace Microsoft.AspNetCore.Sockets.Formatters
 {
     public static class MessageFormatter
     {
+        public static readonly byte TextFormatIndicator = (byte)'T';
+        public static readonly byte BinaryFormatIndicator = (byte)'B';
+
+        public static readonly string TextContentType = "application/vnd.microsoft.aspnetcore.endpoint-messages.v1+text";
+        public static readonly string BinaryContentType = "application/vnd.microsoft.aspnetcore.endpoint-messages.v1+binary";
```
BrennanConroy: Update the spec too

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export function parseMessages(input: string): [string, Message[]] {
+    if (input.length == 0) {
+        return [null, []];
+    }
+
+    if (input[0] == 'T') {
+        return Text.parseMessages(input.substr(1));
+    }
+    else {
```
moozzyk: nit: `else` not needed

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+            if (error) {
+                // Failed to parse message. Report the error and the messages we parsed so far
+                return [error, messages];
+            }
+            messages.push(message);
+        }
+        return [null, messages];
+    }
+
+    export function formatMessages(input: Message[]): string {
```
moozzyk: remove?

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+        return [null, offset, new Message(type, payload)];
+    }
+
+    export function parseMessages(input: string): [string, Message[]] {
+        if (input.length == 0) {
+            return [null, []];
+        }
+
+        let messages = [];
+        var offset = 0;
```
moozzyk: consider using `let` everywhere

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+        while (offset < input.length) {
+            var [error, offset, message] = parseMessage(input, offset);
```
moozzyk: this should be `let`. Also not a huge fan of this tuple/destrurcturing but it is internal so probably is fine.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+        // Required space is: 3 (type flag, ":", ";") + length (payload len)
+        if (!hasSpace(input, offset, 3 + length)) {
+            return ["Message is incomplete", position, null];
+        }
+
+        // Read the type
+        var [typeStr, offset] = splitAt(input, ":", offset);
+
+        // Parse the type
+        var type = knownTypes[typeStr];
```
moozzyk: nit: rename to messageType. `type === undefined` is a bit misleading

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
```diff
@@ -27,43 +29,43 @@ public ServerSentEventsTransport(ReadableChannel<Message> application, ILoggerFa
...
+
             try
             {
                 while (await _application.WaitToReadAsync(token))
                 {
+                    var buffer = pipe.Alloc();
                     while (_application.TryRead(out var message))
                     {
-                        await Send(context, message);
+                        if (!ServerSentEventsMessageFormatter.TryFormatMessage(message, buffer.Memory.Span, out var written))
```
moozzyk: I wonder how often we will confuse these formatters with IStreamFormatters.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return [null, messages];
+    }
+
+    export function formatMessages(input: Message[]): string {
+        throw "Not implemented";
+    }
+}
+
+export function parseMessages(input: string): [string, Message[]] {
```
moozzyk: Why `parseMessage` is inside `Text` while this is outside?

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: Also this name should be more specific

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: Rename to something more specific. Hard to assume what SSE is without looking at the code.

src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
anurse: These are basically internal (and could even be moved into an `Internal` namespace. In fact, I think I will do that). Also, they could end up unifying with `IStreamFormatters` eventually since they are attempting to achieve the same goal, writing an object to a stream.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: This one is designed to handle the Text/Binary format indicator. But since in JavaScript the data types will be different (string vs ArrayBuffer), I'm thinking I'll remove this and merge it in to `Text`

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: sure, but we will need it later :). I can remove it for now though, since nothing uses it and we can just re-add it.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: They aren't "classes" though, the closest match in C# is a `static class`, since you never create an instance, they're just groupings of functions. In TypeScript, the idiomatic way to do that is to put functions in a namespace. I could use a `class` and just only put `static` functions on it, but that seems like trying to hard to make TypeScript look like C#... I'm open to changing if you'd prefer a class though.

I will definitely rename them though.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: Ok, I can do that. I'm not a fan of using exceptions for parsing errors since they aren't necessarily exceptional cases and it's cleaner to do control flow with return values. I can change these around though.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: Yep, this is vestigial from my first pass at some of this, I'll remove it.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: I believe I can't use `let` for this one because of the way I reassign it, but I'll look at it again.

anurse: So I can do it, but it means messing with the return values from the `splitAt` helper. The problem is this pair of lines:

```typescript
let offset = 0;
let [someStr, offset] = splitAt(input, ":", offset);
```

We're trying to introduce a new variable to hold the string before the split and reassign the offset that's tracking our current position in the string (which we are using to avoid allocating too many strings).

Now, we can do this:

```typescript
let offset = 0;

let someStr;
[someStr, offset] = splitAt(input, ":", offset);
```

That looks slightly better than many of the alternatives. Or we could return just the offset, instead of using a tuple, but then we basically have to inline the entire helper and go back to using `indexOf` directly.

Finally, we could introduce an interface rather than using a tuple and do something like:

```typescript
export interface SplitResult {
    prefix: string,
    offset: number
};

let offset = 0;
let result = splitAt(input, 0, offset);
offset = result.offset;
```

Using `var` removes any of these concerns since `var` allows you to redeclare the same variable (which just overwrites it)

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: Without `out` parameters, it's basically this or Objects/Interfaces.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: Removed when the previous comment was addressed

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -0,0 +1,146 @@
...
+        if (input.length == 0) {
+            return [null, []];
+        }
+
+        let messages = [];
+        var offset = 0;
+        while (offset < input.length) {
+            var [error, offset, message] = parseMessage(input, offset);
+            if (error) {
+                // Failed to parse message. Report the error and the messages we parsed so far
```
anurse: Well, that's just because I forgot to do it in the Transport code. If we want to support giving the messages we were able to read to the user, even in case of error, we can do that. It'd be best to avoid exceptions in that case though.

anurse:
Open PR feedback remaining:

* Partial results when an error occurs? https://github.com/aspnet/SignalR/pull/257#discussion_r104051426
* Tuples? https://github.com/aspnet/SignalR/pull/257#discussion_r104058032
* Namespaces of functions vs classes? https://github.com/aspnet/SignalR/pull/257#discussion_r104052307
* `let` vs `var` when reassigning a variable: https://github.com/aspnet/SignalR/pull/257#discussion_r104056789

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: I was not aware of that - if using `let` over `var` requires adding code then we should not do this just for the sake of using `let`

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: Let's leave as it is then.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: I was wondering why not to make these parsefunctions members of the actual transports. At least for ServerSentEvents it  is already coupled to the transport so I am not sure what we gain by keeping it separate. 

I also thought about making them actual classes. In case of long polling you could keep state inside the parser itself instead of outside. You also wouldn't need to return tuples. On the flip side the majority of poll payloads would still contain just one message and we would create a new parser to parse each of these.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: Given that these parsing errors are reported to the connection and result in closing the connection I would say they are not benign.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: I think we should give the user what we were able to read.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: > In case of long polling you could keep state inside the parser itself instead of outside.

There isn't any parser state held outside the parser itself. The only exported function `parseMessages` expects a complete input and returns a complete output. **Internal** functions use tuples to avoid having to slice strings and create a bunch of new string objects, but otherwise, all the state is self-contained in the single public API.

anurse: Anyway, I'm not too worried if you want to refactor this once it gets in, I do want to try and merge this PR fairly soon though so it doesn't go stale.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
anurse: In that case, we can't really use exceptions for error reporting then because they interrupt the control flow. When I had a version that returned a tuple, we could return the errors as well as the individual messages.

Or we could refactor around a different kind of API that parses messages incrementally, but then we have a lot more state to track internally about where we are in the input stream.

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: Sounds good!

src/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: This is what I was thinking about - parsing incrementally. Still, this should be separate - it's better to merge this and then iterate.

anurse:
Sounds like feedback is mostly addressed and the builds are green. Someone going to approve this or are there more comments? :)

moozzyk:
:ship::it:

davidfowl:
👏

PR #258, opened 2017-03-01
Enabling feedback to SendAsync via asymmetric channels

moozzyk:


src/Microsoft.AspNetCore.Sockets.Common/Internal/ChannelConnection.cs
```diff
@@ -7,21 +7,33 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public static class ChannelConnection
     {
+        public static ChannelConnection<TIn, TOut> Create<TIn, TOut>(Channel<TIn> input, Channel<TOut> output)
+        {
+            return new ChannelConnection<TIn, TOut>(input, output);
+        }
+
         public static ChannelConnection<T> Create<T>(Channel<T> input, Channel<T> output)
```
moozzyk: I tried removing this overload but it did not really work because then I could not assign it to `IChannelConnection<T>` without cast.

src/Microsoft.AspNetCore.Sockets.Common/IChannelConnection.cs
```diff
@@ -8,9 +8,13 @@ namespace Microsoft.AspNetCore.Sockets
...
     // REVIEW: These should probably move to Channels. Why not use IChannel? Because I think it's better to be clear that this is providing
     // access to two separate channels, the read end for one and the write end for the other.
-    public interface IChannelConnection<T> : IDisposable
+    public interface IChannelConnection<T> : IChannelConnection<T, T>
+    {
+    }
+
+    public interface IChannelConnection<T, U> : IDisposable
     {
         ReadableChannel<T> Input { get; }
```
mikaelm12: Maybe change these to TIn and TOut like you have elsewhere

moozzyk:
Rebased and :up::date:

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -194,20 +193,21 @@ public Task<bool> SendAsync(byte[] data, MessageType type)
...
-            var message = new Message(data, type);
+            var sendTcs = new TaskCompletionSource<object>();
+            var message = new SendMessage(data, type, sendTcs);
 
             while (await Output.WaitToWriteAsync(cancellationToken))
             {
                 if (Output.TryWrite(message))
                 {
-                    return true;
+                    await sendTcs.Task;
```
halter73: Is it OK for the code following the call Connection.SendAsync() to run inline with Transport.SendMessages? I'm worried about callers blocking.

src/Microsoft.AspNetCore.Sockets.Client/SendMessage.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public struct SendMessage
```
halter73: Nit: Not sure I love the naming of SendMessage and Message. I would suggest renaming Message to ReceiveMessage, but I get that's potentially little confusing from the perspective of the server. Unfortunately this makes the client code a little less clear. Especially considering the client transports receive SendMessages.

src/Microsoft.AspNetCore.Sockets.Client/SendMessage.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public MessageType Type { get; }
+        public byte[] Payload { get; }
+        public TaskCompletionSource<object> SendResult { get; }
```
halter73: Just some questions. So this is just a Message but with the EOM bool replaced with a SendResult TCS? Is the TCS just used to report transport errors to the Connection.SendAsync() caller? Do Channels not have a built in mechanism for this like Pipelines?

moozzyk: We use the channel to send the user request from the application to the transport. Data sent from the transport to the connection are just messages received by the transport from the server. The connection has a loop running on a separate thread that removes these messages from the channel and raise the Received event. The issue was that before this change `SendAsync` would complete once the message was queued to the channel and we were not able to tell the user if sending the message actually succeeded. They would get the exception in the `Closed` event but in general it was impossible to correlate this exception with the actual `SendAsync` call that failed. Now the `SendMessage` contains the tcs which is used by the transport let the user know if sending the message succeeded. 

(EndOfMessage is not relevant here - it is just an artifact of sharing code with the server - this is for partial websocket messages which we don't allow on the client. The WebSocketsTransport on the client will buffer until the entire message is received and only then will push the message to the channel,)

src/Microsoft.AspNetCore.Sockets.Client/SendMessage.cs
moozzyk: I couldn't come up with a better name for the `SendMessage` - I would be more than happy to change it to something better. 

The Message is called just `Message` because it is shared with the server where the channels are symmetric (i.e. objects of the same type are sent both directions).

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: The purpose of this change is to be able to tell the user whether we were able to to send their message. Therefore we need to wait until the message is sent before we can return. I guess they don't have to await the call if they are not interested in the result they would just not await the call.

davidfowl: We need to dispatch the completion so don't dead lock anywhere.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -132,16 +134,19 @@ public Task StartAsync(Uri url, IChannelConnection<Message> application)
 
                         var response = await _httpClient.SendAsync(request);
                         response.EnsureSuccessStatusCode();
+                        sendTcs.SetResult(null);
```
davidfowl: Dispatch the TCS completion. What version of .NET Standard are we using? We may be able to use the new TCS enum.

moozzyk: It is netstandard1.3

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: @davidfowl updated to not `await` on the sendTcs.Task in our code.

moozzyk: Changed how we create tcs to prevent from running user's code on our thread.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -194,20 +193,21 @@ public Task<bool> SendAsync(byte[] data, MessageType type)
 
             if (_connectionState != ConnectionState.Connected)
             {
-                return false;
+                throw new InvalidOperationException(
+                    $"Cannot send messages when the connection is not in the Connected state.");
             }
 
-            var message = new Message(data, type);
+            var sendTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
```
moozzyk: Add comment

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -194,20 +193,21 @@ public Task<bool> SendAsync(byte[] data, MessageType type)
 
             if (_connectionState != ConnectionState.Connected)
             {
-                return false;
+                throw new InvalidOperationException(
+                    $"Cannot send messages when the connection is not in the Connected state.");
```
BrennanConroy: Remove `$`

PR #259, opened 2017-03-01
Switch dotnet-bundle to BuildBundlerMinifier

natemcmaster:
Most recent update of BuildBundlerMinifier chains the bundle step in automatically via MSBuild.

cref https://github.com/madskristensen/BundlerMinifier/issues/230

### DIFF
--- a/samples/ChatSample/ChatSample.csproj
+++ b/samples/ChatSample/ChatSample.csproj
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="BuildBundlerMinifier" Version="2.4.337" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.Cookies" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore" Version="$(AspNetCoreVersion)" />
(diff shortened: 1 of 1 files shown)


samples/ChatSample/ChatSample.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="BuildBundlerMinifier" Version="2.4.337" />
```
moozzyk: How it is being invoked?

samples/ChatSample/ChatSample.csproj
```diff
@@ -53,7 +53,6 @@
   </Target>
 
   <ItemGroup>
-    <DotNetCliToolReference Include="BundlerMinifier.Core" Version="2.2.301" />
     <DotNetCliToolReference Include="Microsoft.EntityFrameworkCore.Tools.DotNet" Version="$(AspNetCoreVersion)" />
     <DotNetCliToolReference Include="Microsoft.Extensions.SecretManager.Tools" Version="1.0.1-*" />
```
moozzyk: Is there a $(VersionVariable) we could use here?

samples/ChatSample/ChatSample.csproj
natemcmaster: This package carries an [MSBuild target](https://github.com/madskristensen/BundlerMinifier/blob/master/src/BundlerMinifier/build/BuildBundlerMinifier.targets) that chains in automatically to "Build" and "Clean".

samples/ChatSample/ChatSample.csproj
natemcmaster: We could make one but we should bump these tools to 1.2.0-* like everything else.

moozzyk: We should but this is not this change.

samples/ChatSample/ChatSample.csproj
moozzyk: I see. Thanks!

moozzyk:
:ship::it:

Issue #248, opened 2017-02-28
Fix a possible race when StartAsync is not awaited and StopAsync/Dispose is called


moozzyk:
Previously tracked as #206

PR #260, opened 2017-03-02
Fixing a race DisposeAsync is called when StartAsync hasn't finished

moozzyk:
Fixes: #248

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO.Pipelines;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
@@ -17,10 +16,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
     {
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
-        private int _connectionState = ConnectionState.Initial;
-        private IChannelConnection<Message> _transportChannel;
-        private ITransport _transport;
-        private Task _receiveLoopTask;
+        private volatile int _connectionState = ConnectionState.Initial;
+        private volatile IChannelConnection<Message> _transportChannel;
+        private volatile ITransport _transport;
+        private volatile Task _receiveLoopTask;
+        private volatile Task _startTask = Task.CompletedTask;
 
         private ReadableChannel<Message> Input => _transportChannel.Input;
         private WritableChannel<Message> Output => _transportChannel.Output;
@@ -47,11 +47,14 @@ namespace Microsoft.AspNetCore.Sockets.Client
         public Task StartAsync(HttpClient httpClient) => StartAsync(transport: null, httpClient: httpClient);
         public Task StartAsync(ITransport transport) => StartAsync(transport: transport, httpClient: null);
 
-        // TODO HIGH: Fix a race when the connection is being stopped/disposed when start has not finished running
-        public async Task StartAsync(ITransport transport, HttpClient httpClient)
+        public Task StartAsync(ITransport transport, HttpClient httpClient)
         {
-            _transport = transport ?? new WebSocketsTransport(_loggerFactory);
+            _startTask = StartAsyncInternal(transport, httpClient);
+            return _startTask;
+        }
 
+        private async Task StartAsyncInternal(ITransport transport, HttpClient httpClient)
+        {
             if (Interlocked.CompareExchange(ref _connectionState, ConnectionState.Connecting, ConnectionState.Initial)
                 != ConnectionState.Initial)
             {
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
```diff
@@ -106,6 +106,37 @@ public void ConnectionReturnsUrlUsedToStartTheConnection()
         }
 
         [Fact]
+        public async Task CanStopStartingConnection()
```
mikaelm12: Looks like you're not actually asserting anything in this test

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -21,6 +20,7 @@ public class Connection: IConnection
         private IChannelConnection<Message> _transportChannel;
         private ITransport _transport;
         private Task _receiveLoopTask;
+        private Task _startTask = Task.FromResult<object>(null);
```
davidfowl: `Task.CompletedTask`

test/Microsoft.AspNetCore.Sockets.Client.Tests/ConnectionTests.cs
moozzyk: This is a race so is hard to test. The main idea is that StartAsync is not called on the transport (awaiting the start task would fail since I am using a mock transport). I changed the test a little bit and now I do verify that the `StartAsync` completes before `DisposeAsync` and verify explicitly that `StartAsync` on the transport is not called.

moozzyk:
:up::date: - I also fixed an issue where in WebSocketsTransport and LongPollingTransport where stopping a transport that has not been started resulted in a first chance exception. It wouldn't bubble up because we have an umbrella catch but still it's better to not have it.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -51,7 +51,7 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
 
             // TODO: Handle TCP connection errors
             // https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Core/Owin/WebSockets/WebSocketHandler.cs#L248-L251
-            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
+            Running =Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
```
BrennanConroy: Revert

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -21,6 +20,7 @@ public class Connection: IConnection
         private IChannelConnection<Message> _transportChannel;
         private ITransport _transport;
         private Task _receiveLoopTask;
+        private Task _startTask = Task.CompletedTask;
```
halter73: Should _startTask and _connectionState be volatile? I would expect in the time it takes an HTTP request to complete in GetConnectUrl, the disposing thread will see the Task, but better to be safe.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -47,20 +47,32 @@ public Connection(Uri url, ILoggerFactory loggerFactory)
...
 
+        private async Task StartAsyncInternal(ITransport transport, HttpClient httpClient)
+        {
             if (Interlocked.CompareExchange(ref _connectionState, ConnectionState.Connecting, ConnectionState.Initial)
                 != ConnectionState.Initial)
             {
                 throw new InvalidOperationException("Cannot start a connection that is not in the Initial state.");
             }
 
+            _transport = transport ?? new WebSocketsTransport(_loggerFactory);
```
halter73: Any reason this isn't done immediately prior to the call to StartTransport?

moozzyk: No, you are right - it should be done as close to starting the transport as possible. I will fix this.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: That's a very good point. I was thinking about `_startTask` a bit but not about `_connectionState`.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -69,16 +81,31 @@ public Connection(Uri url, ILoggerFactory loggerFactory)
...
             }
 
-            // start receive loop
-            _receiveLoopTask = ReceiveAsync();
+            // if the connection is not in the Connecting state here it means the user called DisposeAsync
+            if (Interlocked.CompareExchange(ref _connectionState, ConnectionState.Connected, ConnectionState.Connecting)
+                == ConnectionState.Connecting)
+            {
+                // Do not "simplify" - events can be removed from a different thread
+                var connectedEventHandler = Connected;
```
davidfowl: What thread are you on here?

davidfowl: You might want to dispatch here too

moozzyk: OK. I did not want this initially but now I think it does make more sense.

PR #267, opened 2017-03-04
Dispatching raising Received and Closed events to separate threads

moozzyk:


src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -142,12 +144,15 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
             {
                 Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected);
 
-                // Do not "simplify" - events can be removed from a different thread
-                var closedEventHandler = Closed;
-                if (closedEventHandler != null)
-                {
-                    closedEventHandler(t.IsFaulted ? t.Exception.InnerException : null);
-                }
+                _eventQueue = _eventQueue.ContinueWith(_ => Task.Run(() => {
```
moozzyk: we should protect enqueue with a lock

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -92,6 +90,10 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
             {
                 connectedEventHandler();
             }
+
+            // start receive loop only after the Connected event was raised to
+            // avoid Received event being raised before the Connected event
+            _receiveLoopTask = ReceiveAsync();
```
mikaelm12: 👍

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -21,6 +21,7 @@ public class Connection: IConnection
         private ITransport _transport;
         private Task _receiveLoopTask;
         private volatile Task _startTask = Task.CompletedTask;
+        private volatile Task _eventQueue = Task.CompletedTask;
```
davidfowl: You should just bring back the `TaskQueue`, it's well tested

moozzyk: I can try but it had some things I was not sure about (like TaskMonitor to warn about possible deadlocks). It would also have to be rewritten/cleaned up a bit because it has references to performance counters and TaskAsyncHelper. If the tests can be ported easily then maybe it's fine.

davidfowl: https://github.com/aspnet/SignalR/blob/40ecc9df4de4bfc7394b99401e3bcb406f9debc7/src/Microsoft.AspNetCore.Sockets/TaskQueue.cs

moozzyk: I thought we eventually did not move it. Since we did, I will change the code to use it.

moozzyk:
:up::date: - resurrected TaskQueue and using it now

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -142,12 +144,17 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
             {
                 Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected);
 
-                // Do not "simplify" - events can be removed from a different thread
-                var closedEventHandler = Closed;
-                if (closedEventHandler != null)
-                {
-                    closedEventHandler(t.IsFaulted ? t.Exception.InnerException : null);
-                }
+                var ignore2 = _eventQueue.Enqueue(queue => {
```
moozzyk: 🥇

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -142,12 +144,17 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
             {
                 Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected);
 
-                // Do not "simplify" - events can be removed from a different thread
-                var closedEventHandler = Closed;
-                if (closedEventHandler != null)
-                {
-                    closedEventHandler(t.IsFaulted ? t.Exception.InnerException : null);
-                }
+                _eventQueue.Enqueue(queue => {
```
BrennanConroy: Does this one need `var ignore =` also?

test/Microsoft.AspNetCore.Sockets.Client.Tests/TaskQueueTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            var queue = new TaskQueue();
+            var ignore = queue.Enqueue(() => Task.FromException(exception));
+            var task = queue.Enqueue(() => Task.CompletedTask);
+
+            var actual = await Assert.ThrowsAsync<Exception>(async () => await task);
+            Assert.Same(exception, actual);
+        }
+
+        [Fact]
+        public void TaskQueueRunsTasksInSequence()
```
BrennanConroy: Not a big fan of this test, doesn't really seem to correctly check that tasks run in sequence.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: This is the one that had `ignore2` but since we are in a non-async function (lambda) here the compiler does not complain about not assigning the returned task so I did remove it.

test/Microsoft.AspNetCore.Sockets.Client.Tests/TaskQueueTests.cs
moozzyk: I copied these tests from the old repo but I see what you mean. I could assign a value instead of `n++`.

BrennanConroy:
Looks good to me

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -138,16 +140,21 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
 
             _transportChannel = new ChannelConnection<Message>(applicationToTransport, transportToApplication);
 
-            var ignore = Input.Completion.ContinueWith(t =>
+            var ignore = Input.Completion.ContinueWith(async t =>
```
davidfowl: Why do the hook channel completion? We should change this in another PR.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -166,18 +173,36 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
         {
             try
             {
-                _logger.LogTrace("Beginning receive loop");
+                _logger.LogTrace("Beginning receive loop.");
 
                 while (await Input.WaitToReadAsync())
                 {
+                    if (_connectionState != ConnectionState.Connected)
```
davidfowl: How correct does this need to be? Doesn't it need to be synchronized so that we never call a user callback if they are disconnected?

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: Because the channel can be closed from the other side and this is the only way to know it.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: This prevents from trying to invoke receive events after the user called DisposeAsync. We also need to drain the channel otherwise Input.Completion task will never be completed. 

As for synchronization - we modify the state using InterlockedExchange and _connectionState is volatile. We change the state as the first thing if the user calls DisposeAsync or if the channel is closed by the transport.

davidfowl: I'm pretty sure you can still receive data after the state has shifted into disconnected. Even though the connection state is volatile it doesn't mean that right after this check the state can't shift into disconnected.

moozzyk: As per this check if the data is received after the connection moved to the disconnected state we will reject it - this can only happen when the user is stopping the connection and they basically said they don't want to receive messages. If state changed after this check you received the data before you moved to the disconnected state. Firing the event seems to be fine in that case.
From the functional perspective the behavior I wanted was - receive event should not be scheduled after close event and receive should not be scheduled after DisposeAsync exits even if the invocation is not awaited.

davidfowl: Yes, I just want to call out that's this "stops" it from happening. In quotes because it can still happen. We didn't fix the race, just reduced the annoyance of it.

moozzyk: Let's talk about this tomorrow.

moozzyk:
02df601715f345c0596d2f74e4fb893dfe8827a5

Issue #266, opened 2017-03-03
WebSocketTransport does not create correct url to connect to the server


moozzyk:
The url here: https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs#L117  should be just `/echo`

With the Url above the client does not get the connection id. If the WebSockets transport was replaced with LongPolling transport things wouldn't work.

mikaelm12:
Resolved in: https://github.com/aspnet/SignalR/pull/268

PR #268, opened 2017-03-06
Websockets transport now creates correct url

mikaelm12:
Addresses: https://github.com/aspnet/SignalR/issues/266

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
@@ -37,7 +37,7 @@ namespace Microsoft.AspNetCore.SignalR.Tests
 
         [ConditionalFact]
         [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win2008R2, SkipReason = "No WebSockets Client for this platform")]
-        public async Task BareWebSocketsTest()
+        public async Task WebSocketsTest()
         {
             const string message = "Hello, World!";
             using (var ws = new ClientWebSocket())
--- a/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
@@ -37,7 +37,7 @@ namespace Microsoft.AspNetCore.SignalR.Tests
 
         [ConditionalFact]
         [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win2008R2, SkipReason = "No WebSockets Client for this platform")]
-        public async Task WebSocketsTest()
+        public async Task BareWebSocketsTest()
         {
             const string message = "Hello, World!";
             using (var ws = new ClientWebSocket())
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -37,7 +37,7 @@ public EndToEndTests(ServerFixture serverFixture)
 
         [ConditionalFact]
         [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win2008R2, SkipReason = "No WebSockets Client for this platform")]
-        public async Task WebSocketsTest()
+        public async Task BareWebSocketsTest()
```
moozzyk: Not sure why the test name was changed...

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -100,21 +110,22 @@ public static IEnumerable<object[]> MessageSizesData
         {
             get
             {
-                yield return new object[] { new string('A', 5 * 1024)};
-                yield return new object[] { new string('A', 5 * 1024 * 1024 + 32)};
+                yield return new object[] { new string('A', 5 * 1024), new WebSocketsTransport()};
+                yield return new object[] { new string('A', 5 * 1024 * 1024 + 32), new WebSocketsTransport() };
+                yield return new object[] { new string('A', 5 * 1024), new LongPollingTransport(new HttpClient())};
```
moozzyk: This does not seem right (given the name of the property). If needed at all (and I am not convinced it is) you should create a getter that creates a Cartesian product of what before was `MessageSizesData` and `Transports`

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -100,21 +110,22 @@ public static IEnumerable<object[]> MessageSizesData
...
+                yield return new object[] { new string('A', 5 * 1024 * 1024 + 32), new LongPollingTransport(new HttpClient())};
             }
         }
 
         [ConditionalTheory]
         [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win2008R2, SkipReason = "No WebSockets Client for this platform")]
         [MemberData(nameof(MessageSizesData))]
-        public async Task ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport(string message)
+        public async Task ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport(string message, ITransport transport)
```
moozzyk: This test (as indicated by its name) is used to test the logic we have in the WebSockets transport to combine messages which don't fit the buffer/have endOfMessage flag set to false. As such this test is specific to WebSocketsTransport and doesn't have to be run against other transports. We already have a test above that test that sending and receiving data using different transports works.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
mikaelm12: Just to be a little more explicit that this test is using bare websockets and not testing the transports. I can change this back

moozzyk: Yeah, change it back. Other than that it now looks good.

mikaelm12:
Hmm. Travis Failure https://travis-ci.org/aspnet/SignalR/jobs/208377625#L2744
Any help on this @anurse

anurse:
\*sigh\* it's something timing out on Travis yet again. I can't figure out why these tests are so flaky only on Travis. It's got to be something about slow machines and task scheduling there, but it's almost impossible to reproduce on any of my normal machines. For now, restarting the build (which I just did) normally works.

PR #270, opened 2017-03-07
Removing [Fin]

moozzyk:
[Fin] was removed in favor of the `Close` message type.


moozzyk:
Just noticed I forgot to create a PR for this...

PR #274, opened 2017-03-08
Moving TS client to a dedicated subfolder

moozzyk:


mikaelm12:
Sorry but whats the purpose behind this again?

moozzyk:
To keep the project root clean and have one folder for all TS client related changes.

mikaelm12:
So this will be the strategy for all the additional clients moving forward then as well I'm assuming?

mikaelm12:
Anyways the changes look pretty good to me but I'll let someone more confident in their csproj abilities sign off 😄

BrennanConroy:
:shipit:

moozzyk:
@mikaelm12 - we decided that only TS client will be in this repo but we don't want npm to pollute the tree - especially the root.

PR #279, opened 2017-03-09
Increasing the timeout of host start

moozzyk:
Hoping this will prevent random failures like build SignalR build 1155


davidfowl:
Flaky tests....

moozzyk:
Yeah...
Also increasing log level to debug to see traces like this: https://github.com/aspnet/Hosting/blob/d5ec0859e5496f83ca32bf8bfb68ce3c3efe832f/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs#L50

PR #291, opened 2017-03-11
Removing our own copy of NullLoggerFactory

moozzyk:
Moving TaskQueue to the Internal folder as it is in the Internal namespace


src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -30,7 +30,7 @@ public class LongPollingTransport : ITransport
 
         public Task Running { get; private set; } = Task.CompletedTask;
 
-        public LongPollingTransport(HttpClient httpClient)
```
mikaelm12: It looks like I didn't have my trailing space flagger extension enabled lol

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -28,15 +29,19 @@ public class EndToEndTestsCollection : ICollectionFixture<ServerFixture>
     [Collection(EndToEndTestsCollection.Name)]
     public class EndToEndTests
     {
+        private readonly ITestOutputHelper _output;
+
         private readonly ServerFixture _serverFixture;
 
-        public EndToEndTests(ServerFixture serverFixture)
+        public EndToEndTests(ServerFixture serverFixture, ITestOutputHelper output)
```
mikaelm12: What exactly in this change for?

mikaelm12: Capturing output so we can log inside these tests?

moozzyk: Yes. @BrennanConroy looked at the server logs for [the flaky test failure](https://travis-ci.org/aspnet/SignalR/jobs/209849075) and recommended turning on logs on the client.

src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
```diff
@@ -115,37 +128,46 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
...
         {
+            _logger.LogInformation("Starting the send loop");
+
             while (await _application.Input.WaitToReadAsync(cancellationToken))
             {
                 while (_application.Input.TryRead(out SendMessage message))
                 {
                     try
                     {
+                        _logger.LogDebug("Receive message from application. Message type {0}. Payload size: {1}",
```
BrennanConroy: Received

src/Microsoft.AspNetCore.Sockets.Client/Internal/NullLoggerFactory.cs
```diff
@@ -1,30 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
-
-namespace Microsoft.AspNetCore.Sockets.Client.Internal
-{
-    public class NullLoggerFactory : ILoggerFactory
```
ajaybhargavb: 👍

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -63,17 +64,22 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
...
                 // Connection is being stopped while start was in progress
                 if (_connectionState == ConnectionState.Disconnected)
                 {
+                    _logger.LogDebug("Connection was closed from a different thread.");
                     return;
                 }
 
                 _transport = transport ?? new WebSocketsTransport(_loggerFactory);
+
+                _logger.LogDebug("Staring transport '{0}' with Url: {1}", transport.GetType().Name, connectUrl);
```
mikaelm12: Starting*

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -249,6 +271,8 @@ public Task SendAsync(byte[] data, MessageType type)
             var sendTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
             var message = new SendMessage(data, type, sendTcs);
 
+            _logger.LogDebug("Sending message");
```
mikaelm12: Should this log message go inside the while loop?

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -102,8 +110,20 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
...
                     Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected);
 
+                    // There is an inherent race between receive and close. Removing the last message from the channel
+                    // makes Input.Completion task completed and runs this continuation. We need to await _receiveLoopTask
+                    // to make sure that the message removed from the channel is processed before we drain the queue.
+                    // There is a short window between we start the channel and assign the _receiveLoopTask a value.
+                    // To make sure that _receiveLoopTask can be awaited (i.e. is not null) we need to await _startTask.
+                    _logger.LogDebug("Ensuring all outstanding messages are processed.");
+                    await _startTask;
+                    await _receiveLoopTask;
```
mikaelm12: 👍

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: I don't think so. We have a loop but it is always sending just one message. The only case when you might have more than one iteration will be when you try to send more than one message at the same time and this is OK (and the reason to have the loop)

Issue #37, opened 2016-11-17
Add ability to configure active transports for specific endpoints on the server


davidfowl:
We should allow configuring end points for specific transports

PR #294, opened 2017-03-12
Supported transports

davidfowl:
- Allow turning transports on or off per end point with a flags enum
- It's http specific (maybe it should be called http transport type?)

If we like the direction I'll work on finishing this up.

/cc @anurse @BrennanConroy 

#37


src/Microsoft.AspNetCore.Sockets/TransportType.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    [Flags]
+    public enum TransportType
+    {
+        /// <summary>
+        /// Every transport
+        /// </summary>
+        All = Streaming | LongPolling,
```
mikaelm12: Should the groupings come after the individual transports?

src/Microsoft.AspNetCore.Sockets/TransportType.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.Sockets/TransportType.cs
davidfowl: Yea, this was something that I copied from old new signalr. Wanted to see what people think of the entire approach (enum vs strings etc).

BrennanConroy:
👍

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -345,11 +349,22 @@ private Task ProcessNegotiate(HttpContext context)
...
-        private async Task<bool> EnsureConnectionStateAsync(ConnectionState connectionState, HttpContext context, string transportName)
+        private async Task<bool> EnsureConnectionStateAsync(ConnectionState connectionState, HttpContext context, TransportType transportType, TransportType supportedTransports)
         {
+            var transportName = transportType.ToString().ToLower();
+
+            if ((supportedTransports & transportType) == 0)
+            {
+                context.Response.StatusCode = StatusCodes.Status404NotFound;
+                // REVIEW: Is this information leakage? Should we only show this in "development"?
+                await context.Response.WriteAsync($"{transportName} transport not supported by this end point type");
```
anurse: Pretty sure it is a leak. Maybe just log it?

moozzyk: When we have negotiate that returns supported transports - would it be different?

anurse: Fair point. I suppose it's not really a leak then.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -345,11 +349,22 @@ private Task ProcessNegotiate(HttpContext context)
             }
         }
 
-        private async Task<bool> EnsureConnectionStateAsync(ConnectionState connectionState, HttpContext context, string transportName)
+        private async Task<bool> EnsureConnectionStateAsync(ConnectionState connectionState, HttpContext context, TransportType transportType, TransportType supportedTransports)
         {
+            var transportName = transportType.ToString().ToLower();
```
BrennanConroy: This should go inside the if statement

davidfowl: It's used outside of the if

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -155,7 +159,7 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
                     {
                         _logger.LogDebug("Establishing new connection: {connectionId} on {requestId}", state.Connection.ConnectionId, state.RequestId);
 
-                        state.Connection.Metadata["transport"] = LongPollingTransport.Name;
+                        state.Connection.Metadata["transport"] = "longpolling";
```
anurse: Y U No Constant? Should we use the enum here?

davidfowl: I was thinking we should use the enum instead.

anurse: Yeah, the enum seems more appropriate.

mikaelm12:
AppVeyor and Travis failures 👀

src/Microsoft.AspNetCore.Sockets/TransportType.cs
moozzyk: As per our conversation a few days ago - Transport should be common so that we don't have to copy this enum on the client. Also, All/Streaming should not be part of this enum. Finally I understand why this enum is flags but it's a bit weird.

mikaelm12: Yeah, should this move into `Microsoft,AspNetCore.Sockets.Common`?

davidfowl:
🆙 📅

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -31,6 +31,7 @@ public class HttpConnectionDispatcherTests
             var context = new DefaultHttpContext();
             var services = new ServiceCollection();
             services.AddEndPoint<TestEndPoint>();
+            services.AddOptions();
```
BrennanConroy: Could we put `AddOptions` in `AddEndPoint`?

davidfowl: No, AddOptions is added by hosting usually. It's only needed in unit tests.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -101,6 +104,37 @@ public class HttpConnectionDispatcherTests
             }
         }
 
+        [Theory]
+        [InlineData(TransportType.WebSockets, "/ws", 404)]
+        [InlineData(TransportType.ServerSentEvents, "/sse", 404)]
```
anurse: Add tests for the transports that **are** supported? Our other tests don't set `SupportedTransports`, I think, so we don't test that scenario currently.

src/Microsoft.AspNetCore.Sockets.Common/TransportType.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    [Flags]
+    public enum TransportType
+    {
+        WebSockets = 1,
+        ServerSentEvents = 2,
+        LongPolling = 4,
+        All = WebSockets | ServerSentEvents | LongPolling
```
moozzyk: Suppose we use this enum to allow the user to pick the transport they want to use. What it would mean if the user passes TransportType.All? Can .All be defined outside this enum as a const or not at all? We pollute a public type and then use .All only in one place.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -101,6 +104,37 @@ public class HttpConnectionDispatcherTests
...
+        public async Task EndPointThatOnlySupportsLongPollingRejectsOtherTransports(TransportType transportType, string path, int status)
```
mikaelm12: Would it be a little cleaner to instead provide the supported transport type and the provided type in the test params? This way you don't have to have separate tests for each of the supported transports. You would have to keep the `EndPointThatOnlySupportsWebSocketsAndSSERejectsLongPolling` since you removed the streaming enum though.

src/Microsoft.AspNetCore.Sockets.Common/TransportType.cs
anurse: Well, I think it would work fine. The user is choosing which transports to allow the client to use. `All` (or any combination of the flags really) just means "Pick among these transports based on framework-defined logic"

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
anurse: I _think_ you can OR flags in an Attribute value, but I may be wrong. It's still a compile-time constant expression...

src/Microsoft.AspNetCore.Sockets.Common/TransportType.cs
moozzyk: This would be the default behavior - no need to provide any transport type...

anurse: Sure, but the point is that it being a Flags enum is valid there as well

moozzyk: I didn't like it being Flags either :D but meh...

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -401,7 +420,7 @@ public class HttpConnectionDispatcherTests
                 context.Response.Body = strm;
                 var services = new ServiceCollection();
                 services.AddOptions();
-                services.AddEndPoint<TestEndPoint>(options =>
+                services.AddEndPoint<ImmediatelyCompleteEndPoint>(options =>
```
anurse: See? We even had a thing for exactly this purpose :)

PR #297, opened 2017-03-14
use new protocol for '/send'

anurse:
This uses the transport protocol from long-polling as the payload in `/send`. It's controlled by `Content-Type`, so if the `Content-Type` is empty, we assume the legacy behavior (querystring `format` and body is entire message). If the `Content-Type` is one of the two used for the message formats, we parse it as such.

JavaScript client is NOT updated, but C# server and client ARE updated.

/cc @moozzyk @mikaelm12

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
@@ -221,6 +221,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
                             message.SendResult?.TrySetResult(null);
                         }
                     }
+                    else
+                    {
+                        _logger.LogDebug("No messages in batch to send");
+                    }
                 }
             }
             catch (OperationCanceledException)
--- a/src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
@@ -1,18 +1,23 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Pipelines;
+using System.IO.Pipelines.Text.Primitives;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Text.Formatting;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Sockets.Internal.Formatters;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
-using System;
-using System.Buffers;
-using System.Collections.Generic;
-using System.Net;
-using System.Net.Http;
-using System.Net.Http.Headers;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -316,31 +321,51 @@ private Task ProcessNegotiate(HttpContext context)
...
+                await stream.FlushAsync();
                 buffer = stream.ToArray();
             }
 
-            var format =
-                string.Equals(context.Request.Query["format"], "binary", StringComparison.OrdinalIgnoreCase)
-                    ? MessageType.Binary
-                    : MessageType.Text;
+            IList<Message> messages;
+            if (string.Equals(context.Request.ContentType, MessageFormatter.TextContentType))
```
BrennanConroy: `StringComparison`

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -407,5 +432,29 @@ private Task ProcessNegotiate(HttpContext context)
...
+            }
+
+            if (payload.Unread[0] != MessageFormatter.GetFormatIndicator(messageFormat))
+            {
+                throw new FormatException($"Format indicator '{(char)payload.Unread[0]}' does not match format determined by Content-Type '{MessageFormatter.GetContentType(messageFormat)}'");
+            }
+
+            payload.Advance(1);
+
+            _parser.Reset();
```
BrennanConroy: Shouldn't we be worried about multiple connections using the same `MessageParser`?

anurse: Yes, good catch. In the transports it's OK because each transport instance is per-connection

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -171,43 +176,76 @@ private IList<Message> ParsePayload(byte[] payload)
...
                 {
+                    // Grab as many messages as we can from the channel
+                    messages = new List<SendMessage>();
                     while (!cancellationToken.IsCancellationRequested && _application.Input.TryRead(out SendMessage message))
                     {
-                        sendTcs = message.SendResult;
+                        messages.Add(message);
+                    }
+
+                    if (messages.Count > 0)
```
anurse: Yeah, I guess I just don't want to introduce an assert where one isn't needed. It's benign for the Task to complete and no messages to be available, since we'll just restart the loop.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -171,43 +176,76 @@ private IList<Message> ParsePayload(byte[] payload)
...
-                            request.Content = new ByteArrayContent(message.Payload);
-                        }
+                        // TODO: We can probably use a pipeline here or some kind of pooled memory.
+                        // But where do we get the pool from? ArrayBufferPool.Instance?
+                        var memoryStream = new MemoryStream();
 
-                        _logger.LogDebug("Sending a message to the server using url: '{0}'. Message type {1}", sendUrl, message.Type);
+                        // Write the messages to the stream
+                        var pipe = memoryStream.AsPipelineWriter();
+                        var output = new PipelineTextOutput(pipe, TextEncoder.Utf8); // We don't need the Encoder, but it's harmless to set.
```
anurse: It's only needed because the only implementation of `IOutput` for Pipelines also implements `ITextOutput` which has an Encoder property, which the constructor sets. I set it to `Utf8` because that's the right format for any text we want to output and I didn't want to risk a null-ref (though to be fair, it should work with `null`)

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -316,31 +321,51 @@ private Task ProcessNegotiate(HttpContext context)
...
 
-            // Collect the message and write it to the channel
-            // TODO: Need to use some kind of pooled memory here.
+            // Read the entire payload to a byte array for now because Pipelines and ReadOnlyBytes
+            // don't play well with each other yet.
             byte[] buffer;
             using (var stream = new MemoryStream())
             {
                 await context.Request.Body.CopyToAsync(stream);
+                await stream.FlushAsync();
```
anurse: Yes indeed it is! I puzzled over why I was getting an empty array out of `ToArray` for a few minutes before I realized it was required :).

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -316,31 +321,51 @@ private Task ProcessNegotiate(HttpContext context)
...
+                messages = ParseSendBatch(ref reader, MessageFormat.Text);
+            }
+            else if (string.Equals(context.Request.ContentType, MessageFormatter.BinaryContentType))
+            {
+                var reader = new BytesReader(buffer);
+                messages = ParseSendBatch(ref reader, MessageFormat.Binary);
+            }
+            else
+            {
+                // Legacy, single message raw format
```
anurse: Yes

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -407,5 +432,29 @@ private Task ProcessNegotiate(HttpContext context)
 
             return connectionState;
         }
+
+        private IList<Message> ParseSendBatch(ref BytesReader payload, MessageFormat messageFormat)
+        {
+            var messages = new List<Message>();
+
+            if (payload.Unread.Length == 0)
```
anurse: In general we've tried to be lenient with empty payloads, so I'm just doing the same here. An empty payload here is odd, but not necessarily exceptional in that our behavior is deterministic (unlike a parse error in a payload where we don't know how to recover).

test/Microsoft.AspNetCore.Sockets.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -201,5 +203,64 @@ public class LongPollingTransportTests
...
+                    await connectionToTransport.Out.WriteAsync(new SendMessage(Encoding.UTF8.GetBytes("World"), MessageType.Binary, tcs2)).OrTimeout();
+
+                    // Start the transport
+                    await longPollingTransport.StartAsync(new Uri("http://fakeuri.org"), channelConnection);
+
+                    connectionToTransport.Out.Complete();
+
+                    await longPollingTransport.Running.OrTimeout();
+
+                    await longPollingTransport.Running.OrTimeout();
```
anurse: Yep, in fact I did it again somewhere else but noticed it earlier ;). Thanks!

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
moozzyk: OK

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: OK. Do you think we should log this?

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
moozzyk: OK. I should have checked the API - I thought there was a ctor that didn't take the encoder.

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
anurse: Sure, I added some logging.

PR #300, opened 2017-03-14
Creating TypeScript Client npm module during build

moozzyk:


moozzyk:
I don't think manipulating JSON is common in our build. In this particular case I hoped I would be able to get away without touching package.json at all by providing the version to `npm pack` (a functionality similar to what nuget.exe offers) but apparently it's not possible. 
In general I like using the `npm version` better than a general purpose JsonPoke for updating versions of npm modules since it's part of the ecosystem.

PR #301, opened 2017-03-14
port aspnet/WebSockets#149

anurse:
merging as soon as the builds are green because this is blocking the build.

See aspnet/WebSockets#149 for reasoning.

/cc @JunTaoLuo @moozzyk @BrennanConroy @mikaelm12 @davidfowl


test/Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest/Autobahn/AutobahnExpectations.cs
```diff
@@ -13,16 +13,17 @@ public class AutobahnExpectations
         private Dictionary<string, Expectation> _expectations = new Dictionary<string, Expectation>();
         public bool Ssl { get; }
         public ServerType Server { get; }
+        public string Environment { get; }
```
JunTaoLuo: Why is this added?

anurse: Good catch, it's something the WebSockets tests use but SignalR doesn't. Copy-pasta.

PR #302, opened 2017-03-14
autobahn + appveyor = sadness

anurse:
AppVeyor isn't a fan of autobahn, I'm just going to remove the install command (and the test condition will skip the tests since they aren't available). We test it on Travis and our CI so it's still OK. This has been a consistent problem, and I'm just tired of restarting builds ;)

See https://ci.appveyor.com/project/aspnetci/signalr/build/1.0.502

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,8 +10,6 @@ branches:
 build_script:
   - ps: .\build.ps1
 install:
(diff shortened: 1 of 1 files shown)


moozzyk:
It has been really flaky on AppVeyor recently indeed and does not seem like something we can control. Hopefully running on Travis only is good enough.

PR #306, opened 2017-03-15
tidy up status codes and transport conflicts

### LINKED ISSUE #230: Ensure HTTP status codes are correct as per spec
See the transports spec for details.

Part of #174

anurse:
1. WebSockets and SSE transports now refuse a second connection with `409 Conflict`, as specified in the spec.

2. Added a test to ensure Long Polling does what the spec asks (as updated in #305)

fixes #230 

/cc @moozzyk @mikaelm12


test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    internal class TestWebSocketConnectionFeature : IHttpWebSocketConnectionFeature
+    {
+        public bool IsWebSocketRequest => true;
+
+        public WebSocketConnection Client { get; private set; }
+
+        public ValueTask<IWebSocketConnection> AcceptWebSocketConnectionAsync(WebSocketAcceptContext context)
+        {
+            var pipelineFactory = new PipeFactory(ManagedBufferPool.Shared);
```
mikaelm12: Wrap the PipeFactory in a using

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -314,7 +317,72 @@ public class HttpConnectionDispatcherTests
...
+            await task1.OrTimeout();
+
+            // Send a message from the app to complete Task 2
+            await state.Connection.Transport.Output.WriteAsync(new Message(Encoding.UTF8.GetBytes("Hello, World"), MessageType.Text));
+
+            await task2.OrTimeout();
+
+            // Verify the results
+            Assert.Equal(StatusCodes.Status204NoContent, context1.Response.StatusCode);
+            Assert.Equal("", GetContentAsString(context1.Response.Body));
```
mikaelm12: nit: string.Empty

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
anurse: Well, we can't quite do that, but we do need to handle disposing the factory. The `IWebSocketConnection`s need the `PipeFactory` to stay alive while they run. I'll take a look at where we could dispose them. To be honest though, if we never disposed them it'd be fine. The only resource they hold is the buffer pool and the one we're using in this case is a global one which lives for the life of the app and is never Disposed (which I used because this is test code).

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -345,7 +345,7 @@ private Task ProcessNegotiate(HttpContext context)
             }
         }
 
-        private async Task<bool> EnsureConnectionStateAsync(ConnectionState connectionState, HttpContext context, string transportName)
+        private async Task<bool> EnsureConnectionStateAsync(ConnectionState connectionState, HttpContext context, string transportName, bool allowReconnect)
```
davidfowl: Add a default value that's the most common (false)

davidfowl:
> WebSockets and SSE transports now refuse a second connection with 409 Conflict, as specified in the spec.

Am confused here. Because https://github.com/aspnet/SignalR/blob/a64f73a3b985fa46d89f10ebc7321895ee63c845/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs#L260.

anurse:
Well it wasn't working. I'll debug a bit.

anurse:
Looks like it was just a problem with my first test. I've fixed the test and removed the reconnect code I added.

anurse:
Also there wasn't any test for WebSockets, so I updated that.

SignalR.sln
```diff
@@ -47,7 +47,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "WebSocketSample", "samples\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.Sockets.Client", "src\Microsoft.AspNetCore.Sockets.Client\Microsoft.AspNetCore.Sockets.Client.csproj", "{623FD372-36DE-41A9-A564-F6040D570DBD}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.Client.Tests", "test\Microsoft.AspNetCore.Sockets.Client.Tests\Microsoft.AspNetCore.Client.Tests.csproj", "{B19C15A5-F5EA-4CA7-936B-1166ABEE35C4}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.Sockets.Client.Tests", "test\Microsoft.AspNetCore.Sockets.Client.Tests\Microsoft.AspNetCore.Sockets.Client.Tests.csproj", "{B19C15A5-F5EA-4CA7-936B-1166ABEE35C4}"
```
moozzyk: This project contains both Sockets and HubConnection tests, so I don't think it should be renamed (unless we split it)

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -384,7 +415,7 @@ public class HttpConnectionDispatcherTests
             return messages;
         }
 
-        private static DefaultHttpContext MakeRequest<TEndPoint>(string path, ConnectionState state, string format = null) where TEndPoint : EndPoint
+        private static DefaultHttpContext MakeRequest<TEndPoint>(string path, ConnectionState state, string format = null, bool isWebSocketRequest = false) where TEndPoint : EndPoint
```
moozzyk: you probably could just check if the path ends with /ws

anurse: Yeah, but that seems a little too magic... I don't know why... it's just a feeling :)

SignalR.sln
anurse: I'd rather call it SignalR.Client.Tests then. Sockets is "part" of SignalR (yeah, we can talk about layering, but from the larger perspective it's basically all part of the same product). `Microsoft.AspNetCore.Client.Tests` is just too broad and kinda confusing IMO

moozzyk: I am fine with `SignalR.Client.Tests`. We did not want to split tests for Sockets layer vs. SiganlR layer because it did not seem like it was worth it.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
moozzyk: I am fine either way

PR #307, opened 2017-03-16
Fixes a race where wrong task could be awaited in channel completion 

moozzyk:
No new tests - we have coverage for the scenarios I called out in the commit description.


src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -52,18 +52,31 @@ public Connection(Uri url, ILoggerFactory loggerFactory)
...
             {
-                throw new InvalidOperationException("Cannot start a connection that is not in the Initial state.");
+                return Task.FromException(
+                    new InvalidOperationException("Cannot start a connection that is not in the Initial state."));
             }
 
+            StartAsyncInternal(transport, httpClient)
+                .ContinueWith(t =>
+                {
+                    if (t.IsFaulted)
```
anurse: How about `IsCancelled`?

test/Microsoft.AspNetCore.SignalR.Tests/Microsoft.AspNetCore.SignalR.Tests.csproj
```diff
@@ -13,7 +13,12 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Remove="XUnitLoggerProvider.cs" />
```
anurse: This seems wrong. Is it needed?

moozzyk: Seems wrong indeed. Probably a VS job. Will double check.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: StartAsyncInernal never returns a cancelled task. Do you think we should add this proactively?

anurse: I think it's better to be safe than sorry and go for full completeness. Especially since if `StartAsyncInternal` ever throws an `OperationCancelledException` it will result in it being marked cancelled, not faulted.

This would actually make a useful helper someday (`.ContinueBySignallingTcs` or something). But we're not at the point of needing it in a helper (and we may never be ;)). But just in case we ever did, or we looked at this code and reused it somewhere, I'd rather be complete ;)

anurse:
FYI https://github.com/aspnet/Testing/pull/260

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: OK. I will check if the task is cancelled.

TaskAsyncHelper strikes back!

moozzyk:
I don't want to block this on your change to Testing but once it is merged we would switch.

anurse:
Yeah, that was my thought too, no need to block. Especially since my change may have to go somewhere else, since it would create a circular dependency to put it in Testing.

PR #317, opened 2017-03-17
Replacing onError with onClosed

moozzyk:
Unifying errors in transports

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -1,4 +1,4 @@
-import { DataReceived, ErrorHandler } from "./Common"
+import { DataReceived, TransportClosed } from "./Common"
 import { IHttpClient } from "./HttpClient"
 import * as Formatters from "./Formatters";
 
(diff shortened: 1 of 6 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -39,9 +39,12 @@ export class WebSocketTransport implements ITransport {
 
             webSocket.onclose = (event: CloseEvent) => {
                 // webSocket will be null if the transport did not start successfully
-                if (this.webSocket && (event.wasClean === false || event.code !== 1000)) {
-                    if (this.onError) {
-                        this.onError(event);
+                if (this.onClosed && this.webSocket) {
+                    if (event.wasClean === false || event.code !== 1000) {
+                        this.onClosed(new Error(`Reason: ${event.reason}. Code: ${event.code}`));
```
anurse: Can we put a clearer textual description in there too? Something like `WebSocket connection terminated with status code: ${event.code} - ${event.reason}`

moozzyk: Changing to: `Websocket closed with status code: ${event.code} (${event.reason})`

PR #325, opened 2017-03-21
the simple to use Cors 

WestFive:
yes i hava a project to use core signalr 
 i very love this guys.
when i  build my project the cors is a big trouble .

and use it not  only in localhost:5000=>>>

### DIFF
--- a/samples/ChatSample/Program.cs
+++ b/samples/ChatSample/Program.cs
@@ -19,6 +19,7 @@ namespace ChatSample
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseIISIntegration()
                 .UseStartup<Startup>()
+                .UseKestrel().UseUrls("http://*.5000")//Add this Access to by IP
                 .Build();
 
             host.Run();
--- a/samples/ChatSample/Startup.cs
+++ b/samples/ChatSample/Startup.cs
@@ -39,6 +39,12 @@ namespace ChatSample
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+             // if you want use Cors  must add this first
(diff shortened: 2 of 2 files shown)


samples/ChatSample/Program.cs
```diff
@@ -19,6 +19,7 @@ public static void Main(string[] args)
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseIISIntegration()
                 .UseStartup<Startup>()
+                .UseKestrel().UseUrls("http://*.5000")//Add this Access to by IP
```
moozzyk: I don't think we want to hardcode the port. Also, I believe the `.` is incorrect. I think It should be `*:5000` and not `*.5000`

samples/ChatSample/Startup.cs
```diff
@@ -39,6 +39,12 @@ public Startup(IHostingEnvironment env)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+             // if you want use Cors  must add this first
+            //Add Cors
+            services.AddCors(options => options.AddPolicy("AllowAll", p => p.AllowAnyOrigin()
+                                                                 .AllowAnyMethod()
+                                                                  .AllowAnyHeader()
+                                                                  .AllowCredentials()));
```
moozzyk: I don't think we would want to have something like this in a sample - it allows requests sent in any form sent from anywhere. I think the best practice is to only allow what is required instead of allow anything.

davidfowl:
I don't understand the purpose of this change, why are you adding cors and exposing the sample to the internet? We don't do that in any of our samples.

WestFive:
well thanks i think i understand it.

PR #327, opened 2017-03-21
Completing pending calls when the connection is closed

moozzyk:


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -48,7 +52,7 @@ export class HubConnection {
         var descriptor = JSON.parse(data);
         if (descriptor.Method === undefined) {
             let invocationResult: InvocationResultDescriptor = descriptor;
-            let callback = this.callbacks[invocationResult.Id];
+            let callback = this.callbacks.get(invocationResult.Id);
```
davidfowl: What's the difference?

moozzyk: It did not work when running on node. As per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@iterator it is also not supported on IE, Opera and Safari.

moozzyk:
:watch:

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
davidfowl: I highly doubt that.

moozzyk: I *did* debug this.

moozzyk: And I also pointed to documentation which says it is not supported on some browsers.

moozzyk: @davidfowl:
```
PS C:\source\nodemaptest> node
>
(To exit, press ^C again or type .exit)
> var map = new Map();
undefined
> map['a'] = 'b'
'b'
> map['a'] = 'b';
'b'
> map.length
undefined
> map.size
0
> map.set('a', 'b');
Map { 'a' => 'b', a: 'b' }
> map.size
1
>
```

davidfowl: The problem is new Map(), it should just be var map = {}

moozzyk: HubConnection [uses Map](https://github.com/aspnet/SignalR/blob/dev/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts#L21-L22) not {}.

davidfowl: Why?

moozzyk: Because it is strongly typed (in TS sense) and gives you intellisense in vscode.

PR #330, opened 2017-03-21
Removing site.js

moozzyk:
site.js is basically empty but was configured to create a (empty) site.min.js which would occasionally be marked as modified even there was no content (presumably because of encoding change?)


BrennanConroy:
:shipit:

PR #332, opened 2017-03-21
Handling exceptions thrown when invoking a hub method

moozzyk:


src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -134,15 +132,26 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
-            await _connection.SendAsync(ms.ToArray(), MessageType.Text, cancellationToken);
-            _logger.LogInformation("Sending Invocation #{0} complete", descriptor.Id);
+                // TODO: Format.Text - who, where and when decides about the format of outgoing messages
+                await _connection.SendAsync(ms.ToArray(), MessageType.Text, cancellationToken);
+                _logger.LogInformation("Sending Invocation #{0} complete", descriptor.Id);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(0, ex, "Sending Invocation #{0} failed", descriptor.Id);
+                irq.Completion.SetException(ex);
```
moozzyk: this needs to be TrySetException - the result can be set in the meantime if the connection closes.

davidfowl: Make the change now?

moozzyk: Yes, I will.

Issue #333, opened 2017-03-22
Inheritance in Hubs


Lukino2000:
Hi,
for my understanding at the moment it is not possible using inheritance in hub declaration.

I suppose this is cause of HubEndPoint.DiscoverHubMethods where you use TypeInfo.DeclaredMethods.

is there any way to use a BaseHub with public methods?

thanks in advance and sorry for any misunderstanding

davidfowl:
Sure, that's just an oversight. @BrennanConroy can you take a look at this?

BrennanConroy:
Closed via #334

PR #334, opened 2017-03-22
Allow overridden and new methods on hubs

BrennanConroy:
https://github.com/aspnet/SignalR/issues/333

@anurse @moozzyk @mikaelm12

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
@@ -324,66 +324,6 @@ namespace Microsoft.AspNetCore.SignalR
                     _logger.LogDebug("Hub method '{methodName}' is bound", methodName);
                 }
             }
-
-            // used for checking against methods in base classes
-            //var methods = new Dictionary<string, MethodInfo>();
-            //do
-            //{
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -309,23 +309,65 @@ private void DiscoverHubMethods()
         {
             var typeInfo = typeof(THub).GetTypeInfo();
 
-            foreach (var methodInfo in typeInfo.DeclaredMethods.Where(m => IsHubMethod(m)))
+            // used for checking against methods in base classes
+            var methods = new Dictionary<string, MethodInfo>();
```
davidfowl: This doesn't look right at all. Did you look at MVC by any chance?

anurse: My main question would be if MVC blocks the use of the `new` keyword intentionally or if it is an unintentional omission. It does seem odd to me that we would half-support derived classes by supporting overriding but not shadowing.

BrennanConroy:
:up: :date: No more `new` methods

BrennanConroy:
@rynowak Could you take a look?

PR #338, opened 2017-03-25
Convert to netcoreapp2.0

BrennanConroy:
@davidfowl @anurse @moozzyk @mikaelm12

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/TextMessageParser.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/TextMessageParser.cs
@@ -85,14 +85,15 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
 
         private bool TryReadLength(ref BytesReader buffer)
         {
+            // Read until the first ':' to find the length
             var lengthBuffer = buffer.ReadBytesUntil((byte)TextMessageFormatter.FieldDelimiter);
 
             if (lengthBuffer == null)
             {
+                // Insufficient data
                 return false;
             }
 
-            // Read until the first ':' to find the length
             var lengthSpan = lengthBuffer.Value.ToSingleSpan();
 
             // Parse the length
--- a/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/TextMessageParser.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/TextMessageParser.cs
@@ -85,18 +85,20 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
 
         private bool TryReadLength(ref BytesReader buffer)
         {
-            // Read until the first ':' to find the length
-            var lengthSpan = buffer.ReadBytesUntil((byte)TextMessageFormatter.FieldDelimiter)?.ToSingleSpan();
-            if (lengthSpan == null)
+            var lengthBuffer = buffer.ReadBytesUntil((byte)TextMessageFormatter.FieldDelimiter);
+
+            if (lengthBuffer == null)
             {
-                // Insufficient data
                 return false;
             }
 
+            // Read until the first ':' to find the length
+            var lengthSpan = lengthBuffer.Value.ToSingleSpan();
+
             // Parse the length
-            if (!PrimitiveParser.TryParseInt32(lengthSpan.Value, out var length, out var consumedByLength, encoder: TextEncoder.Utf8) || consumedByLength < lengthSpan.Value.Length)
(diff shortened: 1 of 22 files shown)


src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/TextMessageParser.cs
```diff
@@ -85,18 +85,20 @@ public bool TryParseMessage(ref BytesReader buffer, out Message message)
...
         private bool TryReadLength(ref BytesReader buffer)
         {
-            // Read until the first ':' to find the length
-            var lengthSpan = buffer.ReadBytesUntil((byte)TextMessageFormatter.FieldDelimiter)?.ToSingleSpan();
-            if (lengthSpan == null)
+            var lengthBuffer = buffer.ReadBytesUntil((byte)TextMessageFormatter.FieldDelimiter);
+
+            if (lengthBuffer == null)
             {
-                // Insufficient data
```
moozzyk: is this comment no longer valid?

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
```diff
@@ -701,6 +697,26 @@ private Task SendCoreAsync<T>(bool fin, WebSocketOpcode opcode, int payloadAlloc
...
 
+        private void WritePayload<T>(ref WritableBuffer buffer, int payloadLength, Action<WritableBuffer, uint, int, T> payloadWriter, T payload)
+        {
+            var maskingKey = Span<byte>.Empty;
+            if (_maskingKeyBuffer != null)
+            {
+                // Get a span of the output buffer for the masking key, write it there, then advance the write head.
+                maskingKey = buffer.Buffer.Slice(0, 4).Span;
+                WriteMaskingKey(maskingKey);
+                buffer.Advance(4);
```
moozzyk: I think 4 should be a const (even if local)

test/Microsoft.Extensions.WebSockets.Internal.Tests/WebSocketConnectionTests.SendAsync.cs
```diff
@@ -143,15 +143,15 @@ public partial class WebSocketConnectionTests
...
-            var mask = data.Slice(2, 4);
-            var actualPayload = data.Slice(6);
+            var mask = new byte[] { data[2], data[3], data[4], data[5] };
+            var actualPayload = new byte[data.Length - 6];
 
             // Unmask the payload
             for (int i = 0; i < actualPayload.Length; i++)
             {
-                actualPayload[i] = (byte)(mask[i % 4] ^ actualPayload[i]);
+                actualPayload[i] = (byte)(mask[i % 4] ^ data[i + 6]);
```
moozzyk: is it safe to iterate over `data` like this?

davidfowl:
I can't seem to reproduce these null refs locally.

BrennanConroy:
Nice! How did you figure that out? I would like to know for future debugging, I figured it was something with `_unreadSegments`

davidfowl:
@BrennanConroy I just narrowed it down after realizing that the buffer was reported as empty. This was the original repro https://github.com/dotnet/corefxlab/issues/1374

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
```diff
@@ -700,12 +700,14 @@ private Task SendCoreAsync<T>(bool fin, WebSocketOpcode opcode, int payloadAlloc
         private void WritePayload<T>(ref WritableBuffer buffer, int payloadLength, Action<WritableBuffer, uint, int, T> payloadWriter, T payload)
         {
             var maskingKey = Span<byte>.Empty;
+            var keySize = sizeof(uint);
```
moozzyk: nit: const?

davidfowl: doesn't work

test/Microsoft.Extensions.WebSockets.Internal.Tests/WebSocketConnectionTests.SendAsync.cs
anurse: It should be. We know the data is at least `(6 + payload length)` bytes long. The code's a bit odd, but it's probably like this because of a fast-span issue.

src/Microsoft.Extensions.WebSockets.Internal/WebSocketConnection.cs
anurse: `sizeof(uint)` would be even better :)

davidfowl: That what I did

PR #340, opened 2017-03-27
Use policy names in EndPointOptions

BrennanConroy:


src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -12,14 +14,27 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public static class AuthorizeHelper
     {
-        public static async Task<bool> AuthorizeAsync(HttpContext context, AuthorizationPolicy policy)
+        public static async Task<bool> AuthorizeAsync(HttpContext context, IEnumerable<string> policies)
```
moozzyk: Is `policies` guaranteed to be not null? (asking because policy could have been)

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -1,64 +1,80 @@
...
 
+using System.Collections.Generic;
+using System.Net;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Internal;
+using System;
```
davidfowl: Sort properly

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -1,64 +1,80 @@
...
 
 namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public static class AuthorizeHelper
     {
-        public static async Task<bool> AuthorizeAsync(HttpContext context, AuthorizationPolicy policy)
+        public static async Task<bool> AuthorizeAsync(HttpContext context, List<string> policies)
         {
-            if (policy != null)
+            if (policies.Count == 0)
```
davidfowl: Check null?

src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
```diff
@@ -1,13 +1,14 @@
...
 
+using System.Collections.Generic;
 using Microsoft.AspNetCore.Authorization;
 
 namespace Microsoft.AspNetCore.Sockets
 {
     public class EndPointOptions<TEndPoint> where TEndPoint : EndPoint
     {
-        public AuthorizationPolicy Policy { get; set; }
+        public List<string> Policies { get; set; } = new List<string>();
```
moozzyk: as per our conversation - remove `set`?

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -1,64 +1,80 @@
...
-                    {
-                        newPrincipal = new ClaimsPrincipal(new ClaimsIdentity());
-                    }
+                return true;
+            }
 
-                    context.User = newPrincipal;
-                }
+            var policyProvider = context.RequestServices.GetRequiredService<IAuthorizationPolicyProvider>();
+            if (policyProvider == null)
```
moozzyk: Wouldn't `GetRequiredService` throw if the service was not registered?

src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
```diff
@@ -1,13 +1,14 @@
...
 
+using System.Collections.Generic;
 using Microsoft.AspNetCore.Authorization;
 
 namespace Microsoft.AspNetCore.Sockets
 {
     public class EndPointOptions<TEndPoint> where TEndPoint : EndPoint
     {
-        public AuthorizationPolicy Policy { get; set; }
+        public List<string> Policies { get; } = new List<string>();
```
anurse: Not `IList`?

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -12,53 +14,62 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 {
     public static class AuthorizeHelper
     {
-        public static async Task<bool> AuthorizeAsync(HttpContext context, AuthorizationPolicy policy)
+        public static async Task<bool> AuthorizeAsync(HttpContext context, List<string> policies)
```
anurse: `IList`

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
anurse: `(policies?.Count ?? 0) == 0` ?

moozzyk:
I think you need to add some tests

blowdart:
You changed the default return value for a security function from true to false, and you have no tests? Add some tests.

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -12,53 +13,62 @@ namespace Microsoft.AspNetCore.Sockets.Internal
...
-                    {
-                        newPrincipal = new ClaimsPrincipal(new ClaimsIdentity());
-                    }
+                return true;
+            }
 
-                    context.User = newPrincipal;
-                }
+            var policyProvider = context.RequestServices.GetRequiredService<IAuthorizationPolicyProvider>();
+            var authorizeAttributes = new List<AuthorizeAttribute>();
```
davidfowl: I just noticed this is a list of `AuthorizeAttribute`, that seems bizarre.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -424,6 +426,171 @@ public class HttpConnectionDispatcherTests
             Assert.Equal(MessageType.Close, messages[3].Type);
         }
 
+        [Fact]
+        public async Task UnauthorizedConnectionFailsToStartEndPoint()
```
davidfowl: Why are all of your tests on `/poll`

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
BrennanConroy: Yeah it is odd, https://github.com/aspnet/Security/blob/22d2fe99c6fd9806b36025399a217a3a8b4e50f4/src/Microsoft.AspNetCore.Authorization/AuthorizationPolicy.cs#L111 the only thing that implements IAuthorizeData is AuthorizeAttribute.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
BrennanConroy: Its either `/sse` `/poll` or `/ws` these tests don't test if a transport can send, they test if you can do something with the endpoint after authorization, so we don't care what path as long as it works.

BrennanConroy:
🆙 📅

src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
```diff
@@ -1,13 +1,13 @@
...
 
-using Microsoft.AspNetCore.Authorization;
+using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Sockets
 {
     public class EndPointOptions<TEndPoint> where TEndPoint : EndPoint
     {
-        public AuthorizationPolicy Policy { get; set; }
+        public IList<string> Policies { get; } = new List<string>();
```
anurse: `AuthorizationPolicyNames`. Otherwise its just a random list of strings and I have no idea what is expected there.

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -12,53 +13,62 @@ namespace Microsoft.AspNetCore.Sockets.Internal
...
+            foreach (var policy in policies)
+            {
+                authorizeData.Add(new AuthorizeAttribute(policy));
+            }
 
-                // Challenge
-                if (policy.AuthenticationSchemes != null && policy.AuthenticationSchemes.Count > 0)
+            var authorizePolicy = await AuthorizationPolicy.CombineAsync(policyProvider, authorizeData);
+            if (authorizePolicy.AuthenticationSchemes != null && authorizePolicy.AuthenticationSchemes.Count > 0)
+            {
```
anurse: Looks like the code path inside this `if` isn't tested. Should be a simple matter of adding the ability for the `TestAuthenticationHandler` to authenticate a user with some artificial claims.

BrennanConroy:
🆙 📅

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -452,6 +454,273 @@ public class HttpConnectionDispatcherTests
...
+
+            // would block if EndPoint was executed
+            await dispatcher.ExecuteAsync<BlockingEndPoint>("", context).OrTimeout();
+
+            Assert.Equal(StatusCodes.Status401Unauthorized, context.Response.StatusCode);
+        }
+
+        private class TestAuthenticationHandler : IAuthenticationHandler
+        {
+            private readonly HttpContext Context;
```
anurse: So many contexts that I don't know which context this context is in (contextually). Rename this to `HttpContext` or something.

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -12,53 +13,62 @@ namespace Microsoft.AspNetCore.Sockets.Internal
...
-                }
+            foreach (var policy in policies)
+            {
+                authorizeData.Add(new AuthorizeAttribute(policy));
+            }
 
-                // Challenge
-                if (policy.AuthenticationSchemes != null && policy.AuthenticationSchemes.Count > 0)
+            var authorizePolicy = await AuthorizationPolicy.CombineAsync(policyProvider, authorizeData);
+            if (authorizePolicy.AuthenticationSchemes != null && authorizePolicy.AuthenticationSchemes.Count > 0)
```
anurse: I don't think you could, since it would mean comparing an `int?` with `0`. I'm not positive I know the behavior of that, but even if it works it seems kinda unclear.

Maybe we need an List.IsNullOrEmpty ;P

moozzyk: This works just fine: 
```
            List<int> myList = null;

            if (myList?.Count > 0)
            {
                Console.WriteLine("non-empty");
            }
            else
            {
                Console.WriteLine("empty");
            }
```

moozzyk: Which dotpeek decompiled as:
```
      List<int> list = (List<int>) null;
      // ISSUE: explicit non-virtual call
      if (list != null && __nonvirtual (list.Count) > 0)
        Console.WriteLine("non-empty");
      else
        Console.WriteLine("empty");
```

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -12,53 +13,62 @@ namespace Microsoft.AspNetCore.Sockets.Internal
...
-                        await context.Authentication.ChallengeAsync(scheme, properties: null);
+                        newPrincipal = SecurityHelper.MergeUserPrincipal(newPrincipal, result);
                     }
                 }
-                else
+
+                if (newPrincipal == null)
                 {
-                    await context.Authentication.ChallengeAsync(properties: null);
+                    newPrincipal = new ClaimsPrincipal(new ClaimsIdentity());
```
BrennanConroy: Its just a minor optimization, we don't want to allocate it unless none of the `AuthenticationSchemes` could be authenticated.

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
anurse: Yeah, I guess it's just the fact that we had to decompile just to confirm that it works as expected that worries me. It's more a question of clarity

moozzyk: That's all right - it was just a nit and I am fine with keeping it as is. I only did the decompilation to make sure that I am not crazy ;D

src/Microsoft.AspNetCore.Sockets/Internal/AuthorizeHelper.cs
```diff
@@ -12,53 +13,62 @@ namespace Microsoft.AspNetCore.Sockets.Internal
...
                 }
 
-                return false;
+                context.User = newPrincipal;
+            }
+
+            var authService = context.RequestServices.GetRequiredService<IAuthorizationService>();
+            if (await authService.AuthorizeAsync(context.User, context, authorizePolicy))
```
BrennanConroy: It doesn't seem possible to have `context.User == null`

PR #343, opened 2017-03-28
Upgraded Json.NET version to 10.0.1

kichalla:


moozzyk:
Do we do this everywhere?

kichalla:
Yes, we are doing this all across our repos except a few exceptions (ex: Razor). Any specific concerns?

moozzyk:
No if we are doing this across all/most of the repos.

PR #345, opened 2017-03-28
* Fixing a racey test

moozzyk:
* Fixing a misplaced test file


moozzyk:
:watch:

moozzyk:
Hopefully this will help move this forward:
Scenario:
- client invokes a method on the server
- the method on the server invoked by the client invokes a client method
- the tests verifies that the client method was invoked with the expected argument

The issue here was that the test tried to stop the connection in a client method invoked by the server. However, the client might not have received the result of the server method it invoked (first bullet point). Since now we are completing pending invocations with error when the connection is stopped the client would throw for the unfinished server method invocation and the test would fail. The fix is to wait for both - client method invocation and server method invocation to complete before trying stopping the connection.

BrennanConroy:
:shipit:

PR #356, opened 2017-03-30
Bugfix issue #355: Don't handle Invocation when no InvocationHandler …

### LINKED ISSUE #355: NullReferenceException when HubConnection InvocationHandler is not found
I'm using a HubConnection Client and define no Client method invocation handlers (so no call to .On("SomeFunc", ...))

I got an NullReferenceException when the server calls a client method The HubConnection InvocationHandler is not found (null) but is still getting called.

System.NullReferenceException: 'Object reference not set to an instance of an object.'
   at Microsoft.AspNetCore.SignalR.Client.HubConnection.DispatchInvocation(InvocationDescriptor invocationDescriptor, CancellationToken cancellationToken) in C:\Users\XXX\Source\Repos\SignalR\src\Microsoft.AspNetCore.SignalR.Client\HubConnection.cs:line 220
   at Microsoft.AspNetCore.SignalR.Client.HubConnection.<OnDataReceived>d__29.MoveNext() in C:\Users\XXX\Source\Repos\SignalR\src\Microsoft.AspNetCore.SignalR.Client\HubConnection.cs:line 168

I'm gonna submit a PR to this issue for review.

darxis:
Please review.
Fixes #355

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
@@ -269,6 +269,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
             await hubConnection.StartAsync(Mock.Of<ITransport>());
 
             mockConnection.Raise(c => c.Received += null, new object[] { new byte[] { }, MessageType.Text });
+            mockInvocationAdapter.Verify(a => a.ReadMessageAsync(It.IsAny<Stream>(), It.IsAny<IInvocationBinder>(), It.IsAny<CancellationToken>()), Times.Once());
         }
     }
 }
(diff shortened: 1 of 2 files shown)


moozzyk:
Can we add a test for this? This scenario is quite important...

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -246,5 +247,28 @@ public void CannotCreateHubConnectionWithNullUrl()
             var thrown = await Assert.ThrowsAsync(exception.GetType(), async () => await invokeTask);
             Assert.Same(exception, thrown);
         }
+
+        [Fact]
+        public async Task DoesNotThrowWhenClientMethodCalledButNoInvocationHandlerHasBeenSetUp()
+        {
+            var mockConnection = new Mock<IConnection>();
```
davidfowl: Can you assert something in this test?

davidfowl:
@darxis Thanks!

PR #358, opened 2017-04-01
ServerSentEvents for the C# client

mikaelm12:
+ ServerSentEventsTransport 
+ ServerSentEventsParser
+ Some basic unit tests

I shouldn't reset the parser on incomplete parse results. I need to look into this

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/DefaultTransportFactory.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/DefaultTransportFactory.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             if ((availableServerTransports & TransportType.ServerSentEvents & _requestedTransportType) == TransportType.ServerSentEvents)
             {
-                throw new NotImplementedException();
+                return new ServerSentEventsTransport(_httpClient, _loggerFactory);
             }
 
             if ((availableServerTransports & TransportType.LongPolling & _requestedTransportType) == TransportType.LongPolling)
--- a/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
@@ -1,0 +1,232 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Pipelines;
+using System.IO.Pipelines.Text.Primitives;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Text.Formatting;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public class ServerSentEventsTransport : ITransport
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
+        private static readonly ProductInfoHeaderValue DefaultUserAgentHeader = ProductInfoHeaderValue.Parse(DefaultUserAgent);
+
+        private readonly HttpClient _httpClient;
+        private readonly ILogger _logger;
+        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+
+        private IChannelConnection<SendMessage, Message> _application;
+        private ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
+
+        public Task Running { get; private set; } = Task.CompletedTask;
+
+        public ServerSentEventsTransport(HttpClient httpClient)
+            : this(httpClient, null)
+        { }
+
+        public ServerSentEventsTransport(HttpClient httpClient, ILoggerFactory loggerFactory)
+        {
+            if (httpClient == null)
+            {
+                throw new ArgumentNullException(nameof(_httpClient));
+            }
+
+            _httpClient = httpClient;
+            _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<ServerSentEventsTransport>();
+        }
+
+        public Task StartAsync(Uri url, IChannelConnection<SendMessage, Message> application)
+        {
+            _logger.LogInformation("Starting {0}", nameof(ServerSentEventsTransport));
+
+            _application = application;
+            var sseUrl = Utils.AppendPath(url, "sse");
+            var sendUrl = Utils.AppendPath(url, "send");
+            var sendTask = SendMessages(sendUrl, _transportCts.Token);
+            var receiveTask = OpenConnection(_application, sseUrl, _transportCts.Token);
+
+            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
+            {
+                _logger.LogDebug("Transport stopped. Exception: '{0}'", t.Exception?.InnerException);
+
+                _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
+                return t;
+            }).Unwrap();
+
+            return TaskCache.CompletedTask;
+        }
+
+        public async Task OpenConnection(IChannelConnection<SendMessage, Message> application, Uri url, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("Starting receive loop");
+
+            var request = new HttpRequestMessage(HttpMethod.Get, url);
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
+            var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+
+            var stream = await response.Content.ReadAsStreamAsync();
+
+            var pipelineReader = stream.AsPipelineReader();
+            try
+            {
+
+            while (true)
+            {
+                var result = await pipelineReader.ReadAsync();
+                var input = result.Buffer;
+                var consumed = input.Start;
+                var examined = input.End;
+
+                try
+                {
+                    if (input.IsEmpty && result.IsCompleted)
+                    {
+                        _logger.LogDebug("Done reading from pipeline");
+                        break;
+                    }
+
+                    var parseResult = _parser.ParseMessage(input, out consumed, out examined, out var message);
+
+                    switch (parseResult)
+                    {
+                        case ServerSentEventsMessageParser.ParseResult.Completed:
+                            _application.Output.TryWrite(message);
+                            _parser.Reset();
+                            break;
+                        case ServerSentEventsMessageParser.ParseResult.Incomplete:
+                            if (result.IsCompleted)
+                            {
+                                throw new FormatException("Error parsing data from event stream");
+                            }
+                            break;
+                    }
+                }
+                finally
+                {
+                    pipelineReader.Advance(consumed, examined);
+                }
+            }
+            }
+            finally
+            {
+                _transportCts.Cancel();
+            }
+        }
+
+        public async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("Starting the send loop");
+
+            IList<SendMessage> messages = null;
+            try
+            {
+                while (await _application.Input.WaitToReadAsync(cancellationToken))
+                {
+                    messages = new List<SendMessage>();
+                    while (!cancellationToken.IsCancellationRequested && _application.Input.TryRead(out SendMessage message))
+                    {
+                        messages.Add(message);
+                    }
+
+                    if (messages.Count > 0)
+                    {
+                        _logger.LogDebug("Sending {0} message(s) to the server using url: {1}", messages.Count, sendUrl);
+
+                        var request = new HttpRequestMessage(HttpMethod.Post, sendUrl);
+                        request.Headers.UserAgent.Add(DefaultUserAgentHeader);
+
+                        var memoryStream = new MemoryStream();
+
+                        var pipe = memoryStream.AsPipelineWriter();
+                        var output = new PipelineTextOutput(pipe, TextEncoder.Utf8);
+                        await WriteMessagesAsync(messages, output, MessageFormat.Binary);
+
+                        memoryStream.Seek(0, SeekOrigin.Begin);
+
+                        request.Content = new StreamContent(memoryStream);
+                        request.Content.Headers.ContentType = MediaTypeHeaderValue.Parse(MessageFormatter.GetContentType(MessageFormat.Binary));
+
+                        var response = await _httpClient.SendAsync(request);
+                        response.EnsureSuccessStatusCode();
+
+                        _logger.LogDebug("Message(s) sent successfully");
+                        foreach (var message in messages)
+                        {
+                            message.SendResult?.TrySetResult(null);
+                        }
+                    }
+                }
+            }
+            catch (OperationCanceledException)
+            {
+                _logger.LogError("Send cancelled");
+
+                if (messages != null)
+                {
+                    foreach (var message in messages)
+                    {
+                        message.SendResult?.TrySetCanceled();
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug("Error while sending to '{0}' : '{1}'", sendUrl, ex);
+                if (messages != null)
+                {
+                    foreach (var message in messages)
+                    {
+                        message.SendResult?.TrySetException(ex);
+                    }
+                }
+                throw;
+            }
+            finally
+            {
+                // Make sure the poll loop is terminated
+                _transportCts.Cancel();
+            }
+
+            _logger.LogInformation("Send loop stopped");
+        }
+
+        private async Task WriteMessagesAsync(IList<SendMessage> messages, PipelineTextOutput output, MessageFormat format)
+        {
+            output.Append(MessageFormatter.GetFormatIndicator(format), TextEncoder.Utf8);
+
+            foreach (var message in messages)
+            {
+                _logger.LogDebug("Writing '{0}' message to the server", message.Type);
+
+                var payload = message.Payload ?? Array.Empty<byte>();
+                if (!MessageFormatter.TryWriteMessage(new Message(payload, message.Type, endOfMessage: true), output, format))
+                {
+                    // We didn't get any more memory!
+                    throw new InvalidOperationException("Unable to write message to pipeline");
+                }
+                await output.FlushAsync();
+            }
+        }
+
+        public async Task StopAsync()
+        {
+            _logger.LogInformation("Transport {0} is stopping", nameof(ServerSentEventsTransport));
+            _transportCts.Cancel();
+            _application.Output.TryComplete();
+            await Running;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageFormatter.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageFormatter.cs
@@ -14,11 +14,11 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
         private static readonly byte[] DataPrefix = new byte[] { (byte)'d', (byte)'a', (byte)'t', (byte)'a', (byte)':', (byte)' ' };
         private static readonly byte[] Newline = new byte[] { (byte)'\r', (byte)'\n' };
 
-        private const byte LineFeed = (byte)'\n';
-        private const char TextTypeFlag = 'T';
-        private const char BinaryTypeFlag = 'B';
-        private const char CloseTypeFlag = 'C';
-        private const char ErrorTypeFlag = 'E';
+        static readonly byte LineFeed = (byte)'\n';
+        static readonly char TextTypeFlag = 'T';
+        static readonly char BinaryTypeFlag = 'B';
+        static readonly char CloseTypeFlag = 'C';
+        static readonly char ErrorTypeFlag = 'E';
 
         public static bool TryWriteMessage(Message message, IOutput output)
         {
--- a/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
@@ -1,0 +1,193 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.using System;
+
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
+{
+    public class ServerSentEventsMessageParser
+    {
+        static readonly byte ByteCR = (byte)'\r';
+        static readonly byte ByteLF = (byte)'\n';
+
+        private InternalParseState _internalParserState = InternalParseState.ReadMessageType;
+        private IList<byte[]> _data = new List<byte[]>();
+
+        public ParseResult ParseMessage(ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined, out Message message)
+        { 
+            consumed = buffer.Start;
+            examined = buffer.End;
+            message = new Message();
+            var messageType = MessageType.Text;
+            var reader = new ReadableBufferReader(buffer);
+
+            var start = consumed;
+            var end = examined;
+
+            while (!reader.End)
+            {
+                if (ReadCursorOperations.Seek(start, end, out var lineEnd, ByteLF) == -1)
+                {
+                    //Partial message. We need to read more.
+                    return ParseResult.Incomplete;
+                }
+
+                lineEnd = buffer.Move(lineEnd, 1);
+                var line = ConvertBufferToSpan(buffer.Slice(start, lineEnd));
+                reader.Skip(line.Length);
+
+                //Check for a misplaced '\n'
+                if (line.Length <= 1 || line[line.Length - 2] != ByteCR)
+                {
+                    throw new FormatException("There was an issue with the frame format");
+                }
+
+                //Strip the \r\n from the span
+                line = line.Slice(0, line.Length - 2);
+
+                switch (_internalParserState)
+                {
+                    case InternalParseState.ReadMessageType:
+                        messageType = GetMessageType(line);
+                        start = lineEnd;
+                        _internalParserState = InternalParseState.ReadMessagePayload;
+                        consumed = lineEnd;
+
+                        break;
+                    case InternalParseState.ReadMessagePayload:
+                        if (!StartsWithDataPrefix(line))
+                        {
+                            throw new FormatException("Expected the message prefix 'data: '");
+                        }
+
+                        //Slice away the 'data: '
+                        var newData = line.Slice(6).ToArray();
+
+                        start = lineEnd;
+                        _data.Add(newData);
+                        consumed = lineEnd;
+                        break;
+                    case InternalParseState.ReadEndOfMessage:
+                        if (ReadCursorOperations.Seek(start, end, out lineEnd, ByteLF) == -1)
+                        {
+                            // The message has ended with \r\n\r
+                            return ParseResult.Incomplete;
+                        }
+
+                        if (_data.Count > 0)
+                        {
+                            //Find the final size of the payload
+                            var payloadSize = 0;
+                            foreach (var dataLine in _data)
+                            {
+                                payloadSize += dataLine.Length;
+                            }
+                            var payload = new byte[payloadSize];
+
+                            //Copy the contents of the data array to a single buffer
+                            var offset = 0;
+                            foreach (var dataLine in _data)
+                            {
+                                dataLine.CopyTo(payload, offset);
+                                offset += dataLine.Length;
+                            }
+
+                            message = new Message(payload, messageType);
+                        }
+                        else
+                        {
+                            //Empty message
+                            message = new Message(Array.Empty<byte>(), messageType);
+                        }
+
+                        consumed = buffer.End;
+                        return ParseResult.Completed;
+                }
+
+                //Peek into next byte. If it is a carriage return byte, then advance to the next state
+                if (reader.Peek() == ByteCR)
+                {
+                    _internalParserState = InternalParseState.ReadEndOfMessage;
+                }
+            }
+            return ParseResult.Incomplete;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private Span<byte> ConvertBufferToSpan(ReadableBuffer buffer)
+        {
+            if (buffer.IsSingleSpan)
+            {
+                return buffer.First.Span;
+            }
+            return buffer.ToArray();
+        }
+
+        public void Reset()
+        {
+            _internalParserState = InternalParseState.ReadMessageType;
+            _data.Clear();
+        }
+
+        private bool StartsWithDataPrefix(ReadOnlySpan<byte> line)
+        {
+            var dataPrefix = Encoding.UTF8.GetBytes("data: ");
+            var prefixSpan = new ReadOnlySpan<byte>(dataPrefix);
+            if (line.StartsWith(dataPrefix))
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        private MessageType GetMessageType(ReadOnlySpan<byte> line)
+        {
+            if (!StartsWithDataPrefix(line))
+            {
+                throw new FormatException("Expected the message prefix 'data: '");
+            }
+
+            if (line.Length != 7)
+            {
+                throw new FormatException("There was an error parsing the message type");
+            }
+
+            //Skip the "data: " part of the line
+            var type = (char)line[6];
+            switch (type)
+            {
+                case 'T':
+                    return MessageType.Text;
+                case 'B':
+                    return MessageType.Binary;
+                case 'C':
+                    return MessageType.Close;
+                case 'E':
+                    return MessageType.Error;
+                default:
+                    throw new FormatException($"Unknown message type: '{type}'");
+            }
+        }
+
+        public enum ParseResult
+        {
+            Completed,
+            Incomplete,
+        }
+
+        private enum InternalParseState
+        {
+            Initial,
+            ReadMessageType,
+            ReadMessagePayload,
+            ReadEndOfMessage,
+            Error
+        }
+    }
+}
--- a/test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
+++ b/test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO.Pipelines;
+using System.Text;
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Sockets.Common.Tests.Internal.Formatters
+{
+    public class ServerSentEventsParserTests
+    {
+        [Theory]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: E\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\ndata: Hello\r\ndata: , World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\ndata: Major\r\ndata:  Key\r\ndata:  Alert\r\n\r\n", "Major Key Alert")]
+        [InlineData("data: T\r\n\r\n", "")]
+        public void ParseSSEMessageSuccessCases(string encodedMessage, string expectedMessage)
+        {
+            var buffer = Encoding.UTF8.GetBytes(encodedMessage);
+            var readableBuffer = ReadableBuffer.Create(buffer);
+            var parser = new ServerSentEventsMessageParser();
+            var consumed = new ReadCursor();
+            var examined = new ReadCursor();
+
+            var parsePhase = parser.ParseMessage(readableBuffer, out consumed, out examined, out Message message);
+            Assert.Equal(ServerSentEventsMessageParser.ParseResult.Completed, parsePhase);
+
+            var result = Encoding.UTF8.GetString(message.Payload);
+            Assert.Equal(expectedMessage, result);
+        }
+
+        [Theory]
+        [InlineData("data: X\r\n", "Unknown message type: 'X'")]
+        [InlineData("data: T\n", "There was an issue with the frame format")]
+        [InlineData("data: X\r\n\r\n", "Unknown message type: 'X'")]
+        [InlineData("data: Not the message type\r\n\r\n", "There was an error parsing the message type")]
+        [InlineData("data: T\r\ndata: Hello, World\r\r\n\n", "There was an issue with the frame format")]
+        [InlineData("data: Not the message type\r\r\n", "There was an error parsing the message type")]
+        [InlineData("data: T\r\ndata: Hello, World\n\n", "There was an issue with the frame format")]
+        [InlineData("data: T\r\nfoo: Hello, World\r\n\r\n", "Expected the message prefix 'data: '")]
+        [InlineData("foo: T\r\ndata: Hello, World\r\n\r\n", "Expected the message prefix 'data: '")]
+        [InlineData("food: T\r\ndata: Hello, World\r\n\r\n", "Expected the message prefix 'data: '")]
+        public void ParseSSEMessageFailureCases(string encodedMessage, string expectedExceptionMessage)
+        {
+            var buffer = Encoding.UTF8.GetBytes(encodedMessage);
+            var readableBuffer = ReadableBuffer.Create(buffer);
+            var parser = new ServerSentEventsMessageParser();
+            var consumed = new ReadCursor();
+            var examined = new ReadCursor();
+
+            var ex = Assert.Throws<FormatException>(() => { parser.ParseMessage(readableBuffer, out consumed, out examined, out Message message); });
+            Assert.Equal(expectedExceptionMessage, ex.Message);
+        }
+
+        [Theory]
+        [InlineData("")]
+        [InlineData("data:")]
+        [InlineData("data: T\r\nda")]
+        [InlineData("data: T\r\ndata:")]
+        [InlineData("data: T\r\ndata: Hello, World")]
+        [InlineData("data: T\r\ndata: Hello, World\r")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\\")]
+        [InlineData("data: T\r\ndata: Major\r\ndata:  Key\rndata:  Alert\r\n\r\\")]
+        [InlineData("data: T\r\ndata: Major\r\ndata:  Key\r\ndata:  Alert\r\n\r\\")]
+        public void ParseSSEMessageIncompleteParseResult(string encodedMessage)
+        {
+            var buffer = Encoding.UTF8.GetBytes(encodedMessage);
+            var readableBuffer = ReadableBuffer.Create(buffer);
+            var parser = new ServerSentEventsMessageParser();
+            var consumed = new ReadCursor();
+            var examined = new ReadCursor();
+
+            var parseResult = parser.ParseMessage(readableBuffer, out consumed, out examined, out Message message);
+
+            Assert.Equal(ServerSentEventsMessageParser.ParseResult.Incomplete, parseResult);
+        }
+    }
+}
--- a/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
@@ -78,16 +78,16 @@ namespace Microsoft.AspNetCore.SignalR.Tests
                 var receiveTcs = new TaskCompletionSource<string>();
                 connection.Received += (data, format) => receiveTcs.TrySetResult(Encoding.UTF8.GetString(data));
                 connection.Closed += e =>
+                {
+                    if (e != null)
                     {
-                        if (e != null)
-                        {
-                            receiveTcs.TrySetException(e);
-                        }
-                        else
-                        {
-                            receiveTcs.TrySetResult(null);
-                        }
-                    };
+                        receiveTcs.TrySetException(e);
+                    }
+                    else
+                    {
+                        receiveTcs.TrySetResult(null);
+                    }
+                };
 
                 await connection.StartAsync(transportType);
 
@@ -146,6 +146,7 @@ namespace Microsoft.AspNetCore.SignalR.Tests
             new[]
             {
                 new object[] { TransportType.WebSockets },
+                new object[] { TransportType.ServerSentEvents },
                 new object[] { TransportType.LongPolling }
             };
     }


mikaelm12:
Changing the parser to be stateless, so hold off for a second if you have any concerns there 😄

samples/ClientSample/RawSample.cs
```diff
@@ -39,7 +39,7 @@ internal class RawSample
             using (var httpClient = new HttpClient(new LoggingMessageHandler(loggerFactory, new HttpClientHandler())))
             {
                 logger.LogInformation("Connecting to {0}", baseUrl);
-                var transport = new LongPollingTransport(httpClient, loggerFactory);
```
moozzyk: I am getting rid of that

mikaelm12: I only changed it for testing. I'll undo before checking it in

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public class ServerSentEventsTransport : ITransport
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
```
moozzyk: It would be good to move it to a common place - now we have a copy for ServerSentEvents and LongPolling but we also should use this for negotiate request.

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+
+        private IChannelConnection<SendMessage, Message> _application;
+        private CancellationToken _cancellationToken = new CancellationToken();
+
+        public Task Running { get; private set; } = Task.CompletedTask;
+
+        public ServerSentEventsTransport(HttpClient httpClient)
+            : this(httpClient, null)
+        { }
```
moozzyk: nit: separate ctors with a line

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+        public ServerSentEventsTransport(HttpClient httpClient, ILoggerFactory loggerFactory)
```
moozzyk: We should throw for null httpclient (same in longpolling transport)

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        public Task StartAsync(Uri url, IChannelConnection<SendMessage, Message> application)
+        {
+            _logger.LogInformation("Starting {0}", nameof(ServerSentEventsTransport));
+
+            _application = application;
+            var sseUrl = Utils.AppendPath(url, "sse");
+            var sendUrl = Utils.AppendPath(url, "send");
+            var sendTask = SendMessages(sendUrl, _cancellationToken);
+            var openConnectionTask = OpenConnection(_application, sseUrl, _cancellationToken);
```
moozzyk: consider calling it `receiveTask` - this is our receive channel

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
+        private static readonly ProductInfoHeaderValue DefaultUserAgentHeader = ProductInfoHeaderValue.Parse(DefaultUserAgent);
+
+        private readonly HttpClient _httpClient;
+        private readonly ILogger _logger;
+        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+
+        private IChannelConnection<SendMessage, Message> _application;
+        private CancellationToken _cancellationToken = new CancellationToken();
```
moozzyk: nit: readonly

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+                {
+                    if (input.IsEmpty && result.IsCompleted)
+                    {
+                        _logger.LogDebug("Done reading from pipeline");
+                        break;
+                    }
+
+                    var parseResult = ServerSentEventsMessageParser.ParseMessage(input, out consumed, out examined, out var message);
+
+                    switch (parseResult)
```
moozzyk: Is this switch exhaustive? I think it would be more readable if it was converted to if statements - see next three comments.

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+                    {
+                        case ServerSentEventsMessageParser.ParsePhase.Completed:
+                            break;
+                        case ServerSentEventsMessageParser.ParsePhase.Incomplete:
+                            if (result.IsCompleted)
+                            {
+                                throw new FormatException("There was an error parsing");
+                            }
+
+                            continue;
```
moozzyk: I find it a bit unusual (continue from case).

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+                        case ServerSentEventsMessageParser.ParsePhase.Error:
+                            throw new FormatException("There was an error parsing");
+                    }
+
+                    _application.Output.TryWrite(message);
```
moozzyk: if the `switch` is exhaustive shouldn't this go to `ServerSentEventsMessageParser.ParsePhase.Completed:` case in which case the ` continue;` in `ServerSentEventsMessageParser.ParsePhase.Incomplete:` is not needed (and the switch statement is actually less cryptic)

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+                            break;
+                        case ServerSentEventsMessageParser.ParsePhase.Incomplete:
+                            if (result.IsCompleted)
+                            {
+                                throw new FormatException("There was an error parsing");
+                            }
+
+                            continue;
+                        case ServerSentEventsMessageParser.ParsePhase.Error:
+                            throw new FormatException("There was an error parsing");
```
moozzyk: nit: change message to something like: "Error parsing data from event stream"? it might be possible to merge the code and throw the exception from one place.

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+                    }
+
+                    _application.Output.TryWrite(message);
+                }
+                finally
+                {
+                    pipelineReader.Advance(consumed, examined);
+                }
+
```
moozzyk: nit: remove

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        public async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("Starting the send loop");
```
moozzyk: We will need to merge this later so that we have one copy for both server sent events and long polling.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+using System;
+using System.Buffers;
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
+{
+    public  class ServerSentEventsMessageParser
+    {
+
```
moozzyk: nit: remove empty line

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+        public static ParsePhase ParseMessage(ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined, out Message message)
+        {
+            const byte ByteCR = (byte)'\r';
+            const byte ByteLF = (byte)'\n';
+            const byte ByteSpace = (byte)' ';
+            bool firstLine = true;
```
moozzyk: nit: var

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+            byte[] data = null;
+            consumed = buffer.Start;
+            examined = buffer.End;
+            message = new Message();
+            MessageType messageType = MessageType.Text;
```
moozzyk: nit: `var`

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            var start = consumed;
+            var end = examined;
+
+            while (!reader.End)
+            {
+                var span = reader.Span;
+                var backup = reader;
+                var ch1 = reader.Take();
+                var ch2 = reader.Take();
```
moozzyk: can you do that if there was just one byte in the buffer (which was consumed by the first `.Take()`?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+                if (ch1 == ByteCR)
+                {
+                    if (ch2 == ByteLF)
+                    {
+                        consumed = buffer.End;
+                        message = new Message(data, messageType);
+                        return ParsePhase.Completed;
```
moozzyk: what is the data if the buffer only contains `\r\n` (which should be an invalid message/parse error)?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+                    }
+
+                    //When we are only missing the final \n
+                    return ParsePhase.Incomplete;
+                }
+
+                reader = backup;
```
moozzyk: why do we need `backup`?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+
+                if (firstLine)
+                {
+                    messageType = GetMessageType(line);
+                    start = lineEnd;
+                    firstLine = false;
+                    continue;
+                }
+
+                var nothing = buffer.End;
```
moozzyk: What does it do?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+                //Slice away the 'data: '
+                var newData = line.Slice(line.IndexOf(ByteSpace) + 1).ToArray();
```
moozzyk: how do you know you sliced `data: ` and not `foo: `

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.IO.Pipelines;
+
+namespace Microsoft.AspNetCore.Sockets.Common.Tests.Internal.Formatters
+{
+    public class ServerSentEventsParserTests
+    {
+        [Theory]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", MessageType.Text, "Hello, World")]
+        [InlineData("data: T\r\ndata: Hello\r\ndata: , World\r\n\r\n", MessageType.Text, "Hello, World")]
+        public async Task ParseSSEMessage(string encodedMessage, MessageType messageType, string expectedMessage)
```
moozzyk: You need more tests - especially negative cases

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
mikaelm12: Oops. Should be gone

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
mikaelm12: Yes. `ReadableBufferReader` returns -1 if you call take after you've reached the end of the internal buffer https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines/ReadableBufferReader.cs#L64-L70

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
anurse: Yeah, we should throw if the message does not contain at least a type line. For example, the following is invalid: `\r\n`, but this is a valid zero-byte message: `data: T\r\n\r\n`

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
anurse: After you've done this the simple way, by checking individual bytes, and writing tests for it, ping me if you want to try something clever and slightly faster ;). Kestrel does this kind of stuff a lot and there are some tricks you can use.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+                start = lineEnd;
+
+                if (data?.Length > 0)
+                {
+                    var tempData = new byte[data.Length + newData.Length];
```
anurse: This is going to do a lot of copying for multi-line messages. Ideally, we'd want to simply keep returning `Incomplete` until we know we've got the whole frame, so that we can avoid copying until the end. Though that may be an optimization we add after we've gotten this code in. + @davidfowl for thoughts

For now though, let's avoid some of the extra copying. Use a `List<byte[]>` to hold the payload segments, and just add each line's `data` buffer to it. Then at the end, copy each one into a final buffer.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                return buffer.First.Span;
+            }
+            return buffer.ToArray();
+        }
+
+        private static MessageType GetMessageType(ReadOnlySpan<byte> line)
+        {
+            //Skip the "data: " part of the line
+            var type = (char)line[6];
```
anurse: Check the length. If the line is invalid, we want to throw `FormatException` , not `IndexOutOfRangeException` ;).

anurse: Also, the following is invalid: `data: Tfoo`, so it needs to be **exactly** 7 bytes long (excluding `\r\n`)

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+                // Make sure LF is included in lineEnd
+                lineEnd = buffer.Move(lineEnd, 1);
+                var line = ConvertBufferToSpan(buffer.Slice(start, lineEnd));
+                reader.Skip(line.Length);
+
+                //Strip the \r\n from the span
+                line = line.Slice(0, line.Length - 2);
+
+                if (firstLine)
+                {
```
anurse: Need to check `data:` prefix here as well. Again, there's a clever way we could do the whole thing (checking `data:` **and** figuring out what the type is) if you've got some cycles after getting the simple method implemented and want to try something ;).

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+                case 'T':
+                    return MessageType.Text;
+                case 'B':
+                    return MessageType.Binary;
+                case 'C':
+                    return MessageType.Close;
+                case 'E':
+                    return MessageType.Error;
+                default:
+                    throw new FormatException("Invalid mesage type");
```
anurse: Make this message match the ones in the other parsers: https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/TextMessageParser.cs#L141

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+        public enum ParsePhase
```
anurse: Since this is returned from the Parse method, call it `ParseResult`.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -82,7 +83,10 @@ private static string MakeNewConnectionId()
 
         private static void Scan(object state)
         {
-            ((ConnectionManager)state).Scan();
+            if (!Debugger.IsAttached)
```
anurse: Yank this before you merge. Though @davidfowl , I'm beginning to think this would be useful to add to DEBUG builds... it's a pain to debug when connections keep getting cut out from under you ;)

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
anurse: Yeah, try to hit every branch in your parser with a test. See the tests for the existing formatters and parsers for some ideas of how to mess with the parser :)

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
anurse: Also, space between this and the closing `}` above.

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
mikaelm12: I was thinking that I'd do thatin a subsequent PR to avoid changing the LongPolling transport here and keep this focused on SSE.

moozzyk: sounds good to me

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
mikaelm12: I'll make that change in the LongPolling transport as well

mikaelm12:
Investigating the hang in the EndToEnd tests...

samples/ClientSample/RawSample.cs
```diff
@@ -26,7 +26,7 @@ internal class RawSample
                 args = args.Except(new[] { "--debug" }).ToArray();
             }
 
-            var baseUrl = "http://localhost:5000/chat";
```
anurse: Hmm, why did this change? It should be the default port when starting the SocketsSample up in Kestrel. IIS Express will use a random port, but using Kestrel should have a consistent port number.

Alternatively, we could just drop the default and require a URL on the command line :)

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+            var reader = new ReadableBufferReader(buffer);
+
+            var start = consumed;
+            var end = examined;
+
+            while (!reader.End)
+            {
+                if (ReadCursorOperations.Seek(start, end, out var lineEnd, ByteLF) == -1)
+                {
+                    if(end != buffer.End)
```
anurse: We never set `end`, that I can see. So this condition is never true.

Also, I think the condition and the `throw` are unnecessary. If we ever reach a point where we're looking for the end of a line and don't find it, we want to return `Incomplete`. We don't know, in this function, where the end of the stream is, so all we can do is say we haven't found a full message yet.

The caller will need to be the one who emits a failure of some kind if the stream ends and the last `ParseResult` was `Incomplete`.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                    {
+                        return ParseResult.Incomplete;
+                    }
+                    throw new FormatException("Expected a '\r\n' line ending");
+                }
+
+                lineEnd = buffer.Move(lineEnd, 1);
```
anurse: I assume `lineEnd` was positioned at the `\n`, so this is advancing past it? If so, throw a comment down here to clarify that :)

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                var line = ConvertBufferToSpan(buffer.Slice(start, lineEnd));
+                reader.Skip(line.Length);
+
+                if(!(line.Length > 1))
```
anurse: use `<= 1`, then you can remove the `!` and the extra `()`

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                        }
+                        break;
+                    case InternalParseState.ReadMessagePayload:
+                        if (!StartsWithDataPrefix(line))
+                        {
+                            throw new FormatException("Expected the message prefix 'data: '");
+                        }
+
+                        //Slice away the 'data: '
+                        var newData = line.Slice(line.IndexOf(ByteSpace) + 1).ToArray();
```
anurse: Thanks to `StartsWithDataPrefix` you shouldn't need `IndexOf`, you know it starts with `data: ` and can just use the number of characters directly.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                        break;
+                    case InternalParseState.ReadEndOfMessage:
+                        if (ReadCursorOperations.Seek(start, end, out lineEnd, ByteLF) == -1)
+                        {
+                            // The message has ended with \r\n\r
+                            return ParseResult.Incomplete;
+                        }
+
+                        //To check for the invalid case of there being more data after the frame end \r\n\r\n
+                        if (!reader.End)
```
anurse: There could be another frame in the buffer already, so we can't assume this. We shouldn't need this check at all anyway, all we need to do is set `consumed` to the end of the current frame and let the caller call us again.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                            //Find the final size of the payload
+                            var payloadSize = 0;
+                            foreach (var dataLine in _data)
+                            {
+                                payloadSize += dataLine.Length;
+                            }
+                            var payload = new byte[payloadSize];
+
+                            //Copy the contents of the data array to a single buffer
+                            var marker = 0;
```
anurse: rename to `offset`

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                                dataLine.CopyTo(payload, marker);
+                                marker += dataLine.Length;
+                            }
+
+                            message = new Message(payload, messageType);
+                        }
+                        else
+                        {
+                            //Empty message
+                            message = new Message(new byte[0], messageType);
```
anurse: use `Array.Empty<byte>`, it's cacherific!

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                            throw new FormatException("Expected the message prefix 'data: '");
+                        }
+
+                        //Slice away the 'data: '
+                        var newData = line.Slice(line.IndexOf(ByteSpace) + 1).ToArray();
+                        start = lineEnd;
+                        _data.Add(newData);
+
+                        //Peek into next byte. If it is a carriage return byte, then advance to the next state
+                        if (reader.Peek() == ByteCR)
```
anurse: Since both cases which exit the `switch` via `break` need to do this check, let's move it outside the `switch` entirely. The last case exits with `return` so it will skip it anyway.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,208 @@
...
+                        {
+                            //Empty message
+                            message = new Message(new byte[0], messageType);
+                        }
+
+                        consumed = buffer.End;
+                        return ParseResult.Completed;
+                }
+            }
+            throw new FormatException("Error parsing data from event stream");
```
anurse: This should just be `Incomplete`, right? We reached the end of our buffer but haven't seen a full frame yet.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -82,7 +82,7 @@ private static string MakeNewConnectionId()
 
         private static void Scan(object state)
         {
-            ((ConnectionManager)state).Scan();
+                ((ConnectionManager)state).Scan();
```
anurse: unindent ;)

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,87 @@
...
+            var result = Encoding.UTF8.GetString(message.Payload);
+            Assert.Equal(expectedMessage, result);
+        }
+
+
+        [Theory]
+        [InlineData("", MessageType.Text, "Error parsing data from event stream")]
+        [InlineData("data: X\r\n\r\n", MessageType.Text, "Unknown message type: 'X'")]
+        [InlineData("data: X\r\n", MessageType.Text, "Unknown message type: 'X'")]
+        [InlineData("data:", MessageType.Text, "Expected a '\r\n' line ending")]
```
anurse: This, as mentioned above in the implementation, isn't actually an error case. Any case where everything we've seen is **valid** but we just haven't seen enough to finish the frame is actually a case where we should return `Incomplete`, since we don't know if there's more data available.

Note the difference between the Incomplete and Invalid tests in the TextMessageParser: https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/TextMessageParserTests.cs#L84 - Feel free to ask me if you have questions about those tests or about what should be incomplete vs invalid in SSE.

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        public async Task ParseSSEMessageSuccessCases(string encodedMessage, MessageType messageType, string expectedMessage)
+        {
+            var stream = new MemoryStream();
+            var streamWriter = new StreamWriter(stream);
+            streamWriter.Write(encodedMessage);
+            streamWriter.Flush();
+            stream.Seek(0, SeekOrigin.Begin);
+
+            var pipelineReader = stream.AsPipelineReader();
+            var buffer = await pipelineReader.ReadToEndAsync();
```
anurse: Rather than using a real pipeline, you can just create a `ReadableBuffer` directly from an array using `ReadableBuffer.Create` and save some steps. See https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/TextMessageParserTests.cs#L27 (which uses `BytesReader`, which is a different thing, but the pattern should be about the same for `ReadableBuffer`)

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
mikaelm12: What I was going for here is partial message after we've reached the end of the buffer which should throw. I don't have any proper partial message tests in which we would expect the parser to return Incomplete.

anurse: Chatted in-person and clarified what I was talking about. We're on the same page now :)

mikaelm12: This fixes the issue I was seeing in the EndToEnd test.

mikaelm12:
Rebased on dev. Ready for review 😄

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,232 @@
...
+                _logger.LogDebug("Transport stopped. Exception: '{0}'", t.Exception?.InnerException);
+
+                _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
+                return t;
+            }).Unwrap();
+
+            return TaskCache.CompletedTask;
+        }
+
+        public async Task OpenConnection(IChannelConnection<SendMessage, Message> application, Uri url, CancellationToken cancellationToken)
```
BrennanConroy: Should this be private

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,232 @@
...
+
+            var request = new HttpRequestMessage(HttpMethod.Get, url);
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
+            var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+
+            var stream = await response.Content.ReadAsStreamAsync();
+
+            var pipelineReader = stream.AsPipelineReader();
+            try
+            {
```
BrennanConroy: formatting

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,232 @@
...
+                }
+            }
+            }
+            finally
+            {
+                _transportCts.Cancel();
+            }
+        }
+
+        public async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
```
BrennanConroy: Private

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,232 @@
...
+            finally
+            {
+                // Make sure the poll loop is terminated
+                _transportCts.Cancel();
+            }
+
+            _logger.LogInformation("Send loop stopped");
+        }
+
+        private async Task WriteMessagesAsync(IList<SendMessage> messages, PipelineTextOutput output, MessageFormat format)
```
BrennanConroy: Any reason we use IList and not List?

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,232 @@
...
+                var result = await pipelineReader.ReadAsync();
+                var input = result.Buffer;
+                var consumed = input.Start;
+                var examined = input.End;
+
+                try
+                {
+                    if (input.IsEmpty && result.IsCompleted)
+                    {
+                        _logger.LogDebug("Done reading from pipeline");
```
anurse: Rephrase as "Server-Sent Event Stream ended" (pipeline is just an internal concept here)

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,232 @@
...
+                    switch (parseResult)
+                    {
+                        case ServerSentEventsMessageParser.ParseResult.Completed:
+                            _application.Output.TryWrite(message);
+                            _parser.Reset();
+                            break;
+                        case ServerSentEventsMessageParser.ParseResult.Incomplete:
+                            if (result.IsCompleted)
+                            {
+                                throw new FormatException("Error parsing data from event stream");
```
anurse: Rephrase as "Incomplete message" to match https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs#L174

I'm not 100% happy with that message either, but we should be consistent. Maybe something more like "Stream ended with an incomplete message" would be more appropriate, but if we change it, let's change it everywhere :)

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
moozzyk: Shouldn't this actually be IEnumerable<SendMessage> ?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageFormatter.cs
```diff
@@ -14,11 +14,11 @@ public static class ServerSentEventsMessageFormatter
...
-        private const byte LineFeed = (byte)'\n';
-        private const char TextTypeFlag = 'T';
-        private const char BinaryTypeFlag = 'B';
-        private const char CloseTypeFlag = 'C';
-        private const char ErrorTypeFlag = 'E';
+        static readonly byte LineFeed = (byte)'\n';
+        static readonly char TextTypeFlag = 'T';
+        static readonly char BinaryTypeFlag = 'B';
+        static readonly char CloseTypeFlag = 'C';
+        static readonly char ErrorTypeFlag = 'E';
```
anurse: Add `private` to these. We always put accessibility/visibility modifiers on, even when they are the default.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
+{
+    public class ServerSentEventsMessageParser
+    {
+        static readonly byte ByteCR = (byte)'\r';
+        static readonly byte ByteLF = (byte)'\n';
```
anurse: Add `private` to these. We always put accessibility/visibility modifiers on, even when they are the default.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+                    //Partial message. We need to read more.
+                    return ParseResult.Incomplete;
+                }
+
+                lineEnd = buffer.Move(lineEnd, 1);
+                var line = ConvertBufferToSpan(buffer.Slice(start, lineEnd));
+                reader.Skip(line.Length);
+
+                //Check for a misplaced '\n'
+                if (line.Length <= 1 || line[line.Length - 2] != ByteCR)
```
anurse: Split these conditions and give more descriptive error messages for each case. It's always good to have clearer errors :).

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+                //Strip the \r\n from the span
+                line = line.Slice(0, line.Length - 2);
+
+                switch (_internalParserState)
+                {
+                    case InternalParseState.ReadMessageType:
+                        messageType = GetMessageType(line);
+                        start = lineEnd;
+                        _internalParserState = InternalParseState.ReadMessagePayload;
+                        consumed = lineEnd;
```
anurse: nit: move this to be along side the `start = ` line. I just like to see things like that together. I don't mind which side of the `_internalParserState` line it goes, just together

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+                var line = ConvertBufferToSpan(buffer.Slice(start, lineEnd));
+                reader.Skip(line.Length);
+
+                //Check for a misplaced '\n'
+                if (line.Length <= 1 || line[line.Length - 2] != ByteCR)
+                {
+                    throw new FormatException("There was an issue with the frame format");
+                }
+
+                //Strip the \r\n from the span
```
BrennanConroy: Use `Remove` in the comment to avoid Policheck, also I think a space after `//` is the convention

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+                        {
+                            throw new FormatException("Expected the message prefix 'data: '");
+                        }
+
+                        //Slice away the 'data: '
+                        var newData = line.Slice(6).ToArray();
+
+                        start = lineEnd;
+                        _data.Add(newData);
+                        consumed = lineEnd;
```
anurse: Same nit about line order as above. Just like to see similar assignments together :).

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+                        break;
+                    case InternalParseState.ReadEndOfMessage:
+                        if (ReadCursorOperations.Seek(start, end, out lineEnd, ByteLF) == -1)
+                        {
+                            // The message has ended with \r\n\r
```
anurse: Comment looks wrong ;)

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+
+                            message = new Message(payload, messageType);
+                        }
+                        else
+                        {
+                            //Empty message
+                            message = new Message(Array.Empty<byte>(), messageType);
+                        }
+
+                        consumed = buffer.End;
```
anurse: `consumed = lineEnd` right? You may not have reached the end of the entire buffer.

A good test for this would be to take a buffer with something like `data: T\r\ndata: Foo\r\n\r\ndata: T`, call Parse once and make sure that you haven't consumed the start of the second frame. I think with this code, you will.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        public void Reset()
+        {
+            _internalParserState = InternalParseState.ReadMessageType;
+            _data.Clear();
+        }
+
+        private bool StartsWithDataPrefix(ReadOnlySpan<byte> line)
+        {
+            var dataPrefix = Encoding.UTF8.GetBytes("data: ");
```
BrennanConroy: Can this be cached

anurse: Make this buffer static, it doesn't change :). You can't make the Span static though, since Span can't go on the heap

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
anurse: No, use `List<T>`. For private code, it's best to move the concrete `List<T>` around so that when you a) avoid interface dispatch and b) when you enumerate you can use the `struct` enumerator that `List<T>` provides, rather than boxing it.

src/Microsoft.AspNetCore.Sockets.Client/DefaultTransportFactory.cs
```diff
@@ -39,7 +39,7 @@ public ITransport CreateTransport(TransportType availableServerTransports)
 
             if ((availableServerTransports & TransportType.ServerSentEvents & _requestedTransportType) == TransportType.ServerSentEvents)
             {
-                throw new NotImplementedException();
+                return new ServerSentEventsTransport(_httpClient, _loggerFactory);
```
moozzyk: Add SSE to DefaultFactoryTests

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,232 @@
...
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
+        private static readonly ProductInfoHeaderValue DefaultUserAgentHeader = ProductInfoHeaderValue.Parse(DefaultUserAgent);
+
+        private readonly HttpClient _httpClient;
+        private readonly ILogger _logger;
+        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+
+        private IChannelConnection<SendMessage, Message> _application;
+        private ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
```
moozzyk: readonly

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageFormatter.cs
moozzyk: Why were they changed from const to static readonly?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
moozzyk: why not `const`s?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+
+namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
+{
+    public class ServerSentEventsMessageParser
+    {
+        static readonly byte ByteCR = (byte)'\r';
+        static readonly byte ByteLF = (byte)'\n';
+
+        private InternalParseState _internalParserState = InternalParseState.ReadMessageType;
+        private IList<byte[]> _data = new List<byte[]>();
```
moozzyk: readonly?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        public ParseResult ParseMessage(ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined, out Message message)
+        { 
+            consumed = buffer.Start;
+            examined = buffer.End;
+            message = new Message();
+            var messageType = MessageType.Text;
```
moozzyk: I think messageType should be declared in nested scope not here (it is not used here). Also any reason for assigningthe Message.Text value?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+            if (line.Length != 7)
+            {
+                throw new FormatException("There was an error parsing the message type");
+            }
+
+            //Skip the "data: " part of the line
+            var type = (char)line[6];
+            switch (type)
+            {
+                case 'T':
```
moozzyk: Did you intend to use TextTypeFlag etc. and therefore made them static readonly?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        public enum ParseResult
+        {
+            Completed,
+            Incomplete,
+        }
+
+        private enum InternalParseState
+        {
+            Initial,
```
moozzyk: This state does not seem to be used at all

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,193 @@
...
+                {
+                    case InternalParseState.ReadMessageType:
+                        messageType = GetMessageType(line);
+                        start = lineEnd;
+                        _internalParserState = InternalParseState.ReadMessagePayload;
+                        consumed = lineEnd;
+
+                        break;
+                    case InternalParseState.ReadMessagePayload:
+                        if (!StartsWithDataPrefix(line))
```
moozzyk: You only seem to use  `StartsWithDataPrefix` to throw the exception. Change the method name to EnsureStartsWithDataPrefix and throw directly from this method to avoid throwing different exception for the same condition.

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.IO.Pipelines;
+using System.Text;
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Sockets.Common.Tests.Internal.Formatters
+{
+    public class ServerSentEventsParserTests
+    {
+        [Theory]
```
moozzyk: Add some test cases where you receive partial data and you need to re-enter parsing e.g (square brackets 
"d", "ata: T\r\n", "\r\n"
"data: T\r\n", "\r\n"
"data: T\r\ndata: data\r\n", "\r\n"

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
```
moozzyk: Add cases for standalone `\r` in data
Add a case where the message is "data: data: \r\n"

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: Update existing End2End Tests and DefaultFactory tests to use SSE transport

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,83 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
moozzyk: The only tests you have is for parsing but not for the transport itself. Add ServerSentEventsTransport tests similar to: https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageFormatter.cs
mikaelm12: I initially wanted to use these in the SSE parser as well and was told to make them static readonly. We use bytes in the parser though chars here.

mikaelm12:
Replaced by https://github.com/aspnet/SignalR/pull/401 😄

PR #360, opened 2017-04-01
Fix issue where multiple calls to dispose don't wait properly

### LINKED ISSUE #347: Investigate ungraceful WebSocket connection close on exit
Currently on close with a WebSocket connection open you can get the following error:
```
fail: Microsoft.AspNetCore.Sockets.ConnectionManager[0]
      Failed disposing connection 2511b1b2-1dbb-44ee-805c-e6d629fca076
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'FrameResponseStream'.
   at System.IO.Pipelines.PipeCompletion.ThrowFailed()
   at System.IO.Pipelines.Pipe.System.IO.Pipelines.IWritableBufferAwaiter.GetResult()
   at Microsoft.Extensions.WebSockets.Internal.WebSocketConnection.<SendCoreLockAcquiredAsync>d__38`1.MoveNext() in C:\Github\SignalR\src\Microsoft.Extensions.WebSockets.Internal\WebSocketConnection.cs:line 695
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Microsoft.Extensions.WebSockets.Internal.WebSocketConnection.<CloseAsync>d__24.MoveNext() in C:\Github\SignalR\src\Microsoft.Extensions.WebSockets.Internal\WebSocketConnection.cs:line 214
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
```

davidfowl:
- DisposeAsync returned immediately to anyone but the first caller.
This means that it was possible to end the request before properly
waiting on the transport task which means writing after dispose was possible.
- Added a test

Fixes #347

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
+++ b/src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
@@ -38,8 +38,7 @@ namespace Microsoft.AspNetCore.Sockets.Internal
 
         public async Task DisposeAsync()
         {
-            Task applicationTask = TaskCache.CompletedTask;
-            Task transportTask = TaskCache.CompletedTask;
+            Task disposeTask = TaskCache.CompletedTask;
 
             try
             {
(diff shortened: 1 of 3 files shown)


davidfowl:
Some tests are hanging. Will fix

davidfowl:
@anurse I don't think these tests ever really worked properly. Now that the connection waits on the transport to complete, the `TestWebSocketConnectionFeature` closes the websocket but never receives a close frame so it stays open forever.

src/Microsoft.AspNetCore.Sockets/Internal/ConnectionState.cs
```diff
@@ -75,6 +80,9 @@ public ConnectionState(Connection connection, IChannelConnection<Message> applic
 
             // REVIEW: Add a timeout so we don't wait forever
             await Task.WhenAll(applicationTask, transportTask);
```
BrennanConroy: Couldn't we just move the `Task.WhenAll` inside the `Lock` and it would achieve the same thing without creating a tcs?

davidfowl: You don't mean move, you mean, add another set of waiting. The only downside is that we'd need to replicate a bunch of logic around null checking.

BrennanConroy: No, I meant move this line to right after 69, then anyone calling `DisposeAsync` a second or more times would be on `await Lock.WaitAsync()` and when the `applicationTask` and `transportTask` finish the lock gets released and the other `DisposeAsync` callers finish

BrennanConroy: Ignore me

PR #363, opened 2017-04-03
React to CoreFx byte[] slice

BrennanConroy:
https://github.com/aspnet/SignalR/issues/362


moozzyk:
I guess we need to take it so just merge it.

Issue #361, opened 2017-04-03
Add timeouts to server side waiting logic


davidfowl:
There are a bunch of places we wait on tasks to complete where we should be timing out:

1. The transport completes and the application is running forever.
2. The application completes and the transport is running forever (only happens with websockets).

The first is likely an application bug. We closed the transport either gracefully or ungracefully and have told the application this much but that signalr is being ignored.

The second is our bug since the application told us that it was done sending data but we don't close because nothing is timing out the receive loop.

We *may* not want to tackle 1. because the user should just make their application work. 2 on the other hand is our issue.

/cc @anurse @BrennanConroy

anurse:
There's a more nuanced side to 1 in WebSockets. "The Transport Completes" indicates that both ends have closed. WebSockets allows for a half-close, where one side sends a close frame, indicating that it's finished **sending**, but it still expects to read data **received** from the other side.

In general, in WebSockets, sending a Close frame gives no guarantee as to when or even if a Close frame will come from the other party. EndPoints should be able to support this behavior if they want to.

I am 100% in favor of a system that aborts the socket after the EndPoint returns from `OnConnectionAsync` (after a certain timeout), because then the EndPoint has **also** indicated that it no longer wishes to **read** data. But simply closing the outbound channel does not indicate that the connection is complete and the socket should be terminated.

davidfowl:
Discussed with @anurse and we agree this is a super edge case. We won't be adding an application timeout either.

PR #367, opened 2017-04-04
Added timeout tests

davidfowl:
This is to start a discussion about if we should do dispose timeouts at all. There are 2 places right now where we can time out existing connections.

- The transport is still running after the application closes
- The application is still running after the transport closes

In both cases, we're stuck in dispose async forever. We need to do something about the transport for sure since that's not user code running and as far as the user is concerned, their application code has ended and that should signal to the transport to close. This works well for everything but websockets as it waits forever today. I'll probably just implement that one first.

#361

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
+++ b/src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
@@ -14,9 +14,4 @@ namespace Microsoft.AspNetCore.Sockets
 
         public WebSocketOptions WebSockets { get; } = new WebSocketOptions();
     }
-
-    public class WebSocketOptions
-    {
-        public TimeSpan CloseTimeout { get; set; } = TimeSpan.FromSeconds(5);
-    }
 }
--- a/src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
+++ b/src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Authorization;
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
```diff
@@ -107,9 +114,21 @@ public WebSocketsTransport(IChannelConnection<Message> application, ILoggerFacto
...
 
-                // Wait for the client to close.
-                // TODO: Consider timing out here and cancelling the receive loop.
-                await receiving;
+                // Wait for the client to close or wait for the close timeout
+                var resultTask = await Task.WhenAny(receiving, Task.Delay(_options.CloseTimeout));
+
+                // We timed out waiting for the transport to close so abort the connection so we don't attempt to write anything else
+                if (resultTask != receiving)
+                {
```
moozzyk: Add logging

src/Microsoft.AspNetCore.Sockets/EndPointOptions.cs
```diff
@@ -10,5 +11,12 @@ public class EndPointOptions<TEndPoint> where TEndPoint : EndPoint
         public AuthorizationPolicy Policy { get; set; }
 
         public TransportType Transports { get; set; } = TransportType.All;
+
+        public WebSocketOptions WebSockets { get; } = new WebSocketOptions();
+    }
+
+    public class WebSocketOptions
```
moozzyk: move to a separate file

davidfowl: Forgot I added this here

muratg:
Why is Travis taking so long?

davidfowl:
@moozzyk OSX took longer for some reason. Nothing to do with this change.

moozzyk:
@muratg @davidfowl - it's "normal" that OSX takes longer.

PR #370, opened 2017-04-04
Adding transport factory

moozzyk:


TestTransportFactory.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.AspNetCore.Sockets.Client.Tests;
```
moozzyk: Missing file header (I added that comment before @mikaelm12)

mikaelm12: 😄

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -46,11 +47,16 @@ public Connection(Uri url, ILoggerFactory loggerFactory)
             _logger = _loggerFactory.CreateLogger<Connection>();
         }
 
-        public Task StartAsync() => StartAsync(transport: null, httpClient: null);
-        public Task StartAsync(HttpClient httpClient) => StartAsync(transport: null, httpClient: httpClient);
-        public Task StartAsync(ITransport transport) => StartAsync(transport: transport, httpClient: null);
+        public Task StartAsync() => StartAsync(transportFactory: null, httpClient: null);
```
davidfowl: Overload hell. We need to do an API review

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -44,20 +45,19 @@ public void CannotCreateHubConnectionWithNullUrl()
                 .Returns<HttpRequestMessage, CancellationToken>(async (request, cancellationToken) =>
                 {
                     await Task.Yield();
-                    return ResponseUtils.CreateResponse(HttpStatusCode.OK);
```
mikaelm12: Why did you change this?

samples/ClientSample/HubSample.cs
```diff
@@ -26,44 +24,40 @@ internal class HubSample
...
-                    logger.LogInformation("Connected to {0}", baseUrl);
-
-                    var cts = new CancellationTokenSource();
-                    Console.CancelKeyPress += (sender, a) =>
-                    {
-                        a.Cancel = true;
-                        logger.LogInformation("Stopping loops...");
-                        cts.Cancel();
-                    };
+                await connection.StartAsync();
```
mikaelm12: Do we want this to still use LongPolling here? the TransportFactory uses WebSockets by default? 
I also think it would be good to show usage of the enum in the sample.

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
moozzyk: `StartAsync`
`StartAsyncWithTransportType`
`StartAsyncWithTransportTypeAndHttpClient`
`StartAsyncWithTransportFactory`
`StartAsyncWithTransportFactoryAndHttpClient`

:D

samples/ClientSample/HubSample.cs
moozzyk: Possibly - unfortunately this sample can't run because of authorization - we need to figure this out.

src/Microsoft.AspNetCore.Sockets.Client/DefaultTransportFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            if (requestedTransportType <= 0 || requestedTransportType > TransportType.All)
+            {
+                throw new ArgumentOutOfRangeException(nameof(requestedTransportType));
+            }
+
+            _requestedTransportType = requestedTransportType;
+        }
+
+        public ITransport CreateTransport(TransportType availableServerTransports, ILoggerFactory loggerFactory, HttpClient httpClient)
```
moozzyk: No. Note that this code is not being called by the user. This is user's code being invoked from SignalR and we provide you all the things we might create and you might need to create the transport. You don't have to use them as long as you give us a transport (or throw)

src/Microsoft.AspNetCore.Sockets.Client/ITransportFactory.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using System.Net.Http;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public interface ITransportFactory
+    {
+        ITransport CreateTransport(TransportType availableServerTransports, ILoggerFactory loggerFactory, HttpClient httpClient);
```
moozzyk: why? I don't think it matters... This is code invoked by SignalR and not the user.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
moozzyk: https://msdn.microsoft.com/en-us/library/system.net.transporttype(v=vs.110).aspx

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestTransportFactory.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using System.Net.Http;
```
moozzyk: Looks like my VS code formatting settings were lost...

moozzyk:
:up::date:

src/Microsoft.AspNetCore.Sockets.Client/ITransportFactory.cs
mikaelm12: This is more of a preference thing. Seems more natural for the logger to go last IMO, but that's just me

davidfowl: The logger factory should not be part of the contract for making a transport. It's something that can be passed into the constructor.

moozzyk: OK

moozzyk: Removed ILoggerFactory from the interface

src/Microsoft.AspNetCore.Sockets.Client/Connection.cs
```diff
@@ -59,7 +65,13 @@ public Task StartAsync(ITransport transport, HttpClient httpClient)
...
             }
 
-            StartAsyncInternal(transport, httpClient)
+            if (httpClient == null)
+            {
+                // We are creating the client so store it to be able to dispose
+                _httpClient = httpClient = new HttpClient();
+            }
+
+            StartAsyncInternal(transportFactory ?? new DefaultTransportFactory(TransportType.All, _loggerFactory), httpClient)
```
davidfowl: Why do we use ContinueWith anywhere on the client?

moozzyk: tcs makes it easy to avoid race - https://github.com/aspnet/SignalR/commit/23375c911b4659d171bf8532b824c32771a54cb1 contains full details

moozzyk: I guess it could be converted to try/catch/finally inside StartAsyncInternal but it's not this change.

davidfowl: Yea. We should clean this up. I'm anti continue with.

src/Microsoft.AspNetCore.Sockets.Client/DefaultTransportFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public class DefaultTransportFactory : ITransportFactory
+    {
+        private readonly TransportType _requestedTransportType;
+        private readonly ILoggerFactory _loggerFactory;
+
+
```
davidfowl: nit: extra space

src/Microsoft.AspNetCore.Sockets.Client/IConnection.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public interface IConnection
     {
-        Task StartAsync(ITransport transport, HttpClient httpClient);
+        Task StartAsync(ITransportFactory transportFactory, HttpClient httpClient);
```
davidfowl: Why did we make HttpClient part of the interface instead of a ctor parameter? The interface doesn't need to be tied to a particular http client implementation does it?

src/Microsoft.AspNetCore.Sockets.Client/ITransportFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Net.Http;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public interface ITransportFactory
+    {
+        ITransport CreateTransport(TransportType availableServerTransports, HttpClient httpClient);
```
davidfowl: Why does this need the HttpClient in the method?

moozzyk: Because it would require always providing the HttpClient to Start method or we would use different http clients. Right now you can just provide the factory and if you don't provide http client we will give you the one we created so that the same client is being used across the stack.

src/Microsoft.AspNetCore.Sockets.Client/IConnection.cs
moozzyk: I started without http client and it did not look nice either because of HubConnection. The alternatives are - creating and tracking HttpClient in hubConnection when you use the overload that takes enum. Adding an overload of IConnection.StartAsync that takes transport type in addition to the one that takes factory and the HttpClient. Not having it as part of this interface also means that you basically always need to create your own HttpClient when using the transport factory and pass as the parameter to StartAsync instead of letting us create the HttpClient and give it to you. It seems that it basically always leaks somewhere.

moozzyk:
:up::date: - removed HttpClient from the `ITransportFactory` interface

PR #372, opened 2017-04-05
SignalR Hub Protocol

anurse:
Here's the first pass at the full SignalR Hub Protocol document.

/fyi @DamianEdwards

### DIFF
--- a/specs/HubProtocol.md
+++ b/specs/HubProtocol.md
@@ -52,6 +52,10 @@ On the Caller side, the user code which performs the invocation indicates how it
 
 Errors are indicated by the presence of the `error` field in a `Completion` message. Errors always indicate the immediate end of the invocation. In the case of streamed responses, the arrival of a `Completion` message indicating an error should **not** stop the dispatching of previously-received results. The error is only yielded after the previously-received results have been dispatched.
 
+## Completion and Results
+
+A Invocation is only considered completed when the `Completion` message is recevied. Receiving **any** message using the same `Invocation ID` after a `Completion` message has been received for that invocation is considered a protocol error and the recipient should immediately terminate the connection. For non-streamed results, it is up to the Callee whether to encode the result in a separate `Result` message, or within the `Completion` message.
+
 ## Examples
 
 Consider the following C# methods
@@ -75,8 +79,8 @@ public IEnumerable<int> Batched(int count)
     }
 }
 
-[return: StreamToClient] // This is a made-up attribute that is used to indicate to the .NET Binder that it should stream results
-public IEnumerable<int> Stream(int x, int y)
+[return: Streamed] // This is a made-up attribute that is used to indicate to the .NET Binder that it should stream results
+public IEnumerable<int> Stream(int count)
 {
     for(var i = 0; i < count; i++)
     {
@@ -84,8 +88,8 @@ public IEnumerable<int> Stream(int x, int y)
     }
 }
 
-[return: StreamToClient] // This is a made-up attribute that is used to indicate to the .NET Binder that it should stream results
-public IEnumerable<int> StreamFailure(int x, int y)
+[return: Streamed] // This is a made-up attribute that is used to indicate to the .NET Binder that it should stream results
+public IEnumerable<int> StreamFailure(int count)
 {
     for(var i = 0; i < count; i++)
     {
@@ -102,6 +106,14 @@ In each of the below examples, lines starting `C->S` indicate messages sent from
 ```
 C->S: Invocation { Id = 42, Target = "Add", Arguments = [ 40, 2 ] }
 S->C: Completion { Id = 42, Result = 42 }
+```
+
+The following is also acceptable, since the Callee may choose to encode the `Result` in a separate message
+
+```
+C->S: Invocation { Id = 42, Target = "Add", Arguments = [ 40, 2 ] }
+S->C: Result { Id = 42, Result = 42 }
+S->C: Completion { Id = 42 }
 ```
 
 ### Single Result with Error (`SingleResultFailure` example above)
--- a/specs/HubProtocol.md
+++ b/specs/HubProtocol.md
@@ -1,0 +1,225 @@
+# SignalR Hub Protocol
+
+The SignalR Protocol is a protocol for two-way RPC over any Message-based transport. Either party in the connection may invoke procedures on the other party, and procedures can return zero or more results or an error.
+
+## Terms
+
+* Caller - The node that is issuing an `Invocation` message and receiving `Result` messages (a node can be both Caller and Callee for different invocations simultaneously)
+* Callee - The node that is receiving an `Invocation` message and issuing `Result` messages (a node can be both Callee and Caller for different invocations simultaneously)
+* Binder - The component on each node that handles mapping `Invocation` messages to method calls and return values to `Result` messages
+
+## Transport Requirements
+
+The SignalR Protocol requires the following attributes from the underlying transport. The protocol was primarily designed for use with WebSockets, though it is relatively straightforward to build an adaptor layer for a different transport.
+
+* Message-based (aka Datagram, as opposed to Streaming)
+* A distinction between Text and Binary frames
+* Reliable, in-order, delivery of messages - Specifically, the SignalR protocol provides no facility for retransmission or reordering of messages. If that is important to an application scenario, the application must either use a transport that guarantees it (i.e. TCP) or provide their own system for managing message order.
+
+## Overview
+
+There are two encodings of the SignalR protocol: [JSON](http://www.json.org/) and [Protocol Buffers](https://developers.google.com/protocol-buffers/). Only one format can be used for the duration of a connection, and the format must be negotiated in advance (i.e. using a QueryString value, Header, or other indicator). However, each format shares a similar overall structure.
+
+In the SignalR protocol, the following types of messages can be sent:
+
+* `Invocation` Message - Indicates a request to invoke a particular method (the Target) with provided Arguments on the remote endpoint.
+* `Result` Message - Indicates results from a previous Invocation message.
+* `Completion` Message - Indicates a previous Invocation has completed, and no further Result messages will be received.
+
+In order to perform a single invocation, Caller follows the following basic flow:
+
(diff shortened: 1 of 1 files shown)


specs/HubProtocol.md
```diff
@@ -0,0 +1,225 @@
...
+    "type": 2,
+    "invocationId": 123,
+    "error": "An error occurred!",
+    "payload": 42
+}
+```
+
+### Completion Message Encoding
+
+A `Completion` message is a JSON object with the following properties
```
moozzyk: Is this message required after an error? I propose that errors terminate the invocation and as a result should be on the Completion message and not on the Result message. Also, Completion should also have the payload property to optimize for non-streaming cases (which I think will be the majority) to prevent from sending more than one message.

specs/HubProtocol.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Protocol Buffers (ProtoBuf) Encoding
+
+In order to support ProtoBuf, an application must provide a [ProtoBuf service definition](https://developers.google.com/protocol-buffers/docs/proto3) for the Hub. However, implementations may automatically generate these definitions from reflection information, if the underlying platform supports this. For example, the .NET implementation will attempt to generate service definitions for methods that use only simple primitive and enumerated types. The service definition provides a description of how to encode the arguments and return value for the call. For example, consider the following C# method:
```
moozzyk: I am missing details about how things should work with types not supported out of the box.

specs/HubProtocol.md
```diff
@@ -0,0 +1,225 @@
...
+
+The SignalR Protocol requires the following attributes from the underlying transport. The protocol was primarily designed for use with WebSockets, though it is relatively straightforward to build an adaptor layer for a different transport.
+
+* Message-based (aka Datagram, as opposed to Streaming)
+* A distinction between Text and Binary frames
+* Reliable, in-order, delivery of messages - Specifically, the SignalR protocol provides no facility for retransmission or reordering of messages. If that is important to an application scenario, the application must either use a transport that guarantees it (i.e. TCP) or provide their own system for managing message order.
+
+## Overview
+
+There are two encodings of the SignalR protocol: [JSON](http://www.json.org/) and [Protocol Buffers](https://developers.google.com/protocol-buffers/). Only one format can be used for the duration of a connection, and the format must be negotiated in advance (i.e. using a QueryString value, Header, or other indicator). However, each format shares a similar overall structure.
```
anurse: I don't think that's the plan any more, but it's just from a random chat with @davidfowl , we can clear it up tomorrow. My understanding is that we have a SignalR Protocol, with two encodings (Text/JSON, and Binary/ProtoBuf), and that user-provided protocols are not intended to be supported. Sockets is the layer for you if you need a custom protocol.

specs/HubProtocol.md
```diff
@@ -0,0 +1,225 @@
...
+
+In the SignalR protocol, the following types of messages can be sent:
+
+* `Invocation` Message - Indicates a request to invoke a particular method (the Target) with provided Arguments on the remote endpoint.
+* `Result` Message - Indicates results from a previous Invocation message.
+* `Completion` Message - Indicates a previous Invocation has completed, and no further Result messages will be received.
```
anurse: It has no result. Yes, the Callee must send two messages. This is by design to handle things like `IEnumerable`, `IObservable`, etc. results. If we used a "fin" flag, we'd have to buffer one result when enumerating in order to see if it's the last one in order to set the "fin" flag.

specs/HubProtocol.md
```diff
@@ -0,0 +1,225 @@
...
+
+**NOTE**: `Invocation ID`s are arbitrarily chosen by the Caller and the Callee is expected to use the same string in all response messages. Callees may establish reasonable limits on `Invocation ID` lengths and terminate the connection when an `Invocation ID` that is too long is received.
+
+## Multiple Results
+
+The SignalR protocol allows for multiple `Result` messages to be transmitted in response to an `Invocation` message, and allows the receiver to dispatch these results as they arrive, to allow for streaming data from one endpoint to another.
+
+On the Callee side, it is up to the Callee's Binder to determine if a method call will yield multiple results. For example, in .NET, a return type of `IEnumerable` or `IObservable` will yield multiple results, but other return types will yield a single result. The Callee's Binder must encode each result in separate `Result` messages, indicating the end of results by sending a `Completion` message.
```
anurse: There's a lot in here that is up to the Binder, which is the language-specific piece that maps from language concepts to SignalR protocol concepts. For example, we expect the .NET Binder to consider `IEnumerable<T>` returns as a single value (I'll update the spec), partly for this reason. We may end up using an attribute or something to indicate a fully streamed result.

This spec doesn't technically cover the Binder, it just provides examples from the .NET Binder.

anurse: I've clarified this in the text in the next update.

specs/HubProtocol.md
```diff
@@ -0,0 +1,225 @@
...
+
+On the Caller side, the user code which performs the invocation indicates how it would like to receive the results and it is up the Caller's Binder to determine how to handle the result. If the Caller expects only a single result, but multiple results are returned, the Caller's Binder should yield an error indicating that multiple results were returned. However, if a Caller expects multiple results, but only a single result is returned, the Caller's Binder should yield that single result and indicate there are no further results.
```
anurse: It depends how the result is bound on the client side. In .NET, if you bind to a `Task<T>`, then the `Task` will not be completed until the `Completion` message arrives, at which point we know what the result should be (error or payload). If you bind an `IObservable<T>`, that already has an `OnError` and `OnNext` callback model for yielding results.

specs/HubProtocol.md
moozzyk: OK. We could keep it closed for now and open later if needed. I don't really expect a lot of people asking for it but we will already have 2 implementations so we will probably have a design where things won't be hardcoded and therefore it should be easy to make them customizable if needed.

specs/HubProtocol.md
anurse: Just re-read and noticed this was about expecting a single result and receiving multiple results. Since the caller doesn't yield until the `Completion` message arrives, it knows that multiple results were received but only one was expected and can yield the error then. It could even yield the error as soon as the second `Result` message arrives (since it knows it wasn't expecting it).

Receiving an unsolicited `Result` message for an invocation already completed would be a protocol error and we'd terminate the connection.

specs/HubProtocol.md
anurse: It's not covered here. The spec just expects the binder to have a full ProtoBuf definition for the service, as described below. Our binder will likely have a system for generating these service definitions from arbitrary C# methods (based on the suggested type mappings below, etc.), but that's not part of the protocol.

specs/HubProtocol.md
```diff
@@ -0,0 +1,225 @@
...
+syntax = "proto3";
+
+message SendMessageToUserRequest {
+    string userName = 1;
+    string message = 2;
+}
+
+message SendMessageToUserResponse {
+    bool result = 1;
+}
```
anurse: This is missing the `service` definition, adding it.

specs/HubProtocol.md
anurse: Yeah, I've considered that too. I'm OK with either. I like the simple single-purpose message types for their simplicity, but agree that errors should immediately terminate the call, and that the majority of cases will have a single return value.

specs/HubProtocol.md
anurse: Yep, I do that in the current implementation PR. There's an `IHubProtocol` interface.

specs/HubProtocol.md
moozzyk: I think I worded this incorrectly. I was not missing the details but merely a mention. It is kind of implied in the table below in the line for 'custom `struct` or `class`'. I just think one sentence saying that the user will have to do additional work for types not supported out of the box should be enough.

anurse: Gotcha. That makes sense

specs/HubProtocol.md
moozzyk: I think having just one message for single results will be less error prone and prevent from cases where result was received but the callback never finished because the Completion message never arrived. To be clear we would still have both Result and Completion messages but Result messages will not be used for singleton results.

anurse:
:up: :date:

specs/HubProtocol.md
```diff
@@ -45,10 +44,119 @@ The `Target` of an `Invocation` message must refer to a specific method, overloa
...
 
 On the Caller side, the user code which performs the invocation indicates how it would like to receive the results and it is up the Caller's Binder to determine how to handle the result. If the Caller expects only a single result, but multiple results are returned, the Caller's Binder should yield an error indicating that multiple results were returned. However, if a Caller expects multiple results, but only a single result is returned, the Caller's Binder should yield that single result and indicate there are no further results.
 
+## Errors
+
+Errors are indicated by the presence of the `error` field in a `Completion` message. Errors always indicate the immediate end of the invocation. In the case of streamed responses, the arrival of a `Completion` message indicating an error should **not** stop the dispatching of previously-received results. The error is only yielded after the previously-received results have been dispatched.
+
```
moozzyk: Should you mention that receiving a Result message after the Completion message for a given Id is a protocol violation?

specs/HubProtocol.md
```diff
@@ -0,0 +1,351 @@
...
+public IEnumerable<int> Batched(int count)
+{
+    for(var i = 0; i < count; i++)
+    {
+        yield return i;
+    }
+}
+
+[return: StreamToClient] // This is a made-up attribute that is used to indicate to the .NET Binder that it should stream results
```
davidfowl: `[Streaming]`

specs/HubProtocol.md
```diff
@@ -45,10 +44,119 @@ The `Target` of an `Invocation` message must refer to a specific method, overloa
...
+public IEnumerable<int> Stream(int x, int y)
```
anurse: Yep, copy-🍝

specs/HubProtocol.md
anurse: I'd prefer `Streamed`, since this is a declarative attribute on the return value. Also, we can nit-pick the attribute later, it's not really part of the spec.

anurse:
:up: :date:

PR #374, opened 2017-04-05
Implement new Hub Protocol

### LINKED ISSUE #373: Implement new SignalR Hub Protocol
This issue tracks implementing the SignalR Hub Protocol described in https://github.com/aspnet/SignalR/pull/372

This issue ONLY tracks implementing the new protocol in the existing code, not things like `IObservable` or `Channel` results.

anurse:
Fixes #373 

Note that this does **not** include the stuff about streaming, observables, etc. we've talked about. This is just a simple implementation of the new protocol using the existing code and Hub model.

I also removed the "InvocationAdapter" and the "InvocationAdapterRegistry", since we aren't sure we want to expose it publicly any more. Custom Hub Protocols are not a feature we're intending to support (note that we **will** be allowing apps to control serialization of Hub arguments, just not our outer protocol). There's a bunch of deleted code in the sample now, since the "Protobuf" and "Line" adapters were cleaned up, but we'll have it in history for when we start implementing Protobuf.

There are definitely some things we'll still need to review, for example we'll need to start looking at how to generate Invocation IDs for requests from the server to the client. Right now I use `Guid`s everywhere, but we may not want to do that. The Client uses a monotonically increasing long, which is fine for it since it's sending to one destination, but servers often want to send the same message to multiple destinations.

JavaScript client is **not yet updated**. We don't test it automatically so I'm not quite sure what the best thing to do here is. The changes are fairly minimal, and could happen in a separate PR, but our samples would be broken in between this merging and the JavaScript client being updated. I lean towards doing that since it'll be a brief period of time that we'll have broken samples, and none of our CI builds will be broken.

I do plan to add some tests to `JsonHubProtocol`, which I will do early tomorrow.


SignalR.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.9
```
anurse: Wow...  VS seems to have barfed all over my SLN. Not sure if I should revert it if it's only going to do it again...

davidfowl: Revert

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -105,15 +108,10 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
             _logger.LogTrace("Preparing invocation of '{0}', with return type '{1}' and {2} args", methodName, returnType.AssemblyQualifiedName, args.Length);
 
             // Create an invocation descriptor.
-            var descriptor = new InvocationDescriptor
-            {
-                Id = GetNextId(),
-                Method = methodName,
-                Arguments = args
-            };
+            var invocationMessage = new InvocationMessage(GetNextId().ToString(), methodName, args);
```
davidfowl: Oh mah ghad, stringz

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/IHubProtocol.cs
```diff
@@ -0,0 +1,20 @@
...
+    public interface IHubProtocol
+    {
+        MessageType MessageType { get; }
+
+        bool TryParseMessage(ReadOnlySpan<byte> input, IInvocationBinder binder, out HubMessage message);
+
+        // Need a better API when we have sorted out pooling, etc.
+        // See https://github.com/aspnet/SignalR/issues/126
+        // For exmaple: bool TryWriteMessage(HubMessage message, IOutput output);
+        byte[] WriteMessage(HubMessage message);
```
davidfowl: This should take a `Span<byte>` or `IOutput`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public InvocationMessage(string invocationId, string target, object[] arguments) : base(invocationId)
+        {
+            Target = target;
+            Arguments = arguments;
+        }
+
+        public override string ToString()
+        {
+            return $"Invocation {{ id: \"{InvocationId}\", target: \"{Target}\", arguments: [ {string.Join(", ", Arguments.Select(a => a.ToString()))} ] }}";
```
davidfowl: Needs to handle null args.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/CompletionMessage.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
```
davidfowl: Why internal?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,190 @@
...
+        private const string TypePropertyName = "type";
+        private const string ErrorPropertyName = "error";
+        private const string TargetPropertyName = "target";
+        private const string ArgumentsPropertyName = "arguments";
+
+        private const int InvocationMessageType = 1;
+        private const int ResultMessageType = 2;
+        private const int CompletionMessageType = 3;
+
+        private JsonSerializer _serializer = new JsonSerializer();
```
davidfowl: This is the for outer protocol only right?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            var paramTypes = binder.GetParameterTypes(target);
+            var arguments = new object[paramTypes.Length];
+
+            var args = json.Value<JArray>(ArgumentsPropertyName);
+            for (var i = 0; i < paramTypes.Length; i++)
+            {
+                var paramType = paramTypes[i];
+
+                // TODO(anurse): We can add some DI magic here to allow users to provide their own serialization
```
davidfowl: Link to the bug

src/Microsoft.AspNetCore.SignalR.Common/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -8,11 +8,21 @@
...
+    <RootNamespace>Microsoft.AspNetCore.SignalR</RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All"/>
+    <Compile Include="../Common/IOutputExtensions.cs" Link="IOutputExtensions.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Memory" Version="4.4.0-*" />
```
davidfowl: Use the pattern bro. `$(SystemMemoryVersion)`

src/Microsoft.AspNetCore.SignalR.Common/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -8,11 +8,21 @@
...
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
```
davidfowl: We don't need this anywhere anymore.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -23,18 +24,18 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
-        private readonly InvocationAdapterRegistry _registry;
         private readonly ConnectionMultiplexer _redisServerConnection;
         private readonly ISubscriber _bus;
         private readonly ILoggerFactory _loggerFactory;
         private readonly RedisOptions _options;
 
-        public RedisHubLifetimeManager(InvocationAdapterRegistry registry,
-                                       ILoggerFactory loggerFactory,
+        // TODO: What protocol should we use to publish to the bus?
+        private readonly IHubProtocol _protocol = new JsonHubProtocol();
```
davidfowl: Whatever protocol we want really.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -139,5 +122,10 @@ public override Task OnDisconnectedAsync(Connection connection)
                 }
             }
         }
+
+        private static string GetInvocationId()
+        {
+            return Guid.NewGuid().ToString("N");
```
davidfowl: Perf...

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -36,19 +36,19 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
         private readonly HubLifetimeManager<THub> _lifetimeManager;
         private readonly IHubContext<THub, TClient> _hubContext;
         private readonly ILogger<HubEndPoint<THub, TClient>> _logger;
-        private readonly InvocationAdapterRegistry _registry;
         private readonly IServiceScopeFactory _serviceScopeFactory;
 
+        // TODO: Configuration and Protobuf
+        private readonly JsonHubProtocol _protocol = new JsonHubProtocol();
```
davidfowl: Can we inject this.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -140,115 +140,112 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
+        {
+            // Encode the message
+            var payload = _protocol.WriteMessage(hubMessage);
 
-            var outMessage = new Message(outStream.ToArray(), MessageType.Text, endOfMessage: true);
+            // Create a message and put it on the outbound channel
+            var message = new Message(payload, _protocol.MessageType, endOfMessage: true);
 
-            while (await connection.Transport.Output.WaitToWriteAsync())
+            while (!connection.Transport.Output.TryWrite(message))
```
davidfowl: What is this pattern?

src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
```diff
@@ -16,8 +16,6 @@ public static ISignalRBuilder AddSignalR(this IServiceCollection services)
             services.AddSingleton(typeof(IHubContext<>), typeof(HubContext<>));
             services.AddSingleton(typeof(HubEndPoint<>), typeof(HubEndPoint<>));
             services.AddSingleton<IConfigureOptions<SignalROptions>, SignalROptionsSetup>();
-            services.AddSingleton<JsonNetInvocationAdapter>();
```
davidfowl: Add the `JsonHubProtocol`.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -173,10 +175,10 @@ public class HubEndpointTests
             {
                 var endPointTask = endPoint.OnConnectedAsync(client.Connection);
 
-                var result = await client.Invoke<InvocationResultDescriptor>(nameof(MethodHub.ValueMethod)).OrTimeout();
+                var result = (await client.InvokeAndWait(nameof(MethodHub.ValueMethod)).OrTimeout()).Single();
```
davidfowl: This is kinda ugly. What about InvokeSingle.

berdon:
Is there or could there be a general discussion on InvocationAdapters? It's pretty convenient to be able to use one. I can give an example:

We have the traditional server/client model where communication happens between the clients and the servers. The disparity comes from having two disjoint client groups each of which the server needs to provide backwards compatibility for communication. Meaning, as we change data models being transmitted between client groups we need to be able to translate these versioned client models from older versions to the current version back again to older versions.

This seemed trivial with the InvocationAdapter and being able to map old models to the current model, etc. What would a proposed solution be lacking the construct?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/IHubProtocol.cs
anurse: It would have to be `IOutput`, otherwise we have to handle allocation strategies everywhere it's called. I started with that, but the JSON serialization is all `Stream`-based, so it ends up just using `MemoryStream` and `ToArray` to get the data, then it just bulk-writes that to the `IOutput`.

And then, we have to build our own `IOutput` implementation over a buffer pool, or allocate a pipeline. I didn't feel like we were ready for that so I used this API rather than try to shoe-horn a messy implementation pattern into the "desired" API. We need a proper plan for how to actually handle pooling of memory. If we want to block this change on getting that figured out, I can work on it, but I didn't want to overload this PR.

davidfowl: Thats the pattern we use for everything else at the moment already though.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
anurse: Right now, both, but yes, it is intended to be for the outer protocol only.

davidfowl:
@berdon I don't really understand what you're asking for.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/CompletionMessage.cs
anurse: Do we want this implementation detail to be part of our public API surface?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: We likely will when we have DI-based hub argument serialization, but I didn't want to jump the gun on that when we don't intend for users to be able to replace the outer Hub Protocol logic.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: Sure, just trying to avoid the await if we can write immediately and make it consistent with the read pattern.

src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
anurse: Again, for what purpose? Are we changing our plan on user-replaceable hub protocols?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Doing it in a single place is pretty confusing, revert after do it later, everywhere. Though I hate how it looks.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: image

berdon:
@davidfowl 

I'll try to clarify my use case - which I don't think is atypical.

I have interfaces that I'll use as arguments for Hub methods:

- IModelOne
- IModelTwo
- IModelThree

I have implementations of these interfaces

- ModelOneV1
- ModelOneV2
- ModelTwoV1
- ...

I want to be able to have Hub clients call the Hub method taking in the interface types but passing in their implementations. So far, this works as is.

However, the complication comes in when needing backwards compatibility - where older models no longer contractually agree with the interface types. Ie. We have an old version one client calling a Hub method accepting IModelOne but the new version IModelOne has a breaking change for ModelOneV1.

Currently, we were expecting to write a translator that lived inside a custom InvocationAdapter that could handle translating between older model objects and the current model object.

I don't think the InvocationAdapter necessarily _needs_ to exist especially depending on what your guys' thoughts are on argument serialization.

anurse:
The Invocation Adapter was only intended as a way to serialize/deserialize the hub messages, hence the changes in this PR. However, having a component that allows you to intercept and rewrite those messages does seem like it might be useful and worth discussing, for exactly those scenarios.

Can I suggest moving this discussion to a new Issue? This seems like something we'll want to discuss in depth to figure out the best way to do it, but I don't want to overload this discussion thread with it.

berdon:
Absolutely.

Full disclosure, my team is completely leveraging aspnet/SignalR right now and have been for some time. We've been riding the wave so to speak on the changes but this one presented a pretty breaking change for us which is why I had to pipe up.

I'm hoping we can contribute when able - for instance, I plan on implementing the [Authorize] stuff next week and was planning on submitting a PR.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: I hate how most of the Channel consumption API looks :)

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/CompletionMessage.cs
davidfowl: Yes

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Thats not jumping the gun, we do it for everything else.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,190 @@
...
+            {
+                WriteMessage(message, memoryStream);
+                memoryStream.FlushAsync();
+                return memoryStream.ToArray();
+            }
+        }
+
+        private HubMessage ParseMessage(Stream input, IInvocationBinder binder)
+        {
+            var reader = new JsonTextReader(new StreamReader(input));
```
BrennanConroy: dispose the reader

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -60,60 +61,38 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+        private async Task PublishAsync(string channel, HubMessage message)
         {
-            // TODO: What format??
-            var invocationAdapter = _registry.GetInvocationAdapter("json");
-
             // BAD
-            using (var ms = new MemoryStream())
-            {
-                await invocationAdapter.WriteMessageAsync(message, ms);
+            var payload = _protocol.WriteMessage(message);
```
BrennanConroy: remove comment

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,190 @@
...
+        public JsonHubProtocol()
+        {
+        }
+
+        public bool TryParseMessage(ReadOnlySpan<byte> input, IInvocationBinder binder, out HubMessage message)
+        {
+            // TODO: Need a span-native JSON parser!
+            using (var memoryStream = new MemoryStream(input.ToArray()))
+            {
+                message = ParseMessage(memoryStream, binder);
```
BrennanConroy: This can throw which isn't good in a `Try*` method

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,190 @@
...
+                    // Invocation
+                    return BindInvocationMessage(json, binder);
+                case ResultMessageType:
+                    // Result
+                    return BindResultMessage(json, binder);
+                case CompletionMessageType:
+                    // Completion
+                    return BindCompletionMessage(json);
+                default:
+                    throw new FormatException($"Unknown message type: {type}");
```
moozzyk: Log as error?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
moozzyk: We still need some kind of conversion - if a client speaks protobuf we can't send them JSON.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -14,12 +14,7 @@ namespace Microsoft.AspNetCore.SignalR
     public class DefaultHubLifetimeManager<THub> : HubLifetimeManager<THub>
     {
         private readonly ConnectionList _connections = new ConnectionList();
-        private readonly InvocationAdapterRegistry _registry;
-
-        public DefaultHubLifetimeManager(InvocationAdapterRegistry registry)
-        {
-            _registry = registry;
-        }
+        private readonly IHubProtocol _protocol = new JsonHubProtocol();
```
moozzyk: Shouldn't this be per connection and not global?

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
moozzyk: Why can't we just use (unsigned) int/long here?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -140,115 +140,112 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
                     {
-                        InvocationDescriptor invocationDescriptor;
-                        var inputStream = new MemoryStream(incomingMessage.Payload);
-
-                        // TODO: Handle receiving InvocationResultDescriptor
-                        invocationDescriptor = await invocationAdapter.ReadMessageAsync(inputStream, this) as InvocationDescriptor;
-
-                        // Is there a better way of detecting that a connection was closed?
-                        if (invocationDescriptor == null)
+                        if (!_protocol.TryParseMessage(incomingMessage.Payload, this, out var hubMessage))
```
moozzyk: Are we tying protocol to endpoint instead of connection now?

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -37,58 +37,86 @@ public TestClient(IServiceProvider serviceProvider, string format = "json")
...
+        public async Task<HubMessage> Read()
         {
-            while (await Application.Input.WaitToReadAsync(_cts.Token))
+            while(true)
             {
-                var value = await TryRead<T>();
+                var message = TryRead();
 
-                if (value != null)
+                if(message == null)
```
moozzyk: nit: formatting (same with the `if` below)

samples/SocketsSample/LineInvocationAdapter.cs
```diff
@@ -1,91 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: Is this going away completely or will be replaced? Because if is going away the corresponding javascript code should be removed as well.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -124,12 +107,12 @@ public override Task OnDisconnectedAsync(Connection connection)
             return TaskCache.CompletedTask;
         }
 
-        private static async Task WriteAsync(Connection connection, IInvocationAdapter invocationAdapter, InvocationDescriptor invocation)
+        private async Task WriteAsync(Connection connection, HubMessage hubMessage)
         {
             var stream = new MemoryStream();
```
JanEggers: missing using statement ?

anurse: `MemoryStream` technically doesn't really need it since all it's resources are managed and will be marked for clean up at the end of the method, but yeah, it's good to maintain the pattern. This code has been changed around anyway though :).

Good catch!

anurse:
FYI, this PR is not in a reviewable state right now, I'm going to close it and reopen it when it is. I'm making changes based on #372 feedback

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -122,60 +120,62 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
 
-        private async void OnDataReceived(byte[] data, MessageType messageType)
+        private void OnDataReceived(byte[] data, MessageType messageType)
         {
-            var message
-                = await _adapter.ReadMessageAsync(new MemoryStream(data), _binder, _connectionActive.Token);
+            if (!_protocol.TryParseMessage(data, _binder, out var message))
+            {
+                _logger.LogError("Received invalid message");
+                throw new InvalidOperationException("Received invalid message");
```
anurse: It would depend on what the caller will do with the exception. I can avoid throwing here, but I also think that IConnection shouldn't allow an exception in the Received event bring down the process.

samples/SocketsSample/LineInvocationAdapter.cs
anurse: I'm not seeing the JavaScript code anywhere. Is it in SocketsSample?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: Not sure yet, but most of this logic is incomplete in the PR right now. We can have a quick chat about it with @davidfowl . I think I'd lean towards Connection, but allowing EndPoints to restrict allowed protocols.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
anurse: Yep, this is just incomplete right now.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
anurse: Yeah, we can do that actually. The initial concern was in situations where we were invoking across multiple connections and wanted to generate the ID only once and send the same message to each connection. Originally, @davidfowl and I thought we'd need to use GUIDs because each connection would be on a different Invocation ID, but I was confusing Result messages and Invocation messages :). In either case, we can just generate a seemingly-random integer (using a per-endpoint incrementing int like this) and it'll be fine.

I do still want to keep it a string (which I changed when switching to GUID) because I think it's better to be flexible (and since we don't care about them as numbers, except as a source of uniqueness)

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: I mean, it's still bad ;P

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: Yeah, that'll all be part of adding ProtoBuf support. My thought right now though is that it'll be the responsibility of each node to deserialize the object and reserialize it (once, lazily) where needed. So the protocol we actually put in Redis doesn't really matter (would likely be ProtoBuf for compactness)

samples/SocketsSample/LineInvocationAdapter.cs
moozzyk: It was in SocketSample - looks like it has been removed (by me LOL). In that case I think this: https://github.com/aspnet/SignalR/blob/dev/samples/SocketsSample/wwwroot/hubs.html#L10-L13 can be removed.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: It is an event handler and I was worrying that it will throw an exception on a random thread that no one is able to see. I checked and it actually seems it would fail the receive task in the Connection.cs which would cause the connection to be closed so it should be fine. Can you add a test for this scenario if you have not?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
moozzyk: It's OK for endpoint to restrict allowed protocols but if protocol is not tied to connection you wouldn't be able to have one connection talking JSon and another talking Protobuf, no?

moozzyk: As per the other comments - this will probably come later when we add support for protobuf so I think it is fine to go with JSON protocol hardcoded and tied to EndPoint and then fix it when adding protobuf

anurse:
https://github.com/aspnet/SignalR/pull/390

PR #378, opened 2017-04-08
0b11111000 is not a valid first UTF8-8 byte

Tragetaschen:
I might be overlooking something, or my counting skills are off-by-one, but that are 9 4s and only 7 0s. 0b11111000 (0xF8) is not a valid first byte of a multi-byte character in UTF-8

### DIFF
--- a/src/Microsoft.Extensions.WebSockets.Internal/Utf8Validator.cs
+++ b/src/Microsoft.Extensions.WebSockets.Internal/Utf8Validator.cs
@@ -12,6 +12,9 @@ namespace Microsoft.Extensions.WebSockets.Internal
     public class Utf8Validator
     {
         // Table of UTF-8 code point widths. '0' indicates an invalid first byte.
+        // 0x80 - 0xBF are the continuation bytes and invalid as first byte.
+        // 0xC0 - 0xC1 are overlong encodings of ASCII characters
+        // 0xF5 - 0xFF encode numbers that are larger than the Unicode limit (0x10FFFF)
         private static readonly byte[] _utf8Width = new byte[256]
         {
             /* 0x00 */ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, /* 0x0F */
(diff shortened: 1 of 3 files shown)


anurse:
Good catch! Could you add a quick test (just some additional `InlineData` rows in https://github.com/Tragetaschen/SignalR/blob/e8b19baa1883a35a5ab99f15f9e1512e71e3f1b7/test/Microsoft.Extensions.WebSockets.Internal.Tests/Utf8ValidatorTests.cs#L93 should do) for this and a couple of the other invalid first bytes? Just want to make sure this doesn't regress if we refactor this code :).

Tragetaschen:
I started looking into it more closely and now it's more of an optimization than an actual fix.

I've marked in the array more starting bytes as invalid:
- 0xC0 and 0xC1 are overlong encodings of ASCII
- 0xF5-0xF7 decode to code points outside of UTF-8
- 0xF8 as a five-byte-encoding is outside the UTF-8 standard

These starting bytes now quickly mark the sequence as invalid and don't go through the remaining validation steps. For example [the test for an overlong encoding of 'H'](https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.Extensions.WebSockets.Internal.Tests/Utf8ValidatorTests.cs#L83) now exits earlier, but doesn't otherwise change.

test/Microsoft.Extensions.WebSockets.Internal.Tests/Utf8ValidatorTests.cs
```diff
@@ -98,6 +101,18 @@ public void InvalidSingleFramePayloads(byte[] payload)
 
         [Theory]
 
+        [InlineData(new byte[] { 0xC0 }, new byte[] { 0xA0 })]
+        [InlineData(new byte[] { 0xC1 }, new byte[] { 0x88 })]
+        [InlineData(new byte[] { 0xF5 }, new byte[] { 0x80, 0x80, 0x80 })]
+        public void InvalidMultiFramePayloadsByFirstByte(byte[] payload, byte[] _)
```
anurse: Just go ahead and remove the second parameter from this test. Maybe put a short comment on the test explaining why these cases are invalid.

src/Microsoft.Extensions.WebSockets.Internal/Utf8Validator.cs
```diff
@@ -12,6 +12,9 @@ namespace Microsoft.Extensions.WebSockets.Internal
     public class Utf8Validator
     {
         // Table of UTF-8 code point widths. '0' indicates an invalid first byte.
+        // 0x80 - 0xBF are the continuation bytes and invalid as first byte.
+        // 0xC0 - 0xC1 are overlong encodings of ASCII characters
+        // 0xF5 - 0xFF encode numbers that are larger than the Unicode limit (0x10FFFF)
```
anurse: :+1:

Tragetaschen:
Yeah, I'm seeing those as the CI finds them. My VS2017 won't discover the tests, and running them manually doesn't work either. So I'm blind here :-/

anurse:
Yeah, we've seen problems with Test Discovery in VS 2017. You can also run `.\build.cmd` in the repo, or cd to a test project directory and run `dotnet test` to run from the command line.

Tragetaschen:
Yay! Sorry for occupying the CI for so long...

anurse:
@Tragetaschen no worries at all! That's what the CI is there for! (And I've done it myself a fair share of times ;))

PR #382, opened 2017-04-11
Add non-blocking invocations to the spec

anurse:
@moozzyk noted that we have cases where we need non-blocking invocations (i.e. Invocations where the Caller does not expect a Result or Completion). Server to Client broadcasts are examples of this case.

### DIFF
--- a/specs/HubProtocol.md
+++ b/specs/HubProtocol.md
@@ -51,6 +51,12 @@ The SignalR protocol allows for multiple `Result` messages to be transmitted in 
 
 On the Callee side, it is up to the Callee's Binder to determine if a method call will yield multiple results. For example, in .NET certain return types may indicate multiple results, while others may indicate a single result. Even then, applications may wish for multiple results to be buffered and returned in a single `Completion` frame. It is up to the Binder to decide how to map this. The Callee's Binder must encode each result in separate `Result` messages, indicating the end of results by sending a `Completion` message. Since the `Completion` message accepts an optional payload value, methods with single results can be handled with a single `Completion` message, bearing the complete results.
 
+On the Caller side, the user code which performs the invocation indicates how it would like to receive the results and it is up the Caller's Binder to determine how to handle the result. If the Caller expects only a single result, but multiple results are returned, the Caller's Binder should yield an error indicating that multiple results were returned. However, if a Caller expects multiple results, but only a single result is returned, the Caller's Binder should yield that single result and indicate there are no further results.
+
+## Completion and Results
+
+A Invocation is only considered completed when the `Completion` message is recevied. Receiving **any** message using the same `Invocation ID` after a `Completion` message has been received for that invocation is considered a protocol error and the recipient should immediately terminate the connection.
+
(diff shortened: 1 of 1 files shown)


specs/HubProtocol.md
```diff
@@ -30,16 +30,21 @@ In order to perform a single invocation, Caller follows the following basic flow
...
 2. Send an `Invocation` message containing the `Invocation ID`, the name of the `Target` being invoked, and the `Arguments` to provide to the method.
-3. Wait for a `Result` or `Completion` message with a matching `Invocation ID`
-4. If a `Completion` message arrives, go to 7
-5. If the `Result` message has a payload, dispatch the payload to the application (i.e. by yielding a result to an `IObservable`, or by collecting the result for dispatching in step 8)
-6. Go to 3
-7. Complete the invocation, dispatching the final payload item (if any) or the error (if any) to the application
+3. If the `Invocation` is marked as non-blocking (see "Non-Blocking Invocations" below), stop here and immediately yield back to the application.
+4. Wait for a `Result` or `Completion` message with a matching `Invocation ID`
+5. If a `Completion` message arrives, go to 8
+6. If the `Result` message has a payload, dispatch the payload to the application (i.e. by yielding a result to an `IObservable`, or by collecting the result for dispatching in step 8)
```
moozzyk: This change actually fixed this step before there was no step 8 :)

anurse:
I'm going to incorporate some changes suggested by @moozzyk and require that Streaming data use `Result` frames and non-streaming data use `Completion` frames. It makes things a lot clearer.

specs/HubProtocol.md
anurse: Yeah, I noticed that. Whoops!

anurse:
:up: :date: after discussing with @moozzyk .

specs/HubProtocol.md
```diff
@@ -23,23 +23,28 @@ There are two encodings of the SignalR protocol: [JSON](http://www.json.org/) an
 In the SignalR protocol, the following types of messages can be sent:
 
 * `Invocation` Message - Indicates a request to invoke a particular method (the Target) with provided Arguments on the remote endpoint.
-* `Result` Message - Indicates response data from a previous Invocation message.
-* `Completion` Message - Indicates a previous Invocation has completed, and no further `Result` messages will be received. Optionally contains a final item of response data, or an error.
+* `Result` Message - Indicates individual items of streamed response data from a previous Invocation message.
```
moozzyk: I wonder if we should rename it to StreamingResult/StreamingItem

specs/HubProtocol.md
```diff
@@ -48,13 +53,33 @@ On the Callee side, it is up to the Callee's Binder to determine if a method cal
 
 On the Caller side, the user code which performs the invocation indicates how it would like to receive the results and it is up the Caller's Binder to determine how to handle the result. If the Caller expects only a single result, but multiple results are returned, the Caller's Binder should yield an error indicating that multiple results were returned. However, if a Caller expects multiple results, but only a single result is returned, the Caller's Binder should yield that single result and indicate there are no further results.
 
+## Completion and Results
+
+A Invocation is only considered completed when the `Completion` message is recevied. Receiving **any** message using the same `Invocation ID` after a `Completion` message has been received for that invocation is considered a protocol error and the recipient should immediately terminate the connection.
```
moozzyk: Shouldn't be "A**n** Invocation" ?

moozzyk: I am not sure about the requirement of terminating the connection. In a general case you would have to keep all invocation ids to know if the message you received is for an invocation that has already completed. Rather I would say that the client MAY terminate a connection or ignore a message if the invocation id does not correspond to any pending invocations.

anurse: Sure, that's basically the intent.

Issue #256, opened 2017-03-01
Authorization for EndPoints and Hubs


BrennanConroy:
- [ ] Authorize Attribute on EndPoint
- [ ] Authorize Attribute on Hub class and methods
- [x] Use Policy name in EndPointOptions

berdon:
@BrennanConroy Would you have a few minutes today to discuss the thoughts around this?

I'd like to help out but I want to make sure any changes I make are in line with what you guys are planning. I've done some initial work but with #254 and #255 that might change how or where Hub policies might be stored after checking TypeInfo custom attributes.

BrennanConroy:
Closing in favor of https://github.com/aspnet/SignalR/issues/552

Issue #255, opened 2017-03-01
Add AddHub methods


BrennanConroy:
Do this in the same PR or at the same time as #254
```c#
AddHub<THub>()
AddHub<THub, TClient>()
AddHub<THub>(HubOptions)
AddHub<THub, TClient>(HubOptions)
```

berdon:
How do you envision this being used? Something like:

```
public void ConfigureServices(IServiceCollection services)
{
    services.AddSignalR()
      .AddHub<MyHub>(o -> ...)
      .AddHub<MyOtherHub>();
}

public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
{
  app.UseSignalR(routes =>
  {
    routes.MapHub<MyHub>("/hub");
    routes.MapHub<MyOtherHub>("/other");
  });
}

Would `AddHub()` be optional if you didn't need to specify options?

BrennanConroy:
That snippet looks right. It isn't necessary unless you want DI in your Hub I believe.

grahamehorner:
I think that AddHub should be explicit, meaning that if a developer used a shared assembly that contained two SignalR hubs but only wished to publish/expose one as a service then they can do.  I think adding all SignalR hubs as services automatic is a bad idea for resource management etc. or have I misunderstood what AddHub is for an the above comments ?

muratg:
Needed for #261

muratg:
Closing per today's triage decision.

Issue #254, opened 2017-03-01
Add HubOptions


BrennanConroy:
HubOptions will contain EndPoint options.

Initial set of thoughts for the options:
- [ ] Verbose error for hub methods
- [ ] Adding/Removing supported protocols?
- [x] Negotiate Timeout?
- [X] JsonSerializerSettings

muratg:
Needed for #261

anurse:
We may want both a global `HubOptions` and a `HubOptions<T>` that lets you override options for a specific Hub.

BrennanConroy:
Moving out of alpha, we did what we wanted to for alpha

moozzyk:
https://stackoverflow.com/questions/46766516/get-detailed-error-for-exception-in-signalr-core-hub-method-execution

muratg:
Closing per today's triage decision.

niklr:
What happened to JsonSerializerSettings? Seems to be removed after 1.0.0-alpha2-final

anurse:
Since the protocols are pluggable, the options were moved to a different location. Now you set them in the `.AddJsonProtocol` extension method in `ConfigureServices`, like this:

```csharp
services.AddSignalR()
    .AddJsonProtocol(options => {
        // Edit or replace 'options.PayloadSerializerSettings' here!
    });
```

borissedov:
I want to enable custom exceptions logging for Hub actions.
But my standard approach from Startup.cs
` app.UseExceptionHandler(handler =>
            {
                handler.Run(context =>
                {
                    var exception = context.Features.Get<IExceptionHandlerFeature>()?.Error;
                    ...
               }
           });
`
doesn't work for Hub errors.
Is there any workaround?

PR #386, opened 2017-04-12
Adding support for Hub and HubMethod [Authorize]

berdon:
- Allows for implicit method protection with [Authorize] on Hub
- Allows for [AllowAnonymous] to allow unauthorized access on protected Hub

Addresses #256 

Sneak peak / first draft for authorization implementation on Hub and Hub Methods. I'm hoping this is in alignment with the changes @BrennanConroy is doing in #254 / #255 and doesn't step on any toes there.

I added test cases and updated the ChatSample to include a Role requirement. Presently, [Authorize] on Hub is mostly useful for protecting all underlying methods as it doesn't prevent a connection unlike `EndPointOptions<>.AuthorizationPolicyNames`. There's some loose ends too on what, if at all, Authentication and Challenging looks like within the Hub context.

### DIFF
--- a/samples/ChatSample/Hubs/Chat.cs
+++ b/samples/ChatSample/Hubs/Chat.cs
@@ -8,7 +8,6 @@ using Microsoft.AspNetCore.SignalR;
 
 namespace ChatSample.Hubs
 {
-    // TODO: Make this work
     [Authorize]
     public class Chat : Hub
     {
@@ -30,6 +29,12 @@ namespace ChatSample.Hubs
         public async Task Send(string message)
         {
             await Clients.All.InvokeAsync("Send", $"{Context.User.Identity.Name}: {message}");
+        }
+
+        [Authorize(Roles = "Moderator")]
+        public async Task Shout(string message)
+        {
+            await Clients.All.InvokeAsync("Shout", $"{Context.User.Identity.Name}: {message}");
         }
     }
 }
--- a/samples/ChatSample/Views/Home/Index.cshtml
+++ b/samples/ChatSample/Views/Home/Index.cshtml
@@ -10,6 +10,7 @@
     <form id="sendmessage" action="#">
         <input type="text" id="new-message" />
         <input type="submit" id="send" value="Send" class="send" />
+        <input type="button" id="shout" value="Shout" class="shout" />
     </form>
 </div>
 <script src="lib/signalr-client/signalr-client.js"></script>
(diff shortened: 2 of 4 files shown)


berdon:
Adding more tests around the AuthorizeAttribute:
- Named policies
- Role requirements

davidfowl:
Hey @berdon thanks for the pull request! I think the authorize attribute on the hub should map conceptually to end point auth.

berdon:
@davidfowl 

I had actually originally done that by adding the mapping in the extension method to set the `IEndPointOptions` to whatever `IAuthorizeData` was applied to the Hub. I didn't do that for two reasons:

1. Conceptually, that stage is different: authorization at that stage allows for skipping a "connection" entirely
2. Because of 1, overriding the default Hub AuthorizeAttribute becomes difficult because `EndPoint` has no notion of underlying `HubEndPoint` AuthorizeAttribute`s. It would _need_ to have knowledge to know whether it should forgo a connection or not.

I think it actually makes sense to have three stages of authorization:
- EndPoint
- Hub
- Hub Methods

When @BrennanConroy adds in the HubEndPoint it could still allow for setting the AuthorizationPolicy in that as well.

davidfowl:
While I understand that is conceptually true I'm saying that we should be doing 2 kinds, EndPoint and Hub Methods. We need the plumbing so that endpoints can advertise without hardcoding the policy via options. Maybe an auth selector so that the HubEndPoint can advertise the correct `IAuthorizeData`. 

The reason I want this to *just* work is so that http authentication works by default and when it works, the second level Hub is redundant.

grahamehorner:
I think it would be nice if the endpoint/hub authorisation attribute had a option such as RequireSecureConnection allowing hub to be exposed only on the endpoint it was designed/written for;  as this gives added safeguard when deploying of hubs, should the endpoint be incorrectly configured (for any reason) and therefor preventing data leak as the hub would simply refuse connection giving a  WEB_SOCKET_ENDPOINT_UNAVAILABLE_CLOSE_STATUS    = 1001

[Authorize(RequireSecureConnection=true, Roles = "Moderator")]

IMO this also becomes important when using DependencyInjection of hubs created by 3rd parties as they may need to enforce a secured connection to they hub.

davidfowl:
@berdon lets do the 2 pieces of auth we agree on, what you have is good here. I just want you to remove the Hub level auth or now. We can discuss that on the issue. I think [this comment]( https://github.com/aspnet/SignalR/pull/386#issuecomment-294246476) should be moved to https://github.com/aspnet/SignalR/issues/256 so we can continue discussing it there. 

This also needs a rebase. Let me know if you can do it otherwise I can just take it over.

berdon:
I'll rebase and get this cr updated.

davidfowl:
@berdon This may be a bit more challenging with the layering changes. I'm going to do some more refactoring here to make things a bit clearer.

PR #390, opened 2017-04-13
Implement new Hub Protocol (Part Deux)

anurse:
Take 2! Now up-to-date with the Hub Protocol as defined in the spec.

I still need to do some verification on Redis, but I wanted to get this review out.


anurse:
Verified Redis still works.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -118,81 +118,94 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
         public Task<object> Invoke(string methodName, Type returnType, params object[] args) => Invoke(methodName, returnType, CancellationToken.None, args);
         public async Task<object> Invoke(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
         {
+            if (_connectionActive.Token.IsCancellationRequested)
```
davidfowl: Lets just remove this.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -311,21 +369,62 @@ public InvocationHandler(Type[] parameterTypes, Action<object[]> handler)
             }
         }
 
-        private struct InvocationRequest
+        private class InvocationRequest : IDisposable
```
davidfowl: Why?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -311,21 +369,62 @@ public InvocationHandler(Type[] parameterTypes, Action<object[]> handler)
...
         {
+            private readonly TaskCompletionSource<object> _completionSource = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+            private readonly CancellationTokenRegistration _cancellationTokenRegistration;
+            private object _result;
+            private object _lock = new object();
```
davidfowl: Why?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/IHubProtocol.cs
```diff
@@ -0,0 +1,48 @@
...
+    public interface IHubProtocol
+    {
+        MessageType MessageType { get; }
+
+        bool TryParseMessage(ReadOnlySpan<byte> input, IInvocationBinder binder, out HubMessage message);
+
+        bool TryWriteMessage(HubMessage message, IOutput output);
+    }
+
+    public static class HubProtocolWriteMessageExtensions
```
davidfowl: Move to another class

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: a) it's going in a dictionary, so it'll be boxed anyway
b) we used to put it in once and take it out once and never look at it in between, now we need to keep pulling it out of the dictionary for each stream item we get (in theory).

This is somewhat premature as it happened in my prototype that did streaming, but I think it's still going to be useful. We can always review when we have perf data.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -118,81 +118,94 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
+                    DispatchInvocationCompletion(completion, irq, _connectionActive.Token);
+                    RemoveRequest(completion.InvocationId);
                     break;
+                case StreamItemMessage streamItem:
+                    // Complete the invocation with an error, we don't support streaming (yet)
+                    lock (_pendingCallsLock)
+                    {
+                        if (!_pendingCalls.TryGetValue(streamItem.InvocationId, out irq))
+                        {
+                            _logger.LogWarning("Dropped unsolicited Completion message for invocation '{invocationId}'", streamItem.InvocationId);
```
anurse: "Completion" message? No, that's not right ;).

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: You're right, it is now. This went through a few different iterations and the TCS was my last addition.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -204,69 +217,114 @@ private void Shutdown(Exception ex = null)
...
                 {
-                    _logger.LogInformation("Completing Invocation #{0} with error: {1}", result.Id, result.Error);
-                    irq.Completion.TrySetException(new Exception(result.Error));
+                    _logger.LogWarning("Duplicate request to remove invocation {invocationId} from the queue ignored", invocationId);
                 }
                 else
                 {
-                    _logger.LogInformation("Completing Invocation #{0} with result of type: {1}", result.Id, result.Result?.GetType()?.FullName ?? "<<void>>");
-                    irq.Completion.TrySetResult(result.Result);
+                    _pendingCalls.Remove(invocationId);
```
anurse: Dispose the invocation to release the cancellation token registration.

davidfowl: Do we ever use it in  using?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,227 @@
...
+        private InvocationMessage BindInvocationMessage(JObject json, IInvocationBinder binder)
+        {
+            var invocationId = json.Value<string>(InvocationIdPropertyName);
+            var target = json.Value<string>(TargetPropertyName);
+
+            var nonBlocking = false;
+            var nonBlockingProp = json.Property(NonBlockingPropertyName);
+            if (nonBlockingProp != null)
+            {
+                nonBlocking = nonBlockingProp.Value.ToObject<bool>();
```
davidfowl: ToObject is hella bad. Can you use `.Value<bool>`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -288,6 +312,23 @@ public void Dispose()
             }
         }
 
+        private string GetInvocationId()
+        {
+            var invocationId = Interlocked.Increment(ref _nextInvocationId) - 1;
```
davidfowl: Feels wrong for this to be generated in the redis logic. Maybe we should change the interface to pass this as an argument along with the invocation.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -140,115 +145,110 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
-                        // TODO: Handle receiving InvocationResultDescriptor
-                        invocationDescriptor = await invocationAdapter.ReadMessageAsync(inputStream, this) as InvocationDescriptor;
-
-                        // Is there a better way of detecting that a connection was closed?
-                        if (invocationDescriptor == null)
+                        if (!protocol.TryParseMessage(incomingMessage.Payload, this, out var hubMessage))
                         {
-                            break;
+                            _logger.LogError("Received invalid message");
+                            throw new InvalidOperationException("Received invalid message");
```
davidfowl: How do these exceptions manifest on the other side?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: What do you mean? The token registration? No, there isn't a wrapping lexical scope, so we don't use `using`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: I think it makes perfect sense for it to be in the lifetime manager. The lifetime manager handles all invocations from server to client, right?

src/Microsoft.AspNetCore.SignalR/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.AspNetCore.SignalR.Internal.Protocol;
```
moozzyk: nit: header

src/Microsoft.AspNetCore.SignalR/Internal/IHubProtocolResolver.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Microsoft.AspNetCore.SignalR.Internal.Protocol;
```
moozzyk: nit: header

src/Microsoft.AspNetCore.SignalR/SignalROptions.cs
```diff
@@ -1,20 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: Are we expecting any options here anytime soon? Or should we just remove this class for now?

src/Microsoft.AspNetCore.Sockets/ConnectionMetadataNames.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.AspNetCore.Sockets
+{
+    public static class ConnectionMetadataNames
+    {
+        public static readonly string Format = nameof(Format);
```
moozzyk: should we use case insensitive comparison for these?

src/Microsoft.AspNetCore.Sockets/EndPointDependencyInjectionExtensions.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Extensions.DependencyInjection
 {
-    public static class EndpointDependencyInjectionExtensions
+    public static class EndPointDependencyInjectionExtensions
```
moozzyk: :+1:

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -231,17 +231,12 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
         private ConnectionState CreateConnection(HttpContext context)
         {
-            var format =
-                string.Equals(context.Request.Query["format"], "binary", StringComparison.OrdinalIgnoreCase)
-                    ? MessageType.Binary
-                    : MessageType.Text;
-
             var state = _manager.CreateConnection();
             state.Connection.User = context.User;
+            state.Connection.Metadata[ConnectionMetadataNames.HttpContext] = context;
```
moozzyk: :+1:

test/Common/TaskExtensions.cs
```diff
@@ -10,36 +11,48 @@ public static class TaskExtensions
     {
         private const int DefaultTimeout = 5000;
 
-        public static Task OrTimeout(this Task task, int milliseconds = DefaultTimeout)
+        public static Task OrTimeout(this Task task, int milliseconds = DefaultTimeout, [CallerMemberName] string memberName = null, [CallerFilePath] string filePath = null, [CallerLineNumber] int? lineNumber = null)
```
moozzyk: :+1:
I learned about these attributes from @natemcmaster only this week

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -0,0 +1,144 @@
...
+    public class HubConnectionProtocolTests
+    {
+        [Fact]
+        public async Task InvokeSendsAnInvocationMessage()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(new JsonSerializer()), new LoggerFactory());
+            try
+            {
+                var invokeTask = hubConnection.Invoke("Foo", typeof(void));
```
moozzyk: looks funny because you actually did not start the connection

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        [InlineData(true, NullValueHandling.Ignore, "{\"invocationId\":\"123\",\"type\":1,\"target\":\"Target\",\"arguments\":[{\"stringProp\":\"SignalR!\",\"doubleProp\":6.2831853071,\"intProp\":42,\"dateTimeProp\":\"2017-04-11T00:00:00\"}]}")]
+        [InlineData(false, NullValueHandling.Include, "{\"invocationId\":\"123\",\"type\":1,\"target\":\"Target\",\"arguments\":[{\"StringProp\":\"SignalR!\",\"DoubleProp\":6.2831853071,\"IntProp\":42,\"DateTimeProp\":\"2017-04-11T00:00:00\",\"NullProp\":null}]}")]
+        [InlineData(true, NullValueHandling.Include, "{\"invocationId\":\"123\",\"type\":1,\"target\":\"Target\",\"arguments\":[{\"stringProp\":\"SignalR!\",\"doubleProp\":6.2831853071,\"intProp\":42,\"dateTimeProp\":\"2017-04-11T00:00:00\",\"nullProp\":null}]}")]
+        public async Task CanSerializeCustomObjectInInvocation(bool camelCase, NullValueHandling nullHandling, string expectedOutput)
+        {
+            var message = new InvocationMessage("123", "Target", new[] { new CustomObject() }, nonBlocking: false);
+
+            var payloadSerializer = new JsonSerializer();
+            if(camelCase)
```
moozzyk: formatting

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        [InlineData(true, NullValueHandling.Ignore, "{\"invocationId\":\"123\",\"type\":2,\"result\":{\"stringProp\":\"SignalR!\",\"doubleProp\":6.2831853071,\"intProp\":42,\"dateTimeProp\":\"2017-04-11T00:00:00\"}}")]
+        [InlineData(false, NullValueHandling.Include, "{\"invocationId\":\"123\",\"type\":2,\"result\":{\"StringProp\":\"SignalR!\",\"DoubleProp\":6.2831853071,\"IntProp\":42,\"DateTimeProp\":\"2017-04-11T00:00:00\",\"NullProp\":null}}")]
+        [InlineData(true, NullValueHandling.Include, "{\"invocationId\":\"123\",\"type\":2,\"result\":{\"stringProp\":\"SignalR!\",\"doubleProp\":6.2831853071,\"intProp\":42,\"dateTimeProp\":\"2017-04-11T00:00:00\",\"nullProp\":null}}")]
+        public async Task CanSerializeCustomObjectInResult(bool camelCase, NullValueHandling nullHandling, string expectedOutput)
+        {
+            var message = new StreamItemMessage("123", new CustomObject());
+
+            var jsonSerializer = new JsonSerializer();
+            if(camelCase)
```
moozzyk: nit: format

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        public async Task CanSerializeCustomObjectInInvocation(bool camelCase, NullValueHandling nullHandling, string expectedOutput)
+        {
+            var message = new InvocationMessage("123", "Target", new[] { new CustomObject() }, nonBlocking: false);
+
+            var payloadSerializer = new JsonSerializer();
+            if(camelCase)
+            {
+                payloadSerializer.ContractResolver = new CamelCasePropertyNamesContractResolver();
+            }
+            payloadSerializer.NullValueHandling = nullHandling;
```
moozzyk: You could initialize this inline (i.e.: `var payloadSerializer = new JsonSerializer {NullValueHandling = nullHandling}`)

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        [InlineData(true, NullValueHandling.Ignore, "{\"invocationId\":\"123\",\"type\":3,\"result\":{\"stringProp\":\"SignalR!\",\"doubleProp\":6.2831853071,\"intProp\":42,\"dateTimeProp\":\"2017-04-11T00:00:00\"}}")]
+        [InlineData(false, NullValueHandling.Include, "{\"invocationId\":\"123\",\"type\":3,\"result\":{\"StringProp\":\"SignalR!\",\"DoubleProp\":6.2831853071,\"IntProp\":42,\"DateTimeProp\":\"2017-04-11T00:00:00\",\"NullProp\":null}}")]
+        [InlineData(true, NullValueHandling.Include, "{\"invocationId\":\"123\",\"type\":3,\"result\":{\"stringProp\":\"SignalR!\",\"doubleProp\":6.2831853071,\"intProp\":42,\"dateTimeProp\":\"2017-04-11T00:00:00\",\"nullProp\":null}}")]
+        public async Task CanSerializeCustomObjectInCompletion(bool camelCase, NullValueHandling nullHandling, string expectedOutput)
+        {
+            var message = new CompletionMessage("123", error: null, result: new CustomObject(), hasResult: true);
+
+            var jsonSerializer = new JsonSerializer();
+            if(camelCase)
```
moozzyk: :arrow_up:

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -204,69 +217,114 @@ private void Shutdown(Exception ex = null)
...
+            {
+                return;
+            }
+
+            // If the invocation hasn't been cancelled, dispatch the completion
+            if (!irq.CancellationToken.IsCancellationRequested)
+            {
+                if (!string.IsNullOrEmpty(completion.Error))
+                {
+                    irq.Complete(new Exception(completion.Error));
```
anurse: Given that this is generally an application-defined error (aside from a few cases), we should use `Exception` unless we want to encode an exception type in the Error frame. If we want to throw specific exceptions for our well-known errors, maybe we need to associate a numeric code or symbolic name to some errors?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/CompletionMessage.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
+{
+    public class CompletionMessage : HubMessage
+    {
+        public string Error { get; }
+        public object Result { get; }
+        public bool HasResult { get; }
```
anurse: If you return `null` from a Hub method, we "have" a result, but it's `null`. Need to distinguish that from a CompletionMessage that has no `result` at all

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        public bool TryParseMessage(ReadOnlySpan<byte> input, IInvocationBinder binder, out HubMessage message)
+        {
+            // TODO: Need a span-native JSON parser!
+            using (var memoryStream = new MemoryStream(input.ToArray()))
+            {
+                message = ParseMessage(memoryStream, binder);
+            }
+            return true;
```
anurse: Actually, this doesn't need to be a try method at all, it's like our parser/formatter methods but unlike there we know exactly how much data to parse, so there's no need return `false` to indicate more data is required.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,227 @@
...
+            return new InvocationMessage(invocationId, target, arguments, nonBlocking);
+        }
+
+        private StreamItemMessage BindResultMessage(JObject json, IInvocationBinder binder)
+        {
+            var invocationId = json.Value<string>(InvocationIdPropertyName);
+            var payload = json.Value<JToken>(ResultPropertyName);
+
+            var returnType = binder.GetReturnType(invocationId);
+            return new StreamItemMessage(invocationId, payload?.ToObject(returnType, _payloadSerializer));
```
anurse: Yep, you could want to stream the value `null`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/CompletionMessage.cs
moozzyk: I forgot to remove this comment. Sorry.

src/Microsoft.AspNetCore.Sockets/ConnectionMetadataNames.cs
anurse: IMO, no, ordinal is quicker and these are tokens that are designed to be used from this static list.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
anurse: Good point, I had some code in TestConnection that required calling StartAsync at one point, I'll put that in because it's best to emulate the same call sequence.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: I think we should just create HubException or something to this effect. https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Client/HubException.cs

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
moozzyk: Then the result should be `null`, no?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: This is a case where we can and should synchronously fail.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: They don't today. But we can add that.

anurse: https://github.com/aspnet/SignalR/issues/420

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -70,7 +70,7 @@ public HubConnection(IConnection connection, IInvocationAdapter adapter, ILogger
 
             _connection = connection;
             _binder = new HubBinder(this);
-            _adapter = adapter;
+            _protocol = protocol;
```
anurse: I'd rather check `protocol` for null, since we have overloads that set the default value.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -204,69 +217,114 @@ private void Shutdown(Exception ex = null)
...
+            if (_connectionActive.IsCancellationRequested)
+            {
+                throw new InvalidOperationException("Connection has been terminated.");
+            }
+
+            lock (_pendingCallsLock)
+            {
+                if (_pendingCalls.ContainsKey(invocationId))
+                {
+                    throw new InvalidOperationException("An invocation with this ID has already been queued");
```
anurse: I've changed back to a ConcurrentDictionary, which means `TryAdd` gives this to us for free. I'd rather manifest errors like this when they happen, whereas `Debug.Assert` is hidden from the user in Release builds.

It's the difference between getting a bug report that says "After sending 4 billion messages, I got an Exception saying duplicate invocation IDs were generated" and getting one that says "After sending 4 billion messages, weird stuff started happening" :). Yeah, it's super unlikely, but I'd rather know what's happening.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
anurse: Yeah, we could require that.

anurse:
:up: :date:

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
anurse: I actually don't think I like requiring the property to be present even when the value is null. Certainly our clients/servers can work that way, but most JSON formats generally consider a missing property to be equal to `null`, unless there's a specific reason to treat them differently.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -120,19 +125,24 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
             {
+                _logger.LogError("Invoke was called after the connection was terminated");
                 throw new InvalidOperationException("Connection has been terminated.");
             }
 
             _logger.LogTrace("Preparing invocation of '{0}', with return type '{1}' and {2} args", methodName, returnType.AssemblyQualifiedName, args.Length);
 
             // Create an invocation descriptor. Client invocations are always blocking
-            var invocationMessage = new InvocationMessage(GetNextId().ToString(), methodName, args, nonBlocking: false);
+            var invocationMessage = new InvocationMessage(GetNextId().ToString(), nonBlocking: false, target: methodName, arguments: args);
```
moozzyk: `GetNextId()` already returns string so `.ToString()` is not needed.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -219,19 +219,24 @@ private void Shutdown(Exception ex = null)
 
             _connectionActive.Cancel();
 
-            lock (_pendingCallsLock)
+            // Keys is a snapshot, so we'll need to run this until other threads stop adding (which they will because of the cancellation token above)
```
moozzyk: Isn't it racy though? If you expect/allow other threads to add while you are in this loop the threads can be scheduled in a way that _pendingCalls.Count is 0 so you will exit the loop after which a new item will be added.

src/Microsoft.AspNetCore.SignalR.Client/HubException.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    [Serializable]
+    public class HubException : Exception
```
moozzyk: 👍

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -36,6 +37,29 @@ public InvocationMessage(string invocationId, string target, object[] arguments,
             NonBlocking = nonBlocking;
         }
 
+        public override bool Equals(object obj)
+        {
+            return obj is InvocationMessage m && Equals(m);
+        }
+
+        public override int GetHashCode()
```
moozzyk: Why do we need GetHashCode and Equals?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -36,6 +37,29 @@ public InvocationMessage(string invocationId, string target, object[] arguments,
...
+        {
+            var combiner = new HashCodeCombiner();
+            combiner.Add(InvocationId);
+            combiner.Add(Target);
+            combiner.Add(Arguments);
+            combiner.Add(NonBlocking);
+            return combiner.CombinedHash;
+        }
+
+        public bool Equals(InvocationMessage other)
```
moozzyk: blows up for null

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -36,6 +37,29 @@ public InvocationMessage(string invocationId, string target, object[] arguments,
             NonBlocking = nonBlocking;
         }
 
+        public override bool Equals(object obj)
```
moozzyk: nit: group Equals

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -36,6 +37,29 @@ public InvocationMessage(string invocationId, string target, object[] arguments,
...
+            combiner.Add(InvocationId);
+            combiner.Add(Target);
+            combiner.Add(Arguments);
+            combiner.Add(NonBlocking);
+            return combiner.CombinedHash;
+        }
+
+        public bool Equals(InvocationMessage other)
+        {
+            return string.Equals(InvocationId, other.InvocationId, StringComparison.Ordinal) &&
```
moozzyk: You may want to do reference comparison first - should also take care for the null case.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/StreamItemMessage.cs
```diff
@@ -13,9 +15,28 @@ public StreamItemMessage(string invocationId, object item) : base(invocationId)
             Item = item;
         }
 
+        public override bool Equals(object obj)
```
moozzyk: nit: group equals

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/StreamItemMessage.cs
```diff
@@ -13,9 +15,28 @@ public StreamItemMessage(string invocationId, object item) : base(invocationId)
...
+        public override int GetHashCode()
+        {
+            var combiner = new HashCodeCombiner();
+            combiner.Add(InvocationId);
+            combiner.Add(Item);
+            return combiner.CombinedHash;
+        }
+
+        public bool Equals(StreamItemMessage other)
+        {
```
moozzyk: blows up for null. add reference comparison.

src/Microsoft.AspNetCore.SignalR.Common/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -18,6 +18,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.HashCodeCombiner.Sources" Version="$(AspNetCoreVersion)" />
```
moozzyk: I was not aware of this class.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -159,11 +159,7 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
                 {
                     while (connection.Transport.Input.TryRead(out var incomingMessage))
                     {
-                        if (!protocol.TryParseMessage(incomingMessage.Payload, this, out var hubMessage))
-                        {
-                            _logger.LogError("Received invalid message");
-                            throw new InvalidOperationException("Received invalid message");
-                        }
+                        var hubMessage = protocol.ParseMessage(incomingMessage.Payload, this);
```
moozzyk: Handle null values returned by ParseMessage (+test if possible) - https://github.com/aspnet/SignalR/pull/390/commits/fa7ead1f51ac45c4cc928d6725e43cedc45d66b9#diff-17d494f97b6bf3a80750f65894639dccR77 (or throw at the source?)

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -64,30 +68,24 @@ public bool TryWriteMessage(HubMessage message, IOutput output)
...
-            if (json == null)
+            using (var reader = new JsonTextReader(new StreamReader(input)))
             {
-                return null;
-            }
+                // PERF: Could probably use the JsonTextReader directly for better perf and fewer allocations
+                var json = _payloadSerializer.Deserialize<JObject>(reader);
+                if (json == null)
+                {
+                    return null;
```
moozzyk: Either handle this (e.g here: https://github.com/aspnet/SignalR/pull/390/commits/fa7ead1f51ac45c4cc928d6725e43cedc45d66b9#diff-b8bab5f1c287065aa78fe0918014bb98R162) or maybe just throw here?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -10,138 +13,138 @@ namespace Microsoft.AspNetCore.SignalR.Common.Tests.Internal.Protocol
...
-        [InlineData(true, NullValueHandling.Include, "{\"invocationId\":\"123\",\"type\":2,\"result\":{\"stringProp\":\"SignalR!\",\"doubleProp\":6.2831853071,\"intProp\":42,\"dateTimeProp\":\"2017-04-11T00:00:00\",\"nullProp\":null}}")]
-        public async Task CanSerializeCustomObjectInResult(bool camelCase, NullValueHandling nullHandling, string expectedOutput)
+        [MemberData(nameof(ProtocolTestData))]
+        public void ParseMessage(HubMessage expectedMessage, bool camelCase, NullValueHandling nullValueHandling, string input)
         {
-            var message = new StreamItemMessage("123", new CustomObject());
-
-            var jsonSerializer = new JsonSerializer();
-            if(camelCase)
+            var jsonSerializer = new JsonSerializer()
```
moozzyk: nit: you can remove `()`

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -10,138 +13,138 @@ namespace Microsoft.AspNetCore.SignalR.Common.Tests.Internal.Protocol
...
-            var json = Encoding.UTF8.GetString(encoded);
+            public override int GetHashCode()
+            {
+                // This is never used in a hash table
+                return 0;
+            }
 
-            Assert.Equal(expectedOutput, json);
+            public bool Equals(CustomObject other)
+            {
```
moozzyk: Why do we need it? If we do - is it now a requirement to make custom types equatable if they can be invocation results?

anurse: It's just for the test.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
moozzyk: Just add a comment.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: Yeah... I guess I'll just go back to the locking version.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
anurse: It was for testing, but I think I've got a better way.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
anurse: Won't matter when I refactor this code into the tests.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: I'll add a branch to the `switch` to handle it.

anurse:
:up: :date: The full build won't work because I had to disable some things in the TypeScript projects due to an MSBuild regression (which has been fixed, we just haven't taken an updated CLI)

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -118,81 +123,80 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
...
-                Method = methodName,
-                Arguments = args
-            };
+            // Create an invocation descriptor. Client invocations are always blocking
+            var invocationMessage = new InvocationMessage(GetNextId(), nonBlocking: false, target: methodName, arguments: args);
 
             // I just want an excuse to use 'irq' as a variable name...
-            _logger.LogDebug("Registering Invocation ID '{0}' for tracking", descriptor.Id);
-            var irq = new InvocationRequest(cancellationToken, returnType);
+            _logger.LogDebug("Registering Invocation ID '{0}' for tracking", invocationMessage.InvocationId);
```
BrennanConroy: Since you're here... make `{0}`  `{id}`

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -206,68 +210,112 @@ private void Shutdown(Exception ex = null)
...
 
-        private void DispatchInvocation(InvocationDescriptor invocationDescriptor, CancellationToken cancellationToken)
+        private void DispatchInvocation(InvocationMessage invocation, CancellationToken cancellationToken)
         {
             // Find the handler
-            if (!_handlers.TryGetValue(invocationDescriptor.Method, out InvocationHandler handler))
+            if (!_handlers.TryGetValue(invocation.Target, out InvocationHandler handler))
             {
-                _logger.LogWarning("Failed to find handler for '{0}' method", invocationDescriptor.Method);
+                _logger.LogWarning("Failed to find handler for '{0}' method", invocation.Target);
```
BrennanConroy: same

src/Microsoft.AspNetCore.SignalR.Client/HubException.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
BrennanConroy: Copyright

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        public string Target { get; }
+
+        public object[] Arguments { get; }
+
+        public bool NonBlocking { get; }
+
+        public InvocationMessage(string invocationId, bool nonBlocking, string target, params object[] arguments) : base(invocationId)
+        {
+            if(string.IsNullOrEmpty(invocationId))
```
BrennanConroy: space

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,277 @@
...
+            }
+        }
+
+        public bool TryWriteMessage(HubMessage message, IOutput output)
+        {
+            // TODO: Need IOutput-compatible JSON serializer!
+            using (var memoryStream = new MemoryStream())
+            {
+                WriteMessage(message, memoryStream);
+                memoryStream.FlushAsync();
```
BrennanConroy: FlushAsync without any kind of wait?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,277 @@
...
+                        throw new FormatException($"Unexpected JSON Token Type '{token.Type}'. Expected a JSON Object.");
+                    }
+
+                    var json = (JObject)token;
+
+                    // Determine the type of the message
+                    var type = GetRequiredProperty<int>(json, TypePropertyName, JTokenType.Integer);
+                    switch (type)
+                    {
+                        case InvocationMessageType: return BindInvocationMessage(json, binder);
```
BrennanConroy: nit: newline after case:

src/Microsoft.AspNetCore.Sockets/ConnectionMetadataNames.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.AspNetCore.Sockets
```
BrennanConroy: copyright

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/IHubProtocol.cs
moozzyk: Not addressed.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -150,10 +150,34 @@ public class HubEndpointTests
...
+
+            using (var client = new TestClient(serviceProvider))
+            {
+                var endPointTask = endPoint.OnConnectedAsync(client.Connection);
+
+                var result = (await client.InvokeAsync(methodName).OrTimeout());
 
-                Assert.Null(result.Error);
-                Assert.Equal("hello", result.Result);
+                // json serializer makes this a long
```
moozzyk: Remove?

anurse:
@moozzyk Yeah, I shouldn't have squashed. Things were a little crazy to rebase at one point and squashing made that easier, but it does make incremental review harder. Will avoid in the future.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: Derp. Removed.

anurse:
Taking @moozzyk 's comment as a grudging approval ;). Since folks are busy with BUILD stuff, I'll look at updating the TypeScript PR as well and getting that merged in :)

Issue #48, opened 2016-11-24
Is it possible to use generic Hub<IClientProxy>?


xperiandri:
Is it possible to use a generic SignalR hub from Microsoft.AspNetCore.SignalR?

``` C#
public class DriverHub : Hub<IDriverClient>
```

I cannot instantiate it
I have to map it this way

``` C#
            //app.UseSignalR(routes =>
            //{
            //    routes.MapHub<PassengersHub>("/Passenger");
            //    routes.MapHub<DriversHub>("/Driver");
            //});
            app.UseSockets(routes =>
            {
                routes.MapEndpoint<HubEndPoint<PassengerHub, IPassengerClient>>(Routes.PassengerArea1);
                routes.MapEndpoint<HubEndPoint<DriverHub, IDriverClient>>(Routes.DriverArea1);
            });
```

But adding

``` C#
            services.AddSignalR();
            // Missing in AddSignalR
            services.AddSingleton(typeof(HubEndPoint<,>), typeof(HubEndPoint<,>));
            services.AddSingleton(typeof(IHubContext<,>), typeof(HubContext<>));
```

Does not help, because of generic arguments count mismatch

xperiandri:
As there are no dynamic objects it is too verbose to call `InvokeAsync`

davidfowl:
We'll support both Hub<dynamic> and Hub<someinterface> but they haven't been implemented yet.

muratg:
Dup of https://github.com/aspnet/SignalR/issues/555 and https://github.com/aspnet/SignalR/issues/554

PR #391, opened 2017-04-13
Presence sample

moozzyk:
Adding a sample of how to handle presence in single server and scaleout scenarios

### DIFF
--- a/samples/ChatSample/DefaultPresenceManager.cs
+++ b/samples/ChatSample/DefaultPresenceManager.cs
@@ -1,37 +1,42 @@
-﻿
-using Microsoft.AspNetCore.SignalR;
-using Microsoft.Extensions.DependencyInjection;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
-using System;
+using Microsoft.AspNetCore.SignalR;
 using Microsoft.AspNetCore.Sockets;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
 namespace ChatSample
 {
-    // TODO: not possible to use TClient instead of (implicit) IClientProxy
-    // public class DefaultPresenceManager<THub> : IPresenceManager where THub : HubWithPresence<TClient>
     public class DefaultPresenceManager<THub> : IPresenceManager where THub : HubWithPresence
     {
         private IHubContext<THub> _hubContext;
         private HubLifetimeManager<THub> _lifetimeManager;
         private readonly IServiceScopeFactory _serviceScopeFactory;
+        private readonly ILogger _logger;
 
-        public DefaultPresenceManager(IHubContext<THub> hubContext, HubLifetimeManager<THub> lifetimeManager, IServiceScopeFactory serviceScopeFactory)
+        public DefaultPresenceManager(IHubContext<THub> hubContext, HubLifetimeManager<THub> lifetimeManager,
+            IServiceScopeFactory serviceScopeFactory, ILoggerFactory loggerFactory)
         {
             _hubContext = hubContext;
             _lifetimeManager = lifetimeManager;
             _serviceScopeFactory = serviceScopeFactory;
+            _logger = loggerFactory.CreateLogger<DefaultPresenceManager<THub>>();
         }
 
-        private readonly ConcurrentDictionary<Connection, UserDetails> usersOnline 
+        private readonly ConcurrentDictionary<Connection, UserDetails> usersOnline
             = new ConcurrentDictionary<Connection, UserDetails>();
 
-        public IEnumerable<UserDetails> UsersOnline => usersOnline.Values;
+        public Task<IEnumerable<UserDetails>> UsersOnline()
+            => Task.FromResult(usersOnline.Values.AsEnumerable());
 
         public async Task UserJoined(Connection connection)
         {
-            // `context.User?.Identity?.Name ?? string.Empty` ?
             var user = new UserDetails(connection.ConnectionId, connection.User.Identity.Name);
 
             await Notify(hub => hub.OnUserJoined(user));
@@ -41,9 +46,10 @@ namespace ChatSample
 
         public async Task UserLeft(Connection connection)
         {
-            usersOnline.TryRemove(connection, out UserDetails user);
-
-            await Notify(hub => hub.OnUserLeft(user));
+            if (usersOnline.TryRemove(connection, out UserDetails user))
+            {
+                await Notify(hub => hub.OnUserLeft(user));
+            }
         }
 
         private async Task Notify(Func<THub, Task> invocation)
@@ -63,9 +69,9 @@ namespace ChatSample
                     {
                         await invocation(hub);
                     }
-                    catch
+                    catch (Exception ex)
                     {
-                        // TODO: log
+                        _logger.LogWarning(ex, "Presence notification failed.");
                     }
                     finally
                     {
--- a/samples/ChatSample/DefaultPresenceManager.cs
+++ b/samples/ChatSample/DefaultPresenceManager.cs
@@ -1,0 +1,78 @@
+﻿
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.DependencyInjection;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System;
+using Microsoft.AspNetCore.Sockets;
+
+namespace ChatSample
+{
+    // TODO: not possible to use TClient instead of (implicit) IClientProxy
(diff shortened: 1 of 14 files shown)


moozzyk:
:up::date: - cleaned up

moozzyk:
:watch:

samples/ChatSample/DefaultPresenceManager.cs
```diff
@@ -0,0 +1,83 @@
...
+            var user = new UserDetails(connection.ConnectionId, connection.User.Identity.Name);
+
+            await Notify(hub => hub.OnUserJoined(user));
+
+            usersOnline.TryAdd(connection, user);
+        }
+
+        public async Task UserLeft(Connection connection)
+        {
+            usersOnline.TryRemove(connection, out UserDetails user);
```
BrennanConroy: Wrap this in an if statement and only execute Notify if true

moozzyk: Fixed.

samples/ChatSample/DefaultPresenceManager.cs
```diff
@@ -0,0 +1,84 @@
...
+        public DefaultPresenceManager(IHubContext<THub> hubContext, HubLifetimeManager<THub> lifetimeManager,
+            IServiceScopeFactory serviceScopeFactory, ILoggerFactory loggerFactory)
+        {
+            _hubContext = hubContext;
+            _lifetimeManager = lifetimeManager;
+            _serviceScopeFactory = serviceScopeFactory;
+            _logger = loggerFactory.CreateLogger<DefaultPresenceManager<THub>>();
+        }
+
+        private readonly ConcurrentDictionary<Connection, UserDetails> usersOnline
```
mikaelm12: rename to `_usersOnline`

mikaelm12: rename to `_usersOnline`

samples/ChatSample/DefaultPresenceManager.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                await Notify(hub => hub.OnUserLeft(user));
+            }
+        }
+
+        private async Task Notify(Func<THub, Task> invocation)
+        {
+            foreach (var connection in usersOnline.Keys)
+            {
+                using (var scope = _serviceScopeFactory.CreateScope())
```
davidfowl: This is the grossest part of this change. Creating a new hub instance here feels kinda wrong as it requires mimicking the HubEndPoint logic... At least it's a sample and we get to see what's required to make it work.

src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
```diff
@@ -32,9 +32,14 @@ public HubRouteBuilder(SocketRouteBuilder routes)
             _routes = routes;
         }
 
-        public void MapHub<THub>(string path) where THub : Hub
+        public void MapHub<THub>(string path) where THub : Hub<IClientProxy>
```
davidfowl: why?

src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
```diff
@@ -32,9 +32,14 @@ public HubRouteBuilder(SocketRouteBuilder routes)
...
         {
             _routes.MapEndpoint<HubEndPoint<THub>>(path);
         }
+
+        public void MapHub<THub, TClient>(string path) where THub : Hub<TClient>
```
davidfowl: Isn't this overload broken?

davidfowl:
I think you should make derived a `HubLifetimeManager` that implements the `IPresenceManager` since all of the state is already there. That should remove the need to pass things in the `HubWithPresence` ctor. We also need to figure out how to help people distinguish between UserOnline and OnConnected (same for disconnected, offline)

samples/ChatSample/RedisPresenceManager.cs
```diff
@@ -0,0 +1,175 @@
...
+        private readonly ConnectionMultiplexer _redisConnection;
+        private readonly ISubscriber _redisSubscriber;
+        private readonly ILogger _logger;
+
+        private readonly ConcurrentDictionary<Connection, object> connections
+            = new ConcurrentDictionary<Connection, object>();
+
+        // TODO: subscribe and handle lifecycle events/disconnects
+        // TODO: handle situations where Redis shuts down
+        // TODO: handle situations where a server goes down and the server has zombie connections
```
davidfowl: This is key to the entire feature

moozzyk: This is not part of this PR.

src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
moozzyk: I started with `DefaultPresenceManager<THub>` was `DefaultPresenceManager<THub, TClient>` which required this but this indeed did not work (I think I hit: https://github.com/aspnet/SignalR/issues/48).  Fixing https://github.com/aspnet/SignalR/issues/48 did not seem straightforward and was not the main point of this PR. I will remove this overload for now.

src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
moozzyk: I think it was related to the other change. Seems no longer needed when I dropped the TClient.

samples/ChatSample/DefaultPresenceManager.cs
moozzyk: Yeah. There is not a good way to invoke a hub method from outside.

samples/ChatSample/RedisPresenceManager.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            _logger = loggerFactory.CreateLogger<RedisPresenceManager<THub>>();
+            _redisDatabase = options.Value.Options.DefaultDatabase.GetValueOrDefault();
+            _redisConnection = ConnectToRedis(options.Value, _logger);
+            _redisSubscriber = _redisConnection.GetSubscriber();
+            _redisChannel = new RedisChannel((string)UsersOnlineRedisKey, RedisChannel.PatternMode.Literal);
+
+            _redisSubscriber.Subscribe(_redisChannel, (channel, value) =>
+            {
+                var user = ToUserDetails(((string)value).Substring(1));
```
moozzyk: We can. It's not a cast btw. it is a conversion operator so I don't even know how expensive it is.

samples/ChatSample/Views/Home/Index.cshtml
```diff
@@ -34,6 +64,14 @@ document.getElementById('sendmessage').addEventListener('submit', event => {
         document.getElementById('messages').appendChild(child);
     });
 
+    newMessage.value = ''; 
     event.preventDefault();
 });
+
+let addUserOnline = function (user) {
```
moozzyk: It was introduced by the person who originally submitted the PR and I did not notice it. I don't think it needs to be a variable at all.

src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
moozzyk: This is actually needed if you try to map a hub the derives from `Hub<IClientProxy>` instead of `Hub`.  I have `HubWithPresence : HubWithPresence<IClientProxy>` and `HubWithPresence<TClient> : Hub<TClient>` and it is causing an eror: `There is no implicit reference conversion from 'ChatSample.Hubs.Chat' to 'Microsoft.AspNetCore.SignalR.Hub'.` I can create `HubWithPresence : Hub` and it will work but the other should work as well.

samples/ChatSample/DefaultPresenceManager.cs
```diff
@@ -0,0 +1,84 @@
...
+            var user = new UserDetails(connection.ConnectionId, connection.User.Identity.Name);
+
+            await Notify(hub => hub.OnUserJoined(user));
+
+            usersOnline.TryAdd(connection, user);
+        }
+
+        public async Task UserLeft(Connection connection)
+        {
+            if (usersOnline.TryRemove(connection, out UserDetails user))
```
BrennanConroy: out var

samples/ChatSample/RedisPresenceManager.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            return Task.CompletedTask;
+        }
+
+        public Task UserLeft(Connection connection)
+        {
+            connections.TryRemove(connection, out object _);
+
+            var database = _redisConnection.GetDatabase(_redisDatabase);
+            var user = $"{connection.ConnectionId }|{ connection.User.Identity.Name}";
```
BrennanConroy: format

moozzyk:
Rebased and addressed most of the feeback. Working on _make derived a HubLifetimeManager that implements the IPresenceManager_ which I will send separately.

moozzyk:
I accidentally delete the branch which closed the PR. I opened a new, follow up PR here: https://github.com/aspnet/SignalR/pull/391

Issue #383, opened 2017-04-11
Change JavaScript to use new Transport protocol for send


anurse:
Right now, the JavaScript client is just relying on the legacy behavior here in HttpConnectionDispatcher: https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs#L354

I'd like to remove that code soon, so we should update the JavaScript client. We'll need to write a Text message formatter in JavaScript, but we've already got the parser so it shouldn't be too hard.

/cc @mikaelm12 @moozzyk @davidfowl

anurse:
Sorry if there's already a bug tracking this, but I didn't see it. I wanted to have a bug tracking it so we know to remove the legacy code in HttpConnectionDispatcher when the bug is closed.

moozzyk:
I will do that

moozzyk:
Fixed in e691e1cff83cce9ab1fd294ee787aba822ff0e7b

Issue #127, opened 2017-01-13
Don't use application/x-www-form-urlencoded POST requests for http polling


davidfowl:
CSRF is a concern here. We should just post raw data to the body.

anurse:
See #125 

How does not using `application/x-www-form-urlencoded` prevent CSRF? We need an anti-forgery token no matter how we encode the data if we want to prevent CSRF.

anurse:
Wait, never mind, this is for the poll request. Just got it ;). Never mind!

davidfowl:
@anurse sends and poll. This is a long standing issue https://github.com/SignalR/SignalR/issues/1886.

davidfowl:
@anurse we can close this right?

moozzyk:
No - https://github.com/aspnet/SignalR/blob/a00ec445ef040395faed0b564f098258b8239e60/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts#L21 - I could just remove it but I wanted to understand first how this can be a CSRF and I keep forgetting to ask you.

davidfowl:
Here's the original issue - https://github.com/SignalR/SignalR/issues/1413

https://github.com/SignalR/SignalR/commit/f41abf4a17c91e3f18b5fc8cbdff9c0d3297645a

moozzyk:
Fixed in e691e1c

PR #398, opened 2017-04-17
Removing legacy raw format for `send`

moozzyk:
Fixes issues #383 and #127

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
@@ -338,31 +338,24 @@ namespace Microsoft.AspNetCore.Sockets
                 buffer = stream.ToArray();
             }
 
-            IList<Message> messages;
+            MessageFormat messageFormat;
             if (string.Equals(context.Request.ContentType, MessageFormatter.TextContentType, StringComparison.OrdinalIgnoreCase))
             {
-                var reader = new BytesReader(buffer);
-                messages = ParseSendBatch(ref reader, MessageFormat.Text);
+                messageFormat = MessageFormat.Text;
             }
             else if (string.Equals(context.Request.ContentType, MessageFormatter.BinaryContentType, StringComparison.OrdinalIgnoreCase))
             {
-                var reader = new BytesReader(buffer);
-                messages = ParseSendBatch(ref reader, MessageFormat.Binary);
+                messageFormat = MessageFormat.Binary;
             }
             else
             {
-                // Legacy, single message raw format
-
-                var format =
-                    string.Equals(context.Request.Query["format"], "binary", StringComparison.OrdinalIgnoreCase)
-                        ? MessageType.Binary
-                        : MessageType.Text;
-                messages = new List<Message>()
-                {
-                    new Message(buffer, format, endOfMessage: true)
-                };
+                context.Response.StatusCode = StatusCodes.Status400BadRequest;
+                await context.Response.WriteAsync($"'{context.Request.ContentType}' is not a valid Content-Type for send requests.");
+                return;
             }
 
+            var reader = new BytesReader(buffer);
+            var messages = ParseSendBatch(ref reader, messageFormat);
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -338,31 +338,12 @@ private ConnectionState CreateConnection(HttpContext context)
...
-                        : MessageType.Text;
-                messages = new List<Message>()
-                {
-                    new Message(buffer, format, endOfMessage: true)
-                };
-            }
-
+            var reader = new BytesReader(buffer);
+            var messageFormat =
+                string.Equals(context.Request.ContentType, MessageFormatter.TextContentType, StringComparison.OrdinalIgnoreCase)
```
anurse: We should return `400 Bad Request` if the content type doesn't match either type.

moozzyk:
:up::date:

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -378,6 +359,28 @@ private ConnectionState CreateConnection(HttpContext context)
             }
         }
 
+        private async Task<IList<Message>> ExtractMessages(HttpContext context, byte[] buffer)
```
BrennanConroy: Since this is all internal could we change this to the concrete `List<T>`?

davidfowl: This isn't cachable. We should avoid returning task of t in internal apis

moozzyk: I could inline it back. I only extracted the method because the code is more readable. I could also use ValueTask<T> if it makes it better.

moozzyk: @BrennanConroy  - I will update ParseSendBatch to return `List<Message>` instead of `IList<Message>` because this is where IList is being returned initially from.

anurse: You could also move the writing of the response to outside this method. Other methods do it this way, but I'm not sure I like it there either (even though I wrote it. That way you'd be doing the awaiting in the method that returns `Task` rather than `Task<T>` which **is** cacheable.

moozzyk:
:up::date:

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
moozzyk: This way the code is magic because I would report a detailed error in a different place that I detected it. It may work initially but as code evolves it will be hard to keep the two in sync. Also, we already have methods which return `Task<T>` for exactly the same reason I did - writing to the response in case of error:  https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs#L407

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -338,31 +338,29 @@ private ConnectionState CreateConnection(HttpContext context)
...
-                    new Message(buffer, format, endOfMessage: true)
-                };
+                context.Response.StatusCode = StatusCodes.Status400BadRequest;
+                await context.Response.WriteAsync($"'{context.Request.ContentType}' is not a valid Content-Type for send requests.");
+                return;
             }
 
+            var reader = new BytesReader(buffer);
+            var messages = ParseSendBatch(ref reader, messageFormat);
+            if (messages == null)
```
anurse: This can't be true anymore, right? We can just remove it.

moozzyk: Good catch!

PR #401, opened 2017-04-18
SSE Transport

mikaelm12:
- ServerSentEventsTransport
- ServerSentEventsParser
- Some basic unit tests

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
@@ -27,9 +27,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly HttpClient _httpClient;
         private readonly ILogger _logger;
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+        private readonly ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
+
 
         private IChannelConnection<SendMessage, Message> _application;
-        private ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
 
         public Task Running { get; private set; } = Task.CompletedTask;
 
@@ -69,7 +70,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             return TaskCache.CompletedTask;
         }
 
-        public async Task OpenConnection(IChannelConnection<SendMessage, Message> application, Uri url, CancellationToken cancellationToken)
+        private async Task OpenConnection(IChannelConnection<SendMessage, Message> application, Uri url, CancellationToken cancellationToken)
         {
             _logger.LogInformation("Starting receive loop");
 
@@ -82,43 +83,42 @@ namespace Microsoft.AspNetCore.Sockets.Client
             var pipelineReader = stream.AsPipelineReader();
             try
             {
-
-            while (true)
-            {
-                var result = await pipelineReader.ReadAsync();
-                var input = result.Buffer;
-                var consumed = input.Start;
-                var examined = input.End;
-
-                try
+                while (true)
                 {
-                    if (input.IsEmpty && result.IsCompleted)
+                    var result = await pipelineReader.ReadAsync();
+                    var input = result.Buffer;
+                    var consumed = input.Start;
+                    var examined = input.End;
+
+                    try
                     {
-                        _logger.LogDebug("Done reading from pipeline");
-                        break;
+                        if (input.IsEmpty && result.IsCompleted)
+                        {
+                            _logger.LogDebug("Server-Sent Event Stream ended");
+                            break;
+                        }
+
+                        var parseResult = _parser.ParseMessage(input, out consumed, out examined, out var message);
+
+                        switch (parseResult)
+                        {
+                            case ServerSentEventsMessageParser.ParseResult.Completed:
+                                _application.Output.TryWrite(message);
+                                _parser.Reset();
+                                break;
+                            case ServerSentEventsMessageParser.ParseResult.Incomplete:
+                                if (result.IsCompleted)
+                                {
+                                    throw new FormatException("Incomplete message");
+                                }
+                                break;
+                        }
                     }
-
-                    var parseResult = _parser.ParseMessage(input, out consumed, out examined, out var message);
-
-                    switch (parseResult)
+                    finally
                     {
-                        case ServerSentEventsMessageParser.ParseResult.Completed:
-                            _application.Output.TryWrite(message);
-                            _parser.Reset();
-                            break;
-                        case ServerSentEventsMessageParser.ParseResult.Incomplete:
-                            if (result.IsCompleted)
-                            {
-                                throw new FormatException("Error parsing data from event stream");
-                            }
-                            break;
+                        pipelineReader.Advance(consumed, examined);
                     }
                 }
-                finally
-                {
-                    pipelineReader.Advance(consumed, examined);
-                }
-            }
             }
             finally
             {
@@ -130,7 +130,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         {
             _logger.LogInformation("Starting the send loop");
 
-            IList<SendMessage> messages = null;
+            List<SendMessage> messages = null;
             try
             {
                 while (await _application.Input.WaitToReadAsync(cancellationToken))
@@ -203,7 +203,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             _logger.LogInformation("Send loop stopped");
         }
 
-        private async Task WriteMessagesAsync(IList<SendMessage> messages, PipelineTextOutput output, MessageFormat format)
+        private async Task WriteMessagesAsync(List<SendMessage> messages, PipelineTextOutput output, MessageFormat format)
         {
             output.Append(MessageFormatter.GetFormatIndicator(format), TextEncoder.Utf8);
 
--- a/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
@@ -29,7 +29,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
         private readonly ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
 
-
         private IChannelConnection<SendMessage, Message> _application;
 
         public Task Running { get; private set; } = Task.CompletedTask;
--- a/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
@@ -50,7 +50,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public Task StartAsync(Uri url, IChannelConnection<SendMessage, Message> application)
         {
-            _logger.LogInformation("Starting {0}", nameof(ServerSentEventsTransport));
+            _logger.LogInformation("Starting {transportName}", nameof(ServerSentEventsTransport));
 
             _application = application;
             var sseUrl = Utils.AppendPath(url, "sse");
@@ -60,7 +60,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
             {
-                _logger.LogDebug("Transport stopped. Exception: '{0}'", t.Exception?.InnerException);
+                if (t.Exception != null) { _logger.LogError(t.Exception, "Transport stopped"); }
 
                 _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
                 return t;
@@ -125,7 +125,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
         }
 
-        public async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
+        private async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
         {
             _logger.LogInformation("Starting the send loop");
 
@@ -142,7 +142,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
                     if (messages.Count > 0)
                     {
-                        _logger.LogDebug("Sending {0} message(s) to the server using url: {1}", messages.Count, sendUrl);
+                        _logger.LogDebug("Sending {messageCount} message(s) to the server using url: {url}", messages.Count, sendUrl);
 
                         var request = new HttpRequestMessage(HttpMethod.Post, sendUrl);
                         request.Headers.UserAgent.Add(DefaultUserAgentHeader);
@@ -183,7 +183,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
             catch (Exception ex)
             {
-                _logger.LogDebug("Error while sending to '{0}' : '{1}'", sendUrl, ex);
+                _logger.LogDebug("Error while sending to '{url}' : '{exception}'", sendUrl, ex);
                 if (messages != null)
                 {
                     foreach (var message in messages)
@@ -208,7 +208,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             foreach (var message in messages)
             {
-                _logger.LogDebug("Writing '{0}' message to the server", message.Type);
+                _logger.LogDebug("Writing '{messageType}' message to the server", message.Type);
 
                 var payload = message.Payload ?? Array.Empty<byte>();
                 if (!MessageFormatter.TryWriteMessage(new Message(payload, message.Type, endOfMessage: true), output, format))
@@ -222,7 +222,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public async Task StopAsync()
         {
-            _logger.LogInformation("Transport {0} is stopping", nameof(ServerSentEventsTransport));
+            _logger.LogInformation("Transport {transportName} is stopping", nameof(ServerSentEventsTransport));
             _transportCts.Cancel();
             _application.Output.TryComplete();
             await Running;
--- a/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
@@ -1,0 +1,232 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Pipelines;
+using System.IO.Pipelines.Text.Primitives;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Text.Formatting;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public class ServerSentEventsTransport : ITransport
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
+        private static readonly ProductInfoHeaderValue DefaultUserAgentHeader = ProductInfoHeaderValue.Parse(DefaultUserAgent);
+
+        private readonly HttpClient _httpClient;
+        private readonly ILogger _logger;
+        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+
+        private IChannelConnection<SendMessage, Message> _application;
+        private ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
+
+        public Task Running { get; private set; } = Task.CompletedTask;
+
+        public ServerSentEventsTransport(HttpClient httpClient)
+            : this(httpClient, null)
+        { }
+
+        public ServerSentEventsTransport(HttpClient httpClient, ILoggerFactory loggerFactory)
+        {
+            if (httpClient == null)
+            {
+                throw new ArgumentNullException(nameof(_httpClient));
+            }
+
+            _httpClient = httpClient;
+            _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<ServerSentEventsTransport>();
+        }
+
+        public Task StartAsync(Uri url, IChannelConnection<SendMessage, Message> application)
+        {
+            _logger.LogInformation("Starting {0}", nameof(ServerSentEventsTransport));
+
+            _application = application;
+            var sseUrl = Utils.AppendPath(url, "sse");
+            var sendUrl = Utils.AppendPath(url, "send");
+            var sendTask = SendMessages(sendUrl, _transportCts.Token);
+            var receiveTask = OpenConnection(_application, sseUrl, _transportCts.Token);
+
+            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
+            {
+                _logger.LogDebug("Transport stopped. Exception: '{0}'", t.Exception?.InnerException);
+
+                _application.Output.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
+                return t;
+            }).Unwrap();
+
+            return TaskCache.CompletedTask;
+        }
+
+        public async Task OpenConnection(IChannelConnection<SendMessage, Message> application, Uri url, CancellationToken cancellationToken)
+        {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,231 @@
...
+                        var memoryStream = new MemoryStream();
+
+                        var pipe = memoryStream.AsPipelineWriter();
+                        var output = new PipelineTextOutput(pipe, TextEncoder.Utf8);
+                        await WriteMessagesAsync(messages, output, MessageFormat.Binary);
+
+                        memoryStream.Seek(0, SeekOrigin.Begin);
+
+                        request.Content = new StreamContent(memoryStream);
+                        request.Content.Headers.ContentType = MediaTypeHeaderValue.Parse(MessageFormatter.GetContentType(MessageFormat.Binary));
```
davidfowl: Why is this hard coded to binary?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
...
+        const byte ByteLF = (byte)'\n';
+        const byte ByteT = (byte)'T';
+        const byte ByteB = (byte)'B';
+        const byte ByteC = (byte)'C';
+        const byte ByteE = (byte)'E';
+
+        private static byte[] _dataPrefix = Encoding.UTF8.GetBytes("data: ");
+        private static byte[] _sseLineEnding = Encoding.UTF8.GetBytes("\r\n");
+        private static byte[] _newLine = Encoding.UTF8.GetBytes(Environment.NewLine);
+
```
davidfowl: nit: remove extra space

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\ndata: ", "Hello, World")]
+        public void ParseSSEMessageSuccessCases(string encodedMessage, string expectedMessage)
+        {
+            var buffer = Encoding.UTF8.GetBytes(encodedMessage);
+            var readableBuffer = ReadableBuffer.Create(buffer);
+            var parser = new ServerSentEventsMessageParser();
+            var consumed = new ReadCursor();
+            var examined = new ReadCursor();
+
+            var parsePhase = parser.ParseMessage(readableBuffer, out consumed, out examined, out Message message);
```
davidfowl: parseResult

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        [InlineData("data: T\r\ndata: A\rB\r\n\r\n", "A\rB")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: E\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\ndata: ", "Hello, World")]
+        public void ParseSSEMessageSuccessCases(string encodedMessage, string expectedMessage)
+        {
+            var buffer = Encoding.UTF8.GetBytes(encodedMessage);
+            var readableBuffer = ReadableBuffer.Create(buffer);
+            var parser = new ServerSentEventsMessageParser();
+            var consumed = new ReadCursor();
```
davidfowl: Remove this and use `out var`

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,231 @@
...
+                    }
+                }
+            }
+            finally
+            {
+                _transportCts.Cancel();
+            }
+        }
+
+        public async Task SendMessages(Uri sendUrl, CancellationToken cancellationToken)
```
BrennanConroy: private

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageFormatter.cs
```diff
@@ -14,11 +14,11 @@ public static class ServerSentEventsMessageFormatter
         private static readonly byte[] DataPrefix = new byte[] { (byte)'d', (byte)'a', (byte)'t', (byte)'a', (byte)':', (byte)' ' };
         private static readonly byte[] Newline = new byte[] { (byte)'\r', (byte)'\n' };
 
-        private const byte LineFeed = (byte)'\n';
-        private const char TextTypeFlag = 'T';
-        private const char BinaryTypeFlag = 'B';
-        private const char CloseTypeFlag = 'C';
-        private const char ErrorTypeFlag = 'E';
+        const byte LineFeed = (byte)'\n';
```
BrennanConroy: Why?

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.using System;
```
BrennanConroy: Remove `using System;` from the header

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
+{
+    public class ServerSentEventsMessageParser
+    {
+        const byte ByteCR = (byte)'\r';
```
BrennanConroy: Add visibility qualifier, `public` `private` `internal`

moozzyk:
Rebase on dev and fix this test: 
https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Tests/DefaultTransportFactoryTests.cs#L71
After your changes the attribute should read: 
```
[InlineData(TransportType.WebSockets, typeof(ServerSentEventsTransport))]
```

moozzyk:
Is this PR replacing https://github.com/aspnet/SignalR/pull/358? Because the other is still active and has comments that have not be addressed (e.g. https://github.com/aspnet/SignalR/pull/358#discussion_r111247765) or responded to...

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        [InlineData("data: \r")]
+        [InlineData("data: T\r\nda")]
+        [InlineData("data: T\r\ndata:")]
+        [InlineData("data: T\r\ndata: Hello, World")]
+        [InlineData("data: T\r\ndata: Hello, World\r")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\\")]
+        [InlineData("data: T\r\ndata: Major\r\ndata:  Key\rndata:  Alert\r\n\r\\")]
+        [InlineData("data: T\r\ndata: Major\r\ndata:  Key\r\ndata:  Alert\r\n\r\\")]
```
moozzyk: Is `\\` this correct/intentional? `\n` is not two characters so if the intention was that the first message ended with `\` and the last message was just `n` that it is incorrect. 
This actually might be a bug in the parser if this passes - after `\r\n` the only two expected things are `\r\n`  (end of message) or `data: ` (more data). Since you already parsed `\r` after '\r\n` the only valid value is `\n` but in the last two tests you allow `\`

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
mikaelm12: This was by design from @anurse and @moozzyk . 
The message type is used [here ](https://github.com/aspnet/SignalR/pull/401/files/3165bf1713de5bd2b914a37db78a3e188e969daa#diff-18dc3a080dc48bb38ff05117f88d1495R214) in writeMessagesAsync.

moozzyk: It just says that the entire payload will be sent as binary.

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+    public class ServerSentEventsParserTests
+    {
+        [Theory]
+        [InlineData("data: T\r\n\r\n", "")]
+        [InlineData("data: T\r\ndata: \r\r\n\r\n", "\r")]
+        [InlineData("data: T\r\ndata: A\rB\r\n\r\n", "A\rB")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: E\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\ndata: ", "Hello, World")]
+        public void ParseSSEMessageSuccessCases(string encodedMessage, string expectedMessage)
```
moozzyk: we talked about adding some cases for standalone `\r` (i.e. `\r` not followed by `\n`)

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: I am still missing some general SSE tests - similar to what we have for the longpolling transport:
https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Sockets.Common.Tests.Internal.Formatters
+{
+    public class ServerSentEventsParserTests
+    {
+        [Theory]
+        [InlineData("data: T\r\n\r\n", "")]
```
moozzyk: I don't see any test cases for Binary messages

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,231 @@
...
+                throw new ArgumentNullException(nameof(_httpClient));
+            }
+
+            _httpClient = httpClient;
+            _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<ServerSentEventsTransport>();
+        }
+
+        public Task StartAsync(Uri url, IChannelConnection<SendMessage, Message> application)
+        {
+            _logger.LogInformation("Starting {0}", nameof(ServerSentEventsTransport));
```
anurse: Use `"Starting {transportName}"`. In Logging (and only in logging) you can use names instead of indexes. The order the tokens appear has to match the input array, and the names and values are stored in key-value pairs for loggers that support it.

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,231 @@
...
+
+            _application = application;
+            var sseUrl = Utils.AppendPath(url, "sse");
+            var sendUrl = Utils.AppendPath(url, "send");
+            var sendTask = SendMessages(sendUrl, _transportCts.Token);
+            var receiveTask = OpenConnection(_application, sseUrl, _transportCts.Token);
+
+            Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
+            {
+                _logger.LogDebug("Transport stopped. Exception: '{0}'", t.Exception?.InnerException);
```
anurse: Should this be an error? Also, use `if(t.Exception != null) { LogError(t.Exception, "Transport stopped") }`. There is an explicit override for logging messages containing exceptions.

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
anurse: /cc @BrennanConroy to confirm I'm right here :).

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
anurse: There's little reason for the C# client to use the text format, since it's only really for compatibility with Browsers. We can make it configurable, but file a bug if you want that and let's do it separately.

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,231 @@
...
+            _logger.LogInformation("Send loop stopped");
+        }
+
+        private async Task WriteMessagesAsync(List<SendMessage> messages, PipelineTextOutput output, MessageFormat format)
+        {
+            output.Append(MessageFormatter.GetFormatIndicator(format), TextEncoder.Utf8);
+
+            foreach (var message in messages)
+            {
+                _logger.LogDebug("Writing '{0}' message to the server", message.Type);
```
anurse: Again, use a named token: `"Writing '{messageType}' message to server"`

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -0,0 +1,231 @@
...
+                    // We didn't get any more memory!
+                    throw new InvalidOperationException("Unable to write message to pipeline");
+                }
+                await output.FlushAsync();
+            }
+        }
+
+        public async Task StopAsync()
+        {
+            _logger.LogInformation("Transport {0} is stopping", nameof(ServerSentEventsTransport));
```
anurse: Named token: `"Transport {transportName} is stopping"`

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
...
+
+                if (line.Length <= 1)
+                {
+                    throw new FormatException("There was an error in the frame format");
+                }
+
+                if (IsMessageEnd(line))
+                {
+                    _internalParserState = InternalParseState.ReadEndOfMessage;
+                }
```
anurse: newline after this

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
...
+            examined = buffer.End;
+            message = new Message();
+            var reader = new ReadableBufferReader(buffer);
+
+            var start = consumed;
+            var end = examined;
+
+            while (!reader.End)
+            {
+                var messageType = MessageType.Text;
```
anurse: You need to store this in a field, because if we see `data: T\r\n` in one buffer and `data: Foo\r\n` in the next buffer, we need to store the message type between invocations of the Parse method.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
...
+                    _internalParserState = InternalParseState.ReadEndOfMessage;
+                }
+                // To ensure that the \n was preceded by a \r
+                // since messages can't contain \n.
+                // data: foo\n\bar should be encoded as
+                // data: foo\r\n
+                // data: bar\r\n
+                else if (line[line.Length - _sseLineEnding.Length] != ByteCR)
+                {
+                    throw new FormatException("A '\\n' character can only be used as a line ending");
```
anurse: `"Unexpected '\n' in message. A '\n' character can only be used as part of the newline sequence '\r\n'"` might be a little clearer.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
...
+                        {
+                            // Find the final size of the payload
+                            var payloadSize = 0;
+                            foreach (var dataLine in _data)
+                            {
+                                payloadSize += dataLine.Length + _newLine.Length;
+                            }
+
+                            // Allocate space in the paylod buffer for the data and the new lines 
+                            var payload = new byte[payloadSize - _newLine.Length];
```
anurse: Add a comment as to why you're subtracting `_newLine.Length` again (i.e. because there isn't a final newline)

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -0,0 +1,197 @@
...
+
+                            var offset = 0;
+                            foreach (var dataLine in _data)
+                            {
+                                dataLine.CopyTo(payload, offset);
+                                offset += dataLine.Length;
+                                if (_data.Count > 1 && offset < payload.Length)
```
anurse: I don't think you need `_data.Count > 1` here. If there is only one item, `offset` should equal `payload.Length` now.

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        [InlineData("d", "ata: T\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T", "\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r", "\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\n", "data: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\nd", "ata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\ndata: ", "Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\ndata: Hello, World", "\r\n\r\n", "Hello, World")]
+        [InlineData("data: T\r\ndata: Hello, World\r\n", "\r\n", "Hello, World")]
+        [InlineData("data: T", "\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
+        [InlineData("data: ", "T\r\ndata: Hello, World\r\n\r\n", "Hello, World")]
```
anurse: As mentioned above, you'll encounter an issue with something like `"data:T\r\n", "data: Foo\r\n\r\n"` because you don't store the message type outside the parse function. It's masked right now because you always set `messageType` back to `Text`, so a test like `"data: B\r\n", "data: [base64]\r\n\r\n"` will show the issue right away because you'll get a successful parse but it will be a Text frame instead of a Binary frame.

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+                pipe.Reader.Advance(consumed, examined);
+
+                // Send the rest of the data and parse the complete message
+                await pipe.Writer.WriteAsync(Encoding.UTF8.GetBytes(encodedMessagePart2));
+                result = await pipe.Reader.ReadAsync();
+
+                parseResult = parser.ParseMessage(result.Buffer, out consumed, out examined, out message);
+                Assert.Equal(ServerSentEventsMessageParser.ParseResult.Completed, parseResult);
+
+                var resultMessage = Encoding.UTF8.GetString(message.Payload);
```
anurse: Verify the message type in the test

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+            var buffer = Encoding.UTF8.GetBytes(encodedMessage);
+            var readableBuffer = ReadableBuffer.Create(buffer);
+            var parser = new ServerSentEventsMessageParser();
+            var consumed = new ReadCursor();
+            var examined = new ReadCursor();
+
+            var parsePhase = parser.ParseMessage(readableBuffer, out consumed, out examined, out Message message);
+            Assert.Equal(ServerSentEventsMessageParser.ParseResult.Completed, parsePhase);
+
+            var result = Encoding.UTF8.GetString(message.Payload);
```
anurse: Verify message type.

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
```
anurse: ABVMT - Always Be Verifying Message Type ;)

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -0,0 +1,234 @@
...
+                result = await pipe.Reader.ReadAsync();
+                parseResult = parser.ParseMessage(result.Buffer, out consumed, out examined, out message);
+                Assert.Equal(ServerSentEventsMessageParser.ParseResult.Completed, parseResult);
+                Assert.Equal(MessageType.Text, message.Type);
+                Assert.Equal("bar", Encoding.UTF8.GetString(message.Payload));
+                pipe.Reader.Advance(consumed, examined);
+            }
+        }
+
+        public static IEnumerable<object[]> MessageSizesData
```
mikaelm12: Copy pasta 😄

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
mikaelm12: This is a bug. I've addressed it  in new changes that I'm going to push soon.
I don't think we would have actually run into an issue here because the transport loop would throw because the message is incomplete. Regardless, it shouldn't get outside of the parsing step.

anurse:
I just realized you'll also need to Base64-decode binary frames. When a frame is marked as Binary, the spec says the content is Base64-encoded data. In that case:

* You don't add newlines back when concatenating the payload buffers
* After building the final payload buffer, you need to run it through `Convert.FromBase64String(Encoding.UTF8.GetString(payload))` to get the _actual_ payload.

mikaelm12:
As per our conversation binary messages will be handled in a subsequent PR. Issues being tracked here https://github.com/aspnet/SignalR/issues/403

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
mikaelm12: Good catch 👍

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
mikaelm12: Not supported for now. That will be the next PR.

mikaelm12:
⬆️ 📅

Issue #352, opened 2017-03-29
ServerSentEvents in TypeScript client fails to connect on FireFox


BrennanConroy:
Firefox version 51.0.1

Running ChatSample
```
Failed to start the connection. undefined  signalr-client.js:51:21
Firefox can’t establish a connection to the server at http://localhost:5000/chat/sse?id=c85bda38-7744-471b-918d-0bc92a3ffc38.  signalr-client.js:472:34
uncaught exception: undefined
```

If you connect with a different browser like Chrome and send a message when the Firefox connection "failed" then you can start sending messages with the Firefox connection successfully.

PR #407, opened 2017-04-20
Work around for a Firefox bug

moozzyk:
Firefox won't fire EventSource open event until it receives some data. The workaround is to send an empty comment when starting ServerSentEvent transport.

Fixes: #352

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
@@ -37,6 +37,9 @@ namespace Microsoft.AspNetCore.Sockets.Transports
 
             context.Response.Headers["Content-Encoding"] = "identity";
 
+            // Workaround for a Firefox bug where EventSource won't fire the open event
+            // until it receives some data
+            await context.Response.WriteAsync(":\r\n");
             await context.Response.Body.FlushAsync();
 
             var pipe = context.Response.Body.AsPipelineWriter();
--- a/samples/ChatSample/Views/Home/Index.cshtml
+++ b/samples/ChatSample/Views/Home/Index.cshtml
@@ -14,26 +14,45 @@
 </div>
 <script src="lib/signalr-client/signalr-client.js"></script>
 <script>
+let transportType = signalR.TransportType[getParameterByName('transport')] || signalR.TransportType.WebSockets;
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Sockets/Transports/ServerSentEventsTransport.cs
```diff
@@ -37,6 +37,9 @@ public ServerSentEventsTransport(ReadableChannel<Message> application, ILoggerFa
 
             context.Response.Headers["Content-Encoding"] = "identity";
 
+            // Workaround for a Firefox bug where EventSource won't fire the open event
+            // until it receives some data
+            await context.Response.WriteAsync(":\r\n");
```
davidfowl: Instead of this, just send what old signalr sent. `data: init\r\n`

moozzyk: No. Parsers will try to parse it and will barf so we will have to update them everywhere. We could send data:T\r\ndata:\r\n\r\n but it is more data and I think it will be reported as a new message so we will have to handle this in each client. What I am sending now is a valid event-source comment as per spec so it is not reported in browsers and I updated our parser in C# to ignore comments in general.

mikaelm12: We could add handling for an init data message. We would just add some additional logic in the GetMessageType logic [here](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs#L180) to check for the init message

moozzyk: Why would we add custom magic that fits neither event source protocol nor ours if we can use a built-in feature of a protocol we are using. Note that with custom init TS client parsers would have to be updated as well. Also, if you want to do this correctly you would have to keep the state to detect cases where you get multiple init messages. Oh, in that case should we also send init for longpolling and websockets? But if we do this for webSockets at the Sockets level the IoT scenario - being able to connect to the server with bare webSocket is broken because the user need to react to (ignore) the init message
We just added a new future across the stack that no one (at the moment) needs or asked. 

If we want to have an init message that has a purpose other then working around a FF bug we need to design it.

Issue #408, opened 2017-04-20
SSE message parsing can accept invalid message types


mikaelm12:
When parsing the message type we expect to receive an SSE line of the form: `data: <MessageType>\r\n` where the message type is either T for text, B for binary, E for error or C for close. 
Currently the parser will accept for example an initial line `data: This is wrong\r\n` as an indicator of a text message. 
The issue is in the [GetMessageType method](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs#L180-L185)

PR #410, opened 2017-04-21
Fix SSE Message Type Parsing

mikaelm12:
Currently the parser will accept for  an initial message type line `data: This is not a message type\r\n ` as an indicator of a text message.
This fixes: https://github.com/aspnet/SignalR/issues/408

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
@@ -23,6 +23,8 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
         private static byte[] _sseLineEnding = Encoding.UTF8.GetBytes("\r\n");
         private static byte[] _newLine = Encoding.UTF8.GetBytes(Environment.NewLine);
 
+        private readonly static int _messageTypeLineLength = "data: X\r\n".Length;
+
         private InternalParseState _internalParserState = InternalParseState.ReadMessageType;
         private List<byte[]> _data = new List<byte[]>();
         private MessageType _messageType = MessageType.Text;
@@ -94,7 +96,10 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
                 switch (_internalParserState)
                 {
                     case InternalParseState.ReadMessageType:
-                        _messageType = GetMessageType(line);
+                        EnsureStartsWithDataPrefix(line);
+
+
+                        _messageType = ParseMessageType(line);
 
                         _internalParserState = InternalParseState.ReadMessagePayload;
 
@@ -102,6 +107,8 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
                         consumed = lineEnd;
                         break;
                     case InternalParseState.ReadMessagePayload:
+                        EnsureStartsWithDataPrefix(line);
+
                         // Slice away the 'data: '
                         var payloadLength = line.Length - (_dataPrefix.Length + _sseLineEnding.Length);
                         var newData = line.Slice(_dataPrefix.Length, payloadLength).ToArray();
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -181,6 +183,11 @@ private MessageType GetMessageType(ReadOnlySpan<byte> line)
         {
             EnsureStartsWithDataPrefix(line);
 
+            if (line.Length != _messagTypeLineLength)
```
BrennanConroy: Move this before `EnsureStartsWithDataPrefix`

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -129,17 +130,19 @@ public void ParseSSEMessageIncompleteParseResult(string encodedMessage)
...
         [InlineData("data: T\r\nf", "oo: Hello, World\r\n\r\n", "Expected the message prefix 'data: '")]
         [InlineData("foo", ": T\r\ndata: Hello, World\r\n\r\n", "Expected the message prefix 'data: '")]
         [InlineData("food:", " T\r\ndata: Hello, World\r\n\r\n", "Expected the message prefix 'data: '")]
         [InlineData("data: T\r\ndata: Hello, W", "orld\r\n\n", "There was an error in the frame format")]
         [InlineData("data: T\r\nda", "ta: Hello\n, World\r\n\r\n", "Unexpected '\n' in message. A '\n' character can only be used as part of the newline sequence '\r\n'")]
-        [InlineData("data:", " data: \r\n", "Unknown message type: 'd'")]
+        [InlineData("data:", " data: \r\n", "Expected a data format message of the form 'data: <MesssageType>'")]
         [InlineData("data: ", "T\r\ndata: Major\r\ndata:  Key\r\ndata:  Alert\r\n\r\\", "Expected a \\r\\n frame ending")]
+        [InlineData("data: ", "This is not a message type\r\n", "Expected a data format message of the form 'data: <MesssageType>'")]
```
moozzyk: I am missing a case like this "data: T", "ralala"

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
moozzyk: This is misplaced. GetMessageType is supposed to get the message type and not check the contents that follow the message type. I believe the check should happen here: https://github.com/aspnet/SignalR/pull/410/files#diff-b530db168b0693614299473e8c15d1d7R92 and it should just make sure that the the next two characters are \r\n. It would also remove the cryptic _messagTypeLineLength with the magic `X`.

mikaelm12: I've moved these checks out of this method but I think that the length check should stay. 
Because of the check here https://github.com/aspnet/SignalR/blob/c0af8a3a258a5b0d5d338f7580bf842ad91ed1e1/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs#L79 and here https://github.com/aspnet/SignalR/blob/c0af8a3a258a5b0d5d338f7580bf842ad91ed1e1/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs#L45 we aready know that we end with a \r\n. 

Also to check the next character we would still need a reference to the length of the message type line. So we would still need something resemebling the _messageTypeLineLength.

moozzyk: Maybe just rename this function to `ParseMessageType` then? We are already ensuring that the line starts with "data: " so it is more just a simple conversion function. I am not a big fan of `"data: X\r\n"` but it's probably fine. `_messagTypeLineLength` has a typo.

mikaelm12:
⬆️ 📅

mikaelm12:
Rebased on dev

PR #418, opened 2017-04-25
Support binary messages in SSE parser

mikaelm12:
This add functionality for properly handling binary messages over SSE. 
It introduces a new `MessageFormatUtils` class that handles the binary payload.
It also fixes a bug where the message type was being reset on calls to ParseMessage

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
@@ -128,6 +128,14 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
                                     offset += _newLine.Length;
                                 }
                             }
+
+                            switch (_messageType)
+                            {
+                                case MessageType.Binary:
+                                    payload = MessageFormatUtils.DecodePayload(payload);
+                                    break;
+                            }
+
                             message = new Message(payload, _messageType);
                         }
                         else
@@ -188,7 +196,7 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
                 case ByteT:
                     return MessageType.Text;
                 case ByteB:
-                    throw new NotSupportedException("Support for binary messages has not been implemented yet");
+                    return MessageType.Binary;
                 case ByteC:
                     return MessageType.Close;
                 case ByteE:
--- a/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
@@ -107,26 +107,34 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
                         {
                             payload = _data[0];
                         }
-
                         else if (_data.Count > 1)
                         {
                             // Find the final size of the payload
                             var payloadSize = 0;
                             foreach (var dataLine in _data)
                             {
-                                payloadSize += dataLine.Length + _newLine.Length;
+                                payloadSize += dataLine.Length;
                             }
 
-                            // Allocate space in the paylod buffer for the data and the new lines. 
-                            // Subtract newLine length because we don't want a trailing newline. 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -128,14 +139,16 @@ public ParseResult ParseMessage(ReadableBuffer buffer, out ReadCursor consumed,
                                     offset += _newLine.Length;
                                 }
                             }
+
+                            if( _messageType == MessageType.Binary)
```
anurse: spacing

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -104,7 +104,18 @@ public ParseResult ParseMessage(ReadableBuffer buffer, out ReadCursor consumed,
                         consumed = lineEnd;
                         break;
                     case InternalParseState.ReadEndOfMessage:
-                        if (_data.Count > 0)
+                        if (_data.Count == 1)
+                        {
+                            if (_messageType == MessageType.Binary)
+                            {
+                                payload = MessageFormatUtils.DecodePayload(_data[0]);
```
anurse: Don't do the decoding in the if branches. In the `if` branches, just assign `payload` to the single buffer that represents all the data we got from the client. Then **outside** this `if`, do another `if` on message type and decode in there. That way you only have to call `DecodePayload` once and there's a simple flow of operations:

1. Get the payload into a single buffer (via copying if necessary)
2. Decode the buffer **if necessary**
3. Put the buffer into a `Message` struct and return it.

src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/ServerSentEventsMessageParser.cs
```diff
@@ -104,7 +103,12 @@ public ParseResult ParseMessage(ReadableBuffer buffer, out ReadCursor consumed,
                         consumed = lineEnd;
                         break;
                     case InternalParseState.ReadEndOfMessage:
-                        if (_data.Count > 0)
+                        if (_data.Count == 1)
+                        {
+                            payload = _data[0];
+                        }
+
```
anurse: Remove this line

test/Microsoft.AspNetCore.Sockets.Common.Tests/Internal/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -14,21 +14,23 @@ namespace Microsoft.AspNetCore.Sockets.Common.Tests.Internal.Formatters
...
-        [InlineData("data: T\r\ndata: Hello, World\r\n\r\ndata: ", "Hello, World")]
-        public void ParseSSEMessageSuccessCases(string encodedMessage, string expectedMessage)
+        [InlineData("data: T\r\n\r\n", "", MessageType.Text)]
+        [InlineData("data: B\r\n\r\n", "", MessageType.Binary)]
+        [InlineData("data: T\r\ndata: \r\r\n\r\n", "\r", MessageType.Text)]
+        [InlineData("data: T\r\ndata: A\rB\r\n\r\n", "A\rB", MessageType.Text)]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", "Hello, World", MessageType.Text)]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", "Hello, World", MessageType.Text)]
+        [InlineData("data: T\r\ndata: Hello, World\r\n\r\ndata: ", "Hello, World", MessageType.Text)]
+        [InlineData("data: B\r\ndata: SGVsbG8sIFdvcmxk\r\n\r\n", "Hello, World", MessageType.Binary)]
```
anurse: Can you toss in a test which has base-64 spread across two lines?

```
data: B\r\n
data: SGVsbG8\r\n
data: sIFdvcmxk\r\n
\r\n
```

mikaelm12:
⬆️  📅 
Had to add some hacky code to properly handle not adding new lines to binary messages split across multiple data lines.

Issue #377, opened 2017-04-07
Move the LongPolling and SeverSentEvents SendMessage logic to a common place.


mikaelm12:
The SendMessage logic in LongPolling and ServerSentEvents is the same. After SSE gets merged this logic should be shared.

mikaelm12:
Addressed in https://github.com/aspnet/SignalR/pull/424

PR #424, opened 2017-04-27
Move Long Polling and SSE send logic

mikaelm12:
The code for sending messages in long polling and server sent events was the same. I moved that logic to a static helper.
Tracked here: https://github.com/aspnet/SignalR/issues/377

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/SendUtils.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/SendUtils.cs
@@ -1,0 +1,127 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Pipelines;
+using System.IO.Pipelines.Text.Primitives;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Text.Formatting;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public static class SendUtils
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Sockets.Client/SendUtils.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.Text;
+using System.Text.Formatting;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Sockets.Internal.Formatters;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Sockets.Client
+{
+    public static class SendUtils
```
moozzyk: This class should be internal

src/Microsoft.AspNetCore.Sockets.Client/SendUtils.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
+        private static readonly ProductInfoHeaderValue DefaultUserAgentHeader = ProductInfoHeaderValue.Parse(DefaultUserAgent);
+        public static async Task SendMessages(Uri sendUrl, IChannelConnection<SendMessage, Message> application, HttpClient httpClient, CancellationTokenSource transportCts, ILogger logger)
```
moozzyk: nit: separate variables and function with an empty new line

src/Microsoft.AspNetCore.Sockets.Client/SendUtils.cs
```diff
@@ -17,10 +17,12 @@
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
-    public static class SendUtils
+    internal static class SendUtils
     {
         private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
+
```
moozzyk: nit: these can be together

PR #433, opened 2017-05-02
React to Logging API changes

BrennanConroy:


BrennanConroy:
The filters aren't all equivalent, I mainly focused on the console ones in tests to be similar. I probably should just change all of them.

Eilon:
Ok... I don't personally care what the filter levels are, but if there's some cleanup to do here, might as well do that 😄

PR #437, opened 2017-05-04
Change targets

mikaelm12:
@Eilon @muratg @mikeharder @JunTaoLuo

PR #438, opened 2017-05-04
netcoreapp2.0

anurse:


moozzyk:
Fight https://github.com/aspnet/SignalR/pull/437 vs https://github.com/aspnet/SignalR/pull/438

(@JunTaoLuo comments apply to both)

anurse:
I could not find that PR when I was looking... Weird, but my bad :)

PR #439, opened 2017-05-05
Fixing SocialWeather sample

moozzyk:


samples/SocialWeather/wwwroot/index.html
```diff
@@ -67,7 +67,7 @@
                 span.style.color = color;
             }
 
-            let connectUrl = `ws://${document.location.host}/weather/ws?formatType=json`;
```
moozzyk: Technically not needed since we use Text by default

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -231,17 +231,14 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
 
         private ConnectionState CreateConnection(HttpContext context)
         {
-            var format =
```
davidfowl: Why change this?

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -231,17 +231,14 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
-                    ? MessageType.Binary
-                    : MessageType.Text;
-
             var state = _manager.CreateConnection();
             state.Connection.User = context.User;
 
             // TODO: this is wrong. + how does the user add their own metadata based on HttpContext
             var formatType = (string)context.Request.Query["formatType"];
             state.Connection.Metadata["formatType"] = string.IsNullOrEmpty(formatType) ? "json" : formatType;
+            state.Connection.Metadata["Context"] = context;
```
davidfowl: this should probably be typeof(HttpContext)

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
moozzyk: This variable is not being used at all.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
moozzyk: Sure. I just took this: https://github.com/aspnet/SignalR/pull/390/files#diff-366b10d924ffee156b8529bd1727f6e2R236

moozzyk:
:up::date: Also fixed a couple exceptions caused by not reading/writing from/to the channel correctly (thanks @anurse for helping me with that)

PR #440, opened 2017-05-05
Use common ObjectMethodExecutor

BrennanConroy:


davidfowl:
IT might be worth applying the optimizations here. https://github.com/aspnet/Mvc/pull/6171/files

BrennanConroy:
What optimization? We do the one with `Execute` and casting the result to `Task` to avoid an allocation.

davidfowl:
Ah I see. Merge it then 😄

Issue #428, opened 2017-04-29
Address package downgrade warnings


mikeharder:
Latest build of `UniverseCoherence (2.0.0-preview1)` contained package downgrade warnings in this repo:

http://aspnetci/viewLog.html?buildId=211993&buildTypeId=UniverseCoherence&tab=buildLog

```
Detected package downgrade: System.Net.NameResolution from 4.3.0 to 4.0.0  [C:\b\w\590aeb6bc9b87cb8\.r\SignalR\.build\KoreBuild.proj]
 SocketsSample (>= 1.0.0) -> Microsoft.AspNetCore.SignalR.Redis (>= 1.0.0-preview1-24766) -> StackExchange.Redis.StrongName (>= 1.1.605) -> System.Net.Sockets (>= 4.1.0) -> runtime.win.System.Net.Sockets (>= 4.3.0) -> System.Net.NameResolution (>= 4.3.0)  [C:\b\w\590aeb6bc9b87cb8\.r\SignalR\.build\KoreBuild.proj]
 SocketsSample (>= 1.0.0) -> Microsoft.AspNetCore.SignalR.Redis (>= 1.0.0-preview1-24766) -> StackExchange.Redis.StrongName (>= 1.1.605) -> System.Net.NameResolution (>= 4.0.0) [C:\b\w\590aeb6bc9b87cb8\.r\SignalR\.build\KoreBuild.proj]
Detected package downgrade: System.Net.Primitives from 4.3.0 to 4.0.11  [C:\b\w\590aeb6bc9b87cb8\.r\SignalR\.build\KoreBuild.proj]
 SocketsSample (>= 1.0.0) -> Microsoft.AspNetCore.SignalR.Redis (>= 1.0.0-preview1-24766) -> StackExchange.Redis.StrongName (>= 1.1.605) -> System.Net.Sockets (>= 4.1.0) -> runtime.win.System.Net.Sockets (>= 4.3.0) -> System.Net.Primitives (>= 4.3.0)  [C:\b\w\590aeb6bc9b87cb8\.r\SignalR\.build\KoreBuild.proj]
[...]

PR #442, opened 2017-05-08
Upgrade StackExchange.Redis.StrongName to version 1.2.3

mikeharder:
- Addresses package downgrade warnings (https://github.com/aspnet/SignalR/issues/428)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ branches:
     - release
     - dev
     - /^(.*\/)?ci-.*$/
+    - /^rel\/.*/
 before_install:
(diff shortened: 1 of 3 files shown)


moozzyk:
I think the package should be updated too because it currently brings [StrongName version](https://github.com/aspnet/SignalR/blob/ca9a44afc3bf88c83ff2fc51795a1de2875b5ad4/src/Microsoft.AspNetCore.SignalR.Redis/Microsoft.AspNetCore.SignalR.Redis.csproj#L17) (not that the previous version was correct...)

moozzyk:
Ignore my comment - I must have clicked a wrong *StrongName package where the latest was something like 1.0.1

Issue #271, opened 2017-03-08
Configurable WebSockets Transport Buffer Size


mikaelm12:
The WebSockets transport arbitrarily uses a 1k buffer for reads. We should figure out a better value for this.

muratg:
cc @moozzyk thoughts?

moozzyk:
https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Core/Owin/WebSockets/WebSocketHandler.cs#L20-L21 - possibly configurable but I am not feeling strongly about this.

muratg:
Set the buffer to 4KB.

grahamehorner:
+1 vote for a configurable buffer value with a default value 4k

davidfowl:
@mikaelm12 Add an option to https://github.com/aspnet/SignalR/blob/ccf6cd415e7e79896ba6b47045ecc5a5f1474460/src/Microsoft.AspNetCore.Sockets.Http/WebSocketOptions.cs. 

We may also want to consider a maximum message size https://github.com/SignalR/SignalR/blob/83726e069c2c19a32a2e57a990facad17328906b/src/Microsoft.AspNet.SignalR.Core/Configuration/IConfigurationManager.cs#L51

davidfowl:
Closing this since we're going to just use bounded pipes

SamBallantyne:
@davidfowl how did this end up? 

I'd like to use websocketd for a very low latency application where I am writing to stdout roughly 100 messages of about 100B every second. Websocketd seems to be buffering the input... Any thoughts?

SamBallantyne:
What are the bounds on the pipe? Is it possible to flush?

davidfowl:
@SamBallantyne what exactly are you trying to control? There are various buffer sizes some are configurable and some aren't configurable.

PR #448, opened 2017-05-11
Increase the size of the WebSockets buffer to 4K.

mikaelm12:
Addresses issue: https://github.com/aspnet/SignalR/issues/271

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client/WebSocketsTransport.cs
@@ -68,7 +68,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                const int bufferSize = 1024;
+                const int bufferSize = 4096;
                 var totalBytes = 0;
                 var incomingMessage = new List<ArraySegment<byte>>();
                 WebSocketReceiveResult receiveResult;
--- a/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
@@ -142,8 +142,8 @@ namespace Microsoft.AspNetCore.SignalR.Tests
         {
             get
             {
-                yield return new object[] { new string('A', 5 * 1024) };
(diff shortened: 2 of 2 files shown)


grahamehorner:
I'm wondering why hardcode the buffer size; should it not be configurable ? with a default value of 4k

moozzyk:
Do tests need to be updated? I remember we had tests whose messages sizes were specifically chosen to fit/not fit multiple buffers.

mikaelm12:
> I'm wondering why hardcode the buffer size; should it not be configurable ?

We discussed this, no one felt strongly about it and we decided unless there was a demonstrated interest configuring the buffer size we would just leave it at 4k. That being said we can still change it 😄

mikaelm12:
@moozzyk 

>  I remember we had tests whose messages sizes were specifically chosen to fit/not fit multiple buffers.

Good point. I'll update the EndToEnd tests

grahamehorner:
what would be the impact of a 4k buffer for a high volume service that is only using 256byte messages?
or a service that is using messages larger than 4k ? would GC be working overtime? allocating 4k buffers for a small messages?

moozzyk:
YAGNI. SignalR 2.x has been running with a 4K buffer for webSocket messages from the very beginning and no one has reported any problems with that. If this turns out to be an issue we can think of making it configurable but we should not do this now. Note that this is on the client and not on the server.

davidfowl:
> what would be the impact of a 4k buffer for a high volume service that is only using 256byte messages?
or a service that is using messages larger than 4k ? would GC be working overtime? allocating 4k buffers for a small messages?

Would need to be measured. Historically we don't worry about it as much on the client side but the downside of the entire API surface here is that we're pinning the 4K buffer during the length of the async read.

The other bigger issues with this code IMO are the fact that we're making a new buffer and a new list of buffers per read. On top of that we're not using pooled buffers. In short, this code needs to be  profiled and potentially overhauled in the long run. The initial goal was to get it up and running then we can optimize.

grahamehorner:
I'm happy to do some profiling to assist the team, I agree @davidfowl the allocation of memory read with no memory pooling could cause client side issues especially in this new world of multi-platform .NET where devices such as Raspberry PI has a smaller amount memory where memory fragmentation may be an issues, and the GC would need to run more frequently.

IMO some issues could be reduced if the buffer size was configurable; happy to test with some of the use cases we have on low memory devices.

PR #452, opened 2017-05-15
Change the size header to little endian

Drawaes:
While it has been traditional to use Big Endian byte orders in protocols, it really is only that a tradition. As almost all platforms (ARM can run in both in theory, but isn't normally configured for anything other than Little, and I think the XBOX) will run in little endian is there a good reason to burn ops/the hassle of bit flipping?

### DIFF
--- a/specs/TransportProtocols.md
+++ b/specs/TransportProtocols.md
@@ -174,7 +174,7 @@ The body is encoded as follows. The `Content-Type` response header is set to `ap
 B([Length][Type][Body])([Length][Type][Body])... continues until end of the response body ...
 ```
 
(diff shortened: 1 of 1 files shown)


anurse:
It's not by convention or tradition, it's by specification in [RFC 1700](https://tools.ietf.org/html/rfc1700). That spec defined that Internet protocols would use big-endian and I haven't seen data to show that diverging from this would give us a significant performance impact to justify it. Honestly, if we can get to the point where bit-swapping the length field is our performance bottleneck, I'll be very happy with our performance level (especially given that every protocol layer below us is already dealing with the impact of bit-swapping).

I'm closing this, as we don't expect to diverge from this specification unless there is significant data indicating a value in doing so.

PR #454, opened 2017-05-15
Fixing `dotnet ef database update`

moozzyk:
Currently it is impossible to create database for the chat sample - `dotnet ef` commands fail with: "No parameterless constructor was found on 'ApplicationDbContext'. Either add a parameterless constructor to 'Application
DbContext' or add an implementation of 'IDbContextFactory<ApplicationDbContext>' in the same assembly as 'ApplicationDbC
ontext'." - adding IDbContextFactory implementation to fix this

Also disabling Redis presence in favor of in-memory presence


moozzyk:
Note: Register (and possibly some other pages) are currently broken due to: https://github.com/aspnet/Mvc/issues/6272

ajaybhargavb:
https://github.com/aspnet/Mvc/issues/6272 is fixed

PR #461, opened 2017-05-18
Fix #324 by adding streaming support to the server

### LINKED ISSUE #324: Design and implement support for Hub methods returning Streams and IObservable<T>
We want to support Hub methods returning streams of data, like this:

```csharp
public IObservable<Thingy> Subscribe()
{
}

public Stream DownloadFile()
{
}
```

We need a design and implementation to support this.

anurse:
There are a few small WIP items:

* TypeScript Client support. I'm not totally happy with it. Contemplating just removing it from this PR, since it doesn't affect the server and the server has tests for it's behavior
* Going to change `StreamingAttribute` to target the method, not the return value. The return value may be the [technically correct](https://www.youtube.com/watch?v=hou0lU8WMgo) place to annotate this but the method is the intuitive place
* Haven't written the C# client support. Again, I'm not sure we should do that in this PR.

We only support `ReadableChannel<T>` and `IObservable<T>` right now. `Stream` is a little trickier so I haven't done it yet.


samples/SocketsSample/Hubs/Streaming.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Channels;
+using Microsoft.AspNetCore.SignalR;
+
+namespace SocketsSample.Hubs
+{
+    public class Streaming : Hub
+    {
+        [return: Streaming]
```
davidfowl: Remind me why we need the attribute `[Streaming]`?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -295,6 +307,68 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
             }
         }
 
+        private bool IsChannel(Type type, out Type payloadType)
+        {
+            var channelType = type.AllBaseTypes().FirstOrDefault(t => t.IsGenericType && t.GetGenericTypeDefinition() == typeof(ReadableChannel<>));
```
davidfowl: This doesn't look very efficient 😄

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -295,6 +307,68 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
+            }
+            else if (IsChannel(result.GetType(), out var payloadType))
+            {
+                enumerator = AsyncEnumeratorAdapters.FromChannel(result, payloadType);
+                return true;
+            }
+            else
+            {
+                // This type cannot be streamed
+                throw new NotSupportedException($"Cannot stream results of type: {result.GetType().FullName}");
```
davidfowl: Observable of 1 item?

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
```diff
@@ -0,0 +1,87 @@
...
+                // This will block the thread emitting the object if the channel is bounded and full
+                // I think this is OK, since we want to push the backpressure up. However, we may need
+                // to find a way to force the subscription off to a dedicated thread in order to
+                // ensure we don't block other tasks
+
+                // Right now however, we use unbounded channels, so all of the above is moot because TryWrite will always succeed
+                while (!_output.TryWrite(value))
+                {
+                    // Wait for a spot
+                    if (!_output.WaitToWriteAsync(_cancellationToken).Result)
```
davidfowl: Task queue would be better...

davidfowl:
> Haven't written the C# client support. Again, I'm not sure we should do that in this PR.

We need either the C# client or the typescript client to be part of this PR. Otherwise I'm not sure we have anything real.

davidfowl:
Running the ChannelCounter fails in the sample.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: It'll go away when we have `IAsyncEnumerable<T>`, because we can just cast and call `GetEnumerator`

samples/SocketsSample/Hubs/Streaming.cs
anurse: I originally had that thought but when we talked last we decided to use an attribute to make it clear. It's a little less clear around something like `IAsyncEnumerable<T>`. Do you want to load up all the data before returning a result or stream it out?

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
anurse: But we still have to block this thread to create backpressure. Otherwise the caller will just keep invoking OnNext and the tasks will queue up. That's why `IObservable` is less ideal. If we're OK overrunning the Task Queue, then we can just queue things up.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: If we keep the attribute, sure. But if we remove the attribute then single results just won't be streamed.

anurse:
> We need either the C# client or the typescript client to be part of this PR. Otherwise I'm not sure we have anything real.

We need to review the API before we can go very far there.

anurse:
> co-variance

**contra**-variance :trollface:

samples/SocketsSample/Hubs/Streaming.cs
anurse: But if you think it's clear that `IAsyncEnumerable<T>` should be streamed, we can get rid of the attribute.

anurse:
I'm going to drop the TS client and spike out a C# client API to unblock the PR. The TS client requires building a whole primitive for streaming, whereas the C# one does not.

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
davidfowl: Yes but the you can block after some threshold.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Yes let's nuke it

davidfowl:
Isn't "out" - co-variant and "in" contravarint?

anurse:
That's totally what I said. Check again.

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
anurse: Isn't that threshold just defined by the channel we create? Right now it's unbounded, but we can create a bounded channel. Also, how does the blocking work with the Task queue? Do we need to track how many outstanding tasks we have in the Adapter and then block when it hits a certain threshold?

Using the channel for this seems simpler.

samples/SocketsSample/wwwroot/index.html
```diff
@@ -18,5 +18,11 @@
         <li><a href="hubs.html?transport=ServerSentEvents">Server Sent Events</a></li>
         <li><a href="hubs.html?transport=WebSockets">Web Sockets</a></li>
     </ul>
+    <h1>ASP.NET SignalR (Streaming)</h1>
```
BrennanConroy: Could you add `Core` to all these

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
```diff
@@ -0,0 +1,87 @@
+using System;
```
BrennanConroy: copyright

anurse:
Small incomplete update to fix simple comments (Channel streaming co-variance, some JavaScript bugs).

I think I'm going to add automated tests for the TypeScript client and then look in to C#.

Still to come:

* [ ] Remove Streaming attribute
* [ ] Add TypeScript client tests
* [ ] Add C# client code
* [ ] Add C# client tests.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Observable.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export class Subject<T> implements Observable<T> {
+    observers: Observer<T>[];
+
+    constructor() {
+        this.observers = [];
+    }
+
+    public next(item: T): void {
+        for (let observer of this.observers) {
```
moozzyk: nit: these loops could be just forEach

samples/SocketsSample/wwwroot/streaming.html
```diff
@@ -0,0 +1,105 @@
...
+            click('channelButton', function () {
+                run('ChannelCounter')
+            });
+
+            function run(method) {
+                let id = invocationCounter;
+                invocationCounter += 1;
+                addLine('resultsList', `running ${method}(${id}) ...`);
+                connection.stream(method, 10, (Math.random() * 5) * 200).subscribe({
+                    closed: false,
```
moozzyk: Needed at all? or should be just null because we don't maintain it.

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
```diff
@@ -0,0 +1,125 @@
...
+                .Invoke(null, new[] { observable });
+        }
+
+        public static IAsyncEnumerator<object> FromObservable<T>(IObservable<T> observable)
+        {
+            // TODO: Allow bounding and optimizations?
+            var channel = Channel.CreateUnbounded<object>();
+            var cancellationTokenSource = new CancellationTokenSource();
+
+            var subscription = observable.Subscribe(new ChannelObserver<T>(channel.Out, cancellationTokenSource.Token));
```
moozzyk: Why not just `CancellationToken.None` instead of `cts.Token`?

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
davidfowl: I see, that makes sense.

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
anurse: Because at one point I planned to cancel it, but it turns out there's nowhere to do so. Will fix :)

samples/SocketsSample/wwwroot/streaming.html
anurse: It's part of the interface, but I believe it's optional.

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
```diff
@@ -0,0 +1,125 @@
...
+                // ensure we don't block other tasks
+
+                // Right now however, we use unbounded channels, so all of the above is moot because TryWrite will always succeed
+                while (!_output.TryWrite(value))
+                {
+                    // Wait for a spot
+                    if (!_output.WaitToWriteAsync(_cancellationToken).Result)
+                    {
+                        // Channel was closed.
+                        throw new InvalidOperationException("Output channel was closed");
```
davidfowl: Won't this crash the process?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -518,6 +600,63 @@ private IServiceProvider CreateServiceProvider(Action<ServiceCollection> addServ
             return services.BuildServiceProvider();
         }
 
+        public class StreamingHub : TestHub
+        {
+            [return: Streaming]
```
davidfowl: Remove

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
davidfowl: You actually do need to make cancellation work here. I was just thinking about this a bit. The token here needs to represent the connection lifetime. If you're streaming and the connection dies, we need to notify something. There are some challenges though:
- If the particular stream was created as part of that invocation, then completing the stream makes sense if the connection goes away.
- If the stream is shared, you can't assume that ending a single connection should kill the stream. 

This means that we need to expose a cancellation token on the connection so that the hub can choose what to do when a connection goes away. We can also think about ways to make things a bit more automatic (something declarative).

src/Microsoft.AspNetCore.SignalR/Internal/AsyncEnumeratorAdapters.cs
anurse: It will crash the thing that drives the Observer, which depends on the underlying implementation of IObservable. We could suppress the exception and just drop items I guess.

anurse:
:up: :date: and rebase. I got the C# client support added and tests in place for it and the TS client.

build/dependencies.props
```diff
@@ -12,6 +12,7 @@
     <RedisVersion>1.2.3</RedisVersion>
     <TestSdkVersion>15.3.0-*</TestSdkVersion>
     <XunitVersion>2.3.0-beta2-*</XunitVersion>
+    <RxVersion>3.1.1</RxVersion>
```
davidfowl: Do we need permission to use this? 

/cc @Eilon

src/Microsoft.AspNetCore.SignalR.Client/InvocationRequest.cs
```diff
@@ -0,0 +1,113 @@
...
+        private readonly CancellationTokenRegistration _cancellationTokenRegistration;
+        private readonly bool _streaming;
+        private readonly ILogger _logger;
+        private readonly InvocationSubject _subject;
+
+        public Type ResultType { get; }
+        public CancellationToken CancellationToken { get; }
+        public string InvocationId { get; }
+
+        public Task<object> Task => _completionSource?.Task;
```
davidfowl: A null Task is pretty crappy. Make it return a completed task if there's nothing.

src/Microsoft.AspNetCore.SignalR.Client/InvocationRequest.cs
```diff
@@ -0,0 +1,113 @@
...
+        public IObservable<object> Observable => _subject;
+
+        public InvocationRequest(CancellationToken cancellationToken, Type resultType, string invocationId, ILoggerFactory loggerFactory, bool streaming)
+        {
+            _logger = loggerFactory.CreateLogger<InvocationRequest>();
+            _cancellationTokenRegistration = cancellationToken.Register(() => _completionSource.TrySetCanceled());
```
davidfowl: Remove allocation.

src/Microsoft.AspNetCore.SignalR.Client/InvocationRequest.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+            else
+            {
+                _completionSource.TrySetResult(result);
+            }
+        }
+
+        public void StreamItem(object item)
+        {
+            if(_streaming)
```
davidfowl: nit: space

src/Microsoft.AspNetCore.SignalR.Client/InvocationRequest.cs
```diff
@@ -0,0 +1,113 @@
...
+            if (_streaming)
+            {
+                _subject.TryOnError(new OperationCanceledException("Connection terminated"));
+            }
+            else
+            {
+                _completionSource.TrySetCanceled();
+            }
+
+            _cancellationTokenRegistration.Dispose();
```
davidfowl: We need to make sure things like this don't dead lock (because dispose waits on cancel callbacks to finish).

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: When we get C# 8? 😄 We'll need to make this better in the mean time, file an issue.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -295,6 +307,74 @@ public class HubEndPoint<THub, TClient> : EndPoint, IInvocationBinder where THub
...
+            else
+            {
+                payloadType = channelType.GetGenericArguments()[0];
+                return true;
+            }
+        }
+
+        private async Task StreamResultsAsync(string invocationId, Connection connection, IHubProtocol protocol, IAsyncEnumerator<object> enumerator)
+        {
+            // TODO: Cancellation?
```
davidfowl: File bug.

src/Microsoft.AspNetCore.SignalR/Internal/TypeBaseEnumerationExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    public static class TypeBaseEnumerationExtensions
+    {
+        public static IEnumerable<Type> AllBaseTypes(this Type type)
+        {
+            var current = type;
+            while(current != null)
```
davidfowl: nit: space

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -2,12 +2,16 @@
...
 
 using System;
+using System.Reactive.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.AspNetCore.SignalR.Tests.Common;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Xunit;
+using System.Reactive;
```
davidfowl: nit: usings are in the wrong spot.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -39,6 +43,28 @@ public class HubConnectionProtocolTests
...
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(new JsonSerializer()), new LoggerFactory());
+            try
+            {
+                await hubConnection.StartAsync();
+
+                var observable = hubConnection.Stream<object>("Foo");
+
+                var invokeMessage = await connection.ReadSentTextMessageAsync().OrTimeout();
+
+                Assert.Equal("{\"invocationId\":\"1\",\"type\":1,\"target\":\"Foo\",\"arguments\":[]}", invokeMessage);
```
davidfowl: Why are we comparing text?

davidfowl:
@anurse conflicts bro, my bad 😄

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -236,6 +254,20 @@ private void DispatchInvocation(InvocationMessage invocation, CancellationToken
             handler.Handler(invocation.Arguments);
         }
 
+        private void DispatchInvocationStreamItem(StreamItemMessage streamItem, InvocationRequest irq)
+        {
+            _logger.LogTrace("Received StreamItem for Invocation #{invocationId}", streamItem.InvocationId);
+
+            if (irq.CancellationToken.IsCancellationRequested)
+            {
+                _logger.LogTrace("Cancelling dispatch of StreamItem message for Invocation {invocationId}. The invocation was cancelled.", irq.InvocationId);
```
BrennanConroy: I'd prefer the American spelling of canceled and canceling

src/Microsoft.AspNetCore.SignalR.Client/InvocationSubject.cs
```diff
@@ -0,0 +1,116 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    internal class InvocationSubject : IObservable<object>
+    {
+        private readonly object _lock = new object();
+        private IList<IObserver<object>> _observers = new List<IObserver<object>>();
```
BrennanConroy: Internal, make this concrete `List`

src/Microsoft.AspNetCore.SignalR.Client/InvocationSubject.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+                else if (_error != null)
+                {
+                    observer.OnError(_error);
+                }
+                else
+                {
+                    _observers.Add(observer);
+                }
+                return new Subscription(this, observer);
```
BrennanConroy: Should probably return an empty disposable or something in the cases where `observer` isn't added to the list

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
davidfowl: 🇺🇸

build/dependencies.props
anurse: Shoot, I meant to wait to push this until I got confirmation from @Eilon, but I've already started the process with him.

src/Microsoft.AspNetCore.SignalR.Client/InvocationRequest.cs
anurse: The CTS is set to run continuations asynchronously, and firing that CTS is the only cancel callback.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
anurse: Weird... that one was definitely brought in with Ctrl-. ... at least I'm pretty sure it was.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
anurse: Because this is testing the HubProtocol logic. The only reason we "send" objects is because I added some small JSON serializing to the test code. I don't really think we want to JSON deserialize and then build a comparer.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: Look neighbour. I'll spell words how I want to spell them damnit!

anurse:
Rebased and :up: :date:d

anurse:
Derp. Fixing build.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: https://github.com/aspnet/SignalR/issues/481

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -125,17 +127,42 @@ export class HubConnection {
...
+                    // TODO: Log a warning if there's a payload?
+                    subject.complete();
+                }
+            }
+            else {
+                subject.next(<T>(<ResultMessage>invocationEvent).item);
+            }
+        });
+
+        //TODO: separate conversion to enable different data formats
```
davidfowl: Did you file a bug for this?

anurse: It's not a new comment actually: https://github.com/aspnet/SignalR/blob/dev/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts#L156

@moozzyk was an issue filed for this at the time it was added?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
anurse: To be fair, the type doesn't have to wait for C# 8, and wouldn't require a new netstandard version.

build/dependencies.props
anurse: Got approval from @Eilon

anurse:
Rebased. Going to squash and merge shortly, so last call for comments.

client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
```diff
@@ -2,85 +2,120 @@ const TESTHUBENDPOINT_URL = `http://${document.location.host}/testhub`;
...
+                        done();
+                    });
+            });
 
-            let callbackPromise = new Promise((resolve, reject) => {
-                client.on("Message", msg => {
-                    expect(msg).toBe(message);
-                    resolve();
-                });
+            it(`rethrows an exception from the server`, done => {
```
moozzyk: Consider adding the same test for streaming

src/Microsoft.AspNetCore.SignalR.Client/CastObservable.cs
```diff
@@ -0,0 +1,54 @@
...
+                _innerObserver.OnError(error);
+            }
+
+            public void OnNext(object value)
+            {
+                try
+                {
+                    _innerObserver.OnNext((TResult)value);
+                }
+                catch(Exception ex)
```
moozzyk: nit: space

src/Microsoft.AspNetCore.SignalR.Client/InvocationRequest.cs
```diff
@@ -0,0 +1,113 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    internal class InvocationRequest : IDisposable
```
moozzyk: Should we have two requests implementing the same interface? We fork the code basically in each function based on the _streaming flag...

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
moozzyk: There is no specific issue. We will hit this as soon as we want to add protobuf support. If someone tries adding support for a non-built-in format they will hit this too.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
```diff
@@ -2,150 +2,290 @@ import { IConnection } from "../Microsoft.AspNetCore.SignalR.Client.TS/IConnecti
...
 
-        try {
-            await invokePromise;
-            fail(); // exception is expected because the call has not completed
-        }
-        catch (e) {
-        }
-        done();
+            let hubConnection = new HubConnection(connection);
+            var invokePromise = hubConnection.invoke("testMethod", "foo", 42);
```
moozzyk: Either remove these parameters or rename "foo" to something else in tests where the expected result or error is also "foo".

src/Microsoft.AspNetCore.SignalR.Client/InvocationSubject.cs
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    internal class InvocationSubject : IObservable<object>
+    {
```
davidfowl: This looks like deadlock central. Triggering callbacks inside of locks smells like badness to me and will likely end being a large source of errors. Can we instead trigger them outside of the lock after we determine that they need to be called.

Also triggering callbacks inline on subscribe is basically the default tcs behavior which we try to change everywhere. Do we want to dispatch by default. Does user code run directly on these callbacks?

anurse:
> We need client tests that do bad things like block etc.

You want this PR in this week or you want these things fixed now? One or the other :)

anurse:
We can just remove the C# client support for now if you think that's cleaner.

src/Microsoft.AspNetCore.SignalR.Client/InvocationSubject.cs
anurse: Yep, we should capture the observers into a new local list and call them outside the lock

anurse:
Resolved pr feedback around locking, parameter names and InvocationRequest. I'll work on adding client tests to verify things next.

anurse:
There are other issues with the observable on the client. For example, if one of the observers blocks and a new item arrives, it could result in some observers getting the items out of order. In general, if we're going to use observable to represent streams on the client, we may need to either use Rx or reimplement complex portions of it.

Given that, I think we have a few options with this PR;

1. Change to use Channel to represent the results of `Stream`
2. Remove the C# streaming client for now and file an issue for adding it properly
3. Leave the C# streaming client as-is for now and file an issue for fixing it
4. Fix it in this PR (this is a lot of work)

src/Microsoft.AspNetCore.SignalR.Client/InvocationSubject.cs
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    return false;
+                }
+
+                localObservers = new List<IObserver<object>>(_observers);
+            }
+
+            foreach (var observer in localObservers)
+            {
+                observer.OnNext(value);
```
moozzyk: should we catch exceptions?

davidfowl:
I think I like changing the client representation to Channel for now (on both the server and client). I think we should commit what you have in this PR and keep iterating.

anurse:
> (on both the server and client)

The server representation is actually `IAsyncEnumerator<T>` currently, so it wouldn't need to change.

src/Microsoft.AspNetCore.SignalR.Client/InvocationSubject.cs
anurse: Yes, we should, but it won't matter now anyway :)

anurse:
Rebased and reworked the client to use Channel as the primitive for Streaming. Will merge when I have green builds (as per @davidfowl 's comment), and we can iterate on it moving forward.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -115,7 +116,7 @@ public void On(string methodName, Type[] parameterTypes, Action<object[]> handle
             _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
         }
 
-        public IObservable<object> Stream(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
+        public ReadableChannel<object> Stream(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
         {
             var irq = InvocationRequest.Stream(cancellationToken, returnType, GetNextId(), _loggerFactory, out var observable);
```
moozzyk: nit: rename observable to channel

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -203,7 +204,7 @@ private void OnDataReceived(byte[] data, MessageType messageType)
                         _logger.LogWarning("Dropped unsolicited Stream Item message for invocation '{invocationId}'", streamItem.InvocationId);
                         return;
                     }
-                    DispatchInvocationStreamItem(streamItem, irq);
+                    DispatchInvocationStreamItemAsync(streamItem, irq);
```
moozzyk: I think this introduces a race which can result in items not being reported in order. We might need a queue for this.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -252,17 +253,19 @@ private void DispatchInvocation(InvocationMessage invocation, CancellationToken
...
+        private async void DispatchInvocationStreamItemAsync(StreamItemMessage streamItem, InvocationRequest irq)
         {
             _logger.LogTrace("Received StreamItem for Invocation #{invocationId}", streamItem.InvocationId);
 
             if (irq.CancellationToken.IsCancellationRequested)
             {
                 _logger.LogTrace("Canceling dispatch of StreamItem message for Invocation {invocationId}. The invocation was cancelled.", irq.InvocationId);
             }
-            else
+            else if (!await irq.StreamItem(streamItem.Item))
```
moozzyk: Can this throw?

src/Microsoft.AspNetCore.SignalR.Client/InvocationRequest.cs
```diff
@@ -0,0 +1,153 @@
...
+                _completionSource.TrySetException(exception);
+            }
+
+            public override Task<bool> StreamItem(object item)
+            {
+                Logger.LogError("Invocation {invocationId} received stream item but was invoked as a non-streamed invocation.", InvocationId);
+                _completionSource.TrySetException(new InvalidOperationException("Streaming methods must be invoked using HubConnection.Stream"));
+
+                // We "delivered" the stream item successfully as far as the caller cares
+                return Task.FromResult(true);
```
davidfowl: This isn't cached

anurse: Boo ... https://source.dot.net/#System.Private.CoreLib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs,b003e738047b6fd3,references

anurse: Double Boo... https://source.dot.net/#System.Private.CoreLib/src/System/Threading/Tasks/future.cs,98

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: If we want to consider ordering of streamed items, that's a whole new design concept as the frames could get reordered in transit (so it doesn't even matter how we queue on the client). It's not necessarily required, as the user can stream items tagged with an order and do buffering.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
anurse: Only in the Streaming case, if someone else closes the write end of the channel (which only we control). I'll add some guard code though since this runs in a background thread.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: In what circumstances would frames be reordered in transit?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: Thanks - I was afraid of an unhandled exception on a background thread which would crash the process and which could not be caught/handled in any way.

davidfowl:
Weird MSBuild hang:

```
  Microsoft.AspNetCore.SignalR.Client.TS -> /home/travis/build/aspnet/SignalR/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/bin/Release/netstandard2.0/Microsoft.AspNetCore.SignalR.Client.TS.dll
MSBUILD : error MSB4166: Child node "32" exited prematurely. Shutting down. Diagnostic information may be found in files in the temporary files directory named MSBuild_*.failure.txt.
/home/travis/build/aspnet/SignalR/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj(27,5): warning MSB5021: "sh" and its child processes are being terminated in order to cancel the build.
/home/travis/build/aspnet/SignalR/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj(27,5): warning MSB4220: Waiting for the currently executing task "Exec" to cancel.
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/Roslyn/Microsoft.CSharp.Core.targets(71,5): warning MSB5021: "bash" and its child processes are being terminated in order to cancel the build. [/home/travis/build/aspnet/SignalR/samples/WebSocketSample/WebSocketSample.csproj]
  Microsoft.AspNetCore.Sockets.Common -> /home/travis/build/aspnet/SignalR/src/Microsoft.AspNetCore.Sockets.Common/bin/Release/netstandard1.3/Microsoft.AspNetCore.Sockets.Common.dll
  Microsoft.Extensions.WebSockets.Internal -> /home/travis/build/aspnet/SignalR/src/Microsoft.Extensions.WebSockets.Internal/bin/Release/netstandard2.0/Microsoft.Extensions.WebSockets.Internal.dll
  WebSocketSample -> /home/travis/build/aspnet/SignalR/samples/WebSocketSample/bin/Release/netcoreapp2.0/WebSocketSample.dll
AutobahnTests.cs(26,26): warning xUnit1004: Test methods should not be skipped [/home/travis/build/aspnet/SignalR/test/Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest/Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest.csproj]
  Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest -> /home/travis/build/aspnet/SignalR/test/Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest/bin/Release/netcoreapp2.0/Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest.dll
Build FAILED.
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/NuGet.targets(98,5): warning : Detected package downgrade: NETStandard.Library from 2.0.0-preview2-25316-01 to 1.6.1  [/home/travis/build/aspnet/SignalR/.build/KoreBuild.proj]
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/NuGet.targets(98,5): warning :  Microsoft.AspNetCore.SignalR.Client (>= 1.0.0-preview2-t004899032) -> Microsoft.AspNetCore.Sockets.Client (>= 1.0.0-preview2-t004899032) -> Microsoft.Extensions.Logging.Abstractions (>= 2.0.0-preview2-25326) -> NETStandard.Library (>= 2.0.0-preview2-25316-01)  [/home/travis/build/aspnet/SignalR/.build/KoreBuild.proj]
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/NuGet.targets(98,5): warning :  Microsoft.AspNetCore.SignalR.Client (>= 1.0.0-preview2-t004899032) -> NETStandard.Library (>= 1.6.1) [/home/travis/build/aspnet/SignalR/.build/KoreBuild.proj]
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/NuGet.targets(98,5): warning : Detected package downgrade: NETStandard.Library from 2.0.0-preview2-25316-01 to 1.6.1  [/home/travis/build/aspnet/SignalR/.build/KoreBuild.proj]
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/NuGet.targets(98,5): warning :  Microsoft.AspNetCore.Sockets.Client (>= 1.0.0-preview2-t004899032) -> Microsoft.Extensions.Logging.Abstractions (>= 2.0.0-preview2-25326) -> NETStandard.Library (>= 2.0.0-preview2-25316-01)  [/home/travis/build/aspnet/SignalR/.build/KoreBuild.proj]
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/NuGet.targets(98,5): warning :  Microsoft.AspNetCore.Sockets.Client (>= 1.0.0-preview2-t004899032) -> NETStandard.Library (>= 1.6.1) [/home/travis/build/aspnet/SignalR/.build/KoreBuild.proj]
/home/travis/build/aspnet/SignalR/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj(27,5): warning MSB5021: "sh" and its child processes are being terminated in order to cancel the build.
/home/travis/build/aspnet/SignalR/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj(27,5): warning MSB4220: Waiting for the currently executing task "Exec" to cancel.
/home/travis/.dotnet/sdk/2.0.0-preview2-006127/Roslyn/Microsoft.CSharp.Core.targets(71,5): warning MSB5021: "bash" and its child processes are being terminated in order to cancel the build. [/home/travis/build/aspnet/SignalR/samples/WebSocketSample/WebSocketSample.csproj]
AutobahnTests.cs(26,26): warning xUnit1004: Test methods should not be skipped [/home/travis/build/aspnet/SignalR/test/Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest/Microsoft.AspNetCore.WebSockets.Internal.ConformanceTest.csproj]
MSBUILD : error MSB4166: Child node "32" exited prematurely. Shutting down. Diagnostic information may be found in files in the temporary files directory named MSBuild_*.failure.txt.
    6 Warning(s)
    1 Error(s)
Time Elapsed 00:13:48.97
'dotnet msbuild @/home/travis/build/aspnet/SignalR/artifacts/msbuild/msbuild.rsp' failed with exit code 1
The command "./build.sh" exited with 1.
```

moozzyk:
I have seen this happen twice yesterday. I don't know how to get this file (or if it is even possible): `Diagnostic information may be found in files in the temporary files directory named MSBuild_*.failure.txt.`. I wonder if this: _This job ran on our container-based trusty beta_ can be the culprit.

anurse:
We've been running on the container-based infrastructure across the platform for a while now and problems haven't been that consistent. Plus, MSBuild should work in containers :).

moozzyk:
I agree in general. I was just wondering if because it's beta it still may have some issues where a process can be randomly killed which would affect the build.

PR #463, opened 2017-05-19
Make the http end points more resty

davidfowl:
# Resty end points

## Negotiation

```
OPTIONS /{path}
```

## Long polling

```
GET /{path}
```

## Server Sent Events

```
GET /{path}
Accept: text/event-stream
```

## HTTP Send

```
POST /{path}
```

## WebSockets

```
GET /{path}
Upgrade: websocket
Connection: Upgrade
```

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -22,7 +22,7 @@ export class WebSocketTransport implements ITransport {
     connect(url: string, queryString: string = ""): Promise<void> {
         return new Promise<void>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
-            let connectUrl = url + "/ws?" + queryString;
+            let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
 
             let webSocket = new WebSocket(connectUrl);
 
@@ -94,10 +94,10 @@ export class ServerSentEventsTransport implements ITransport {
 
         this.queryString = queryString;
         this.url = url;
-        let tmp = `${this.url}/sse?${this.queryString}`;
 
         return new Promise<void>((resolve, reject) => {
-            let eventSource = new EventSource(`${this.url}/sse?${this.queryString}`);
+            let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
+            let eventSource = new EventSource(connectUrl);
 
             try {
                 eventSource.onmessage = (e: MessageEvent) => {
@@ -139,7 +139,8 @@ export class ServerSentEventsTransport implements ITransport {
     }
 
     async send(data: any): Promise<void> {
-        return send(this.httpClient, `${this.url}/send?${this.queryString}`, data);
+        let sendUrl = this.url + (this.queryString == "" ? "" : "?" + this.queryString);
+        return send(this.httpClient, sendUrl, data);
     }
 
     stop(): void {
@@ -168,7 +169,8 @@ export class LongPollingTransport implements ITransport {
         this.url = url;
         this.queryString = queryString;
         this.shouldPoll = true;
-        this.poll(url + "/poll?" + this.queryString)
+        let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
+        this.poll(connectUrl);
         return Promise.resolve();
     }
 
@@ -231,7 +233,8 @@ export class LongPollingTransport implements ITransport {
     }
 
     async send(data: any): Promise<void> {
-        return send(this.httpClient, `${this.url}/send?${this.queryString}`, data);
+        let sendUrl = this.url + (this.queryString == "" ? "" : "?" + this.queryString);
+        return send(this.httpClient, sendUrl, data);
     }
 
     stop(): void {
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -22,7 +22,7 @@ export class WebSocketTransport implements ITransport {
     connect(url: string, queryString: string = ""): Promise<void> {
         return new Promise<void>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
-            let connectUrl = url + (queryString ? "" : "?" + queryString);
+            let connectUrl = url + (queryString ? "?" + queryString : "");
 
             let webSocket = new WebSocket(connectUrl);
 
@@ -95,7 +95,7 @@ export class ServerSentEventsTransport implements ITransport {
 
         this.queryString = queryString;
         this.url = url;
-        this.fullUrl = url + (queryString ? "" : "?" + queryString);
+        this.fullUrl = url + (queryString ? "?" + queryString : "");
 
         return new Promise<void>((resolve, reject) => {
             let eventSource = new EventSource(this.fullUrl);
@@ -170,7 +170,7 @@ export class LongPollingTransport implements ITransport {
         this.url = url;
         this.queryString = queryString;
         this.shouldPoll = true;
-        this.fullUrl = url + (queryString ? "" : "?" + queryString);
+        this.fullUrl = url + (queryString ? "?" + queryString : "");
         this.poll(this.fullUrl);
         return Promise.resolve();
     }
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -22,7 +22,7 @@ export class WebSocketTransport implements ITransport {
     connect(url: string, queryString: string = ""): Promise<void> {
         return new Promise<void>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
-            let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
+            let connectUrl = url + (queryString ? "" : "?" + queryString);
 
             let webSocket = new WebSocket(connectUrl);
 
@@ -81,6 +81,7 @@ export class ServerSentEventsTransport implements ITransport {
     private eventSource: EventSource;
     private url: string;
     private queryString: string;
+    private fullUrl: string;
     private httpClient: IHttpClient;
 
     constructor(httpClient: IHttpClient) {
@@ -94,10 +95,10 @@ export class ServerSentEventsTransport implements ITransport {
 
         this.queryString = queryString;
         this.url = url;
+        this.fullUrl = url + (queryString ? "" : "?" + queryString);
 
         return new Promise<void>((resolve, reject) => {
-            let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
-            let eventSource = new EventSource(connectUrl);
+            let eventSource = new EventSource(this.fullUrl);
 
             try {
                 eventSource.onmessage = (e: MessageEvent) => {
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -63,7 +66,10 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
             var supportedTransports = options.Transports;
 
             // Server sent events transport
-            if (context.Request.Path.StartsWithSegments(path + "/sse"))
+            // GET /{path}
+            // Accept: text/event-stream
+            var headers = context.Request.GetTypedHeaders();
+            if (headers.Accept?.Contains(new Net.Http.Headers.MediaTypeHeaderValue("text/event-stream")) == true)
```
davidfowl: Might be overkill for typed headers...

anurse:
> GET /{path}/negotiate

Psh, the **real** REST way to do this would be an `OPTIONS` request and encoding the Connection info in HTTP headers in the response :trollface:

src/Microsoft.AspNetCore.Sockets.Client/LongPollingTransport.cs
```diff
@@ -45,9 +44,9 @@ public Task StartAsync(Uri url, IChannelConnection<SendMessage, Message> applica
             _application = application;
 
             // Start sending and polling (ask for binary if the server supports it)
-            var pollUrl = Utils.AppendQueryString(Utils.AppendPath(url, "poll"), "supportsBinary=true");
+            var pollUrl = Utils.AppendQueryString(url, "supportsBinary=true");
```
anurse: Use the `Accept` header to indicate binary vs text long polling as well.

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
```diff
@@ -50,11 +50,6 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
         public async Task ProcessRequestAsync(HttpContext context, CancellationToken token)
         {
             var feature = context.Features.Get<IHttpWebSocketConnectionFeature>();
-            if (feature == null || !feature.IsWebSocketRequest)
```
anurse: Should keep this as an `Assert` or an `Exception`

anurse:
Also, update https://github.com/aspnet/SignalR/blob/dev/specs/TransportProtocols.md in this PR so they stay in sync.

src/Microsoft.AspNetCore.Sockets/Transports/WebSocketsTransport.cs
davidfowl: Will assert.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -22,7 +22,7 @@ export class WebSocketTransport implements ITransport {
     connect(url: string, queryString: string = ""): Promise<void> {
         return new Promise<void>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
-            let connectUrl = url + "/ws?" + queryString;
+            let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
```
moozzyk: queryString still can be null. I think something like this should work:
`(queryString ? "" : "?" + queryString);`

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -94,10 +94,10 @@ export class ServerSentEventsTransport implements ITransport {
 
         this.queryString = queryString;
         this.url = url;
-        let tmp = `${this.url}/sse?${this.queryString}`;
 
         return new Promise<void>((resolve, reject) => {
-            let eventSource = new EventSource(`${this.url}/sse?${this.queryString}`);
+            let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
```
moozzyk: store this url and re-use instead of creating it each time for send?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -168,7 +169,8 @@ export class LongPollingTransport implements ITransport {
         this.url = url;
         this.queryString = queryString;
         this.shouldPoll = true;
-        this.poll(url + "/poll?" + this.queryString)
+        let connectUrl = url + (queryString == "" ? "" : "?" + queryString);
```
moozzyk: store the url and reuse

samples/ClientSample/ClientSample.csproj
```diff
@@ -1,11 +1,12 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk">
```
moozzyk: What is the difference?

src/Microsoft.AspNetCore.Sockets.Client/ServerSentEventsTransport.cs
```diff
@@ -44,8 +44,8 @@ public Task StartAsync(Uri url, IChannelConnection<SendMessage, Message> applica
             _logger.LogInformation("Starting {transportName}", nameof(ServerSentEventsTransport));
 
             _application = application;
-            var sseUrl = Utils.AppendPath(url, "sse");
-            var sendUrl = Utils.AppendPath(url, "send");
+            var sseUrl = url;
+            var sendUrl = url;
```
moozzyk: we now just need url only. No reason to have both.

src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs
```diff
@@ -104,8 +110,10 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
 
                 await DoPersistentConnection(endpoint, ws, context, state);
             }
-            else if (context.Request.Path.StartsWithSegments(path + "/poll"))
+            else if (string.Equals(context.Request.Method, HttpMethods.Get))
```
BrennanConroy: `StringComparison`?

samples/ClientSample/ClientSample.csproj
davidfowl: It's a console application.

moozzyk: Yeah, but I was curious what the difference between these Skds was - clearly a console app can use .Web SDK just fine...

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -479,7 +480,7 @@ public void ConnectionReturnsUrlUsedToStartTheConnection()
                 .Returns<HttpRequestMessage, CancellationToken>(async (request, cancellationToken) =>
                 {
                     await Task.Yield();
-                    if (request.RequestUri.AbsolutePath.EndsWith("/send"))
+                    if (request.Method == HttpMethod.Post)
```
mikaelm12: You can use the HttpMethods class here. 
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs#L50

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -555,7 +557,7 @@ public void ConnectionReturnsUrlUsedToStartTheConnection()
...
```
mikaelm12: Here too

test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -137,7 +137,7 @@ public class LongPollingTransportTests
                 .Returns<HttpRequestMessage, CancellationToken>(async (request, cancellationToken) =>
                 {
                     await Task.Yield();
-                    var statusCode = request.RequestUri.AbsolutePath.EndsWith("send")
+                    var statusCode = request.Method == HttpMethod.Post
```
mikaelm12: And here etc..

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
davidfowl: Does it really matter? This is once per connection.

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
davidfowl: That's not necessary, the client has an HttpClient enum and that's what's being used here.

samples/ClientSample/ClientSample.csproj
davidfowl: It actually doesn't work well. I'm not sure why we made it web in the first place. The default is launching IIS express which makes 0 sense here.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
davidfowl: Will do

davidfowl:
🆙 📅

davidfowl:
> Psh, the real REST way to do this would be an OPTIONS request and encoding the Connection info in HTTP headers in the response :trollface:

@anurse how does that play with CORS?

specs/TransportProtocols.md
```diff
@@ -72,7 +72,7 @@ foo: boz
...
+In this transport, the client establishes an SSE connection to `[endpoint-base]`, and the server responds with an HTTP response with a `Content-Type` of `text/event-stream`. The **mandatory** `connectionId` query string value is used to identify the connection to send to. If there is no `connectionId` query string value, a `400 Bad Request` response is returned, if there is no connection with the specified ID, a `404 Not Found` response is returned. Each SSE event represents a single frame from client to server. The transport uses unnamed events, which means only the `data` field is available. Thus we use the first line of the `data` field for frame metadata. The frame body starts on the **second** line of the `data` field value. The first line has the following format (Identifiers in square brackets `[]` indicate fields defined below):
```
anurse: Clarify that the `Accept` header needs to include `text/event-stream`. That's technically part of the SSE "spec" but it's good to be clear.

specs/TransportProtocols.md
```diff
@@ -117,16 +117,17 @@ Long Polling requires that the client poll the server for new messages. Unlike t
...
 
-A Poll is established by sending an HTTP GET request to `[endpoint-base]/poll` with the following query string parameters
+A Poll is established by sending an HTTP GET request to `[endpoint-base]` with the following query string parameters
 
 * `connectionId` (Required) - The Connection ID of the destination connection.
-* `supportsBinary` (Optional: default `false`) - A boolean indicating if the client supports raw binary data in responses
 
-When messages are available, the server responds with a body in one of the two formats below (depending upon the value of `supportsBinary`). The response may be chunked, as per the chunked encoding part of the HTTP spec.
+Adding the `Accept` header with value `application/vnd.microsoft.aspnetcore.endpoint-messages.v1+binary` indicates if the client supports raw binary data in responses.
```
anurse: It's not totally clear that the `Accept` header is required in both cases, even when doing text.

davidfowl: Was it before?

davidfowl:
@anurse I tried the OPTIONS request and it seems to work. It cleans things up a bit (not sure if I'm violating the spec missing Allow headers though). Will update

davidfowl:
Superseded by https://github.com/aspnet/SignalR/pull/470

samples/ClientSample/ClientSample.csproj
moozzyk: I see.

PR #468, opened 2017-05-19
Socket sample nit

mikaelm12:
Just a nit in the Socket Sample that was bothering me 😄

### DIFF
--- a/samples/SocketsSample/EndPoints/MessagesEndPoint.cs
+++ b/samples/SocketsSample/EndPoints/MessagesEndPoint.cs
@@ -29,7 +29,7 @@ namespace SocketsSample.EndPoints
                         // We can avoid the copy here but we'll deal with that later
                         var text = Encoding.UTF8.GetString(message.Payload);
                         text = $"{connection.ConnectionId}: {text}";
-                        await Broadcast(Encoding.UTF8.GetBytes(text), message.Type, message.EndOfMessage);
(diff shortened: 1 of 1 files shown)


moozzyk:
https://github.com/aspnet/SignalR/commit/2d278009b253f3e2e5a1b8bfa61cf3918e414e11#diff-0af8ac0fde56f734a7300aff70a9086a

samples/SocketsSample/EndPoints/MessagesEndPoint.cs
```diff
@@ -26,10 +26,7 @@ public class MessagesEndPoint : EndPoint
                     Message message;
                     if (connection.Transport.Input.TryRead(out message))
                     {
-                        // We can avoid the copy here but we'll deal with that later
-                        var text = Encoding.UTF8.GetString(message.Payload);
-                        text = $"{connection.ConnectionId}: {text}";
-                        await Broadcast(Encoding.UTF8.GetBytes(text), message.Type, message.EndOfMessage);
+                        await Broadcast(message.Payload, message.Type, message.EndOfMessage);
```
davidfowl: Why? It was originally modifying the message to add the connection id.

mikaelm12:
Yeah, I just wanted to avoid the conversion to a string and back to bytes and didn't think we needed the connection id in the message body. But I just closed it in favor of debug-ability.

PR #473, opened 2017-05-20
Progress towards splitting the layers

davidfowl:
- This is based on the work @anurse did in [anurse/endpoint-middleware-spike](https://github.com/aspnet/SignalR/tree/anurse/endpoint-middleware-spike) to
introduce a connection middleware pipeline that mimics much of our http
pipeline. The intent is that this layer will be generic enough to build both
SignalR and Kestrel on top of but we're not there yet. This change makes incremental
progress towards splitting apart sockets and http so that we can add the tcp transport
without breaking everything all at once.
- Created **Microsoft.AspNetCore.Sockets.Abstractions** where the primitives for
sockets live. That includes, `ConnectionContext` (formerly Connection), `EndPoint`,
`ISocketBuilder`, `SocketDelegate`, etc.
- `ConnectionContext` isn't in it's final form as yet, it still very closely mirrors
the original Connection object we had so that tests continue to pass.
- The `HttpConnectionDispatcher` doesn't know about `EndPoint` anymore, it just cares
about invoking the `SocketDelegate`.
- `EndPointOptions` has been removed as part of this change as it coupled http specific configuration
to the end point type. There's a new `HttpSocketOptions` that needs to be passed into `MapSocket` calls.
- Updated the tests to deal with the API changes.


src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
```diff
@@ -0,0 +1,27 @@
...
+            Transport = transport;
+            ConnectionId = id;
+        }
+
+        public override string ConnectionId { get; }
+
+        public override IFeatureCollection Features { get; } = new FeatureCollection();
+
+        public override ClaimsPrincipal User { get; set; }
+        public override ConnectionMetadata Metadata { get; } = new ConnectionMetadata();
```
mikaelm12: nit: Space

src/Microsoft.AspNetCore.Sockets.Abstractions/ISocketBuilder.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.Sockets.Abstractions/Microsoft.AspNetCore.Sockets.Abstractions.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
mikaelm12: Don't forget to add this to the Coherence packages list

src/Microsoft.AspNetCore.Sockets.Abstractions/SocketBuilder.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.Sockets.Abstractions/SocketBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.Sockets.Abstractions/SocketDelegate.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.Sockets/HttpSocketBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.Sockets/HttpSocketBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    public static class HttpSocketBuilderExtensions
+    {
+        public static ISocketBuilder UseEndPoint<TEndPoint>(this ISocketBuilder socketBuilder) where TEndPoint : EndPoint
```
mikaelm12: Isn't the convention for terminal middleware to start with "Run" instead of "Use"

samples/SocialWeather/Startup.cs
```diff
@@ -32,7 +32,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 app.UseDeveloperExceptionPage();
             }
 
-            app.UseSockets(o => { o.MapEndpoint<SocialWeatherEndPoint>("weather"); });
+            app.UseSockets(o => { o.MapEndPoint<SocialWeatherEndPoint>("weather"); });
```
davidfowl: Not in this PR. It's been o the whole time.

src/Microsoft.AspNetCore.Sockets/HttpSocketBuilderExtensions.cs
davidfowl: Yep!

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public abstract class ConnectionContext : IDisposable
+    {
+        public abstract string ConnectionId { get; }
+
+        public abstract IFeatureCollection Features { get; }
+
+        public abstract ClaimsPrincipal User { get; set; }
+
+        // REVIEW: Should this be changed to items
```
anurse: I think consistency with HttpContext is valuable here, so I do think we should rename it to `Items`.

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            Transport = transport;
+            ConnectionId = id;
+        }
+
+        public override string ConnectionId { get; }
+
+        public override IFeatureCollection Features { get; } = new FeatureCollection();
+
+        public override ClaimsPrincipal User { get; set; }
```
anurse: Is there a reason these aren't backed by Features anymore?

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
davidfowl: I don't want to do that in this change. The main thing here is the assembly changes. I can file an issue to follow up (like getting rid of Metatdata altogether).

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
davidfowl: Didn't want to make those changes yet.

src/Microsoft.AspNetCore.Sockets/HttpSocketBuilderExtensions.cs
davidfowl: I don't think we do this anywhere.

mikaelm12: https://github.com/aspnet/Proxy/blob/dev/src/Microsoft.AspNetCore.Proxy/ProxyExtension.cs#L16

PR #475, opened 2017-05-22
Handle case where scanning happens after dispose.

davidfowl:
- ConnectionManager.Scan would null ref because the timer.Change was being called
after closing all connections.


davidfowl:
@BrennanConroy it is.

BrennanConroy:
Yeah ignore me. I just realized after posting that. Tired

Get Outlook for Android<https://aka.ms/ghei36>

Issue #469, opened 2017-05-20
The server does not terminate long polling connection cleanly 


moozzyk:
If the user send a message and immediately leaves the OnConnectedAsync method like this: https://github.com/aspnet/SignalR/blob/8f18ff44234e038c6728f9798faf970df3244fad/test/Microsoft.AspNetCore.SignalR.Tests/EchoEndPoint.cs#L11-L14 the connection always closes with 404 error on the client. The problem is that when the user leaves the OnConnectedAsync we immediately clean up the connection but since we just sent data to the client there is no pending poll. As a result when the poll request arrives from the client the connection has already been cleaned up and the poll request is rejected with 404.

davidfowl:
Good catch.

PR #476, opened 2017-05-22
Fix long polling graceful shutdown

davidfowl:
- If a message was sent and the application was ended in the same
poll request, the connection would be closed by the time the next poll
came in which resulted in a 404. This change waits until the transport writes
all data before closing it.
- Also fixed a test so that the exception started to show client side.

#469


moozzyk:
👍  for fixing the test. I found this bug with JS client but then was surprised that this test did not fail and found that the exception always happened but the test always ignored it.

Issue #416, opened 2017-04-24
Webpack builds signalr-client incorrectly 


SergeiNET:
I tried to use signalr-client with Angular 2 and webpack. 
Webpack throws following warnings:

> WARNING in ./~/signalr-client/dist/src/index.js
> 3:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
> 
> WARNING in ./~/signalr-client/dist/src/Connection.js
> 11:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
> 
> WARNING in ./~/signalr-client/dist/src/HttpClient.js
> 3:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
> 
> WARNING in ./~/signalr-client/dist/src/HubConnection.js
> 3:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
> 
> WARNING in ./~/signalr-client/dist/src/Transports.js
> 11:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
> 
> WARNING in ./~/signalr-client/dist/src/Formatters.js
> 3:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
> 
> WARNING in ./~/signalr-client/dist/src/Message.js
> 3:24-31 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

When I tried to load my application I got following error: 

> Uncaught Error: Cannot find module "."
>     at webpackMissingModule (vendor.bundle.js:83335) [<root>]
>     at Object.defineProperty.value (vendor.bundle.js:83335) [<root>]
>     at Object.<anonymous> (vendor.bundle.js:83344) [<root>]
>     at __webpack_require__ (inline.bundle.js:53) [<root>]
>     at Object.173 (main.bundle.js:177) [<root>]
>     at __webpack_require__ (inline.bundle.js:53) [<root>]
>     at Object.172 (main.bundle.js:122) [<root>]
>     at __webpack_require__ (inline.bundle.js:53) [<root>]
>     at Object.171 (main.bundle.js:78) [<root>]
>     at __webpack_require__ (inline.bundle.js:53) [<root>]
>     at Object.162 (main.bundle.js:24) [<root>]
>     at __webpack_require__ (inline.bundle.js:53) [<root>]
>     at Object.559 (main.bundle.js:335) [<root>]
[...]

PR #483, opened 2017-05-23
Adding Webpack2/jest + latest typescript

tebeco:
* Updating Typescript to ^2.3.2 (latest ?)
* DOES NOT emit .js .mas.js .d.ts unless explicit build
* Adding TsLint (fell free to update tslint.json)
* for vsCode integration:
 Install TsLint extension
 edit settings like 
```
    "tslint.exclude": "**/node_modules/**/*.ts",
    "tslint.configFile": "tslint.json",
    "tslint.alwaysShowRuleFailuresAsWarnings": true
```

* Adding jest on top of jasmine
* Addin test watcher mode (over jest)
* Adding test coverage (jest)
* Adding Webpack2 for building and bundling: UMD, (see webpack.config.js)
* Adding npm scripts integration :
`npm run build` : clean / check lint / run tests / webpack build
`npm run lint` : run linter
`npm test` to run once the test
`npm run test:watch` : to run a watcher dedicated to test : you'll have to select a mode (`a` for example for all test)
`npm run test:cover` : to get details on the code coverage


tebeco:
poke @anurse (i see you're working on typescript stuff)
i putted stuff in a new folder for now
and i did not integrated in either with VS or in the CI <= maybe @davidfowl / @DamianEdwards  on that one

waiting for feedback before doing so

tebeco:
Just squashed and force push + Merge to avoid confusion

anurse:
I'm not sure I understand exactly what the goal here is. Are you changing the TypeScript client to build using WebPack? If so, we need to evaluate the reasons for doing so. Also, @moozzyk is the main owner of the TypeScript client.

tebeco:
I tried to integrate SignalR in a React application
So I created a simple "create-react-app" in typescript
And did the npm install (with the specified npm registry)

It seems that the imports / intellisens / Goto definitions / build works
Like :
import { HubConnection } from 'signalr-client';
Let x = new HubConnection(......)

But it goes really bad at runtime there was console error about not finding modules and so on

I've changed the import to something dirty like :
const signalR = require('/node_modules/signalr-client/signalr-client.js');
let x = new signalr.HubConnection(....);

Works but it's JavaScript no intelisens, no typeguard,....
But it works
So something is going on with exports modules

That's why I moved to webpack2 to check the bundling with another tool/config

And I just tested it ... works really well, I can add a react sample of a Tchat application in react to the pull request

So far I'm using Npm link to test my produced package

Conclusion so far :
There's something wrong with either module/export/bundling

Also webpack comes with tree shaking so it will bundle only imported code from dependencies not everything
I don't know for browserify, never used it

Also it seems that here tsc was used to produce umd
I took another approach to let tsc produce common js 
And webpack is responsible to produce UMD

moozzyk:
There is an [issue](https://github.com/aspnet/SignalR/issues/416#issuecomment-303677411) about using ts-client with webpack which definitely needs to be looked at. However we cannot merge this PR. We can't maintain multiple copies of the same code. Rather we need to figure out how to build things so that they work for most of the users. One of the reasons I have not looked at the user experience when working with TS client yet is that we are still doing big changes at the lower layers on the stack and things are changing rapidly. I am going to close this PR for now but will link it to the bug tracking the webpack issue and we may use some ideas when we finally get to this.

tebeco:
@moozzyk

I can gladly make it in the original ts-client folder
I created it in a separate folder to start this discussion because I'm stuck for now when trying to actually use signalR

I understand works has to be done anyway but nothing is lost after this integration right ?

moozzyk:
As I said, I have not looked into user experience yet and webpack yet. There are some ideas about how to make the ts-client work with webpack described [here](https://github.com/aspnet/SignalR/issues/416#issuecomment-303677411). If you are still blocked and you want to have something that works in your scenario you can use code you wrote but we are not going to merge these changes for the reasons I stated above.

tebeco:
Do you think a PR to fix linting / upgrading typescript etc but nos touching tools is acceptable ?

moozzyk:
Sure. Just make it a single purpose. Make clear what you are trying to achieve and why. Make sure existing test pass. If you change code/add code add tests.

PR #484, opened 2017-05-23
Add data received logs to SSE and LongPolling in TS client

BrennanConroy:


mikaelm12:
Can we also add logging for the on closed events?

anurse:
filed #485 because using `console.log` makes me nervous.

tebeco:
Yup
console.xxx are forbidden by default in most linter ;)
Exception done for console.error()

PR #486, opened 2017-05-24
Change targets

mikaelm12:
This is still blocked by some upstream repos but making the pr now so it's ready to go.
@muratg @Eilon @JunTaoLuo

### DIFF
--- a/samples/ChatSample/ChatSample.csproj
+++ b/samples/ChatSample/ChatSample.csproj
@@ -3,12 +3,16 @@
   <Import Project="..\..\build\dependencies.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <UserSecretsId>aspnet-ChatSample-f11cf018-e0a8-49fa-b749-4c0eb5c9150b</UserSecretsId>
     <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8</PackageTargetFallback>
     <!-- Don't create a NuGet package -->
     <IsPackable>false</IsPackable>
   </PropertyGroup>
+
+ <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+     <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETStandardLibraryNETFrameworkVersion)" PrivateAssets="All" />
+ </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Redis\Microsoft.AspNetCore.SignalR.Redis.csproj" />
--- a/samples/ChatSample/ChatSample.csproj
+++ b/samples/ChatSample/ChatSample.csproj
@@ -5,7 +5,6 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <UserSecretsId>aspnet-ChatSample-f11cf018-e0a8-49fa-b749-4c0eb5c9150b</UserSecretsId>
-    <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8</PackageTargetFallback>
     <!-- Don't create a NuGet package -->
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/samples/ClientSample/ClientSample.csproj
+++ b/samples/ClientSample/ClientSample.csproj
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\dependencies.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <!-- Don't create a NuGet package -->
     <IsPackable>false</IsPackable>
     <OutputType>Exe</OutputType>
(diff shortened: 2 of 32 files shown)


src/Microsoft.AspNetCore.SignalR.Client/Microsoft.AspNetCore.SignalR.Client.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>Client for ASP.NET Core SignalR</Description>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
moozzyk: Shouldn't the client remain targeting netstandard1.3?

test/Microsoft.AspNetCore.SignalR.Tests/Microsoft.AspNetCore.SignalR.Tests.csproj
```diff
@@ -3,7 +3,8 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
```
moozzyk: This does not seem right. Wouldn't the second TargetFrameworks overwrite the first and you would effectively always end up with `netcoreapp2.0;net461`? Note the order in: https://github.com/aspnet/SignalR/pull/486/files#diff-8a366c3f3e35af025106ceaa9af5f7afR6

src/Microsoft.AspNetCore.SignalR.Client/Microsoft.AspNetCore.SignalR.Client.csproj
mikaelm12: So, should the client and common projects still target 1.3?

test/Microsoft.AspNetCore.SignalR.Tests/Microsoft.AspNetCore.SignalR.Tests.csproj
JunTaoLuo: Yea this is wrong

samples/ClientSample/ClientSample.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\dependencies.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
```
JunTaoLuo: How does this compile without NS.Library.NETFramework? I think we need that reference here.

JunTaoLuo: Same with all the samples when targeting net461

samples/SocketsSample/SocketsSample.csproj
```diff
@@ -3,7 +3,8 @@
   <Import Project="..\..\build\dependencies.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <PackageTargetFallback Condition=" '$(TargetFrameworks)' == 'netcoreapp2.0' ">$(PackageTargetFallback);portable-net45+win8+wp8+wpa81</PackageTargetFallback>
```
JunTaoLuo: I misspoke about this yesterday, you need `<PackageTargetFallback Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">$(PackageTargetFallback);portable-net45+win8+wp8+wpa81</PackageTargetFallback>`

moozzyk: Out of curiosity - why do we need it all?

JunTaoLuo: We might not need it  which is why I wanted @mikaelm12 to try removing it first. But even if we do include it, it should be TargetFramework singular not TargetFrameworks plural.

samples/ChatSample/ChatSample.csproj
```diff
@@ -3,13 +3,17 @@
   <Import Project="..\..\build\dependencies.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <UserSecretsId>aspnet-ChatSample-f11cf018-e0a8-49fa-b749-4c0eb5c9150b</UserSecretsId>
     <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8</PackageTargetFallback>
```
moozzyk: I still don't know why we need this fallback (what happens if it were removed). It also looks different from the fallback you added here: https://github.com/aspnet/SignalR/pull/486/files/46a28efc05b931c8b0389fa7b1e7041a3e6bd8c1..574e53b6514e34f40f4d04c38b8b72818afc8446#diff-502c00f36596d80fb0e0159c90a0b43dR7

samples/WebSocketSample/WebSocketSample.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\dependencies.props" />
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
```
JunTaoLuo: Is this sample still missing the NS.Library.NETFramework?

JunTaoLuo: Actually maybe since there are no dependencies, it's not needed. Make sure you can run this sample and it should be sufficient.

PR #487, opened 2017-05-24
Split http and non-http layers

davidfowl:
- This change introduces Microsoft.AspNetCore.SignalR.Http
and Microsoft.AspNetCore.Sockets.Http which expose extension methods
on IAppBuilder for wiring up a sockets and signalr pipeline.


davidfowl:
@anurse am merging this before @mikaelm12 's ns2.0 changes. Feel free to make changes will I enjoy the 🌞 in hawaii 😄

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionList.cs
```diff
@@ -34,8 +33,7 @@ public void Add(ConnectionContext connection)
 
         public void Remove(ConnectionContext connection)
         {
-            ConnectionContext dummy;
-            _connections.TryRemove(connection.ConnectionId, out dummy);
+            _connections.TryRemove(connection.ConnectionId, out var dummy);
```
anurse: Given that @davidfowl is in Hawaii, I think the only way this nit is being addressed is if you change it yourself :P

moozzyk: Hawaii can't be an excuse for flaws of this caliber! ;)

davidfowl:
Oh and can somebody update Coherence.csv

PR #490, opened 2017-05-24
AuthZ: Use new PolicyEvaluator

HaoK:
@BrennanConroy 

Depends on https://github.com/aspnet/Security/pull/1193

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Extensions.DependencyInjection
         public static ISignalRBuilder AddSignalR(this IServiceCollection services)
         {
             services.AddSockets();
+            services.AddAuthorizationPolicyEvaluator();
             return services.AddSignalRCore();
         }
     }
--- a/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
@@ -10,7 +10,6 @@ namespace Microsoft.Extensions.DependencyInjection
         public static ISignalRBuilder AddSignalR(this IServiceCollection services)
         {
             services.AddSockets();
-            services.AddAuthorizationPolicyEvaluator();
             return services.AddSignalRCore();
         }
     }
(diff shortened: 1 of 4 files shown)


BrennanConroy:
Looks good, should we add a test or 2 for forbids?

davidfowl:
Why are the builds still failing?

BrennanConroy:
> Depends on aspnet/Security#1193

Which hasn't been merged

src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
```diff
@@ -10,6 +10,7 @@ public static class SignalRDependencyInjectionExtensions
         public static ISignalRBuilder AddSignalR(this IServiceCollection services)
         {
             services.AddSockets();
+            services.AddAuthorizationPolicyEvaluator();
```
davidfowl: This is in the wrong place. It should be in AddSockets not AddSignalR

davidfowl:
The dev branch still looks red to me. Do we just need to force a rebuild?

HaoK:
Yeah individual repos broken by other repos need to be manually triggered

HaoK:
I'll add a few forbid tests and move the Add to the right place

HaoK:
Updated PR, looks like the tests needed to be switched over to Auth 2.0 as well

moozzyk:
@HaoK - I can no longer build SignalR repo. The build fails with: Unable to resolve 'Microsoft.AspNetCore.Authorization.Policy (>= 2.0.0)' for '.NETCoreApp,Version=v2.0'. [C:\source\SignalR-Core\.build\KoreBuild.proj]. The package is missing from aspnetcore-ci-dev.

HaoK:
Yup, still working on the universe coherence build, was a 11 repo change, so I wouldn't expect packages til later tonight

HaoK:
If you want to work against the volitale packages so you can work against the new policy, you can add <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-volatile-dev/api/v3/index.json" /> to your nuget.config

moozzyk:
Thanks! Will try the volatile feed.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -693,31 +691,23 @@ public class HttpConnectionDispatcherTests
...
+            // would hang if EndPoint was running
+            await dispatcher.ExecuteAsync(context, options, app).OrTimeout();
 
-            Assert.Equal(StatusCodes.Status200OK, context.Response.StatusCode);
-            Assert.Equal("T12:T:Hello, World;", GetContentAsString(context.Response.Body));
+            Assert.Equal(StatusCodes.Status403Forbidden, context.Response.StatusCode);
         }
 
         [Fact]
-        public async Task AllPoliciesRequiredForAuthorizedEndPoint()
```
BrennanConroy: Why did you remove this test?

HaoK: Oops, I used it as the basis for the new forbid test, but must have deleted the original accidentally when I did the conversion to the new auth 2.0 stuff.

PR #492, opened 2017-05-25
Change Base64 to Base64Encoder

BrennanConroy:


src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/MessageFormatUtils.cs
```diff
@@ -13,22 +13,12 @@ public static byte[] DecodePayload(byte[] inputPayload)
...
-                if (inputPayload.Length > 1 && inputPayload[inputPayload.Length - 2] == '=')
-                {
-                    decodedLength -= 1;
-                }
+                var decodedLength = Base64Encoder.ComputeDecodedLength(inputPayload);
 
                 // Allocate a new buffer to decode to
                 var decodeBuffer = new byte[decodedLength];
-                if (Base64.Decode(inputPayload, decodeBuffer) != decodedLength)
+                if (!Base64Encoder.TryDecode(inputPayload, decodeBuffer, out _, out var bytesWritten) ||
```
anurse: I don't think you need to capture `bytesWritten`. That was only there when I didn't have `TryDecode` and wanted some way to make sure the decode succeeded.

Issue #163, opened 2017-01-27
When a hub cannot be created longPolling transport returns 204 instead of an error.


moozzyk:
The hub could not be created and the server closed the connection. In case of longPolling the connection was closed with 204 even though there was an error. 
For SSE server returns 404: GET http://localhost:5000/notification/sse?id=2d19ab96-1679-485d-b8df-0bd9bcbed0fd 404 (Not Found) - does not seem to be correct.
For websockets transport the the webSocket is closed with 1011 (which is correct)

![image](https://cloud.githubusercontent.com/assets/1438884/22381896/4b49b840-e478-11e6-90f1-fc3a9deb7629.png)

```
fail: Microsoft.AspNetCore.SignalR.HubEndPoint[0]
      Error when invoking OnConnectedAsync on hub.
System.InvalidOperationException: Unable to resolve service for type 'Microsoft.AspNetCore.Http.IHttpContextAccessor' while attempting to activate 'SignalRCoreTest.NotificationHub'.
   at Microsoft.Extensions.Internal.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.Internal.ActivatorUtilities.CreateInstance[T](IServiceProvider provider, Object[] parameters)
   at Microsoft.AspNetCore.SignalR.DefaultHubActivator`2.Create()
   at Microsoft.AspNetCore.SignalR.HubEndPoint`2.<HubOnConnectedAsync>d__9.MoveNext()
info: Microsoft.AspNetCore.Sockets.Transports.LongPollingTransport[0]
      Terminating Long Polling connection by sending 204 response.
fail: Microsoft.AspNetCore.Sockets.ConnectionManager[0]
      Failed disposing connection e4eb6bf0-b08c-4899-a8ba-04648ec1f114
System.InvalidOperationException: Unable to resolve service for type 'Microsoft.AspNetCore.Http.IHttpContextAccessor' while attempting to activate 'SignalRCoreTest.NotificationHub'.
   at Microsoft.Extensions.Internal.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.Internal.ActivatorUtilities.CreateInstance[T](IServiceProvider provider, Object[] parameters)
   at Microsoft.AspNetCore.SignalR.DefaultHubActivator`2.Create()
```

Also each time it happens you will see:

```
[...]

PR #496, opened 2017-05-27
Closing transports with error if hubs cannot be created

moozzyk:
- Preventing from closing long polling transport with 204 in case of
error
- Reporting an error to the client if WebSocket was not closed normally

Note in case of ServerSentEvents it is not possible on the client to
tell the difference between when the server closed event stream due to
an exception or because the client left OnConnectedAsync. In both cases
the client sees only that the stream was closed.

Part of: #163

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -125,7 +125,7 @@ export class ServerSentEventsTransport implements ITransport {
 
                     // don't report an error if the transport did not start successfully
                     if (this.eventSource && this.onClosed) {
-                        this.onClosed(new Error(e.message));
+                        this.onClosed(new Error(e.message || "Error occurred"));
                     }
                 }
 
--- a/client-ts/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
@@ -25,10 +25,8 @@ namespace Microsoft.AspNetCore.SignalR.Test.Server
 
             app.UseFileServer();
             app.UseSockets(options => options.MapEndPoint<EchoEndPoint>("echo"));
-            app.UseSignalR(routes =>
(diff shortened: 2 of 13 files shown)


davidfowl:
The change looks good overall but I'm wondering if there's a usage pattern issue with channels. This effectively means that end point code has to do the same. /cc @stephentoub

davidfowl:
Don't we also have a first class way to close with an error?

moozzyk:
We are already [closing](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets/ConnectionState.cs#L57-L68) the channels with errors in the dispatcher in case of failures.

davidfowl:
That's not the issue, the issue is the fact that the exception isn't observable via WaitForRead/WaitToWrite

moozzyk:
Ahh, this. Yes, this would be useful.

stephentoub:
What's the desired behavior?  You want the exceptions that failed the channel to be returned (asynchronously thrown) from WaitForRead/WriteAsync rather than it returning false?

davidfowl:
> What's the desired behavior? You want the exceptions that failed the channel to be returned (asynchronously thrown) from WaitForRead/WriteAsync rather than it returning false?

I think WaitForRead/WaitForWriteAsync should throw if the channel is completed with an error. If you want a model where you want to be exception free then just don't complete the channel with an error.

stephentoub:
> I think WaitForRead/WaitForWriteAsync should throw if the channel is completed with an error

Right, that's what I was asking, but specifically, what exception?  You'd want it to be the one that faulted the channel, or some new "the channel has faulted" exception?  I assume the former; that's actually how I used to have it, and I changed it, but we can go back.

davidfowl:
The former, we're trying to flow the exception from the application to the transport in the case of SignalR (and vice versa). So if for example the websocket connection dies in the middle of the application processing, the app should see that exception.

stephentoub:
Ok, thanks.  I'll put up a PR in corefxlab as soon as I figure out how to workaround the infrastructure failures I'm currently hitting in the repo :)

PR #499, opened 2017-05-31
Preserve Order and remove task queue in Redis

BrennanConroy:
We shouldn't need the task queues anymore, its an artifact from when we had `connection.Channel.Output.WriteAsync(...)`. Also added `PreserveAsyncOrder` to the redis connection so messages are received in order

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -59,29 +59,23 @@ namespace Microsoft.AspNetCore.SignalR.Redis
                 // TODO: We could support reconnecting, like old SignalR does.
                 throw new InvalidOperationException("Connection to redis failed.");
             }
+            _redisServerConnection.PreserveAsyncOrder = true;
             _bus = _redisServerConnection.GetSubscriber();
-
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -59,29 +59,23 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
 
                 var message = DeserializeMessage(data);
 
                 // TODO: This isn't going to work when we allow JsonSerializer customization or add Protobuf
-                var tasks = new List<Task>(_connections.Count);
 
                 foreach (var connection in _connections)
                 {
-                    tasks.Add(WriteAsync(connection, message));
+                    await WriteAsync(connection, message);
```
davidfowl: Does subscribe support an async `Task` callback? If no, things can still happen in the wrong order since the callback would be async void and the library would think the call is done after the first await.

BrennanConroy:
Yep, looking into alternatives

BrennanConroy:
Closing for now

BrennanConroy:
https://github.com/StackExchange/StackExchange.Redis/issues/639

PR #503, opened 2017-06-01
Discover hub methods at startup time

davidfowl:
- Errors will show up earlier as a result instead of cryptic
first connect errors.


moozzyk:
Do we have any tests for this scenario?

davidfowl:
@moozzyk for what? Startup failures specifically or failures in general?

moozzyk:
Errors in discovering hub methods + a simple E2E test showing that it actually fails with the right exception at the right time.

davidfowl:
We don't have any end to end tests that test startup failure, I'll add one. We do have tests that verify errors with hub discovery https://github.com/aspnet/SignalR/blob/091158fbb0694bab10edd64be34df23e8cf1bd39/test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs#L324

moozzyk:
Travis failed again with "Child node "19" exited prematurely.". I sent an email to figure out how to investigate. If no one knows we will probably need to talk to Travis support or MSBuild folks - this is happening for more than half of our builds...

davidfowl:
🆙 📅

PR #504, opened 2017-06-01
Remove unnecessary build step and reference

BrennanConroy:
Python is already installed and openssl was only needed for 1.X I believe.


BrennanConroy:
Merging since the only failure is the flaky websocket test

PR #505, opened 2017-06-01
IGNORE - Dumping msbuild failure logs

moozzyk:
Testing...


davidfowl:
@moozzyk this worked out?

moozzyk:
Yes. The reason was the binary logger.

PR #507, opened 2017-06-02
Removed custom websocket implementation

davidfowl:
- Use the default websocket middleware
- Rewrote TestWebSocketConnectionFeature to use Channels instead of pipes
- The server basically copies what the client did (which isn't optimized at the moment)


src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
```diff
@@ -126,69 +117,93 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
...
+                    var truncBuffer = new ArraySegment<byte>(buffer.Array, 0, receiveResult.Count);
+                    incomingMessage.Add(truncBuffer);
+                    totalBytes += receiveResult.Count;
+                } while (!receiveResult.EndOfMessage);
 
-            // Write the message to the channel
-            return _application.Output.WriteAsync(message);
-        }
+                // Making sure the message type is either text or binary
+                Debug.Assert((receiveResult.MessageType == WebSocketMessageType.Binary || receiveResult.MessageType == WebSocketMessageType.Text), "Unexpected message type");
```
anurse: This isn't going to work for multi-part messages. Only the first message will have a `Binary` or `Text` message type, the remaining messages are of type `Continuation`. You need to capture the type of the first frame of the message for use here.

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
```diff
@@ -126,69 +117,93 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
...
+                const int bufferSize = 4096;
+                var totalBytes = 0;
+                WebSocketReceiveResult receiveResult;
+                do
+                {
+                    var buffer = new ArraySegment<byte>(new byte[bufferSize]);
 
-            LogFrame("Receiving", frame);
+                    // Exceptions are handled above where the send and receive tasks are being run.
+                    receiveResult = await socket.ReceiveAsync(buffer, CancellationToken.None);
```
anurse: We should enforce that after the first frame of a message, all remaining frames **must** be of type `Continuation` or `Close`. Receiving `Text (Fin=false)` followed by `Text (Fin=true)` is a protocol error (the second message must be `Continuation`) and we should close the socket when that occurs, with status code `1002` (Protocol Error)

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
```diff
@@ -126,69 +117,93 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
...
+            while (true)
             {
-                return Task.CompletedTask;
-            }
+                const int bufferSize = 4096;
+                var totalBytes = 0;
+                WebSocketReceiveResult receiveResult;
+                do
+                {
+                    var buffer = new ArraySegment<byte>(new byte[bufferSize]);
```
anurse: I assume we'll want to pool this memory eventually.

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
```diff
@@ -126,69 +117,93 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
...
-                                opcode: _lastFrameIncomplete ? WebSocketOpcode.Continuation : opcode,
-                                payload: ReadableBuffer.Create(message.Payload));
+                            var messageType = message.Type == MessageType.Binary ?
+                                WebSocketMessageType.Binary :
+                                WebSocketMessageType.Text;
 
-                            _lastFrameIncomplete = !message.EndOfMessage;
+                            if (_logger.IsEnabled(LogLevel.Debug))
+                            {
+                                _logger.LogDebug(
```
anurse: Use the logging syntax that maintains structured log messages: `_logger.LogDebug("Sending frame: Fin={fin}, Payload={length} bytes", message.EndOfMessage, message.Payload.Length);`

I wish we could some how tap in to `FormattedString` to make this work while still preserving structured logging :)

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
```diff
@@ -126,69 +117,93 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
...
+                    message = new Message(messageBuffer, messageType, receiveResult.EndOfMessage);
+                }
+                else
+                {
+                    var buffer = new byte[incomingMessage[0].Count];
+                    Buffer.BlockCopy(incomingMessage[0].Array, incomingMessage[0].Offset, buffer, 0, incomingMessage[0].Count);
+                    message = new Message(buffer, messageType, receiveResult.EndOfMessage);
+                }
+
+                _logger.LogInformation("Passing message to application. Payload size: {0}", message.Payload.Length);
```
anurse: Give `0` a name: `{length}`

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
```diff
@@ -126,69 +117,93 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
...
-            var effectiveOpcode = frame.Opcode;
-            if (frame.Opcode == WebSocketOpcode.Continuation)
-            {
-                effectiveOpcode = _lastOpcode;
-            }
-            else
-            {
-                _lastOpcode = frame.Opcode;
-            }
+                    _logger.LogDebug("Message received. Type: {messageType}, size: {size}, EndOfMessage: {endOfMessage}",
```
anurse: See, you do it right here :).

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
```diff
@@ -88,23 +79,23 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<Message>
 
                 // Shutting down because we received a close frame from the client.
                 // Complete the input writer so that the application knows there won't be any more input.
-                _logger.LogDebug("Client closed connection with status code '{0}' ({1}). Signaling end-of-input to application", receiving.Result.Status, receiving.Result.Description);
+                _logger.LogDebug("Client closed connection with status code '{0}' ({1}). Signaling end-of-input to application", receiving.Result.CloseStatus, receiving.Result.CloseStatusDescription);
```
anurse: I know this was already wrong when you started, but update this one to use named format specifiers too.

BrennanConroy:
Remove `pip install` from travis.yml!

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
moozzyk: `WebSocketMessageType` enum [does not have](https://msdn.microsoft.com/en-us/library/system.net.websockets.websocketmessagetype(vs.110).aspx) the `Continuation` member...

src/Microsoft.AspNetCore.Sockets.Http/Transports/WebSocketsTransport.cs
davidfowl: Already works

PR #510, opened 2017-06-02
React to Base64 API changes

BrennanConroy:


src/Microsoft.AspNetCore.Sockets.Common/Internal/Formatters/MessageFormatUtils.cs
```diff
@@ -13,11 +14,11 @@ public static byte[] DecodePayload(byte[] inputPayload)
             if (inputPayload.Length > 0)
             {
                 // Determine the output size
-                var decodedLength = Base64Encoder.ComputeDecodedLength(inputPayload);
+                var decodedLength = Base64Decoder.ComputeDecodedLength(inputPayload);
 
                 // Allocate a new buffer to decode to
                 var decodeBuffer = new byte[decodedLength];
-                if (!Base64Encoder.TryDecode(inputPayload, decodeBuffer, out _, out var _))
+                if (Base64.Decoder.Transform(inputPayload, decodeBuffer, out _, out var _) != TransformationStatus.Done)
```
moozzyk: what are the other values that can be returned here? Do we need to handle them?

BrennanConroy: Not sure, made the PR to unblock the mirror

moozzyk: I think it is fine to merge like this to unblock the mirror but then make sure that we don't need to handle other cases. If we do, we will do it separately.

Issue #506, opened 2017-06-02
Port IsValidHubMethod from SignalR/SignalR


davidfowl:
Right now we detect too many methods:

```

dbug: Microsoft.AspNetCore.SignalR.HubEndPoint[0]
      Hub method 'ToString' is bound
dbug: Microsoft.AspNetCore.SignalR.HubEndPoint[0]
      Hub method 'Equals' is bound
dbug: Microsoft.AspNetCore.SignalR.HubEndPoint[0]
      Hub method 'GetHashCode' is bound
dbug: Microsoft.AspNetCore.SignalR.HubEndPoint[0]
      Hub method 'GetType' is bound
```

https://github.com/SignalR/SignalR/blob/b73cbabe69bc48c4cb8d768148e157712a264ce9/src/Microsoft.AspNet.SignalR.Core/Hubs/ReflectionHelper.cs#L30

BrennanConroy:
https://github.com/aspnet/SignalR/commit/3cabb6aeb11d0c80617c31c4eea909ce80d823b9

PR #511, opened 2017-06-02
Don't over-discover Hub methods

BrennanConroy:
#506

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/Internal/HubReflectionHelper.cs
+++ b/src/Microsoft.AspNetCore.SignalR/Internal/HubReflectionHelper.cs
@@ -40,12 +40,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
             // removes methods such as Hub<TClient>.OnConnectedAsync
             var baseType = baseDefinition.GetTypeInfo().IsGenericType ? baseDefinition.GetGenericTypeDefinition() : baseDefinition;
-            if (typeof(Hub<>) == baseType)
-            {
-                return false;
-            }
-
-            return true;
+            return typeof(Hub<>) != baseType;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


davidfowl:
I think it's time to move this state into another type.

BrennanConroy:
Moved to helper class

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -337,6 +337,80 @@ public void HubsCannotHaveOverloadedMethods()
         }
 
         [Fact]
+        public async Task CannotCallStaticHubMethods()
+        {
+            var serviceProvider = CreateServiceProvider();
```
moozzyk: I thought the tests would be more unit-testy..  For instance the original bug is about discovering methods defined on `Object` but there is no test testing this. This can be easily tested without doing clients etc. just by invoking HubReflectionHelper and verifying what method it discovered.

src/Microsoft.AspNetCore.SignalR/Internal/HubReflectionHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            var baseDefinition = methodInfo.GetBaseDefinition().DeclaringType;
+            if (typeof(object) == baseDefinition || methodInfo.IsSpecialName)
+            {
+                return false;
+            }
+
+            // removes methods such as Hub<TClient>.OnConnectedAsync
+            var baseType = baseDefinition.GetTypeInfo().IsGenericType ? baseDefinition.GetGenericTypeDefinition() : baseDefinition;
+            if (typeof(Hub<>) == baseType)
```
moozzyk: nit: just return `typeof(Hub<>) != baseType`?

BrennanConroy:
:up: 📅

PR #514, opened 2017-06-03
Merge DefaultConnectionContext with ConnectionState

davidfowl:
- Removed ConnectionState as a result


samples/ChatSample/Hubs/Chat.cs
```diff
@@ -20,7 +20,7 @@ public Chat(IUserTracker<Chat> userTracker)
         {
             if (!Context.User.Identity.IsAuthenticated)
             {
-                Context.Connection.Dispose();
```
moozzyk: A general comment: How is the user supposed to know that they should do that? What would happen if they threw an exception - would the transport be disposed?

PR #517, opened 2017-06-04
Merge transport and hub protocols

davidfowl:
- This change merges the transport and hub protocols into a single protocol. The
idea being that sockets in a purely streaming layer that sends frames from the underlying
transport. This makes things like TCP possible and doesn't impose a framing layer at the lowest
level. This will make it possible to build servers like kestrel on top of the TCP layer.
- The Message was removed from the lowest layer of the stack and pushed into the hubs layer. Hub invocations
are framed with what was before the transport protocol. Connections also need to state upfront if they support
binary or not. This will determine how data will be serialized to the specific connection.
- Changed the SSE parser and writer to be strictly SSE without any of the transport protocol specific
information.

Notes:
- Out for review early because it's disruptive and cross cutting
- ~Still working on test failures (it just compiled at time of writing)~
- The underlying transport behavior behavior hasn't changed so individual `byte[]` payloads should contain complete messages. We'll remove this assumption in another PR.
- ~We need to remove the outer most T/B from the payload, the transport shouldn't have to write it anymore as the connection is responsible for saying what it supports.~
- ~The protocol spec needs to be updated.~


specs/TransportProtocols.md
```diff
@@ -7,24 +7,15 @@ This document describes the protocols used by the three ASP.NET Endpoint Transpo
...
-
 The only transport which fully implements the duplex requirement is WebSockets, the others are "half-transports" which implement one end of the duplex connection. They are used in combination to achieve a duplex connection.
 
 Throughout this document, the term `[endpoint-base]` is used to refer to the route assigned to a particular end point. The term `[connection-id]` is used to refer to the connection ID provided by the `OPTIONS [endpoint-base]` request.
 
 **NOTE on errors:** In all error cases, by default, the detailed exception message is **never** provided; a short description string may be provided. However, an application developer may elect to allow detailed exception messages to be emitted, which should only be used in the `Development` environment. Unexpected errors are communicated by HTTP `500 Server Error` status codes or WebSockets `1008 Policy Violation` close frames; in these cases the connection should be considered to be terminated.
 
-## `Close` and `Error` frames
```
davidfowl: Decided to kill these since this is being merged into the hub protocol and the individual transport protocols weren't being used. I didn't remove it in the code yet though (we weren't sending these today anyways). I can do that if we agree it's a good idea.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Formatters.spec.ts
```diff
@@ -20,46 +20,20 @@ describe("Text Message Formatter", () => {
...
-        ["T1:T:A;1:asdf:", new Error("Unknown type value: 'asdf'")],
-        ["T1:T:A;1::", new Error("Message is incomplete")],
-        ["T1:T:A;1:AB:", new Error("Message is incomplete")],
-        ["T1:T:A;5:T:A", new Error("Message is incomplete")],
-        ["T1:T:A;5:T:AB", new Error("Message is incomplete")],
-        ["T1:T:A;5:T:ABCDE", new Error("Message is incomplete")],
-        ["T1:T:A;5:X:ABCDE", new Error("Message is incomplete")],
-        ["T1:T:A;5:T:ABCDEF", new Error("Message missing trailer character")],
+        ["ABC", new Error("Invalid length: 'ABC'")],
+        ["1:X:A", new Error("Message is incomplete")],
```
moozzyk: Why is this 'Message is incomplete'? Isn't X supposed to be message type in which case it would be invalid?

davidfowl: Thats the error that gets thrown from the parser.

specs/TransportProtocols.md
moozzyk: At the hub level - Close is when the connection is closed. Error is if we can read/understand the payload.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Formatters.spec.ts
moozzyk: But this error does not seem to be correct - the message type was 'X' which is not a valid message type in the first place.

specs/TransportProtocols.md
```diff
@@ -71,50 +62,17 @@ foo: boz
...
 
-Since there is such a long round-trip-time for messages, given that the client must issue a request before the server can transmit a message back, Long Polling responses contain batches of multiple messages. Also, in order to support browsers which do not support XHR2, which provides the ability to read binary data, there are two different modes for the polling transport.
+In order to support browsers which do not support XHR2, which provides the ability to read binary data, there are two different modes for the polling transport.
```
anurse: Is this still something we care about at this layer? I think the format of the data is negotiated outside this protocol now (JSON/ProtoBuf). The `Content-Type`s would be different, but I think that's actually outside this protocol.

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
```diff
@@ -22,6 +22,6 @@ public abstract class ConnectionContext
         public abstract ConnectionMetadata Metadata { get; }
 
         // TEMPORARY
-        public abstract IChannelConnection<Message> Transport { get; set; }
+        public abstract IChannelConnection<byte[]> Transport { get; set; }
```
anurse: So with this, `Message` is dead yes? Why keep the `T` in the text protocol then? We can just make it `{length}:{payload}` now rather than leaving a vestigial "type" flag in.

src/Microsoft.AspNetCore.Sockets.Common/ContentTypes.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    public static class ContentTypes
+    {
+        public static readonly string TextContentType = "application/vnd.microsoft.aspnetcore.endpoint-messages.v1+text";
```
anurse: This could be simpler now that it's at a higher level. It could live in SignalR and be `application/vnd.microsoft.signalr.v1+json`, and similar for ProtoBuf.

We'd need to expose `Content-Type` in the metadata as well. Also, we need to think about how protocols are negotiated in this model. Is it specified in a header/query string value? `Accept` headers? We can do standard content negotiation for Long Polling, but WebSockets and SSE have fixed header sets and we can't apply request headers easily.

If you want to file an issue for this and do it in a separate PR that's fine, but we should tidy this all up.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Formatters.spec.ts
davidfowl: Yea, no argument from me, fixing it just wasn't part of this change.

davidfowl: Oh wait, there's a bug 😄

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
davidfowl: Message was moved up a layer and still has the type in it.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Formatters.spec.ts
davidfowl: Nope, bad test, missing `;` at the end, fixed it.

specs/TransportProtocols.md
davidfowl: This *is* the transport protocol spec though.

src/Microsoft.AspNetCore.Sockets.Common/ContentTypes.cs
davidfowl: That's a TBD filed https://github.com/aspnet/SignalR/issues/520

test/Microsoft.AspNetCore.SignalR.Tests/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -14,62 +14,53 @@ namespace Microsoft.AspNetCore.Sockets.Common.Tests.Internal.Formatters
...
-        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n", "Hello, World", MessageType.Text)]
-        [InlineData("data: T\r\ndata: Hello, World\r\n\r\ndata: ", "Hello, World", MessageType.Text)]
-        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n:comment\r\ndata: ", "Hello, World", MessageType.Text)]
-        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n:comment", "Hello, World", MessageType.Text)]
-        [InlineData("data: T\r\ndata: Hello, World\r\n\r\n:comment\r\n", "Hello, World", MessageType.Text)]
-        [InlineData("data: T\r\ndata: Hello, World\r\n:comment\r\n\r\n", "Hello, World", MessageType.Text)]
-        [InlineData("data: B\r\ndata: SGVsbG8sIFdvcmxk\r\n\r\n", "Hello, World", MessageType.Binary)]
-        [InlineData("data: B\r\ndata: SGVsbG8g\r\ndata: V29ybGQ=\r\n\r\n", "Hello World", MessageType.Binary)]
-        public void ParseSSEMessageSuccessCases(string encodedMessage, string expectedMessage, MessageType messageType)
+        [InlineData("\r\n", "")]
```
moozzyk: nit: remove duplicate cases

test/Microsoft.AspNetCore.SignalR.Tests/Formatters/ServerSentEventsParserTests.cs
```diff
@@ -191,16 +173,15 @@ public void ParseSSEMessageIncompleteParseResult(string encodedMessage)
...
                 Assert.Equal(expectedMessage, ex.Message);
 
             }
         }
 
         [Theory]
-        [InlineData("data: T\r\ndata: foo\r\n\r\n", "data: T\r\ndata: bar\r\n\r\n", MessageType.Text)]
-        [InlineData("data: B\r\ndata: Zm9v\r\n\r\n", "data: B\r\ndata: YmFy\r\n\r\n", MessageType.Binary)]
-        public async Task ParseMultipleMessagesText(string message1, string message2, MessageType expectedMessageType)
+        [InlineData("data: foo\r\n\r\n", "data: bar\r\n\r\n")]
```
moozzyk: nit: just make it fact?

davidfowl:
@moozzyk I will address your nits in another PR coming soon.

specs/HubProtocol.md
```diff
@@ -321,6 +321,72 @@ Example - The following `Completion` message is a protocol error because it has
...
+* `[Length]` - Length of the `[Body]` field in bytes, specified as UTF-8 digits (`0`-`9`, terminated by `:`). If the body is a binary frame, this length indicates the number of Base64-encoded characters, not the number of bytes in the final decoded message!
+* `[Type]` - A single-byte UTF-8 character indicating the type of the frame, see the list of frame Types below
+* `[Body]` - The body of the message, the content of which depends upon the value of `[Type]`
+
+The following values are valid for `[Type]`:
+
+* `T` - Indicates a text frame, the `[Body]` contains UTF-8 encoded text data.
+* `B` - Indicates a binary frame, the `[Body]` contains Base64 encoded binary data.
+
+Note: If there is no `[Body]` for a frame, there does still need to be a `:` and `;` delimiting the body. So, for example, the following is an encoding of a single text frame `A`: `T1:T:A;`
```
davidfowl: Not designed, could be.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -152,7 +160,10 @@ export class HubConnection {
         });
 
         //TODO: separate conversion to enable different data formats
-        this.connection.send(JSON.stringify(invocationDescriptor))
+        let data = JSON.stringify(invocationDescriptor);
```
davidfowl: I think we should mirror the C# and make an IHubProtocol.

PR #522, opened 2017-06-05
Change Group Add and Remove API

mikaelm12:
Changing the Group add and remove API to accept a connectionId instead of a `ConnectionContext`. This lets you remove any connection from a group.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
@@ -20,8 +20,9 @@ namespace Microsoft.AspNetCore.SignalR
         private long _nextInvocationId = 0;
         private readonly ConnectionList _connections = new ConnectionList();
 
-        public override Task AddGroupAsync(ConnectionContext connection, string groupName)
+        public override Task AddGroupAsync(string connectionId, string groupName)
         {
+            var connection = _connections[connectionId];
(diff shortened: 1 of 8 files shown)


grahamehorner:
IMHO the group functionality needs wider discussion around the process of allowing/granting a connection to a group, which may/may not be via a SignalR message/reponse to the owner/creator of group or some claims based logic prior to adding the connection.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -15,8 +15,9 @@ public class DefaultHubLifetimeManager<THub> : HubLifetimeManager<THub>
         private long _nextInvocationId = 0;
         private readonly ConnectionList _connections = new ConnectionList();
 
-        public override Task AddGroupAsync(ConnectionContext connection, string groupName)
+        public override Task AddGroupAsync(string connectionId, string groupName)
         {
+            var connection = _connections[connectionId];
```
moozzyk: what if the connectionId is incorrect (or null)?

mikaelm12: The `ConnectionList` type will return null if it can't resolve the connectionId that was passed in. I should add a null check here.

mikaelm12: And the other places that I get connections like this

davidfowl: You're going to handle that right and noop.

PR #532, opened 2017-06-06
Add support for Hub Authorize Attribute

BrennanConroy:
Currently redirect is broken for Auth, unrelated to this change

Q: Exposing HttpSocketOptions.AuthorizeData is a little strange


src/Microsoft.AspNetCore.SignalR.Http/HubRouteBuilder.cs
```diff
@@ -16,7 +19,21 @@ public HubRouteBuilder(SocketRouteBuilder routes)
 
         public void MapHub<THub>(string path) where THub : Hub<IClientProxy>
         {
-            _routes.MapSocket(path, builder =>
+            MapHub<THub>(path, null);
```
davidfowl: `socketOptions: null`

src/Microsoft.AspNetCore.SignalR.Http/HubRouteBuilder.cs
```diff
@@ -16,7 +19,21 @@ public HubRouteBuilder(SocketRouteBuilder routes)
...
         public void MapHub<THub>(string path) where THub : Hub<IClientProxy>
         {
-            _routes.MapSocket(path, builder =>
+            MapHub<THub>(path, socketOptions: null);
+        }
+
+        public void MapHub<THub>(string path, Action<HttpSocketOptions> socketOptions) where THub : Hub<IClientProxy>
+        {
+            // find auth attributes
+            var authorizeAttribute = typeof(THub).GetCustomAttribute<AuthorizeAttribute>();
```
davidfowl: pass the inherit flag? here

PR #534, opened 2017-06-06
Add the package-lock.json file to source

natemcmaster:
Per its spec: https://docs.npmjs.com/files/package-lock.json

> This file is intended to be committed into source repositories


moozzyk:
I am sure what the benefit is given that our dependencies are only dev dependencies...

natemcmaster:
It's part of the faster install experience in npm5. Here are the other benefits they describe:

> This file is intended to be committed into source repositories, and serves various purposes:
>
> Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> 
> Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
> 
> To facilitate greater visibility of tree changes through readable source control diffs.
> 
> And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

anurse:
In theory it can still speed up the restoring of dev dependencies, no? Seems good.

moozzyk:
In theory, yes. As pointed by Nate it works in npm5 only which is still in beta and we don't use it anywhere...

.travis.yml
```diff
@@ -6,7 +6,7 @@ env:
     - DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     - DOTNET_CLI_TELEMETRY_OPTOUT: 1
     - SIGNALR_TESTS_VERBOSE: 1
-    - TRAVIS_NODE_VERSION: 6.9.1
+    - TRAVIS_NODE_VERSION: 8.0.0
```
BrennanConroy: More importantly, we need to update all of our agents

natemcmaster:
FYI current stable version of npm is 5.0.3

.travis.yml
moozzyk: Does this work? We currently have this warning: 
`npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.`

https://github.com/gulpjs/gulp/issues/1571

I tried to fix it but it is a dependency that we don't control.

natemcmaster: The travis and appveyor builds passed fine, and it builds locally too.

PR #537, opened 2017-06-07
Added IHubProtocol to ts client

davidfowl:
- Mimics the .NET client structure wise
- Also adds a JsonHubProtocol implementation


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -13,7 +13,6 @@
 
   <ItemGroup>
     <Inputs Include="*.ts;" />
-    <Inputs Remove="Observable.ts" />
```
anurse: \*sigh* VS...

davidfowl: Yea, it *really* can't handle the whole "using tooling" and globbing support....

PR #539, opened 2017-06-07
Fix SocketSample group send

mikaelm12:
The socket sample group send wasn't grabbing the message properly. Just a wrong id

### DIFF
--- a/samples/SocketsSample/wwwroot/hubs.html
+++ b/samples/SocketsSample/wwwroot/hubs.html
@@ -130,7 +130,7 @@ click('leave-group', event => {
 
 click('groupmsg', event => {
     let groupName = getText('target');
-    let message = getText('messages');
+    let message = getText('message');
     invoke(connection, 'SendToGroup', groupName, message);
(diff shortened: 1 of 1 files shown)


samples/SocketsSample/wwwroot/hubs.html
```diff
@@ -130,7 +130,7 @@ <h1 id="head1"></h1>
 
 click('groupmsg', event => {
     let groupName = getText('target');
-    let message = getText('messages');
+    let message = getText('message');
```
moozzyk: I wonder if it would be better to rename this. We have two ids - "message" and "messages" and they are being confused. I think this is the second or third PR trying to get things straight.

mikaelm12: Alright. I'll go through quickly and give more specific ids to the message inputs.

moozzyk:
but you still have two almost identical ids - message and messages. My comment was about these two since they are being confused all the time.

Issue #518, opened 2017-06-05
Remove ITransportFactory and HttpClient from IConnection.StartAsync


davidfowl:
Thinking forward to the future of non-http, we should decouple `IConnection` from http as well. That way Hubs can be used on top of any arbitrary `IConnection` as well.

https://github.com/aspnet/SignalR/blob/831fa7289319b890122161ad02dd17222dfbbca4/src/Microsoft.AspNetCore.Sockets.Client/IConnection.cs#L13

davidfowl:
Notes:
- `IConnection.StartAsync` shouldn't take any arguments and should be moved to Sockets.Abstractions
- `Connection` should be renamed to `HttpConnection`
- `HttpConnection` should let you replace the `HttpMessageHandler` but possibly not the HttpClient itself. This solves a bunch of the ownership issues around disposing the `HttpClient`. We always own the `HttpClient` and you get pluggability via the message handler.

Some design issues with this:
- Should `HubConnection` be tied to Http? Do we need an `HttpHubConnection`?
- How do we pick the transport in the `HttpConnection` case? Is it a constructor argument instead of an argument to StartAsync? That would be much cleaner layering wise.

/cc @moozzyk

moozzyk:
Done.

PR #544, opened 2017-06-09
Re-layer the .NET Client into Http and non-Http

davidfowl:
- Moved IConnection to Sockets.Abstractions and removed
HttpConnection and TransportType dependency.
- Renamed Sockets.Client to Sockets.Client.Http
- Renamed Sockets.Common to Sockets.Common.Http
- Renamed Connection to HttpConnection
- Removed HTTP dependency from HubConnection
- Removed tests that were testing connection logic in HubConnection

#518


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -4,130 +4,26 @@
...
         [Fact]
-        public void CannotCreateHubConnectionWithNullUrl()
-        {
-            var exception = Assert.Throws<ArgumentNullException>(
-                () => new HubConnection((Uri)null, Mock.Of<ILoggerFactory>()));
-            Assert.Equal("url", exception.ParamName);
-        }
-
-        [Fact]
-        public async Task CanDisposeNotStartedHubConnection()
```
moozzyk: I think at least some of these tests should stay. At the moment StartAsync or DisposeAsync just call into corresponding methods in the wrapped connection but if we ever add code to these methods I am almost sure we will forget testing these scenarios.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -15,18 +15,19 @@
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
-    public class Connection : IConnection
+    public class HttpConnection : IConnection
```
moozzyk: Had exactly the same thought actually (including the type name)...

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -37,34 +38,46 @@ public class Connection : IConnection
...
+        {
+        }
+
+        public HttpConnection(Uri url, TransportType transportType, ILoggerFactory loggerFactory, HttpMessageHandler httpMessageHandler)
         {
             Url = url ?? throw new ArgumentNullException(nameof(url));
 
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
-            _logger = _loggerFactory.CreateLogger<Connection>();
+            _logger = _loggerFactory.CreateLogger<HttpConnection>();
```
davidfowl: The HttpMessageHandler makes it a PITA, since it needs to be passed to the `DefaultTransportFactory`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -37,34 +38,46 @@ public class Connection : IConnection
...
-        public Task StartAsync(TransportType transportType) => StartAsync(transportType, httpClient: null);
-        public Task StartAsync(TransportType transportType, HttpClient httpClient)
+        public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactory loggerFactory, HttpMessageHandler httpMessageHandler)
         {
-            if (httpClient == null)
-            {
-                // We are creating the client so store it to be able to dispose
-                _httpClient = httpClient = new HttpClient();
-            }
+            Url = url ?? throw new ArgumentNullException(nameof(url));
```
davidfowl: It was there before 😄

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: I see it now. It's fine.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: Yes, you copied it and have not changed it and also left one that was there - so you are progressing ;)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
davidfowl: I'll add tests that verify those call the underlying connection start/disposeasync

moozzyk: Yeah. I thought about this - it will be easy with Mock.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: but but that's what namespaces are for...

anurse: yeah but nobody uses namespaces below the `using` declarations, so they don't work that way.

mikaelm12:
Don't forget to update the [packages list in Coherence](https://github.com/aspnet/Coherence/blob/dev/packages/packages.csv)

PR #549, opened 2017-06-09
Reorganize and add more logging to Sockets.Http

BrennanConroy:
Moved Sockets.Transports to pubternal also.


davidfowl:
ConnectionId should be used everywhere. Request is is only interesting for long polling.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -46,14 +48,14 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<byte[]>
...
 
-                await ProcessSocketAsync(ws);
+                await ProcessSocketCoreAsync(ws);
             }
-            _logger.LogInformation("Socket closed.");
+            _logger.SocketClosed(context.TraceIdentifier);
         }
 
-        public async Task ProcessSocketAsync(WebSocket socket)
+        public async Task ProcessSocketCoreAsync(WebSocket socket)
```
davidfowl: Why was this renamed?

src/Microsoft.AspNetCore.Sockets.Common.Http/Internal/SocketLoggerExtensions.cs
```diff
@@ -0,0 +1,313 @@
...
+using System;
+using System.Net.WebSockets;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Sockets.Internal
+{
+    public static class SocketLoggerExtensions
+    {
+        // Category: LongPollingTransport
+        private static readonly Action<ILogger, DateTime, string, Exception> _longPolling204 =
```
davidfowl: Where are the connection ids for these logs?

src/Microsoft.AspNetCore.Sockets.Common.Http/Microsoft.AspNetCore.Sockets.Common.Http.csproj
```diff
@@ -13,6 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
```
davidfowl: This shouldn't be in here. It should be in both client and server.

BrennanConroy:
Kestrel has this format:
```
info: Microsoft.AspNetCore.Server.Kestrel[27]
      Connection id "0HL5ILL9H6928", Request id "0HL5ILL9H6928:00000001": request body incoming data rate dropped below 1 bytes/second.
```

I think it looks good, thoughts?

davidfowl:
@BrennanConroy rebase, this change looks good.

BrennanConroy:
@davidfowl Was that in reference to my previous comment?

davidfowl:
Both, the overall change looks good and that format looks good.

PR #559, opened 2017-06-13
Proxy generation Part 1

moozzyk:
Proxy generation Part 1. Contains
- dotnet signalr tool
- metadata based hub discovery

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -13,6 +13,7 @@
     <TestSdkVersion>15.3.0-*</TestSdkVersion>
     <XunitVersion>2.3.0-beta2-*</XunitVersion>
     <RxVersion>3.1.1</RxVersion>
+    <MonoCecilVersion>0.10.0-beta6</MonoCecilVersion>
 
     <!--
       TODO remove in next update of xunit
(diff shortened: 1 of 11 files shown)


build/dependencies.props
```diff
@@ -13,6 +13,7 @@
     <TestSdkVersion>15.3.0-*</TestSdkVersion>
     <XunitVersion>2.3.0-beta2-*</XunitVersion>
     <RxVersion>3.1.1</RxVersion>
+    <MonoCecilVersion>0.10.0-beta6</MonoCecilVersion>
```
moozzyk: @Eilon  - Can we get this approved?

src/Microsoft.AspNetCore.SignalR.Tools/HubDiscovery.cs
```diff
@@ -0,0 +1,225 @@
...
+                                var assemblyDefintion = ModuleDefinition.ReadModule(fullPath, parameters).Assembly;
+                                _assemblyCache[assemblyReferenceName.FullName] = assemblyDefintion;
+                                return assemblyDefintion;
+                            }
+                            catch
+                            {
+                                // Unable to read module
+                                // TODO: Log?
+                            }
+                        }
```
mikaelm12: Remove new line

muratg:
@moozzyk what are we doing about this PR? It's quite stale.

moozzyk:
We need to decide if we want to move forward with this feature in general and if we want (and can) use Roslyn instead of Cecil.

muratg:
I think the feature would be valuable.  @davidfowl thoughts?

davidfowl:
Yes, we still want to do this

Eilon:
I would strongly prefer if we use Roslyn and not Mono Cecil because it's easier for us to bundle everything up to ship.

davidfowl:
We may as well use reflection at this point. Run like an ef tool.

Issue #555, opened 2017-06-12
Hub<T> where T is an interface implementation


muratg:

mikaelm12:
Added: 32ef3eb35522f1f3bf1c90a82c9656cc4ffe57fe

PR #560, opened 2017-06-14
Hub<dynamic>

mikaelm12:
Initial PR for review. Still kind of hacky and still needs tests but I wanted to get some feedback.


anurse:
I'm not sure I like how the `TClient` keeps having to flow through. That may just be a quirk of hacking it up right now, but I think something a little cleaner might be to use wrapping rather than flowing the generics all the way through. I sketched something out here: https://gist.github.com/anurse/1c3b1ad0fe1b7c488ed9a31faca500ad

The idea is that the infrastructure just gives an `IClientProxy` to the `Hub`, and it's just `Hub<T>` that does the magic when needed. It pulls an `IClientProxyFactory` (or maybe `IClientProxyAdaptor`?) out of DI with the appropriate `T` and tells it to "convert" the `IClientProxy` into a `T` for it. We'd register a specific `IClientProxyFactory<dynamic>` as well as an open-generic `IClientProxyFactory<>` implementation for #555 (when we get there). This would also allow users to code-gen their proxies in advance (possibly via our proxy generation tool?) and compile them in.

I did find some stuff about it not being possible to implement an interface using `dynamic` as the generic parameter, which will be tricky (but it's going to affect us no matter what; apparently abstract classes work better). We could get around that by using a separate base class for `Hub<dynamic>` (`DynamicHub`?), or by just putting the `dynamic` logic separately into an extension method, so that any Hub can use it: `Clients.All.Dynamic().Send("foo")`

Thoughts? @moozzyk @davidfowl

PR #563, opened 2017-06-15
Temporarily add a direct reference to NETStandard.Library to unblock build

mikeharder:


mikeharder:
Fixed by: https://github.com/aspnet/SignalR/commit/4812d205eadc1eda69de52b6605e6e41d9c62ecd

PR #565, opened 2017-06-15
Simplify delegate invoking

### LINKED ISSUE #561: Delegate invoking CAN be simplified with thread safety.
From https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs#L143 and https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs#L174 I read the comments saying `Do not "simplify" - events can be removed from a different thread`.

I think these can be simplified using new "null conditional operator" actually. From https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators, these new operators are thread-safe.

huww98:
Simplify delegate invoking

fixes #561


davidfowl:
Thanks!

PR #566, opened 2017-06-15
Decouple IConnection.ts and HubConnection.ts from http

davidfowl:
- Removed TransportType from start
- Renamed Connection.ts to HttpConnection.ts

The creation of the SignalR client is a bit uglier but I plan to fix that in both C# and JS (and other clients) with some sort of builder pattern.


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -96,8 +89,8 @@ export class HubConnection {
         }
     }
 
-    start(transportType?: TransportType): Promise<void> {
```
moozzyk: How do you select transport in HubConnection now?

moozzyk: I missed the change to the ctor - you have to pass IConnection

PR #567, opened 2017-06-15
Always close websocket on server

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
```diff
@@ -27,7 +27,7 @@ public WebSocketsTransportTests(ServerFixture serverFixture)
             _serverFixture = serverFixture;
         }
 
-        [ConditionalFact(Skip = "WebsocketClient.CloseAsync never returns - investigating")]
+        [ConditionalFact]
```
moozzyk: Why did you remove changes to add logging to this test class?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -86,10 +86,17 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<byte[]>
 
                 // Wait for the application to finish sending.
                 _logger.WaitingForSend(_connectionId);
-                await sending;
 
-                // Send the server's close frame
-                await socket.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "", CancellationToken.None);
+                try
+                {
+                    await sending;
```
davidfowl: Instead of throwing and catching  a first chance exception, let's write a custom awaiter for this that does .Ignore(). I spoke to @stephentoub and there's a proposal in corefx https://github.com/dotnet/corefx/issues/20732. In the meantime, lets copy that code:

```C#
internal struct NoThrowAwaiter : ICriticalNotifyCompletion
{
    private readonly Task _task;
    public NoThrowAwaiter(Task task) { _task = task; }
    public NoThrowAwaiter GetAwaiter() => this;
    public bool IsCompleted => _task.IsCompleted;
    public void GetResult() { }
    public void OnCompleted(Action continuation) => _task.GetAwaiter().OnCompleted(continuation);
    public void UnsafeOnCompleted(Action continuation) => OnCompleted(continuation);
}
```

And add an extension method .NoThrow to task.

src/Microsoft.AspNetCore.Sockets.Http/Internal/NoThrowTaskExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Sockets.Http.Internal
+{
+    public static class NoThrowTaskExtensions
+    {
+        public static NoThrowAwaiter NoThrow(this Task task)
```
moozzyk: Is this being used anywhere?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -66,69 +67,51 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<byte[]>
                 receiving,
                 sending);
 
-            // What happened?
+            var failed = false;
+            var task = Task.CompletedTask;
             if (trigger == receiving)
             {
                 if (receiving.IsCanceled || receiving.IsFaulted)
```
davidfowl: Remove this.

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
```diff
@@ -98,48 +98,6 @@ public class WebSocketsTests
             }
         }
 
-        [Theory]
-        [InlineData(WebSocketMessageType.Text)]
-        [InlineData(WebSocketMessageType.Binary)]
-        public async Task FrameReceivedAfterServerCloseSent(WebSocketMessageType webSocketMessageType)
```
davidfowl: Why did this get removed?

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
```diff
@@ -232,5 +190,34 @@ public class WebSocketsTests
                 serverSocket.Dispose();
             }
         }
+
+        [Fact]
+        public async Task Server()
```
davidfowl: Is this really a test name?

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
BrennanConroy: It probably should have been removed when we added a timeout to `ProcessSocketAsync`

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
BrennanConroy: I added it back because if the client disconnects abnormally via `socket.Abort()` like one of our tests does, then `sending` never completes and we just timeout on the server. If that's the behavior we want I can change the test to not expect an `OperationCanceledException`

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -66,69 +67,51 @@ public WebSocketsTransport(WebSocketOptions options, IChannelConnection<byte[]>
...
                 {
-                    // The receiver faulted or cancelled. This means the socket is probably broken. Abort the socket and propagate the exception
+                    // The receiver faulted or canceled. This means the socket is probably broken. Abort the socket and propagate the exception
                     receiving.GetAwaiter().GetResult();
```
moozzyk: When we are in this branch - do we ever await the `sending` task? What if it throws? Also, how do we make sure the task (and therefore the send loop ever) finishes

davidfowl: Thats why this code should be removed.

BrennanConroy:
🆙 📅

Issue #568, opened 2017-06-15
Revisit how we close the webSockets transport on the client


moozzyk:
Currently the webSockets transport on the client uses [CloseAsync](https://github.com/aspnet/SignalR/blob/0298868c00c4254ae283cf3669eb7bdb7a341ff4/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs#L230) to close the webSocket. `CloseAsync` initiates closing the webSocket and waits until the server sends a close frame to the client. `CloseAsync` can block forever if the server never sends the close frame and there is no good way to time out this call (CancellationToken is not respected). A better way of Closing the webSocket is to use `CloseOutputAsync` which initiates closing the websocket but does not wait for the Close frame from the server. We should then try to drain the webSocket by reading until we receive a close frame or a timeout occurs. Only when we finished draining the webSocket we should `Dispose()` the webSocket.

moozzyk:
Add https://github.com/aspnet/SignalR/pull/575/files#diff-f81419b5767af008b375f60977aad33b as part of this fix.

PR #575, opened 2017-06-21
Passing Transport.All as a default C# client transport

moozzyk:
It will make the client pick WebSockets where available and
ServerSentEvents where not (e.g. Win7)

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
@@ -75,6 +75,15 @@ namespace Microsoft.AspNetCore.SignalR.Tests
             }
         }
 
+        [Fact]
+        public async Task CanStartConnectionUsingDefaultTransport()
+        {
+            var url = _serverFixture.BaseUrl + "/echo";
+            var connection = new HttpConnection(new Uri(url));
+            await connection.StartAsync().OrTimeout();
+            await connection.DisposeAsync().OrTimeout();
+        }
+
         [ConditionalTheory]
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -75,6 +75,15 @@ public EndToEndTests(ServerFixture serverFixture, ITestOutputHelper output) : ba
             }
         }
 
+        [Fact]
+        public async Task CanStartConnectionUsingDefaultTransport()
+        {
+            var url = _serverFixture.BaseUrl + "/echo";
+            var connection = new HttpConnection(new Uri(url));
```
BrennanConroy: Add a `loggerFactory` to the `HttpConnection`, also this test passes on my home machine

BrennanConroy: Did a little digging and server side we are sending the close frame in response to `CloseAsync` from the client, but the client never finishes the `CloseAsync` call, we filed an issue to change it to `CloseOutputAsync` and this test will probably continue being flaky until we make that change.

moozzyk: If it is going to be flaky we cannot merge it. We need to fix #568.

mikaelm12:
Whats the status of this fix?

moozzyk:
Blocked by https://github.com/aspnet/SignalR/issues/568

moozzyk:
The fix has already been merged and the test is part of https://github.com/aspnet/SignalR/pull/695

PR #576, opened 2017-06-21
Remove dependencies on a bunch of corefxlab things

davidfowl:
- Used Stream instead of IOutput
- Removed pipelines dependency in most places.


src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubProtocolWriteMessageExtensions.cs
```diff
@@ -3,34 +3,22 @@
...
-using System.IO.Pipelines.Text.Primitives;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
 {
     public static class HubProtocolWriteMessageExtensions
     {
-        public static async ValueTask<byte[]> WriteToArrayAsync(this IHubProtocol protocol, HubMessage message)
+        public static byte[] WriteToArrayAsync(this IHubProtocol protocol, HubMessage message)
```
moozzyk: remove `Async` suffix

BrennanConroy: Change method name, no longer async

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/TextMessageFormatter.cs
```diff
@@ -2,37 +2,42 @@
...
         {
             // Calculate the length, it's the number of characters for text messages, but number of base64 characters for binary
-            var length = payload.Length;
 
             // Write the length as a string
-            output.Append(length, TextEncoder.Utf8);
+
+            // Super inefficient...
+            var lengthString = payload.Length.ToString(CultureInfo.InvariantCulture);
+            var buffer = ArrayPool<byte>.Shared.Rent(10);
```
BrennanConroy: comment why this is 10, `// 2^(32-1) is 10 digits etc.`

Issue #546, opened 2017-06-09
Consider making .NET Client .On callbacks and Receive callbacks async


davidfowl:
If you want to write async logic in a callback then it'll be async void and events will fire in an overlapping way. We should consider async overloads to the On methods.

moozzyk:
This is already biting us [here](https://github.com/aspnet/SignalR/blob/59a8f5f3b58f3f3c7b33898bc3fbe8158a0ddf9b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs#L209).

davidfowl:
We'd also need to make the Received callback async too.

mikaelm12:
Addressed in: https://github.com/aspnet/SignalR/pull/578

PR #578, opened 2017-06-22
Changed the handler to be Func<object[], Task>

mikaelm12:
Addresses issue: https://github.com/aspnet/SignalR/issues/546

Changed the receive event on IConnection to be a `Func<object[], Task>`. 
The overload of HubConnection's On method that takes an action is now available as an extension method on HubConnection now.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
@@ -85,7 +85,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
         }
 
         // TODO: Client return values/tasks?
-        public void On(string methodName, Type[] parameterTypes, Action<object[]> handler)
+
+        public void On(string methodName, Type[] parameterTypes, Func<object[], Task> handler)
         {
             var invocationHandler = new InvocationHandler(parameterTypes, handler);
             _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
@@ -148,7 +149,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        private void OnDataReceived(byte[] data)
+        private async Task OnDataReceived(byte[] data)
         {
             if (_protocol.TryParseMessages(data, _binder, out var messages))
             {
@@ -163,7 +164,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                                 var argsList = string.Join(", ", invocation.Arguments.Select(a => a.GetType().FullName));
                                 _logger.LogTrace("Received Invocation '{invocationId}': {methodName}({args})", invocation.InvocationId, invocation.Target, argsList);
                             }
-                            DispatchInvocation(invocation, _connectionActive.Token);
+                            await DispatchInvocation(invocation, _connectionActive.Token);
                             break;
                         case CompletionMessage completion:
                             if (!TryRemoveInvocation(completion.InvocationId, out irq))
@@ -218,18 +219,18 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        private void DispatchInvocation(InvocationMessage invocation, CancellationToken cancellationToken)
+        private Task DispatchInvocation(InvocationMessage invocation, CancellationToken cancellationToken)
         {
             // Find the handler
             if (!_handlers.TryGetValue(invocation.Target, out InvocationHandler handler))
             {
                 _logger.LogWarning("Failed to find handler for '{target}' method", invocation.Target);
-                return;
+                return Task.CompletedTask;
             }
 
             // TODO: Return values
             // TODO: Dispatch to a sync context to ensure we aren't blocking this loop.
-            handler.Handler(invocation.Arguments);
+            return handler.Handler(invocation.Arguments);
         }
 
         // This async void is GROSS but we need to dispatch asynchronously because we're writing to a Channel
@@ -353,12 +354,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        private struct InvocationHandler
+        public struct InvocationHandler
         {
-            public Action<object[]> Handler { get; }
+            public Func<object[], Task> Handler { get; }
             public Type[] ParameterTypes { get; }
 
-            public InvocationHandler(Type[] parameterTypes, Action<object[]> handler)
+            public InvocationHandler(Type[] parameterTypes, Func<object[], Task> handler)
             {
                 Handler = handler;
                 ParameterTypes = parameterTypes;
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
@@ -354,7 +354,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        public struct InvocationHandler
+        private struct InvocationHandler
         {
             public Func<object[], Task> Handler { get; }
             public Type[] ParameterTypes { get; }
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
@@ -70,7 +70,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _protocol = protocol;
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
-            _connection.Received += OnDataReceived;
+            _connection.Received += OnDataReceivedAsync;
             _connection.Closed += Shutdown;
         }
 
@@ -85,7 +85,6 @@ namespace Microsoft.AspNetCore.SignalR.Client
         }
 
         // TODO: Client return values/tasks?
-
         public void On(string methodName, Type[] parameterTypes, Func<object[], Task> handler)
         {
             var invocationHandler = new InvocationHandler(parameterTypes, handler);
@@ -149,7 +148,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        private async Task OnDataReceived(byte[] data)
+        private async Task OnDataReceivedAsync(byte[] data)
         {
             if (_protocol.TryParseMessages(data, _binder, out var messages))
             {
@@ -164,7 +163,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                                 var argsList = string.Join(", ", invocation.Arguments.Select(a => a.GetType().FullName));
                                 _logger.LogTrace("Received Invocation '{invocationId}': {methodName}({args})", invocation.InvocationId, invocation.Target, argsList);
                             }
-                            await DispatchInvocation(invocation, _connectionActive.Token);
+                            await DispatchInvocationAsync(invocation, _connectionActive.Token);
                             break;
                         case CompletionMessage completion:
                             if (!TryRemoveInvocation(completion.InvocationId, out irq))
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -218,18 +219,18 @@ private void Shutdown(Exception ex = null)
             }
         }
 
-        private void DispatchInvocation(InvocationMessage invocation, CancellationToken cancellationToken)
+        private Task DispatchInvocation(InvocationMessage invocation, CancellationToken cancellationToken)
```
BrennanConroy: `DispatchInvocationAsync`

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -148,7 +149,7 @@ private void InvokeCore(string methodName, InvocationRequest irq, object[] args)
             }
         }
 
-        private void OnDataReceived(byte[] data)
+        private async Task OnDataReceived(byte[] data)
```
BrennanConroy: `OnDataReceivedAsync`

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -85,7 +85,8 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
         }
 
         // TODO: Client return values/tasks?
-        public void On(string methodName, Type[] parameterTypes, Action<object[]> handler)
+
```
BrennanConroy: newline

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionExtensions.cs
```diff
@@ -89,6 +90,11 @@ public static ReadableChannel<TResult> Stream<TResult>(this HubConnection hubCon
             return outputChannel.In;
         }
 
+        private static void On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Action<object[]> handler)
```
BrennanConroy: Shouldn't this be public

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -285,13 +285,11 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
                     if (Input.TryRead(out var buffer))
                     {
                         _logger.LogDebug("Scheduling raising Received event.");
-                        var ignore = _eventQueue.Enqueue(() =>
+                        var ignore = _eventQueue.Enqueue(async () =>
```
anurse: When the only task involved is the final thing you run, we can avoid all the boilerplate C# sets up for an async method by just rewriting this like so:

```csharp
var ignore = _eventQueue.Enqueue(() =>
{
    _logger.LogDebug("Raising Received event.");
    return Received?.Invoke(buffer);
});
```

This means you run "synchronously" up to the call to Invoke and then just invoke it and pass it's `Task` back to the caller.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionExtensions.cs
mikaelm12: Do we need to expose this overload? Shouldn't people use the generic overloads rather than pass in the parameter type array?

davidfowl: No, keep it private, there's no reason to have the sync overload, they can use the async one.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: While you're at it, change ignore to `_`.

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -406,9 +410,9 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
             await connection.StartAsync();
             channel.Output.TryWrite(Array.Empty<byte>());
             channel.Output.TryWrite(Array.Empty<byte>());
-            await allowDisposeTcs.Task.OrTimeout();
```
davidfowl: Why?

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -406,9 +410,9 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+
+            Assert.False(channel.Input.TryRead(out var message));
             await connection.DisposeAsync();
-            Assert.Equal(2, receivedInvocationCount);
```
davidfowl: Why?

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -350,7 +350,11 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
 
             var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
             var receivedInvoked = false;
-            connection.Received += (m) => receivedInvoked = true;
+            connection.Received += (m) =>
```
moozzyk: nit: remove brackets

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionExtensions.cs
```diff
@@ -89,6 +90,11 @@ public static ReadableChannel<TResult> Stream<TResult>(this HubConnection hubCon
...
+        {
+            hubConnetion.On(methodName, parameterTypes, (parameters) => { handler(parameters); return Task.CompletedTask; });
```
moozzyk: formatting

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -389,7 +393,7 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
 
             var closedTcs = new TaskCompletionSource<object>();
             var allowDisposeTcs = new TaskCompletionSource<object>();
-            int receivedInvocationCount = 0;
+            var receivedInvocationCount = 0;
 
             var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
             connection.Received +=
```
moozzyk: allowDisposeTcs and receivedInvocationCount need to be removed because they are no longer needed.

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
moozzyk: The idea behind this test now is that we would block in the user code and will check if is HttpConnection continues to consume incoming messages. If it did not it would mean the user's code blocked the SignalR thread.
@mikaelm12 - please add a comment explaining how this test is supposed to work. This is probably the most convoluted test we currently have.

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -406,9 +410,9 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
             await connection.StartAsync();
             channel.Output.TryWrite(Array.Empty<byte>());
             channel.Output.TryWrite(Array.Empty<byte>());
```
moozzyk: I think the second write needs to happen after the user's callback has been invoked.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
anurse: Oh, and use more throw expressions :trollface:

davidfowl: Hmm no middle finger reaction

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -361,21 +367,22 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-            mockHttpHandler.Protected()
-                .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.IsAny<HttpRequestMessage>(), ItExpr.IsAny<CancellationToken>())
-                .Returns<HttpRequestMessage, CancellationToken>(async (request, cancellationToken) =>
-                {
-                    await Task.Yield();
-                    return request.Method == HttpMethod.Options
-                        ? ResponseUtils.CreateResponse(HttpStatusCode.OK, ResponseUtils.CreateNegotiationResponse())
-                        : ResponseUtils.CreateResponse(HttpStatusCode.OK);
-                });
+                var mockHttpHandler = new Mock<HttpMessageHandler>();
```
moozzyk: This formatting is off

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -593,7 +601,12 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
             try
             {
                 var receiveTcs = new TaskCompletionSource<string>();
-                connection.Received += (data) => receiveTcs.TrySetResult(Encoding.UTF8.GetString(data));
+                connection.Received += (data) =>
```
moozzyk: nit: (because you touched it) - remove parenthesis

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -284,13 +284,11 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
...
                         _logger.LogDebug("Scheduling raising Received event.");
-                        var ignore = _eventQueue.Enqueue(() =>
+                        var _ = _eventQueue.Enqueue(() =>
                         {
                             _logger.LogDebug("Raising Received event.");
 
-                            Received?.Invoke(buffer);
-
-                            return Task.CompletedTask;
+                            return Received?.Invoke(buffer);
```
davidfowl: nit: What happens when Received is null and we return a null task in Enqueue?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -284,11 +284,15 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
                     if (Input.TryRead(out var buffer))
                     {
                         _logger.LogDebug("Scheduling raising Received event.");
-                        var ignore = _eventQueue.Enqueue(() =>
+                        var _ = _eventQueue.Enqueue(() =>
```
BrennanConroy: nit: remove var

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Added a null check for the receive func.

mikaelm12:
Updated 😄

mikaelm12:
Also, in the TaskQueue if the func returns null, should we throw or just return a completed task? 
Talking about this line specifically https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/TaskQueue.cs#L52

moozzyk:
I don't think we ever Enqueue a func that does not return an actual task.

BrennanConroy:
We Enqueue a user callback so who knows what they do

mikaelm12:
Yeah, this ^

moozzyk:
(Obviously I looked at dev) We should return a CompletedTask if they return null.

PR #579, opened 2017-06-23
Use Channel<byte[]> as the abstraction

davidfowl:
ChannelConnection<T> is pubternal (but should really be internal).


src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -9,13 +9,14 @@
 using System.Collections.Generic;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using System.Threading.Tasks.Channels;
```
BrennanConroy: sort

PR #587, opened 2017-06-23
Adding MsgPack HubProtocol

moozzyk:
MsgPackHubProtocol - not plugged in yet.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
@@ -1,0 +1,233 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using MsgPack;
+using MsgPack.Serialization;
+
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
+{
+    public class MessagePackHubProtocol : IHubProtocol
+    {
+        private const int InvocationMessageType = 1;
+        private const int StreamItemMessageType = 2;
+        private const int CompletionMessageType = 3;
+
+        public bool TryParseMessages(ReadOnlySpan<byte> input, IInvocationBinder binder, out IList<HubMessage> messages)
+        {
+            messages = new List<HubMessage>();
+
+            using (var memoryStream = new MemoryStream(input.ToArray()))
+            {
+                messages.Add(ParseMessage(memoryStream, binder));
+            }
+
+            return messages.Count > 0;
+        }
+
+        private static HubMessage ParseMessage(Stream input, IInvocationBinder binder)
+        {
+            var unpacker = Unpacker.Create(input);
+            if (!unpacker.ReadInt32(out var messageType))
+            {
+                throw new FormatException("Message type is missing.");
+            }
+
+            switch (messageType)
+            {
+                case InvocationMessageType:
+                    return CreateInvocationMessage(unpacker, binder);
+                case StreamItemMessageType:
+                    return CreateStreamItemMessage(unpacker, binder);
+                case CompletionMessageType:
+                    return CreateCompletionMessage(unpacker, binder);
+                default:
+                    throw new FormatException($"Invalid message type: {messageType}.");
+            }
+        }
+
+        private static InvocationMessage CreateInvocationMessage(Unpacker unpacker, IInvocationBinder binder)
+        {
+            var invocationId = ReadInvocationId(unpacker);
+            var nonBlocking = ReadBoolean(unpacker, "nonBlocking");
+            var target = ReadString(unpacker, "target");
+            var argumentCount = ReadArrayLength(unpacker, "arguments");
+            var parameterTypes = binder.GetParameterTypes(target);
+
+            if (parameterTypes.Length != argumentCount)
+            {
+                throw new FormatException(
+                    $"Target method expects {parameterTypes.Length} arguments(s) but invocation has {argumentCount} argument(s).");
+            }
+
+            var arguments = new object[argumentCount];
+            for (var i = 0; i < argumentCount; i++)
+            {
+                arguments[i] = DeserializeObject(unpacker, parameterTypes[i], "argument");
+            }
+
+            return new InvocationMessage(invocationId, nonBlocking, target, arguments);
+        }
+
+        private static StreamItemMessage CreateStreamItemMessage(Unpacker unpacker, IInvocationBinder binder)
+        {
+            var invocationId = ReadInvocationId(unpacker);
+            var itemType = binder.GetReturnType(invocationId);
+            var value = DeserializeObject(unpacker, itemType, "item");
+            return new StreamItemMessage(invocationId, value);
+        }
+
+        private static CompletionMessage CreateCompletionMessage(Unpacker unpacker, IInvocationBinder binder)
+        {
+            var invocationId = ReadInvocationId(unpacker);
+            var error = ReadString(unpacker, "error");
+
+            var hasResult = false;
+            object result = null;
+            if (error == null)
+            {
+                hasResult = ReadBoolean(unpacker, "hasResult");
+                if (hasResult)
+                {
+                    var itemType = binder.GetReturnType(invocationId);
+                    result = DeserializeObject(unpacker, itemType, "argument");
+                }
+            }
+
+            return new CompletionMessage(invocationId, error, result, hasResult);
+        }
+
+        // TODO: when to return false?
+        public bool TryWriteMessage(HubMessage message, Stream output)
+        {
+            var packer = Packer.Create(output);
+            switch (message)
+            {
+                case InvocationMessage invocationMessage:
+                    WriteInvocationMessage(invocationMessage, packer, output);
+                    break;
+                case StreamItemMessage streamItemMessage:
+                    WriteStremingItemMessage(streamItemMessage, packer, output);
+                    break;
+                case CompletionMessage completionMessage:
+                    WriteCompletionMessage(completionMessage, packer, output);
+                    break;
+                default:
+                    throw new FormatException($"Unexpected message type: {message.GetType().Name}");
+            }
+
+            return true;
+        }
+
+        private static void WriteInvocationMessage(InvocationMessage invocationMessage, Packer packer, Stream output)
+        {
+            packer.Pack(InvocationMessageType);
+            packer.PackString(invocationMessage.InvocationId);
+            packer.Pack(invocationMessage.NonBlocking);
+            packer.PackString(invocationMessage.Target);
+            packer.PackObject(invocationMessage.Arguments);
+        }
+
+        private void WriteStremingItemMessage(StreamItemMessage streamItemMessage, Packer packer, Stream output)
+        {
+            packer.Pack(StreamItemMessageType);
+            packer.PackString(streamItemMessage.InvocationId);
+            packer.PackObject(streamItemMessage.Item);
+        }
+
+        private void WriteCompletionMessage(CompletionMessage completionMessage, Packer packer, Stream output)
+        {
+            packer.Pack(CompletionMessageType);
+            packer.PackString(completionMessage.InvocationId);
+            packer.PackString(completionMessage.Error);
+            if (completionMessage.Error == null)
+            {
+                packer.Pack(completionMessage.HasResult);
+                if (completionMessage.HasResult)
+                {
+                    packer.PackObject(completionMessage.Result);
+                }
+            }
+        }
+
+        private static string ReadInvocationId(Unpacker unpacker)
+        {
+            return ReadString(unpacker, "invocationId");
+        }
+
+        private static string ReadString(Unpacker unpacker, string field)
+        {
+            Exception msgPackException = null;
+            try
+            {
+                if (unpacker.ReadString(out var value))
+                {
+                    return value;
+                }
+            }
+            catch (Exception e)
+            {
+                msgPackException = e;
+            }
+
+            throw new FormatException($"Reading '{field}' as String failed.", msgPackException);
+        }
+
+        private static bool ReadBoolean(Unpacker unpacker, string field)
+        {
+            Exception msgPackException = null;
+            try
+            {
+                if (unpacker.ReadBoolean(out var value))
+                {
+                    return value;
+                }
+            }
+            catch (Exception e)
+            {
+                msgPackException = e;
+            }
+
+            throw new FormatException($"Reading '{field}' as Boolean failed.", msgPackException);
+        }
+
+        private static long ReadArrayLength(Unpacker unpacker, string field)
+        {
+            Exception msgPackException = null;
+            try
+            {
+                if (unpacker.ReadArrayLength(out var value))
+                {
+                    return value;
+                }
+            }
+            catch (Exception e)
+            {
+                msgPackException = e;
+            }
+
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.SignalR.Common/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -16,6 +16,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetVersion)" />
     <PackageReference Include="System.Buffers.Primitives" Version="$(CoreFxLabsVersion)" />
     <PackageReference Include="System.Binary" Version="$(CoreFxLabsVersion)" />
+    <PackageReference Include="MsgPack.Cli" Version="0.9.0-beta2" />
```
moozzyk: @Eilon - can we get this dependency approved?

davidfowl: @moozzyk Can you put this in dependencies.props?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -0,0 +1,237 @@
...
+                {
+                    return value;
+                }
+            }
+            catch (Exception e)
+            {
+                msgPackException = e;
+            }
+
+            throw new FormatException($"Reading '{field}' as String failed.", msgPackException);
```
davidfowl: Why not throw directly from the catch?

davidfowl: Also, is there a specific exception type?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -0,0 +1,237 @@
...
+{
+    public class MessagePackHubProtocol : IHubProtocol
+    {
+        private const int InvocationMessageType = 1;
+        private const int StreamItemMessageType = 2;
+        private const int CompletionMessageType = 3;
+
+        public bool TryParseMessages(ReadOnlySpan<byte> input, IInvocationBinder binder, out IList<HubMessage> messages)
+        {
+            messages = new List<HubMessage>();
```
davidfowl: All it needs is the BinaryMessageParser created here to parse length prefixed messages.

davidfowl:
Are you going to update the spec?

moozzyk:
Yes, I will update the spec separately.

src/Microsoft.AspNetCore.SignalR.Common/Microsoft.AspNetCore.SignalR.Common.csproj
moozzyk: I will.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
moozzyk: yes

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
moozzyk: ReadString may return false if we are at the end of the stream and there is nothing to read. We should throw in this case because we require the value to be there. ReadString also throws if the string is corrupted or the data to read is not string. We don't want to throw this exception because it is a proprietary exception (MessagePackException). Rather it will be an inner exception.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/MessagePackHubProtocolTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        };
+
+        [Theory]
+        [MemberData(nameof(TestMessages))]
+        public void CanRoundTripInvocationMessage(HubMessage hubMessage)
+        {
+            using (var memoryStream = new MemoryStream())
+            {
+                _hubProtocol.TryWriteMessage(hubMessage, memoryStream);
+                memoryStream.Position = 0;
```
davidfowl: This isn't required since you do ToArray()

neuecc:
Is the msgpack data is valid?
I think it packed 5 separated message block.

```csharp
packer.Pack(InvocationMessageType);
packer.PackString(invocationMessage.InvocationId);
packer.Pack(invocationMessage.NonBlocking);
packer.PackString(invocationMessage.Target);
packer.PackObject(invocationMessage.Arguments);
```

The debugging performance of MessagePack is degraded.

```csharp
// dump result only returns '1' (InvocationMessageType)

// MsgPack-Cli
Console.WriteLine(MsgPack.Serialization.MessagePackSerializer.UnpackMessagePackObject(stream).ToString());

// MessagePack for C#
Console.WriteLine(MessagePack.MessagePackSerializer.ToJson(stream.ToArray()));
```

Message should use array header.

```csharp
packer.PackArrayHeader(5);
packer.Pack(InvocationMessageType);
packer.PackString(invocationMessage.InvocationId);
packer.Pack(invocationMessage.NonBlocking);
packer.PackString(invocationMessage.Target);
packer.PackObject(invocationMessage.Arguments);
```

It can dump, can read from deserializer and it's just 1 byte.

neuecc:
Although it may be unnecessary, this is an example of WriteMessage when described with MessagePack for C#.

```csharp
MessagePackBinary.WriteArrayHeader(output, 5);
MessagePackBinary.WriteInt32(output, InvocationMessageType);
MessagePackBinary.WriteString(output, invocationMessage.InvocationId);
MessagePackBinary.WriteBoolean(output, invocationMessage.NonBlocking);
MessagePackBinary.WriteString(output, invocationMessage.Target);

MessagePackBinary.WriteArrayHeader(output, invocationMessage.Arguments.Length);
for (int i = 0; i < invocationMessage.Arguments.Length; i++)
{
    if (invocationMessage.Arguments[i] == null)
    {
        MessagePackBinary.WriteNil(output);
    }
    else
    {
        // when WriteMessage can get IInvocationBinder.GetParameters?
        // GetType from argument directly(concrete type) is not good way(lost interface declaration)
        MessagePack.MessagePackSerializer.NonGeneric.Serialize(invocationMessage.Arguments[i].GetType(), output, invocationMessage.Arguments[i]);
    }
}
```

moozzyk:
@neuecc - we prefix the message data with the message length and we write and read/interpret the values so I did not make the message an array. Note that you are not really able to deserialize these messages using high level deserializer since each of these messages may contain objects whose types can only be known after we start reading the invocation id.

I started with MessagePack for C# but was blocked by https://github.com/neuecc/MessagePack-CSharp/issues/63 and eventually went with the MsgPack-Cli. I kept MsgPack usage contained so that it is easy to replace. It should also make it easy to compare the two MessagePack libraries in our scenarios.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/CustomObject.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.SignalR.Common.Tests.Internal.Protocol
+{
+    public class CustomObject //: IEquatable<CustomObject>
+    {
+        // Not intended to be a full set of things, just a smattering of sample serializations
```
mikaelm12: +1 for use of "smattering"

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            using (var memoryStream = new MemoryStream(input.ToArray()))
+            {
+                messages.Add(ParseMessage(memoryStream, binder));
```
mikaelm12: What happens if this Parse fails?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -0,0 +1,237 @@
...
+                    var itemType = binder.GetReturnType(invocationId);
+                    result = DeserializeObject(unpacker, itemType, "argument");
+                }
+            }
+
+            return new CompletionMessage(invocationId, error, result, hasResult);
+        }
+
+        // TODO: when to return false?
+        public bool TryWriteMessage(HubMessage message, Stream output)
```
mikaelm12: Do we need TryWriteMessage method wrapping WriteMessage? You throw in the failure case

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/CustomObject.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.SignalR.Common.Tests.Internal.Protocol
+{
+    public class CustomObject //: IEquatable<CustomObject>
```
mikaelm12: Remove comment?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/MessagePackHubProtocolTests.cs
```diff
@@ -0,0 +1,111 @@
...
+                _hubProtocol.TryWriteMessage(hubMessage, memoryStream);
+                _hubProtocol.TryParseMessages(
+                    new ReadOnlySpan<byte>(memoryStream.ToArray()), new TestBinder(hubMessage), out var messages);
+
+                Assert.Equal(1, messages.Count);
+                Assert.Equal(hubMessage, messages[0], TestHubMessageEqualityComparer.Instance);
+            }
+        }
+
+
```
mikaelm12: nit: Remove extra newline

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+            if (leftEnumerable == null || rightEnumerable == null)
+            {
+                return false;
+            }
+
+            var leftEnumerator = leftEnumerable.GetEnumerator();
+            var rightEnumerator = rightEnumerable.GetEnumerator();
+            var leftMoved = leftEnumerator.MoveNext();
+            var rightMoved = rightEnumerator.MoveNext();
+            for (; leftMoved && rightMoved; leftMoved = leftEnumerator.MoveNext(), rightMoved = rightEnumerator.MoveNext())
```
mikaelm12: ?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                if (!Equals(leftEnumerator.Current, rightEnumerator.Current))
+                {
+                    return false;
+                }
+            }
+
+            return !leftMoved && !rightMoved;
```
mikaelm12: nit: You can simplify this from !A && !B to !(A && B)

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
mikaelm12: Oh, parse message just throws

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
moozzyk: no, as per De Morgan law it would be !(A || B). I had it but I reverted since I find it cleaner.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
moozzyk: move iterators in tandem

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
mikaelm12: 😅

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
moozzyk: They could be merged indeed. I am planning to review the Try* methods we have since they oftentimes never return false (anymore).

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/CustomObject.cs
moozzyk: Credits should go to @anurse - I just moved this code to a separate file

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/CustomObject.cs
moozzyk: Need to uncomment actually.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/CustomObject.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.SignalR.Common.Tests.Internal.Protocol
+{
+    public class CustomObject : IEquatable<CustomObject>
```
anurse: Now that this is separate from the test file, can we call it `TestCustomObject` or something?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
moozzyk: The inner exception will tell the details - it is MsgPack specific. If we were at the end of the stream and could not read there won't be any exception. This basically should not happen unless the other side sent a broken message.

I might just add comment in the code.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/CustomObject.cs
moozzyk: Sure.

PR #590, opened 2017-06-25
WIP: Use the client on the server

davidfowl:
- Move all dispatching logic into the HubConnection and use it on
the server side.
- Adding callbacks per connection for every set of hub methods isn't
the most efficient thing we could do but the API churn is low as a result.
- Formalized blocking, non blocking and streaming result behavior. The server
gets client results for free. It also means that crazy peer to peer scenarios can
use the HubConnection API.
- Added support for Client side hubs. Streaming works in both directions for free.

PS: This isn't finished yet but I wanted people to start taking a look at what code sharing would look like.


src/Microsoft.AspNetCore.SignalR.Common/IHubProxy.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.AspNetCore.SignalR
+{
+    public interface IHubProxy
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="methodName"></param>
+        /// <param name="args"></param>
+        Task Invoke(string methodName, params object[] args);
```
davidfowl: This is a non blocking invocation. The naming is bad, the Task result tells you when it was sent but doesn't tell you anything else. I started off with void but the name is just confusing.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -92,13 +93,28 @@ public override Task InvokeAllAsync(string methodName, object[] args)
             return PublishAsync(typeof(THub).FullName, message);
         }
 
-        public override Task InvokeConnectionAsync(string connectionId, string methodName, object[] args)
+        public override IHubClientProxy GetConnectionProxy(string connectionId)
         {
-            var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
-
-            return PublishAsync(typeof(THub).FullName + "." + connectionId, message);
+            throw new NotImplementedException();
```
davidfowl: This is still broken.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -81,13 +85,9 @@ private Task InvokeAllWhere(string methodName, object[] args, Func<ConnectionCon
...
 
-        public override Task InvokeConnectionAsync(string connectionId, string methodName, object[] args)
+        public override IHubClientProxy GetConnectionProxy(string connectionId)
         {
-            var connection = _connections[connectionId];
-
-            var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
-
-            return WriteAsync(connection, message);
+            return _connections[connectionId].Metadata.Get<IHubClientProxy>(typeof(HubConnection));
```
davidfowl: This is where it gets crazy. We just return the HubConnection itself as the proxy since it has all of the logic for writing outgoing messages.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -415,41 +240,6 @@ private void DiscoverHubMethods()
             }
         }
 
-        private async Task<bool> IsHubMethodAuthorized(IServiceProvider provider, ClaimsPrincipal principal, IList<IAuthorizeData> policies)
```
davidfowl: Will put this back.

samples/ChatSample/Hubs/Chat.cs
```diff
@@ -16,24 +16,24 @@ public Chat(IUserTracker<Chat> userTracker)
 
         public override async Task OnConnectedAsync()
         {
-            await Clients.Client(Context.ConnectionId).InvokeAsync("SetUsersOnline", await GetUsersOnline());
```
anurse: I ThoughtAsync we LikedAsync the Async suffix now?

src/Microsoft.AspNetCore.SignalR.Common/IHubProxy.cs
anurse: Not sure I like this. It throws me off completely, as seen by my earlier comments about the naming. The name is just totally wrong. I'm not yet convinced that changing the name of this would be any worse than the other breaking changes we're making (heck, it's probably easier to resolve than any of those, just find/replace) and would mean we could make it something clearer like `SendAsync`, which I would expect to resolve when a message is **sent** vs. `InvokeAsync` which to me implies that it resolves when the underlying method finishes.

src/Microsoft.AspNetCore.SignalR.Common/IHubProxy.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="methodName"></param>
+        /// <param name="returnType"></param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="args"></param>
+        /// <returns></returns>
+        Task<object> InvokeAsync(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args);
```
anurse: Why is this one `InvokeAsync`? It waits for the return?

src/Microsoft.AspNetCore.SignalR.Common/IHubProxy.cs
anurse: And if we're going to fix this naming issue, now's our best opportunity to do so. It's not like we can't revert it after getting feedback.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
anurse: That doesn't seem so crazy to me. It's what HubConnection represents: A handle to a hub on the other peer.

anurse: Do you need the two types though? If `HubConnection` implements `IHubClientProxy`, then shouldn't `Get<HubConnection>` be sufficient? Or am I assuming we have a `Get<T>() => (T)Get(typeof(T))` in the `Metadata` object when we don't? :)

src/Microsoft.AspNetCore.SignalR/HubLifetimeManager.cs
```diff
@@ -14,7 +18,15 @@ public abstract class HubLifetimeManager<THub>
...
-        public abstract Task InvokeConnectionAsync(string connectionId, string methodName, object[] args);
+        // public abstract IHubProxy GetAllProxy();
+
+        public abstract IHubClientProxy GetConnectionProxy(string connectionId);
+
+        // public abstract IHubProxy GetGroupProxy(string groupName);
+
+        // public abstract IHubProxy GetUserProxy(string userId);
+
+        // public abstract IGroupManager GetGroupManager();
```
anurse: Composition is good :) 👍

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -288,7 +288,8 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
                         {
                             _logger.LogDebug("Raising Received event.");
 
-                            Received?.Invoke(buffer);
+                            // TODO: Return this
```
anurse: Why TODO? Why not just return it?

src/Microsoft.AspNetCore.SignalR.Common/IHubProxy.cs
davidfowl: Yea, the name was never going to stick as Invoke when there's an InvokeAsync.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
davidfowl: Both work.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: This PR stomps on @mikaelm12 's PR which does this. Returning this requires more test changes.

samples/ClientSample/HubSample.cs
```diff
@@ -37,6 +41,9 @@ internal static void Register(CommandLineApplication app)
             var connection = new HubConnection(httpConnection, loggerFactory);
             try
             {
+                // Wire up client side Chat
+                connection.Bind(new Chat());
```
davidfowl: Client side hubs

samples/ClientSample/HubSample.cs
```diff
@@ -76,4 +91,35 @@ internal static void Register(CommandLineApplication app)
...
+        {
+            Console.WriteLine(message);
+        }
+
+        public int Add(int a, int b)
+        {
+            return a + b;
+        }
+
+        public ReadableChannel<int> ChannelCounter(int count, int delay)
```
davidfowl: Stream results to the server

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
anurse: Right, so then how about skipping the `typeof` and just using `Get<HubConnection>`?

davidfowl:
Closing this out for now. I'll cherry pick some of the changes.

Issue #580, opened 2017-06-23
Closing websocket in the browser causing failure on server


davidfowl:
Run the hubs sample in SocketSample and click connect then disconnect. This exception is thrown:

```
fail: Microsoft.AspNetCore.Sockets.Internal.Transports.WebSocketsTransport[11]
      06/23/2017 03:16:49: Connection Id 8b302138-d905-4238-a966-5f5cb37331f8: Error writing frame.
System.Net.WebSockets.WebSocketException (0x80004005): TODO, RESX: net_WebSockets_InvalidState; (CloseSent,Open, CloseReceived)
   at System.Net.WebSockets.WebSocketValidate.ThrowIfInvalidState(WebSocketState currentState, Boolean isDisposed, WebSocketState[] validStates)
   at System.Net.WebSockets.ManagedWebSocket.SendAsync(ArraySegment`1 buffer, WebSocketMessageType messageType, Boolean endOfMessage, CancellationToken cancellationToken)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Sockets.Internal.Transports.WebSocketsTransport.<StartSending>d__8.MoveNext() in C:\Users\dfowler\Documents\GitHub\SignalR\src\Microsoft.AspNetCore.Sockets.Http\Internal\Transports\WebSocketsTransport.cs:line 180
```

BrennanConroy:
That doesn't happen on the dev branch, but I tried your PR #579 and it happens there

davidfowl:
Ahhhh I though I cleaned. I guess not.

BrennanConroy:
Oh wait, its just a flaky error. I managed to repro it on dev

davidfowl:
!!!

davidfowl:
The fact that none of our tests catch that means we have a gap as well.

muratg:
Assigning to @BrennanConroy for investigation.  Possible fix is catching the exception and ignoring it (if WebSocket is closed.)

BrennanConroy:
https://github.com/aspnet/SignalR/commit/e36da3007214db1b3c085972cd88be23c7a2a97e

PR #593, opened 2017-06-26
Ignore failed send after close (WebSockets)

BrennanConroy:
#580

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
@@ -181,7 +181,9 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Transports
                             await ws.SendAsync(new ArraySegment<byte>(buffer), _options.WebSocketMessageType, endOfMessage: true, cancellationToken: CancellationToken.None);
                         }
                         catch (WebSocketException socketException)
-                        when (socketException.WebSocketErrorCode == WebSocketError.InvalidState)
+                        when (ws.State == WebSocketState.Aborted ||
+                              ws.State == WebSocketState.Closed ||
+                              ws.State == WebSocketState.CloseSent)
                         {
                             // this can happen when we send the CloseFrame to the client and try to write afterwards
                             _logger.SendFailed(_connectionId, socketException);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -180,6 +180,13 @@ public WebSocketsTransport(WebSocketOptions options, Channel<byte[]> application
 
                             await ws.SendAsync(new ArraySegment<byte>(buffer), _options.WebSocketMessageType, endOfMessage: true, cancellationToken: CancellationToken.None);
                         }
+                        catch (WebSocketException socketException)
+                        when (socketException.WebSocketErrorCode == WebSocketError.InvalidState)
```
moozzyk: shouldn't you also check that the transport/channel was actually closed to avoid swallowing exceptions in case where they should be thrown?

davidfowl: I don't like this fix at all. We should optimistically avoid throwing an exception here even though there is a race. Also why don't we look at the actual state instead of looking at the websocket error code?

BrennanConroy: We never throw here for any exception, we just log it and exit. I wasn't sure what states were relevant
```
None = 0,
Connecting = 1,
Open = 2,
CloseSent = 3,
CloseReceived = 4,
Closed = 5,
Aborted = 6
```

I think `Closed Aborted CloseSent` would be the ones to cause the offending exception.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -178,7 +178,17 @@ public WebSocketsTransport(WebSocketOptions options, Channel<byte[]> application
...
                             _logger.SendPayload(_connectionId, buffer.Length);
 
-                            await ws.SendAsync(new ArraySegment<byte>(buffer), _options.WebSocketMessageType, endOfMessage: true, cancellationToken: CancellationToken.None);
+                            if (WebSocketCanSend(ws))
+                            {
+                                await ws.SendAsync(new ArraySegment<byte>(buffer), _options.WebSocketMessageType, endOfMessage: true, cancellationToken: CancellationToken.None);
+                            }
+                        }
+                        catch (WebSocketException socketException)
+                        when (!WebSocketCanSend(ws))
```
davidfowl: Fix this formatting, put the when on the same line as the catch.

PR #594, opened 2017-06-26
Enabling length prefixing, parsing multiple messages

moozzyk:


moozzyk:
71949129eab690cb45f63911bccd70d2ad656634

PR #597, opened 2017-06-27
Pin Roslyn compiler version

BrennanConroy:


build/dependencies.props
```diff
@@ -15,6 +15,7 @@
     <TestSdkVersion>15.3.0-*</TestSdkVersion>
     <XunitVersion>2.3.0-beta2-*</XunitVersion>
     <RxVersion>3.1.1</RxVersion>
+    <RoslynVersion>2.6.0-rdonly-ref-61826-08</RoslynVersion>
```
moozzyk: How do we not forget about reverting this when it's time (and when we know when this time is)?

BrennanConroy: I'll file an issue, not sure how we'll know though

mikeharder:
@BrennanConroy : Should we delete dev-gate branch?

PR #598, opened 2017-06-27
MsgPack end-to-end

moozzyk:
Plugging in MsgPack protocol
Adding negotiation message to hub protocol

The following comes next/separately:
- TS client
- Version in negotiate
- Updating the protocol spec

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -77,6 +78,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
         public async Task StartAsync()
         {
             await _connection.StartAsync();
+
+            using (var memoryStream = new MemoryStream())
+            {
+                NegotiationProtocol.TryWriteProtocolNegotiationMessage(new NegotiationMessage(_protocol.Name), memoryStream);
+                await _connection.SendAsync(memoryStream.ToArray(), _connectionActive.Token);
+            }
         }
 
         public async Task DisposeAsync()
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -77,6 +78,12 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
         public async Task StartAsync()
         {
             await _connection.StartAsync();
+
+            using (var memoryStream = new MemoryStream())
+            {
+                NegotiationProtocol.TryWriteProtocolNegotiationMessage(new NegotiationMessage(_protocol.GetType().Name), memoryStream);
+                await _connection.SendAsync(memoryStream.ToArray(), CancellationToken.None);
```
moozzyk: Failures here will most likely need to be handled

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -44,6 +44,8 @@ public JsonHubProtocol(JsonSerializer payloadSerializer)
             _payloadSerializer = payloadSerializer;
         }
 
+        public string Name { get => "Json"; }
```
davidfowl: `Name => "json"`

src/Microsoft.AspNetCore.SignalR/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -9,10 +10,17 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 {
     public class DefaultHubProtocolResolver : IHubProtocolResolver
     {
-        public IHubProtocol GetProtocol(ConnectionContext connection)
+        public IHubProtocol GetProtocol(string protocolName, ConnectionContext connection)
         {
-            // TODO: Allow customization of this serializer!
-            return new JsonHubProtocol(new JsonSerializer());
+            switch (protocolName)
```
davidfowl: `ToLowerInvariant()` this shouldn't be case sensitive

moozzyk:
f8d91b54d4b677b619d716f5003f2ab98b614829

Issue #589, opened 2017-06-24
We need an API to make a non blocking invocation


davidfowl:
Today we infer that client to server calls are blocking invocations (they need a response from the caller) and server to client calls are non blocking. Instead of this being implicit, we should have an explicit API for describing this.

PR #603, opened 2017-06-28
Support non blocking send on client

davidfowl:
https://github.com/aspnet/SignalR/issues/589

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
@@ -101,29 +101,46 @@ namespace Microsoft.AspNetCore.SignalR.Client
         public ReadableChannel<object> Stream(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
         {
             var irq = InvocationRequest.Stream(cancellationToken, returnType, GetNextId(), _loggerFactory, out var channel);
-            InvokeCore(methodName, irq, args);
+            _ = InvokeCore(methodName, irq, args, nonBlocking: false);
             return channel;
         }
 
-        public Task<object> Invoke(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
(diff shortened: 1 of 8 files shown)


moozzyk:
You changed only client. Shouldn't the server be changed as well to not send a response if this is a nonBlocking request? No one is waiting for this response and you will likely receive a [warning](https://github.com/aspnet/SignalR/blob/f8d91b54d4b677b619d716f5003f2ab98b614829/src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs#L178) for each such response.

davidfowl:
Why would that warning ever happen? The sever never sends a message back for non blocking invocations and the client never sends a response back to the sever for a non blocking invocation.

anurse:
I don't think the server honors `nonBlocking` right now. It'll send a completion anyway, which is a protocol error (warning, in this case)

moozzyk:
What @anurse said.

davidfowl:
I don't understand.

davidfowl:
I do now 😄

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -101,29 +101,46 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
...
+            return InvokeCore(methodName, irq, args, nonBlocking: true);
+        }
+
+        private Task InvokeCore(string methodName, InvocationRequest irq, object[] args, bool nonBlocking)
         {
             ThrowIfConnectionTerminated();
-            _logger.LogTrace("Preparing invocation of '{target}', with return type '{returnType}' and {argumentCount} args", methodName, irq.ResultType.AssemblyQualifiedName, args.Length);
+            if (nonBlocking)
+            {
+                _logger.LogTrace("Preparing invocation of '{target}' and {argumentCount} args", methodName, irq.ResultType.AssemblyQualifiedName, args.Length);
```
moozzyk: super nit: Preparing non-blocking...

Issue #552, opened 2017-06-12
Auth


moozzyk:
- [x] - Auth
  - [x] - Authorize Attribute on EndPoints
  - [x] - Authorize Attribute on Hub
  - [x] - Authorize Attribute on Hub methods

muratg:
Hub and hub method auth done.

Rest of the work is punted to RTM: https://github.com/aspnet/SignalR/issues/583

PR #606, opened 2017-06-28
Support Authorize attribute on EndPoints

BrennanConroy:
https://github.com/aspnet/SignalR/issues/552

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/SocketRouteBuilder.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/SocketRouteBuilder.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Reflection;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Routing;
 
 namespace Microsoft.AspNetCore.Sockets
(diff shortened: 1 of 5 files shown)


davidfowl:
Why is appveyor failing?

BrennanConroy:
Looks like a flaky test? :(

davidfowl:
Can you file an issue for it and re-run appveyor

BrennanConroy:
https://github.com/aspnet/SignalR/issues/608

src/Microsoft.AspNetCore.Sockets.Http/SocketRouteBuilder.cs
```diff
@@ -30,7 +32,20 @@ public void MapSocket(string path, HttpSocketOptions options, Action<ISocketBuil
 
         public void MapEndPoint<TEndPoint>(string path) where TEndPoint : EndPoint
         {
-            MapSocket(path, builder =>
+            MapEndPoint<TEndPoint>(path, socketOptions: null);
+        }
+
+        public void MapEndPoint<TEndPoint>(string path, Action<HttpSocketOptions> socketOptions) where TEndPoint : EndPoint
+        {
+            var authorizeAttributes = typeof(TEndPoint).GetCustomAttributes<AuthorizeAttribute>(inherit: true);
```
moozzyk: Why even do that if socketOptions are null?

src/Microsoft.AspNetCore.Sockets/SocketBuilderExtensions.cs
```diff
@@ -12,12 +9,9 @@ public static class SocketBuilderExtensions
     {
         public static ISocketBuilder UseEndPoint<TEndPoint>(this ISocketBuilder socketBuilder) where TEndPoint : EndPoint
         {
+            var endpoint = socketBuilder.ApplicationServices.GetRequiredService<TEndPoint>();
```
moozzyk: Was there any specific reason you changed it? Asking because you extended scope of endpoint which is technically not needed,

src/Microsoft.AspNetCore.Sockets.Http/SocketRouteBuilder.cs
BrennanConroy: What? `socketOptions` is just a way for the user to add there own options to the one we make, so we still add our attribute info

src/Microsoft.AspNetCore.Sockets/SocketBuilderExtensions.cs
BrennanConroy: I was making it consistent with the `UseHub` method

src/Microsoft.AspNetCore.Sockets.Http/SocketRouteBuilder.cs
moozzyk: ahhh. I read the code incorrectly.

PR #607, opened 2017-06-28
Moving typings to the TS client folder

moozzyk:


davidfowl:
I assume this does something useful?

moozzyk:
Clean-up. The client was moved to the client-ts folder but the typings folder which is only used by the client-ts was missed and is still in the repo root.

Issue #588, opened 2017-06-24
.NET Core Signalr C# Client - Make overload for specifying an HttpClientHandler more convenient


pwen090:
Hello I have a C# asp.net core (full framework) SignalR server; using SSL and requiring TLS 1.2. I have an existing C# (full framework) console application using SignalR client. This works fine without any issues. I have a new client in C# (.net core) but the same code is returning an exception when I call my hubconnection.Start() method.

`A security error occurred
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at System.Net.Http.WinHttpHandler.<StartRequest>d__105.MoveNext()
`
The "only" difference between the code for the .NET Core and .NET Full client is in the .NET Full client we have:

`ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
            ServicePointManager.ServerCertificateValidationCallback += (sender, certx, chain, sslPolicyErrors) => true;`

This is obviously a big difference and most likely the cause for the .NET Core client failing given our server enforces TLS 1.2 and also returns a cert that might look invalid to default cert validation methods. 

What I am unclear on is how with the .NET Core SignalR Client do we perform the same configuration like we have done in the past with ServicePointManager? thank you

pwen090:
Using version 2.2.2 off standard nuget feed for Microsoft.AspNet.SignalR.Client. Is there a newer version I should try to give you better feedback? 

I saw those constructor options and gave it a good try but was not sure what/how it should really be setup. Do you have any more information on that so i can try to maybe get something working to share with people? Even if it is a high level note of what should be chained together. Is the top most piece the actual hubconnection.Start(httpclient or clientransport) and then what from there? thank you for any advice.

I had to do some tweaks for autorest and ssl under .net core but it was more a matter of changing my existing WebRequestHandler to be a HttpClientHandler. And since HttpClientHandler lets you specify SslProtocols and ServerCertificateCustomValidationCallback there was nothing else to do to overcome the previous usage of ServicePointManager.

davidfowl:
Sounds like you're using the non ASP.NET Core version of SignalR. This isn't the right repository for that project. You should file this issue at github.com/SignalR/SignalR

[...]

PR #609, opened 2017-06-28
Added a Uri,HttpMessageHandler ctor for HttpConnection

mikaelm12:
Addresses: https://github.com/aspnet/SignalR/issues/588
The only bad thing is that the default transport is WebSockets instead of `TransportType.All`. 
PR https://github.com/aspnet/SignalR/pull/575 will have to be updated after its unblocked.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -39,11 +39,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
         public event Action<Exception> Closed;
 
         public HttpConnection(Uri url)
-            : this(url, TransportType.WebSockets)
+            : this(url, TransportType.All)
         { }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -42,6 +42,10 @@ public HttpConnection(Uri url)
             : this(url, TransportType.WebSockets)
         { }
 
+        public HttpConnection(Uri url, HttpMessageHandler httpMessageHandler)
+            : this(url, TransportType.WebSockets, loggerFactory: null, httpMessageHandler: httpMessageHandler)
```
moozzyk: This should be TransportType.All. I think you should fix it everywhere and we will add the test that is failing in my PR that did that separately.

mikaelm12: Cool, will do.

PR #611, opened 2017-06-28
Remove NETStandard.Library.NETFramework and update Moq

ryanbrandenburg:


build/dependencies.props
```diff
@@ -7,7 +7,7 @@
     <GoogleProtobufVersion>3.1.0</GoogleProtobufVersion>
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
     <JsonNetVersion>10.0.1</JsonNetVersion>
-    <MoqVersion>4.7.1</MoqVersion>
+    <MoqVersion>4.7.49</MoqVersion>
```
moozzyk: Is there any specific reason to update this version? Or it is just that everyone is using it. 
I am asking because I hit a weird issue with Moq yesterday and was wondering if the new version might be the answer.

davidfowl: It's across the board.

ryanbrandenburg: The specific reason we're updating Moq everywhere is that 4.7.1 causes a NuGet warning with Castle.Core, which is resolved in 4.7.49.

Issue #585, opened 2017-06-23
Add a timeout to the EventQueue drain task


mikaelm12:
To avoid a possible deadlock here.
https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs#L163

mikaelm12:
Done: 8fc2cd98b613b5562be4d9b4045b9bb1b2d5683d

PR #619, opened 2017-06-30
Add timeout to Event Queue drain

mikaelm12:
Addresses issue: https://github.com/aspnet/SignalR/issues/585

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -33,7 +33,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private TaskQueue _eventQueue = new TaskQueue();
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
-        private TimeSpan _timeout = TimeSpan.FromSeconds(5);
+        private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
 
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
@@ -175,7 +175,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                     _logger.DrainEvents(_connectionId);
                     await _eventQueue.Drain();
 
-                    await Task.WhenAny(_eventQueue.Drain().NoThrow(), Task.Delay(_timeout));
+                    await Task.WhenAny(_eventQueue.Drain().NoThrow(), Task.Delay(_eventQueueDrainTimeout));
                     _httpClient.Dispose();
 
                     _logger.RaiseClosed(_connectionId);
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -11,6 +11,7 @@ using System.Threading.Tasks.Channels;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Sockets.Features;
 using Microsoft.AspNetCore.Sockets.Client.Internal;
+using Microsoft.AspNetCore.Sockets.Http.Internal;
 using Microsoft.AspNetCore.Sockets.Internal;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
@@ -173,6 +174,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                     _logger.DrainEvents(_connectionId);
                     await _eventQueue.Drain();
 
+                    await Task.WhenAny(_eventQueue.Drain().NoThrow(), Task.Delay(5000));
                     _httpClient.Dispose();
 
                     _logger.RaiseClosed(_connectionId);
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -33,6 +33,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private TaskQueue _eventQueue = new TaskQueue();
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
+        private TimeSpan _timeout = TimeSpan.FromSeconds(5);
 
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
@@ -174,7 +175,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                     _logger.DrainEvents(_connectionId);
                     await _eventQueue.Drain();
 
-                    await Task.WhenAny(_eventQueue.Drain().NoThrow(), Task.Delay(5000));
+                    await Task.WhenAny(_eventQueue.Drain().NoThrow(), Task.Delay(_timeout));
                     _httpClient.Dispose();
 
                     _logger.RaiseClosed(_connectionId);
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/TaskExtensions.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/TaskExtensions.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -162,11 +162,11 @@ public Task StartAsync()
 
                     await _startTcs.Task;
                     await _receiveLoopTask;
-
                     _logger.LogDebug("Draining event queue");
-                    await _eventQueue.Drain();
 
-                    _httpClient.Dispose();
+                    await Task.WhenAny(
```
davidfowl: This line of code is unreadable and too long. We should be using async await and removing the continue with. The task returned from WhenAny won't throw anyways. We should just add the helper that @BrennanConroy added to avoid throwing exceptions for faulted tasks.

We should also discuss configuring options for the client. We might need a builder API or something to make it so we can compose options.

This should look like `Task.WhenAny(eventQueue.Drain().NoThrow(), Task.Delay(Timeout))`

mikaelm12: So I actually started with exactly what you suggested. I copied over the NoThrowAwaiter from [here](https://github.com/aspnet/SignalR/blob/9a53598d8a9a77d676f9d1726f1a45edb60725be/src/Microsoft.AspNetCore.Sockets.Http/Internal/NoThrowTaskExtensions.cs) but you can't pass that into Task.WhenAny.

davidfowl: Yea, because it's not a task but you can easily wrap that.

mikaelm12: @moozzyk preferred this, but he said that you would probably hate it because of the ContinueWith 😄

davidfowl: Yes, the client code is riddled with them and it burns my eyes.

moozzyk: I spent too much time debugging [TaskAsynchHelper](https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Core/TaskAsyncHelper.cs)...

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -136,7 +137,7 @@ public Task StartAsync()
                 throw;
             }
 
-            // if the connection is not in the Connecting state here it means the user called DisposeAsync
```
moozzyk: revert :)

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/TaskExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            return new NoThrowAwaiter(task);
+        }
+        public static async Task NoThrow(this Task task)
+        {
+            await new NoThrowAwaiter(task);
+        }
+    }
+
+    public struct NoThrowAwaiter : ICriticalNotifyCompletion
```
moozzyk: I don't think it should be public. Either make it a private class or move to a different file.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/TaskExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Sockets.Http.Internal
+{
+    public static class NoThrowTaskExtensions
```
moozzyk: the class should be called TaskExtensions. We may want to add things (like OrTimeout()) here in the future - you could say `_eventQueue.Drain().NoThrow().OrTimeout()` (the order is important).

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -437,6 +437,118 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+            var callbackInvokedTcs = new TaskCompletionSource<object>();
+            var closedTcs = new TaskCompletionSource<object>();
+
+            var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
+            connection.Received +=
+                async m =>
+                {
+                    await Task.Delay(6000);
+                    callbackInvokedTcs.SetResult(null);
+                    await closedTcs.Task;
```
moozzyk: why do you need await this?

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -437,6 +437,118 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+                };
+
+            await connection.StartAsync();
+            channel.Out.TryWrite(Array.Empty<byte>());
+
+            // Ensure that the Received callback has been called before attempting the second write
```
moozzyk: I don't think you should send twice. (Does it mean the test takes now 12 seconds?)

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -437,6 +437,118 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+
+
+            var callbackInvokedTcs = new TaskCompletionSource<object>();
+            var closedTcs = new TaskCompletionSource<object>();
+
+            var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
+            connection.Received +=
+                async m =>
+                {
+                    await Task.Delay(6000);
```
moozzyk: don't delay - use a tcs.

test/Microsoft.AspNetCore.SignalR.Client.Tests/ConnectionTests.cs
```diff
@@ -437,6 +437,118 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+
+            closedTcs.SetResult(null);
+
+            await connection.DisposeAsync();
+        }
+
+        [Fact]
+        public async Task EventQueueTimeoutWithException()
+        {
+            var mockHttpHandler = new Mock<HttpMessageHandler>();
```
moozzyk: similar comments as above

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -173,6 +174,7 @@ public Task StartAsync()
                     _logger.DrainEvents(_connectionId);
                     await _eventQueue.Drain();
 
+                    await Task.WhenAny(_eventQueue.Drain().NoThrow(), Task.Delay(5000));
```
davidfowl: Declare 5000 in a file variable.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -32,6 +33,7 @@ public class HttpConnection : IConnection
         private TaskQueue _eventQueue = new TaskQueue();
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
+        private TimeSpan _timeout = TimeSpan.FromSeconds(5);
```
davidfowl: which timeout?

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/TaskExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public static class TaskExtensions
+    {
+        public static async Task NoThrow(this Task task)
+        {
+            await new NoThrowAwaiter(task);
+        }
+    }
+
+    struct NoThrowAwaiter : ICriticalNotifyCompletion
```
davidfowl: internal?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: (also `readonly`)

mikaelm12:
Appveyor is having issues getting things from Nuget. Travis is passing and local build also passes. Unless there are any objections, I'm just gonna merge this.

Issue #547, opened 2017-06-09
We don't set X-Requested-With in our XHR requests


anurse:
We don't set the `X-Requested-With` in our XHR requests. As a result, the `OPTIONS` request we do for negotiate on an authenticated Endpoint tries to do a 302 redirect to the login page, which blows up hard.

We can fix this by explicitly calling `xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest")` in our `HttpClient` code in the TypeScript client

Also, the `HttpClient` currently rejects its Promises with a custom object rather than an `Error` object. We should fix that.

/cc @moozzyk

davidfowl:
So you're basically saying we can keep `OPTIONS` 😄

moozzyk:
Yes

anurse:
![https://media.tenor.com/images/a3c06d4e50bf02b2ba66240dc4938703/tenor.gif](https://media.tenor.com/images/a3c06d4e50bf02b2ba66240dc4938703/tenor.gif)

davidfowl:
Moving this to you @mikaelm12

mikaelm12:
Done: ed6badbabeb876592f2fbe3919b5831292c7da05

PR #623, opened 2017-06-30
Set X-Requested-With header

mikaelm12:
Addresses issue: https://github.com/aspnet/SignalR/issues/547

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
@@ -23,6 +23,7 @@ export class HttpClient implements IHttpClient {
 
             xhr.open(method, url, true);
 
+            headers.set("X-Requested-With", "XMLHttpRequest")
             if (headers) {
                 headers.forEach((value, header) => xhr.setRequestHeader(header, value));
             }
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
@@ -34,11 +34,15 @@ export class HttpClient implements IHttpClient {
                     resolve(xhr.response);
                 }
                 else {
-                    reject(new Error(xhr.statusText));
+                    var errorMessage = xhr.statusText + " Error Status Code: " + xhr.status;
+                    reject(new Error(errorMessage));
                 }
             };
 
-            xhr.onerror = () => reject(new Error(xhr.statusText));
+            xhr.onerror = () => {
+                var errorMessage = xhr.statusText + " Error Status Code: " + xhr.status;
+                reject(new Error(errorMessage));
+            }
         });
     }
 }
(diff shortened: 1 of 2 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
```diff
@@ -23,6 +23,7 @@ export class HttpClient implements IHttpClient {
 
             xhr.open(method, url, true);
 
+            headers.set("X-Requested-With", "XMLHttpRequest")
```
moozzyk: headers is optional so can be undefined/null

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
```diff
@@ -23,6 +23,7 @@ export class HttpClient implements IHttpClient {
...
             if (headers) {
```
moozzyk: this will never be false after we add a header uncoditionally

moozzyk:
There was this part to the bug: _Also, the HttpClient currently rejects its Promises with a custom object rather than an Error object. We should fix that._ which was the reason why the bug was assigned to me. If you don't want to fix it in this change open a new issue on this so we will do this.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
```diff
@@ -34,19 +34,11 @@ export class HttpClient implements IHttpClient {
                     resolve(xhr.response);
                 }
                 else {
-                    reject({
-                        status: xhr.status,
-                        statusText: xhr.statusText
-                    });
+                    reject(new Error(xhr.statusText));
```
moozzyk: With this change you are losing status code. We need both. I would add it to the message details. There might be value in storing the status code in a property as well.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
```diff
@@ -34,19 +34,11 @@ export class HttpClient implements IHttpClient {
...
                 }
             };
 
-            xhr.onerror = () => {
-                reject({
-                    status: xhr.status,
-                    statusText: xhr.statusText
-                });
-            };
+            xhr.onerror = () => reject(new Error(xhr.statusText));
```
moozzyk: It would be nice to have some tests for these but I am not sure how hard it would be to add them...

davidfowl: I was going to suggest tests but I didn't know if we had any for this.

moozzyk: We could add tests for this in the end-to-end tests - there is a web server there and it is easy to request a non-existing page...

davidfowl:
I think we'll need to explicitly look at the error code, as long as that's still possible then 👍

anurse:
Yeah, as @davidfowl said we should add the status code and status text to the error as properties. Easiest way to do that is to create a custom `HttpError` class that extends `Error` and adds those two properties.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpError.ts
```diff
@@ -0,0 +1,7 @@
+﻿export class HttpError extends Error {
+    statusCode: number;
+    constructor(errorMessage: string, errorStatusCode: number) {
```
anurse: `errorStatusCode` can just be `statusCode` here.

PR #628, opened 2017-07-02
Make parsers stateless

davidfowl:
- Changed parser input to ReadOnlySpan<byte> to ReadOnlyBuffer<byte>

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
@@ -6,69 +6,41 @@ using System.Binary;
 
 namespace Microsoft.AspNetCore.Sockets.Internal.Formatters
 {
-    public class BinaryMessageParser
+    public static class BinaryMessageParser
     {
-        private ParserState _state;
-
-        public void Reset()
+        public static bool TryParseMessage(ref ReadOnlyBuffer<byte> buffer, out ReadOnlyBuffer<byte> payload)
         {
-            _state = default(ParserState);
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
```diff
@@ -6,69 +6,41 @@
...
-                length = buffer.Slice(0, sizeof(long)).ReadBigEndian<long>();
-
-                if (length > Int32.MaxValue)
-                {
-                    throw new FormatException("Messages over 2GB in size are not supported");
-                }
-
-                buffer = buffer.Slice(sizeof(long));
-                _state.Length = (int)length;
+                return false;
```
moozzyk: Shouldn't these methods basically throw in this case? At least at the moment we are reading the data and we are not returning it to the buffer so Try* does not make sense because we are discarding the data and hence corrupting messages received in multiple chunks.

davidfowl: It doesn't matter right now because we never send partial messages. Once we move to pipelines, the logic will change and Try* will make sense.

moozzyk: This is what I thought. We will just need to make sure that we don't consume bytes if the message is not complete.

davidfowl: See the pipelines branch for and sneak peek of what that looks like.

PR #629, opened 2017-07-02
Replace ConnectionContext with HubConnectionContext

davidfowl:
- The SocketDelegate implementation owns the transport pipe,
it's a single producer single consumer model. SignalR needs to support
multiple producers so that broadcast, return values and sending to individual
connections works. This change introduces a multi producer channel that is used
by all producers to copy data to the transport safely. This will make the move
to pipelines easier.


src/Microsoft.AspNetCore.SignalR/HubConnectionList.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubConnectionList : IReadOnlyCollection<HubConnectionContext>
```
davidfowl: I could have made:

```C#
public interface ICanHasConnectionId
{
    string ConnectionId { get; }
}

public ConnectionList<TConnection> where TConnection : ICanHasConnectionId { }
```

But I decided not to.

src/Microsoft.AspNetCore.SignalR/HubConnectionContext.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.AspNetCore.Sockets;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubConnectionContext
+    {
+        private readonly WritableChannel<byte[]> _output;
+        private readonly ConnectionContext _connectionContext;
+
+        public HubConnectionContext()
```
moozzyk: remove - I don't see it being used anywhere and it creates an unusable instance.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -59,24 +59,54 @@ public class HubEndPoint<THub, TClient> : IInvocationBinder where THub : Hub<TCl
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
-            await ProcessNegotiate(connection);
+            var output = Channel.CreateUnbounded<byte[]>();
+            var connectionContext = new HubConnectionContext(output, connection);
+
+            await ProcessNegotiate(connectionContext);
```
moozzyk: Almost not related: We need to be able to switch the mode of the transport (text vs. binary) here depending on the negotiated protocol. I can't find a good way to do this.

src/Microsoft.AspNetCore.SignalR/HubConnectionList.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
```
moozzyk: nit: header
(also System might not be needed)

src/Microsoft.AspNetCore.SignalR/HubConnectionList.cs
```diff
@@ -0,0 +1,49 @@
...
+        public int Count => _connections.Count;
+
+        public void Add(HubConnectionContext connection)
+        {
+            _connections.TryAdd(connection.ConnectionId, connection);
+        }
+
+        public void Remove(HubConnectionContext connection)
+        {
+            _connections.TryRemove(connection.ConnectionId, out var dummy);
```
moozzyk: nit: _

src/Microsoft.AspNetCore.SignalR/HubConnectionList.cs
moozzyk: Isn't the HubConnectionContext the real deal here though? This is basically a copy of ConnectionList and we need it because HubConnectionContext wraps ConnectionContext...

src/Microsoft.AspNetCore.SignalR/HubConnectionContext.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
```
moozzyk: nit: make sure to add headers.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
davidfowl: Why do we need to do this? We moved to a place where the protocol itself is all we need. Once you negotiate json or protobuf, nothing else needs to be implied.

moozzyk: The problem is that negotiation happens over a connection that has already been opened. You can set mode in WebSocketOptions but it is before negotiation happens and this is the client who requests a protocol so you don't know what to set it to. For webSockets we could try always sending binary messages but it requires converting to string on the client. This won't work with SSE though. SSE either needs to reject binary protocols or base64 encode but since the transport is decoupled from connection neither side knows if it should do that.

davidfowl: I think the mode is known by the client up front. I would just throw on the client side and not introduced any sever handshake dance for this.

moozzyk: It is known on the client but it is not known on the server - the server picks the protocol upon receiving negotiate from the client. (i.e. the server cannot use _options.WebSocketMessageType here: https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs#L183 because it was set before negotiation and therefore no-one knew what protocol the client will request).

moozzyk: in other words 
- if the client requested Protobuf, but webSocketOptions.WebSocketMessageType was set to text (which is the default) the server will send binary data over text violating the protocol. 
- if the user sets webSocketOptions.WebSocketMessageType to binary but the client requests JSon we will send text messages over binary which requires additional processing on the client to convert to string
- with SSE it does not work at all because it is a text protocol. We can't base64 encode everything though since we don't want to base64 encode text messages

davidfowl: It's not negotiation, that's a misnomer. They client tells the sever what protocol it supports, the sever either has that exact protocol or it throws.

Seems like we're only talking about websockets here. The sever has settting per endpoint that determines what to send (binary or text) and there's no way to do this in a per connection level. Maybe that's what you mean? Assume we have a way to set that in some connection metadata or feature so that it can be read by the websocket transport.

For SSE why would the client bother to send a request to the server if we ended up with an invalid combination, say "protobuf and SSE" or msgpack and SSE? 

Were we going to support both http transport and hub protocol negotiation?

moozzyk: > The sever has settting per endpoint that determines what to send (binary or text) and there's no way to do this in a per connection level. Maybe that's what you mean? Assume we have a way to set that in some connection metadata or feature so that it can be read by the websocket transport.
Yes, not being able to set it per connection is the first blocker. And yes, I wanted to set it on Metadata (only it felt quite gross).

> For SSE why would the client bother to send a request to the server if we ended up with an invalid combination, say "protobuf and SSE" or msgpack and SSE?
- you request Transport.All and wanted to use ProtoBuf. Then you run on Win7 and you get the best available transport which is SSE. You end up with SSE and binary. 
- another problem is that if it is not supported the server has not enough information to reject invalid combinations if this is not supported

davidfowl: Why wouldn't the client just outright reject SSE + protobuf?

moozzyk: Our client could (3rd party clients might decide not to) but it only feels like a convenience - IMHO the server should always reject invalid requests and not rely on the client.

davidfowl: Not sure the complexity is worth it.

src/Microsoft.AspNetCore.SignalR/HubConnectionList.cs
davidfowl: I'm just saying that I could have reused ConnectionList if I made it generic and made an interface for things that have connection ids.

src/Microsoft.AspNetCore.SignalR/HubConnectionContext.cs
davidfowl: This was for testing. Maybe it should be internal.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
moozzyk: Actually even our client may not be able to reject SSE + binary protocol:

```CSharp
var httpConnection = new HttpConnection(new Uri(url), TransportType.ServerSentEvents);
var hubConnection = new HubConnection(httpConnection, new MessagePackHubProtocol());
```
who is supposed to reject this combination? HttpConnection has no notion of the protocol, HubConnection has no notion of transport....

davidfowl: Killed by decoupling 😉

moozzyk: Prototype is here - https://github.com/aspnet/SignalR/pull/632/files. I marked what I find gross. Open to ideas.

test/Microsoft.AspNetCore.SignalR.Tests/Internal/DefaultHubProtocolResolverTests.cs
```diff
@@ -18,7 +19,7 @@ public class DefaultHubProtocolResolverTests
         [MemberData(nameof(HubProtocols))]
         public void DefaultHubProtocolResolverTestsCanCreateSupportedProtocols(IHubProtocol protocol)
         {
-            var mockConnection = new Mock<HubConnectionContext>();
+            var mockConnection = new Mock<HubConnectionContext>(Channel.CreateUnbounded<byte[]>().Out, new Mock<ConnectionContext>().Object);
```
moozzyk: Have you tried `var mockConnection = Mock.Of<ConnectionContext>()` ? - asking because it blew up on me on full .NET...

src/Microsoft.AspNetCore.SignalR/HubConnectionList.cs
anurse: Well, if you're going to strawman the name, then it sounds ridiculous, but I'm not so sure it's a bad idea. Something like `IConnectionContext` with a `ConnectionId` property. Or a utility base class that exposes a single `protected abstract` method like `GetConnectionId` (kinda like `KeyedCollection` does)

davidfowl: It's just not important enough at the moment. ConnectionList isn't really that fundamental, it's there for convenience.

PR #632, opened 2017-07-04
WIP plugging things in

moozzyk:
**Prototype**

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IConnection.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IConnection.ts
@@ -2,7 +2,7 @@ import { DataReceived, ConnectionClosed } from "./Common"
 import { TransportType, ITransport } from  "./Transports"
 
 export interface IConnection {
-    start(): Promise<void>;
+    start(isBinary: boolean): Promise<void>;
     send(data: any): Promise<void>;
     stop(): void;
 
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
@@ -30,6 +30,6 @@ export interface NegotiationMessage {
 
 export interface IHubProtocol {
     name(): string;
-    parseMessages(input: string): HubMessage[];
-    writeMessage(message: HubMessage): string;
+    parseMessages(input: any): HubMessage[];
+    writeMessage(message: HubMessage): any;
 }
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
@@ -30,6 +30,7 @@ export interface NegotiationMessage {
 
 export interface IHubProtocol {
     name(): string;
+    isBinary(): boolean;
     parseMessages(input: any): HubMessage[];
     writeMessage(message: HubMessage): any;
 }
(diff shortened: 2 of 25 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -196,6 +207,8 @@ export class LongPollingTransport implements ITransport {
             }
             else {
                 if (this.onClosed) {
+                    // TODO: Uncaught DOMException: Failed to read the 'responseText' property from 'XMLHttpRequest':
+                    // The value is only accessible if the object's 'responseType' is '' or 'text' (was 'arraybuffer').
                     this.onClosed(new Error(`Status: ${pollXhr.status}, Message: ${pollXhr.responseText}`));
```
moozzyk: @mikaelm12 - this tells me that putting statusText to error message is more tricky than it seems.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IConnection.ts
```diff
@@ -2,7 +2,7 @@ import { DataReceived, ConnectionClosed } from "./Common"
 import { TransportType, ITransport } from  "./Transports"
 
 export interface IConnection {
-    start(): Promise<void>;
+    start(isBinary: boolean): Promise<void>;
```
davidfowl: I don't like this.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
```diff
@@ -30,6 +30,7 @@ export interface NegotiationMessage {
 
 export interface IHubProtocol {
     name(): string;
+    isBinary(): boolean;
```
moozzyk: 💣

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IConnection.ts
moozzyk: :bomb:

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/IHubProtocol.cs
```diff
@@ -11,6 +11,8 @@ public interface IHubProtocol
     {
         string Name { get; }
 
+        ProtocolType ProtocolType { get; }
```
moozzyk: :bomb:

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -83,9 +83,9 @@ public class HubEndPoint<THub, TClient> : IInvocationBinder where THub : Hub<TCl
                         // Resolve the Hub Protocol for the connection and store it in metadata
                         // Other components, outside the Hub, may need to know what protocol is in use
                         // for a particular connection, so we store it here.
-                        connection.Metadata[HubConnectionMetadataNames.HubProtocol] =
-                            _protocolResolver.GetProtocol(negotiationMessage.Protocol, connection);
-
+                        var protocol = _protocolResolver.GetProtocol(negotiationMessage.Protocol, connection);
+                        connection.Metadata[HubConnectionMetadataNames.HubProtocol] = protocol;
+                        connection.Metadata["Binary"] = protocol.ProtocolType == ProtocolType.Binary;
```
moozzyk: :bomb:

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IConnection.ts
moozzyk: neither do I but is required to make longPolling work.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
moozzyk: I think this could/should be done as part of `OPTIONS` in which case we would just create the transport using the binary/text mode.

moozzyk:
This was only a prototype. Closing.

PR #639, opened 2017-07-06
Features everywhere

davidfowl:
- The goal here is to move things closer to the final design where
ConnectionContext represents a very low level primitive that represents
any connection like transport. As part of that change, we remove unnecessary
properties like User and move those into features. They temporarily live in the same
assembly but they are not required by ConnectionContext.
- Used features for Hubs instead of Metadata
- Metadata is no longer thread safe


src/Microsoft.AspNetCore.Sockets/ConnectionMetadata.cs
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    internal class ConnectionMetadata : IDictionary<object, object>
```
davidfowl: This is pretty much a clone from https://github.com/aspnet/HttpAbstractions/blob/a79aea3bf379ef0752567474d9490279c6818ab1/src/Microsoft.AspNetCore.Http/Internal/ItemsDictionary.cs

src/Microsoft.AspNetCore.Sockets/ConnectionMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        public ConnectionMetadata(IDictionary<object, object> items)
+        {
+            Items = items;
+        }
+
+        public IDictionary<object, object> Items { get; }
+
+        // Replace the indexer with one that returns null for missing values
```
moozzyk: Is this comment still valid?

src/Microsoft.AspNetCore.Sockets/ConnectionMetadata.cs
```diff
@@ -0,0 +1,119 @@
...
+        object IDictionary<object, object>.this[object key]
+        {
+            get
+            {
+                object value;
+                if (Items.TryGetValue(key, out value))
```
moozzyk: nit: out var

samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
```diff
@@ -60,7 +61,7 @@ public Task InvokeUserAsync(string userId, object data)
 
         public void AddGroupAsync(ConnectionContext connection, string groupName)
         {
-            var groups = connection.Metadata.GetOrAdd("groups", _ => new HashSet<string>());
+            var groups = (HashSet<string>)connection.Metadata;
```
moozzyk: shouldn't this be: `(HashSet<string>)connection.Metadata["groups"];`?

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -91,8 +90,9 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
         {
             return InvokeAllWhere(methodName, args, connection =>
             {
-                var groups = connection.Metadata.Get<HashSet<string>>(HubConnectionMetadataNames.Groups);
-                return groups?.Contains(groupName) == true;
+                // REVIEW: Not thread safe... (we should be locking on groups)
```
moozzyk: Why not do it now?

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -106,6 +106,9 @@ public override Task InvokeUserAsync(string userId, string methodName, object[]
 
         public override Task OnConnectedAsync(HubConnectionContext connection)
         {
+            // Set the hub groups feature
+            connection.Features.Set<IHubGroupsFeature>(new HubGroupsFeature());
```
moozzyk: Why not set it when creating the HubConnectionContext? This is creating a pit of failure (NRE or races) in cases where people want to add the connection to a group in their OnConnectedAsync which is common.

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
```diff
@@ -1,25 +1,20 @@
...
+using System.Collections.Generic;
 using System.Threading.Tasks.Channels;
 using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Sockets
 {
     public abstract class ConnectionContext
     {
-        public abstract string ConnectionId { get; }
+        public abstract string ConnectionId { get; set; }
```
moozzyk: Why does this need to be settable now?

src/Microsoft.AspNetCore.Sockets/DefaultConnectionContext.cs
```diff
@@ -36,13 +49,13 @@ public DefaultConnectionContext(string id, Channel<byte[]> transport, Channel<by
 
         public ConnectionStatus Status { get; set; } = ConnectionStatus.Inactive;
 
-        public override string ConnectionId { get; }
+        public override string ConnectionId { get; set; }
```
moozzyk: Should not be settable

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionTransportFeature.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks.Channels;
+
+namespace Microsoft.AspNetCore.Sockets.Features
+{
+    public interface IConnectionTransportFeature
+    {
+        Channel<byte[]> Transport { get; set; }
```
moozzyk: remove setter

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionMetadataFeature.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Sockets.Features
+{
+    public interface IConnectionMetadataFeature
+    {
+        IDictionary<object, object> Metadata { get; set; }
```
moozzyk: setter should not be needed.

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionIdFeature.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Sockets.Features
+{
+    public interface IConnectionIdFeature
+    {
+        string ConnectionId { get; set; }
```
moozzyk: remove setter

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -25,7 +25,7 @@ public class TestClient : IDisposable, IInvocationBinder
 
         public DefaultConnectionContext Connection { get; }
         public Channel<byte[]> Application { get; }
-        public Task Connected => Connection.Metadata.Get<TaskCompletionSource<bool>>("ConnectedTask").Task;
+        public Task Connected => (Connection.Metadata["ConnectedTask"] as TaskCompletionSource<bool>).Task;
```
moozzyk: why `as` casting? Implies that ConnectedTask might not be a `TaskCompletionSource<bool>`. If this was true we should handle this.

src/Microsoft.AspNetCore.Sockets/DefaultConnectionContext.cs
```diff
@@ -36,13 +49,13 @@ public DefaultConnectionContext(string id, Channel<byte[]> transport, Channel<by
...
 
-        public override IFeatureCollection Features { get; } = new FeatureCollection();
+        public override IFeatureCollection Features { get; }
 
-        public override ClaimsPrincipal User { get; set; }
+        public ClaimsPrincipal User { get; set; }
 
-        public override ConnectionMetadata Metadata { get; } = new ConnectionMetadata();
+        public override IDictionary<object, object> Metadata { get; set; } = new ConnectionMetadata();
```
moozzyk: remove setter

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
```diff
@@ -1,25 +1,20 @@
...
 
         public abstract IFeatureCollection Features { get; }
 
-        public abstract ClaimsPrincipal User { get; set; }
+        public abstract IDictionary<object, object> Metadata { get; set; }
```
moozzyk: has to be settable?

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
davidfowl: Like most (if not all) the things on the HttpContext, this is settable so that middleware can mutate.

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
davidfowl: Like most (if not all) the things on the HttpContext, this is settable so that middleware can mutate.

src/Microsoft.AspNetCore.Sockets/DefaultConnectionContext.cs
davidfowl: Like most (if not all) the things on the HttpContext, this is settable so that middleware can mutate.

src/Microsoft.AspNetCore.Sockets/DefaultConnectionContext.cs
davidfowl: Like most (if not all) the things on the HttpContext, this is settable so that middleware can mutate.

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionTransportFeature.cs
davidfowl: Like most (if not all) the things on the HttpContext, this is settable so that middleware can mutate.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
davidfowl: I removed the `Get<T>` methods. This is functionally equivalent.

src/Microsoft.AspNetCore.Sockets/ConnectionMetadata.cs
davidfowl: Yep

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
davidfowl: Pit of failure for who exactly? This runs before application code runs and it's private to the implementation of this implementation of the HubLifetimeManager.

samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
davidfowl: Good catch

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionUserFeature.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Sockets.Features
+{
+    public interface IConnectionUserFeature
+    {
+        ClaimsPrincipal User { get; }
```
anurse: Why is this **not** settable? Of all the feature properties, it's the one middleware is most likely to want to mutate.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
anurse: Should use a regular cast then, since you'll fail if this returns null.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
moozzyk: I confused this OnConnectedAsync with the one that is on Hub so ignore my comment.

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionContext.cs
moozzyk: I did not have the bigger picture - these setters are not needed not but will be used in the future by the pipeline.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
moozzyk: I am saying it should be: `TaskCompletionSource<bool>)Connection.Metadata["ConnectedTask"]` because we don't expect this to be of a different type?

davidfowl: Ah sure.

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionUserFeature.cs
davidfowl: Should be, good catch.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
davidfowl: Will do

davidfowl:
🆙 📅

PR #642, opened 2017-07-06
Downgrade special version to alpha1

pranavkm:


davidfowl:
ugh.

davidfowl:
Can you write an announcement plz.

PR #645, opened 2017-07-07
Pawelka/transfer mode

moozzyk:
Yet another prototype for binary protocol support. Works end-to-end - possible to send  base64 encoded MessagePack messages over SSE. 

Issues:
- cheating - messages sent from the client are just binary because post can handle that - only messages from the server over event-stream are base64 encoded. Hard to make it symmetric because of the first message the client sends to the server. This message contains the protocol to be used and is supposed to be text only. This protocol is on Hub level and encoding happens in the HttpConnection.
- ~The SSE transport could not be stopped from the client (not sure if any of the end-to-end tests run against SSE - the one I tried always hanged/timed out)~ fixed in a41ef82f19e76ed13fb8d276fa3d1ec3803179ff
- Need to pass ConnectionContext to transports to be able to get the ITransferModeFeature
- ~On the server side need to base64 encode in the SSE transport because we give the channel to the user~
- half of the tests fail due to NotImplementedException, the other half was blindly updated to compile - they need to be reviewed whether slapping TransferMode.Text is OK and whether we need to test the other mode. 
- need to find the right place for TransferMode et al.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/Microsoft.AspNetCore.SignalR.Client.csproj
+++ b/src/Microsoft.AspNetCore.SignalR.Client/Microsoft.AspNetCore.SignalR.Client.csproj
@@ -14,6 +14,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.SignalR.Common\Microsoft.AspNetCore.SignalR.Common.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Sockets.Abstractions\Microsoft.AspNetCore.Sockets.Abstractions.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Sockets.Common.Http\Microsoft.AspNetCore.Sockets.Common.Http.csproj" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
 
--- a/src/Microsoft.AspNetCore.SignalR/DataEncoderFeature.cs
+++ b/src/Microsoft.AspNetCore.SignalR/DataEncoderFeature.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Text;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public interface IDataEncoderFeature
+    {
+        IDataEncoder Encoder { get; set; }
+    }
+
+    public class DataEncoderFeature : IDataEncoderFeature
+    {
+        public IDataEncoder Encoder { get; set; }
+    }
+
+    public interface IDataEncoder
+    {
+        byte[] Encode(byte[] message);
+    }
+
+    public class PassThroughEncoder : IDataEncoder
+    {
+        public byte[] Encode(byte[] message)
+        {
+            return message;
+        }
+    }
+
+    public class Base64Encoder : IDataEncoder
+    {
+        public byte[] Encode(byte[] message)
+        {
+            return Encoding.UTF8.GetBytes(Convert.ToBase64String(message));
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using System.Net.Http;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Channels;
@@ -50,7 +51,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         { }
 
         public HttpConnection(Uri url, TransportType transportType)
-                    : this(url, transportType, loggerFactory: null)
+            : this(url, transportType, loggerFactory: null)
         {
         }
 
@@ -292,9 +293,13 @@ namespace Microsoft.AspNetCore.Sockets.Client
                     if (Input.TryRead(out var buffer))
                     {
                         var transferMode = GetTransferMode();
-                        if (transferMode != _transport.Mode)
+                        if (transferMode == TransferMode.Binary && _transport.Mode == TransferMode.Text)
                         {
-                            // TODO: Binary <--> Text conversion happens here
+                            buffer = Convert.FromBase64String(Encoding.UTF8.GetString(buffer));
+                        }
+                        else if (transferMode == TransferMode.Text && _transport.Mode == TransferMode.Binary)
+                        {
+                            // This should never happen with currently supported transports
                             throw new NotImplementedException();
                         }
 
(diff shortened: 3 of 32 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -344,12 +349,17 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
                     "Cannot send messages when the connection is not in the Connected state.");
             }
 
+            /* Gross - send as binary HttpRequest can handle that
```
moozzyk: The problem is that the first message sent by HubConnection is supposed to be only text but HttpConnection will convert everything.

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -77,11 +78,18 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
...
 
             var pipelineReader = stream.AsPipelineReader();
+            // HACK: CancellationToken on ReadAsync is ignored
+            _pipeReader = pipelineReader;
             try
             {
                 while (true)
                 {
-                    var result = await pipelineReader.ReadAsync();
+                    var result = await pipelineReader.ReadAsync(cancellationToken);
```
moozzyk: https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs#L289-L292

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -391,15 +388,19 @@ private static string GetNegotiatePayload(string connectionId, HttpSocketOptions
 
             _logger.ReceivedBytes(connection.ConnectionId, buffer.Length);
 
-            var transferModeFeature = connection.Features.Get<ITransferModeFeature>();
-            if (transferModeFeature != null)
+            // uber uber gross - send is binary only receive is base64 encoded
+            // the issue: The first message - clients encodes but the server does not know it needs to decode
+            /*
+            // TODO: uber gross
+            if ((TransportType)connection.Metadata[ConnectionMetadataNames.Transport] == TransportType.ServerSentEvents)
```
moozzyk: didn't work because of the first message

src/Microsoft.AspNetCore.SignalR.Client/Microsoft.AspNetCore.SignalR.Client.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.SignalR.Common\Microsoft.AspNetCore.SignalR.Common.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Sockets.Abstractions\Microsoft.AspNetCore.Sockets.Abstractions.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Sockets.Common.Http\Microsoft.AspNetCore.Sockets.Common.Http.csproj" />
```
moozzyk: Is this package name correct (i.e. Microsoft.AspNetCore.Sockets.Common.Http vs Microsoft.AspNetCore.Sockets.Http.Common)? Nota bene - it only contains TransportType at the moment...

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -40,11 +41,13 @@ public ServerSentEventsTransport(HttpClient httpClient, ILoggerFactory loggerFac
             _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<ServerSentEventsTransport>();
         }
 
-        public Task StartAsync(Uri url, Channel<byte[], SendMessage> application)
+        public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode)
         {
             _logger.LogInformation("Starting {transportName}", nameof(ServerSentEventsTransport));
 
             _application = application;
+            Mode = TransferMode.Text; // SSE transport supports only text mode
```
moozzyk: The actual mode can be different from the requested mode which tells the connection it needs to convert data

src/Microsoft.AspNetCore.Sockets.Common.Http/TransferMode.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    public enum TransferMode
+    {
+        Binary = 0x01,
```
moozzyk: these are not supposed to be flags (anymore)

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -82,8 +82,11 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
                     return;
                 }
 
+                // TODO: We need to handle conversion for binary protocols. We may need to give the transport a different
```
moozzyk: to be removed - the conversion was pushed to the transport

moozzyk:
@davidfowl - moved up base64 up the stack on the server side so that it does not happen in the transport but on the other side of the pipe (i.e. before writing to the channel)

src/Microsoft.AspNetCore.SignalR/DataEncoderFeature.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
```
mikaelm12: nit: Header

src/Microsoft.AspNetCore.SignalR/DataEncoderFeature.cs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public interface IDataEncoder
+    {
+        byte[] Encode(byte[] message);
+    }
+
+    public class PassThroughEncoder : IDataEncoder
+    {
+        public byte[] Encode(byte[] message)
```
mikaelm12: You can make this one line

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -82,8 +82,12 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
+                // channel/adapter to be able to do Binary <--> Text conversion
+
+                connection.Metadata["TransportCapabilities"] = TransferMode.Text;
```
davidfowl: Why not just put this on a feature?

moozzyk: It can go on TransportFeature.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -120,12 +120,13 @@ public HubConnectionTests()
 
         [Theory]
         [MemberData(nameof(HubProtocols))]
+        //TODO: should be across transports
```
mikaelm12: You introduce this in your other pr, right?

moozzyk: Yes. This PR is only a very hacky prototype. I am using this test to verify things work end-to-end.

src/Microsoft.AspNetCore.SignalR/DataEncoderFeature.cs
moozzyk: Yes. Will do if we decide to go in this direction.

src/Microsoft.AspNetCore.SignalR/DataEncoderFeature.cs
moozzyk: It's just a prototype so I didn't care.

mikaelm12: lol

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
davidfowl: Let's do that.

moozzyk:
This was only a prototype and is no longer needed.

PR #658, opened 2017-07-12
Check for the existence of a git repository before invoking git commands

pranavkm:


pranavkm:
Getting the change in to unblock the build. I'll keep the PR open to verify the change makes sense.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -40,12 +40,17 @@
     <PropertyGroup>
       <ModuleVersion>@(ExecOutput)</ModuleVersion>
       <ModuleVersion Condition="'$(VersionSuffix)' != ''">$(ModuleVersion)-$(VersionSuffix)</ModuleVersion>
+      <IsGitRepository Condition="EXISTS('$(MSBuildThisFileDirectory)..\..\.git\HEAD')">true</IsGitRepository>
+      <IsGitRepository Condition="'$(IsGitRepository)'==''">false</IsGitRepository>
```
moozzyk: nit: you could flip these and set unconditionally IsGitRepository to false and set it to true only if condition is met.

PR #659, opened 2017-07-13
Plumbing TransferMode on the client side

moozzyk:


src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -34,8 +36,16 @@ public LongPollingTransport(HttpClient httpClient, ILoggerFactory loggerFactory)
...
+        public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode)
         {
+            if (requestedTransferMode != TransferMode.Binary && requestedTransferMode != TransferMode.Text)
+            {
+                throw new ArgumentException("Invalid transfer mode.", nameof(requestedTransferMode));
+            }
+
+            Mode = requestedTransferMode;
+
+            // TODO: log mode
```
moozzyk: Need to rebase on @BrennanConroy 's changes

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -269,6 +278,30 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
...
+            if (transferModeFeature == null)
+            {
+                // REVIEW: use TransferMode.Text as default if the feature is missing?
+                throw new InvalidOperationException("TransferModeFeature not set.");
+            }
+
+            return transferModeFeature.TransferMode;
+        }
+
+        private void SetTransferMode(TransferMode transferMode)
```
davidfowl: Why doesn't it just add it?

moozzyk: not sure what you are asking...

davidfowl: The feature, why does this method throw if the feature isn't there instead of just adding the feature on the fly.

moozzyk: Oh that. This code cannot be reached at the moment. We first call GetTransferMode() first and this one will throw. The question is (and the reason why both methods have the `REVIEW` comment) - should GetTransferMode() throw or should it assume Text if the feature is missing (I am leaning towards using text). If it does not throw (i.e. we allow to start the connection without setting the ITransferModeFeature) this one should just add the feature.

moozzyk: This is now fixed.

moozzyk:
:up::date: - rebased on latest dev and enabled starting HttpConnection without TransferMode feature (defaults to text)

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/ITransferModeFeature.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Sockets.Abstractions.Features
```
moozzyk: Fix the namespace - all other features are in Microsoft.AspNetCore.Sockets.Features

moozzyk:
:watch:

moozzyk:
In-person OK.

PR #660, opened 2017-07-13
Accounting for ReadAsync throwing OperationCanceledException

moozzyk:
Because https://travis-ci.org/aspnet/SignalR/jobs/253286760


src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -113,6 +113,10 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, string
                     }
                 }
             }
+            catch (OperationCanceledException)
+            {
+                // transport is being closed
```
BrennanConroy: Log here!

moozzyk: Fixed

PR #668, opened 2017-07-14
Support for binary protocols on the server side

moozzyk:
(reopening after rebasing on dev)

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -89,6 +89,9 @@ namespace Microsoft.AspNetCore.Sockets
 
                 _logger.EstablishedConnection(connection.ConnectionId, context.TraceIdentifier);
 
+                // ServerSentEvents is a text protocol only
+                connection.TransportCapabilities = TransferMode.Text;
+
                 // We only need to provide the Input channel since writing to the application is handled through /send.
                 var sse = new ServerSentEventsTransport(connection.Application.In, connection.ConnectionId, _loggerFactory);
 
@@ -112,7 +115,7 @@ namespace Microsoft.AspNetCore.Sockets
 
                 _logger.EstablishedConnection(connection.ConnectionId, context.TraceIdentifier);
 
-                var ws = new WebSocketsTransport(options.WebSockets, connection.Application, connection.ConnectionId, _loggerFactory);
+                var ws = new WebSocketsTransport(options.WebSockets, connection.Application, connection, _loggerFactory);
 
                 await DoPersistentConnection(socketDelegate, ws, context, connection);
             }
@@ -330,7 +333,7 @@ namespace Microsoft.AspNetCore.Sockets
 
             // Establish the connection
             var connection = _manager.CreateConnection();
-            
+
             // Set the Connection ID on the logging scope so that logs from now on will have the
             // Connection ID metadata set.
             logScope.ConnectionId = connection.ConnectionId;
@@ -432,6 +435,9 @@ namespace Microsoft.AspNetCore.Sockets
             // Setup the connection state from the http context
             connection.User = context.User;
             connection.SetHttpContext(context);
+
+            // this is the default setting which should be overwritten by transports that have different capabilities (e.g. SSE)
+            connection.TransportCapabilities = TransferMode.Binary | TransferMode.Text;
 
             // Set the Connection ID on the logging scope so that logs from now on will have the
             // Connection ID metadata set.
--- a/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
@@ -18,9 +18,9 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Transports
         private readonly WebSocketOptions _options;
         private readonly ILogger _logger;
         private readonly Channel<byte[]> _application;
-        private readonly string _connectionId;
+        private readonly DefaultConnectionContext _connection;
 
-        public WebSocketsTransport(WebSocketOptions options, Channel<byte[]> application, string connectionId, ILoggerFactory loggerFactory)
+        public WebSocketsTransport(WebSocketOptions options, Channel<byte[]> application, DefaultConnectionContext connection, ILoggerFactory loggerFactory)
         {
             if (options == null)
             {
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Transports
 
             _options = options;
             _application = application;
-            _connectionId = connectionId;
+            _connection = connection;
             _logger = loggerFactory.CreateLogger<WebSocketsTransport>();
         }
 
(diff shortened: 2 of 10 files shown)


moozzyk:
:watch:
appveyor is failing because it is broken and can't restore packages

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -110,9 +113,11 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
                     return;
                 }
 
+
```
davidfowl: nit: Remove space.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -18,9 +19,9 @@ public class WebSocketsTransport : IHttpTransport
         private readonly WebSocketOptions _options;
         private readonly ILogger _logger;
         private readonly Channel<byte[]> _application;
-        private readonly string _connectionId;
+        private readonly ConnectionContext _connection;
```
davidfowl: Change to `DefaultConnectionContext` and just access the property. Or pass the mode in directly.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -176,22 +177,32 @@ public WebSocketsTransport(WebSocketOptions options, Channel<byte[]> application
                     {
                         try
                         {
-                            _logger.SendPayload(_connectionId, buffer.Length);
+                            _logger.SendPayload(_connection.ConnectionId, buffer.Length);
+
+                            var webSocketMessageType = _options.WebSocketMessageType;
```
davidfowl: Why do we need `WebSocketMessageType` still?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -176,22 +177,32 @@ public WebSocketsTransport(WebSocketOptions options, Channel<byte[]> application
...
+                            var transferModeFeature = _connection.Features.Get<ITransferModeFeature>();
+                            if (transferModeFeature != null)
```
davidfowl: Just pass this directly into the ctor of the WebSocketTransport instead of getting the feature each time. It doesn't change dynamically (or shouldn't).

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -110,9 +113,11 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
                 _logger.EstablishedConnection(connection.ConnectionId, context.TraceIdentifier);
 
-                var ws = new WebSocketsTransport(options.WebSockets, connection.Application, connection.ConnectionId, _loggerFactory);
+                connection.TransportCapabilities = TransferMode.Binary | TransferMode.Text;
```
davidfowl: Why do we need to set this in both places?

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
```diff
@@ -29,7 +31,8 @@ public class WebSocketsTests
             using (var applicationSide = ChannelConnection.Create<byte[]>(transportToApplication, applicationToTransport))
             using (var feature = new TestWebSocketConnectionFeature())
             {
-                var ws = new WebSocketsTransport(new WebSocketOptions(), transportSide, connectionId: string.Empty, loggerFactory: new LoggerFactory());
+                var connectionContext = new DefaultConnectionContext(string.Empty, null, null);
```
davidfowl: This isn't required, we should just pass the connection id and the enum directly.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
moozzyk: The problem is that we don't know the mode when the transport is created. We only know it after receiving the first message from the client (negotiation) which contains the protocol to use.

davidfowl: Then change it to `DefaultConnectionContext` for now but this is a bit of a hack 😄 .

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
moozzyk: I left it so that the user can set the WebSocketMessageType when using HttpConnection. I don't like this setting though at all since it is very general and applies to all WebSockets.  I am happy to remove if we are OK with always using the feature.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
moozzyk: It actually change dynamically so can't be passed to the ctor since we don't know. We have a negotiation message where the client selects the protocol which tells us what WebSocketMessageType to use.

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
moozzyk: we don't know what enum to pass here.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
davidfowl: It can't change dynamically. It's set once. You mean it's set later than the ctor.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
davidfowl: Just nuke it.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
moozzyk: where is the second place?

davidfowl: EnsureConnectionStateAsync

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
moozzyk: OK

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
moozzyk: Correct. We set it after the transport is created bug before the first message is sent. It should not change after that.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
moozzyk: The one in EnsureConnectionStateAsync sets it for long polling. But I agree it is misplaced.

davidfowl: EnsureConnectionStateAsync is called for all transports, we should just set it there and override it where it needs to be different.

moozzyk:
:up::date:

PR #670, opened 2017-07-18
Add logging scope

BrennanConroy:


src/Microsoft.AspNetCore.Sockets.Http/Internal/ConnectionLogScope.cs
```diff
@@ -0,0 +1,85 @@
...
+            return _cachedToString;
+        }
+
+        private string ConnectionId
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(_connectionId))
+                {
+                    _connectionId = _context.Request.Query["id"];
```
davidfowl: I was thinking about any potential concurrency issues with this. This might potentially access the `HttpContext` on different threads (in parallel) right? Specifically here https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs#L218. The saving grace is likely that the feature is already populated *and* the value is likely stored before then. We should put a comment in here....

BrennanConroy: Maybe we should add a lock after checking the first `IsNullOrEmpty`

davidfowl: Ugh, no 😄

BrennanConroy: Ok... I'll add a comment "This most likely works, hopefully the id is set before multiple threads attempt to access the scope for logging"

davidfowl: Lets verify its true

src/Microsoft.AspNetCore.Sockets.Http/Internal/ConnectionLogScope.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Sockets.Internal
+{
+    public class ConnectionLogScope : IReadOnlyList<KeyValuePair<string, object>>
+    {
+        private string _connectionId;
+
```
mikaelm12: nit: Get rid of this new line

src/Microsoft.AspNetCore.Sockets.Http/Internal/ConnectionLogScope.cs
BrennanConroy: Good news, I can move `_connectionId = _context.Request.Query["id"];` to the constructor since it wont change during the lifetime of the scope. Bad news, it is technically possible to access this in parallel.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -30,33 +30,37 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
 
         public async Task ExecuteAsync(HttpContext context, HttpSocketOptions options, SocketDelegate socketDelegate)
         {
-            if (!await AuthorizeHelper.AuthorizeAsync(context, options.AuthorizationData))
+            var logScope = new ConnectionLogScope(context.Request.Query["id"]);
```
davidfowl: Add a comment explaining this. Maybe write a helper called GetConnectionId and call it from both places in this file.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -420,6 +423,7 @@ private static string GetNegotiatePayload(string connectionId, HttpSocketOptions
             // Setup the connection state from the http context
             connection.User = context.User;
             connection.SetHttpContext(context);
+            logScope.ConnectionId = connection.ConnectionId;
```
davidfowl: Add a comment explaining why it needs to be set (if it's null). Should we just check for null?

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -30,33 +30,37 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
...
-                // GET /{path}
-                await ExecuteEndpointAsync(context, socketDelegate, options);
-            }
-            else
-            {
-                context.Response.StatusCode = StatusCodes.Status405MethodNotAllowed;
+                if (HttpMethods.IsOptions(context.Request.Method))
+                {
+                    // OPTIONS /{path}
+                    await ProcessNegotiate(context, options);
```
davidfowl: We should update the connection id here and log as well.

BrennanConroy: So should we change the property name in our logs, or pass it here?

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -364,6 +374,11 @@ private static string GetNegotiatePayload(string connectionId, HttpSocketOptions
             return sb.ToString();
         }
 
+        private static string GetConnectionId(HttpContext context)
```
davidfowl:
```C#
private static string GetConnectionId(HttpContext context) => context.Request.Query["id"];
```

PR #674, opened 2017-07-20
Enabling binary protocols over text transports

moozzyk:


src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -201,6 +224,8 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                 {
                     while (connection.Input.TryRead(out var buffer))
                     {
+                        buffer = connection.DataEncoder.Decode(buffer);
```
davidfowl: This isn't right. This needs to be per decoded message inside of TryParseMessages

moozzyk: I don't want to couple encoding with protocol. Also, if you encode each message separately how would you know the boundary between encoded messages - the length prefix will be encoded. Would you introduce another length prefix for the encoded message? (We may still need to do that when pipelines are introduced though because we will not know where the message ends). 
Maybe the protocol should parse just a single message from a stream.

PR #675, opened 2017-07-20
React to AuthZ change

HaoK:
@davidfowl is there an appropriate resource to use for SignalR or is null sufficient?


moozzyk:
@HaoK - please open an issue to track this.

PR #677, opened 2017-07-21
Enabling binary protocols over text transports

moozzyk:
The joy of sending PRs on top of non-dev branches...

It was https://github.com/aspnet/SignalR/pull/674 previously which was closed because the branch was deleted. Reopening for discussion.


moozzyk:
:up::date:

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -100,6 +107,20 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             }
         }
 
+        IDataEncoder GetDataEncoder(TransferMode requestedTransferMode, TransferMode actualTransferMode)
```
davidfowl: private?

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
```diff
@@ -100,6 +107,20 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
...
+        {
+            if (requestedTransferMode == TransferMode.Binary && actualTransferMode == TransferMode.Text)
+            {
+                // This is for instance for SSE which is a Text protocol and the user wants to use a binary
+                // protocol so we need to encode messages.
+                return new Base64Encoder();
```
davidfowl: Should these be static since they are stateless?

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -104,10 +111,16 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                 {
                     if (NegotiationProtocol.TryParseMessage(buffer, out var negotiationMessage))
                     {
-                        // Resolve the Hub Protocol for the connection and store it in metadata
-                        // Other components, outside the Hub, may need to know what protocol is in use
-                        // for a particular connection, so we store it here.
-                        connection.Protocol = _protocolResolver.GetProtocol(negotiationMessage.Protocol, connection);
+                        var protocol = _protocolResolver.GetProtocol(negotiationMessage.Protocol, connection);
+
+                        var transportCapabilities = connection.Features.Get<IConnectionTransportFeature>()?.TransportCapabilities
```
davidfowl: I still hate these throw expressions.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -122,9 +124,8 @@ public Task<string> SendInvocationAsync(string methodName, params object[] args)
         public async Task<string> SendInvocationAsync(string methodName, bool nonBlocking, params object[] args)
         {
             var invocationId = GetInvocationId();
-            var payload = _protocol.WriteToArray(new InvocationMessage(invocationId, nonBlocking, methodName, args));
 
-            await Application.Out.WriteAsync(payload);
+            await Application.Out.WriteAsync(_protocolReaderWriter.WriteMessage(new InvocationMessage(invocationId, nonBlocking, methodName, args)));
```
davidfowl: Keep the 2 lines, this line is too long.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -320,5 +324,57 @@ public class HubConnectionProtocolTests
...
+            try
+            {
+                await hubConnection.StartAsync().OrTimeout();
+                await hubConnection.SendAsync("MyMethod", 42).OrTimeout();
+
+                await connection.ReadSentTextMessageAsync().OrTimeout();
+                var invokeMessage = await connection.ReadSentTextMessageAsync().OrTimeout();
+
+                // this throws if the message is not a valid base64 string
+                Convert.FromBase64String(invokeMessage);
```
davidfowl: Verify the message.

src/Microsoft.AspNetCore.SignalR.Client/HubConnection.cs
moozzyk: They implement an interface so they can either be singletons or some code changes will be required

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
moozzyk: I could add a separate test but it is not our code that throws but Convert itself so not sure what the value would be. The comment was supposed to only explain why this test does not have any `Assert`s

PR #694, opened 2017-08-07
Fix SSE hang on close

### LINKED ISSUE #526: Fix and unskip flaky test: SSETransportStopsSendAndReceiveLoopsWhenTransportStopped

BrennanConroy:
#526 
@moozzyk @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -136,6 +136,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 readCancellationRegistration.Dispose();
                 _transportCts.Cancel();
 
+                if (System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription.StartsWith(".NET Framework"))
+                {
+                    stream.Dispose();
+                }
+
                 await writeTask;
                 _logger.ReceiveStopped(_connectionId);
             }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -21,6 +21,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly ILogger _logger;
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
         private readonly ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
+        private readonly PipeFactory _pipelineFactory = new PipeFactory();
         private string _connectionId;
 
         private Channel<byte[], SendMessage> _application;
@@ -80,7 +81,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
             var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
 
             var stream = await response.Content.ReadAsStreamAsync();
-            var pipelineReader = stream.AsPipelineReader(cancellationToken);
+            var pipe = _pipelineFactory.Create();
+
+            var writeTask = stream.CopyToEndAsync(pipe.Writer, cancellationToken);
+
+            var pipelineReader = pipe.Reader;
             var readCancellationRegistration = cancellationToken.Register(
                 reader => ((IPipeReader)reader).CancelPendingRead(), pipelineReader);
             try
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -80,7 +81,14 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
             var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
 
             var stream = await response.Content.ReadAsStreamAsync();
-            var pipelineReader = stream.AsPipelineReader(cancellationToken);
+            var pipe = _pipelineFactory.Create();
+
+            _ = Task.Run(async () =>
```
moozzyk: Why is this needed?

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -132,6 +137,16 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
                 _transportCts.Cancel();
                 stream.Dispose();
                 _logger.ReceiveStopped(_connectionId);
+                pipe.Writer.Complete();
+
+                try
+                {
+                    await writeTask;
```
BrennanConroy: Should we handle cases where this task takes too long/never finishes, or is that not something that happens except in test code

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -140,6 +137,16 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
...
+                }
+                catch (Exception ex)
```
moozzyk: I looked at code [here](https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines.Extensions/StreamExtensions.cs#L71-L83) and it seems that exception handling is actually not needed. Also, the writer will be completed so it seems that the only thing to do is to await the task...

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -21,6 +21,7 @@ public class ServerSentEventsTransport : ITransport
         private readonly ILogger _logger;
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
         private readonly ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
+        private readonly PipeFactory _pipelineFactory = new PipeFactory();
```
davidfowl: Who disposes this?

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -132,6 +137,16 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
                 _transportCts.Cancel();
                 stream.Dispose();
                 _logger.ReceiveStopped(_connectionId);
+                pipe.Writer.Complete();
```
davidfowl: Seems likt CopyToEndAsync already completes the writer

davidfowl: https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines.Extensions/StreamExtensions.cs#L71-L83

BrennanConroy: Is it safe to assume that the stream will end and complete will actually be called? In our tests the wait for `writeTask` hangs because the stream is constantly being written to.

davidfowl: I see, so if we abort, this will complete the pipe writer

muratg:
@BrennanConroy *ping*

BrennanConroy:
Blocked on 2.1 runtime

muratg:
OK thanks. @natemcmaster will be looking into bringing 2.1 runtime soon.

davidfowl:
@BrennanConroy whats the deal with this?

BrennanConroy:
Made a different PR https://github.com/aspnet/SignalR/pull/1279
Not happy with either one though

davidfowl:
Ok closing this

PR #695, opened 2017-08-07
Use CloseOutputAsync to close client transport WebSocket

### LINKED ISSUE #568: Revisit how we close the webSockets transport on the client

moozzyk:
Fixes #568

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
@@ -67,6 +67,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             // https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Core/Owin/WebSockets/WebSocketHandler.cs#L248-L251
             Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
             {
+                _webSocket.Dispose();
                 _logger.TransportStopped(_connectionId, t.Exception?.InnerException);
                _application.Out.TryComplete(t.IsFaulted ? t.Exception.InnerException : null);
                return t;
@@ -221,7 +222,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
             _logger.TransportStopping(_connectionId);
 
             await CloseWebSocket();
-            _webSocket.Dispose();
 
             try
             {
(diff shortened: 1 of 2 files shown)


davidfowl:
Whats the deal with the build?

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -242,7 +242,9 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
                 if (_webSocket.State != WebSocketState.Closed)
                 {
                     _logger.ClosingWebSocket(_connectionId);
-                    await _webSocket.CloseAsync(WebSocketCloseStatus.NormalClosure, null, CancellationToken.None);
+                    await _webSocket.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, null, CancellationToken.None);
+                    // shutdown the transport after a timeout in case the server does not send close frame
+                    _transportCts.CancelAfter(TimeSpan.FromSeconds(5));
```
davidfowl: Why not pass this token into CloseOutputAsync?

moozzyk:
IntegrationTesting - all repos where it was used broke due to using an outdated version. AFAICT @JunTaoLuo has [fixed](47d14e5d4545c475f787b8a0efd4b6acd1a13ac0) it.

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
moozzyk: Actually, if the token is cancelled due to an exception in Send we would never try to close the websocket gracefully

moozzyk:
Need to investigate build failures...

BrennanConroy:
Here is a branch I made last month https://github.com/aspnet/SignalR/compare/brecon/websocketsClose
Don't remember why I moved the dispose but perhaps there is a good reason?

moozzyk:
The reason to move Dispose is to prevent from disposing immediately after CloseOutputAsync is closed - this is why CI was not happy. It is a race - it happens to pass locally.

moozzyk:
:up::date:

PR #699, opened 2017-08-09
Integrating MsgPack support in TS client

moozzyk:


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
```diff
@@ -28,8 +28,14 @@ export interface NegotiationMessage {
...
 }
 
+export const enum ProtocolType {
+    Text = 1,
+    Binary
+}
+
 export interface IHubProtocol {
     name(): string;
+    type(): ProtocolType;
```
davidfowl: You can do real properties in typescript.

moozzyk: I did not want them to be settable.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -1,28 +1,41 @@
...
+}
+
 export interface ITransport {
-    connect(url: string): Promise<void>;
+    connect(url: string, requestedTransferMode: TransferMode): Promise<void>;
     send(data: any): Promise<void>;
     stop(): void;
     onDataReceived: DataReceived;
     onClosed: TransportClosed;
+    transferMode(): TransferMode;
```
moozzyk: I am thinking about removing this and just returning Promise<TransferMode> from connect. Same for connection.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
moozzyk: (from JS that is...)

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
```diff
@@ -112,10 +112,10 @@ describe("Connection", () => {
             }
         } as IHttpConnectionOptions;
 
-        var connection = new HttpConnection("http://tempuri.org", options);
+        let connection = new HttpConnection("http://tempuri.org", options);
```
muratg: `const connection` instead? (here and elsewhere)

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
davidfowl: Cant you do get only properties?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpConnection.ts
```diff
@@ -32,18 +32,18 @@ export class HttpConnection implements IConnection {
         this.options = options;
     }
 
-    async start(): Promise<void> {
+    async start(requestedTransferMode: TransferMode): Promise<void> {
```
davidfowl: This feels wrong.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
moozzyk: I don't think so - this is not an object literal so I don't think it makes any difference. Conceptually this object is not a const since it exposes `onClosed` and `onDataReceived` properties which are settable.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpConnection.ts
moozzyk: What's the alternative? Features? Adding Features felt to heavy for this. We can do this but it would come separately.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
moozzyk: You can do `readonly` properties but they are only `readonly` in TypeScript but not in Javascript.

moozzyk: on the other hand you could replace the entire function in JS so meh - can change to a property (same for `name()`)

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
muratg: I don't mind either way, but `const` only makes the variable constant (you can't assign something else to the name), not what the variable is referencing. (See `readonly` for that.)

moozzyk: I am leaving it as it is then. If we decide to change it to const we would probably have to change almost all variables which is out of scope for this PR. I was just annoyed by mixing `let` and `var` without any reason.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol.ts
moozzyk: Done

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
moozzyk: Done()

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -120,6 +120,14 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                             ? (IDataEncoder)Base64Encoder
                             : PassThroughEncoder;
 
+                        var transferMode = connection.Features.Get<ITransferModeFeature>() ??
```
moozzyk: I know understand how this sneaked in - https://github.com/aspnet/SignalR/pull/711

moozzyk:
:up::date: - Added features to IConnection

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -196,3 +210,38 @@ export class HubConnection {
...
+        this.wrappedProtocol = protocol;
+        this.name = this.wrappedProtocol.name;
+        this.type = ProtocolType.Text;
+    }
+
+    readonly name: string;
+    readonly type: ProtocolType;
+
+    parseMessages(input: any): HubMessage[] {
+        // atob/btoa are browsers APIs but they can be polyfilled. If this becomes problematic we can use
```
davidfowl: File a bug for that. I don't want too many untracked browser deps

moozzyk: Are we going to do anything about this though? I did try base64-js package and reverted the change since it added 2.2 KB to the library (minified) which is wasteful on browsers since the functionality is built-in. If you hit this on node you would need to polyfill. Unless the bug is supposed to be a doc bug which will be turned into a an article about dependencies a user needs to polyfill when running on node.

davidfowl: I can think of a few things we might do here:
- Polyfill on nodejs - I hate this option
- Dependency injection
- Build a nodejs version with different sources

moozzyk: ...we could also not support binary over SSE on node. I think XmlHttpRequest and WebSockets are more important - once we decide about these the atob/btoa replacement will just follow the suit.

moozzyk: https://github.com/aspnet/SignalR/issues/716

Issue #683, opened 2017-07-31
Add HubLifetimeManager.InvokeConnectionsAsync()


KPixel:
Hello

Can we add the following method to the class HubLifetimeManager<THub>?

`Task InvokeConnectionsAsync(IEnumerable<string> connectionIds, string methodName, object[] args);
`

This will enable dynamic "grouping" where the application decides to message connections arbitrarily. We could use the InvokeConnectionAsync() but this one can be far more efficient.

Here is an example of how it could be used (INode is like a tree node which recursively find all attached connections):
```
        public static async Task InvokeNodeAsync<THub, T>(this HubLifetimeManager<THub> hub, INode<T> node, string methodName, object[] args) where T : INode<T>
        {
            var connectionIds = await node.GetConnectionIdsAsync();
            await hub.InvokeConnectionsAsync(connectionIds, methodName, args);
        }
```

Some other practical scenarios that this API will enable:

- Noisy group: Even in a "static" group, we could choose to dynamically exclude some connections that already received another message recently (like a notification system where if you haven't read the previous notification, we don't send another one)
- Too many groups: In a large forum where you auto-subscribe to every thread you participate in, you can be part of so many groups, that it is impractical to load them when you connect. Another variant of this scenario is to want an implementation where the group has a list of (users who have a list of) connections...
- Undefinable "group": An app using geo-localisation may allow a user to send a message to everyone who is X kilometers close to a point of interest. So, the app will dynamically find all the devices in that area, and message them.

I am giving these scenarios because they may suggest other ways to improve SignalR.

davidfowl:
Sure this is part of work here to round out the APIs https://github.com/aspnet/SignalR/issues/394.

KPixel:
Indeed.

If we take it one step further, I don't think that ALL Hubs should be forced to implement the current Group(s) & User(s) APIs. I think they are opinionated ways to group connectionIds and should be optional, so each application can decide what a "group" is for them.

However, I understand that people coming from the previous version of SignalR will expect these familiar APIs. So, hopefully, you can figure out a balanced design.

davidfowl:
This is already true today if you ignore Add/Remove Group on the lifetime manager. You can make the group logic do whatever you wanted to without a specific InvokeConnectionsAsync call. It's actually a bit more flexible than what you suggest since you can map that to anything really. 

[...]

PR #700, opened 2017-08-09
Client Subset - AllExcept

mikaelm12:
Wanted some initial review + wanted to get some discussion on how to do this for the RedisHubLifetimeManager. Discussed this with @moozzyk yesterday a little but I figured it would be good to put it out here.

### DIFF
--- a/samples/ChatSample/PresenceHubLifetimeManager.cs
+++ b/samples/ChatSample/PresenceHubLifetimeManager.cs
@@ -142,7 +142,7 @@ namespace ChatSample
             return _wrappedHubLifetimeManager.InvokeAllAsync(methodName, args);
         }
 
-        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyCollection<string> excludedIds)
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
         {
             return _wrappedHubLifetimeManager.InvokeAllExceptAsync(methodName, args, excludedIds);
         }
--- a/samples/ChatSample/PresenceHubLifetimeManager.cs
+++ b/samples/ChatSample/PresenceHubLifetimeManager.cs
@@ -2,12 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.AspNetCore.SignalR.Redis;
-using System.Linq;
 
 namespace ChatSample
 {
@@ -139,6 +140,11 @@ namespace ChatSample
         public override Task InvokeAllAsync(string methodName, object[] args)
         {
             return _wrappedHubLifetimeManager.InvokeAllAsync(methodName, args);
+        }
+
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyCollection<string> excludedIds)
+        {
+            return _wrappedHubLifetimeManager.InvokeAllExceptAsync(methodName, args, excludedIds);
         }
 
         public override Task InvokeConnectionAsync(string connectionId, string methodName, object[] args)
--- a/samples/SocketsSample/Hubs/Chat.cs
+++ b/samples/SocketsSample/Hubs/Chat.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using System.Collections.Generic;
 using Microsoft.AspNetCore.SignalR;
 
 namespace SocketsSample.Hubs
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -82,11 +82,12 @@ namespace Microsoft.AspNetCore.SignalR.Redis
                 previousBroadcastTask = Task.WhenAll(tasks);
             });
 
+            var allExceptTask = Task.CompletedTask;
             channelName = _channelNamePrefix + ".AllExcept";
             _logger.LogInformation("Subscribing to channel: {channel}", channelName);
             _bus.Subscribe(channelName, async (c, data) =>
             {
-                await previousBroadcastTask;
+                await allExceptTask;
 
                 _logger.LogTrace("Received message from redis channel {channel}", channelName);
 
@@ -105,7 +106,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
                     }
                 }
 
-                previousBroadcastTask = Task.WhenAll(tasks);
+                allExceptTask = Task.WhenAll(tasks);
             });
         }
 
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -93,7 +93,6 @@ namespace Microsoft.AspNetCore.SignalR.Redis
                 var message = DeserializeMessage<RedisExcludeClientMessage>(data);
                 var invocationMessage = (InvocationMessage)message.HubMessage;
                 var excludedIds = message.ExcludedIds;
-                var excludedIdsSet = new HashSet<string>(excludedIds);
 
                 //// TODO: This isn't going to work when we allow JsonSerializer customization or add Protobuf
 
@@ -146,13 +145,13 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             return PublishAsync(_channelNamePrefix + ".user." + userId, message);
         }
 
-        private async Task PublishAsync(string channel, object args)
+        private async Task PublishAsync(string channel, HubMessage hubMessage)
         {
             byte[] payload;
             using (var stream = new MemoryStream())
             using (var writer = new JsonTextWriter(new StreamWriter(stream)))
             {
-                _serializer.Serialize(writer, args);
+                _serializer.Serialize(writer, hubMessage);
                 await writer.FlushAsync();
                 payload = stream.ToArray();
             }
@@ -394,12 +393,13 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             }
         }
 
-        private class RedisExcludeClientMessage
+        private class RedisExcludeClientMessage : HubMessage
         {
             public HubMessage HubMessage;
             public IReadOnlyList<string> ExcludedIds;
 
-            public RedisExcludeClientMessage(HubMessage message, IReadOnlyList<string> ids)
+            public RedisExcludeClientMessage(HubMessage message, IReadOnlyList<string> ids) :
+                base(message.InvocationId)
             {
                 HubMessage = message;
                 ExcludedIds = ids;
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -94,7 +94,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
                 var message = DeserializeMessage<RedisExcludeClientsMessage>(data);
                 var excludedIds = message.ExcludedIds;
 
-                //// TODO: This isn't going to work when we allow JsonSerializer customization or add Protobuf
+                // TODO: This isn't going to work when we allow JsonSerializer customization or add Protobuf
 
                 var tasks = new List<Task>(_connections.Count);
 
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -118,7 +118,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
 
         public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
         {
-            var message = new RedisExcludeClientsMessage(GetInvocationId(), nonBlocking: true, target: methodName, excludedIds: excludedIds ,arguments: args);
+            var message = new RedisExcludeClientsMessage(GetInvocationId(), nonBlocking: true, target: methodName, excludedIds: excludedIds, arguments: args);
             return PublishAsync(_channelNamePrefix + ".AllExcept", message);
         }
 
@@ -395,7 +395,8 @@ namespace Microsoft.AspNetCore.SignalR.Redis
         {
             public IReadOnlyList<string> ExcludedIds;
 
-            public RedisExcludeClientsMessage(string invocationId, bool nonBlocking, string target, IReadOnlyList<string> excludedIds, params object[] arguments) : base(invocationId, nonBlocking, target, arguments)
+            public RedisExcludeClientsMessage(string invocationId, bool nonBlocking, string target, IReadOnlyList<string> excludedIds, params object[] arguments)
+                : base(invocationId, nonBlocking, target, arguments)
             {
                 ExcludedIds = excludedIds;
             }
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -26,6 +26,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
         private readonly ISubscriber _bus;
         private readonly ILogger _logger;
         private readonly RedisOptions _options;
+        private readonly string _channelNamePrefix = typeof(THub).FullName;
 
         // This serializer is ONLY use to transmit the data through redis, it has no connection to the serializer used on each connection.
         private readonly JsonSerializer _serializer = new JsonSerializer
@@ -60,7 +61,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
 
             var previousBroadcastTask = Task.CompletedTask;
 
-            var channelName = typeof(THub).FullName;
+            var channelName = _channelNamePrefix;
             _logger.LogInformation("Subscribing to channel: {channel}", channelName);
             _bus.Subscribe(channelName, async (c, data) =>
             {
@@ -68,7 +69,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
 
                 _logger.LogTrace("Received message from redis channel {channel}", channelName);
 
-                var message = DeserializeMessage(data);
+                var message = DeserializeMessage<HubMessage>(data);
 
                 // TODO: This isn't going to work when we allow JsonSerializer customization or add Protobuf
                 var tasks = new List<Task>(_connections.Count);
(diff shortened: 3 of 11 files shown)


src/Microsoft.AspNetCore.SignalR/IHubClients.cs
```diff
@@ -1,12 +1,16 @@
...
 
+using System.Collections.Generic;
+
 namespace Microsoft.AspNetCore.SignalR
 {
     public interface IHubClients
     {
         IClientProxy All { get; }
 
+        IClientProxy AllExcept(List<string> excludedIds);
```
moozzyk: `params string[]` as per: https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Core/Hubs/IHubConnectionContext.cs#L14

samples/SocketsSample/Hubs/Chat.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using System.Collections.Generic;
```
moozzyk: needed?

samples/ChatSample/PresenceHubLifetimeManager.cs
```diff
@@ -141,6 +142,11 @@ public override Task InvokeAllAsync(string methodName, object[] args)
             return _wrappedHubLifetimeManager.InvokeAllAsync(methodName, args);
         }
 
+        public override Task InvokeAllExceptAsync(List<string> excludedIds, string methodName, object[] args)
+        {
+            return _wrappedHubLifetimeManager.InvokeAllExceptAsync( excludedIds, methodName, args);
```
moozzyk: format

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -140,6 +140,14 @@ private string GetInvocationId()
             return invocationId.ToString();
         }
 
+        public override Task InvokeAllExceptAsync(List<string> excludedIds, string methodName, object[] args)
+        {
+            return InvokeAllWhere(methodName, args, connection =>
+            {
+                return !excludedIds.Contains(connection.ConnectionId);
```
moozzyk: Not sure if this is for now but this feels slow

BrennanConroy:
For redis; servers can subscribe to a new channel `typeof(THub).FullName + ".AllExcept"` and when AllExcept is called we can put the list of connections in the message and then when the servers receive it they could parse the message for the connections...

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -476,6 +476,51 @@ public void HubsCannotHaveOverloadedMethods()
             }
         }
 
+        [Fact]
+        public async Task SendToAllExcept()
+        {
+            var serviceProvider = CreateServiceProvider();
+
+            dynamic endPoint = serviceProvider.GetService(GetEndPointType(typeof(MethodHub)));
```
BrennanConroy: `var endPoint = serviceProvider.GetService<HubEndPoint<MethodHub>>();`

KPixel:
Great to see this feature coming.

I think its implementation should be the reverse of `InvokeConnectionsAsync()`. I discussed that one here: #683. Can you please include it?

I also like the idea of a `HubMessage` wrapper that takes the list of `connectionIds`. We can reuse it there too.

Can you change the type of `excludedIds` to `ICollection<string>`? It will enable using `HashSet`, which makes sense here.

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
anurse: Yeah, using a `HashSet<string>` for `excludedIds` would be faster, since it has a O(1) contains test

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -340,6 +339,11 @@ private HubMessage DeserializeMessage(RedisValue data)
             return message;
         }
 
+        public override Task InvokeAllExceptAsync(List<string> excludedIds, string methodName, object[] args)
+        {
+            throw new NotImplementedException();
```
anurse: Assuming this is part of the WIP-ness of this PR ;)

src/Microsoft.AspNetCore.SignalR/HubContext.cs
```diff
@@ -20,6 +22,11 @@ public HubContext(HubLifetimeManager<THub> lifetimeManager)
 
         public virtual IGroupManager Groups { get; }
 
+        public IClientProxy AllExcept(List<string> excludedIds)
```
anurse: Should use `IReadOnlyList<string>` or `IEnumerable<string>` here. We never plan to write to this list, and we don't care exactly how it's implemented. In fact, if our implementations are going to use `HashSet<string>`, `IEnumerable` would be better since we're going to need to convert it to a different type of collection anyway.

anurse: Also, should have a `params string[] excludedIds` overload (possibly via an extension method) so you can call `.AllExcept(connection1, connection2, connection3)...`

src/Microsoft.AspNetCore.SignalR/HubLifetimeManager.cs
```diff
@@ -13,6 +14,8 @@ public abstract class HubLifetimeManager<THub>
 
         public abstract Task InvokeAllAsync(string methodName, object[] args);
 
+        public abstract Task InvokeAllExceptAsync(List<string> excludedIds, string methodName, object[] args);
```
anurse: `IEnumerable<string>`/`IReadOnlyList<string>`

src/Microsoft.AspNetCore.SignalR/IHubClients.cs
anurse: I don't think having `params` on a public interface API is a great plan actually... I'd rather use `IEnumerable` or `IReadOnlyList` and have a `params` variant implemented in an extension method. `params` forces you to allocate an array, if you don't already have one (until we get `params IEnumerable` someday...)

src/Microsoft.AspNetCore.SignalR/Proxies.cs
```diff
@@ -54,12 +55,28 @@ public Task InvokeAsync(string method, params object[] args)
         }
     }
 
+    public class AllClientsExceptProxy<THub> : IClientProxy
+    {
+        private readonly HubLifetimeManager<THub> _lifetimeManager;
+        private readonly List<string> _excludedIds;
+
+        public AllClientsExceptProxy(HubLifetimeManager<THub> lifetimeManager, List<string> excludedIds)
```
anurse: `IEnumerable<string>`/`IReadOnlyList<string>`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: Yeah lol. Working this out now 😄

src/Microsoft.AspNetCore.SignalR/IHubClients.cs
moozzyk: I thought about `IEnumerable` but there is always a problem with lazy evaluation. `IReadOnlyList` is a good option. Agree that `params` is a little weird (you don't know connectionIds upfront but you need to know the count) but it saves you a few keystrokes if you want to exclude just one connectionId you store in a variable (e.g. `.Others`).

src/Microsoft.AspNetCore.SignalR/Proxies.cs
KPixel: Why not `ICollection<string>`? It is the lowest level interface that offers `Contains()`.
Performance-wise, the dev can choose between sending a List<> or a Set<> or something else.

anurse: We never need to write to the collection, so we should avoid using a writable interface, hence `IReadOnlyCollection` or `IReadOnlyList` would be a better option.

In general, we want to lead users towards good performance without them having to know how our code is implemented, so we try to avoid leading users to a path where they have to consider the performance of collection objects they give us. It's our job to make processing the collection fast, not the user's.

KPixel: `IReadOnlyCollection` and `IReadOnlyList` do not have a `Contains()` method, which is essential to the implementation.

And if we care about performance, we must allow passing an `ISet`-based collection.

So, I think it is better to take a (potentially) writable interface that performs well than take a read-only collection that SignalR must first convert to something like `HashSet`.

That interface ends up being `ICollection` because we also want to allow `IList` for convenience. Otherwise, the overload method taking `params string[]` cannot pass its parameter directly.

anurse: The point is moot until we have benchmarks to properly understand the performance characteristics in practical scenarios. Given that the most common scenario we've seen for this involves a very small number of entries in the list, we may just spin our wheels here for a while with almost no gain :).

The original comment is just about using an interface over the concrete `List<T>` type currently used. `IReadOnlyCollection<T>`/`IReadOnlyList<T>` are the "correct" expressions of what this public API requires, which is generally our first stop when building an API. Performance decisions to compromise that have to be data-driven and based on actual usage.

samples/SocketsSample/Hubs/Chat.cs
mikaelm12: Nope!

mikaelm12:
Tested this functionality with the presence chat sample and it looks pretty good 😄

mikaelm12:
⬆️ 📅

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,43 +81,78 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
             var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
 
-            return PublishAsync(typeof(THub).FullName, message);
+            return PublishAsync(_channelNamePrefix, message);
+        }
+
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyCollection<string> excludedIds)
+        {
+            var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
+            var data = new object[] { message, excludedIds };
```
BrennanConroy: This looks... What was wrong with making a class that wraps InvocationMessage and adds a IReadOnlyCollection?

mikaelm12: This was a WIP hack that I didn't update 😄

src/Microsoft.AspNetCore.SignalR/HubContext.cs
```diff
@@ -20,6 +22,11 @@ public HubContext(HubLifetimeManager<THub> lifetimeManager)
 
         public virtual IGroupManager Groups { get; }
 
+        public IClientProxy AllExcept(IReadOnlyList<string> excludedIds)
```
mikaelm12: Should all be `IReadOnlyList` now

mikaelm12:
🆙 📅

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,43 +81,78 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+            _bus.Subscribe(channelName, async (c, data) =>
+            {
+                await previousBroadcastTask;
+
+                _logger.LogTrace("Received message from redis channel {channel}", channelName);
+
+                var message = DeserializeMessage<RedisExcludeClientMessage>(data);
+                var invocationMessage = (InvocationMessage)message.HubMessage;
+                var excludedIds = message.ExcludedIds;
+                var excludedIdsSet = new HashSet<string>(excludedIds);
```
BrennanConroy: This isn't used

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,43 +81,78 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
         public override Task InvokeUserAsync(string userId, string methodName, object[] args)
         {
             var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
 
-            return PublishAsync(typeof(THub).FullName + ".user." + userId, message);
+            return PublishAsync(_channelNamePrefix + ".user." + userId, message);
         }
 
-        private async Task PublishAsync(string channel, HubMessage hubMessage)
+        private async Task PublishAsync(string channel, object args)
```
BrennanConroy: Revert

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -362,6 +394,17 @@ public override void WriteLine(string value)
             }
         }
 
+        private class RedisExcludeClientMessage
```
BrennanConroy: This should be able to inherit from InvocationMessage

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -140,6 +141,15 @@ private string GetInvocationId()
             return invocationId.ToString();
         }
 
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
+        {
+            return InvokeAllWhere(methodName, args, connection =>
+            {
+                var excludedIdsSet = new HashSet<string>(excludedIds);
```
BrennanConroy: Should just get rid of this and call `.Contains` on `excludedIds`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: Would it make more sense to inherit from HubMessage?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,34 +81,68 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
             var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
 
-            return PublishAsync(typeof(THub).FullName, message);
+            return PublishAsync(_channelNamePrefix, message);
+        }
+
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
+        {
+            var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
+            var redisExcludeMessage = new RedisExcludeClientMessage(message, excludedIds);
```
BrennanConroy: The point of making it inherit from `InvocationMessage` was that you could avoid newing up two objects, and just do a single `RedisExcludeClientMessage`

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -476,6 +476,50 @@ public void HubsCannotHaveOverloadedMethods()
...
+
+                await firstClient.SendInvocationAsync("SendToAllExcept", "To second", excludeThirdClientId).OrTimeout();
+                await firstClient.SendInvocationAsync("SendToAllExcept", "To third", excludeSecondClientId).OrTimeout();
+
+                var secondClientResult = await secondClient.ReadAsync().OrTimeout();
+                var invocation = Assert.IsType<InvocationMessage>(secondClientResult);
+                Assert.Equal("Send", invocation.Target);
+                Assert.Equal("To second", invocation.Arguments[0]);
+
+                var thirdClientResult = await thirdClient.ReadAsync().OrTimeout(); ;
```
BrennanConroy: extra semi-colon

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -362,6 +391,16 @@ public override void WriteLine(string value)
             }
         }
 
+        public class RedisExcludeClientsMessage : InvocationMessage
+        {
+            public IReadOnlyList<string> ExcludedIds;
+
+            public RedisExcludeClientsMessage(string invocationId, bool nonBlocking, string target, IReadOnlyList<string> excludedIds, params object[] arguments) : base(invocationId, nonBlocking, target, arguments)
```
BrennanConroy: Should probably add a newline

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,34 +81,66 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
         {
             var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, arguments: args);
 
-            return PublishAsync(typeof(THub).FullName, message);
+            return PublishAsync(_channelNamePrefix, message);
+        }
+
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
+        {
+            var message = new RedisExcludeClientsMessage(GetInvocationId(), nonBlocking: true, target: methodName, excludedIds: excludedIds ,arguments: args);
```
BrennanConroy: formatting `excludedIds ,arguments:`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,34 +81,68 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+            {
+                await previousBroadcastTask;
+
+                _logger.LogTrace("Received message from redis channel {channel}", channelName);
+
+                var message = DeserializeMessage<RedisExcludeClientMessage>(data);
+                var invocationMessage = (InvocationMessage)message.HubMessage;
+                var excludedIds = message.ExcludedIds;
+
+                //// TODO: This isn't going to work when we allow JsonSerializer customization or add Protobuf
```
anurse: How is it not going to work? If we're just writing serialized data directly to the connection, then we're broken for MsgPack too, if one node is using JSON and the other is using MsgPack. If we're serializing separately for every connection, then it should work fine with JsonSerializer customization and Protobuf because each send uses the client's configured serializers.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: Hrm... feels like a a premature optimization, but I guess we can do it.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -362,6 +393,18 @@ public override void WriteLine(string value)
...
+            public HubMessage HubMessage;
+            public IReadOnlyList<string> ExcludedIds;
+
+            public RedisExcludeClientMessage(HubMessage message, IReadOnlyList<string> ids) :
+                base(message.InvocationId)
+            {
+                HubMessage = message;
+                ExcludedIds = ids;
+            }
+        }
```
anurse: need an extra line

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
anurse: Yeah, this will create a new HashSet for every check. If we decide we need a HashSet for optimization, we would want to initialize it outside the lambda so there's only one instance.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: I derped and commented on outdated stuff. Disregard

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,34 +81,66 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+            _bus.Subscribe(channelName, async (c, data) =>
+            {
+                await previousBroadcastTask;
+
+                _logger.LogTrace("Received message from redis channel {channel}", channelName);
+
+                var message = DeserializeMessage<RedisExcludeClientsMessage>(data);
+                var excludedIds = message.ExcludedIds;
+
+                //// TODO: This isn't going to work when we allow JsonSerializer customization or add Protobuf
```
anurse: Can you clarify how this is broken with JsonSerializer customization and Protobuf? I'd think it'd already be broken for MsgPack if that was the case. We should be re-serializing for each client in the current implementation, right? It's slower, but guarantees it's correct and should work fine with serializer customization.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,34 +81,66 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
 
                 previousBroadcastTask = Task.WhenAll(tasks);
             });
+
+            channelName = _channelNamePrefix + ".AllExcept";
+            _logger.LogInformation("Subscribing to channel: {channel}", channelName);
+            _bus.Subscribe(channelName, async (c, data) =>
+            {
+                await previousBroadcastTask;
```
moozzyk: Is this correct to use the same task as broadcast?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -80,34 +81,66 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+
+                foreach (var connection in _connections)
+                {
+                    if (!excludedIds.Contains(connection.ConnectionId))
+                    {
+                        tasks.Add(WriteAsync(connection, message));
+                    }
+                }
+
+                previousBroadcastTask = Task.WhenAll(tasks);
```
moozzyk: I think this is wrong/racey - you can replace a task that has not been finished with another task that is not finished. Sharing the same task between broadcast and .allexcept makes it even more racey

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: This comment is a copy pasta result. Not sure tbh

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: Oh yeah. This doesn't look good.

mikaelm12:
If there's no further discussion I'm gonna merge this eod

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
davidfowl: super nit: the comment has too many `////`

PR #703, opened 2017-08-09
Update protocol spec for Message Pack

moozzyk:


specs/HubProtocol.md
```diff
@@ -16,7 +16,7 @@ The SignalR Protocol requires the following attributes from the underlying trans
 
 ## Overview
 
-There are two encodings of the SignalR protocol: [JSON](http://www.json.org/) and [Protocol Buffers](https://developers.google.com/protocol-buffers/). Only one format can be used for the duration of a connection, and the format must be negotiated in advance (i.e. using a QueryString value, Header, or other indicator). However, each format shares a similar overall structure.
```
moozzyk: I am not sure if we want to remove protobuf from the spec or just move it somewhere - it still can be useful if we want to add a sample ProtoBufHubProtocol.

davidfowl: We shouldn't move it. Lets keep the text here all in a single place.

specs/HubProtocol.md
```diff
@@ -348,101 +363,176 @@ World;0:;
...
+| `System.Single`                                 | `Number`                     | `float`                   |
+| `System.Double`                                 | `Number`                     | `float`                   |
+| `System.Boolean`                                | `true` or `false`            | `true`, `false`           |
+| `System.String`                                 | `String`                     | `fixstr`, `str`           |
+| `System.Byte`[]                                 | `String` (Base64-encoded)    | `bin`                     |
+| `IEnumerable<T>`                                | `Array`                      | `bin`                     |
+| custom `enum`                                   | `Number`                     | `fixint`, `int`           |
+| custom `struct` or `class`                      | `Object`                     | `fixmap`, `map`           |
+
+Message Pack payloads are wrapped in an outer message framing described below.
```
mikaelm12: 😄

moozzyk:
:up::date:

specs/HubProtocol.md
```diff
@@ -16,15 +16,30 @@ The SignalR Protocol requires the following attributes from the underlying trans
...
+
+Example:
+
+```json
+{
+    "protocol": "messagepack"
+}
+```
+
+In order to perform a single invocation, the Caller follows the following basic flow:
```
davidfowl: It's unclear that this is the end of negotiation and the beginning of a new section.

specs/HubProtocol.md
```diff
@@ -348,8 +363,160 @@ World;0:;
 
 Note that the final frame still ends with the `;` terminator, and that since the body may contain `;`, newlines, etc., the length is specified in order to know exactly where the body ends.
 
+## MessagePack (MsgPack) encoding
+
+In the MsgPack Encoding of the SignalR Protocol, each Message is represented as a single MsgPack array containing items that correspond to properties of the given hub protocol message. The array items may be primitive values, arrays (e.g. method arguments) or objects (e.g. argument value). The first item in the array is the message type.
```
davidfowl: Hmm, so this is basically a length prefix (conceptually). The array size is the length prefix surrogate. Is this how you'd usually represent an object in message pack? If you convert this to JSON would it be any array instead of an object? Is that weird or normal? Did we save space?

specs/HubProtocol.md
```diff
@@ -348,8 +363,160 @@ World;0:;
...
+is decoded as follows:
+
+* `0x94` - 4-element array
+* `0x03` - `3` (Message Type - `Result` message)
+* `0xa3` - string of length 3 (Target)
+* `0x78` - `x`
+* `0x79` - `y`
+* `0x7a` - `z`
+* `0x01` - `1` (ResultKind - Error result)
+* `0xa5` - string of lenght 5
```
BrennanConroy: length

specs/HubProtocol.md
```diff
@@ -348,8 +363,160 @@ World;0:;
...
+is decoded as follows:
+
+* `0x94` - 4-element array
+* `0x03` - `3` (Message Type - `Result` message)
+* `0xa3` - string of length 3 (Target)
+* `0x78` - `x`
+* `0x79` - `y`
+* `0x7a` - `z`
+* `0x03` - `3` (ResultKind - Non-Void result)
+* `0x2a` - `42` (Result)
```
BrennanConroy: Result always going to be between 0 and 255?

specs/HubProtocol.md
moozzyk: Yes, this is one of the representations of objects in MsgPack. Previously, I had only values which worked well in C# but was hard to handle in JS (and also see this [comment](https://github.com/aspnet/SignalR/pull/587#issuecomment-311338633)). You can't convert this directly to object because you don't have fields names - so yes it will be an array. A different way of doing this is to send an object (i.e. including field names etc.) serialized to MsgPack format. I did it originally and the messages were much bigger than doing arrays - so there was not a real benefit of using message pack. I also don't think it is needed to send hub protocol messages as objects since we control the protocol.

specs/HubProtocol.md
moozzyk: Will fix.

specs/HubProtocol.md
moozzyk: no - result can be a primitive value, an array or on object (it's a method invocation result). The result is encoded using MsgPack rules. Describing how MsgPack encodes values is out of scope for this document by I added a link at the top of this section (https://github.com/msgpack/msgpack/blob/master/spec.md#formats).

specs/HubProtocol.md
davidfowl: So, how does versioning work? Always add array items to the end? Treat indexes as field names?

davidfowl: The interesting thing is that we could do the exact same thing with JSON for the exact same reasons, so why not do that too? Why not use arrays instead of objects with field names.

moozzyk: 1. Versioning is a different story (and not part of this change). The way I think it should work is that the Negotiation message contains the version and then we could have version-specific implementations of HubProtocol. This way a new version of protocol is completely independent of the old version.

2. We could but we will give up readability. Also requires a little bit more code on our side (especially in TS client).

davidfowl: > Versioning is a different story (and not part of this change). The way I think it should work is that the Negotiation message contains the version and then we could have version-specific implementations of HubProtocol. This way a new version of protocol is completely independent of the old version.

The benefit of the self describing format is that certain types of changes don't require a version bump (like additing properties).

> We could but we will give up readability. Also requires a little bit more code on our side (especially in TS client).

It would be trivial since we have a layer that turns the protocol format into messages.

BTW I'm not saying we should do this, I just wanted to understand the general approach of using arrays vs objects since msgpack supports both.

moozzyk: > The benefit of the self describing format is that certain types of changes don't require a version bump (like additing properties).
Agreed but one of the main objectives of the binary format was efficiency/small messages

> It would be trivial since we have a layer that turns the protocol format into messages.
Yes it wouldn't be hard.

> I just wanted to understand the general approach of using arrays vs objects since msgpack supports both.
I guess it felt natural to use objects when using JSON and it was easy to implement. I was kind of fixed on this approach when starting with MsgPack but then I realized using arrays will be more efficient so switched to this.

moozzyk:
:up:date:

Issue #704, opened 2017-08-10
Potential Null ref in RedisHublifetimeManager


mikaelm12:
Inverted boolean logic could result in a null ref in the RedisHubLifetimeManager 😦 
https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs#L277-L282

mikaelm12:
Fixed: https://github.com/aspnet/SignalR/pull/706

Issue #705, opened 2017-08-10
Redis Tests


mikaelm12:
We need tests for our Redis scenarios. Just found this issue https://github.com/aspnet/SignalR/issues/704 which slid by.
We need a testing strategy. Redis needs some love

muratg:
We have some tests. Closing this bug.

PR #706, opened 2017-08-10
Fix potential null ref in RedisHublifetimeManager RemoveGroup

mikaelm12:
We should be making sure the connection is not null before trying to remove it from the group. 
issue: https://github.com/aspnet/SignalR/issues/704


mikaelm12:
No tests because we don't have tests for Redis 😞 https://github.com/aspnet/SignalR/issues/705

PR #710, opened 2017-08-10
Length prefixing base64 encoded messages

moozzyk:
... in preparation for pipeline conversion


moozzyk:
There will be much more work required to handle cases where we receive partial payloads. I believe this will require changing some APIs to take buffers instead of `byte[]` and buffering in TS.

PR #711, opened 2017-08-11
Enabling tests which would have prevented a bug

moozzyk:


test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -168,6 +180,12 @@ public EndToEndTests(ServerFixture<Startup> serverFixture, ITestOutputHelper out
                     logger.LogInformation("Disposed Connection");
                 }
             }
+
+            bool base64Encoded(TransferMode transferMode, IConnection connection)
```
davidfowl: nit: casing this isn't typescript 😉

moozzyk: LOL

moozzyk: Fixed.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -301,5 +319,14 @@ public static IEnumerable<object[]> MessageSizesData
                 new object[] { TransportType.ServerSentEvents },
                 new object[] { TransportType.LongPolling }
             };
+
+        public static IEnumerable<object[]> TransportTypesXTransferModes()
```
muratg: XTransfer?

moozzyk: read: TransportTypes times TransferModes

davidfowl: Maybe call it And

moozzyk: The thing is that this is not *And* - this is Cartesian product. We also have [one more instance](https://github.com/aspnet/SignalR/blob/3a1d4c5dd66cdcccbc61dc8e530784c8b79b9b1b/test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs#L220) of this so if we want to change the convention it should be in both places.

davidfowl: I know but I don't think it matters.

moozzyk: ok - I can change isn't that important..

muratg: Thanks for the Xplanation :)

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -120,6 +120,14 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                             ? (IDataEncoder)Base64Encoder
                             : PassThroughEncoder;
 
+                        var transferMode = connection.Features.Get<ITransferModeFeature>() ??
```
davidfowl: nit: rename to transferModeFeature

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -120,6 +120,14 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
+                            throw new InvalidOperationException("Unable to read transfer mode.");
+
+                        transferMode.TransferMode =
```
davidfowl: Where is this used? Or are we just updating it to be correct? I can't see who uses this thing.

moozzyk: https://github.com/aspnet/SignalR/blob/29f9c54b864d3f86522b15651acf3b6adc57c8be/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs#L181-L183
- I must have lost it in one of rebases/merges.

moozzyk:
Addressed feedback

PR #717, opened 2017-08-11
Length prefixing base64 encoded messages

moozzyk:
... in preparation for pipeline conversion

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Base64EncodedHubProtocol.spec.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Base64EncodedHubProtocol.spec.ts
@@ -1,0 +1,71 @@
+import { Base64EncodedHubProtocol } from "../Microsoft.AspNetCore.SignalR.Client.TS/Base64EncodedHubProtocol"
+import { IHubProtocol, HubMessage, ProtocolType } from "../Microsoft.AspNetCore.SignalR.Client.TS/IHubProtocol"
+
+class FakeHubProtocol implements IHubProtocol {
+    name: "fakehubprotocol";
+    type: ProtocolType;
+
+    parseMessages(input: any): HubMessage[] {
+        let s = "";
+
+        new Uint8Array(input).forEach((item: any) => {
+            s += String.fromCharCode(item);
+        });
+
+        return JSON.parse(s);
+    }
+
+    writeMessage(message: HubMessage): any {
+        let s = JSON.stringify(message);
+        let payload = new Uint8Array(s.length);
+        for (let i = 0; i < payload.length; i++) {
+            payload[i] = s.charCodeAt(i);
+        }
+        return payload;
+    }
+}
+
+describe("Base64EncodedHubProtocol", () => {
+    ([
+        ["ABC", new Error("Invalid payload.")],
+        ["3:ABC", new Error("Invalid payload.")],
+        [":;", new Error("Invalid length: ''")],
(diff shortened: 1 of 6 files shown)


moozzyk:
Resurrecting #710 after rebasing on dev.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -224,9 +224,18 @@ class Base64EncodedHubProtocol implements IHubProtocol {
...
 
     parseMessages(input: any): HubMessage[] {
+
+        // The format of the message is `size:message;`
+        let pos = input.indexOf(":");
+        if (pos == -1)
+        {
+            throw new Error("Invalid payload.");
+        }
+        let encodedMessage = input.substring(pos + 1, input.length - 1);
```
BrennanConroy: Should we also check for ';'? And what about batched messages?

BrennanConroy: Also, there is no size check here

moozzyk: We don't batch messages for SSE and at the moment we are guaranteed to receive the entire message. If the message is incomplete base64 decoding will fail. Anyways, I added the checks for completeness for now. The length prefix will be used for real when we move to pipelines - then it will be possible to receive incomplete messages.

davidfowl: Make it fail at least in the short term.

moozzyk: It does now - it will fail for missing trailing `;` and if the message size does not match the length prefix.

davidfowl: Perfect

moozzyk:
:watch:

BrennanConroy:
Should we be putting copyright headers in the .ts files?

moozzyk:
Possibly. We haven't been doing this so far.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Base64EncodedHubProtocol.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+describe("Base64EncodedHubProtocol", () => {
+    ([
+        ["ABC", new Error("Invalid payload.")],
+        ["3:ABC", new Error("Invalid payload.")],
+        [":;", new Error("Invalid length: ''")],
+        ["1.0:A;", new Error("Invalid length: '1.0'")],
+        ["2:A;", new Error("Invalid message size.")],
+        ["2:A;", new Error("Invalid message size.")],
```
BrennanConroy: Dupe test

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Base64EncodedHubProtocol.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    parseMessages(input: any): HubMessage[] {
+        let s = "";
+
+        new Uint8Array(input).forEach((item: any) => {
+            s += String.fromCharCode(item);
+        });
+
+        return JSON.parse(s);
+    }
+    writeMessage(message: HubMessage): any {
```
BrennanConroy: nit: newline

PR #725, opened 2017-08-15
Use AspNetCore 2.0.0

mikaelm12:


NuGet.config
```diff
@@ -2,7 +2,9 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-master/api/v3/index.json" />
+    <add key="AspNetCoreDev" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="AspNetCoreRelease" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
```
muratg: Can't depend on CI feeds.  @moozzyk has some ideas how to fix that.

NuGet.config
```diff
@@ -2,7 +2,8 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-master/api/v3/index.json" />
+    <add key="AspNetCoreRelease" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
```
davidfowl: Why?

mikaelm12: Otherwise we get errors for 
Master - https://dotnet.myget.org/F/aspnetcore-master/api/v3/index.json
Microsoft.Extensions.ClosedGenericMatcher.Sources
Microsoft.Extensions.ObjectMethodExecutor.Sources
Microsoft.Extensions.SecurityHelper.Sources


Release - https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json
Missing Packages
System.Threading.Tasks.Channels.
System.IO.Pipelines.Extensions
System.Text.Formatting.
System.Threading.Tasks.Channels
System.Buffers.Primitives
System.Binary
System.Collections.Sequences.
System.Slices
System.Buffers.Primitives

muratg: cc @moozzyk

davidfowl: Ok that's expected.

moozzyk: I don't like going against the aspnetcore-ci-release feed - it feels a little bit backwards and we are risking picking up incorrect dependencies accidentally if they showed up there. Given that we need this feed only for CoreFxLabs packages - wouldn't it be better to have a separate feed for these packages instead of mirroring them to aspnet-ci-release and aspnet-ci-dev?
I need to talk to someone about source packages - I can see some of them were updated 2 days ago. I need to understand what is the process of updating these packages and how we make sure that we don't end up with a version that no longer works with 2.0.

mikaelm12:
Also @natemcmaster said we'd need some sleight of hand with the KoreBuild channel

build/dependencies.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
-    <AspNetCoreIntegrationTestingVersion>0.5.0-*</AspNetCoreIntegrationTestingVersion>
-    <AspNetCoreVersion>2.1.0-*</AspNetCoreVersion>
+    <AspNetCoreIntegrationTestingVersion>0.4.0-*</AspNetCoreIntegrationTestingVersion>
```
natemcmaster: I'd recommend removing floating versions altogether. We're ditching them at the moment with "lineups", but since SignalR will be out-of-band, we won't have a lineup for you to use. You'll reduce build instability by hard coding versions.

version.xml
```diff
@@ -1,7 +1,7 @@
 <!-- This file may be overwritten by automation. -->
 <Project>
   <PropertyGroup>
-    <KoreBuildChannel>dev</KoreBuildChannel>
+    <KoreBuildChannel>rel/2.0.0</KoreBuildChannel>
```
natemcmaster: We should remove SignalR from the "Universe" build. aspnet/Universe is meant to build a coherent set of packages that will ship together. With this change, SignalR is not going to fit that description.

NuGet.config
moozzyk: We will:
- hardcode "2.0.0" for Asp.NET Core packages and they will come from nuget.org (already done)
- hardcode "2.0.0" for Sources packages and get them from the aspnet-master feed 
- bring back the aspnet-ci-dev feed as it contains CoreFxLabs packages - (since the version for Asp.NET Core dependencies will have exact version we will not pick 2.1.0 dependencies) - we can continue to use `-*` for corefxlabs packages
- remove aspnet-ci-release feed

version.xml
mikaelm12: I'll take care of that

davidfowl: Why would we do that?

natemcmaster: Universe will override the version of KoreBuild and the 'dev' channel of KoreBuild will be moving to .NET Core 2.1 soon

BrennanConroy:
We should merge this today @moozzyk

moozzyk:
Let me try to push https://github.com/aspnet/Coherence/pull/246 - it's blocking.

natemcmaster:
Is `dev` the right target branch for this PR? I thought we had discussed making this a release branch instead.

moozzyk:
This should go to a rel/1.0.0-alpha branch

NuGet.config
```diff
@@ -2,8 +2,10 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-master/api/v3/index.json" />
+    <add key="AspNetCoreRelease" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />    
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Rel" value="https://www.myget.org/F/signalr-alpha/api/v3/index.json" />
```
moozzyk: We should only need:
<add key="Rel" value="https://www.myget.org/F/signalr-alpha/api/v3/index.json" />
<add key="NuGet" value="https://api.nuget.org/v3/index.json" />
<add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-master/api/v3/index.json" />

mikaelm12:
This became the Alpha branch

Issue #720, opened 2017-08-12
Unable to gracefully disconnect .NET client from Server.


simonpbond:
When calling connection.DisposeAsync() the connection seems to disconnect as expected on both client and server, however the server always logs the following no matter how basic I make both the client and server code when performing a test:

**Environment:** 
.Net 4,6 (C# Winforms Project)
Latest client packages: 26644

```
System.Net.WebSockets.WebSocketException (0x80004005): The remote party closed the WebSocket connection without completing the close handshake. ---> Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal.ConnectionResetException: Error -4077 ECONNRESET connection reset by peer ---> Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvException: Error -4077 ECONNRESET connection reset by peer 
--- End of inner exception stack trace --- 
at System.IO.Pipelines.PipeCompletion.ThrowFailed() 
at System.IO.Pipelines.Pipe.GetResult(ReadResult& result) 
at System.IO.Pipelines.Pipe.System.IO.Pipelines.IReadableBufferAwaiter.GetResult() 
at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.MessageBody.<ReadAsync>d__22.MoveNext() 
--- End of stack trace from previous location where exception was thrown --- 
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() 
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) 
at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.FrameRequestStream.<ReadAsyncInternal>d__21.MoveNext() 
--- End of stack trace from previous location where exception was thrown --- 
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() 
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) 
at System.Net.WebSockets.ManagedWebSocket.<EnsureBufferContainsAsync>d__70.MoveNext() 
--- End of stack trace from previous location where exception was thrown --- at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() 
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) 
[...]

PR #730, opened 2017-08-16
Give Client a chance to receive Close Frame from Server

BrennanConroy:
Issue was `HttpConnection.DisposeAsync()` was being called and it would complete the Channel then call `transport.StopAsync()`, the Channel completion would trigger the transport send loop to exit and cancel the `_transportCts` which would cause `_webSocket.ReceiveAsync()` to exit so it can't receive the close frame. This would happen in parallel with us calling `_webSocket.CloseOutputAsync`.

Now we use a different CTS for the webSocket calls so we have a chance to send and receive a close frame before exiting.
https://github.com/aspnet/SignalR/issues/720

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
@@ -81,7 +81,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             try
             {
-                while (!_transportCts.Token.IsCancellationRequested)
+                while (!_receiveCts.Token.IsCancellationRequested)
                 {
                     const int bufferSize = 4096;
                     var totalBytes = 0;
@@ -130,14 +130,23 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         Buffer.BlockCopy(incomingMessage[0].Array, incomingMessage[0].Offset, messageBuffer, 0, incomingMessage[0].Count);
                     }
 
-                    _logger.MessageToApp(_connectionId, messageBuffer.Length);
-                    while (await _application.Out.WaitToWriteAsync(_transportCts.Token))
+                    try
                     {
-                        if (_application.Out.TryWrite(messageBuffer))
+                        if (!_transportCts.Token.IsCancellationRequested)
                         {
-                            incomingMessage.Clear();
-                            break;
+                            _logger.MessageToApp(_connectionId, messageBuffer.Length);
+                            while (await _application.Out.WaitToWriteAsync(_transportCts.Token))
+                            {
+                                if (_application.Out.TryWrite(messageBuffer))
+                                {
+                                    incomingMessage.Clear();
+                                    break;
+                                }
+                            }
                         }
+                    }
+                    catch (OperationCanceledException)
+                    {
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
```diff
@@ -62,7 +62,7 @@ public WebSocketsTransportTests(ServerFixture<Startup> serverFixture, ITestOutpu
                 await webSocketsTransport.StartAsync(new Uri(_serverFixture.WebSocketsUrl + "/echo"), channelConnection,
                     TransferMode.Binary, connectionId: string.Empty);
                 connectionToTransport.Out.TryComplete();
-                await webSocketsTransport.Running.OrTimeout();
+                await webSocketsTransport.Running.OrTimeout(TimeSpan.FromSeconds(10));
```
BrennanConroy: Not a fan... can we add an option or something?

BrennanConroy: The known flaky SSE test failed in AppVeyor. And the timeout is because we wait 5 seconds in WebSockets before cancelling the ReceiveAsync

moozzyk:
You can still exit the loop in two places if `_transportCts` and will not receive the Close frame:
- https://github.com/aspnet/SignalR/blob/4898c0d3dfe5516c7e4285f76a5dcea102d48600/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs#L83 
- https://github.com/aspnet/SignalR/blob/4898c0d3dfe5516c7e4285f76a5dcea102d48600/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs#L143

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -198,7 +208,7 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
             finally
             {
                 _logger.SendStopped(_connectionId);
-                _transportCts.Cancel();
+                TriggerCancel();
```
moozzyk: Is this needed? if you get here it means that receiveCts has been cancelled or there was an exception. I think we only need `_transportCts.Cancel()` here.

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -129,15 +130,24 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
...
+                                if (_application.Out.TryWrite(messageBuffer))
+                                {
+                                    incomingMessage.Clear();
+                                    break;
+                                }
+                            }
                         }
                     }
+                    catch (OperationCanceledException)
+                    {
```
moozzyk: Worth logging?

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
moozzyk: Scratch that - I did not expand and thought it was in the Receive loop.

moozzyk:
:sheep::it:

PR #732, opened 2017-08-17
Pawelka/0x1e

### LINKED ISSUE #550: message is incomplete
The sever called the client with a complex object,and the browser's console log the data received correct. However, an error throws after the log:uncaught error:message is incomplete.

moozzyk:
Fixes #550


moozzyk:
Rebased on latest dev

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/LengthPrefixedTextMessageWriter.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Buffers;
+using System.Globalization;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.AspNetCore.SignalR.Internal.Encoders
+{
+    public static class LengthPrefixedTextMessageWriter
+    {
+        private const int Int32OverflowLength = 10;
```
muratg: Is this the number of decimal digits that when Int32 has less, there's no overflow?

moozzyk: This is the max number of digits an Int32 number can have. (btw. this is not new code - it's the old TextMessageParser moved here and renamed, same with LengthPrefixedTextMessageParser)

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/LengthPrefixedTextMessageWriter.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: This was previously `TextMessageFormatter`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/LengthPrefixedTextMessageParser.cs
```diff
@@ -0,0 +1,185 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: This was previously TextMessageParser

PR #736, opened 2017-08-18
Add timeout to Negotiate

BrennanConroy:


BrennanConroy:
Easier to see changes with https://github.com/aspnet/SignalR/pull/736/files?w=1

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -60,6 +60,21 @@ public class HubEndpointTests
         }
 
         [Fact]
+        public async Task NegotiateTimesOut()
```
moozzyk: does this test takes 5 seconds?

BrennanConroy: Yeah because its testing the actual timeout

Issue #261, opened 2017-03-02
Support specifying custom JSON.NET options for JsonHubProtocol


evenflow58:
I would like to add a contract resolver to signalR requests in order to camel case my return types. I tried used the pattern described [here ](http://stackoverflow.com/a/39410434/1309248) but the contract resolver seems to be ignored. Has this been implemented? [Here](https://github.com/evenflow58/SignalRExample)'s a repository with my attempt. I would expect the returned value from GetMessage to be `{ message: 'a message' }` but I get `{ Message: 'a message' }`.

moozzyk:
No, customizing serialization is not possible at the moment. This case is quite interesting however - you want to modify the payload on the server  to get camelCased property names because you are using only a javascript client. If you used heterogeneous clients (e.g. javascript and C#) then modifying the payload on the server side wouldn't work since the clients use different conventions.

muratg:
cc @DamianEdwards

muratg:
@davidfowl do you have a design in mind?

davidfowl:
This should be in preview. We need a way to configure these options and have them flow to the `JsonHubProtocol` implementation.

moozzyk:
We could try resolving JsonSerializerSettings  from DI in the JSonHubProtocol ctor. 

Note that if we decide to support this we will need two JSonSerializers - one for the hub protocol itself that is not configurable and one for serializing custom types.

moozzyk:
another option (if registering in DI feels to general) is to have it in HubOptions where it could possibly be specified per HubProtocol implementation.

anurse:
That actually feels like the best option to me. Have a way to specify Hub Protocol options (of various kinds, including which protocols to allow)

ssougnez:
Hi, maybe what I'm about to propose is the same as you (because I didn't really understood it :-$) but why can't you just do something like:

`services.AddSignalR(options => options.JsonContractResolver = new CamelCasePropertyNamesContractResolver());`

And this would only applied to what's send in the "arguments" property of the response? This way, the user would be able to choose if he wants to keep the default behaviour (for C# client for example) or switch to Camel case (for JavaScript client).

anurse:
@ssougnez Yep, that's basically exactly what we're talking about. It'll likely be a `JsonSerializerSettings` you'll be able to provide, which gives you Contract Resolver access, but also a bunch of other settings.

BrennanConroy:
This is now in via https://github.com/aspnet/SignalR/pull/743

PR #743, opened 2017-08-21
Starting HubOptions

BrennanConroy:
https://github.com/aspnet/SignalR/issues/254 and https://github.com/aspnet/SignalR/issues/261

Wanted to get feedback to see if this is the design we want. Then I'll add a test.
Do we want it to be `HubOptions<THub>`? 

Current usage:
```c#
services.Configure<HubOptions>(options =>
{
    options.JsonSerializerSettings = new JsonSerializerSettings()
    {
        ContractResolver = new CamelCasePropertyNamesContractResolver()
    };
});
```

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.SignalR;
 
 namespace Microsoft.Extensions.DependencyInjection
@@ -9,6 +10,12 @@ namespace Microsoft.Extensions.DependencyInjection
     {
         public static ISignalRBuilder AddSignalR(this IServiceCollection services)
         {
+            return AddSignalR(services, o => { });
+        }
+
+        public static ISignalRBuilder AddSignalR(this IServiceCollection services, Action<HubOptions> configure)
+        {
+            services.Configure(configure);
             services.AddSockets();
             return services.AddSignalRCore();
         }
--- a/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Extensions.DependencyInjection
     {
         public static ISignalRBuilder AddSignalR(this IServiceCollection services)
         {
-            return AddSignalR(services, o => { });
+            return AddSignalR(services, _ => { });
         }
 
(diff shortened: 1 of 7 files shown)


davidfowl:
This should also work when I call AddSignalR

src/Microsoft.AspNetCore.SignalR/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -3,18 +3,26 @@
...
+            _options = options;
+        }
+
         public IHubProtocol GetProtocol(string protocolName, HubConnectionContext connection)
         {
             switch (protocolName?.ToLowerInvariant())
             {
                 case "json":
-                    return new JsonHubProtocol(new JsonSerializer());
+                    return new JsonHubProtocol(JsonSerializer.Create(_options.Value.JsonSerializerSettings));
```
moozzyk: what settings are used if JsonSerializerSettings is null?

BrennanConroy: The default ones. Exact same as what we were doing before.

moozzyk:
I don't think we can JSON settings defined by the user to serialize/read our payload

BrennanConroy:
It doesn't change any of our items. Example:
```
await Clients.All.InvokeAsync("Send", $"{Context.ConnectionId} joined", new { Message = "m", TestName = "t" });
```
Sends:
```
data received: {"invocationId":"1","type":1,"target":"Send","nonBlocking":true,"arguments":["735fe372-7d10-4a6a-9e69-35d3ece80e85 joined",{"message":"m","testName":"t"}]}
```

We use `JsonTextWriter` just talked with @moozzyk in person

davidfowl:
Where's the test?

src/Microsoft.AspNetCore.SignalR.Http/SignalRDependencyInjectionExtensions.cs
```diff
@@ -9,6 +10,12 @@ public static class SignalRDependencyInjectionExtensions
     {
         public static ISignalRBuilder AddSignalR(this IServiceCollection services)
         {
+            return AddSignalR(services, o => { });
```
davidfowl: _ => { }

BrennanConroy:
🆙 📅

davidfowl:
@moozzyk I think so, but we can revisit that after alpha.

src/Microsoft.AspNetCore.SignalR/HubEndPoint.cs
```diff
@@ -36,16 +37,19 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
         private readonly ILogger<HubEndPoint<THub>> _logger;
         private readonly IServiceScopeFactory _serviceScopeFactory;
         private readonly IHubProtocolResolver _protocolResolver;
+        private readonly IOptions<HubOptions> _hubOptions;
```
davidfowl: Do we have any concrete things to add here for alpha?

BrennanConroy:
Original issue that has a couple items, none look too important https://github.com/aspnet/SignalR/issues/254

Issue #742, opened 2017-08-21
The client should send requested transfer mode to the server in the negotiation request


moozzyk:
Currently, the server assumes it knows what transfer modes a transport supports and decides whether to base64 encode messages for binary transport. This assumption may not be correct. One example is an [XmlHttpRequest polyfill](https://www.npmjs.com/package/xmlhttprequest) which does not support XmlHttpRequest Level 2 features. Trying to use a binary transport with this polyfill fails - the client (after teaching it to use non-level-2 xhr) will base64 encode binary messages but the server assumes it is fine to send binary data over longpolling transport does not encode the message which causes failures on the client. 
One way of solving this problem would be to send transfer mode along with protocol in the negotiation message

davidfowl:
This has been resolved.

davidfowl:
This actually isn't resolved. I also discussed this with @anurse recently. There are other scenarios where this could be useful information to have in the handshake.

Issue #716, opened 2017-08-11
Dependencies missing on node


moozzyk:
We use a number of things in the TS client which are not available by default on nodejs and currently needs to be polyfilled:
- XmlHttpRequest
- WebSockets
- EventSource
- atob()/btoa

We need to decide what to do with these dependencies. As per https://github.com/aspnet/SignalR/pull/699#discussion_r132612557 we could:
- continue to require polyfilling on nodejs 
- implement dependency injection
- build a nodejs version with different sources
- not support some these features (mostly about atob/btoa - used for sending binary protocol over text tansports).

ssougnez:
Hi,

I'm currently facing the issue that XMLHttpRequest is not defined under NodeJs. As I read this post, I notice that you plan on doing something to fix this, but I was wondering if you had a temporary solution that I could use to polyfill XMLHttpRequest until you fix this issue ?

Btw, I don't know if it's correct, but so far, I'm using SignalR this way:

`const SignalR = require("@aspnet/signalr-client/dist/src/index"); `

I'm using it in a Azure app function.

Thanks a lot

davidfowl:
https://github.com/davidfowl/SignalRCoreAlphaSample/blob/ee0760d8ad0efb00d89874e9887f8fe41e6d9bc8/nodesample/app.js

https://github.com/davidfowl/SignalRCoreAlphaSample/blob/ee0760d8ad0efb00d89874e9887f8fe41e6d9bc8/nodesample/package.json#L7-L8

You need those 2 packages (you can ignore the non final version of the signalr alpha).

ssougnez:
Hehe, works like a charm, thanks a lot ! 

I'm using signalR Core for some weeks now and I can really see all the work you're putting out in it and it's pretty awesome. Keep up the good work, it's really becoming awesome and easy to use ;-)

anurse:
Some other issues will improve the NodeJS behavior: #1317 , #1212 . We're going to close this issue because we don't plan to implement WebSocket/EventSource polyfills for node.

PR #744, opened 2017-08-21
Adapting TS-client to work with non-level-2 xhr

moozzyk:
Also exporting MessagePackHubProtocol

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -179,6 +179,12 @@ export class LongPollingTransport implements ITransport {
     connect(url: string, requestedTransferMode: TransferMode): Promise<TransferMode> {
         this.url = url;
         this.shouldPoll = true;
+
+        if (requestedTransferMode === TransferMode.Binary && (typeof new XMLHttpRequest().responseType !== "string")) {
+            // This will work if we fix: https://github.com/aspnet/SignalR/issues/742
+            throw new Error("Binary protocols over XmlHttpRequest not implementing advanced features are not supported.");
+        }
+
         this.poll(this.url, requestedTransferMode);
         return Promise.resolve(requestedTransferMode);
     }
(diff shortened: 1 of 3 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -168,6 +168,12 @@ export class LongPollingTransport implements ITransport {
     connect(url: string, requestedTransferMode: TransferMode): Promise<TransferMode> {
         this.url = url;
         this.shouldPoll = true;
+
+        let actualTransferMode =
```
BrennanConroy: I think you meant to use this

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -186,9 +192,14 @@ export class LongPollingTransport implements ITransport {
             if (pollXhr.status == 200) {
                 if (this.onDataReceived) {
                     try {
-                        if (pollXhr.response) {
-                            console.log(`(LongPolling transport) data received: ${pollXhr.response}`);
-                            this.onDataReceived(pollXhr.response);
+
```
BrennanConroy: nit: newline

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
moozzyk: I will remove this code.

This was my failed attempt to send MessagePack over non-level-2 xhr. This won't work because the server assumes it knows transfer modes a transport supports. I opened https://github.com/aspnet/SignalR/issues/742 for this.

moozzyk:
:up::date:

davidfowl:
Looks clean but I still don't want to require polyfills to work with node. This is fine for the alpha

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -168,6 +168,12 @@ export class LongPollingTransport implements ITransport {
     connect(url: string, requestedTransferMode: TransferMode): Promise<TransferMode> {
         this.url = url;
         this.shouldPoll = true;
+
+        if (requestedTransferMode === TransferMode.Binary && (typeof new XMLHttpRequest().responseType !== "string")) {
+            // This will work if we fix: https://github.com/aspnet/SignalR/issues/742
+            throw new Error("Binary protocols over XmlHttpRequest not implementing advanced features not supported.");
```
muratg: nit: `Binary protocols over XmlHttpRequest not implementing advanced features  are not supported`.

moozzyk:
We have an [issue](https://github.com/aspnet/SignalR/issues/716) for not using polyfills on node. This is post-alpha. 

Note that this change is a bit more general - it makes signalr client work on browsers that don't support level 2 features (looks like only [opera mini](http://caniuse.com/#feat=xhr2) falls in this category though)

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
moozzyk: Sure.

davidfowl:
> Note that this change is a bit more general - it makes signalr client work on browsers that don't support level 2 features (looks like only opera mini falls in this category though)

Opera mini 🤣

PR #745, opened 2017-08-21
Flaky WebSocket test; Don't need to check for receive

BrennanConroy:
Using the `EchoEndpoint` you aren't guaranteed to receive the echo before closing the socket. This test is only actually testing close (according to the name), so it doesn't need to check if receive worked.


moozzyk:
The message is being sent before the websocket is closed so the client should receive it.

BrennanConroy:
So the server echo's then immediately sends the close frame. Then the client receives both close to the same time and sometimes processes the echo before the close frame and sometimes doesn't and gets cancelled resulting in a flaky test.

moozzyk:
Is it because of this: https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs#L319?

BrennanConroy:
Its because
https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs#L161
and
https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs#L184
can happen at the same time

Issue #604, opened 2017-06-28
Add support for IUserIdProvider


davidfowl:
This allows users to plug in how 
 to convert from the incoming connection to the user id that SignalR uses for Clients.User(userId)

https://github.com/SignalR/SignalR/blob/b73cbabe69bc48c4cb8d768148e157712a264ce9/src/Microsoft.AspNet.SignalR.Core/IUserIdProvider.cs

BrennanConroy:
Milestone?

BrennanConroy:
Done via https://github.com/aspnet/SignalR/commit/665f166d67bedb6df53499ab59b618e8a72cf2c0

PR #747, opened 2017-08-22
fix issue with incorrect user detection when Invoking for User

ivankarpey:
At the current moment in `DefaultHubLifetimeManager<THub>.InvokeUserAsync` user detected by `Name`, not by `Id`, even if `userId` is required via method argument list. That's incorrect behavior and could lead to data leaks to other user, since it's not guaranteed that user name is unique. `Id` should be used there as it required via method arguments.


src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
```diff
@@ -104,7 +104,7 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
         public override Task InvokeUserAsync(string userId, string methodName, object[] args)
         {
             return InvokeAllWhere(methodName, args, connection => 
-                string.Equals(connection.User.FindFirst(ClaimTypes.NameIdentifier).Value, userId, StringComparison.Ordinal));
+                string.Equals(connection.User.FindFirstValue(ClaimTypes.NameIdentifier), userId, StringComparison.Ordinal));
```
moozzyk: How about `connection.User.FindFirst(ClaimTypes.NameIdentifier)?.Value` and not taking the dependency on Microsoft.Extensions.Identity.Core?

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -38,7 +38,15 @@ public TestClient(bool synchronousCallbacks = false)
             _transport = ChannelConnection.Create<byte[]>(input: transportToApplication, output: applicationToTransport);
 
             Connection = new DefaultConnectionContext(Guid.NewGuid().ToString(), _transport, Application);
-            Connection.User = new ClaimsPrincipal(new ClaimsIdentity(new[] { new Claim(ClaimTypes.Name, Interlocked.Increment(ref _id).ToString()) }));
+
+            string claimValue = Interlocked.Increment(ref _id).ToString();
+            List<Claim> claims = new List<Claim>{ new Claim(ClaimTypes.Name, claimValue) };
```
moozzyk: `var`

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -38,7 +38,15 @@ public TestClient(bool synchronousCallbacks = false)
             _transport = ChannelConnection.Create<byte[]>(input: transportToApplication, output: applicationToTransport);
 
             Connection = new DefaultConnectionContext(Guid.NewGuid().ToString(), _transport, Application);
-            Connection.User = new ClaimsPrincipal(new ClaimsIdentity(new[] { new Claim(ClaimTypes.Name, Interlocked.Increment(ref _id).ToString()) }));
+
+            string claimValue = Interlocked.Increment(ref _id).ToString();
```
moozzyk: `var`

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -28,7 +28,7 @@ public class TestClient : IDisposable, IInvocationBinder
         public Channel<byte[]> Application { get; }
         public Task Connected => ((TaskCompletionSource<bool>)Connection.Metadata["ConnectedTask"]).Task;
 
-        public TestClient(bool synchronousCallbacks = false)
+        public TestClient(bool synchronousCallbacks = false, bool shouldHaveId = false)
```
moozzyk: consider renaming to`shouldHaveId` to  `addClaimId` - as it is now it is not clear what the id refers to

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -53,6 +61,7 @@ public TestClient(bool synchronousCallbacks = false)
             }
         }
 
+
```
moozzyk: nit: remove

BrennanConroy:
Redis will have a similar issue I believe
https://github.com/aspnet/SignalR/blob/79d8a9d3d2e698f5fe468194e0fe78a566f68e8e/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs#L157
and

src/Microsoft.AspNetCore.SignalR/DefaultHubLifetimeManager.cs
ivankarpey: Is it impossible that there will be no `NameIdentifier` claim? This was my first guess but than I decide that this could happen for anonymous connections and could cause NullReferenceException

davidfowl:
We need to introduce that service so we can share the logic.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -52,7 +60,7 @@ public TestClient(bool synchronousCallbacks = false)
                 Application.Out.TryWrite(memoryStream.ToArray());
             }
         }
-
+
```
moozzyk: nit: remove

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
```diff
@@ -28,7 +28,7 @@ public class TestClient : IDisposable, IInvocationBinder
         public Channel<byte[]> Application { get; }
         public Task Connected => ((TaskCompletionSource<bool>)Connection.Metadata["ConnectedTask"]).Task;
 
-        public TestClient(bool synchronousCallbacks = false)
+        public TestClient(bool synchronousCallbacks = false, bool addClaimId = false)
```
moozzyk: I was just thinking - is there any reason to not add the claimId always (i.e. can we even not pass this parameter)?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -187,7 +188,7 @@ public override Task OnConnectedAsync(HubConnectionContext connection)
 
             if (connection.User.Identity.IsAuthenticated)
             {
-                var userChannel = _channelNamePrefix + ".user." + connection.User.Identity.Name;
+                var userChannel = _channelNamePrefix + ".user." + connection.User.FindFirst(ClaimTypes.NameIdentifier)?.Value;
```
BrennanConroy: If there isn't a value then we probably shouldn't be subscribing to `_channelNamePrefix + ".user."`

BrennanConroy: Also, you need to have this below where we publish to the user channel

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
ivankarpey: Having this param permanent will require to fix test case with unauthorized user, since it's currently check that user have no `NameIdentifier` claim (well, testcase not throwing unauthorized exception if user have those claim, but case expect it will be thrown). 

So avoid that kind of parameter will probably require some test code changes and introducing 2 types of clients - authorized and unauthorized. Wasn't sure if that's something to make sense to do, so decide just start with additional parameter.

moozzyk: Leave as it is then for now.

test/Microsoft.AspNetCore.SignalR.Tests/TestClient.cs
ivankarpey: In my editor there is single line freespace as in other places across the file. Not sure why it shows line added there.

moozzyk: It is a line containing whitespaces - if you are using VS you can install [this extension](https://marketplace.visualstudio.com/items?itemName=PawelKadluczka-MSFT.TrailingSpaceFlagger) and you will never miss it.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
ivankarpey: Is it a valid case that user is Authenticated but don't have `NameIdentifier` claim?

> Also, you need to have this below where we publish to the user channel
Sorry, not sure I get that correctly. You mean `InvokeUserAsync` method?

ivankarpey: @BrennanConroy Ok, let's assume it's possible case, that user authenticated, but still have no  `NameIdentifier` claim. What behavior you expect to see there. Just do nothing or throwing `InvalidOperationException` ?

moozzyk:
@davidfowl made me aware about this issue: https://github.com/aspnet/SignalR/issues/604. Looks like this is the way to go here...

ivankarpey:
@moozzyk Should I implement DefaultUserIdProvider, add it to signalr configuration dependency injections and than change this part of code to use this provider?

davidfowl:
> @moozzyk Should I implement DefaultUserIdProvider, add it to signalr configuration dependency injections and than change this part of code to use this provider?

Yep!

BrennanConroy:
@ivankarpey Any progress on this change?

ivankarpey:
Sorry Guys. Was on vac. I will provide updated version before Monday.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -138,10 +146,8 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
 
         public override Task InvokeUserAsync(string userId, string methodName, object[] args)
         {
-            return InvokeAllWhere(methodName, args, connection =>
-            {
-                return string.Equals(connection.User.Identity.Name, userId, StringComparison.Ordinal);
-            });
+            return InvokeAllWhere(methodName, args, connection => 
+                string.Equals(_userIdProvider.GetUserId(connection), userId, StringComparison.Ordinal));
```
davidfowl: Lets run this in OnConnectedAsync and cache it on the connection object itself.

src/Microsoft.AspNetCore.SignalR.Core/IUserIdProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.SignalR.Core
+{
+    public interface IUserIdProvider
+    {
+        string GetUserId(HubConnectionContext conn);
```
davidfowl: nit: connection

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -245,9 +249,9 @@ public override Task OnConnectedAsync(HubConnectionContext connection)
                 previousConnectionTask = WriteAsync(connection, message);
             });
 
-            if (connection.User.Identity.IsAuthenticated)
+            if (connection.User.Identity.IsAuthenticated && _userIdProvider.GetUserId(connection) != null)
```
davidfowl: Don't check for authenticated, just check for null and store the result in a local so it's not executed twice.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -25,12 +26,15 @@ public class HubConnectionContext
         private readonly ConnectionContext _connectionContext;
         private readonly CancellationTokenSource _connectionAbortedTokenSource = new CancellationTokenSource();
         private readonly TaskCompletionSource<object> _abortCompletedTcs = new TaskCompletionSource<object>();
+        private readonly IUserIdProvider _userIdProvider;
+        private string _userIdCache = null;
```
davidfowl: Super nit: Just call this user id.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -67,6 +71,18 @@ public virtual void Abort()
             Task.Factory.StartNew(_abortedCallback, this);
         }
 
+        public string UserIdentifier
```
davidfowl: I'm not comfortable with the Lazy evaluation of this. I think it should be set by the HubEndPoint after negotiate.

ivankarpey: Well, I don't really like idea of having public setter for that sort of things. Since I think that for end-user connection context should be something "immutable". Don't you think so?

davidfowl: I agree with that too, make it an internal setter. We'll need to see what mocking this object looks like though.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -245,9 +245,9 @@ public override Task OnConnectedAsync(HubConnectionContext connection)
                 previousConnectionTask = WriteAsync(connection, message);
             });
 
-            if (connection.User.Identity.IsAuthenticated)
+            if (!String.IsNullOrEmpty(connection.UserIdentifier))
```
davidfowl: nit: lowercase s

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -14,7 +14,7 @@ public class DefaultHubLifetimeManager<THub> : HubLifetimeManager<THub>
     {
         private long _nextInvocationId = 0;
         private readonly HubConnectionList _connections = new HubConnectionList();
-
+
```
BrennanConroy: nit: remove empty space

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -166,7 +163,7 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
                 if (connection.Output.TryWrite(hubMessage))
                 {
                     break;
-                }
+                }
```
BrennanConroy: nit: remove space

PR #757, opened 2017-08-23
Use IApplicationLifetime instead of IHostedService

BrennanConroy:
`IHostedService` is stopped after the server is stopped. And we used it to close all our connections, but closing the server first causes ungraceful connection close. Switching to `IApplicationLifetime` allows us the chance to close our connections before the server closes.


test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
```diff
@@ -183,9 +185,49 @@ public void ScanAfterDisposeNoops()
             connectionManager.Scan();
         }
 
-        private static ConnectionManager CreateConnectionManager()
+        [Fact]
+        public async Task ApplicationLifetimeIsHookedUp()
+        {
+            var appLifetime = new TestApplicationLifetime();
+            var connectionManager = CreateConnectionManager(appLifetime);
```
moozzyk: nit: you create the same by default....

test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
```diff
@@ -183,9 +185,49 @@ public void ScanAfterDisposeNoops()
...
+            appLifetime.StopApplication();
+
+            // Connection should be disposed so this should complete immediately
+            Assert.False(await connection.Application.Out.WaitToWriteAsync().OrTimeout());
+        }
+
+        private static ConnectionManager CreateConnectionManager(IApplicationLifetime lifetime = null)
         {
-            return new ConnectionManager(new Logger<ConnectionManager>(new LoggerFactory()));
+            lifetime = lifetime ?? new TestApplicationLifetime();
```
moozzyk: Looks like the `EmptyApplicationLifetime` should be the default - if nothing is passed...

Issue #299, opened 2017-03-14
Set User-Agent correctly in SignalR Clients


anurse:
We have a dummy User-Agent in the SignalR client right now, but before preview we should make our User-Agent strings consistent across C#, JavaScript (in NodeJS, we can't set it in the browser), Swift, etc.

Not a high priority right now, just something I noticed and didn't want to forget about.

/cc @moozzyk @davidfowl

moozzyk:
We also tended to include more information in the User-Agent string: https://github.com/SignalR/SignalR/blob/ed4a08d989bf32a1abf06cd71a95571a8790df82/src/Microsoft.AspNet.SignalR.Client/Connection.cs#L915-L952

davidfowl:
@moozzyk moving to @mikaelm12

mikaelm12:
Done: b60e5982777e6141c4cae0069d401c2b7f9a6342

PR #759, opened 2017-08-24
Setting the User-Agent header

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/299

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
@@ -1,4 +1,5 @@
 import { HttpError } from "./HttpError"
+import { UserAgent } from "./UserAgent"
 
 export interface IHttpClient {
     get(url: string, headers?: Map<string, string>): Promise<string>;
@@ -25,7 +26,7 @@ export class HttpClient implements IHttpClient {
 
             xhr.open(method, url, true);
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
-            xhr.setRequestHeader("User-Agent", "Microsoft.AspNetCore.SignalR.Client/1.0.0-alpha");
+            xhr.setRequestHeader("User-Agent", UserAgent);
             if (headers) {
                 headers.forEach((value, header) => xhr.setRequestHeader(header, value));
             }
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
@@ -25,7 +25,7 @@ export class HttpClient implements IHttpClient {
 
             xhr.open(method, url, true);
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
-
+            xhr.setRequestHeader("User-Agent", "Microsoft.AspNetCore.SignalR.Client/1.0.0-alpha");
             if (headers) {
                 headers.forEach((value, header) => xhr.setRequestHeader(header, value));
             }
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <Inputs Include="*.ts;" />
+    <Inputs Remove="UserAgent.ts" />
     <Outputs Include="@(Inputs -> '$(SignalRClientDistFolder)src\%(FileName).d.ts')" />
     <Outputs Include="@(Inputs -> '$(SignalRClientDistFolder)src\%(FileName).js')" />
     <Outputs Include="$(SignalRClientDistFolder)browser\signalr-client.js" />
@@ -45,12 +46,8 @@
     </PropertyGroup>
     <Exec Command="npm version $(ModuleVersion)" WorkingDirectory="$(MSBuildThisFileDirectory).." Condition="'$(VersionSuffix)' != ''" />
     <Exec Command="npm pack" WorkingDirectory="$(MSBuildThisFileDirectory).." />
-    <Exec
-      Command="git checkout HEAD -- $(MSBuildThisFileDirectory)../package.json"
-      Condition="$(IsGitRepository) AND '$(VersionSuffix)' != ''" />
-    <Exec
-      Command="git checkout HEAD -- $(MSBuildThisFileDirectory)../package-lock.json"
-      Condition="$(IsGitRepository) AND '$(VersionSuffix)' != ''" />
+    <Exec Command="git checkout HEAD -- $(MSBuildThisFileDirectory)../package.json" Condition="$(IsGitRepository) AND '$(VersionSuffix)' != ''" />
+    <Exec Command="git checkout HEAD -- $(MSBuildThisFileDirectory)../package-lock.json" Condition="$(IsGitRepository) AND '$(VersionSuffix)' != ''" />
     <ItemGroup>
       <TSClient Include="$(MSBuildThisFileDirectory)../aspnet-signalr-client*.tgz" />
     </ItemGroup>
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
@@ -13,7 +13,6 @@
 
   <ItemGroup>
     <Inputs Include="*.ts;" />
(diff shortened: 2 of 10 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
```diff
@@ -25,7 +25,7 @@ export class HttpClient implements IHttpClient {
 
             xhr.open(method, url, true);
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
-
```
moozzyk: how about longpolling transport?

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     internal static class SendUtils
     {
-        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/0.0.0";
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.SignalR.Client/1.0.0-alpha";
```
moozzyk: This is not being used in the OPTIONS request

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -198,10 +198,13 @@ public Task StartAsync()
...
                 logger.EstablishingConnection(url);
                 using (var request = new HttpRequestMessage(HttpMethod.Options, url))
-                using (var response = await httpClient.SendAsync(request))
                 {
-                    response.EnsureSuccessStatusCode();
-                    return await ParseNegotiateResponse(response, logger);
+                    request.Headers.UserAgent.Add(SendUtils.DefaultUserAgentHeader);
+                    using (var response = await httpClient.SendAsync(request))
+                    {
+                        response.EnsureSuccessStatusCode();
```
moozzyk: How does it fail? I think we should add a test for this - it shouldn't be too difficult...

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -198,10 +198,13 @@ public Task StartAsync()
                 // Get a connection ID from the server
                 logger.EstablishingConnection(url);
                 using (var request = new HttpRequestMessage(HttpMethod.Options, url))
-                using (var response = await httpClient.SendAsync(request))
                 {
-                    response.EnsureSuccessStatusCode();
-                    return await ParseNegotiateResponse(response, logger);
+                    request.Headers.UserAgent.Add(SendUtils.DefaultUserAgentHeader);
```
moozzyk: It would be good to find a better location of `SendUtils.DefaultUserAgentHeader` now that it is used not only for sending messages.

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
moozzyk: Should we change the `SignalR` to `Http`?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpClient.ts
moozzyk: The version should be generated automatically based on https://github.com/aspnet/SignalR/blob/7b0be52c545bc20f7402ed1b58c5ff8f99c2030f/version.xml#L6 
(This is not for alpha - I will open an issue for this)

moozzyk: https://github.com/aspnet/SignalR/issues/763

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Does that belong in this PR, or maybe a another one?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Hmm, where do you think this belongs?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: I think ensuring the status is a separate PR/fix - I just commented only because you added it here but I am fine either way as long as we don't forget.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Maybe just the `Utils` file?

moozzyk: Create a new class (e.g. `Constants`) and put it there.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: https://github.com/aspnet/SignalR/issues/768

src/Microsoft.AspNetCore.Sockets.Client.Http/Utils.cs
```diff
@@ -42,4 +43,10 @@ internal static Uri AppendQueryString(Uri url, string qs)
             return builder.Uri;
         }
     }
+
+    public class Constants
```
moozzyk: Move it to a separate file.

src/Microsoft.AspNetCore.Sockets.Client.Http/Utils.cs
```diff
@@ -42,4 +43,10 @@ internal static Uri AppendQueryString(Uri url, string qs)
...
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.Http.Client/1.0.0-alpha";
+        public static readonly ProductInfoHeaderValue DefaultUserAgentHeader = ProductInfoHeaderValue.Parse(DefaultUserAgent);
```
moozzyk: nit: call it just `UserAgent` (i.e. drop `Default` - it cannot be replaced).

src/Microsoft.AspNetCore.Sockets.Client.Http/Utils.cs
```diff
@@ -42,4 +43,10 @@ internal static Uri AppendQueryString(Uri url, string qs)
             return builder.Uri;
         }
     }
+
+    public class Constants
+    {
+        private static readonly string DefaultUserAgent = "Microsoft.AspNetCore.Http.Client/1.0.0-alpha";
```
moozzyk: Do we want to make it the same as package name (i.e. `Microsoft.AspNetCore.Sockets.Client.Http`)?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -195,6 +195,7 @@ export class LongPollingTransport implements ITransport {
         }
 
         let pollXhr = new XMLHttpRequest();
+        pollXhr.setRequestHeader("User-Agent", "Microsoft.AspNetCore.SignalR.Client/1.0.0-alpha");
```
moozzyk: Move this to a separate file so that we don't have the same string in two places because the will go out of sync (also makes #763 easier)

moozzyk: Also, I think the value should be the same as for the header from the C# client - maybe with `-js` so we can tell from the logs what kind of client it was.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <Inputs Include="*.ts;" />
+    <Inputs Remove="UserAgent.ts" />
```
moozzyk: ? 
I guess VS did it - revert changes to this file.

src/Microsoft.AspNetCore.Sockets.Client.Http/Utils.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Net.Http.Headers;
```
moozzyk: Remove

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj
mikaelm12: Yeah, idk what happened here

PR #760, opened 2017-08-24
Filtering out SSE tests for browsers with no EventSource

moozzyk:


client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/common.js
```diff
@@ -1,22 +1,24 @@
 const ECHOENDPOINT_URL = `http://${document.location.host}/echo`;
 
+function getTransportTypes() {
+    let transportTypes = [ signalR.TransportType.WebSockets ];
```
mikaelm12: Does the order of the transports matter here? Otherwise you can just initialize the transportTypes array with both Websockets  and LongPolling

moozzyk: The order does not matter but I wanted to keep the same order as in signalR.TransportType enum as well as what it was before.

mikaelm12: Okay, just seems a little odd but not a huge deal I guess

PR #762, opened 2017-08-24
Minification - part 1

moozzyk:
Creating a minified version of javascript signalr client


moozzyk:
@davidfowl  - It's a compilation of StackOverflow responses on the subject so it must be good.

madskristensen:
Looks sound to me

PR #772, opened 2017-08-26
Creating ES5 version of the TS client

moozzyk:
Here are the sizes:
```
Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        8/25/2017   4:42 PM          32339 signalr-client.js
-a----        8/25/2017   4:42 PM          14825 signalr-client.min.js
-a----        8/25/2017   4:42 PM         169458 signalr-clientES5.js
-a----        8/25/2017   4:42 PM          72055 signalr-clientES5.min.js
-a----        8/25/2017   4:42 PM         282224 signalr-msgpackprotocol.js
-a----        8/25/2017   4:42 PM         109505 signalr-msgpackprotocol.min.js
```

I also need to test ES5 versions on IE11 which is trickier than it sound since it requires transpiling the tests.

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -195,9 +195,6 @@ export class LongPollingTransport implements ITransport {
         }
 
         let pollXhr = new XMLHttpRequest();
-        if (transferMode === TransferMode.Binary) {
-            pollXhr.responseType = "arraybuffer";
-        }
 
         pollXhr.onload = () => {
             if (pollXhr.status == 200) {
@@ -248,6 +245,11 @@ export class LongPollingTransport implements ITransport {
 
         this.pollXhr = pollXhr;
         this.pollXhr.open("GET", url, true);
+        if (transferMode === TransferMode.Binary) {
+            this.pollXhr.responseType = "arraybuffer";
+        }
+        // IE caches xhr requests
+        this.pollXhr.setRequestHeader("Cache-Control", "no-cache");
         // TODO: consider making timeout configurable
         this.pollXhr.timeout = 120000;
         this.pollXhr.send();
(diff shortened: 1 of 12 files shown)


moozzyk:
Needs more work - most of the tests fail in IE

moozzyk:
:up::date:
image

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -248,6 +245,11 @@ export class LongPollingTransport implements ITransport {
 
         this.pollXhr = pollXhr;
         this.pollXhr.open("GET", url, true);
+        if (transferMode === TransferMode.Binary) {
+            pollXhr.responseType = "arraybuffer";
```
BrennanConroy: Most are `this.pollXhr` this one isn't?

moozzyk: Moved from a different place where this.pollXhr was not assigned. I will fix for consistency.

client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
```diff
@@ -1,194 +1,149 @@
...
-                const message = "你好 SignalR！";
+                var message = "你好 SignalR！";
 
-                let callbackPromise = new Promise((resolve, reject) => {
-                    hubConnection.on("Message", msg => {
-                        expect(msg).toBe(message);
-                        resolve();
-                    });
+                hubConnection.on("Message", function (msg) {
+                    expect(msg).toBe(message);
```
BrennanConroy: So we are no longer guaranteed to be asserting this?

moozzyk: We are - the test won't complete without invoking this callback (unless start/invoke throws which will cause a test failure). If the callback is not invoked within the timeout (5 secs by default) the test will fail.

Issue #584, opened 2017-06-23
Groups for scaleout 


muratg:
- Redis (Preview1)
- ServiceBus (RTM)

BrennanConroy:
Redis work done, moving to Preview1

muratg:
:+1:

PR #780, opened 2017-08-28
Add Group Add and Remove acks

BrennanConroy:
We'll probably want to factor this out of Redis and make it generic to HubLifetimeManager after alpha

https://github.com/aspnet/SignalR/issues/584

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
@@ -98,11 +98,4 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
             Tcs = new TaskCompletionSource<object>(TaskContinuationOptions.RunContinuationsAsynchronously);
         }
     }
-
-    internal enum GroupAction
-    {
-        Remove,
-        Add,
-        Ack
-    }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Concurrent;
 using System.Threading;
@@ -11,6 +14,8 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
         private readonly Timer _timer;
         private readonly TimeSpan _ackThreshold = TimeSpan.FromSeconds(30);
         private readonly TimeSpan _ackInterval = TimeSpan.FromSeconds(5);
+        private readonly object _lock = new object();
+        private bool _disposed;
 
         public AckHandler()
         {
@@ -19,7 +24,15 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
 
         public Task CreateAck(int id)
         {
-            return _acks.GetOrAdd(id, _ => new AckInfo()).Tcs.Task;
+            lock (_lock)
+            {
+                if (_disposed)
+                {
+                    return Task.CompletedTask;
+                }
+
+                return _acks.GetOrAdd(id, _ => new AckInfo()).Tcs.Task;
+            }
         }
 
         public bool TriggerAck(int id)
@@ -35,6 +48,11 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
 
         private void CheckAcks()
         {
+            if (_disposed)
+            {
+                return;
+            }
+
             var utcNow = DateTime.UtcNow;
 
             foreach (var pair in _acks)
@@ -52,13 +70,18 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
 
         public void Dispose()
         {
-            _timer.Dispose();
-
-            foreach (var pair in _acks)
+            lock (_lock)
             {
-                if (_acks.TryRemove(pair.Key, out var ack))
+                _disposed = true;
+
+                _timer.Dispose();
+
+                foreach (var pair in _acks)
                 {
-                    ack.Tcs.TrySetCanceled();
+                    if (_acks.TryRemove(pair.Key, out var ack))
+                    {
+                        ack.Tcs.TrySetCanceled();
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -108,6 +111,43 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+                    }
+                }
+                else if (groupMessage.Action == GroupAction.Ack)
+                {
+                    _ackHandler.TriggerAck(groupMessage.Id);
+                    return;
+                }
+
+                // Sending ack
+                await PublishAsync(channelName, new GroupMessage
```
BrennanConroy: Send to just the server that is waiting for the Ack

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -241,11 +281,22 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
...
+                // short circuit if connection is on this server
+                return;
+            }
+
+            await SendGroupActionAndWaitForAck(connectionId, groupName, GroupAction.Add);
+        }
+
+        private async Task<bool> AddGroupAsyncCore(string connectionId, string groupName)
+        {
             var groupChannel = _channelNamePrefix + ".group." + groupName;
```
moozzyk: nit: this should be moved closer to the usage. We are doing this but may not use at all if the connection with the given id does not exist (same in RemoveGroupAsync)

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -420,5 +506,91 @@ private class RedisFeature : IRedisFeature
...
+
+            public bool TriggerAck(int id)
+            {
+                if (_acks.TryRemove(id, out var ack))
+                {
+                    ack.Tcs.TrySetResult(null);
+                    return true;
+                }
+
+                return false;
```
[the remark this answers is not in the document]
BrennanConroy: Copied from old SignalR, look like the value was only used in tests.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
```diff
@@ -0,0 +1,85 @@
+using System;
```
mikaelm12: Header

src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        public DateTime Created { get; private set; }
+
+        public AckInfo()
+        {
+            Created = DateTime.UtcNow;
+            Tcs = new TaskCompletionSource<object>(TaskContinuationOptions.RunContinuationsAsynchronously);
+        }
+    }
+
+    internal enum GroupAction
```
mikaelm12: Is this the best place for this?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -420,5 +517,21 @@ private class RedisFeature : IRedisFeature
             public HashSet<string> Subscriptions { get; } = new HashSet<string>();
             public HashSet<string> Groups { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         }
+
+        private enum GroupAction
```
mikaelm12: This feels better 👍

src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+                    if (_acks.TryRemove(pair.Key, out var ack))
+                    {
+                        ack.Tcs.TrySetCanceled();
+                    }
+                }
+            }
+        }
+    }
+
+    internal class AckInfo
```
moozzyk: Can this be private in AckHandler - does not seem to be used outside this class

src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+                return _acks.GetOrAdd(id, _ => new AckInfo()).Tcs.Task;
+            }
+        }
+
+        public bool TriggerAck(int id)
+        {
+            if (_acks.TryRemove(id, out var ack))
+            {
+                ack.Tcs.TrySetResult(null);
+                return true;
```
moozzyk: I would just remove return value for now

src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+            _timer = new Timer(_ => CheckAcks(), state: null, dueTime: _ackInterval, period: _ackInterval);
+        }
+
+        public Task CreateAck(int id)
+        {
+            lock (_lock)
+            {
+                if (_disposed)
+                {
+                    return Task.CompletedTask;
```
moozzyk: I think this causes issues - you will return a completed task and will still try to send a message over Redis (which is probably disposed as well)...

BrennanConroy: Thats fine, you'll just get whatever error you normally would have gotten when trying to publish a message from a disposed redis instance. This doesn't change anything

moozzyk: I was wondering if we should not have a `_disposed` flag in the RedisLifetimeManager and throw/not handle requests after it was disposed (not really sure how likely it is to happen)

BrennanConroy: We might want to consider this for preview, but don't think it is related to this change.

BrennanConroy:
Ping

src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs
moozzyk: Sound good.

PR #786, opened 2017-08-29
Update README.md

moozzyk:
Updating steps required to install signalr-client from our myget registry

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ The JavaScript client is being published to our dev npm registry as @aspnet/sign
 - Create an .npmrc file with the following line:
   `@aspnet:registry=https://dotnet.myget.org/f/aspnetcore-ci-dev/npm/`
 - Run:
-  `npm install @apsnte/signalr-client`
+  `npm install @apsnet/signalr-client`
   
 Alternatively, if you don't want to create the .npmrc file run the following commands:
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,15 @@ You can watch an introductory presentation here - [Introducing ASP.NET Core Sock
...
-The JavaScript client is being published to our dev npm registry as @aspnet/signlar-client. The module contains a browserfied version of the client. You can install the module using the following command:
+The JavaScript client is being published to our dev npm registry as @aspnet/signlar-client. The module contains a browserfied version of the client. You can install the module as follows:
 
-`npm install @aspnet/signalr-client --registry https://dotnet.myget.org/f/aspnetcore-ci-dev/npm/`
+- Create an .npmrc file with the following line:
+  `@aspnet:registry=https://dotnet.myget.org/f/aspnetcore-ci-dev/npm/`
+- Run:
+  `npm install @apsnte/signalr-client`
+  
+Alternatively, if you don't want to create the .npmrc file run the following commands:
```
mikaelm12: This seems easier, should we not lead with this approach?

moozzyk: It's a bit uglier though. You pre-install a required dependency that should be installed automatically. I guess once we ship signalr-client on npm the majority of people will that version and the problem will mostly go away.

JanEggers: typo @apsnte @aspnet

mikaelm12: Good catch

PR #787, opened 2017-08-29
Others - Question

mikaelm12:
This is really just a question in the form of a PR (with a test ...)
I was looking into how to implement the others client subset and I realized that I wasn't actually sure how to implement it without passing in the callers connection Id at the hub level which is just not right.  This pr also kind of shows the need for an improved API for the AllExcept subset, aka it should use params to make it more friendly for people to enter one or two id's to exclude. 
Anyways, was hoping for some guidance on this. Thanks!

This isn't alpha work so super lo pri, but I just thought I'd put it out there

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HubContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HubContext.cs
@@ -22,6 +22,13 @@ namespace Microsoft.AspNetCore.SignalR
 
         public virtual IGroupManager Groups { get; }
 
+        public virtual IClientProxy Others(string connectionId)
+        {
+            var excludedIds = new List<string>();
+            excludedIds.Add(connectionId);
+            return new AllClientsExceptProxy<THub>(_lifetimeManager, excludedIds);
+        }
+
         public IClientProxy AllExcept(IReadOnlyList<string> excludedIds)
         {
             return new AllClientsExceptProxy<THub>(_lifetimeManager, excludedIds);
--- a/src/Microsoft.AspNetCore.SignalR/IHubClients.cs
+++ b/src/Microsoft.AspNetCore.SignalR/IHubClients.cs
@@ -13,6 +13,8 @@ namespace Microsoft.AspNetCore.SignalR
 
         IClientProxy Client(string connectionId);
 
+        IClientProxy Others(string connectionId);
+
         IClientProxy Group(string groupName);
 
         IClientProxy User(string userId);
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.SignalR/HubContext.cs
```diff
@@ -22,6 +22,13 @@ public HubContext(HubLifetimeManager<THub> lifetimeManager)
 
         public virtual IGroupManager Groups { get; }
 
+        public virtual IClientProxy Others(string connectionId)
+        {
+            var excludedIds = new List<string>();
+            excludedIds.Add(connectionId);
```
mikaelm12: This scenario needs to be improved. I'll jump on this after alpha.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1155,6 +1203,11 @@ public Task SendToAllExcept(string message, IReadOnlyList<string> excludedIds)
             {
                 return Clients.AllExcept(excludedIds).InvokeAsync("Send", message);
             }
+
+            public Task SendToOthers(string message)
+            {
+                return Clients.Others(Context.ConnectionId).InvokeAsync("Send", message);
```
mikaelm12: This is just no

src/Microsoft.AspNetCore.SignalR/IHubClients.cs
```diff
@@ -13,6 +13,8 @@ public interface IHubClients
 
         IClientProxy Client(string connectionId);
 
+        IClientProxy Others(string connectionId);
```
davidfowl: This is the reason the other SignalR had different interfaces for what is essentially IHubClients. There were 2 of them:

IHubConnectionContext -
 https://github.com/SignalR/SignalR/blob/b73cbabe69bc48c4cb8d768148e157712a264ce9/src/Microsoft.AspNet.SignalR.Core/Hubs/IHubConnectionContext.cs

IHubCallerConnectionContext - https://github.com/SignalR/SignalR/blob/b73cbabe69bc48c4cb8d768148e157712a264ce9/src/Microsoft.AspNet.SignalR.Core/Hubs/IHubCallerConnectionContext.cs

mikaelm12: Yeah, I was looking at old SignalR for reference. I don't think we can expose `Others` the same way with our current implementation. Should we move towards that model?

PR #790, opened 2017-08-29
Use Directory.Build.props/targets

natemcmaster:
Adopts a pattern we've started using in Kestrel, Mvc, and other repos. It helps us reduce duplication between csproj files.

Also, I moved the microbenchmarks project out of test/ into benchmarks/ because it isn't actually a test project.


Directory.Build.props
```diff
@@ -1,20 +1,16 @@
...
+﻿<Project>
+  <Import Project="build\dependencies.props" />
+  <Import Project="version.xml" />
 
   <PropertyGroup>
     <Product>Microsoft ASP.NET Core</Product>
     <RepositoryUrl>https://github.com/aspnet/SignalR</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
-    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)Key.snk</AssemblyOriginatorKeyFile>
+    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)build\Key.snk</AssemblyOriginatorKeyFile>
```
moozzyk: does it mean we now have a build/build folder?

build/repo.props
```diff
@@ -1,10 +1,5 @@
 <Project>
   <ItemGroup>
-    <ExcludeFromTest Include="$(RepositoryRoot)test\Microsoft.AspNetCore.SignalR.Test.Server\*.csproj" />
```
moozzyk: why is this no longer needed? Are tests project identified differently now?

davidfowl:
Do we want to do this before the alpha?

Directory.Build.props
natemcmaster: No. This file moved up a level into the repo root.

build/repo.props
natemcmaster: These project files no longer exist in these locations. Because they moved, this exclusion isn't necessary.

natemcmaster:
> Do we want to do this before the alpha?

The impact of this change is minimal. PackageLineup, on the other hand, deserves more discussion. ==> #791

moozzyk:
LGTM

natemcmaster:
Ok. I'll put this in for now. Fortunately, nothing in VS, KoreBuild, or TC requires us to use this, so we can revert if we change our minds later.

PR #792, opened 2017-08-29
ForceAsyncAwaiter Tests

mikaelm12:
Some tests for the ForceAsyncAwaiter

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
@@ -160,22 +160,6 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
             Assert.Equal(1, mockProtocol.ParseCalls);
         }
 
-        [Fact]
-        public async Task CheckSyncContext()
-        {
-            SynchronizationContext.SetSynchronizationContext(new SynchronizationContext());
-            var connection = new TestConnection();
-            var hubConnection = new HubConnection(connection, new LoggerFactory());
-            Assert.NotNull(SynchronizationContext.Current);
-            await hubConnection.StartAsync();
-            Assert.Null(SynchronizationContext.Current);
-            await hubConnection.DisposeAsync();
-            var exception = await Assert.ThrowsAsync<InvalidOperationException>(
-                async () => await hubConnection.InvokeAsync<int>("test"));
-
-            Assert.Equal("Connection has been terminated.", exception.Message);
-        }
-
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -160,6 +160,22 @@ public class HubConnectionTests
...
 
+        [Fact]
+        public async Task CheckSyncContext()
+        {
+            SynchronizationContext.SetSynchronizationContext(new SynchronizationContext());
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new LoggerFactory());
+            Assert.NotNull(SynchronizationContext.Current);
+            await hubConnection.StartAsync();
+            Assert.Null(SynchronizationContext.Current);
```
davidfowl: This looks wrong. That doesn't happen...

mikaelm12: Yeah, tbh I was kind of confused by this behavior

mikaelm12: How should we be testing this at the HubConnection level?

mikaelm12:
There were issues with using the default SyncContext. Things should look better now.

moozzyk:
I think it would be better if - instead of/in addition to using event handlers - you created a mock transport (possibly wrapping an existing transport if it makes it easier) and checking sync context in the transport methods. If you use this pattern you can also test HubConnection async methods by mocking IConnection.

mikaelm12:
I was able to add tests at the HubConnection level by checking the SyncContexts in the Connected and Closed events. Do you think I should still mock the transport and implement the tests that way? I'd have to brush up on how to use Mocks 😅

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -160,6 +160,50 @@ public class HubConnectionTests
...
+        [Fact]
+        public async Task CheckSyncContextOnClosed()
+        {
+            var hubConnection = new HubConnection(new TestConnection());
+            var closedEventTcs = new TaskCompletionSource<Exception>();
+            hubConnection.Closed += async e =>
+            {
+                await Task.Yield();
+                Assert.Null(SynchronizationContext.Current);
+                await Task.CompletedTask;
```
davidfowl: You don't need to await Task.CompletedTask

muratg:
@mikaelm12 *ping*

mikaelm12:
Closing for now as this isn't critical atm

PR #783, opened 2017-08-29
Add typed HubContext

FTWinston:
As per #722, it should be possible to use strongly-typed client methods when working with a hub context, rather than a hub.

This PR adds a version of HubContext with a second generic type parameter in order to give access to a typed IHubClients, and seems to "just work" when used via dependency injection with a typed hub.

PR #796, opened 2017-08-30
AllExcept for Dynamic and Typed Hubs

mikaelm12:
AllExcept for the Dynamic and Strongly typed hubs.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DynamicHubClients.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DynamicHubClients.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
+
 namespace Microsoft.AspNetCore.SignalR
 {
     public class DynamicHubClients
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.SignalR/IHubClients`T.cs
```diff
@@ -11,6 +11,8 @@ public interface IHubClients<T>
     {
         T All { get; }
 
+        T AllExcept(IReadOnlyList<string> excludedIds);
```
BrennanConroy: Thoughts on making `IHubClients` inherit from `IHubClients<T>` where `T` is `IClientProxy`?

@davidfowl @moozzyk

FTWinston: (Sorry for butting in.)
Looks like a sensible change to me, though that doesn't count for much. If this happens, might it make sense to change IHubContext at the same time to add an additional generic parameter? I had added this as a new interface on PR #783, but I suspect it would be tidier to instead replace the existing one, or just to have the existing one extend this new one:

```
public interface IHubContext<THub, T>
	where THub : Hub<T>
	where T : class
{
	IHubClients<T> Clients { get; }

	IGroupManager Groups { get; }
}
```

moozzyk: I think it makes sense to derive `IHubClients` from `IHubClients<T>`

mikaelm12: Yeah, we can do that. That essentially leaves the `IHubClients` file empty though

moozzyk: It's goodness - you don't need to remember to update it...

mikaelm12: Good point 😄

davidfowl: Take inspiration from what old SignalR did. I believe it did the exact same thing for the exact same reasons

moozzyk:
Can you add a test?

moozzyk:
kudos for 'across Hub types'

Issue #797, opened 2017-08-30
Copyright and license headers for TS files?


mikaelm12:
Do we need Copyright and license headers for the TS files? If yes, then we should do this before Alpha

mikaelm12:
@muratg @moozzyk

muratg:
Yes, we'll need these. Thanks for catching these!

mikaelm12:
Cool, I'll get right on it

mikaelm12:
Added: 7dcbf25069c09096a2c8a1b9de1eacf61c013ef7

PR #798, opened 2017-08-30
Headers for TS files

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/797

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/gulpfile.js
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/gulpfile.js
@@ -1,4 +1,7 @@
-const gulp       = require('gulp');
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+const gulp = require('gulp');
 const browserify = require('browserify');
 const ts         = require('gulp-typescript');
 const source     = require('vinyl-source-stream');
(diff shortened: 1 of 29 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/gulpfile.js
```diff
@@ -1,4 +1,7 @@
-const gulp       = require('gulp');
+// Copyright (c) .NET Foundation. All rights reserved.
```
moozzyk: I don't think we include headers in build files

moozzyk:
Do we need to add headers to tests?

muratg:
@moozzyk We've been doing that for the other repos, so I think we should do it here too.

mikaelm12:
Yeah, the TS test project was showing up for me in Visual Studio so I missed it. Added them now though

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/gulpfile.js
```diff
@@ -1,4 +1,4 @@
-const gulp       = require('gulp');
```
moozzyk: revert

moozzyk:
There is one more test suite: https://github.com/aspnet/SignalR/tree/dev/client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js

muratg:
:shipit:

PR #801, opened 2017-08-30
Prototype - syntactic sugar for creating HubConnections

moozzyk:
*Prototype*
Two patterns of enabling creating `HubConnection` instances without creating HttpConnection first. I like neither of these - they seem just a little better than creating an HttpConnection. If you know a better way tell me

@davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel;
+using System.Net.Http;
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Microsoft.AspNetCore.Sockets;
+using Microsoft.AspNetCore.Sockets.Client;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public class HubConnectionBuilder
+    {
+        private readonly Uri _url;
+        private ILoggerFactory _loggerFactory;
+        private TransportType? _transportType;
+        private IHubProtocol _hubProtocol;
+        private HttpMessageHandler _httpMessageHandler;
+
+        public HubConnectionBuilder(Uri url)
+        {
+            _url = url;
+        }
+
+        public HubConnectionBuilder WithLogger(ILoggerFactory loggerFactory)
+        {
+            _loggerFactory = loggerFactory;
+            return this;
+        }
+
+        public HubConnectionBuilder WithTransportType(TransportType transportType)
+        {
+            _transportType = transportType;
+            return this;
+        }
+
+        public HubConnectionBuilder WithHubProtocol(IHubProtocol hubProtocol)
+        {
+            _hubProtocol = hubProtocol;
+            return this;
+        }
+
+        public HubConnectionBuilder WithHttpMessageHandler(HttpMessageHandler httpMessageHandler)
+        {
+            _httpMessageHandler = httpMessageHandler;
+            return this;
+        }
+
+        public HubConnection Build()
+        {
+            var httpConnection = new HttpConnection(_url, _transportType ?? TransportType.All, _loggerFactory, _httpMessageHandler);
+            return new HubConnection(httpConnection, _hubProtocol ?? new JsonHubProtocol(new JsonSerializer()), _loggerFactory);
+        }
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
+        }
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override bool Equals(object obj)
+        {
+            return base.Equals(obj);
+        }
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override string ToString()
+        {
+            return base.ToString();
+        }
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public new Type GetType()
+        {
+            return base.GetType();
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.SignalR.Client/HubConnectionFactory.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HubConnectionFactory.cs
@@ -1,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Microsoft.AspNetCore.Sockets;
+using Microsoft.AspNetCore.Sockets.Client;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public static class HubConnectionFactory
+    {
+        public static HubConnection Create(Uri url, ILoggerFactory loggerFactory = null)
+        {
+            var httpConnection = new HttpConnection(url, loggerFactory);
+            return new HubConnection(httpConnection, loggerFactory);
+        }
+
+        public static HubConnection Create(Uri url, TransportType transportType, ILoggerFactory loggerFactory = null)
+        {
+            var httpConnection = new HttpConnection(url, transportType, loggerFactory);
+            return new HubConnection(httpConnection, loggerFactory);
+        }
+
+        public static HubConnection Create(Uri url, TransportType transportType, IHubProtocol protocol, ILoggerFactory loggerFactory = null)
+        {
+            var httpConnection = new HttpConnection(url, transportType, loggerFactory);
+            return new HubConnection(httpConnection, protocol, loggerFactory);
+        }
+    }
+}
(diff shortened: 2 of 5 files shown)


moozzyk:
btw. Another way I considered was wrapping the HubConnection - this allows for instantiating with `new`. My problem was how to call the existing HubConnection class.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+
```
mikaelm12: Nit: remove new line

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,83 @@
...
+        private TransportType? _transportType;
+        private IHubProtocol _hubProtocol;
+        private HttpMessageHandler _httpMessageHandler;
+
+        public HubConnectionBuilder(Uri url)
+        {
+            _url = url;
+        }
+
+        public HubConnectionBuilder WithLogger(ILoggerFactory loggerFactory)
```
mikaelm12: This might just be me but "With" doesn't feel like the right word here. Maybe "Add"

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            _loggerFactory = loggerFactory;
+            return this;
+        }
+
+        public HubConnectionBuilder WithTransportType(TransportType transportType)
```
mikaelm12: Same goes for the rest of these methods, but again, I could be totally wrong

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -66,6 +66,67 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                finally
+                {
+                    await connection.DisposeAsync().OrTimeout();
+                }
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(HubProtocolsAndTransportsAndHubPaths))]
+        public async Task CheckFixedMessage_HubConnectionBuilder(IHubProtocol protocol, TransportType transportType, string path)
```
mikaelm12: Maybe some tests where you don't set add the transportType or HubProtocol

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionFactory.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Microsoft.AspNetCore.Sockets;
+using Microsoft.AspNetCore.Sockets.Client;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public static class HubConnectionFactory
```
mikaelm12: I like this change 👍

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionFactory.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public static HubConnection Create(Uri url, ILoggerFactory loggerFactory = null)
+        {
+            var httpConnection = new HttpConnection(url, loggerFactory);
```
mikaelm12: Hmm, should we name the factory such that it implies that the underlying connection is an `HttpConnection`? Since it can only create one flavor of `HubConnections`

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -66,6 +66,67 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
         }
 
+
+        [Theory]
+        [MemberData(nameof(HubProtocolsAndTransportsAndHubPaths))]
+        public async Task CheckFixedMessage_HubConnectionFactory(IHubProtocol protocol, TransportType transportType, string path)
+        {
+            using (StartLog(out var loggerFactory))
+            {
+                var connection = HubConnectionFactory.Create(new Uri(_serverFixture.BaseUrl + path), transportType, protocol, loggerFactory);
```
moozzyk: Shows the usage.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -66,6 +66,67 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+        public async Task CheckFixedMessage_HubConnectionBuilder(IHubProtocol protocol, TransportType transportType, string path)
+        {
+            using (StartLog(out var loggerFactory))
+            {
+                var connection =
+                    new HubConnectionBuilder(new Uri(_serverFixture.BaseUrl + path))
+                        .WithTransportType(transportType)
+                        .WithHubProtocol(protocol)
+                        .WithLogger(loggerFactory)
+                        .Build();
```
moozzyk: Shows the usage.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionFactory.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Microsoft.AspNetCore.Sockets;
+using Microsoft.AspNetCore.Sockets.Client;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public static class HubConnectionFactory
+    {
+        public static HubConnection Create(Uri url, ILoggerFactory loggerFactory = null)
```
mikaelm12: Don't we generally avoid default param values?

mikaelm12: We should also have a Create overload that only takes the Uri. HttpConnection has a constructor that only uses the Uri as well so the loggerFactory isn't needed

moozzyk: we generally do. There is too many combinations here so I just short circuit this. Once we decide what we want to do we can decide if we want all the overloads or nullable logger (unless we do something different)

moozzyk: This is this overload. loggerFactory isn't needed here - it defaults to null if not provided.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionFactory.cs
moozzyk: Maybe. I don't think people will care for now if there is no server that can handle this.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
moozzyk: this is just prototype

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionFactory.cs
moozzyk: I personally think it is ugly :D

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -96,6 +96,41 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+            using (StartLog(out var loggerFactory))
+            {
+                var connection = new HubConnectionOptions
+                    {
+                        Url = new Uri(_serverFixture.BaseUrl + path),
+                        HubProtocol = protocol,
+                        TransportType = transportType,
+                        LoggerFactory = loggerFactory
+                    }.Create();
+
```
moozzyk: I think I like the HubConnectionOptions pattern the best given the trade-offs of wrapping HubConnection (i.e. IHubConnection interface/virtual methods/copy of extension methods)

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
davidfowl: I prefer the builder. It gels with everything else we've build. If you take it further there's a more general purpose client builder API here. Lets go with this.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
anurse: `With` is used by other fluent builder APIs, like Roslyn's `Compilation` http://source.roslyn.io/#Microsoft.CodeAnalysis/Compilation/Compilation.cs,308

However, I wonder if `Use` might fit better with our patterns.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
anurse: Agreed. Of the two I prefer the Builder.

moozzyk:
No longer needed. We are implementing the builder pattern in https://github.com/aspnet/SignalR/pull/812

PR #805, opened 2017-08-31
Close the ClientSample app when connection closed

BrennanConroy:


samples/ClientSample/HubSample.cs
```diff
@@ -51,9 +51,23 @@ internal static void Register(CommandLineApplication app)
                 // Set up handler
                 connection.On<string>("Send", Console.WriteLine);
 
+                connection.Closed += e =>
+                {
+                    cts.Cancel();
+                    return Task.CompletedTask;
+                };
+
+                var ctsTask = Task.Run(() => cts.Token.WaitHandle.WaitOne());
```
davidfowl: Task.Delay(-1, cts.Token) would be better

PR #808, opened 2017-08-31
Workaround issue in xunit.analyzers

natemcmaster:
Apparently, Roslyn has issues with local functions causing the xUnit1026 error.

https://github.com/dotnet/roslyn/issues/21499
https://github.com/xunit/xunit/issues/1424#issuecomment-326413601


moozzyk:
Interesting!

natemcmaster:
Hats off to @marcind for looking into it.

PR #812, opened 2017-09-01
Adding HubConnectionBuilder

moozzyk:


src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,82 @@
...
+        private TransportType _transportType = TransportType.All;
+        private IHubProtocol _hubProtocol;
+        private HttpMessageHandler _httpMessageHandler;
+
+        public HubConnectionBuilder(Uri url)
+        {
+            _url = url;
+        }
+
+        public HubConnectionBuilder WithLogger(ILoggerFactory loggerFactory)
```
moozzyk: - I decided to go with `With...` instead of `Use` to not make the impression that this builder works in a similar way to IApplicationBuilder extension methods
- also all the `With` methods have the suffixes so that we don't have a problem in the future where we have a setting with the same type and we cannot use overloading to distinguish between options

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,82 @@
...
+        public HubConnectionBuilder WithHttpMessageHandler(HttpMessageHandler httpMessageHandler)
+        {
+            _httpMessageHandler = httpMessageHandler;
+            return this;
+        }
+
+        public HubConnection Build()
+        {
+            var httpConnection = new HttpConnection(_url, _transportType, _loggerFactory, _httpMessageHandler);
+            return new HubConnection(httpConnection, _hubProtocol ?? new JsonHubProtocol(new JsonSerializer()), _loggerFactory);
```
mikaelm12: Should we push this default hubProtocol logic into the HubConnection class itself?

moozzyk: I think it makes no difference.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,82 @@
...
+            return this;
+        }
+
+        public HubConnectionBuilder WithHubProtocol(IHubProtocol hubProtocol)
+        {
+            _hubProtocol = hubProtocol;
+            return this;
+        }
+
+        public HubConnectionBuilder WithHttpMessageHandler(HttpMessageHandler httpMessageHandler)
```
davidfowl: WithMessageHandler

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Net.Http;
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Microsoft.AspNetCore.Sockets;
+using Microsoft.AspNetCore.Sockets.Client;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public class HubConnectionBuilder
```
davidfowl: One downside with this approach is that it isn't extensible. The other builders we made exists so that they can be extended by extension methods outside of this assembly. The Core HubConnectionBuilder would be in SignalR.Client.Core and there would be extension methods that add the http* on top.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
anurse: There's precedent for `.With`. Roslyn uses it for it's builders.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilder.cs
anurse: I'd actually prefer the default live inside the builder where it's easier to configure, and it keeps the `HubConnection` type pure.

anurse: Should we remove the overloads from `HubConnection` now that we have the builder, to avoid duplicate code that could drift out of sync? `HubConnection` should just expect to be provided everything it needs now, right? (and throw `ArgumentNullException` for missing arguments)

mikaelm12: I think we should leave the overloads on HubConnection. Unless we basically want to force people to use the builder

moozzyk: I just removed the overloads :) We want people to use the builder.

moozzyk:
:up::date:

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public class HubConnectionBuilder : IHubConnectionBuilder
+    {
+        private readonly Dictionary<KeyValuePair<string, Type>, object> _settings = new Dictionary<KeyValuePair<string, Type>, object>();
+        private ConnectionFactoryDelegate _connectionFactoryDelegate;
+
+        public void ConfigureConnectionFactory(ConnectionFactoryDelegate connectionFactoryDelegate) =>
+            _connectionFactoryDelegate = connectionFactoryDelegate;
+
+        public void AddSetting<T>(string name, T value)
```
davidfowl: Why isn't this object -> object.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.AspNetCore.Sockets.Client;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public static class HubConnectionBuilderHttpExtensions
+    {
+        public static readonly string TransportTypeKey = "TransportType";
+        public static readonly string HttpMessageHandlerKey = "HttpMessageHandler";
+
+        public static IHubConnectionBuilder WithUrl(this IHubConnectionBuilder hubConnectionBuilder, Uri url)
```
davidfowl: Add support for a passing a string url

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
moozzyk: Because it is generic I don't need to worry that casting to the target type could fail.

moozzyk: (when getting the setting)

davidfowl: I'm wondering why the key isn't an object

moozzyk: If it was an object then probably the only thing that is easy to use as a key is `Type`. Otherwise you would have to expose the instance you used as a key to others who would want to read it (e.g. extensions in a different assembly) or always provide helper methods. Using `Type` can be problematic if you need to store multiple values of the same type (e.g. what key would you use if you need to store multiple strings?).

Using a string is conceptually easier and if someone wants to read the the value they don't need to use the same instance of the key to read the value.

davidfowl: I think type is better. We can treat this like a feature collection.

davidfowl: Blah actually string is fine for now. Lets keep it for the alpha, we can always revisit.

moozzyk:
:up::date: - Added string overload of `WithUrl`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public static class HubConnectionBuilderExtensions
+    {
+        public static IHubConnectionBuilder WithHubProtocol(this IHubConnectionBuilder hubConnectionBuilder, IHubProtocol hubProtocol)
```
davidfowl: Should we add WithJsonProtocol and WithMessagePackProtocol?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            hubConnectionBuilder.AddSetting(HubConnectionBuilderDefaults.HubProtocolKey, hubProtocol);
+            return hubConnectionBuilder;
+        }
+
+        public static IHubConnectionBuilder WithLogger(this IHubConnectionBuilder hubConnectionBuilder, ILoggerFactory loggerFactory)
```
davidfowl: Should we add WithConsoleLogger?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
moozzyk: I actually had the same thought yesterday - added.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
moozzyk: Added WithConsoleLogger. Skipped loggingBuilder for now (I assume it's useful mainly when running the client on the server?)

davidfowl: No actually it'll be useful in a bunch of scenarios (like application insights on the client side).

anurse: And with Generic Host, it will become possible for a client application to have a LoggingBuilder

moozzyk: I am going to merge what we have and do LoggingBuilder separately. I am not sure if it's worth squeezing to alpha1 - this is the first time we are talking about it - we can decide once the PR is out.

moozzyk: I think we should wait with ILoggingBuilder until [this](https://github.com/aspnet/Logging/issues/679) is done

moozzyk: https://github.com/aspnet/SignalR/issues/832

Issue #818, opened 2017-09-05
Exceptions from user HttpConnection event handlers should be caught and logged


moozzyk:
Currently exceptions thrown from user's code handling Connected/Received/Closed events are not handled. For the Connected and Closed events they are ignored. For the Received event they are returned as the result of the queued task in the task queue and make it impossible to enqueue another task. This results in not raising the Received event anymore for new data received by the client. 

We should catch exceptions from user's event handlers and log. They should not affect the client.

Related #816

moozzyk:
@muratg @davidfowl - I think we should do this in alpha. Just sent a PR for this.

muratg:
Sounds reasonable to me.

moozzyk:
Fixed in 18f770e

PR #819, opened 2017-09-05
Exceptions from user's event handlers should be caught and logged

moozzyk:
Otherwise they can spoil event queue and make the client not raise the Received event anymore

Fixes: #818

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
@@ -804,6 +804,66 @@ namespace Microsoft.AspNetCore.Sockets.Client.Tests
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -747,6 +747,128 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
         }
 
         [Fact]
+        public async Task CanReceiveDataEvenIfUserThrowsInConnectedEvent()
```
davidfowl: Add a test for a synchronous exception.

moozzyk:
:up::date:

BrennanConroy:
This PR should be on alpha1. You can change it by "editing" the title

moozzyk:
ceff4bd7da2527dbb463019691d7176b826550a7

Issue #821, opened 2017-09-06
Default log output of SignalR js client is too much


davidfowl:
Seems like we log a bunch of things by default:

![image](https://user-images.githubusercontent.com/95136/30100893-384addc0-929f-11e7-840a-7147f9cad803.png)

We shouldn't log the received messages as information that could be super noisy.

davidfowl:
This should be fixed for the alpha.

/cc @moozzyk @muratg

moozzyk:
I will add Trace level and will log these as Trace.

PR #824, opened 2017-09-06
Logging traffic at Trace level

moozzyk:
Fixes: #821

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
@@ -11,7 +11,13 @@ describe('hubConnection', function () {
 
             it('can invoke server method and receive result', function (done) {
                 var message = "你好，世界！";
-                var hubConnection = new signalR.HubConnection(new signalR.HttpConnection(TESTHUBENDPOINT_URL, { transport: transportType }), { protocol: protocol});
+
+                var options = {
+                    transport: transportType,
+                    protocol: protocol,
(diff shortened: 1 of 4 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/hubConnectionTests.js
```diff
@@ -62,7 +74,12 @@ describe('hubConnection', function () {
 
             it('rethrows an exception from the server when invoking', function (done) {
                 var errorMessage = "An error occurred.";
-                var hubConnection = new signalR.HubConnection(new signalR.HttpConnection(TESTHUBENDPOINT_URL, { transport: transportType }), { protocol: protocol});
+                var options = {
+                    transport: transportType,
+                    protocol: protocol,
+                    logging: signalR.LogLevel.Trace
+                };
+                var hubConnection = new signalR.HubConnection(new signalR.HttpConnection(TESTHUBENDPOINT_URL, options), options);
```
moozzyk: I hate how hard (and weird (look at the `options), options);`) it is to create hubConnection in ts/js. 
Should we change the API so that IConnection is part of options and if not passed we would create an HttpConnection for the user?

davidfowl: We need the builder api on the js client too...

moozzyk: ... or blur the separation a little bit as per https://github.com/aspnet/SignalR/issues/825

PR #830, opened 2017-09-07
Make sure long polling transport can survive http client timeout

### LINKED ISSUE #829: The C# client running longpolling transport closes if the server does not send any messages for 100+ secs
Looks like the default [HttpClient.Timeout](https://msdn.microsoft.com/en-us/library/system.net.http.httpclient.timeout(v=vs.118).aspx) for http request is 100 secs. The HttpClient will then cancel the request. The default timeout for poll requests is 110 seconds. As a result the client cannot be idle for more than 100 seconds otherwise HttpClient timeout kicks in and cancels the request which then will stop the transport and as a result the  client. 

We should:
- re-poll on the client if HttpClient timeout occurs instead of stopping the transport
- set the HttpClient timeout to be greater than the default timeout for poll requests on the server side (e.g. 120 seconds) to make the server close poll requests cleanly

moozzyk:
This should suffice to fix #829 for alpha1. Post alpha we need to decide if we want to send the poll timeout value from the server in the response to the OPTIONS request and, if this is the case, how to pass it to the TransportFactory


test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -437,5 +437,53 @@ public class LongPollingTransportTests
...
+                    if (Interlocked.Increment(ref numPolls) < 3)
+                    {
+                        throw new OperationCanceledException();
+                    }
+
+                    completionTcs.SetResult(null);
+                    return ResponseUtils.CreateResponse(HttpStatusCode.OK);
+                });
+
+            // Task transportActiveTask;
```
muratg: ?

test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -437,5 +437,53 @@ public class LongPollingTransportTests
...
+
+                    var completedTask = await Task.WhenAny(completionTcs.Task, longPollingTransport.Running).OrTimeout();
+                    Assert.Equal(completionTcs.Task, completedTask);
+                }
+                finally
+                {
+                    await longPollingTransport.StopAsync();
+                }
+
+                // await transportActiveTask.OrTimeout();
```
muratg: ?

moozzyk: oops - leftovers. Will remove.

PR #839, opened 2017-09-10
Renamed private fields according to coding guidelines

MrFrank75:
Hi guys, if of any interest, there is this small renaming needed in the ServerFixture class.
Of course just a minimal change, feel free to drop the PR if is not relevant for you.
Cheers
Livio


moozzyk:
Thanks!

Issue #837, opened 2017-09-09
No easy way to get the HttpContext from a HubConnectionContext


davidfowl:
We added GetHttpContext on ConnectionContext but then we changed the Hub to expose HubConnectionContext instead of ConnectionContext:

https://github.com/aspnet/SignalR/blob/eec4b33cff09382285766844b8054d3e65fa1131/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionContextExtensions.cs

This makes it harder to get at the HttpContext from inside of the hub. We should expose GetHttpContext on HubConnectionContext.

davidfowl:
We should fix this for the alpha /cc @moozzyk @muratg

BrennanConroy:
https://github.com/aspnet/SignalR/commit/393ab6a4f03b68d2a55336fc7ccf8c89f7c546ae

PR #849, opened 2017-09-11
Add extension method to get HttpContext on HubConnectionContext

BrennanConroy:
https://github.com/aspnet/SignalR/issues/837

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/HttpConnectionContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR/HttpConnectionContextExtensions.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Sockets.Http.Features;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public static class HttpConnectionContextExtensions
+    {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR/HttpConnectionContextExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Sockets.Http.Features;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public static class HttpConnectionContextExtensions
+    {
+        public static HttpContext GetHttpContext(this HubConnectionContext connection)
+        {
+            return connection.Features.Get<IHttpContextFeature>().HttpContext;
```
moozzyk: `.?` in case the feature is not there?

src/Microsoft.AspNetCore.SignalR/HttpConnectionContextExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Sockets.Http.Features;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public static class HttpConnectionContextExtensions
+    {
+        public static HttpContext GetHttpContext(this HubConnectionContext connection)
```
moozzyk: Can you add a test for this?

Issue #845, opened 2017-09-11
JS client does not work with IE9 and IE10


muratg:
@moozzyk has the details.

moozzyk:
Fixed in 54c71c4c103e0406e5f6b8ee39c1a3e5e93b61de

PR #851, opened 2017-09-11
Replacing Cache-Control header with nonce

moozzyk:
Makes long polling work on IE10 and IE9

Fixes: #845

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -247,12 +247,12 @@ export class LongPollingTransport implements ITransport {
         }
 
         this.pollXhr = pollXhr;
-        this.pollXhr.open("GET", url, true);
+
+        this.pollXhr.open("GET", `${url}&_=${Date.now()}`, true);
         if (transferMode === TransferMode.Binary) {
             this.pollXhr.responseType = "arraybuffer";
         }
-        // IE caches xhr requests
-        this.pollXhr.setRequestHeader("Cache-Control", "no-cache");
+
         // TODO: consider making timeout configurable
         this.pollXhr.timeout = 120000;
         this.pollXhr.send();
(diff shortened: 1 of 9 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/js/connectionTests.js
```diff
@@ -4,30 +4,32 @@
...
-            expect(error).toBeUndefined();
-            done();
-        };
-
-        connection.start().then(function () {
-            connection.send(message);
-        }).catch(function (e) {
-            fail();
-            done();
+    if (typeof WebSocket !== 'undefined') {
```
BrennanConroy: Can this be removed since you added the same check to `getTransportTypes`?

BrennanConroy: Oh nevermind

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -247,12 +247,12 @@ export class LongPollingTransport implements ITransport {
         }
 
         this.pollXhr = pollXhr;
-        this.pollXhr.open("GET", url, true);
+
+        this.pollXhr.open("GET", `${url}&_=${Date.now()}`, true);
```
BrennanConroy: Is there ever a case where someone would use the LongPolling class directly instead of from HttpConnection? Then we wouldn't be guaranteed that a query part already exists

moozzyk: Each poll request needs a connection id which guarantees that there is query part.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -247,12 +247,12 @@ export class LongPollingTransport implements ITransport {
...
         if (transferMode === TransferMode.Binary) {
             this.pollXhr.responseType = "arraybuffer";
         }
-        // IE caches xhr requests
-        this.pollXhr.setRequestHeader("Cache-Control", "no-cache");
```
davidfowl: Keep this?

moozzyk: why? it's not needed anymore and because the url is now different every time nothing on the way should cache anything either.

davidfowl: OK

PR #857, opened 2017-09-12
Reduce C# client logging and add HubProtocol log

BrennanConroy:
@moozzyk @davidfowl


src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -143,6 +143,8 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
 
                                 connection.ProtocolReaderWriter = new HubProtocolReaderWriter(protocol, dataEncoder);
 
+                                _logger.LogInformation("Using HubProtocol '{protocol}'.", protocol.Name);
```
davidfowl: Server and client.

Issue #863, opened 2017-09-12
MsgPack doesn't include property names in packed objects


BrennanConroy:
```c#
public class Stock
{
  public string Symbol {}
  public decimal DayOpen {}
  public decimal DayLow {}
  public decimal DayHigh {}
  public decimal LastChange {}
  public decimal Change {}
  public double PercentChange {}
  public decimal Price {}
}
```

Gets decoded without the names like `Symbol` and `DayOpen`. Looks like it decodes in alphabetical order to deserialize accurately.

moozzyk:
Fixed in dbe4b4025bb9518f816f4762bf53c1f8ae02407e

Issue #862, opened 2017-09-12
Investigate parsing strategy in MsgPack related to ReadOnlyCollections


BrennanConroy:
I had a streaming hub method returning `IReadOnlyCollection<T>` when using MessagePack a parsing exception occurs. We need to investigate this and figure out how to proceed.

```
System.Runtime.Serialization.SerializationException: Type 'System.Collections.ObjectModel.ReadOnlyCollection`1[Microsoft.AspNet.SignalR.StockTicker.Stock]' does
 not have both of default (parameterless) public constructor and  public constructor with an Int32 parameter.
   at MsgPack.Serialization.SerializationExceptions.ThrowTargetDoesNotHavePublicDefaultConstructorNorInitialCapacity(Type type)
   at MsgPack.Serialization.UnpackHelpers.DetermineCollectionSerializationStrategy(Type instanceType, Boolean allowAsymmetricSerializer)
   at MsgPack.Serialization.ReflectionSerializers.ReflectionSerializerHelper.CreateCollectionSerializer[T](SerializationContext context, Type targetType, CollectionTraits traits, PolymorphismSchema schema)
   at MsgPack.Serialization.MessagePackSerializer.CreateReflectionInternal[T](SerializationContext context, Type concreteType, PolymorphismSchema schema)
   at MsgPack.Serialization.MessagePackSerializer.CreateInternal[T](SerializationContext context, PolymorphismSchema schema)
   at MsgPack.Serialization.SerializationContext.GetSerializer[T](Object providerParameter)
   at MsgPack.Serialization.SerializationContext.SerializerGetter.Get(SerializationContext context, Type targetType, Object providerParameter)
   at MsgPack.PackerUnpackerExtensions.PackObject(Packer source, Object value)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.MessagePackHubProtocol.WriteStreamingItemMessage(StreamItemMessage streamItemMessage, Packer packer, Stream output) in C:\Github\SignalR\src\Microsoft.AspNetCore.SignalR.Common\Internal\Protocol\MessagePackHubProtocol.cs:line 164
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.MessagePackHubProtocol.WriteMessageCore(HubMessage message, Stream output) in C:\Github\SignalR\src\Microsoft.AspNetCore.SignalR.Common\Internal\Protocol\MessagePackHubProtocol.cs:line 139
[...]

PR #864, opened 2017-09-13
MessagePack Serialization fixes

moozzyk:
* serializing user objects as maps (#863)
* enabling serializing objects that potentially cannot be deserialized (#862)


src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -20,10 +20,18 @@ public class MessagePackHubProtocol : IHubProtocol
         private const int VoidResult = 2;
         private const int NonVoidResult = 3;
 
+        private static readonly SerializationContext _serializationContext;
```
davidfowl: Is this like Json formatting options or is it just for our part of the protocol?

moozzyk: This is part of the protocol but we will probably need Json like formatting on top of that. This setting just says to write key/property value pairs instead of just values.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -20,10 +20,18 @@ public class MessagePackHubProtocol : IHubProtocol
...
+
         public string Name => "messagepack";
 
         public ProtocolType Type => ProtocolType.Binary;
 
+        static MessagePackHubProtocol()
+        {
+            _serializationContext = new SerializationContext { SerializationMethod = SerializationMethod.Map };
+            _serializationContext.CompatibilityOptions.AllowAsymmetricSerializer = true;
```
davidfowl: Add comments about why this works?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
davidfowl: Does this also make C# less efficient?

moozzyk: The payload is bigger but without it is impossible to deserialize the payload if you don't have a strong type that can be reflected on. Note that this is only for custom/structural objects in arguments and result types.

PR #865, opened 2017-09-13
Appending version to browser javascript file names

moozzyk:


moozzyk:
We should somehow automate this post alpha.

PR #869, opened 2017-09-13
Made tweaks to .NET client logging API

davidfowl:
- Removed ConnectionFactoryDelegate and used Func<IConnection>
- Changed WithLogger that accepts ILoggerFactory to WithLoggerFactory
- Made UseLogger configure the existing ILoggerFactory or create a LoggerFactory
- Add support for setting the log level for console logs
- Updated tests

### DIFF
--- a/samples/ClientSample/HubSample.cs
+++ b/samples/ClientSample/HubSample.cs
@@ -31,16 +31,11 @@ namespace ClientSample
             baseUrl = string.IsNullOrEmpty(baseUrl) ? "http://localhost:5000/default" : baseUrl;
 
             Console.WriteLine("Connecting to {0}", baseUrl);
-            var httpConnection = new HttpConnection(new Uri(baseUrl));
-            var connection = new HubConnectionBuilder()
-                .WithUrl(baseUrl)
-                .WithConsoleLogger()
-                .Build();
+            HubConnection connection = await ConnectAsync(baseUrl);
+            Console.WriteLine("Connected to {0}", baseUrl);
 
             try
             {
-                await connection.StartAsync();
-                Console.WriteLine("Connected to {0}", baseUrl);
 
                 var cts = new CancellationTokenSource();
                 Console.CancelKeyPress += (sender, a) =>
@@ -91,6 +86,27 @@ namespace ClientSample
                 await connection.DisposeAsync();
             }
             return 0;
+        }
+
+        private static async Task<HubConnection> ConnectAsync(string baseUrl)
+        {
+            // Keep trying to until we can start
+            while (true)
+            {
+                var connection = new HubConnectionBuilder()
+                                .WithUrl(baseUrl)
+                                .WithConsoleLogger(LogLevel.Trace)
+                                .Build();
+                try
+                {
+                    await connection.StartAsync();
+                    return connection;
+                }
+                catch (Exception)
+                {
+                    await Task.Delay(1000);
+                }
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/ConnectionFactoryDelegate.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/ConnectionFactoryDelegate.cs
@@ -1,9 +1,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using Microsoft.AspNetCore.Sockets.Client;
-
-namespace Microsoft.AspNetCore.SignalR.Client
-{
-    public delegate IConnection ConnectionFactoryDelegate();
-}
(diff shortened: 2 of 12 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -287,7 +287,7 @@ public static IEnumerable<object[]> MessageSizesData
                 var connection = new HubConnectionBuilder()
                         .WithUrl(new Uri(url))
                         .WithTransportType(transportType)
```
moozzyk: Not related but I was wondering if we should change WithTransportType to just WithTransport

davidfowl: Want me to make that change?

davidfowl: I'm also wondering if With should be Use to match the rest of our stack 😄

samples/ClientSample/HubSample.cs
```diff
@@ -31,16 +31,11 @@ internal static void Register(CommandLineApplication app)
             baseUrl = string.IsNullOrEmpty(baseUrl) ? "http://localhost:5000/default" : baseUrl;
 
             Console.WriteLine("Connecting to {0}", baseUrl);
-            var httpConnection = new HttpConnection(new Uri(baseUrl));
-            var connection = new HubConnectionBuilder()
-                .WithUrl(baseUrl)
-                .WithConsoleLogger()
-                .Build();
+            HubConnection connection = await ConnectAsync(baseUrl);
```
moozzyk: nit: var

samples/ClientSample/HubSample.cs
```diff
@@ -92,5 +87,26 @@ internal static void Register(CommandLineApplication app)
...
+                                .WithConsoleLogger(LogLevel.Trace)
+                                .Build();
+                try
+                {
+                    await connection.StartAsync();
+                    return connection;
+                }
+                catch (Exception)
+                {
+                    await Task.Delay(1000);
```
moozzyk: `Console.WriteLine("Retrying...")`?
Also should we try a different transport when retrying?

src/Microsoft.AspNetCore.SignalR.Client.Core/ConnectionFactoryDelegate.cs
```diff
@@ -1,9 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using Microsoft.AspNetCore.Sockets.Client;
-
-namespace Microsoft.AspNetCore.SignalR.Client
-{
-    public delegate IConnection ConnectionFactoryDelegate();
```
moozzyk: Out of curiosity - is it being removed because we just expose Func everywhere else?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -25,12 +26,20 @@ public static IHubConnectionBuilder WithMessagePackProtocol(this IHubConnectionB
             return hubConnectionBuilder.WithHubProtocol(new MessagePackHubProtocol());
         }
 
-        public static IHubConnectionBuilder WithLogger(this IHubConnectionBuilder hubConnectionBuilder, ILoggerFactory loggerFactory)
+        public static IHubConnectionBuilder WithLoggerFactory(this IHubConnectionBuilder hubConnectionBuilder, ILoggerFactory loggerFactory)
         {
-            hubConnectionBuilder.AddSetting(HubConnectionBuilderDefaults.LoggerFactoryKey, loggerFactory);
+            hubConnectionBuilder.AddSetting(HubConnectionBuilderDefaults.LoggerFactoryKey,
+                loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory)));
```
moozzyk: In general we allow passing null logger factory to HttpConnection and HubConnection so not sure why we need to throw here?

samples/ClientSample/HubSample.cs
BrennanConroy: This is mostly useful for when you launch the C# client and haven't started the server yet, don't think we should be changing transports.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
moozzyk: https://github.com/aspnet/SignalR/pull/812#discussion_r136641177 but maybe it's fine.

samples/ClientSample/HubSample.cs
davidfowl: Yea, I just wanted to make that work so it would be easy to test.

src/Microsoft.AspNetCore.SignalR.Client.Core/ConnectionFactoryDelegate.cs
davidfowl: Yeap

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
davidfowl: I'm not sure it makes sense to allow a null logger factory, what would be the point of that

moozzyk: OK

PR #889, opened 2017-09-15
SignalR Core Logging

BrennanConroy:


src/Microsoft.AspNetCore.SignalR.Core/Internal/SignalRCoreLoggerExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+            LoggerMessage.Define<string, string>(LogLevel.Trace, 9, "{invocationId}: Streaming result of type '{resultType}'.");
+
+        private static readonly Action<ILogger, string, string, Exception> _sendingResult =
+            LoggerMessage.Define<string, string>(LogLevel.Trace, 10, "{invocationId}: Sending result of type '{resultType}'.");
+
+        private static readonly Action<ILogger, string, Exception> _failedInvokingHubMethod =
+            LoggerMessage.Define<string>(LogLevel.Error, 11, "Failed to invoke hub method '{hubMethod}'.");
+
+        private static readonly Action<ILogger, string, Exception> _hubMethodBound =
+            LoggerMessage.Define<string>(LogLevel.Debug, 12, "Hub method '{hubMethod}' is bound.");
```
moozzyk: I think this should at .Trace

PR #890, opened 2017-09-15
Ft winston typed hub context

moozzyk:
This is https://github.com/aspnet/SignalR/pull/783 rebased on the latest dev


moozzyk:
I will squash and merge. @davidfowl - do we want this for alpha2? I think someone was already asking about that...

moozzyk:
I propose we take it for alpha2
/cc @davidfowl

FTWinston:
Cool, thanks for the help @moozzyk

Issue #886, opened 2017-09-15
Add CancellationToken representing the lifetime of a specific HubConnection


davidfowl:
It should be possible for streaming results to detect the client going away without manually keeping track of the OnConnectedAsync/OnDisconnectedAsync calls (you can do it yourself today).

Issue #885, opened 2017-09-15
Add support for unsubscribing from an IObservable<T> streaming result


Tragetaschen:
This is likely related to #481

I have ported my `PersistentConnection` from SignalR 2.2 to a Hub method returning an `IObservable<T>`. By desing, this observable doesn't complete throughout the lifetime of the application.

Whenever a client disconnects (I call connection.stop() in TS), the subscribed observer doesn't unsubscribe and all of them just pile up in my observable. SignalR just loses the subscription:
https://github.com/aspnet/SignalR/blob/rel/1.0.0-alpha1/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs#L39
I've quickly hacked an unsubscribe tied to the Hub's `Dispose` method:

```cs
public class MyHub : Hub
{
    private readonly MyObservable myObservable;

    private IDisposable disposeSubscription;

    public MyHub(MyObservable myObservable) => this.myObservable = myObservable;

    public IObservable<object> Data() => new wrappingObservable(myObservable, this);

    protected override void Dispose(bool disposing)
    {
        disposeSubscription?.Dispose();
        base.Dispose(disposing);
    }

    private class wrappingObservable : IObservable<object>
    {
        private readonly IObservable<object> original;
        private readonly MyHub myHub;

        public wrappingObservable(
            IObservable<object> original,
            MyHub myHub
        )
        {
            this.original = original;
            this.myHub = myHub;
        }

        public IDisposable Subscribe(IObserver<object> observer)
        {
            var disposable = original.Subscribe(observer);
            myHub.disposeSubscription = disposable;
            return disposable;
        }
    }
}
```

moozzyk:
This is related to #481 indeed.

davidfowl:
We could actually have handled this without designing cancellation in the protocol. This is just something we missed.

davidfowl:
#481 is about cancellation separate from the connection lifetime. Right now we just leak unnecessarily.

moozzyk:
I think we should have a common mechanism of cancelling streaming which could be invoked in different ways
- on the server - automatically when the connection is stopped
- on the server - on demand by the user
- from the client - by sending a cancellation (which can actually be invoking a hub method that does no. 2 above and we don't need any special cancellation protocol)

davidfowl:
I don't agree that it should be common. It needs to map to whatever primitive makes sense for the data type.

> on the server - automatically when the connection is stopped

[...]

Issue #481, opened 2017-05-23
Cancellation in Streamed results


anurse:
We need to review how cancellation works in Streamed results. If the connection terminates, we need to be able to stop enumerating the sequence the user gave us, and ideally signal cancellation back to it in order to cancel any other async operations.

On a similar note, we may want to add support for passing a `CancellationToken` in to a Hub method.

grahamehorner:
@anurse +1 for passing a CancellationToken into a hub method.   question would this be a token from a client or a server side token?  could this be used by an administrator to terminate badly behaved SignalR hub streaming methods? or place a upper limit on time of execution etc.

moozzyk:
Need to expose a CancellationToken to the hub in some way. Needs some design/discussion.

davidfowl:
> On a similar note, we may want to add support for passing a CancellationToken in to a Hub method.

I think we should just expose it as a property on the Context instead of being a parameter.

anurse:
Agreed. If we were going to put it as an argument it would be confused with something the client could control.

PR #897, opened 2017-09-16
Added Cancellation support

davidfowl:
- Added ConnectionAbortedToken to the HubConnectionContext. This allows
arbitrary code to access a handle that represents the connection lifetime
without handling OnDisconnectedAsync on the hub itself.
- Expose Abort on HubConnectionContext to allow server side methods to
abort the connection.
- Use the Abort to stop the main loop when unexpected invocation errors happen.
- Use the connection aborted token as unsubscribe from the IObservable and to complete
the IAsyncEnumerator for streaming results.

#886 #481 #885

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -86,16 +86,15 @@ namespace Microsoft.AspNetCore.SignalR
             try
             {
                 connection._connectionAbortedTokenSource.Cancel();
+
+                // Communicate the fact that we're finished triggering abort callbacks
+                connection._abortCompletedTcs.TrySetResult(null);
             }
-            catch
+            catch (Exception ex)
             {
                 // TODO: Should we log if the cancellation callback fails? This is more preventative to make sure
                 // we don't end up with an unobserved task
-            }
-            finally
-            {
-                // Communicate the fact that we're finished triggering abort callbacks
-                connection._abortCompletedTcs.TrySetResult(null);
+                connection._abortCompletedTcs.TrySetException(ex);
             }
         }
     }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -1,8 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Runtime.ExceptionServices;
 using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
 using System.Threading.Tasks.Channels;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.SignalR.Features;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -40,5 +53,50 @@ public HubConnectionContext(WritableChannel<HubMessage> output, ConnectionContex
...
+        private static void AbortConnection(object state)
+        {
+            var connection = (HubConnectionContext)state;
+            try
+            {
+                connection._connectionAbortedTokenSource.Cancel();
+            }
+            catch
+            {
+                // TODO: Should we log if the cancellation callback fails? This is more preventative to make sure
```
anurse: "Should we log if?" YES. YES YOU SHOULD LOG IT. ALWAYS LOG IT. LOG ALL THE THINGS. If you're worried about spam, log to `Debug` or `Trace`. But always log it.

An exception from the abort callback should be an error IMO, as the user callback could always capture it and suppress it.

davidfowl: > An exception from the abort callback should be an error IMO, as the user callback could always capture it and suppress it.

Yes it's their code they can catch the exceptions.

> "Should we log if?" YES. YES YOU SHOULD LOG IT. ALWAYS LOG IT. LOG ALL THE THINGS. If you're worried about spam, log to Debug or Trace. But always log it.

I think we'll log from the HubEndpoint instead of logging from here.

anurse: Where we log it is negotiable, as long as we log it ;)

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -40,5 +53,49 @@ public HubConnectionContext(WritableChannel<HubMessage> output, ConnectionContex
...
+            try
+            {
+                connection._connectionAbortedTokenSource.Cancel();
+
+                // Communicate the fact that we're finished triggering abort callbacks
+                connection._abortCompletedTcs.TrySetResult(null);
+            }
+            catch (Exception ex)
+            {
+                // TODO: Should we log if the cancellation callback fails? This is more preventative to make sure
```
moozzyk: Log as debug or trace

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -258,7 +267,7 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
 
                                         // Don't wait on the result of execution, continue processing other
                                         // incoming messages on this connection.
-                                        var ignore = ProcessInvocation(connection, invocationMessage, cts, completion);
+                                        var ignore = ProcessInvocation(connection, invocationMessage);
```
moozzyk: nit: consider `_`

src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
```diff
@@ -21,32 +21,33 @@ internal static class AsyncEnumeratorAdapters
...
+        public static IAsyncEnumerator<object> FromObservable<T>(IObservable<T> observable, CancellationToken cancellationToken)
         {
             // TODO: Allow bounding and optimizations?
             var channel = Channel.CreateUnbounded<object>();
 
-            var subscription = observable.Subscribe(new ChannelObserver<T>(channel.Out, CancellationToken.None));
+            var subscription = observable.Subscribe(new ChannelObserver<T>(channel.Out, cancellationToken));
+
+            // Dispose the subscription when the token is cancelled
+            cancellationToken.Register(state => ((IDisposable)state).Dispose(), subscription);
```
moozzyk: good catch

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -43,6 +44,167 @@ public class HubEndpointTests
...
+            var serviceProvider = CreateServiceProvider();
+            var endPoint = serviceProvider.GetService<HubEndPoint<AbortHub>>();
+
+            using (var client = new TestClient())
+            {
+                var endPointTask = endPoint.OnConnectedAsync(client.Connection);
+
+                await client.InvokeAsync(nameof(AbortHub.Kill));
+
+                await endPointTask;
```
moozzyk: `.OrTimeout()` - everywhere where applicable

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -43,6 +44,167 @@ public class HubEndpointTests
...
+        public async Task AbortFromHubMethodForcesClientDisconnect()
+        {
+            var serviceProvider = CreateServiceProvider();
+            var endPoint = serviceProvider.GetService<HubEndPoint<AbortHub>>();
+
+            using (var client = new TestClient())
+            {
+                var endPointTask = endPoint.OnConnectedAsync(client.Connection);
+
+                await client.InvokeAsync(nameof(AbortHub.Kill));
```
moozzyk: `.OrTimeout()`

Issue #842, opened 2017-09-11
JS/TS client HubConnection.onClosed should be a method instead of a property


davidfowl:
It should be possible to wire up multiple handlers to the closed event. We could even consider using on instead of having special events. That *does* mean we would need reserved names but it would be less concepts.

Today:

```TS
let http = new signalR.HttpConnection(document.location.protocol + '//' + document.location.host + '/chat');
let connection = new signalR.HubConnection(http);

connection.onClosed = e => {
        
};
```

Proposal:

```TS
let http = new signalR.HttpConnection(document.location.protocol + '//' + document.location.host + '/chat');
let connection = new signalR.HubConnection(http);

connection.on('close', e => {
        
});
```

We could also change the nomenclature and use connect/disconnect like the server side does.

/cc @moozzyk

moozzyk:
I am not a fan of `on('close'` - it will get confused with regular hub methods.

davidfowl:
Yea, I get it. socket.io does something similar for disconnect.

davidfowl:
I think we should do the method change for all events in the .NET client as well. It's just a better API in general.

moozzyk:
I don't like using the string for the name of the method for close if we can have just a strongly typed method. I also don't think we are reducing the number of concepts, rather we are conflating concepts. We could do:
```
connection.onClose(e => {
});
```
if we wanted to. One thing that is not clear from this signature is what happens if invoke this method multiple times. With the property it was clear that you are replacing the handler with the method it is not clear if you are adding or replacing.

davidfowl:
It should match the on overload and it should be possible to subscribe multiple handlers. I think we should do the same for the .NET client as well.

moozzyk:
I am opposed to changing `connection.onClose` to `connection.on('close'...)`. In my mind this is similar to changing `connection.start()` to `connection.invoke('start'...)` which we hopefully are not doing.

davidfowl:
@moozzyk yea I think that's fine. Lets go with:

```JS
connection.closed(e => {
});
```

PR #914, opened 2017-09-19
Enabling multiple callbacks for HubConnection.onClose

moozzyk:
Fixes: #842

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
@@ -152,8 +152,8 @@ describe("Connection", () => {
                 return Promise.reject("");
             },
             stop(): void { },
-            onDataReceived: undefined,
-            onClosed: undefined,
+            onreceive: undefined,
+            onclose: undefined,
         }
 
         let options: IHttpConnectionOptions = {
@@ -249,8 +249,8 @@ describe("Connection", () => {
                 connect(url: string, requestedTransferMode: TransferMode): Promise<TransferMode> { return Promise.resolve(transportTransferMode); },
                 send(data: any): Promise<void> { return Promise.resolve(); },
                 stop(): void {},
-                onDataReceived: null,
-                onClosed: null,
+                onreceive: null,
+                onclose: null,
                 mode: transportTransferMode
             } as ITransport;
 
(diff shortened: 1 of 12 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -239,8 +238,10 @@ export class HubConnection {
         }
     }
 
-    set onClosed(callback: ConnectionClosed) {
-        this.connectionClosedCallback = callback;
+    onClosed(callback: ConnectionClosed) {
```
davidfowl: I prefer onclose

moozzyk: This is for consistency - we use camelCase for all APIs so it would be weird to use lowercase here. (Granted, most of methods have just one words so it is not as visible)

davidfowl: > (Granted, most of methods have just one words so it is not as visible)

Which makes that point moot 😄 . onclose looks more like other browser APIs as well.

https://www.tutorialspoint.com/html5/html5_websocket.htm

Looks more javascripty.

moozzyk: We already had this discussion and we decided to use camelCase because the old JS client [uses camelCase](https://github.com/SignalR/SignalR/blob/d463e0967edb416362c27b58e0a9515cbae698fe/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js#L54-L63). What I am trying to say is that if we now want event names to be lowercase we should probably change it everywhere (i.e. `IConnection` and `ITransport`) and not only here.

(I find it quite amusing you bring webocket API to make your point in a PR where we change the onClose property to a function since "it is a better API" while websocket do use properties and not functions for
 events 🤣)

davidfowl: 😄 Camel case looks so weird for these methods. I dunno what it is but it feels like it doesn't fit.

moozzyk: I am fine with changing it everywhere.

moozzyk:
Need to fix samples

moozzyk:
:up::date:

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
```diff
@@ -249,8 +249,8 @@ describe("Connection", () => {
                 connect(url: string, requestedTransferMode: TransferMode): Promise<TransferMode> { return Promise.resolve(transportTransferMode); },
                 send(data: any): Promise<void> { return Promise.resolve(); },
                 stop(): void {},
-                onDataReceived: null,
-                onClosed: null,
+                onreceive: null,
```
davidfowl: ondata?

PR #915, opened 2017-09-19
Progress towards unsubscribing from streaming messages

BrennanConroy:
This allows users to locally unsubscribe from streaming messages so their callbacks aren't called anymore.
Server still sends messages because we don't have a way of telling it to stop yet.

We also still need a way of completing individual subscriptions on a single stream.

Current usage:
```typescript
var stream = connection.stream("method");
stream.subscribe({
    next: function (item) {
        console.log(item);
    }
});

// do things

stream.complete();
```

~~Need to add tests.~~


BrennanConroy:
Needs work

PR #917, opened 2017-09-20
Enabling customizing serialization settings in MessagePack protocol

moozzyk:


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -23,7 +23,8 @@ public static IHubConnectionBuilder WithJsonProtocol(this IHubConnectionBuilder
 
         public static IHubConnectionBuilder WithMessagePackProtocol(this IHubConnectionBuilder hubConnectionBuilder)
```
davidfowl: Add overloads of this and JsonProtocol that take the settings so this can be configured on the client as well?

moozzyk:
We could. I did not add it because when resolving protocols we need to use the settings from hub options so MessagePackHubProtocol.CreateDefaultSerializationContext() regardless stays because it prevents from default settings between the client and the server going out of sync. (Interestingly the settings are always out of sync for JSON due to camelCase set on the server always but it JSON.NET does not care about this one too much)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
moozzyk: OK. Was on the fence if want it or not.

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -24,7 +24,7 @@ public IHubProtocol GetProtocol(string protocolName, HubConnectionContext connec
                 case "json":
                     return new JsonHubProtocol(JsonSerializer.Create(_options.Value.JsonSerializerSettings));
                 case "messagepack":
-                    return new MessagePackHubProtocol();
+                    return new MessagePackHubProtocol(_options.Value.MessagePackSerializationContext);
```
BrennanConroy: If user doesn't add `HubOptions` then this can be null and `MessagePackHubProtocol` wont have any serialization settings.

moozzyk: Seems to work just fine. [OptionsFactory](https://github.com/aspnet/Options/blob/97a81e4603e42d6d8ad3539ba3255cce2bc57fc4/src/Microsoft.Extensions.Options/OptionsFactory.cs#L30) creates options and we initialize the `MessagePackSerializationContext` property to a default value when HubOptions instance is created.

moozzyk:
:up::date:
- added default ctors to HubProtocol implementations
- added overloads to HubConnectionBuilder to pass additional configuration for serializers

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -18,14 +19,29 @@ public static IHubConnectionBuilder WithHubProtocol(this IHubConnectionBuilder h
 
         public static IHubConnectionBuilder WithJsonProtocol(this IHubConnectionBuilder hubConnectionBuilder)
         {
-            return hubConnectionBuilder.WithHubProtocol(new JsonHubProtocol(new JsonSerializer()));
+            return hubConnectionBuilder.WithHubProtocol(new JsonHubProtocol());
+        }
+
+        public static IHubConnectionBuilder WithJsonProtocol(this IHubConnectionBuilder hubConnectionBuilder, JsonSerializer jsonSerializer)
```
davidfowl: I think we need an overload that takes the settings only.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1017,6 +1020,46 @@ public void HubsCannotHaveOverloadedMethods()
         }
 
         [Fact]
+        public async Task HubOptionsCanUseCustomMessagePackSettings()
```
davidfowl: Add a test that verifies it's not null by default.

moozzyk:
:up::date: Added an overload taking JsonSerializerSettings

moozzyk:
⌚️

PR #918, opened 2017-09-20
Update bootstrappers

ryanbrandenburg:


moozzyk:
@ryanbrandenburg - Is this something that needs to be reviewed or we are just replacing this in all repos (and it already runs in some of the repos)?

PR #922, opened 2017-09-21
Remove extra slice

Drawaes:
Removed extra slice
Also removed the loop for the size test instead of just an if?


davidfowl:
Thanks!

PR #927, opened 2017-09-21
add names to EventIds used by logger extensions

### LINKED ISSUE #876: Use names in EventIds for logging
In https://github.com/aspnet/SignalR/blob/11cc57ee0e4ebc38c2cf5877b1df19e4e686f446/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs we use `int`s for the `eventId` parameter, but the `EventId` type that is used can actually also support holding a name. Names are easier to query and inspect in structure log storage. So, as an example, a `LoggerMessage.Define` should change to something like this:

```csharp
LoggerMessage.Define<...>(LogLevel.Trace, new EventId(1, "[Event Name]", "Preparing blocking invocation '{invocationId}' of '{target}', with return type '{returnType}' and {argumentCount} argument(s).");
```

(Event ID names don't need a prefix because `EventId` is only unique within a Logger, so there's no need to prefix it).

anurse:
fixes #876

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
```diff
@@ -514,7 +514,7 @@ public static void StoppingClient(this ILogger logger, string connectionId)
             }
         }
 
-        public static void ExceptionThrownFromEventHandler(this ILogger logger, string connectionId, string eventHandlerName, Exception exception)
+        public static void ExceptionThrownFromHandler(this ILogger logger, string connectionId, string eventHandlerName, Exception exception)
```
anurse: Literally every other method name matches the field name of the delegate, except this one, so it felt right to get them in alignment. Happy to align the other way if that makes sense :)

Issue #861, opened 2017-09-12
Exceptions thrown from Protocols on the Server-Side hang


BrennanConroy:
The exception is observed when the connection closes.

https://github.com/aspnet/SignalR/blob/30f16fcf4be46d118d34645a77afc31143525250/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs#L110
This line observes the exception. But only happens when the connection closes.

Related to https://github.com/aspnet/SignalR/issues/816

moozzyk:
Fixed with 02676956565e85d806c85e1235cb2998fd5a5caa

Issue #916, opened 2017-09-20
Log exceptions thrown from HubProtocolReaderWriter methods


AndriySvyryd:
These exceptions are quite common while configuring the protocol to serialize the data correctly and are hard to debug without adding catching code.

moozzyk:
These exceptions are not handled in our code so should bubble up and should be automatically logged. Are you not seeing this? Can you show example/repro?

moozzyk:
I think I just saw it with MessagePack where it failed to serialize an object of anonymous type due to:
```
Non-public type '<>f__AnonymousType0`2[System.String,System.String]' cannot be serialized.
```
AFAIR @BrennanConroy told me that they (MessagePack Cli) decided to throw for non public types.

AndriySvyryd:
@moozzyk That's exactly what I see when I invoke something like this on the Hub:
```C#
public async Task<object> GetState() => new { state: "foo" };
```
It's not logged.

AndriySvyryd:
Actually it does appear in the Debug provider, but without the message:
`Exception thrown: 'System.NotSupportedException' in MsgPack.dll`

However it doesn't appear at all in Console.

This is my config:
```JSON
  "Logging": {
    "IncludeScopes": false,
    "Debug": {
      "LogLevel": {
        "Default": "Warning",
        "System": "Information",
        "Microsoft": "Information"
      }
    },
    "Console": {
      "LogLevel": {
        "Default": "Warning",
        "System": "Information",
        "Microsoft": "Information"
      }
    }
  }
```

moozzyk:
@AndriySvyryd - Thanks for the details. I only got my message because I turned breaking on all exceptions. If this is MsgPack that catches it and turns into NotSupportedException without an error message we may not be able to do anything on our side. I will need to investigate to see what happens and maybe open an issue/send a PR to MsgPack to improve the experience/debugability if needed.

AndriySvyryd:
@moozzyk It's not MsgPack. I see the exception message correctly in Diagnostic Tools pane.

moozzyk:
Need to dig more - searched quickly and I don't see any place where we throw the NotSupportedException without a message.
```
C:\source\SignalR-Core\src\Microsoft.AspNetCore.SignalR.Core\Internal\DefaultHubProtocolResolver.cs(29):                    throw new NotSupportedException($"The protocol '{protocolName ?? "(null)"}' is not supported.");
[...]

PR #930, opened 2017-09-22
Exceptions thrown during writing should close the connection

moozzyk:
We need to close the connection if there is an exception when writing to the transport on the server side. Currently if an exception happens it leaves the connection in an unsable state - after the exception no messages from the server will be sent to the client because the writing loop is terminated. Ignoring the message could cause hangs on the client side since we can fail while writing a completion message. In this case if the client is awaiting the invocation it will hang because the task will never be completed.

Fixes: #861 and #916

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -76,19 +76,26 @@ namespace Microsoft.AspNetCore.SignalR
             var protocolReaderWriter = connectionContext.ProtocolReaderWriter;
             async Task WriteToTransport()
             {
-                while (await output.In.WaitToReadAsync())
+                try
                 {
-                    while (output.In.TryRead(out var hubMessage))
+                    while (await output.In.WaitToReadAsync())
                     {
-                        var buffer = protocolReaderWriter.WriteMessage(hubMessage);
-                        while (await connection.Transport.Out.WaitToWriteAsync())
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -22,6 +22,7 @@
 using Newtonsoft.Json.Serialization;
 using Xunit;
 using Microsoft.AspNetCore.Hosting;
+using System.Runtime.Serialization;
```
davidfowl: nit: Sort usings

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -76,20 +76,27 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
+                                {
+                                    break;
+                                }
                             }
                         }
                     }
                 }
+                catch (Exception ex)
+                {
+                    connectionContext.Abort(ex);
```
davidfowl: Came in handy 😄

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
moozzyk: VS lost the setting again...

davidfowl:
@moozzyk does the error get logged?

moozzyk:
Yes, the error is logged.

Issue #402, opened 2017-04-18
Should the Sockets and SignalR protocol be subprotocols in WebSockets


davidfowl:
Should we set the subprotocol when talking over websockets?

/cc @anurse

anurse:
Yes, I think we should, just to make sure we're clear.

davidfowl:
Related to https://github.com/aspnet/SignalR/issues/520

moozzyk:
- The client should send a versioned subprotocol name for SignalR.
- Endpoint should be able to read the subprotocol from metadata

davidfowl:
@moozzyk how do you feel about this? This requires some back and forth between the endpoint and the transport.

moozzyk:
I think it depends on how much of _back and forth_ is required. Not required but would be nice. I wonder if it could be useful at the http connection level for IoT scenarios.

anurse:
What if this was just an EndPoint option and `.MapHub` configured it?

davidfowl:
I'm not sure you want to do it that globally instead of per connection.

anurse:
My thought is you'd have a list of subprotocols on the Endpoint options that are allowed, and the transport would negotiate it against the list that the client presents in it's request. The actual subprotocol decision is per-connection.

Also, we could take a page out of the Auth Handlers and provide events on the Endpoint options for custom subprotocol negotiation.

PR #933, opened 2017-09-23
Initial support for websocket subprotocols

davidfowl:
- Exposes a SubProtocol property on WebSocketOptions that picks the
protocol for all connections on the end point.
- This is required for things like mqtt over websockets (the SubProtocol in
this case is something like mqtt or mqttv3.1)
- Added test

#402 

PS: We could look at making this a Func that selects the subprotocol based on the requested protocols but this seems like it could be a start.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
@@ -47,7 +47,7 @@ namespace Microsoft.AspNetCore.Sockets.Internal.Transports
         {
             Debug.Assert(context.WebSockets.IsWebSocketRequest, "Not a websocket request");
 
-            using (var ws = await context.WebSockets.AcceptWebSocketAsync())
+            using (var ws = await context.WebSockets.AcceptWebSocketAsync(_options.SubProtocol))
             {
                 _logger.SocketOpened(_connection.ConnectionId);
 
--- a/test/Microsoft.AspNetCore.Sockets.Tests/Microsoft.AspNetCore.Sockets.Tests.csproj
+++ b/test/Microsoft.AspNetCore.Sockets.Tests/Microsoft.AspNetCore.Sockets.Tests.csproj
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.0</TargetFrameworks>
+    
+    <RuntimeIdentifier Condition="'$(TargetFramework)' != 'netcoreapp2.0'">win7-x64</RuntimeIdentifier>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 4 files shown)


test/Microsoft.AspNetCore.Sockets.Tests/Microsoft.AspNetCore.Sockets.Tests.csproj
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.0</TargetFrameworks>
+    
+    <RuntimeIdentifier Condition="'$(TargetFramework)' != 'netcoreapp2.0'">win7-x64</RuntimeIdentifier>
```
davidfowl: @natemcmaster should this be in all of our test projects aka Directory.Build.props

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -47,7 +47,7 @@ public WebSocketsTransport(WebSocketOptions options, Channel<byte[]> application
         {
             Debug.Assert(context.WebSockets.IsWebSocketRequest, "Not a websocket request");
 
-            using (var ws = await context.WebSockets.AcceptWebSocketAsync())
+            using (var ws = await context.WebSockets.AcceptWebSocketAsync(_options.SubProtocol))
```
moozzyk: Is this going to be used to OPTION-less websocket connections to negotiate hub protocol?

test/Microsoft.AspNetCore.Sockets.Tests/Microsoft.AspNetCore.Sockets.Tests.csproj
moozzyk: Do we need it now? Why didn't we need it before?

davidfowl: Because we never started the server before in this project. The other tests just build the pipeline.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
davidfowl: Nope. This is for lower level things that use the websocket protocol (like MQTT)

test/Microsoft.AspNetCore.Sockets.Tests/Microsoft.AspNetCore.Sockets.Tests.csproj
natemcmaster: The preferred way to do this is with HasRuntimeOutput. Example:

https://github.com/aspnet/SignalR/blob/d3ae3abe2acb97bf69143c85de9d4843e4d54e10/test/Microsoft.AspNetCore.SignalR.Tests/Microsoft.AspNetCore.SignalR.Tests.csproj#L7-L11

This should be set automatically in future versions of Microsoft.NET.Test.Sdk

PR #934, opened 2017-09-23
Block POST requests for websocket connections

davidfowl:


test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -150,6 +150,40 @@ public class HttpConnectionDispatcherTests
             }
         }
 
+        [Fact]
```
BrennanConroy: Does this need to be a `ConditionalFact` to disable the test on Win7?

anurse: No, this doesn't do a real WebSocket connection, it's just setting up a connection context and http context to run through the code path.

Issue #905, opened 2017-09-18
Application crashes on azure app service with signalr alpha


yosigolan:
setup:
server: asp.net core app version 2.0
            signalr 1.00 alpha
client: Angular 4.3

symptom: when deploying the server to azure app service, server crashes with 2 connections (the entire app crashes and restarts).
i couldn't receive any exception or error from azure app service, just: application stopped unexpectedly.

when using the same setup only with the old signalR (i know its not certified), the app does not crash.
i think its related to the number of connections, because i see it happens when i am connecting the second client and playing a bit.

sorry i dont have too much details, i tried getting some logs from azure but haven't succeeded after few hours...

davidfowl:
When you get some logs, let us know. Turn up the application logs to trace. This isn't very actionable feedback at the moment.

yosigolan:
Do you mean the application logs inside the asp.net core app? they are currently on Information for Microsoft. the last logs i saw from the signalr was connection closed and then the application crashes and restarts (on azure). if there was any warning on error i guess i would see it (since its above information).
i will try to reproduce it tomorrow locally, maybe than i will receive something.

Yosi

davidfowl:
Any logs you can get, crank up the default logging level to "Trace" and look at either app insights (what I recommend) or use the stdout log

yosigolan:
Hey @davidfowl , good news i think i got it.
i reproduced it locally. it happens only on iis (on kestrel works fine). when connecting with 2 clients an exception is thrown inside signalr (i see it stops in visual studio):
{System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.SignalR.DefaultHubLifetimeManager`1.<WriteAsync>d__11.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Travideo.AppsWebService.Hubs.TripPlanSystemViewConnector.<SetTripGeneralData>d__17.MoveNext() in C:\GitRepos\Backend\AppsWebService\AppsWebService\Hubs\TripPlanSystemViewConnector.cs:line 122}

once this exception is thrown the application is dead (this what caused it to crash on azure).
in the logs i get nothing... even when loglevel for Microsoft and system is debug.
let me know if you need anything else.
Yosi

[...]

PR #935, opened 2017-09-24
Check for null before sending to a specific connection id

davidfowl:
- Added some tests for the DefaultHubLifetimeManager

#905

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -301,6 +301,16 @@ namespace Microsoft.AspNetCore.SignalR.Redis
 
         public override async Task AddGroupAsync(string connectionId, string groupName)
         {
+            if (connectionId == null)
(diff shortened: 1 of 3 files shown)


BrennanConroy:
Add similar checks for `RedisHubLifetimeManager`?

davidfowl:
Which checks? The argument checks?

BrennanConroy:
Yeah, null checks

davidfowl:
Yea I originally didn't want to add them since I'd have to add them in every provider but that's fine.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -302,6 +312,16 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
 
         private async Task<bool> AddGroupAsyncCore(string connectionId, string groupName)
         {
+            if (connectionId == null)
```
BrennanConroy: For consistency, move this to `AddGroupAsync`

PR #946, opened 2017-09-26
Remove the params argument from IClientProxy

davidfowl:
- This allows passing arrays without having to explicitly ToArray() or AsEnumerable()
- Added overloads up to 10 arguments
- Added tests

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/IClientProxyExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/IClientProxyExtensions.cs
@@ -165,7 +165,7 @@ namespace Microsoft.AspNetCore.SignalR
         /// <param name="arg7">The seventh argument</param>
         /// <param name="arg8">The eigth argument</param>
         /// <param name="arg9">The ninth argument</param>
-        /// <param name="arg10">The ninth argument</param>
+        /// <param name="arg10">The tenth argument</param>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.SignalR.Core/IClientProxyExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <param name="arg1">The first argument</param>
+        /// <param name="arg2">The second argument</param>
+        /// <param name="arg3">The third argument</param>
+        /// <param name="arg4">The fourth argument</param>
+        /// <param name="arg5">The fifth argument</param>
+        /// <param name="arg6">The sixth argument</param>
+        /// <param name="arg7">The seventh argument</param>
+        /// <param name="arg8">The eigth argument</param>
+        /// <param name="arg9">The ninth argument</param>
+        /// <param name="arg10">The ninth argument</param>
```
BrennanConroy: This is the tenth argument :trollface:

BrennanConroy: Also, these usually have a period at the end but I'll ignore them for this change...

davidfowl: Good catch, I figure the copy pasta would catch up with me

PR #948, opened 2017-09-26
Replacing 8-byte-long length prefix with varint

moozzyk:


src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
```diff
@@ -12,16 +11,29 @@ public static class BinaryMessageFormatter
     {
         public static void WriteMessage(ReadOnlySpan<byte> payload, Stream output)
         {
-            // TODO: Optimize for size - (e.g. use Varints)
-            var length = sizeof(long);
-            var buffer = ArrayPool<byte>.Shared.Rent(length);
-            BufferWriter.WriteBigEndian<long>(buffer, payload.Length);
-            output.Write(buffer, 0, length);
-            ArrayPool<byte>.Shared.Return(buffer);
+            var lenBuffer = new byte[5];
```
BrennanConroy: Looks like a possible candidate for `stackalloc`? Maybe too early to think about that though

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Formatters/BinaryMessageFormatterTests.cs
```diff
@@ -75,5 +101,38 @@ public void WriteTextMessage(int offset, byte[] encoded, string payload)
...
+                BinaryMessageFormatter.WriteMessage(payload, ms);
+                var buffer = new ReadOnlyBuffer<byte>(ms.ToArray());
+                Assert.True(BinaryMessageParser.TryParseMessage(ref buffer, out var roundtripped));
+                Assert.Equal(payload, roundtripped.ToArray());
+            }
+        }
+
+        public static IEnumerable<object[]> RandomPayloads()
+        {
+            var r = new Random();
```
BrennanConroy: unused

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
davidfowl: This is a perfect case for stackalloc, soon we'll be using a new enough compiler to stackalloc `span<byte>` directly.

davidfowl:
@moozzyk you forgot to enable unsafe code.

davidfowl:
A few things:
- We need to update the protocol spec saying what algorithm we're using for variable sized ints.
- We should put a comment in the source code pointing to the spec.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
```diff
@@ -2,26 +2,38 @@
...
-        public static void WriteMessage(ReadOnlySpan<byte> payload, Stream output)
+        public unsafe static void WriteMessage(ReadOnlySpan<byte> payload, Stream output)
         {
-            // TODO: Optimize for size - (e.g. use Varints)
-            var length = sizeof(long);
-            var buffer = ArrayPool<byte>.Shared.Rent(length);
-            BufferWriter.WriteBigEndian<long>(buffer, payload.Length);
-            output.Write(buffer, 0, length);
-            ArrayPool<byte>.Shared.Return(buffer);
+            var lenBuffer = stackalloc byte[5];
```
davidfowl: A short english description of the pieces would be helpful.

moozzyk:
Weird I though I [enabled](https://github.com/aspnet/SignalR/pull/948/commits/40a4d8d8d36b8af1201c3866e692fbf3ebd3c1ad#diff-bb0ffdf19d2c1a4d96407da302323af5R10) unsafe and build passed locally. Will look into it.

davidfowl:
@moozzyk You used the UI and it ended up in the debug only section. *never* use the UI

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
```diff
@@ -10,36 +10,47 @@ public static class BinaryMessageParser
...
+            var length = 0U;
+            var numBytes = 0;
 
-            if (length > Int32.MaxValue)
+            byte byteRead;
+            do
             {
-                throw new FormatException("Messages over 2GB in size are not supported");
+                byteRead = buffer.Span.Slice(numBytes, sizeof(byte)).Read<byte>();
+                length = length | (((uint)(byteRead & 0x7f)) << (numBytes * 7));
```
davidfowl: Since we're bit twiddling, you can multiply by 7 like so `((n<<3) - n)` instead of using *.

moozzyk: Yes

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
```diff
@@ -10,36 +10,47 @@ public static class BinaryMessageParser
...
-            length = buffer.Span.Slice(0, sizeof(long)).ReadBigEndian<long>();
+            var length = 0U;
+            var numBytes = 0;
 
-            if (length > Int32.MaxValue)
+            byte byteRead;
+            do
             {
-                throw new FormatException("Messages over 2GB in size are not supported");
+                byteRead = buffer.Span.Slice(numBytes, sizeof(byte)).Read<byte>();
```
davidfowl: I'm worried about the efficiency of this.

moozzyk: We could try reading up Min(len, 5) bytes instead of slicing one byte at a time...

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
BrennanConroy: And if we know the range of `numBytes` (between 0 and 5) then you can use a lookup table for possibly better speed. These need to be tested before actually done though...

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
davidfowl: I don't like the efficiency of this...

src/Microsoft.AspNetCore.SignalR.Common/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -6,6 +6,10 @@
     <RootNamespace>Microsoft.AspNetCore.SignalR</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
davidfowl: Put this in the existing property group

src/Microsoft.AspNetCore.SignalR.Common/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -4,6 +4,7 @@
     <Description>Common serialiation primitives for SignalR Clients Servers</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
     <RootNamespace>Microsoft.AspNetCore.SignalR</RootNamespace>
+     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
davidfowl: nit: spacing

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
```diff
@@ -2,44 +2,70 @@
...
 
-            if (buffer.Length < sizeof(long))
+            if (buffer.IsEmpty)
             {
                 return false;
             }
 
-            // Read the length
-            length = buffer.Span.Slice(0, sizeof(long)).ReadBigEndian<long>();
+            // The payload starts with a length prefix encoded as a VarInt. VarInts use the most significant bit
```
davidfowl: Your future self thanks you.

davidfowl: We need a spec update as well.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
```diff
@@ -2,44 +2,70 @@
...
+            // as a marker whether the byte is the last byte of the VarInt or if it spans to the next byte. Bytes
+            // appear in the reverse order - i.e. the first byte contains the least significant bits of the value
+            // Examples:
+            // VarInt: 0x35 - %00111001 - the most significant bit is 0 so the value is %x0111001 i.e. 0x35 (53)
```
davidfowl: Isn't 0x35 - `00110101`?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
moozzyk: I will update the spec separately.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
moozzyk: Yes. (I am not fixing this today though - tomorrow will be a better day 😉 )

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -21,55 +21,71 @@ export namespace TextMessageFormat {
...
         let result: Uint8Array[] = [];
         let uint8Array = new Uint8Array(input);
-        // 8 - the length prefix size
-        for (let offset = 0; offset < input.byteLength;) {
+        const MaxLengthPrefixSize = 5;
+        const NumBitsToShift = [0, 7, 14, 21, 28 ];
 
-            if (input.byteLength < offset + 8) {
-                throw new Error("Cannot read message size")
+        for (let offset = 0; offset < input.byteLength;) {
```
davidfowl: Why doesn't this also look at byteLength and length?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
```diff
@@ -21,55 +21,71 @@ export namespace TextMessageFormat {
...
+        buffer.set(output, lenBuffer.length);
         return buffer.buffer;
     }
 
     export function parse(input: ArrayBuffer): Uint8Array[] {
         let result: Uint8Array[] = [];
         let uint8Array = new Uint8Array(input);
-        // 8 - the length prefix size
-        for (let offset = 0; offset < input.byteLength;) {
+        const MaxLengthPrefixSize = 5;
```
davidfowl: I don't think Pascal case is idiomatic typescript

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
```diff
@@ -2,44 +2,70 @@
...
+            // size bytes are missing
+            if ((byteRead & 0x80) != 0 && (numBytes < MaxLengthPrefixSize))
             {
-                throw new FormatException("Messages over 2GB in size are not supported");
+                return false;
             }
 
-            // Skip over the length
-            var remaining = buffer.Slice(sizeof(long));
+            if ((byteRead & 0x80) != 0 || (numBytes == MaxLengthPrefixSize && byteRead > 7))
```
davidfowl: Why not just check the length? `length > Int32.MaxValue`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
```diff
@@ -2,26 +2,41 @@
...
+            var lenNumBytes = 0;
+            var length = payload.Length;
+            do
+            {
+                lenBuffer[lenNumBytes] = (byte)(length & 0x7f);
+                length >>= 7;
+                if (length > 0)
+                {
+                    lenBuffer[lenNumBytes] = (byte)(lenBuffer[lenNumBytes] | 0x80);
```
davidfowl: Why not:

```
lenBuffer[lenNumBytes] |= 0x80;
```

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Formatters/BinaryMessageParserTests.cs
```diff
@@ -35,14 +58,35 @@ public void ReadBinaryMessage(byte[] encoded, byte[] payload)
             Assert.Equal(payload, message.ToArray());
         }
 
+        [Theory]
+        [InlineData(new byte[] { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF })]
```
davidfowl: Add one that has > 5 elements.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
```diff
@@ -2,26 +2,41 @@
...
-            ArrayPool<byte>.Shared.Return(buffer);
+            // This code writes length prefix of the message as a VarInt. Read the comment in
+            // the BinaryMessageParser.TryParseMessage for details.
+
+            var lenBuffer = stackalloc byte[5];
+            var lenNumBytes = 0;
+            var length = payload.Length;
+            do
+            {
+                lenBuffer[lenNumBytes] = (byte)(length & 0x7f);
```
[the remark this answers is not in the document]
moozzyk: I prefer hex.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
moozzyk: `ref` local 😍

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Formatters.ts
moozzyk: Need to double check - most likely it should.

moozzyk: I was baffled by this because the tests did not fail. Now I know why. For IE10 msgpack returns Buffer and not UInt8Array and Buffer does not have byteLength. We create an Uint8Array directly which does have byteLength.  They have a special flag TYPED_ARRAY_SUPPORT they use to decide what to return and the comment says:

/**
 * If `Buffer.TYPED_ARRAY_SUPPORT`:
 *   === true    Use Uint8Array implementation (fastest)
 *   === false   Use Object implementation (most compatible, even IE6)
 *
 * Browsers that support typed arrays are IE 10+, Firefox 4+, Chrome 7+, Safari 5.1+,
 * Opera 11.6+, iOS 4.2+.
 *
 * Due to various browser bugs, sometimes the Object implementation will be used even
 * when the browser supports typed arrays.
 *
 * Note:
 *
 *   - Firefox 4-29 lacks support for adding new properties to `Uint8Array` instances,
 *     See: https://bugzilla.mozilla.org/show_bug.cgi?id=695438.
 *
 *   - Chrome 9-10 is missing the `TypedArray.prototype.subarray` function.
 *
 *   - IE10 has a broken `TypedArray.prototype.subarray` function which returns arrays of
 *     incorrect length in some situations.

 * We detect these buggy browsers and set `Buffer.TYPED_ARRAY_SUPPORT` to `false` so they
 * get the Object implementation, which is slower but behaves correctly.
 */

We do use subarray in one place because IE does not support to slice so I need to dig a bit and understand how subarray is broken in IE10.

PR #949, opened 2017-09-26
Added overloads from 0-n arguments

davidfowl:
- Added 0 arg overload for InvokeAsync

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -131,32 +131,32 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -131,32 +131,32 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
             _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
         }
 
-        public async Task<ReadableChannel<object>> StreamAsync(string methodName, Type returnType, CancellationToken cancellationToken, params object[] args)
+        public async Task<ReadableChannel<object>> StreamAsync(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default(CancellationToken))
```
BrennanConroy: We can use just `default` now right?

davidfowl: are we c# 7.1 everywhere now?

BrennanConroy: Oh, maybe not alpha2

Issue #957, opened 2017-09-27
React Native support


SSkovboiSS:
Hi there

I'm developing a React Native mobile app and need a simple chat for it.  So i am interested in a way @aspnet/signalr-client can be integrated with React Native applications. And if not maybe, is there any workarounds? E. g. without using @aspnet/signalr-client.

davidfowl:
What exactly is the problem?

SSkovboiSS:
I've just setup sample project.

First thing is that msgpack5 depends on core node modules so seems like it's compatible with react-native only through special shim.

Secondly, when I've setup shim I ran into exception when trying to start connection:

Error
    at new HttpError (D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\@aspnet\signalr-client\dist\src\HttpError.js:6)
    at XMLHttpRequest.xhr.onerror (D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\@aspnet\signalr-client\dist\src\HttpClient.js:34)
    at XMLHttpRequest.dispatchEvent (D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\event-target-shim\lib\event-target.js:172)
    at XMLHttpRequest.setReadyState (D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\react-native\Libraries\Network\XMLHttpRequest.js:546)
    at XMLHttpRequest.__didCompleteResponse (D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\react-native\Libraries\Network\XMLHttpRequest.js:381)
    at D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\react-native\Libraries\Network\XMLHttpRequest.js:485
    at RCTDeviceEventEmitter.emit (D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\react-native\Libraries\EventEmitter\EventEmitter.js:181)
    at MessageQueue.__callFunction (D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\react-native\Libraries\BatchedBridge\MessageQueue.js:260)
    at D:\Programs\VSProjects\Empeek\kidscareapp-mobileapp\node_modules\react-native\Libraries\BatchedBridge\MessageQueue.js:101
[...]

PR #966, opened 2017-09-28
Checking if window.document defined before trying to resolve url

moozzyk:


.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
+{
```
moozzyk: I had to copy this from my gist too many times

moozzyk:
[Looks](https://github.com/aspnet/SignalR/issues/957#issuecomment-332648083) like when using React Native you can have `window` but without the `document` property.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpConnection.ts
```diff
@@ -162,7 +162,7 @@ export class HttpConnection implements IConnection {
             return url;
         }
 
-        if (typeof window === 'undefined') {
+        if (typeof window === 'undefined' || !window || !window.document) {
```
davidfowl: do we need the undefined check

moozzyk: It's more idiomatic. I guess `!window` will would work for both `null` and `undefined` but the semantic is a bit different.

Issue #939, opened 2017-09-24
Except functionality


Gurgen:
1. I want to send message to all authorized users
2. I want to send message to specified user except some connection ids

I mean, it will be excellent to add except clients to all clients filter, we have AllExcept() but can't add except to groups and other filters

Like this

T Group(string groupName, IReadOnlyList<string> excludedIds = null);
T User(string userId, IReadOnlyList<string> excludedIds = null);

BrennanConroy:
`User(string userId)` is to a single user so an except list doesn't make sense.
Groups one is questionable, since you could manage your groups differently so the exclude list is automatically there.

davidfowl:
Right the goal is to provide very basic building blocks so you can do more complex groupings using add/remove to group.

Gurgen:
Look. Client User-X connected to hub and I want to inform all other connected clients of User-X except connected user

Gurgen:
`public override Task OnConnectedAsync()
        {
            Clients.Client(Context.ConnectionId).InvokeAsync($"Hello {Context.User.Identity.Name}, you are connected!");
            Clients.User(Context.User.Identity.Name).InvokeAsync($"You are connected from another device!", except: Context.ConnectionId);
            return base.OnConnectedAsync();
        }`

KPixel:
@Gurgen This is not possible yes, but eventually, I think you should be able to find the connection ids of the other connected clients to send the second message, so the code could look like this:

```C#
public override async Task OnConnectedAsync()
{
    await Clients.Client(Context.ConnectionId).InvokeAsync($"Hello {Context.User.Identity.Name}, you are connected!");

    var otherClients = Connections.Where(c => c.User.Identity.Name == Context.User.Identity.Name
        && c.ConnectionId != Context.ConnectionId).Select(c => c.ConnectionId);

    await Clients.Clients(otherClients).InvokeAsync($"You are connected from another device!");
    await base.OnConnectedAsync();
}
```

I discussed an API for that here #683.

And note that this code would have to be written very differently if you need to scale out...

Gurgen:
OK, I understand that we can implement a Base class derived from Hub and implement there functionality as we want. I mean with this issue/improvement will be more simple working with hubs if we have except list of connections to every sub collection of connections.

Gurgen:
Also I created library to work with .net core signal r hubs from android, if you interested.
The repository is here, and at the moment supports only connections via websockets
[...]

PR #970, opened 2017-09-29
Exclude connection ids in users

Gurgen:


davidfowl:
We didn't approve this API, nor did you write a description in the pull request, nor does the code actually work. 

What feedback are you looking for?

Gurgen:
This is a implementation of this [Issue](https://github.com/aspnet/SignalR/issues/939).

added excludedConnectionIds to User() method with default null parameter, so if need we can put there a collection of connection ids which we want to exclude.

davidfowl:
> This is a implementation of this Issue.

Generally when you send a pull request you should write a description and that description should point to the issue you're solving. 

If you send a pull request as a demonstration of your idea that's fine, but please make it clear as part of the description. 

We have no plans to merge this.

Gurgen:
I did it for me and wanted to share with others. This is a very useful functionality not only for me, I think.

davidfowl:
> I did it for me and wanted to share with others. This is a very useful functionality not only for me, I think.

We'll think about the implications. I'm not a fan of adding exclude ids everywhere. We'll think about this more and come up with a recommendation. There might be a more general purpose feature here.

Gurgen:
I have another view of the selection of message recipients, something like a Builder where you can add, as well as exclude recipients. For example, even if you do as recommended by KPixel in [Issue](https://github.com/aspnet/SignalR/issues/683) there is still no freedom of choice. That's what I think about it.

davidfowl:
@Gurgen how is that different from a group?

Gurgen:
And if we want to send a few groups?

Gurgen:
Or to few groups and a few users? Or to users, who has some role?

Gurgen:
With builder we can build list of recipients.
`RecipientBuilder.AddGroup().ExcludeConnectionId().AddUser()....Build()`

Something like this

PR #979, opened 2017-10-03
Adding CancelInvocation message

BrennanConroy:
Opening this early to get feedback on the spec mostly and to see if I'm missing anything major with the feature.

Working on adding a couple more tests and reviewing my changes.

### DIFF
--- a/specs/HubProtocol.md
+++ b/specs/HubProtocol.md
@@ -354,13 +354,13 @@ Example - The following `Completion` message is a protocol error because it has 
 ### CancelInvocation Message Encoding
 A `CancelInvocation` message is a JSON object with the following properties
 
-* `type` - A `Number` with the literal value `4`, indicationg that this is a `CancelInvocation`.
+* `type` - A `Number` with the literal value `5`, indicationg that this is a `CancelInvocation`.
 * `invocationId` - A `String` encoding the `Invocation ID` for a message.
 
 Example
 ```json
 {
-    "type": 4,
+    "type": 5,
     "invocationId": "123"
 }
 ```
@@ -528,23 +528,23 @@ is decoded as follows:
 `CancelInvocation` messages have the following structure
 
 ```
-[4, InvocationId]
+[5, InvocationId]
 ```
 
-* `4` - Message Type - `4` indicates this is a `CancelInvocation` message
+* `5` - Message Type - `4` indicates this is a `CancelInvocation` message
 * InvocationId - A `String` encoding the Invocation ID for the message
 
 Example:
 
 The following payload:
 ```
-0x92 0x04 0xa3 0x78 0x79 0x7a
+0x92 0x05 0xa3 0x78 0x79 0x7a
 ```
 
 is decoded as follows:
 
 * `0x92` - 2-element array
-* `0x04` - `4` (Message Type `CancelInvocation` message)
+* `0x04` - `5` (Message Type `CancelInvocation` message)
 * `0xa3` - string of length 3 (InvocationId)
 * `0x78` - `x`
 * `0x79` - `y`
--- a/specs/HubProtocol.md
+++ b/specs/HubProtocol.md
@@ -403,7 +403,7 @@ is decoded as follows:
 
 * `0x95` - 5-element array
 * `0x01` - `1` (Message Type - `Invocation` message)
-* `0xa3` - string of length 3 (Target)
+* `0xa3` - string of length 3 (InvocationId)
 * `0x78` - `x`
 * `0x79` - `y`
 * `0x7a` - `z`
@@ -422,7 +422,9 @@ is decoded as follows:
 
 `StreamItem` messages have the following structure:
 
+```
 [2, InvocationId, Item]
+```
 
 * `2` - Message Type - `2` indicates this is a `StreamItem` message
 * InvocationId - A `String` encoding the Invocation ID for the message
@@ -439,7 +441,7 @@ is decoded as follows:
 
 * `0x93` - 3-element array
 * `0x02` - `2` (Message Type - `StreamItem` message)
-* `0xa3` - string of length 3 (Target)
+* `0xa3` - string of length 3 (InvocationId)
 * `0x78` - `x`
 * `0x79` - `y`
 * `0x7a` - `z`
@@ -474,7 +476,7 @@ is decoded as follows:
 
 * `0x94` - 4-element array
 * `0x03` - `3` (Message Type - `Result` message)
-* `0xa3` - string of length 3 (Target)
+* `0xa3` - string of length 3 (InvocationId)
 * `0x78` - `x`
 * `0x79` - `y`
 * `0x7a` - `z`
@@ -497,7 +499,7 @@ is decoded as follows:
 
 * `0x93` - 3-element array
 * `0x03` - `3` (Message Type - `Result` message)
-* `0xa3` - string of length 3 (Target)
+* `0xa3` - string of length 3 (InvocationId)
 * `0x78` - `x`
 * `0x79` - `y`
 * `0x7a` - `z`
@@ -514,12 +516,39 @@ is decoded as follows:
 
 * `0x94` - 4-element array
 * `0x03` - `3` (Message Type - `Result` message)
-* `0xa3` - string of length 3 (Target)
+* `0xa3` - string of length 3 (InvocationId)
 * `0x78` - `x`
 * `0x79` - `y`
 * `0x7a` - `z`
 * `0x03` - `3` (ResultKind - Non-Void result)
 * `0x2a` - `42` (Result)
+
+### CancelInvocation Message Encoding
+
+`CancelInvocation` messages have the following structure
+
+```
+[4, InvocationId]
+```
+
+* `4` - Message Type - `4` indicates this is a `CancelInvocation` message
+* InvocationId - A `String` encoding the Invocation ID for the message
+
+Example:
+
+The following payload:
+```
+0x92 0x04 0xa3 0x78 0x79 0x7a
+```
+
+is decoded as follows:
+
+* `0x92` - 2-element array
+* `0x04` - `4` (Message Type `CancelInvocation` message)
+* `0xa3` - string of length 3 (InvocationId)
+* `0x78` - `x`
+* `0x79` - `y`
+* `0x7a` - `z`
 
 ## Protocol Buffers (ProtoBuf) Encoding
 
--- a/specs/HubProtocol.md
+++ b/specs/HubProtocol.md
@@ -20,10 +20,11 @@ This document describes three encodings of the SignalR protocol: [JSON](http://w
 
 In the SignalR protocol, the following types of messages can be sent:
 
-* `Negotiation` Message - Sent by the client to negotiate the message format
+* `Negotiation` Message - Sent by the client to negotiate the message format.
 * `Invocation` Message - Indicates a request to invoke a particular method (the Target) with provided Arguments on the remote endpoint.
 * `StreamItem` Message - Indicates individual items of streamed response data from a previous Invocation message.
 * `Completion` Message - Indicates a previous Invocation has completed, and no further `StreamItem` messages will be received. Contains an error if the invocation concluded with an error, or the result if the invocation is not a streaming invocation.
+* `CancelInvocation` Message - Sent by the client to cancel a streaming invocation on the server.
 
 After opening a connection to the server the client must send a `Negotiation` message to the server as its first message. The negotiation message is **always** a JSON message and contains the name of the format (protocol) that will be used for the duration of the connection. If the server does not support the protocol requested by the client or the first message received from the client is not a `Negotiation` message the server must close the connection.
 
@@ -41,9 +42,9 @@ Example:
 
 ## Communication between the Caller and the Callee
 
-There a three kinds of interactions between the Caller and the Calle:
+There a three kinds of interactions between the Caller and the Callee:
 
-* Invocations - the Caller sends a message to the Calle and expects a message indicating that the invocation has been completed and optionally a result of the invocation
+* Invocations - the Caller sends a message to the Callee and expects a message indicating that the invocation has been completed and optionally a result of the invocation
 * Non-Blocking Invocations - the Caller sends a message to the Callee and does not expect any further messages for this invocation
 * Streaming Invocations - the Caller sends a message to the Callee and expects one or more results returned by the Callee followed by a message indicating the end of invocation
 
@@ -74,7 +75,7 @@ The SignalR protocol allows for multiple `StreamItem` messages to be transmitted
 
 On the Callee side, it is up to the Callee's Binder to determine if a method call will yield multiple results. For example, in .NET certain return types may indicate multiple results, while others may indicate a single result. Even then, applications may wish for multiple results to be buffered and returned in a single `Completion` frame. It is up to the Binder to decide how to map this. The Callee's Binder must encode each result in separate `StreamItem` messages, indicating the end of results by sending a `Completion` message.
 
-On the Caller side, the user code which performs the invocation indicates how it would like to receive the results and it is up the Caller's Binder to determine how to handle the result. If the Caller expects only a single result, but multiple results are returned, the Caller's Binder should yield an error indicating that multiple results were returned. However, if a Caller expects multiple results, but only a single result is returned, the Caller's Binder should yield that single result and indicate there are no further results.
+On the Caller side, the user code which performs the invocation indicates how it would like to receive the results and it is up the Caller's Binder to determine how to handle the result. If the Caller expects only a single result, but multiple results are returned, the Caller's Binder should yield an error indicating that multiple results were returned. However, if a Caller expects multiple results, but only a single result is returned, the Caller's Binder should yield that single result and indicate there are no further results. If the Caller wants to stop receiving `StreamItem` messages before the Callee sends a `Completion` message, the Caller can send a `CancelInvocation` message with the same `Invocation ID` used for the `Invocation` message that started the stream. It is possible to receive `StreamItem` messages or a `Completion` message after a `CancelInvocation` message has been sent, these can be ignored.
 
 ## Completion and results
 
@@ -221,6 +222,16 @@ S->C: StreamItem { Id = 42, Item = 4 }
 S->C: Completion { Id = 42, Error = "Ran out of data!" }
 ```
 
+### Streamed Result closed early (`Stream` example above)
+
+```
+C->S: Invocation { Id = 42, Target = "Stream", Arguments = [ 5 ] }
+S->C: StreamItem { Id = 42, Item = 0 }
+S->C: StreamItem { Id = 42, Item = 1 }
+C->S: CancelInvocation { Id = 42 }
+S->C: StreamItem { Id = 42, Item = 2} // This can be ignored
+```
+
 This should manifest to the Calling code as a sequence which emits `0`, `1`, `2`, `3`, `4`, but then fails with the error `Ran out of data!`.
 
 ### Non-Blocking Call (`NonBlocking` example above)
@@ -337,6 +348,20 @@ Example - The following `Completion` message is a protocol error because it has 
     "invocationId": 123,
     "result": 42,
     "error": "It didn't work!"
+}
+```
+
+### CancelInvocation Message Encoding
+A `CancelInvocation` message is a JSON object with the following properties
+
+* `type` - A `Number` with the literal value `4`, indicationg that this is a `CancelInvocation`.
+* `invocationId` - A `String` encoding the `Invocation ID` for a message.
+
+Example
+```json
+{
+    "type": 4,
+	"invocationId": 123
 }
 ```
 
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -138,7 +138,17 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
 
         private async Task<ReadableChannel<object>> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken)
         {
-            var irq = InvocationRequest.Stream(cancellationToken, returnType, GetNextId(), _loggerFactory, out var channel);
+            var irq = InvocationRequest.Stream(CancellationToken.None, returnType, GetNextId(), _loggerFactory, out var channel);
+            if (cancellationToken.CanBeCanceled)
```
anurse: See, the fact that we're plumbing it through as a `CancellationToken` even when streaming indicates to me that we should consider supporting it in regular invocations. We can talk about it after this PR, but just pointing out that these are basically the same.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -193,7 +203,7 @@ private Task InvokeCore(string methodName, InvocationRequest irq, object[] args,
             return SendInvocation(invocationMessage, irq);
         }
 
-        private async Task SendInvocation(InvocationMessage invocationMessage, InvocationRequest irq)
+        private async Task SendInvocation(HubMessage invocationMessage, InvocationRequest irq)
```
anurse: Rename to `SendHubMessage`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -42,4 +42,11 @@ public override string ToString()
             return $"Invocation {{ {nameof(InvocationId)}: \"{InvocationId}\", {nameof(NonBlocking)}: {NonBlocking}, {nameof(Target)}: \"{Target}\", {nameof(Arguments)}: [ {string.Join(", ", Arguments.Select(a => a?.ToString()))} ] }}";
         }
     }
+
+    public class CancelInvocationMessage : HubMessage
```
anurse: Put it in it's own file. I know it's small, but it'd be the only one that is in a different file if we didn't

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -54,6 +55,9 @@ public HubConnectionContext(WritableChannel<HubMessage> output, ConnectionContex
 
         public virtual WritableChannel<HubMessage> Output => _output;
 
+        // Currently used only for streaming methods
+        internal ConcurrentDictionary<string, CancellationTokenSource> ActiveRequests { get; } = new ConcurrentDictionary<string, CancellationTokenSource>();
```
anurse: `ActiveRequestCancellationSources`?

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -275,6 +275,20 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
+                                        //log
+                                        // check and cancel stream
+                                        if (connection.ActiveRequests.TryRemove(cancelInvocationMessage.InvocationId, out var cts))
+                                        {
+                                            cts.Cancel();
+                                        }
+                                        else
+                                        {
+                                            // stream can be canceled on the server while client is canceling stream
+                                            _logger.LogDebug("CancelInvocationMessage received unexpectedly.");
```
anurse: 👍

specs/HubProtocol.md
```diff
@@ -489,13 +516,40 @@ is decoded as follows:
...
 
+### CancelInvocation Message Encoding
+
+`CancelInvocation` messages have the following structure
+
+```
+[4, InvocationId]
+```
+
+* `4` - Message Type - `4` indicates this is a `CancelInvocation` message
```
moozzyk: I need 4 for StreamCompletion

anurse: You should start registering these with IANA :trollface:

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -138,7 +138,17 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
...
         {
-            var irq = InvocationRequest.Stream(cancellationToken, returnType, GetNextId(), _loggerFactory, out var channel);
+            var irq = InvocationRequest.Stream(CancellationToken.None, returnType, GetNextId(), _loggerFactory, out var channel);
+            if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken.Register(async () =>
+                {
+                    if (!_connectionActive.IsCancellationRequested)
+                    {
+                        await SendInvocation(new CancelInvocationMessage(irq.InvocationId), irq);
```
moozzyk: Rename SendInvocation to something else because now we send more than invocations.

moozzyk: Shouldn't we remove invocation from pending calls? If I understand correctly the server will no longer send completion so no one removes it...

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -144,20 +153,23 @@ private void WriteMessageCore(HubMessage message, Stream output)
             switch (message)
             {
                 case InvocationMessage invocationMessage:
-                    WriteInvocationMessage(invocationMessage, packer, output);
+                    WriteInvocationMessage(invocationMessage, packer);
```
moozzyk: :+1: - not sure why there were no warnings.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
BrennanConroy: Yeah, I just noticed this and was adding it

specs/HubProtocol.md
```diff
@@ -414,7 +441,7 @@ is decoded as follows:
 
 * `0x93` - 3-element array
 * `0x02` - `2` (Message Type - `StreamItem` message)
-* `0xa3` - string of length 3 (Target)
+* `0xa3` - string of length 3 (InvocationId)
```
moozzyk: :+1:

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -438,8 +452,23 @@ private bool IsChannel(Type type, out Type payloadType)
...
             try
             {
-                while (await enumerator.MoveNextAsync())
+                bool hasItem;
+                while (true)
                 {
+                    try
+                    {
+                        hasItem = await enumerator.MoveNextAsync();
+                    } catch (Exception)
```
moozzyk: nit: fomatting

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
anurse: I don't think that analysis is on by default, it was an FxCop rule back in the day.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -138,7 +138,17 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
 
         private async Task<ReadableChannel<object>> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken)
         {
-            var irq = InvocationRequest.Stream(cancellationToken, returnType, GetNextId(), _loggerFactory, out var channel);
+            var irq = InvocationRequest.Stream(CancellationToken.None, returnType, GetNextId(), _loggerFactory, out var channel);
+            if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken.Register(async () =>
+                {
```
davidfowl: Async void. Let’s discuss this. There’s also a closure allocation?

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -438,8 +452,23 @@ private bool IsChannel(Type type, out Type payloadType)
         {
             try
             {
-                while (await enumerator.MoveNextAsync())
+                bool hasItem;
```
moozzyk: this flag is a ugly...

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -275,6 +275,20 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                                         _ = ProcessInvocation(connection, invocationMessage);
                                         break;
 
+                                    case CancelInvocationMessage cancelInvocationMessage:
+                                        //log
+                                        // check and cancel stream
+                                        if (connection.ActiveRequests.TryRemove(cancelInvocationMessage.InvocationId, out var cts))
+                                        {
+                                            cts.Cancel();
```
moozzyk: no need to dispose?

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -450,30 +479,34 @@ private bool IsChannel(Type type, out Type payloadType)
             {
                 await SendMessageAsync(connection, CompletionMessage.WithError(invocationId, ex.Message));
             }
+            finally
+            {
+                if (connection.ActiveRequests.TryRemove(invocationId, out var cts))
```
moozzyk: Maybe you can add a method on connection to cancel the invocation - you now do it from two places...?

specs/HubProtocol.md
moozzyk: I registered with @BrennanConroy last week :)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
moozzyk: What happens if this throws?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -138,7 +138,28 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
 
         private async Task<ReadableChannel<object>> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken)
         {
-            var irq = InvocationRequest.Stream(cancellationToken, returnType, GetNextId(), _loggerFactory, out var channel);
+            var irq = InvocationRequest.Stream(CancellationToken.None, returnType, GetNextId(), _loggerFactory, out var channel);
+
+            if (cancellationToken.CanBeCanceled)
```
davidfowl: After a bit more thinking I think this is wrong for streaming. We need to cancel the stream itself, not the invocation to start streaming. It’s weird to have the token affect both the initial call and subsequent calls via the channel

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -275,6 +275,21 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                                         _ = ProcessInvocation(connection, invocationMessage);
                                         break;
 
+                                    case CancelInvocationMessage cancelInvocationMessage:
+                                        //log
+                                        // check and cancel stream
+                                        if (connection.ActiveRequestCancellationSources.TryRemove(cancelInvocationMessage.InvocationId, out var cts))
+                                        {
+                                            cts.Cancel();
+                                            cts.Dispose();
```
davidfowl: Remove the dispose, it’s not needed

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
davidfowl: Make this message part of the other logging messages

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -450,30 +481,34 @@ private bool IsChannel(Type type, out Type payloadType)
             {
                 await SendMessageAsync(connection, CompletionMessage.WithError(invocationId, ex.Message));
             }
+            finally
+            {
+                if (connection.ActiveRequestCancellationSources.TryRemove(invocationId, out var cts))
```
davidfowl: Ah wait we do this!

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -482,6 +517,13 @@ private bool IsStreamed(HubConnectionContext connection, ObjectMethodExecutor me
                 enumerator = null;
                 return false;
             }
+
+            CancellationToken CreateCancellation()
+            {
```
davidfowl: Why is this a local function?

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -275,6 +275,21 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                                         _ = ProcessInvocation(connection, invocationMessage);
                                         break;
 
+                                    case CancelInvocationMessage cancelInvocationMessage:
+                                        //log
+                                        // check and cancel stream
```
davidfowl: nit: fix the casing of these comments

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -438,8 +453,24 @@ private bool IsChannel(Type type, out Type payloadType)
...
             {
-                while (await enumerator.MoveNextAsync())
+                while (true)
                 {
+                    bool hasItem;
+                    try
+                    {
+                        hasItem = await enumerator.MoveNextAsync();
+                    }
+                    catch (Exception)
```
davidfowl: When does this happen?

BrennanConroy: Good point, I don't think it can.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
BrennanConroy: ?

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
```diff
@@ -27,11 +27,12 @@ internal static class SendUtils
                 {
                     // Grab as many messages as we can from the channel
                     messages = new List<SendMessage>();
-                    while (!transportCts.Token.IsCancellationRequested && application.In.TryRead(out SendMessage message))
+                    while (!transportCts.IsCancellationRequested && application.In.TryRead(out SendMessage message))
                     {
                         messages.Add(message);
                     }
 
+                    transportCts.Token.ThrowIfCancellationRequested();
```
BrennanConroy: This is to fix an infinite loop we had

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
BrennanConroy: Didn't like having the exact same code in both of the above if statements

BrennanConroy:
🆙 📅

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
davidfowl: What infinite loop? Also accessing the token off the CTS every time isn't great.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -138,8 +138,35 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
 
         private async Task<ReadableChannel<object>> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken)
         {
-            var irq = InvocationRequest.Stream(cancellationToken, returnType, GetNextId(), _loggerFactory, out var channel);
-            await InvokeCore(methodName, irq, args, nonBlocking: false);
+            var invokeCts = new CancellationTokenSource();
+            var irq = InvocationRequest.Stream(invokeCts.Token, returnType, GetNextId(), _loggerFactory, out var channel);
```
davidfowl: We should discuss this.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -138,8 +138,35 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
...
+            {
+                await InvokeCore(methodName, irq, args, nonBlocking: false);
+            }
+
+            if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken.Register(state =>
+                {
+                    var connection = (HubConnection)state;
+                    if (!connection._connectionActive.IsCancellationRequested)
```
davidfowl: I was thinking of a way to remove the closure, and we can attach the connection object to the irq.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -175,12 +176,76 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                var httpConnection = new HttpConnection(new Uri(_serverFixture.BaseUrl + path), transportType, loggerFactory);
+                var connection = new HubConnection(httpConnection, protocol, loggerFactory);
+                try
+                {
+                    await connection.StartAsync().OrTimeout();
+
+                    var cts = new CancellationTokenSource();
+                    cts.Cancel();
+
+                    var channel = await connection.StreamAsync<int>("Stream", 5, cts.Token).OrTimeout();
```
davidfowl: Shouldn't this invocation throw?

davidfowl:
Where is the test that verifies the IObservable subscription is removed when a streaming invocation is cancelled.

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
BrennanConroy: If the connection is closed but a message was being sent (still in application channel) then we would loop forever. We could instead let `TryRead` happen which will drain the channel and let `httpClient.SendAsync` throw the exception.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
davidfowl: So we'll get warnings if the steam on the server ends and then the client sends a cancellation?

BrennanConroy: I made it a debug log https://github.com/aspnet/SignalR/pull/979/files#diff-3f9cc189cbcb6b225bb8d6cb6ecdeb21R282

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -138,8 +138,35 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
...
+
+            if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken.Register(state =>
+                {
+                    var invocationReq = (InvocationRequest)state;
+                    if (!invocationReq.HubConnection._connectionActive.IsCancellationRequested)
+                    {
+                        // Fire and forget, if it fails that means we aren't connected anymore.
+                        _ = invocationReq.HubConnection.SendHubMessage(new CancelInvocationMessage(invocationReq.InvocationId), invocationReq);
```
davidfowl: Log failures here.

BrennanConroy: Failures are logged here https://github.com/aspnet/SignalR/pull/979/files#diff-2e56c8934f766e010c7bd3b60cfdb4e3R235 is that enough?

davidfowl: Yep

PR #990, opened 2017-10-05
Fix the VarInt example for the 2GiB

Tragetaschen:
`0x7fffffff` has 31 binary ones. The first four bytes of the VarInt account for 4*7=28, so the fifth byte only needs the remaining three


moozzyk:
Interestingly I just copied the comment from here: https://github.com/aspnet/SignalR/blob/05da66ada08a5a4ae74118a470551d255e21274f/src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs#L32

Tragetaschen:
Ah, I have changed that comment as well

PR #992, opened 2017-10-06
Update TransportProtocols.md

markrendle:
Four ASP.NET Endpoint Transports. There are four.


moozzyk:
I think we need to review (and possibly remove) the `## HTTP Post (Client-to-Server only)` it seems out-of-date (long polling is no longer using Content-Type or the first byte to mark whether the protocol is Text or Binary)  and I am not sure if the description is accurate in other aspects.

moozzyk:
I opened https://github.com/aspnet/SignalR/issues/993 to track this.

Issue #860, opened 2017-09-12
Issue with disposing the HttpConnection


davidfowl:
```
Failed   Microsoft.AspNetCore.SignalR.Client.FunctionalTests.HubConnectionTests.CheckFixedMessage(protocol: JsonHubProtocol { Name = "json", Type = Text }, transportType: ServerSentEvents, path: "/dynamic")
Error Message:
 System.ObjectDisposedException : The CancellationTokenSource has been disposed.
Stack Trace:
   at System.Threading.CancellationTokenSource.ThrowObjectDisposedException()
   at System.Threading.CancellationTokenSource.Cancel()
   at System.Net.Http.HttpClient.Dispose(Boolean disposing)
   at Microsoft.AspNetCore.Sockets.Client.HttpConnection.<DisposeAsyncCore>d__53.MoveNext() in 
```

BrennanConroy:
Looks like this can happen when `HttpClient.Dispose` is called more than once at the same time. In `HttpConnection` we have two spots we dispose the `HttpClient` we should refactor it to only have 1.

moozzyk:
Some notes before we remove one of these. The Dispose in the DisposeAsync is for the cases when the connection did not start successfully (e.g. failure during negotiation) or if the user called `DisposeAsync` between negotiation and starting the transport. In these cases we won't have the continuation on channel completion. The dispose in the channel completion continuation is for cases when the connection was closed abruptly and the user did not call DisposeAsync.

BrennanConroy:
This was fixed when we made the HttpConnection reusable

PR #999, opened 2017-10-09
Dispose HttpClient properly in HttpConnection

BrennanConroy:
#860 
Added tests for both scenarios @moozzyk called out in the linked issue.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -148,6 +148,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             catch
             {
                 Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected);
+                _httpClient.Dispose();
                 throw;
             }
 
@@ -415,7 +416,8 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             if (Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected) == ConnectionState.Initial)
             {
-                // the connection was never started so there is nothing to clean up
+                _httpClient.Dispose();
+                // The connection was never started.
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -365,6 +364,10 @@ private void SetTransferMode(TransferMode transferMode)
                 Output.TryComplete(ex);
                 _logger.ErrorReceiving(_connectionId, ex);
             }
+            finally
+            {
+                _httpClient.Dispose();
```
moozzyk: How is the client disposed when the connection is never started? I think it won't be...

BrennanConroy: Interesting point, I don't think it ever was before?

moozzyk: It did - this is why we had it in the `DisposeAsync`

BrennanConroy: Why would you call `DisposeAsync` when the connection isn't started

moozzyk: I don't think it matters (maybe the app has some additional logic between creating and starting the connection to decide) - we should dispose properly when the user called `DisposeAsync()`...

davidfowl: So now we call Dispose in 3 places!? Can you write a comment saying why?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -412,7 +419,9 @@ private void SetTransferMode(TransferMode transferMode)
 
             if (Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected) == ConnectionState.Initial)
             {
-                // the connection was never started so there is nothing to clean up
+                // The connection was never started.
+                // Connection is now unusable, so cleanup resources.
+                _httpClient.Dispose();
```
moozzyk: I think the client will not be disposed if DisposeAsync is called when the connection is in the connecting state - if this [condition](https://github.com/aspnet/SignalR/pull/999/files#diff-ffe775cce0a6518263da4dfc756b867eR157) is not true.

moozzyk:
I propose to postpone this change for now until we decide whether we want to make the connection instance reusable. If we will make the connection instance reusable (the connection is restartable) we will have to change how we dispose the HttpClient - it probably will be much easier.

BrennanConroy:
Postponing

Issue #813, opened 2017-09-01
Register default functions for observers 'complete', 'error', and 'next' functions


BrennanConroy:
We should register default functions for `complete`, `error`, and `next` in streaming calls for Typescript.

If the user doesn't define a method like `error` then when an error occurs we attempt to call a non-existent function. We should instead have a default function that logs a warning if the user didn't provide a function.

https://github.com/aspnet/SignalR/blob/93b6bcc523811e884403d580507893d67d68e2bc/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Observable.ts#L32-L33

BrennanConroy:
https://github.com/aspnet/SignalR/commit/de535422d733eec0b6dc703764b7324a9ba946bd

PR #1000, opened 2017-10-09
Optional error and complete functions for observables

BrennanConroy:
#813

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
@@ -390,6 +390,31 @@ describe("HubConnection", () => {
             connection.receive({ type: 3, invocationId: connection.lastInvocationId });
             expect(await observer.completed).toEqual([1, 2, 3]);
         });
+
+        it("does not require error function registered", async () => {
+            let connection = new TestConnection();
+
+            let hubConnection = new HubConnection(connection);
+            let observer = hubConnection.stream("testMethod").subscribe({
+                next: val => { }
+            });
+
+            // Typically this would be called by the transport
+            // triggers observer.error()
+            connection.onclose(new Error("Connection lost"));
+        });
+
+        it("does not require complete function registered", async () => {
+            let connection = new TestConnection();
+
+            let hubConnection = new HubConnection(connection);
+            let observer = hubConnection.stream("testMethod").subscribe({
+                next: val => { }
+            });
+
+            // Send completion to trigger observer.complete()
+            connection.receive({ type: 3, invocationId: connection.lastInvocationId });
+        });
     });
 
     describe("onClose", () => {
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
@@ -391,7 +391,7 @@ describe("HubConnection", () => {
             expect(await observer.completed).toEqual([1, 2, 3]);
         });
 
-        it("does not require error function registered", async () => {
+        it("does not require error function registered", (): Promise<any> => {
             let connection = new TestConnection();
 
             let hubConnection = new HubConnection(connection);
@@ -401,10 +401,15 @@ describe("HubConnection", () => {
 
             // Typically this would be called by the transport
             // triggers observer.error()
-            connection.onclose(new Error("Connection lost"));
+            try {
+                connection.onclose(new Error("Connection lost"));
+            } catch (err) {
+                return Promise.reject(err);
+            }
+            return Promise.resolve();
         });
 
-        it("does not require complete function registered", async () => {
+        it("does not require complete function registered", (): Promise<any> => {
             let connection = new TestConnection();
 
             let hubConnection = new HubConnection(connection);
@@ -413,7 +418,12 @@ describe("HubConnection", () => {
             });
 
             // Send completion to trigger observer.complete()
-            connection.receive({ type: 3, invocationId: connection.lastInvocationId });
+            try {
+                connection.receive({ type: 3, invocationId: connection.lastInvocationId });
+            } catch (err) {
+                return Promise.reject(err);
+            }
+            return Promise.resolve();
         });
     });
 
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
@@ -391,25 +391,25 @@ describe("HubConnection", () => {
             expect(await observer.completed).toEqual([1, 2, 3]);
         });
 
-        it("does not require error function registered", (): Promise<any> => {
-            let connection = new TestConnection();
-
-            let hubConnection = new HubConnection(connection);
-            let observer = hubConnection.stream("testMethod").subscribe({
-                next: val => { }
-            });
-
-            // Typically this would be called by the transport
-            // triggers observer.error()
+        it("does not require error function registered", async () => {
             try {
+                let connection = new TestConnection();
+
+                let hubConnection = new HubConnection(connection);
+                let observer = hubConnection.stream("testMethod").subscribe({
+                    next: val => { }
+                });
+
+                // Typically this would be called by the transport
+                // triggers observer.error()
                 connection.onclose(new Error("Connection lost"));
-            } catch (err) {
-                return Promise.reject(err);
+                await Promise.resolve();
+            } catch (e) {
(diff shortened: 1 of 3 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
```diff
@@ -390,6 +390,31 @@ describe("HubConnection", () => {
             connection.receive({ type: 3, invocationId: connection.lastInvocationId });
             expect(await observer.completed).toEqual([1, 2, 3]);
         });
+
+        it("does not require error function registered", async () => {
```
moozzyk: how would these tests fail before the change?

BrennanConroy:
```
1) HubConnection stream does not require error function registered
  Message:
    Failed
  Stack:
    Error: Failed
        at assertion.then.catch (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNetCore.SignalR.Client.TS.Tests\
JasmineUtils.js:28:34)
        at runMicrotasksCallback (internal/process/next_tick.js:58:5)
        at _combinedTickCallback (internal/process/next_tick.js:67:7)
        at process._tickCallback (internal/process/next_tick.js:98:9)
  Message:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
  Stack:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
        at ontimeout (timers.js:365:14)
        at tryOnTimeout (timers.js:237:5)
        at Timer.listOnTimeout (timers.js:207:5)
```

moozzyk: Oh, they are run via the asyncit wrapper... 

can you make the tests more explicit by taking the `done` function and invoking it in the `hubConnection.onclose` handler after making sure that the error is null?

BrennanConroy: I will try!

BrennanConroy: Is there a way to access the normal `it` function when you have `import { asyncit as it, captureException } from './JasmineUtils';`

moozzyk: I think you can just do:
```
it("does not require error function registered", async (done) => {
   ...   
  hubConnection.onclose(e => {
      expect(e).toBe(null);
      done();
  })
```

BrennanConroy: Nope, because normal `it` function is hidden by `asyncit as it`

moozzyk: I [see](https://github.com/aspnet/SignalR/blob/dev/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/JasmineUtils.ts#L8) what you are saying

BrennanConroy: Slightly unrelated
https://github.com/aspnet/SignalR/blob/6d9dd3c1cc5fc85667383bf3ce7e4aeda1158162/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/JasmineUtils.ts#L10
calling `done()` after `fail()` completes failures without waiting for timeout

moozzyk: You would have to probably return a promise that is fulfilled in onclose and rejected on an exception. In this case the test would not be async (interesetingly it is marked as async but it does not have any async calls).

BrennanConroy:
Error would now look like:
```
Failures:
1) HubConnection stream does not require complete function registered
  Message:
    Failed: observer.complete is not a function
  Stack:
    TypeError: observer.complete is not a function
        at Subject.complete (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNetCore.SignalR.Client.TS\Observable
.js:33:26)
        at callbacks.set (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNetCore.SignalR.Client.TS\HubConnection
.js:139:33)
        at HubConnection.processIncomingData (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNetCore.SignalR.Cli
ent.TS\HubConnection.js:74:29)
        at TestConnection.HubConnection.connection.onreceive (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNet
Core.SignalR.Client.TS\HubConnection.js:51:56)
        at TestConnection.receive (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNetCore.SignalR.Client.TS.Test
s\HubConnection.spec.js:416:18)
        at JasmineUtils_1.asyncit.err (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNetCore.SignalR.Client.TS.
Tests\HubConnection.spec.js:349:32)
        at UserContext.testFunction (E:\ASPNETCORE\signalr\client-ts\dist\test\Microsoft.AspNetCore.SignalR.Client.TS.Te
sts\JasmineUtils.js:26:17)

100 specs, 1 failure
Finished in 0.399 seconds
```

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
```diff
@@ -391,7 +391,7 @@ describe("HubConnection", () => {
             expect(await observer.completed).toEqual([1, 2, 3]);
         });
 
-        it("does not require error function registered", async () => {
+        it("does not require error function registered", (): Promise<any> => {
```
moozzyk: Sorry, I was not clear enough. I was thinking about something like this:
```
        it("does not require error function registered", (): Promise<any> => {

	    return new Promise((resolve, reject) => {
               try {
                 let connection = new TestConnection();

                 let hubConnection = new HubConnection(connection);
	         hubConnection.onclose(e => {
                   if (e) {
                     reject(e);
                   }
                   resolve();
                 });
                 let observer = hubConnection.stream("testMethod").subscribe({
                   next: val => { }
                 });

   	         connection.onclose(new Error("Connection lost"));
               }
               catch(e) {
                 reject(e);
               }                
            });
        });
```

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
```diff
@@ -401,10 +401,15 @@ describe("HubConnection", () => {
...
+            try {
+                connection.onclose(new Error("Connection lost"));
+            } catch (err) {
+                return Promise.reject(err);
+            }
+            return Promise.resolve();
         });
 
-        it("does not require complete function registered", async () => {
+        it("does not require complete function registered", (): Promise<any> => {
```
moozzyk: I would leave this test as it was before - because what you have is semantically identical so there is no need to make it more complicated. Just add a comment saying that the expectation is that connection.receive does not throw.
(I focused only on the first test and was not clear that this one was fine - my comment about `async` was just an observation).

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
moozzyk: The idea was to complete the test from onclose handler (as per my code sample). Otherwise we could keep what you originally had since it is basically the same thing as this one...

BrennanConroy: Oh I see. So that doesn't make sense for this test because `e` is always going to have a value since we call `onclose` with an error in order to trigger the `observer.error` call.

moozzyk: If this is the case let's get back to the original version. I thought we could end up with a different error than the one we used to close the underlying connection but it does not seem to be the case.
(Sorry for asking for changes that were not needed).

BrennanConroy: I think its at the original version now?

moozzyk: Yes, it's good to go. Again, sorry for adding more work.

PR #1002, opened 2017-10-10
LoggerDefine for Redis

BrennanConroy:


src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisLoggerExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+            LoggerMessage.Define<string>(LogLevel.Information, new EventId(2, nameof(Subscribing)), "Subscribing to channel: {channel}.");
+
+        private static readonly Action<ILogger, string, Exception> _receivedFromChannel =
+            LoggerMessage.Define<string>(LogLevel.Trace, new EventId(3, nameof(ReceivedFromChannel)), "Received message from redis channel {channel}.");
+
+        private static readonly Action<ILogger, string, Exception> _publishToChannel =
+            LoggerMessage.Define<string>(LogLevel.Trace, new EventId(4, nameof(PublishToChannel)), "Publishing message to redis channel {channel}.");
+
+        private static readonly Action<ILogger, string, Exception> _unsubscribe =
+            LoggerMessage.Define<string>(LogLevel.Information, new EventId(5, nameof(Unsubscribe)), "Unsubscribing from channel: {channel}.");
```
davidfowl: Not sure about this subscribe and unsubscribe being information

BrennanConroy: Debug or Trace?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -279,7 +279,7 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
             {
                 foreach (var subscription in redisSubscriptions)
                 {
-                    _logger.LogInformation("Unsubscribing from channel: {channel}", subscription);
+                    _logger.Unsubscribe(subscription);
```
moozzyk: We are losing the ability to distinguish before unsubscribing from channel and from groupChannel. This can be deduced based on the channel name but I wonder if this is good enough.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisLoggerExtensions.cs
moozzyk: I would go with Trace

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
muratg: Can you share log outputs from before and after the change?

BrennanConroy: Considering the fact that we're changing these to Trace now so they are less visible, and that the channel name will be `hubname.group.groupname` e.g. `ChatHub.group.Admins` so it is pretty easy to tell what the group channels are. I don't think it matters that we lose the "unsubscribe from group" log.

moozzyk: Sounds good.

BrennanConroy:
Ping, approval?

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisLoggerExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+            LoggerMessage.Define<string>(LogLevel.Information, new EventId(0, nameof(ConnectingToEndpoints)), "Connecting to redis endpoints: {endpoints}.");
+
+        private static readonly Action<ILogger, Exception> _connected =
+            LoggerMessage.Define(LogLevel.Information, new EventId(1, nameof(Connected)), "Connected to redis.");
+
+        private static readonly Action<ILogger, string, Exception> _subscribing =
+            LoggerMessage.Define<string>(LogLevel.Trace, new EventId(2, nameof(Subscribing)), "Subscribing to channel: {channel}.");
+
+        private static readonly Action<ILogger, string, Exception> _receivedFromChannel =
+            LoggerMessage.Define<string>(LogLevel.Trace, new EventId(3, nameof(ReceivedFromChannel)), "Received message from redis channel {channel}.");
```
moozzyk: nit: should Redis be capitalized?

PR #1003, opened 2017-10-10
Add negotiate timeout option to hub options

BrennanConroy:
Part of #254

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -38,7 +38,7 @@ namespace Microsoft.AspNetCore.SignalR
         private readonly ILogger<HubEndPoint<THub>> _logger;
         private readonly IServiceScopeFactory _serviceScopeFactory;
         private readonly IHubProtocolResolver _protocolResolver;
-        private readonly IOptions<HubOptions> _hubOptions;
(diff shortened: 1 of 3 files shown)


moozzyk:
What happens if the timeout is greater than 5 seconds and the negotiate is received after 5 seconds (e.g. the timeout is 10 seconds and negotiate is received 8 seconds after starting the connection)? I am asking because the connection manager [removes inactive connections after 5 seconds](https://github.com/aspnet/SignalR/blob/b8d2d24b6777766547149b169d46583b6db6d592/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs#L136) so it may clean up the connection while it is waiting for negotiate.

BrennanConroy:
By the time negotiate in the HubEndpoint is called the connection has already been marked as active so the check will ignore it.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -131,7 +130,7 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
             {
                 using (var cts = new CancellationTokenSource())
                 {
-                    cts.CancelAfter(NegotiateTimeout);
+                    cts.CancelAfter(_hubOptions.Value.NegotiateTimeout);
```
davidfowl: Store _hubOptions as HubOptions not IOptions\<HubOptions\>

Issue #900, opened 2017-09-16
New client side API design


davidfowl:
I'm proposing a new client side API design to solve a bunch of issues:
- https://github.com/aspnet/SignalR/issues/858
- https://github.com/aspnet/SignalR/issues/842
- https://github.com/aspnet/SignalR/issues/836
- https://github.com/aspnet/SignalR/issues/212
- The fact that subscriptions cannot be removed once added
- Passing state to callback to avoid closers

The big changes are:
- Events registration method calls instead of .NET events. This allows taking a state parameter to avoid closure allocations.
- Event registration methods also return IDisposable to allow removing the underlying subscription.
- The other big benefit of methods for event registration are the ability to add overloads. We can support non async overloads of event handlers which means callers aren't forced to return Task.CompletedTask.
- The Connected event has been removed since StartAsync represents the same thing. If we need to support a model where outside actors need to subscribe to the event we can consider making Start return the same task instead of throwing.
- The closed event is removed in favor of a cancellation token. It means you can poll for closure as well as use a callback. It also allows passing the connection lifetime to other APIs. It also means you can get notified that the connection has closed after it has closed. One downside of this is that you can't run async logic that we wait on.

```C#
public interface IConnection
{
    Task StartAsync(CancellationToken cancellationToken = default);
    Task DisposeAsync(CancellationToken cancellationToken = default);
    Task SendAsync(byte[] data, CancellationToken cancellationToken = default);

    public CancellationToken ClosedToken { get; }

    public IDisposable OnReceived(Func<byte[], object, Task> callback, object state);

    IFeatureCollection Features { get; }
}

public class HubConnection
{
    Task StartAsync(CancellationToken cancellationToken = default);
    Task DisposeAsync(CancellationToken cancellationToken = default);
    Task InvokeAsync(string method, CancellationToken cancellationToken = default, params object[] args);
    Task SendAsync(string method, CancellationToken cancellationToken = default, params object[] args);
    Task<ReadableChannel<object>> StreamAsync(string method, CancellationToken cancellationToken = default, params object[] args);

    public CancellationToken ClosedToken { get; }

    IDisposable On(string methodName, Type[] parameterTypes, Func<object[], object, Task> callback, object state);
}
```

PS: This doesn't take into account the possibility that we might want to change `IConnection` to be pipe based in the near future (this means Send and Receive would go away).

/cc @moozzyk

KPixel:
Looking good.

[...]

PR #1006, opened 2017-10-11
OnReceived

mikaelm12:
Removed the Receive event in favor of an OnReceived method as detailed in https://github.com/aspnet/SignalR/issues/900

Roast Me

### DIFF
--- a/samples/ClientSample/RawSample.cs
+++ b/samples/ClientSample/RawSample.cs
@@ -40,7 +40,7 @@ namespace ClientSample
             try
             {
                 var cts = new CancellationTokenSource();
-                connection.Received += data => Console.Out.WriteLineAsync($"{Encoding.UTF8.GetString(data)}");
+                connection.OnReceived(data => Console.Out.WriteLineAsync($"{Encoding.UTF8.GetString(data)}"));
                 connection.Closed += e =>
                 {
                     cts.Cancel();
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -325,7 +325,14 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
             foreach (var handler in copiedHandlers)
             {
-                await handler.InvokeAsync(invocation.Arguments);
+                try
+                {
+                    await handler.InvokeAsync(invocation.Arguments);
+                }
+                catch (Exception ex)
+                {
+                    _logger.ExceptionThrownFromCallback(nameof(On), ex);
+                }
             }
         }
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private readonly object _pendingCallsLock = new object();
         private readonly CancellationTokenSource _connectionActive = new CancellationTokenSource();
         private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>();
-        private readonly ConcurrentDictionary<string, InvocationHandler> _handlers = new ConcurrentDictionary<string, InvocationHandler>();
+        private readonly ConcurrentDictionary<string, List<InvocationHandler>> _handlers = new ConcurrentDictionary<string, List<InvocationHandler>>();
 
         private int _nextId = 0;
 
@@ -62,7 +62,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _protocol = protocol;
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
-            _connection.Received += OnDataReceivedAsync;
+            _connection.OnReceived((data, state) => ((HubConnection)state).OnDataReceivedAsync(data), this);
             _connection.Closed += Shutdown;
         }
 
@@ -119,10 +119,17 @@ namespace Microsoft.AspNetCore.SignalR.Client
         }
 
         // TODO: Client return values/tasks?
-        public void On(string methodName, Type[] parameterTypes, Func<object[], Task> handler)
+        public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], object, Task> handler, object state)
         {
-            var invocationHandler = new InvocationHandler(parameterTypes, handler);
-            _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
+            var invocationHandler = new InvocationHandler(parameterTypes, handler, state);
+            var invocationList = _handlers.AddOrUpdate(methodName,  _ =>  new List<InvocationHandler> { invocationHandler }, 
+                (_, invocations) =>
+                {
+                    invocations.Add(invocationHandler);
+                    return invocations;
+                });
+
+            return new Subscription(invocationHandler, invocationList);
         }
 
         public async Task<ReadableChannel<object>> StreamAsync(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default(CancellationToken))
@@ -299,18 +306,26 @@ namespace Microsoft.AspNetCore.SignalR.Client
             return Task.CompletedTask;
         }
 
-        private Task DispatchInvocationAsync(InvocationMessage invocation, CancellationToken cancellationToken)
+        private async Task DispatchInvocationAsync(InvocationMessage invocation, CancellationToken cancellationToken)
         {
             // Find the handler
-            if (!_handlers.TryGetValue(invocation.Target, out InvocationHandler handler))
+            if (!_handlers.TryGetValue(invocation.Target, out var handlers))
             {
                 _logger.MissingHandler(invocation.Target);
-                return Task.CompletedTask;
+                return;
             }
 
-            // TODO: Return values
-            // TODO: Dispatch to a sync context to ensure we aren't blocking this loop.
-            return handler.Handler(invocation.Arguments);
+            // Copying the callbacks to avoid concurrency issues
+            InvocationHandler[] copiedHandlers;
+            lock (handlers)
+            {
+                copiedHandlers = new InvocationHandler[handlers.Count];
+                handlers.CopyTo(copiedHandlers);
+            }
+            foreach (var handler in copiedHandlers)
+            {
+                await handler.InvokeAsync(invocation.Arguments);
+            }
         }
 
         // This async void is GROSS but we need to dispatch asynchronously because we're writing to a Channel
@@ -404,6 +419,26 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
+        private class Subscription : IDisposable
+        {
+            private readonly InvocationHandler _handler;
+            private readonly List<InvocationHandler> _handlerList;
+
+            public Subscription(InvocationHandler handler, List<InvocationHandler> handlerList)
+            {
+                _handler = handler;
+                _handlerList = handlerList;
+            }
+
+            public void Dispose()
+            {
+                lock (_handlerList)
+                {
+                    _handlerList.Remove(_handler);
+                }
+            }
+        }
+
         private class HubBinder : IInvocationBinder
         {
             private HubConnection _connection;
@@ -415,7 +450,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             public Type GetReturnType(string invocationId)
             {
-                if (!_connection._pendingCalls.TryGetValue(invocationId, out InvocationRequest irq))
+                if (!_connection._pendingCalls.TryGetValue(invocationId, out var irq))
                 {
                     _connection._logger.ReceivedUnexpectedResponse(invocationId);
                     return null;
@@ -425,24 +460,40 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             public Type[] GetParameterTypes(string methodName)
             {
-                if (!_connection._handlers.TryGetValue(methodName, out InvocationHandler handler))
+                if (!_connection._handlers.TryGetValue(methodName, out var handlers))
                 {
                     _connection._logger.MissingHandler(methodName);
                     return Type.EmptyTypes;
                 }
-                return handler.ParameterTypes;
+
+                // We use the parameter types of the first handler
+                lock (handlers)
+                {
+                    if (handlers.Count > 0)
+                    {
+                        return handlers[0].ParameterTypes;
+                    }
+                    throw new InvalidOperationException($"There are no callbacks registered for the method '{methodName}'");
+                }
             }
         }
 
         private struct InvocationHandler
         {
-            public Func<object[], Task> Handler { get; }
+            private Func<object[], object, Task> _callback { get; }
             public Type[] ParameterTypes { get; }
+            private readonly object _state;
 
-            public InvocationHandler(Type[] parameterTypes, Func<object[], Task> handler)
+            public InvocationHandler(Type[] parameterTypes, Func<object[], object, Task> callback, object state)
             {
-                Handler = handler;
+                _callback = callback;
                 ParameterTypes = parameterTypes;
+                _state = state;
+            }
+
+            public async Task InvokeAsync(object[] parameters)
+            {
+                await _callback(parameters, _state);
             }
         }
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -481,7 +481,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private struct InvocationHandler
         {
-            private Func<object[], object, Task> _callback { get; }
+            private Func<object[], object, Task> _callback;
             public Type[] ParameterTypes { get; }
             private readonly object _state;
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -125,7 +125,10 @@ namespace Microsoft.AspNetCore.SignalR.Client
             var invocationList = _handlers.AddOrUpdate(methodName,  _ =>  new List<InvocationHandler> { invocationHandler }, 
                 (_, invocations) =>
                 {
-                    invocations.Add(invocationHandler);
+                    lock (invocations)
+                    {
+                        invocations.Add(invocationHandler);
+                    }
                     return invocations;
                 });
 
@@ -323,6 +326,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 copiedHandlers = new InvocationHandler[handlers.Count];
                 handlers.CopyTo(copiedHandlers);
             }
+
             foreach (var handler in copiedHandlers)
             {
                 try
@@ -488,8 +492,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private struct InvocationHandler
         {
-            private Func<object[], object, Task> _callback;
             public Type[] ParameterTypes { get; }
+            private readonly Func<object[], object, Task> _callback;
             private readonly object _state;
 
             public InvocationHandler(Type[] parameterTypes, Func<object[], object, Task> callback, object state)
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -315,6 +315,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 return;
             }
 
+            //TODO: Optimize this!
             // Copying the callbacks to avoid concurrency issues
             InvocationHandler[] copiedHandlers;
             lock (handlers)
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -492,9 +492,9 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 _state = state;
             }
 
-            public async Task InvokeAsync(object[] parameters)
+            public Task InvokeAsync(object[] parameters)
             {
-                await _callback(parameters, _state);
+                return _callback(parameters, _state);
             }
         }
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
@@ -10,69 +10,70 @@ namespace Microsoft.AspNetCore.SignalR.Client
 {
     public static partial class HubConnectionExtensions
     {
-        private static void On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Action<object[]> handler)
+        private static IDisposable On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Action<object[]> handler)
         {
-            hubConnetion.On(methodName, parameterTypes, (parameters) =>
+            return hubConnetion.On(methodName, parameterTypes, (parameters, state) =>
             {
-                handler(parameters);
+                var currentHandler = (Action<object[]>)state;
+                currentHandler(parameters);
                 return Task.CompletedTask;
-            });
+            }, handler);
         }
 
-        public static void On(this HubConnection hubConnection, string methodName, Action handler)
+        public static IDisposable On(this HubConnection hubConnection, string methodName, Action handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName, Type.EmptyTypes, args => handler());
+            return hubConnection.On(methodName, Type.EmptyTypes, args => handler());
         }
 
-        public static void On<T1>(this HubConnection hubConnection, string methodName, Action<T1> handler)
+        public static IDisposable On<T1>(this HubConnection hubConnection, string methodName, Action<T1> handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName,
+            return hubConnection.On(methodName,
                 new[] { typeof(T1) },
                 args => handler((T1)args[0]));
         }
 
-        public static void On<T1, T2>(this HubConnection hubConnection, string methodName, Action<T1, T2> handler)
+        public static IDisposable On<T1, T2>(this HubConnection hubConnection, string methodName, Action<T1, T2> handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName,
+            return hubConnection.On(methodName,
                 new[] { typeof(T1), typeof(T2) },
                 args => handler((T1)args[0], (T2)args[1]));
         }
 
-        public static void On<T1, T2, T3>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3> handler)
+        public static IDisposable On<T1, T2, T3>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3> handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName,
+            return hubConnection.On(methodName,
                 new[] { typeof(T1), typeof(T2), typeof(T3) },
                 args => handler((T1)args[0], (T2)args[1], (T3)args[2]));
         }
 
-        public static void On<T1, T2, T3, T4>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4> handler)
+        public static IDisposable On<T1, T2, T3, T4>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4> handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName,
+            return hubConnection.On(methodName,
                 new[] { typeof(T1), typeof(T2), typeof(T3), typeof(T4) },
                 args => handler((T1)args[0], (T2)args[1], (T3)args[2], (T4)args[3]));
         }
@@ -89,41 +90,50 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 args => handler((T1)args[0], (T2)args[1], (T3)args[2], (T4)args[3], (T5)args[4]));
         }
 
-        public static void On<T1, T2, T3, T4, T5, T6>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4, T5, T6> handler)
+        public static IDisposable On<T1, T2, T3, T4, T5, T6>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4, T5, T6> handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName,
+            return hubConnection.On(methodName,
                 new[] { typeof(T1), typeof(T2), typeof(T3), typeof(T4), typeof(T5), typeof(T6) },
                 args => handler((T1)args[0], (T2)args[1], (T3)args[2], (T4)args[3], (T5)args[4], (T6)args[5]));
         }
 
-        public static void On<T1, T2, T3, T4, T5, T6, T7>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4, T5, T6, T7> handler)
+        public static IDisposable On<T1, T2, T3, T4, T5, T6, T7>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4, T5, T6, T7> handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName,
+            return hubConnection.On(methodName,
                 new[] { typeof(T1), typeof(T2), typeof(T3), typeof(T4), typeof(T5), typeof(T6), typeof(T7) },
                 args => handler((T1)args[0], (T2)args[1], (T3)args[2], (T4)args[3], (T5)args[4], (T6)args[5], (T7)args[6]));
         }
 
 
-        public static void On<T1, T2, T3, T4, T5, T6, T7, T8>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4, T5, T6, T7, T8> handler)
+        public static IDisposable On<T1, T2, T3, T4, T5, T6, T7, T8>(this HubConnection hubConnection, string methodName, Action<T1, T2, T3, T4, T5, T6, T7, T8> handler)
         {
             if (hubConnection == null)
             {
                 throw new ArgumentNullException(nameof(hubConnection));
             }
 
-            hubConnection.On(methodName,
+            return hubConnection.On(methodName,
                 new[] { typeof(T1), typeof(T2), typeof(T3), typeof(T4), typeof(T5), typeof(T6), typeof(T7), typeof(T8) },
                 args => handler((T1)args[0], (T2)args[1], (T3)args[2], (T4)args[3], (T5)args[4], (T6)args[5], (T7)args[6], (T8)args[7]));
+        }
+
+        public static IDisposable On(this HubConnection hubConnection, string methodName, Type[] parameterTypes, Func<object[], Task> handler)
+        {
+            return hubConnection.On(methodName, parameterTypes, (parameters, state) =>
+            {
+                var currentHandler = (Func<object[], Task>)state;
+                return currentHandler(parameters);
+            }, handler);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
@@ -104,8 +104,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.Internal
             LoggerMessage.Define<string>(LogLevel.Error, new EventId(4, nameof(StreamItemOnNonStreamInvocation)), "Invocation {invocationId} received stream item but was invoked as a non-streamed invocation.");
 
         private static readonly Action<ILogger, string, Exception> _exceptionThrownFromCallback =
-           LoggerMessage.Define<string>(LogLevel.Error, new EventId(23, nameof(ExceptionThrownFromCallback)), "An exception was thrown from the '{callback}' callback");
-
+           LoggerMessage.Define<string>(LogLevel.Error, new EventId(5, nameof(ExceptionThrownFromCallback)), "An exception was thrown from the '{callback}' callback");
 
         public static void PreparingNonBlockingInvocation(this ILogger logger, string invocationId, string target, int count)
         {
@@ -267,7 +266,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.Internal
 
         public static void ExceptionThrownFromCallback(this ILogger logger, string callbackName, Exception exception)
         {
-                _exceptionThrownFromCallback(logger, callbackName, exception);
+             _exceptionThrownFromCallback(logger, callbackName, exception);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
@@ -103,6 +103,10 @@ namespace Microsoft.AspNetCore.SignalR.Client.Internal
         private static readonly Action<ILogger, string, Exception> _streamItemOnNonStreamInvocation =
             LoggerMessage.Define<string>(LogLevel.Error, new EventId(4, nameof(StreamItemOnNonStreamInvocation)), "Invocation {invocationId} received stream item but was invoked as a non-streamed invocation.");
 
+        private static readonly Action<ILogger, string, Exception> _exceptionThrownFromCallback =
+           LoggerMessage.Define<string>(LogLevel.Error, new EventId(23, nameof(ExceptionThrownFromCallback)), "An exception was thrown from the '{callback}' callback");
+
+
         public static void PreparingNonBlockingInvocation(this ILogger logger, string invocationId, string target, int count)
         {
             _preparingNonBlockingInvocation(logger, invocationId, target, count, null);
@@ -259,6 +263,11 @@ namespace Microsoft.AspNetCore.SignalR.Client.Internal
         public static void StreamItemOnNonStreamInvocation(this ILogger logger, string invocationId)
         {
             _streamItemOnNonStreamInvocation(logger, invocationId, null);
+        }
+
+        public static void ExceptionThrownFromCallback(this ILogger logger, string callbackName, Exception exception)
+        {
+                _exceptionThrownFromCallback(logger, callbackName, exception);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -187,7 +187,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         }
                         catch (Exception ex)
                         {
-                            _logger.ExceptionThrownFromCallback(_connectionId, nameof(OnReceived), ex);
+                            _logger.ExceptionThrownFromCallback(_connectionId, nameof(Closed), ex);
                         }
                     }
                 });
@@ -355,7 +355,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                                 }
                                 catch (Exception ex)
                                 {
-                                    _logger.ExceptionThrownFromCallback(_connectionId, "OnReceived", ex);
+                                    _logger.ExceptionThrownFromCallback(_connectionId, nameof(OnReceived), ex);
                                 }
                             }
                         });
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Net.Http;
@@ -37,15 +39,14 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
         private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
-
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
+        private List<ReceiveCallBack> _callbacks = new List<ReceiveCallBack>();
 
         public Uri Url { get; }
 
         public IFeatureCollection Features { get; } = new FeatureCollection();
 
-        public event Func<byte[], Task> Received;
         public event Func<Exception, Task> Closed;
 
         public HttpConnection(Uri url)
@@ -338,16 +339,23 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         {
                             _logger.RaiseReceiveEvent(_connectionId);
 
-                            var receivedHandler = Received;
-                            if (receivedHandler != null)
+                            // Copying the callbacks to avoid concurrency issues
+                            ReceiveCallBack[] callBackCopies;
+                            lock (_callbacks)
+                            {
+                                callBackCopies = new ReceiveCallBack[_callbacks.Count];
+                                _callbacks.CopyTo(callBackCopies);
+                            }
+
+                            foreach (var callBackObject in callBackCopies)
                             {
                                 try
                                 {
-                                    await receivedHandler(buffer);
+                                    await callBackObject.InvokeAsync(buffer);
                                 }
                                 catch (Exception ex)
                                 {
-                                    _logger.ExceptionThrownFromHandler(_connectionId, nameof(Received), ex);
+                                    _logger.ExceptionThrownFromHandler(_connectionId, nameof(callBackObject), ex);
                                 }
                             }
                         });
@@ -442,6 +450,61 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
 
             _httpClient.Dispose();
+        }
+
+        public IDisposable OnReceived(Func<byte[], object, Task> callback, object state)
+        {
+            var callBack = new ReceiveCallBack(callback, state);
+            lock (_callbacks)
+            {
+                _callbacks.Add(callBack);
+            }
+            return new Subscription(callBack, _callbacks);
+        }
+
+        public IDisposable OnReceived(Func<byte[], Task> callback)
+        {
+            return OnReceived((data, state) => 
+            {
+                var currentCallback = (Func<byte[], Task>)state;
+                return currentCallback(data);
+            }, callback);
+        }
+
+        private class ReceiveCallBack
+        {
+            private readonly Func<byte[], object, Task> _callback;
+            private readonly object _state;
+
+            public ReceiveCallBack(Func<byte[], object, Task> callBack, object state)
+            {
+                _callback = callBack;
+                _state = state;
+            }
+
+            public Task InvokeAsync(byte[] data)
+            {
+                return _callback(data, _state);
+            }
+        }
+
+        private class Subscription : IDisposable
+        {
+            private readonly ReceiveCallBack _callback;
+            private readonly List<ReceiveCallBack> _callbacks;
+            public Subscription(ReceiveCallBack callback, List<ReceiveCallBack> callbacks)
+            {
+                _callback = callback;
+                _callbacks = callbacks;
+            }
+
+            public void Dispose()
+            {
+                lock (_callbacks)
+                {
+                    _callbacks.Remove(_callback);
+                }
+            }
         }
 
         private class ConnectionState
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -41,7 +41,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
-        private List<ReceiveCallback> _callbacks = new List<ReceiveCallback>();
+        private readonly List<ReceiveCallback> _callbacks = new List<ReceiveCallback>();
 
         public Uri Url { get; }
 
@@ -454,12 +454,12 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public IDisposable OnReceived(Func<byte[], object, Task> callback, object state)
         {
-            var callBack = new ReceiveCallback(callback, state);
+            var receiveCallback = new ReceiveCallback(callback, state);
             lock (_callbacks)
             {
-                _callbacks.Add(callBack);
+                _callbacks.Add(receiveCallback);
             }
-            return new Subscription(callBack, _callbacks);
+            return new Subscription(receiveCallback, _callbacks);
         }
 
         private class ReceiveCallback
@@ -467,9 +467,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
             private readonly Func<byte[], object, Task> _callback;
             private readonly object _state;
 
-            public ReceiveCallback(Func<byte[], object, Task> callBack, object state)
+            public ReceiveCallback(Func<byte[], object, Task> callback, object state)
             {
-                _callback = callBack;
+                _callback = callback;
                 _state = state;
             }
 
@@ -481,11 +481,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         private class Subscription : IDisposable
         {
-            private readonly ReceiveCallback _callback;
+            private readonly ReceiveCallback _receiveCallback;
             private readonly List<ReceiveCallback> _callbacks;
             public Subscription(ReceiveCallback callback, List<ReceiveCallback> callbacks)
             {
-                _callback = callback;
+                _receiveCallback = callback;
                 _callbacks = callbacks;
             }
 
@@ -493,7 +493,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             {
                 lock (_callbacks)
                 {
-                    _callbacks.Remove(_callback);
+                    _callbacks.Remove(_receiveCallback);
                 }
             }
         }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -462,15 +462,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
             return new Subscription(callBack, _callbacks);
         }
 
-        public IDisposable OnReceived(Func<byte[], Task> callback)
-        {
-            return OnReceived((data, state) => 
-            {
-                var currentCallback = (Func<byte[], Task>)state;
-                return currentCallback(data);
-            }, callback);
-        }
-
         private class ReceiveCallBack
         {
             private readonly Func<byte[], object, Task> _callback;
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -41,7 +41,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
-        private List<ReceiveCallBack> _callbacks = new List<ReceiveCallBack>();
+        private List<ReceiveCallback> _callbacks = new List<ReceiveCallback>();
 
         public Uri Url { get; }
 
@@ -187,7 +187,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         }
                         catch (Exception ex)
                         {
-                            _logger.ExceptionThrownFromHandler(_connectionId, nameof(Closed), ex);
+                            _logger.ExceptionThrownFromCallback(_connectionId, nameof(Closed), ex);
                         }
                     }
                 });
@@ -340,22 +340,22 @@ namespace Microsoft.AspNetCore.Sockets.Client
                             _logger.RaiseReceiveEvent(_connectionId);
 
                             // Copying the callbacks to avoid concurrency issues
-                            ReceiveCallBack[] callBackCopies;
+                            ReceiveCallback[] callbackCopies;
                             lock (_callbacks)
                             {
-                                callBackCopies = new ReceiveCallBack[_callbacks.Count];
-                                _callbacks.CopyTo(callBackCopies);
+                                callbackCopies = new ReceiveCallback[_callbacks.Count];
+                                _callbacks.CopyTo(callbackCopies);
                             }
 
-                            foreach (var callBackObject in callBackCopies)
+                            foreach (var callbackObject in callbackCopies)
                             {
                                 try
                                 {
-                                    await callBackObject.InvokeAsync(buffer);
+                                    await callbackObject.InvokeAsync(buffer);
                                 }
                                 catch (Exception ex)
                                 {
-                                    _logger.ExceptionThrownFromHandler(_connectionId, nameof(callBackObject), ex);
+                                    _logger.ExceptionThrownFromCallback(_connectionId, "OnReceived", ex);
                                 }
                             }
                         });
@@ -454,7 +454,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public IDisposable OnReceived(Func<byte[], object, Task> callback, object state)
         {
-            var callBack = new ReceiveCallBack(callback, state);
+            var callBack = new ReceiveCallback(callback, state);
             lock (_callbacks)
             {
                 _callbacks.Add(callBack);
@@ -462,12 +462,12 @@ namespace Microsoft.AspNetCore.Sockets.Client
             return new Subscription(callBack, _callbacks);
         }
 
-        private class ReceiveCallBack
+        private class ReceiveCallback
         {
             private readonly Func<byte[], object, Task> _callback;
             private readonly object _state;
 
-            public ReceiveCallBack(Func<byte[], object, Task> callBack, object state)
+            public ReceiveCallback(Func<byte[], object, Task> callBack, object state)
             {
                 _callback = callBack;
                 _state = state;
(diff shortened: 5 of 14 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -62,7 +63,7 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             _protocol = protocol;
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
-            _connection.Received += OnDataReceivedAsync;
+            _connection.OnReceived((data, state) => OnDataReceivedAsync(data), null);
```
BrennanConroy: Is `this` captured in the closure? If so you could pass `this` as the state and then cast it to `HubConnection` to call `OnDataReceivedAsync`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -275,15 +287,19 @@ private Task Shutdown(Exception ex = null)
...
             // Find the handler
-            if (!_handlers.TryGetValue(invocation.Target, out InvocationHandler handler))
+            if (!_handlers.TryGetValue(invocation.Target, out List<InvocationHandler> handlers))
             {
                 _logger.MissingHandler(invocation.Target);
                 return Task.CompletedTask;
             }
-
+            var tasks = new List<Task>();
+            foreach(var handler in handlers)
```
davidfowl: nit: formatting

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -333,6 +349,7 @@ private void ThrowIfConnectionTerminated(string invocationId)
         }
 
         private string GetNextId() => Interlocked.Increment(ref _nextId).ToString();
+        private string GetNextCallBackId() => Interlocked.Increment(ref _nextCallBackId).ToString();
```
davidfowl: What is this?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -377,6 +394,21 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
             }
         }
 
+        private class Subscription : IDisposable
+        {
+            private InvocationHandler _handler;
```
davidfowl: readonly

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +430,30 @@ public Type GetReturnType(string invocationId)
...
             public Type[] GetParameterTypes(string methodName)
             {
-                if (!_connection._handlers.TryGetValue(methodName, out InvocationHandler handler))
+                if (!_connection._handlers.TryGetValue(methodName, out List<InvocationHandler> handlers))
                 {
                     _connection._logger.MissingHandler(methodName);
                     return Type.EmptyTypes;
                 }
-                return handler.ParameterTypes;
+                if (handlers.Count > 0)
```
davidfowl: Write a comment about the fact that we're using the first handler to determine the parameter types.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
```diff
@@ -10,69 +10,70 @@ namespace Microsoft.AspNetCore.SignalR.Client
 {
     public static partial class HubConnectionExtensions
     {
-        private static void On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Action<object[]> handler)
+        private static IDisposable On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Action<object[]> handler)
         {
-            hubConnetion.On(methodName, parameterTypes, (parameters) =>
+            //string methodName, Type[] parameterTypes, Func<object[], object, Task> handler, object state)
```
davidfowl: Dead code?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
```diff
@@ -10,69 +10,70 @@ namespace Microsoft.AspNetCore.SignalR.Client
...
+            return hubConnetion.On(methodName, parameterTypes, (parameters, _) =>
             {
                 handler(parameters);
                 return Task.CompletedTask;
-            });
+            }, null);
```
davidfowl: You can avoid the closure now and pass the handler as state.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +440,51 @@ private void SetTransferMode(TransferMode transferMode)
             _httpClient.Dispose();
         }
 
+        private int GetNextCallBackId() => Interlocked.Increment(ref _callBackId);
```
davidfowl: Why?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -40,12 +42,13 @@ public class HttpConnection : IConnection
 
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
+        private List<ReceiveCallBack> _callBacks = new List<ReceiveCallBack>();
```
davidfowl: readonly

davidfowl: nit: style these should be with the other fields.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -275,15 +287,19 @@ private Task Shutdown(Exception ex = null)
...
-
+            var tasks = new List<Task>();
+            foreach(var handler in handlers)
+            {
+                tasks.Add(handler.Handler(invocation.Arguments, null));
+            }
             // TODO: Return values
             // TODO: Dispatch to a sync context to ensure we aren't blocking this loop.
-            return handler.Handler(invocation.Arguments);
+            return Task.WhenAll(tasks);
```
davidfowl: Lets avoid calling these in parallel. Call each one sequentially.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +440,51 @@ private void SetTransferMode(TransferMode transferMode)
...
+        {
+            var id = GetNextCallBackId();
+            var callBack = new ReceiveCallBack(callback, state, id);
+            _callBacks.Add(callBack);
+            return new Subscription(id, _callBacks);
+        }
+
+        private class ReceiveCallBack
+        {
+            public int CallBackID { get; }
```
davidfowl: This isn't needed.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +430,30 @@ public Type GetReturnType(string invocationId)
...
                 {
                     _connection._logger.MissingHandler(methodName);
                     return Type.EmptyTypes;
                 }
-                return handler.ParameterTypes;
+                if (handlers.Count > 0)
+                {
+                    return handlers[0].ParameterTypes;
+                }
+                throw new FormatException($"There are no callback registered for the method '{methodName}'");
```
BrennanConroy: "callbacks"

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -275,15 +287,19 @@ private Task Shutdown(Exception ex = null)
...
+            if (!_handlers.TryGetValue(invocation.Target, out List<InvocationHandler> handlers))
             {
                 _logger.MissingHandler(invocation.Target);
                 return Task.CompletedTask;
             }
-
+            var tasks = new List<Task>();
+            foreach(var handler in handlers)
+            {
+                tasks.Add(handler.Handler(invocation.Arguments, null));
```
BrennanConroy: null seems wrong, it should be `handler.State`?

BrennanConroy: Which also means you should add some tests using state

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -338,17 +341,10 @@ private void SetTransferMode(TransferMode transferMode)
                         {
                             _logger.RaiseReceiveEvent(_connectionId);
 
-                            var receivedHandler = Received;
-                            if (receivedHandler != null)
+                            //First item in the tuple is the callback, second item is the state object.
```
BrennanConroy: Remove

samples/ClientSample/RawSample.cs
```diff
@@ -40,7 +40,7 @@ internal static void Register(CommandLineApplication app)
             try
             {
                 var cts = new CancellationTokenSource();
-                connection.Received += data => Console.Out.WriteLineAsync($"{Encoding.UTF8.GetString(data)}");
+                connection.OnReceived((data, state) => Console.Out.WriteLineAsync($"{Encoding.UTF8.GetString(data)}"), null);
```
davidfowl: Add an extension method that doesn't require state

BrennanConroy:
Agreed

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
moozzyk: Isn't it super racey? The list could be modified if someone happens to be adding or removing a handler while a message is received.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +430,30 @@ public Type GetReturnType(string invocationId)
 
             public Type[] GetParameterTypes(string methodName)
```
moozzyk: probably needs a lock - if the last subscription for the method happens to be removed on a different thread accessing the handler will fail.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +430,30 @@ public Type GetReturnType(string invocationId)
...
 
         private struct InvocationHandler
         {
-            public Func<object[], Task> Handler { get; }
+            public Func<object[], object, Task> Handler { get; }
             public Type[] ParameterTypes { get; }
+            public object State { get; }
 
-            public InvocationHandler(Type[] parameterTypes, Func<object[], Task> handler)
+            public InvocationHandler(Type[] parameterTypes, Func<object[], object, Task> handler, object state)
```
moozzyk: nit: I would rename Handler/handler to Callback/callback

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: What if we just had just one callback (a Func) on HttpConnection? Just wondering when/if multiple callbacks are useful at this level...

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -338,17 +341,10 @@ private void SetTransferMode(TransferMode transferMode)
...
+                            foreach (var callBackObject in _callBacks)
```
moozzyk: needs a lock

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +440,51 @@ private void SetTransferMode(TransferMode transferMode)
...
+            var callBack = new ReceiveCallBack(callback, state, id);
+            _callBacks.Add(callBack);
+            return new Subscription(id, _callBacks);
+        }
+
+        private class ReceiveCallBack
+        {
+            public int CallBackID { get; }
+            private Func<byte[], object, Task> CallBack { get; set; }
+            private object State { get; set; }
```
moozzyk: These are private - why they need to be properties instead of readonly fields?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +440,51 @@ private void SetTransferMode(TransferMode transferMode)
...
+            private int _callBackId;
+            private List<ReceiveCallBack> _handlers;
+            public Subscription(int callBackId, List<ReceiveCallBack> handler)
+            {
+                _callBackId = callBackId;
+                _handlers = handler;
+            }
+            public void Dispose()
+            {
+                _handlers.RemoveAll((callBack) => { return callBack.CallBackID == _callBackId; });
```
moozzyk: use just reference to remove  (and also needs a lock)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +440,51 @@ private void SetTransferMode(TransferMode transferMode)
...
+
+        private class Subscription : IDisposable
+        {
+            private int _callBackId;
+            private List<ReceiveCallBack> _handlers;
+            public Subscription(int callBackId, List<ReceiveCallBack> handler)
+            {
+                _callBackId = callBackId;
+                _handlers = handler;
+            }
```
moozzyk: nit: separate with empty line

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -289,16 +289,16 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
 
             var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
 
-            var receivedInvoked = false;
-            connection.Received += m =>
+            var onReceivedInvoked = false;
+            connection.OnReceived((_, __) =>
```
moozzyk: I thought you could use `(_, _)`..

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -338,17 +341,10 @@ private void SetTransferMode(TransferMode transferMode)
...
                             {
-                                try
-                                {
-                                    await receivedHandler(buffer);
-                                }
-                                catch (Exception ex)
-                                {
-                                    _logger.ExceptionThrownFromHandler(_connectionId, nameof(Received), ex);
-                                }
+                                await callBackObject.Invoke(buffer);
```
moozzyk: What happens if it throws?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -672,132 +669,6 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
         }
 
         [Fact]
-        public async Task CanReceiveDataEvenIfExceptionThrownFromPreviousReceivedEvent()
```
moozzyk: I think these tests are still relevant - user code can throw and I don't see exceptions to be handled.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -672,132 +669,6 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-                Assert.Equal("42", await receiveTcs.Task.OrTimeout());
-            }
-            finally
-            {
-                await connection.DisposeAsync();
-            }
-        }
-
-        [Fact]
-        public async Task CanReceiveDataEvenIfExceptionThrownSynchronouslyFromPreviousReceivedEvent()
```
moozzyk: ditto

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -120,23 +120,6 @@ public class HubConnectionTests
             Assert.Same(exception, thrown);
         }
 
-        [Fact]
-        public async Task DoesNotThrowWhenClientMethodCalledButNoInvocationHandlerHasBeenSetUp()
```
moozzyk: This test should not have been removed.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
moozzyk: What happens if a user throws?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
mikaelm12: Nope. They have to be different

samples/ClientSample/RawSample.cs
```diff
@@ -40,7 +40,7 @@ internal static void Register(CommandLineApplication app)
             try
             {
                 var cts = new CancellationTokenSource();
-                connection.Received += data => Console.Out.WriteLineAsync($"{Encoding.UTF8.GetString(data)}");
+                connection.OnReceived((data, _) => Console.Out.WriteLineAsync($"{Encoding.UTF8.GetString(data)}"));
```
davidfowl: This should be:

```
connection.OnReceived(data => Console.Out.WriteLineAsync($"{Encoding.UTF8.GetString(data)}"))
```

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -272,18 +284,27 @@ private Task Shutdown(Exception ex = null)
...
-            // TODO: Dispatch to a sync context to ensure we aren't blocking this loop.
-            return handler.Handler(invocation.Arguments);
+            //Copying the callbacks to avoid concurrency issues
+            InvocationHandler[] copiedHandlers;
+            lock (_handlerLock)
+            {
+                copiedHandlers = new InvocationHandler[handlers.Count];
+                handlers.CopyTo(copiedHandlers);
+            }
+            var tasks = new List<Task>();
```
davidfowl: Not needed.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -35,7 +35,8 @@ public class HubConnection
         private readonly object _pendingCallsLock = new object();
         private readonly CancellationTokenSource _connectionActive = new CancellationTokenSource();
         private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>();
-        private readonly ConcurrentDictionary<string, InvocationHandler> _handlers = new ConcurrentDictionary<string, InvocationHandler>();
+        private readonly ConcurrentDictionary<string, List<InvocationHandler>> _handlers = new ConcurrentDictionary<string, List<InvocationHandler>>();
+        private object _handlerLock = new object();
```
davidfowl: What is this for?

mikaelm12: For safely copying and removing callbacks in dispose

davidfowl: The lock should be per list not per connection.

davidfowl: Just use the list as the lock.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +440,32 @@ public Type GetReturnType(string invocationId)
...
                     return Type.EmptyTypes;
                 }
-                return handler.ParameterTypes;
+
+                // We use the parameter types of the first handler
+                if (handlers.Count > 0)
+                {
+                    return handlers[0].ParameterTypes;
+                }
+                throw new FormatException($"There are no callbacks registered for the method '{methodName}'");
```
davidfowl: FormatException?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
```diff
@@ -10,69 +10,79 @@ namespace Microsoft.AspNetCore.SignalR.Client
...
-                handler(parameters);
+                var currentHandler = (Action<object[]>)state;
+                currentHandler(parameters);
                 return Task.CompletedTask;
-            });
+            }, handler);
         }
 
-        public static void On(this HubConnection hubConnection, string methodName, Action handler)
+        public static IDisposable On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Func<object[], object, Task> handler)
```
BrennanConroy: This extension is silly. Change it to also take in `object state`
Also, add one that takes `Func<object[], Task> handler`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
```diff
@@ -10,69 +10,79 @@ namespace Microsoft.AspNetCore.SignalR.Client
...
+        {
+            return hubConnetion.On(methodName, parameterTypes, (parameters, _) =>
+            {
+                handler(parameters, null);
```
BrennanConroy: return the handler call, right now you fire and forget

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -37,15 +39,15 @@ public class HttpConnection : IConnection
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
         private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
-
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
+        private List<ReceiveCallBack> _callBacks = new List<ReceiveCallBack>();
+        private object _callbackLock = new object();
```
BrennanConroy: I think you can just use the list as the lock, same as mentioned above.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -338,17 +340,17 @@ private void SetTransferMode(TransferMode transferMode)
                         {
                             _logger.RaiseReceiveEvent(_connectionId);
 
-                            var receivedHandler = Received;
-                            if (receivedHandler != null)
+                            //Copying the callbacks to avoid concurrency issues
```
BrennanConroy: nit: space

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +446,56 @@ private void SetTransferMode(TransferMode transferMode)
             _httpClient.Dispose();
         }
 
+        public IDisposable OnReceived(Func<byte[], object, Task> callback, object state)
+        {
+            var callBack = new ReceiveCallBack(callback, state);
+            _callBacks.Add(callBack);
```
BrennanConroy: lock needed

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +446,56 @@ private void SetTransferMode(TransferMode transferMode)
...
+            return new Subscription(callBack, _callBacks, _callbackLock);
+        }
+
+        public IDisposable OnReceived(Func<byte[], object, Task> callback)
```
BrennanConroy: This should be an extensions method probably

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +446,56 @@ private void SetTransferMode(TransferMode transferMode)
...
+            public readonly Func<byte[], object, Task> CallBack;
+            private readonly object State;
+
+            public ReceiveCallBack(Func<byte[], object, Task> callBack, object state)
+            {
+                CallBack = callBack;
+                State = state;
+            }
+
+            public Task Invoke(byte[] data)
```
BrennanConroy: nit: `InvokeAsync`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -289,16 +289,16 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-            var receivedInvoked = false;
-            connection.Received += m =>
+            var onReceivedInvoked = false;
+            connection.OnReceived((_, __) =>
             {
-                receivedInvoked = true;
+                onReceivedInvoked = true;
                 return Task.CompletedTask;
-            };
+            }, null);
```
BrennanConroy: super nit: pass `onReceivedInvoked` as the state
And for other tests

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -336,12 +336,11 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
 
             var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
 
-            connection.Received +=
-                async m =>
+            connection.OnReceived(async (_, __) =>
                 {
```
BrennanConroy: format

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -30,14 +32,16 @@ internal class TestConnection : IConnection
...
         public event Func<Task> Connected;
-        public event Func<byte[], Task> Received;
         public event Func<Exception, Task> Closed;
 
         public Task Started => _started.Task;
         public Task Disposed => _disposed.Task;
         public ReadableChannel<byte[]> SentMessages => _sentMessages.In;
         public WritableChannel<byte[]> ReceivedMessages => _receivedMessages.Out;
 
+        private ConcurrentDictionary<int, Tuple<Func<byte[], object, Task>, object>> _callBacks = new ConcurrentDictionary<int, Tuple<Func<byte[], object, Task>, object>>();
```
BrennanConroy: two lines

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -136,5 +143,30 @@ private byte[] FormatMessageToArray(byte[] message)
                 Closed?.Invoke(ex);
             }
         }
+
+        private int GetNextCallBackId() => Interlocked.Increment(ref _callBackId);
+
+        public IDisposable OnReceived(Func<byte[], object, Task> callback, object state)
+        {
+            var id = GetNextCallBackId(); ;
```
BrennanConroy: remove extra `;`

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
BrennanConroy: I think `KeyValuePair` is more appropriate

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -136,5 +143,30 @@ private byte[] FormatMessageToArray(byte[] message)
...
+
+        private class Subscription : IDisposable
+        {
+            private int _callBackId;
+            private ConcurrentDictionary<int, Tuple<Func<byte[], object, Task>, object>> _handler;
+            public Subscription(int callBackId, ConcurrentDictionary<int, Tuple<Func<byte[], object, Task>, object>> handler)
+            {
+                _callBackId = callBackId;
+                _handler = handler;
+            }
```
BrennanConroy: newline

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -62,7 +63,7 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             _protocol = protocol;
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
-            _connection.Received += OnDataReceivedAsync;
+            _connection.OnReceived((data, state) => OnDataReceivedAsync(data), this);
```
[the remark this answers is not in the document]
mikaelm12: Got it 😄

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -119,10 +120,21 @@ private IDataEncoder GetDataEncoder(TransferMode requestedTransferMode, Transfer
         }
 
         // TODO: Client return values/tasks?
-        public void On(string methodName, Type[] parameterTypes, Func<object[], Task> handler)
+        public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], object, Task> handler, object state)
         {
-            var invocationHandler = new InvocationHandler(parameterTypes, handler);
-            _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
+            var invocationHandler = new InvocationHandler(parameterTypes, handler, state);
+            if (!_handlers.TryGetValue(methodName, out var callBackList))
```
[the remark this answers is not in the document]
mikaelm12: I did it this way so that I would conditionally create the `List<InvocationHandler>`.  If I just call add or update I have to create a new list every time.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
mikaelm12: Talked offline. Doesn't apply for this test

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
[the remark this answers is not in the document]
BrennanConroy: `AddOrUpdate` has an overload for a factory for the add part, so you don't need to create a new list every time.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -272,18 +283,26 @@ private Task Shutdown(Exception ex = null)
...
+            // Copying the callbacks to avoid concurrency issues
+            InvocationHandler[] copiedHandlers;
+            lock (handlers)
+            {
+                copiedHandlers = new InvocationHandler[handlers.Count];
+                handlers.CopyTo(copiedHandlers);
+            }
+            foreach (var handler in copiedHandlers)
+            {
+                await handler.InvokeAsync(invocation.Arguments, handler.State);
```
BrennanConroy: Should we aggregate exceptions here? @moozzyk

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +436,37 @@ public Type GetReturnType(string invocationId)
 
             public Type[] GetParameterTypes(string methodName)
             {
-                if (!_connection._handlers.TryGetValue(methodName, out InvocationHandler handler))
+                if (!_connection._handlers.TryGetValue(methodName, out List<InvocationHandler> handlers))
```
BrennanConroy: `out var`

mikaelm12:
Updated. Addressed everything but skipping the `DoesNotThrowWhenClientMethodCalledButNoInvocationHandlerHasBeenSetUp` test in HubConnectionTests atm 😅

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -119,10 +119,20 @@ private IDataEncoder GetDataEncoder(TransferMode requestedTransferMode, Transfer
...
         {
-            var invocationHandler = new InvocationHandler(parameterTypes, handler);
-            _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
+            var invocationHandler = new InvocationHandler(parameterTypes, handler, state);
+            var invocationList = _handlers.AddOrUpdate(methodName, 
+                _ => 
+                {
+                    return new List<InvocationHandler> { invocationHandler };
+                }, 
+                (_, invocations) => {
```
BrennanConroy: `{` on new line

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -377,6 +395,25 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
...
+        private class Subscription : IDisposable
+        {
+            private readonly InvocationHandler _handler;
+            private readonly List<InvocationHandler> _handlerList;
+
+            public Subscription(InvocationHandler handler, List<InvocationHandler> handlerList)
+            {
+                _handler = handler;
+                _handlerList = handlerList;
+            }
```
BrennanConroy: nit: newline

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +435,40 @@ public Type GetReturnType(string invocationId)
...
-            public InvocationHandler(Type[] parameterTypes, Func<object[], Task> handler)
+            public InvocationHandler(Type[] parameterTypes, Func<object[], object, Task> callback, object state)
             {
-                Handler = handler;
+                Callback = callback;
                 ParameterTypes = parameterTypes;
+                State = state;
+            }
+
+            public async Task InvokeAsync(object[] parameters, object state)
```
BrennanConroy: You already have state stored, don't need to pass it in

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
```diff
@@ -10,69 +10,78 @@ namespace Microsoft.AspNetCore.SignalR.Client
...
+            }, handler);
         }
 
-        public static void On(this HubConnection hubConnection, string methodName, Action handler)
+        public static IDisposable On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Func<object[], Task> handler)
+        {
+            return hubConnetion.On(methodName, parameterTypes, (parameters, _) =>
+            {
+                return handler(parameters);
+            }, null);
```
BrennanConroy: pass handler as state

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -338,16 +339,23 @@ private void SetTransferMode(TransferMode transferMode)
...
                             {
                                 try
                                 {
-                                    await receivedHandler(buffer);
+                                    await callBackObject.InvokeAsync(buffer);
                                 }
                                 catch (Exception ex)
                                 {
-                                    _logger.ExceptionThrownFromHandler(_connectionId, nameof(Received), ex);
+                                    _logger.ExceptionThrownFromHandler(_connectionId, "Test", ex);
```
BrennanConroy: "Test"?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,57 @@ private void SetTransferMode(TransferMode transferMode)
...
+            }
+            return new Subscription(callBack, _callBacks);
+        }
+
+        public IDisposable OnReceived(Func<byte[], Task> callback)
+        {
+            return OnReceived((data, _) => { return callback(data); }, null);
+        }
+
+        public class ReceiveCallBack
```
BrennanConroy: private

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,57 @@ private void SetTransferMode(TransferMode transferMode)
...
+                State = state;
+            }
+
+            public Task InvokeAsync(byte[] data)
+            {
+                return CallBack(data, State);
+            }
+        }
+
+        public class Subscription : IDisposable
```
BrennanConroy: private

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: oops

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: So I'm also using these classes in TestConnection. I'll just copy them then

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +437,40 @@ public Type GetReturnType(string invocationId)
...
+                }
             }
         }
 
         private struct InvocationHandler
         {
-            public Func<object[], Task> Handler { get; }
+            public Func<object[], object, Task> Callback { get; }
             public Type[] ParameterTypes { get; }
+            public object State { get; }
```
BrennanConroy: nit: `private readonly`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -338,16 +339,23 @@ private void SetTransferMode(TransferMode transferMode)
...
                             {
                                 try
                                 {
-                                    await receivedHandler(buffer);
+                                    await callBackObject.InvokeAsync(buffer);
                                 }
                                 catch (Exception ex)
                                 {
-                                    _logger.ExceptionThrownFromHandler(_connectionId, nameof(Received), ex);
+                                    _logger.ExceptionThrownFromHandler(_connectionId, nameof(callBackObject), ex);
```
BrennanConroy: Log message was about the 'name' event handler, something needs to change here.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,57 @@ private void SetTransferMode(TransferMode transferMode)
...
+            lock (_callBacks)
+            {
+                _callBacks.Add(callBack);
+            }
+            return new Subscription(callBack, _callBacks);
+        }
+
+        public IDisposable OnReceived(Func<byte[], Task> callback)
+        {
+            return OnReceived((data, _) => { return callback(data); }, null);
```
BrennanConroy: `callback` can be passed as state

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,57 @@ private void SetTransferMode(TransferMode transferMode)
...
+        }
+
+        private class ReceiveCallBack
+        {
+            public readonly Func<byte[], object, Task> CallBack;
```
BrennanConroy: nit: private

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -119,10 +119,21 @@ private IDataEncoder GetDataEncoder(TransferMode requestedTransferMode, Transfer
...
 
         // TODO: Client return values/tasks?
-        public void On(string methodName, Type[] parameterTypes, Func<object[], Task> handler)
+        public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], object, Task> handler, object state)
         {
-            var invocationHandler = new InvocationHandler(parameterTypes, handler);
-            _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
+            var invocationHandler = new InvocationHandler(parameterTypes, handler, state);
+            var invocationList = _handlers.AddOrUpdate(methodName, 
+                _ =>
```
moozzyk: nit: could be just `_ => new List<InvocationHandler>{ invocationHandler }`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +437,40 @@ public Type GetReturnType(string invocationId)
...
+                }
             }
         }
 
         private struct InvocationHandler
         {
-            public Func<object[], Task> Handler { get; }
+            public Func<object[], object, Task> Callback { get; }
             public Type[] ParameterTypes { get; }
+            private readonly object State;
```
moozzyk: nit: should be `_state`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -398,24 +437,40 @@ public Type GetReturnType(string invocationId)
...
+                    }
+                    throw new InvalidOperationException($"There are no callbacks registered for the method '{methodName}'");
+                }
             }
         }
 
         private struct InvocationHandler
         {
-            public Func<object[], Task> Handler { get; }
+            public Func<object[], object, Task> Callback { get; }
```
moozzyk: I think this should not be public.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -37,15 +39,14 @@ public class HttpConnection : IConnection
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
         private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
-
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
+        private List<ReceiveCallBack> _callBacks = new List<ReceiveCallBack>();
```
moozzyk: nit: casing - `_callbacks`  `ReceiveCallback` etc. (including below)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,61 @@ private void SetTransferMode(TransferMode transferMode)
...
+        {
+            var callBack = new ReceiveCallBack(callback, state);
+            lock (_callBacks)
+            {
+                _callBacks.Add(callBack);
+            }
+            return new Subscription(callBack, _callBacks);
+        }
+
+        public IDisposable OnReceived(Func<byte[], Task> callback)
```
moozzyk: I don't think this should be on HttpConnection - this is just syntactic sugar.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,61 @@ private void SetTransferMode(TransferMode transferMode)
...
+            {
+                var currentCallback = (Func<byte[], Task>)state;
+                return currentCallback(data);
+            }, callback);
+        }
+
+        private class ReceiveCallBack
+        {
+            private readonly Func<byte[], object, Task> CallBack;
+            private readonly object State;
```
moozzyk: naming - `_state`, `_callback` - these are class variables

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -762,17 +760,17 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-                    {
-                        receivedRaised = true;
-                        throw new InvalidOperationException();
-                    }
-
-                    receiveTcs.TrySetResult(Encoding.UTF8.GetString(data));
-                    return Task.CompletedTask;
-                };
+                connection.OnReceived((data) =>
+               {
```
moozzyk: formatting

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -120,7 +120,7 @@ public class HubConnectionTests
             Assert.Same(exception, thrown);
         }
 
-        [Fact]
+        [Fact(Skip = "Will add this back")]
```
moozzyk: Fix the test, remove `Skip`

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -13,6 +16,7 @@
 using Microsoft.AspNetCore.Sockets.Client;
 using Microsoft.AspNetCore.Sockets.Features;
 using Newtonsoft.Json;
+using static Microsoft.AspNetCore.Sockets.Client.HttpConnection;
```
moozzyk: needed?

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
```
moozzyk: at least some of these don't seem to be needed

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
```diff
@@ -10,69 +10,79 @@ namespace Microsoft.AspNetCore.SignalR.Client
...
-                handler(parameters);
+                var currentHandler = (Action<object[]>)state;
+                currentHandler(parameters);
                 return Task.CompletedTask;
-            });
+            }, handler);
         }
 
-        public static void On(this HubConnection hubConnection, string methodName, Action handler)
+        public static IDisposable On(this HubConnection hubConnetion, string methodName, Type[] parameterTypes, Func<object[], Task> handler)
```
moozzyk: Can you move this to the bottom of this file (i.e. as the last method)? This separates closely related extension methods.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
```diff
@@ -10,69 +10,79 @@ namespace Microsoft.AspNetCore.SignalR.Client
...
+        {
+            return hubConnetion.On(methodName, parameterTypes, (parameters, state) =>
```
moozzyk: typo: should be `hubConnection`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
moozzyk: Should we actually have overloads for different parameters instead in addition to this - similarly to what we have for `Actions`?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Should I make a extensions file for just this method?

moozzyk: Yes.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.cs
mikaelm12: Hmm. Maybe I can add this in another PR?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: readonly

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -272,18 +279,26 @@ private Task Shutdown(Exception ex = null)
...
                 _logger.MissingHandler(invocation.Target);
-                return Task.CompletedTask;
+                return;
             }
 
-            // TODO: Return values
-            // TODO: Dispatch to a sync context to ensure we aren't blocking this loop.
-            return handler.Handler(invocation.Arguments);
+            // Copying the callbacks to avoid concurrency issues
+            InvocationHandler[] copiedHandlers;
```
davidfowl: Put a todo here to optimize this

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
mikaelm12: Replaced this test with a functional test  https://github.com/aspnet/SignalR/blob/mikaelm12/ClientAPIPart2/test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs#L169

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -166,6 +166,37 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                }
+                catch (Exception ex)
+                {
+                    exceptionThrown = true;
+                    loggerFactory.CreateLogger<HubConnectionTests>().LogError(ex, "Exception from test");
+                    throw;
+                }
+                finally
+                {
+                    Assert.False(exceptionThrown);
```
BrennanConroy: Remove

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -166,6 +166,37 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                try
+                {
+                    await connection.StartAsync().OrTimeout();
+
+                    await connection.InvokeAsync("CallHandlerThatDoesntExist").OrTimeout();
+
+                }
+                catch (Exception ex)
+                {
+                    exceptionThrown = true;
```
BrennanConroy: not needed

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
```diff
@@ -24,7 +24,15 @@ public string Echo(string message)
             await Clients.Client(Context.ConnectionId).InvokeAsync("Echo", message);
         }
 
+        public async Task CallHandlerThatDoesntExist()
+        {
+            await Clients.Client(Context.ConnectionId).InvokeAsync("NoClientHandler");
+        }
+
+        public void NoClientHandler() { }
```
BrennanConroy: These shouldn't be needed right?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
```diff
@@ -24,7 +24,15 @@ public string Echo(string message)
...
+
         public IObservable<int> Stream(int count)
+
```
BrennanConroy: revert

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -425,24 +461,40 @@ public Type GetReturnType(string invocationId)
...
             {
-                Handler = handler;
+                _callback = callback;
                 ParameterTypes = parameterTypes;
+                _state = state;
+            }
+
+            public async Task InvokeAsync(object[] parameters)
+            {
+                await _callback(parameters, _state);
```
BrennanConroy: nit: just return, no need to await

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
BrennanConroy: Message is now "An exception was thrown from the 'callBackObject' callback". I think 'OnReceived' is more appropriate. And the function name should be `ExceptionThrownFromCallback`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,52 @@ private void SetTransferMode(TransferMode transferMode)
...
+        {
+            var callBack = new ReceiveCallBack(callback, state);
+            lock (_callbacks)
+            {
+                _callbacks.Add(callBack);
+            }
+            return new Subscription(callBack, _callbacks);
+        }
+
+        private class ReceiveCallBack
```
BrennanConroy: nit: ReceiveCallback

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
```diff
@@ -151,7 +151,7 @@ internal static class SocketClientLoggerExtensions
             LoggerMessage.Define<DateTime, string>(LogLevel.Information, new EventId(18, nameof(StoppingClient)), "{time}: Connection Id {connectionId}: Stopping client.");
 
         private static readonly Action<ILogger, DateTime, string, string, Exception> _exceptionThrownFromHandler =
-            LoggerMessage.Define<DateTime, string, string>(LogLevel.Error, new EventId(19, nameof(ExceptionThrownFromHandler)), "{time}: Connection Id {connectionId}: An exception was thrown from the '{eventHandlerName}' event handler.");
+            LoggerMessage.Define<DateTime, string, string>(LogLevel.Error, new EventId(19, nameof(ExceptionThrownFromHandler)), "{time}: Connection Id {connectionId}: An exception was thrown from the '{eventHandlerName}' callback");
```
BrennanConroy: `{eventHandlerName}` will need to be changed also

mikaelm12:
Updated

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
```diff
@@ -103,6 +103,10 @@ internal static class SignalRClientLoggerExtensions
         private static readonly Action<ILogger, string, Exception> _streamItemOnNonStreamInvocation =
             LoggerMessage.Define<string>(LogLevel.Error, new EventId(4, nameof(StreamItemOnNonStreamInvocation)), "Invocation {invocationId} received stream item but was invoked as a non-streamed invocation.");
 
+        private static readonly Action<ILogger, string, Exception> _exceptionThrownFromCallback =
+           LoggerMessage.Define<string>(LogLevel.Error, new EventId(23, nameof(ExceptionThrownFromCallback)), "An exception was thrown from the '{callback}' callback");
+
+
```
BrennanConroy: nit: remove newline

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -299,18 +306,34 @@ private Task Shutdown(Exception ex = null)
...
-            // TODO: Dispatch to a sync context to ensure we aren't blocking this loop.
-            return handler.Handler(invocation.Arguments);
+            //TODO: Optimize this!
+            // Copying the callbacks to avoid concurrency issues
+            InvocationHandler[] copiedHandlers;
+            lock (handlers)
+            {
+                copiedHandlers = new InvocationHandler[handlers.Count];
+                handlers.CopyTo(copiedHandlers);
+            }
```
moozzyk: nit: add a new line

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -119,10 +119,17 @@ private IDataEncoder GetDataEncoder(TransferMode requestedTransferMode, Transfer
...
-        public void On(string methodName, Type[] parameterTypes, Func<object[], Task> handler)
+        public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], object, Task> handler, object state)
         {
-            var invocationHandler = new InvocationHandler(parameterTypes, handler);
-            _handlers.AddOrUpdate(methodName, invocationHandler, (_, __) => invocationHandler);
+            var invocationHandler = new InvocationHandler(parameterTypes, handler, state);
+            var invocationList = _handlers.AddOrUpdate(methodName,  _ =>  new List<InvocationHandler> { invocationHandler }, 
+                (_, invocations) =>
+                {
+                    invocations.Add(invocationHandler);
```
moozzyk: doesn't this need a lock? You are updating the a which can be used elsewhere.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -425,24 +468,40 @@ public Type GetReturnType(string invocationId)
...
+                    }
+                    throw new InvalidOperationException($"There are no callbacks registered for the method '{methodName}'");
+                }
             }
         }
 
         private struct InvocationHandler
         {
-            public Func<object[], Task> Handler { get; }
+            private Func<object[], object, Task> _callback;
```
moozzyk: nit: _callback should be `readonly` 
also group private variables together

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -37,15 +39,14 @@ public class HttpConnection : IConnection
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
         private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
-
         private ReadableChannel<byte[]> Input => _transportChannel.In;
         private WritableChannel<SendMessage> Output => _transportChannel.Out;
+        private List<ReceiveCallback> _callbacks = new List<ReceiveCallback>();
```
moozzyk: `readonly`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,52 @@ private void SetTransferMode(TransferMode transferMode)
             _httpClient.Dispose();
         }
 
+        public IDisposable OnReceived(Func<byte[], object, Task> callback, object state)
+        {
+            var callBack = new ReceiveCallback(callback, state);
```
moozzyk: you need to rename this variable - you have two different variables which differ just by casing in one place. Call it `receiveCallback` or something to that effect.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -444,6 +452,52 @@ private void SetTransferMode(TransferMode transferMode)
...
+            }
+            return new Subscription(callBack, _callbacks);
+        }
+
+        private class ReceiveCallback
+        {
+            private readonly Func<byte[], object, Task> _callback;
+            private readonly object _state;
+
+            public ReceiveCallback(Func<byte[], object, Task> callBack, object state)
```
moozzyk: `callBack` fix casing

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -166,6 +166,33 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+            using (StartLog(out var loggerFactory))
+            {
+                var httpConnection = new HttpConnection(new Uri(_serverFixture.BaseUrl + path), transportType, loggerFactory);
+                var connection = new HubConnection(httpConnection, protocol, loggerFactory);
+                try
+                {
+                    await connection.StartAsync().OrTimeout();
+
+                    await connection.InvokeAsync("CallHandlerThatDoesntExist").OrTimeout();
+
```
moozzyk: nit: add empty line
I think you may also want to check that there is no exception in the close event. I think if the connection was closed due to an exception it would show up there.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -139,7 +139,7 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
 
             var transport = new Mock<ITransport>();
             transport.Setup(t => t.StopAsync()).Returns(async () => { await releaseDisposeTcs.Task; });
-            var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(transport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
+            var connection = new HttpConnection(new Uri("http://fakeuri.org/"), transportFactory: new TestTransportFactory(transport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
```
moozzyk: I don't think you need the name - the type we new up tells you what you need to know. (I don't care too much though)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -255,7 +255,7 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
         }
 
         [Fact]
-        public async Task ReceivedEventNotRaisedAfterConnectionIsDisposed()
+        public async Task ReceivedCallBackNotRaisedAfterConnectionIsDisposed()
```
moozzyk: nit: Callback casing.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -336,12 +336,11 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
 
             var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
 
-            connection.Received +=
-                async m =>
+            connection.OnReceived(async _ =>
                 {
                     callbackInvokedTcs.SetResult(null);
                     await closedTcs.Task;
```
moozzyk: you could just return the task I think.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -392,11 +391,10 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
             var closedTcs = new TaskCompletionSource<object>();
 
             var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
-            connection.Received +=
-                async m =>
+            connection.OnReceived(async _ =>
```
moozzyk: `connection.OnReceived(_ => blockReceiveCallbackTcs.Task);`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -762,17 +760,17 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
                 var receivedRaised = false;
-                connection.Received += data =>
+                connection.OnReceived((data) =>
                 {
-                    if (!receivedRaised)
-                    {
-                        receivedRaised = true;
-                        throw new InvalidOperationException();
-                    }
+                   if (!receivedRaised)
```
moozzyk: nit: indentation is off

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -30,14 +31,15 @@ internal class TestConnection : IConnection
...
         public event Func<Task> Connected;
-        public event Func<byte[], Task> Received;
         public event Func<Exception, Task> Closed;
 
         public Task Started => _started.Task;
         public Task Disposed => _disposed.Task;
         public ReadableChannel<byte[]> SentMessages => _sentMessages.In;
         public WritableChannel<byte[]> ReceivedMessages => _receivedMessages.Out;
 
+        private List<ReceiveCallBack> _callbacks = new List<ReceiveCallBack>();
```
moozzyk: `readonly`

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -136,5 +141,48 @@ private byte[] FormatMessageToArray(byte[] message)
...
         }
+
+        public IDisposable OnReceived(Func<byte[], object, Task> callback, object state)
+        {
+            var receiveCallBack = new ReceiveCallBack(callback, state);
+            _callbacks.Add(receiveCallBack);
+            return new Subscription(receiveCallBack, _callbacks);
+        }
+
+        private class ReceiveCallBack
```
moozzyk: casing

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -136,5 +141,48 @@ private byte[] FormatMessageToArray(byte[] message)
...
+        {
+            public readonly Func<byte[], object, Task> CallBack;
```
moozzyk: make it a getter (and fix casing)

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -136,5 +141,48 @@ private byte[] FormatMessageToArray(byte[] message)
...
+            private readonly List<ReceiveCallBack> _callbacks;
+            public Subscription(ReceiveCallBack callback, List<ReceiveCallBack> callbacks)
+            {
+                _callback = callback;
+                _callbacks = callbacks;
+            }
+
+            public void Dispose()
+            {
+                lock (_callbacks)
```
moozzyk: locking in just one place does not make much sense. I was thinking we should make the callback list threadsafe but since we don't add/remove callbacks in our tests after the connection is started we should be fine. We might need to do this in the future if we start seeing errors caused by this.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -224,12 +225,12 @@ public static IEnumerable<object[]> MessageSizesData
...
                 {
                     var receiveTcs = new TaskCompletionSource<byte[]>();
-                    connection.Received += data =>
+                    connection.OnReceived((data, state) =>
                     {
                         logger.LogInformation("Received {length} byte message", data.Length);
                         receiveTcs.TrySetResult(data);
                         return Task.CompletedTask;
-                    };
+                    }, null);
```
moozzyk: you could pass `receiveTcs` as `state`

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -127,7 +127,8 @@ public EndToEndTests(ServerFixture<Startup> serverFixture, ITestOutputHelper out
 
                         receiveTcs.TrySetResult(Encoding.UTF8.GetString(data));
                         return Task.CompletedTask;
-                    };
+                    }, null);
```
moozzyk: you could pass `receiveTcs` as `state`

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
mikaelm12: This should actually be private since I've added the InvokeAsync method

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -122,7 +122,14 @@ private byte[] FormatMessageToArray(byte[] message)
                     {
                         while (_receivedMessages.In.TryRead(out var message))
                         {
-                            foreach (var callback in _callbacks)
+                            ReceiveCallback[] callbackCopies;
+                            lock (_callbacks)
+                            {
+                                callbackCopies = new ReceiveCallback[_callbacks.Count];
+                                _callbacks.CopyTo(callbackCopies);
```
moozzyk: Can you just do `_callbacks.ToArray()`?

PR #1007, opened 2017-10-11
AllExcept params

mikaelm12:
Changes `return Clients.AllExcept(new List<string> { Context.ConnectionId }).InvokeAsync(...)`
To           `return Clients.AllExcept(Context.ConnectionId).InvokeAsync(...)`

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -141,7 +141,7 @@ namespace Microsoft.AspNetCore.SignalR
             return invocationId.ToString();
         }
 
-        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, params string[] excludedIds)
         {
             return InvokeAllWhere(methodName, args, connection =>
             {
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -141,7 +141,7 @@ private string GetInvocationId()
             return invocationId.ToString();
         }
 
-        public override Task InvokeAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
+        public override Task InvokeAllExceptAsync(string methodName, object[] args, params string[] excludedIds)
```
BrennanConroy: I don't think lifetime managers need `params`

moozzyk:
I am not sure how useful `params` is in this case. Because to use params I would have to know the connection ids upfront and typically I cannot because they are generated by the server. If I tracked connections somehow then I would already have a collection (and not a handful of variables) which I could inject to the hub or store somehow externally. If the connections to be used in InokeAllExcept are stored somewhere externally as a collection then taking an array likely will require allocating the array and copying references instead of just passing the collection. I think we had a discussion why we chose `IReadOnlyList<T>` over `IReadOnlyCollection<T>` but I need to find it.

`params` might be useful for emulating `.Others` now but in a long run I think we should have a built in support for `.Others`.

moozzyk:
Here is the discussion on this: https://github.com/aspnet/SignalR/pull/700#discussion_r132744060 - I think we needs some discussion to decide what the method should take.

PR #1009, opened 2017-10-11
Short Circuit message to local connection in Redis

BrennanConroy:


moozzyk:
We should figure out this: https://github.com/aspnet/SignalR/issues/705

BrennanConroy:
Yeah, I'll look into tests during my redis work.

PR #1010, opened 2017-10-11
[Redis] Remove groups properly from connection disconnection

BrennanConroy:
Before we would remove the connection from the `_connections` list which would prevent `RemoveGroupAsync` from removing the group locally and cause it to send a message to other servers which would make `OnDisconnectedAsync` wait for 30 seconds for an ACK it would never receive. And then `OnDisconnectedAsync` would throw a `TaskCanceledException`. The reason I didn't move the `_connections.Remove(connection);` to the end of `OnDisconnectedAsync` is to prevent groups being added to the connection while it is clearing out groups.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -122,7 +122,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
                 if (groupMessage.Action == GroupAction.Remove)
                 {
                     var connection = _connections[groupMessage.ConnectionId];
-                    if (connection != null && !await RemoveGroupAsyncCore(connection, groupMessage.Group))
+                    if (connection == null || !await RemoveGroupAsyncCore(connection, groupMessage.Group))
                     {
                         // user not on this server
                         return;
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -121,7 +121,8 @@ namespace Microsoft.AspNetCore.SignalR.Redis
 
                 if (groupMessage.Action == GroupAction.Remove)
                 {
-                    if (!await RemoveGroupAsyncCore(groupMessage.ConnectionId, groupMessage.Group))
+                    var connection = _connections[groupMessage.ConnectionId];
+                    if (connection != null && !await RemoveGroupAsyncCore(connection, groupMessage.Group))
                     {
                         // user not on this server
                         return;
@@ -274,7 +275,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             return Task.WhenAll(connectionTask, userTask);
         }
 
-        public override Task OnDisconnectedAsync(HubConnectionContext connection)
+        public override async Task OnDisconnectedAsync(HubConnectionContext connection)
         {
             _connections.Remove(connection);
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -408,7 +413,11 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
             await SendGroupActionAndWaitForAck(connectionId, groupName, GroupAction.Remove);
         }
 
-        private async Task<bool> RemoveGroupAsyncCore(string connectionId, string groupName)
+        /// <summary>
+        /// This takes <see cref="HubConnectionContext"/> because we want to remove the connection from the
+        /// _connections list in OnDisconnectedAsync and still be able to remove groups with this method.
+        /// </summary>
+        private async Task<bool> RemoveGroupAsyncCore(HubConnectionContext connection, string groupName)
```
mikaelm12: Should this be called `TryRemoveGroupAsyncCore` ?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -121,7 +121,8 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
 
                 if (groupMessage.Action == GroupAction.Remove)
                 {
-                    if (!await RemoveGroupAsyncCore(groupMessage.ConnectionId, groupMessage.Group))
+                    var connection = _connections[groupMessage.ConnectionId];
+                    if (connection == null || !await RemoveGroupAsyncCore(connection, groupMessage.Group))
```
mikaelm12: Can you put parents around `await RemoveGroupAsyncCore(connection, groupMessage.Group)`. Looks weird to me to see the ! next to the await

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
BrennanConroy: I like it better as is since it maps to the public method `RemoveGroupAsync`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -121,7 +121,8 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
 
                 if (groupMessage.Action == GroupAction.Remove)
                 {
-                    if (!await RemoveGroupAsyncCore(groupMessage.ConnectionId, groupMessage.Group))
+                    var connection = _connections[groupMessage.ConnectionId];
+                    if (connection == null || !(await RemoveGroupAsyncCore(connection, groupMessage.Group)))
                     {
                         // user not on this server
```
moozzyk: update the comment? `RemoveGroupAsyncCore` returns false if the group does not exist.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -399,7 +402,9 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
                 throw new ArgumentNullException(nameof(groupName));
             }
 
-            if (await RemoveGroupAsyncCore(connectionId, groupName))
+
+            var connection = _connections[connectionId];
+            if (connection != null && await RemoveGroupAsyncCore(connection, groupName))
```
moozzyk: Is this correct - if the connection is local (i.e. connection is not null) but the group does not exist we will send to other servers. Shouldn't this just be:

```
if (connection != null)
{
    await RemoveGroupAsyncCore(connection, groupName);
    return;
}
```
?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -501,7 +504,7 @@ private T DeserializeMessage<T>(RedisValue data)
         {
             using (var reader = new JsonTextReader(new StreamReader(new MemoryStream(data))))
             {
-                return (T)_serializer.Deserialize(reader);
+                return _serializer.Deserialize<T>(reader);
```
moozzyk: 👍

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
BrennanConroy: Good catch

BrennanConroy:
Title is still accurate

moozzyk:
'diconnection' is not a word AFAICT :P

PR #1013, opened 2017-10-12
Adding Redis Unit tests

BrennanConroy:
There will be more tests, but that can be in a later PR. For now want to get this in as 'Proof of concept'


test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,264 @@
...
+        [Fact]
+        public async Task InvokeAllAsyncWritesToAllConnectionsOutput()
+        {
+            using (var client1 = new TestClient())
+            using (var client2 = new TestClient())
+            {
+                var output1 = Channel.CreateUnbounded<HubMessage>();
+                var output2 = Channel.CreateUnbounded<HubMessage>();
+
+                var manager = new RedisHubLifetimeManager<MyHub>(new LoggerFactory().CreateLogger<RedisHubLifetimeManager<MyHub>>(), Options.Create(new RedisOptions()
```
davidfowl: Can you make this line longer 😄

test/Microsoft.AspNetCore.SignalR.Redis.Tests/TestConnectionMultiplexer.cs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        public long Publish(RedisChannel channel, RedisValue message, CommandFlags flags = CommandFlags.None)
+        {
+            if (_globalSubscriptions.TryGetValue(channel, out var handlers))
+            {
+                foreach (var handler in handlers)
+                {
+                    handler(channel, message);
+                }
```
moozzyk: Should you return `handlers.Count` here?

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,264 @@
...
+                await manager.OnConnectedAsync(connection);
+
+                await manager.InvokeConnectionAsync(connection.ConnectionId, "Hello", new object[] { "World" });
+
+                AssertMessage(output);
+            }
+        }
+
+        [Fact]
+        public async Task InvokeConnectionAsyncOnNonExistentConnectionNoops()
```
moozzyk: This test is a bit weird - the only thing you check here is really that it does not throw and not that it no ops.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/TestConnectionMultiplexer.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        public void WaitAll(params Task[] tasks)
+        {
+            throw new NotImplementedException();
+        }
+    }
+
+    public class TestSubscriber : ISubscriber
+    {
+        private static ConcurrentDictionary<RedisChannel, List<Action<RedisChannel, RedisValue>>> _globalSubscriptions =
```
moozzyk: Add a comment why do you need both `_globalSubscriptions` and `_subscriptions`. If I read the code correctly it is to prevent from removing a subscription using a different instance that was used to add it.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/TestConnectionMultiplexer.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        public bool PreserveAsyncOrder { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+        public bool IsConnected => true;
+
+        public bool IncludeDetailInExceptions { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+        public int StormLogThreshold { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+        // The event '---' is never used
+#pragma warning disable CS0067
```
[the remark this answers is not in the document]
BrennanConroy: I'm not removing them from the interface, I'm shutting up the compiler because it complains about us not using them. When we do something with these in the `RedisHubLifetimeManager` then we can implement the behavior in the `TestConnectionMultiplexer` and remove the warning maybe.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/TestConnectionMultiplexer.cs
BrennanConroy: Will add.
The reason is we are faking a redis server via the `_globalSubscriptions` property. So individual servers need a local copy of their own subscriptions and when publishing need the "server" subscriptions to write to.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/TestConnectionMultiplexer.cs
BrennanConroy: I guess

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
BrennanConroy: It is the exact same in DefaultHubLifetimeManagerTests but I'll change it to say DoesNotThrow

test/Microsoft.AspNetCore.SignalR.Redis.Tests/TestConnectionMultiplexer.cs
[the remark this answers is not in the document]
moozzyk: I understand that. I am saying the suppression should not be needed because we have to have these because they are part of the interface. We don't use them but we cannot remove because the code wouldn't compile. This is why I find it weird that we need to suppress warnings.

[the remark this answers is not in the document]
moozzyk: OK. I think I [found it](https://blogs.msdn.microsoft.com/trevor/2008/08/14/c-warning-cs0067-the-event-event-is-never-used/). You need: 
```
public event EventHandler Unimportant
{
    add { }
    remove { }
}
```

BrennanConroy:
🆙 📅 
Rebased on dev and added a couple new tests that https://github.com/aspnet/SignalR/pull/1010 fixed

PR #1020, opened 2017-10-13
Wait for test server to finish starting up before starting tests

BrennanConroy:
On slower machines (our Win7 machine) the first test can run and try to start a connection before the in-proc server finished starting up. This would of course cause a test failure.

@davidfowl @moozzyk

### DIFF
--- a/test/Common/ServerFixture.cs
+++ b/test/Common/ServerFixture.cs
@@ -3,9 +3,8 @@
 
 using System;
 using System.IO;
-using System.Threading.Tasks;
+using System.Threading;
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Testing;
 
@@ -18,7 +17,6 @@ namespace Microsoft.AspNetCore.SignalR.Tests.Common
         private ILoggerFactory _loggerFactory;
         private ILogger _logger;
         private IWebHost _host;
-        private IApplicationLifetime _lifetime;
         private readonly IDisposable _logToken;
 
         public string BaseUrl => "http://localhost:3000";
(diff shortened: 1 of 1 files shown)


test/Common/ServerFixture.cs
```diff
@@ -63,6 +63,8 @@ private void StartServer()
                 _logger.LogInformation("Test server shut down");
                 _logToken.Dispose();
             });
+
+            t.Wait();
```
davidfowl: Change all of the code to use the new assemblies.

BrennanConroy: What?

moozzyk: aren't we actually waiting already on line 50?

BrennanConroy: Stuff still happens after registered `IApplicationLifetime`s are started.

moozzyk: Let's talk about this on Monday. This at least needs a timeout but if _lifetime.ApplicationStarted is not good enough then shouldn't we replace it entirely with waiting on `t` ?

moozzyk: Only this happens in a ctor and I am not sure if this can be changed easily since it is an XUnit fixture.

davidfowl: We should be able to  remove our custom logic and use the new methods. It should be enough to cover these scenarios.

BrennanConroy: How, the new APIs I see are `Run` and `RunAsync` both of which just call `Start` but give no indication of when `Start` finishes.

BrennanConroy:
:up: 📅 Removed old logic, just call `host.StartAsync` now with a 30 second timeout

test/Common/ServerFixture.cs
```diff
@@ -44,32 +42,22 @@ private void StartServer()
...
-            if (!_lifetime.ApplicationStarted.WaitHandle.WaitOne(TimeSpan.FromSeconds(5)))
+            using (var cts = new CancellationTokenSource())
             {
-                // t probably faulted
-                if (t.IsFaulted)
-                {
-                    throw t.Exception.InnerException;
-                }
-                throw new TimeoutException("Timed out waiting for application to start.");
+                cts.CancelAfter(TimeSpan.FromSeconds(30));
```
davidfowl: Why do we need to time this out ever?

BrennanConroy: We want to be safe, don't want it to hang for any reason.

test/Common/ServerFixture.cs
```diff
@@ -44,32 +42,22 @@ private void StartServer()
...
+                _host.StartAsync(cts.Token).GetAwaiter().GetResult();
```
moozzyk: Wouldn't throw when the token is cancelled? Do we need to catch this exception (or any other exception actually) log and swallow?

BrennanConroy: We do something similar to what we did before
```
try
{
    _host.StartAsync(cts.Token).GetAwaiter().GetResult();
}
catch (TaskCanceledException)
{
    throw new TimeoutException("Timed out waiting for application to start.");
}
```

moozzyk: I think it would be useful because it immediately tells what happened. With TaskCanceledException you would have to go to the code to understand what really happened.

test/Common/ServerFixture.cs
davidfowl: Lets remove this.

test/Common/ServerFixture.cs
davidfowl: Lets just remove the whole thing. I don't understand why our Startup would ever hang. If you have a good reason, let me know because none of the other repositories do this AFAIK.

moozzyk: I just talked with @BrennanConroy and I think we need to take a second look at this because we could not find the reason why the original approach would not work and why the failures would be fixed by this change.

davidfowl: We should simplify the code regardless. None of the existing code is required.

moozzyk: Copied from MVC :P

Issue #802, opened 2017-08-31
TS ConsoleLogger could use console.log()/console.warn()/console.error()


moozzyk:
At the moment the ConsoleLogger logs everything [using `console.log()`](https://github.com/aspnet/SignalR/blob/7db480b23e8d11a384b88a34704468341429c562/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Loggers.ts#L17). We could use `console.warn()` for warnings, `console.error()` for errors and `console.log()` for everything else.

moozzyk:
Fixed with a709158

PR #1021, opened 2017-10-14
Improved ts client console logging

RoBackman:
Summary of the changes
    - Using console specific methods for logging Error, Warning and Information messages

Addresses #802

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Loggers.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Loggers.ts
@@ -17,7 +17,20 @@ export class ConsoleLogger implements ILogger {
 
     log(logLevel: LogLevel, message: string): void {
         if (logLevel >= this.minimumLogLevel) {
-            console.log(`${LogLevel[logLevel]}: ${message}`);
+            switch (logLevel) {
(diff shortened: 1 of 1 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Loggers.ts
```diff
@@ -17,7 +17,20 @@ export class ConsoleLogger implements ILogger {
 
     log(logLevel: LogLevel, message: string): void {
         if (logLevel >= this.minimumLogLevel) {
-            console.log(`${LogLevel[logLevel]}: ${message}`);
+            switch (logLevel) {
+                case LogLevel.Error:
+                    console.error(`${LogLevel[logLevel]}: ${message}`);
```
muratg: @moozzyk, did we have any reason not to use console.error/warn/info calls instead of log?  Otherwise this looks good to me.

@RoBackman, thanks for the PR!

moozzyk: The reason was that I only learned about console.error etc. after I merge the original PR and I did not want to add churn to alpha releases - so I opened #802

moozzyk:
Thanks @RoBackman!

PR #1023, opened 2017-10-16
Add RepositoryRoot

ryanbrandenburg:


Directory.Build.props
```diff
@@ -1,10 +1,11 @@
-﻿<Project>
+<Project>
   <Import Project="version.xml" />
 
   <PropertyGroup>
     <Product>Microsoft ASP.NET Core</Product>
     <RepositoryUrl>https://github.com/aspnet/SignalR</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
+    <RepositoryRoot>$(MSBuildThisFileDirectory)</RepositoryRoot>
```
moozzyk: Out of curiosity - how this is being used?

ryanbrandenburg: The backstory here is that when we added SourceLink we used the RepositoryRoot property in the target. RepositoryRoot is set by KoreBuild, but if you launch your sln in VS you'll get a warning for every project. Therefore we added the property here to make SourceLink work in both cases.

moozzyk: I see. I actually started seeing this warnings and they are quite annoying. I was wondering what was the cause. Now I know. Thanks for the explanation!

PR #1024, opened 2017-10-17
Client API Cancellation tokens

mikaelm12:
Part of #900

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -66,11 +66,11 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _connection.Closed += Shutdown;
         }
 
-        public async Task StartAsync()
+        public async Task StartAsync(CancellationToken cancellationToken = default)
         {
             try
             {
-                await StartAsyncCore().ForceAsync();
+                await StartAsyncCore(cancellationToken).ForceAsync();
             }
             finally
             {
@@ -78,7 +78,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        private async Task StartAsyncCore()
+        private async Task StartAsyncCore(CancellationToken cancellationToken)
         {
             var transferModeFeature = _connection.Features.Get<ITransferModeFeature>();
             if (transferModeFeature == null)
@@ -93,7 +93,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                     : TransferMode.Text;
 
             transferModeFeature.TransferMode = requestedTransferMode;
-            await _connection.StartAsync();
+            await _connection.StartAsync(cancellationToken);
             var actualTransferMode = transferModeFeature.TransferMode;
 
             _protocolReaderWriter = new HubProtocolReaderWriter(_protocol, GetDataEncoder(requestedTransferMode, actualTransferMode));
@@ -103,7 +103,15 @@ namespace Microsoft.AspNetCore.SignalR.Client
             using (var memoryStream = new MemoryStream())
             {
                 NegotiationProtocol.WriteMessage(new NegotiationMessage(_protocol.Name), memoryStream);
-                await _connection.SendAsync(memoryStream.ToArray(), _connectionActive.Token);
+                if (cancellationToken.CanBeCanceled)
+                {
+                    cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _connectionActive.Token).Token;
+                }
+                else
+                {
+                    cancellationToken = _connectionActive.Token;
+                }
+                await _connection.SendAsync(memoryStream.ToArray(), cancellationToken);
             }
         }
 
@@ -121,11 +129,11 @@ namespace Microsoft.AspNetCore.SignalR.Client
             return new PassThroughEncoder();
         }
 
-        public async Task DisposeAsync() => await DisposeAsyncCore().ForceAsync();
+        public async Task DisposeAsync(CancellationToken cancellationToken = default) => await DisposeAsyncCore(cancellationToken).ForceAsync();
 
-        private async Task DisposeAsyncCore()
+        private async Task DisposeAsyncCore(CancellationToken cancellationToken)
         {
-            await _connection.DisposeAsync();
+            await _connection.DisposeAsync(cancellationToken);
         }
 
         // TODO: Client return values/tasks?
@@ -157,6 +165,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 throw new InvalidOperationException($"The '{nameof(StreamAsync)}' method cannot be called before the connection has been started.");
             }
 
+            cancellationToken.ThrowIfCancellationRequested();
             var invokeCts = new CancellationTokenSource();
             var irq = InvocationRequest.Stream(invokeCts.Token, returnType, GetNextId(), _loggerFactory, this, out var channel);
             // After InvokeCore we don't want the irq cancellation token to be triggered.
@@ -185,7 +194,6 @@ namespace Microsoft.AspNetCore.SignalR.Client
                     }
                 }, irq);
             }
-
             return channel;
         }
 
@@ -199,6 +207,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 throw new InvalidOperationException($"The '{nameof(InvokeAsync)}' method cannot be called before the connection has been started.");
             }
 
+            cancellationToken.ThrowIfCancellationRequested();
             var irq = InvocationRequest.Invoke(cancellationToken, returnType, GetNextId(), _loggerFactory, this, out var task);
             await InvokeCore(methodName, irq, args);
             return await task;
--- a/src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
@@ -10,9 +10,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public interface IConnection
     {
-        Task StartAsync();
+        Task StartAsync(CancellationToken cancellationToken);
         Task SendAsync(byte[] data, CancellationToken cancellationToken);
-        Task DisposeAsync();
+        Task DisposeAsync(CancellationToken cancellationToken);
 
         IDisposable OnReceived(Func<byte[], object, Task> callback, object state);
 
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -99,9 +99,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
             _transportFactory = transportFactory ?? throw new ArgumentNullException(nameof(transportFactory));
         }
 
-        public async Task StartAsync() => await StartAsyncCore().ForceAsync();
+        public async Task StartAsync(CancellationToken cancellationToken = default) => await StartAsyncCore(cancellationToken).ForceAsync();
 
-        private Task StartAsyncCore()
+        private Task StartAsyncCore(CancellationToken cancellationToken)
         {
             if (Interlocked.CompareExchange(ref _connectionState, ConnectionState.Connecting, ConnectionState.Initial)
                 != ConnectionState.Initial)
@@ -110,7 +110,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                     new InvalidOperationException("Cannot start a connection that is not in the Initial state."));
             }
 
-            StartAsyncInternal()
+            StartAsyncInternal(cancellationToken)
                 .ContinueWith(t =>
                 {
                     if (t.IsFaulted)
@@ -130,10 +130,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
             return _startTcs.Task;
         }
 
-        private async Task StartAsyncInternal()
+        private async Task StartAsyncInternal(CancellationToken cancellationToken)
         {
             _logger.HttpConnectionStarting();
-
             try
             {
                 var connectUrl = Url;
@@ -143,7 +142,8 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 }
                 else
                 {
-                    var negotiationResponse = await Negotiate(Url, _httpClient, _logger);
+                    var negotiationResponse = await Negotiate(Url, _httpClient, _logger, cancellationToken);
+
                     _connectionId = negotiationResponse.ConnectionId;
 
                     // Connection is being stopped while start was in progress
@@ -158,7 +158,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 }
 
                 _logger.StartingTransport(_connectionId, _transport.GetType().Name, connectUrl);
-                await StartTransport(connectUrl);
+                await StartTransport(connectUrl, cancellationToken);
             }
             catch
             {
@@ -212,7 +212,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
         }
 
-        private async static Task<NegotiationResponse> Negotiate(Uri url, HttpClient httpClient, ILogger logger)
+        private async static Task<NegotiationResponse> Negotiate(Uri url, HttpClient httpClient, ILogger logger, CancellationToken cancellationToken)
         {
             try
             {
@@ -221,7 +221,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 using (var request = new HttpRequestMessage(HttpMethod.Options, url))
                 {
                     request.Headers.UserAgent.Add(Constants.UserAgentHeader);
-                    using (var response = await httpClient.SendAsync(request))
+                    using (var response = await httpClient.SendAsync(request, cancellationToken))
                     {
                         response.EnsureSuccessStatusCode();
                         return await ParseNegotiateResponse(response, logger);
@@ -284,7 +284,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             return Utils.AppendQueryString(url, "id=" + negotiationResponse.ConnectionId);
         }
 
-        private async Task StartTransport(Uri connectUrl)
+        private async Task StartTransport(Uri connectUrl, CancellationToken cancellationToken)
         {
             var applicationToTransport = Channel.CreateUnbounded<SendMessage>();
             var transportToApplication = Channel.CreateUnbounded<byte[]>();
@@ -294,7 +294,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             // Start the transport, giving it one end of the pipeline
             try
             {
-                await _transport.StartAsync(connectUrl, applicationSide, requestedTransferMode: GetTransferMode(), connectionId: _connectionId);
+                await _transport.StartAsync(connectUrl, applicationSide, requestedTransferMode: GetTransferMode(), connectionId: _connectionId, cancellationToken: cancellationToken);
 
                 // actual transfer mode can differ from the one that was requested so set it on the feature
                 Debug.Assert(_transport.Mode.HasValue, "transfer mode not set after transport started");
@@ -426,9 +426,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
         }
 
-        public async Task DisposeAsync() => await DisposeAsyncCore().ForceAsync();
+        public async Task DisposeAsync(CancellationToken cancellationToken = default) => await DisposeAsyncCore(cancellationToken).ForceAsync();
 
-        private async Task DisposeAsyncCore()
+        private async Task DisposeAsyncCore(CancellationToken cancellationToken)
         {
             _logger.StoppingClient(_connectionId);
 
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ITransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ITransport.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Channels;
 
@@ -9,7 +10,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public interface ITransport
     {
-        Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId);
+        Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId, CancellationToken cancellationToken);
         Task StopAsync();
         TransferMode? Mode { get; }
     }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -50,11 +50,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
             Mode = requestedTransferMode;
             _connectionId = connectionId;
 
-            if (cancellationToken.CanBeCanceled)
-            {
-                _transportCts = CancellationTokenSource.CreateLinkedTokenSource(_transportCts.Token, cancellationToken);
-            }
-
             _logger.StartTransport(_connectionId, Mode.Value);
 
             // Start sending and polling (ask for binary if the server supports it)
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private Task _poller;
         private string _connectionId;
 
-        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+        private CancellationTokenSource _transportCts = new CancellationTokenSource();
 
         public Task Running { get; private set; } = Task.CompletedTask;
 
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<LongPollingTransport>();
         }
 
-        public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId)
+        public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId, CancellationToken cancellationToken = default)
         {
             if (requestedTransferMode != TransferMode.Binary && requestedTransferMode != TransferMode.Text)
             {
@@ -49,6 +49,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
             _application = application;
             Mode = requestedTransferMode;
             _connectionId = connectionId;
+
+            if (cancellationToken.CanBeCanceled)
+            {
+                _transportCts = CancellationTokenSource.CreateLinkedTokenSource(_transportCts.Token, cancellationToken);
+            }
 
             _logger.StartTransport(_connectionId, Mode.Value);
 
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -57,11 +57,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             _logger.StartTransport(_connectionId, Mode.Value);
 
-            if (cancellationToken.CanBeCanceled)
-            {
-                _transportCts = CancellationTokenSource.CreateLinkedTokenSource(_transportCts.Token, cancellationToken);
-            }
-
             var sendTask = SendUtils.SendMessages(url, _application, _httpClient, _transportCts, _logger, _connectionId);
             var receiveTask = OpenConnection(_application, url, _transportCts.Token);
 
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -19,7 +19,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
     {
         private readonly HttpClient _httpClient;
         private readonly ILogger _logger;
-        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+        private CancellationTokenSource _transportCts = new CancellationTokenSource();
         private readonly ServerSentEventsMessageParser _parser = new ServerSentEventsMessageParser();
         private string _connectionId;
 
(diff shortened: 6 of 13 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -24,22 +25,22 @@ public class HubConnectionTests
         {
             var connection = new Mock<IConnection>();
             connection.SetupGet(p => p.Features).Returns(new FeatureCollection());
-            connection.Setup(m => m.StartAsync()).Returns(Task.CompletedTask).Verifiable();
+            connection.Setup(m => m.StartAsync(new CancellationToken())).Returns(Task.CompletedTask).Verifiable();
```
davidfowl: Why are you passing them.

mikaelm12: I was getting a weird error that said something about not being able to use default values...

BrennanConroy: Use `It.IsAny<CancellationToken>()`?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -24,22 +25,22 @@ public class HubConnectionTests
...
             var hubConnection = new HubConnection(connection.Object, Mock.Of<IHubProtocol>(), null);
             await hubConnection.StartAsync();
 
-            connection.Verify(c => c.StartAsync(), Times.Once());
+            connection.Verify(c => c.StartAsync(new CancellationToken()), Times.Once());
```
moozzyk: I think it passes by accident (if it does) - you need yo use `It.IsAny<CancellationToken>()` here as well

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -404,9 +403,9 @@ private void SetTransferMode(TransferMode transferMode)
             }
         }
 
-        public async Task DisposeAsync() => await DisposeAsyncCore().ForceAsync();
+        public async Task DisposeAsync(CancellationToken cancellationToken = default) => await DisposeAsyncCore(cancellationToken).ForceAsync();
```
moozzyk: What is the point of taking cancellation token here if it is not used? Is it for custom implementations?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -123,13 +123,12 @@ private Task StartAsyncCore()
             return _startTcs.Task;
         }
 
-        private async Task StartAsyncInternal()
+        private async Task StartAsyncInternal(CancellationToken cancellationToken)
```
moozzyk: This cancellation token seems only to be used for the negotiate request. How about cancelling starting the transport?

moozzyk:
I don't see any tests testing cancellation
Also, should we proactively check at entry points for cancelled tokens and throw before doing any work?

muratg:
Do we have a sample API usage for cancellation?  Perhaps a functional test could demonstrate this...

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -404,9 +403,9 @@ private void SetTransferMode(TransferMode transferMode)
             }
         }
 
-        public async Task DisposeAsync() => await DisposeAsyncCore().ForceAsync();
+        public async Task DisposeAsync(CancellationToken cancellationToken = default) => await DisposeAsyncCore(cancellationToken).ForceAsync();
 
-        private async Task DisposeAsyncCore()
+        private async Task DisposeAsyncCore(CancellationToken cancellationToken)
```
muratg: Not used inside the method -- so it it actually required?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: That's something I wanted to ask about. It's listed in the redesign notes but I'm unsure about the actual usage..

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Adding it to the the transports StartAsync methods

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: We want the cancellation token in dispose in case the receive task takes a while. @davidfowl wants to make the API change now but to hold off implementing the logic now because it might be a bit trickier than we think right now

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -93,7 +93,8 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             using (var memoryStream = new MemoryStream())
             {
                 NegotiationProtocol.WriteMessage(new NegotiationMessage(_protocol.Name), memoryStream);
-                await _connection.SendAsync(memoryStream.ToArray(), _connectionActive.Token);
+                var linkedTcs = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _connectionActive.Token);
```
BrennanConroy: We should be checking if the `cancellationToken` passed in is cancel able, and if not avoid the linkedtcs allocations.

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -57,10 +57,11 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
             _application = application;
             Mode = requestedTransferMode;
             _connectionId = connectionId;
+            var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(_transportCts.Token, cancellationToken);
 
             _logger.StartTransport(_connectionId, Mode.Value);
 
-            await Connect(url);
+            await Connect(url, linkedCts.Token);
             var sendTask = SendMessages(url);
```
BrennanConroy: Token should also be passed to SendMessages and ReceiveMessages

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -279,7 +278,7 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
             // Start the transport, giving it one end of the pipeline
             try
             {
-                await _transport.StartAsync(connectUrl, applicationSide, requestedTransferMode: GetTransferMode(), connectionId: _connectionId);
+                await _transport.StartAsync(connectUrl, applicationSide, requestedTransferMode: GetTransferMode(), connectionId: _connectionId,  cancellationToken: cancellationToken);
```
BrennanConroy: nit: double space

moozzyk:
What state is the client left in if StartAsync is cancelled? Do we care? Do we need to clean up anything? 
Canceling in DisposeAsync may result in weird exceptions (especially for websockets)
Did you add tests to test cancellation at all?

mikaelm12:
Need to add tests still

> What state is the client left in if StartAsync is cancelled? Do we care? Do we need to clean up anything?
Good question. If the token is fired should you be able to try and start the connection again, or should we just dispose it?

> Canceling in DisposeAsync may result in weird exceptions (especially for websockets)
Yeah, still a little iffy about canceling in dispose. @davidfowl has an idea though..

test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -43,8 +43,9 @@ public class LongPollingTransportTests
                 {
                     var connectionToTransport = Channel.CreateUnbounded<SendMessage>();
                     var transportToConnection = Channel.CreateUnbounded<byte[]>();
+                    var tcs = new CancellationTokenSource();
                     var channelConnection = new ChannelConnection<SendMessage, byte[]>(connectionToTransport, transportToConnection);
-                    await longPollingTransport.StartAsync(new Uri("http://fakeuri.org"), channelConnection, TransferMode.Binary, connectionId: string.Empty);
+                    await longPollingTransport.StartAsync(new Uri("http://fakeuri.org"), channelConnection, TransferMode.Binary, connectionId: string.Empty, cancellationToken: tcs.Token);
```
BrennanConroy: `CancellationToken.None`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -100,13 +101,70 @@ public class HubConnectionTests
...
+        public async Task CancelStartAsync()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            var tcs = new CancellationTokenSource();
+            tcs.Cancel();
+            var startTask = hubConnection.StartAsync(tcs.Token);
+
+            await hubConnection.DisposeAsync();
```
BrennanConroy: Move this after the `Assert`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -100,13 +101,70 @@ public class HubConnectionTests
...
+        [Fact]
+        public async Task CancelStreamAsync()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            var tcs = new CancellationTokenSource();
+            tcs.Cancel();
+            await hubConnection.StartAsync();
+            var cancelledStreamTask = hubConnection.StreamAsync("test", typeof(string), new object[] { }, tcs.Token);
```
BrennanConroy: canceled

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -93,7 +93,11 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             using (var memoryStream = new MemoryStream())
             {
                 NegotiationProtocol.WriteMessage(new NegotiationMessage(_protocol.Name), memoryStream);
-                await _connection.SendAsync(memoryStream.ToArray(), _connectionActive.Token);
+                if (cancellationToken.CanBeCanceled)
+                {
+                    cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _connectionActive.Token).Token;
+                }
```
BrennanConroy: missing the else

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -125,19 +129,23 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
...
 
-        public async Task<ReadableChannel<object>> StreamAsync(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default(CancellationToken))
+        public async Task<ReadableChannel<object>> StreamAsync(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default)
         {
             return await StreamAsyncCore(methodName, returnType, args, cancellationToken).ForceAsync();
         }
 
         private async Task<ReadableChannel<object>> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken)
         {
+            if (cancellationToken.IsCancellationRequested)
```
BrennanConroy: Not sure if this is needed and if it is just use `cancellationToken.ThrowIfCancellationRequested()`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -147,11 +155,16 @@ public void On(string methodName, Type[] parameterTypes, Func<object[], Task> ha
             return await task;
         }
 
-        public async Task SendAsync(string methodName, object[] args, CancellationToken cancellationToken = default(CancellationToken)) =>
+        public async Task SendAsync(string methodName, object[] args, CancellationToken cancellationToken = default) =>
             await SendAsyncCore(methodName, args, cancellationToken).ForceAsync();
 
         private Task SendAsyncCore(string methodName, object[] args, CancellationToken cancellationToken)
         {
+            if (cancellationToken.IsCancellationRequested)
```
BrennanConroy: Same

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -54,13 +54,18 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
                 throw new ArgumentException("Invalid transfer mode.", nameof(requestedTransferMode));
             }
 
+            if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_transportCts.Token, cancellationToken).Token;
+            }
```
BrennanConroy: missing the else

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -50,10 +50,19 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
...
+            {
+                cancellationToken = _transportCts.Token;
+            }
+
             _logger.StartTransport(_connectionId, Mode.Value);
 
             // Start sending and polling (ask for binary if the server supports it)
-            _poller = Poll(url, _transportCts.Token);
+            _poller = Poll(url, cancellationToken);
             _sender = SendUtils.SendMessages(url, _application, _httpClient, _transportCts, _logger, _connectionId);
```
BrennanConroy: SendMessages should probably get the LinkedTcs

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -57,8 +57,17 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
...
+            if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_transportCts.Token, cancellationToken).Token;
+            }
+            else
+            {
+                cancellationToken = _transportCts.Token;
+            }
+
             var sendTask = SendUtils.SendMessages(url, _application, _httpClient, _transportCts, _logger, _connectionId);
```
BrennanConroy: Same

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -170,25 +179,26 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
             return await task;
         }
 
-        public async Task SendAsync(string methodName, object[] args, CancellationToken cancellationToken = default(CancellationToken)) =>
+        public async Task SendAsync(string methodName, object[] args, CancellationToken cancellationToken = default) =>
             await SendAsyncCore(methodName, args, cancellationToken).ForceAsync();
 
         private Task SendAsyncCore(string methodName, object[] args, CancellationToken cancellationToken)
         {
+            cancellationToken.ThrowIfCancellationRequested();
```
moozzyk: I think this is not good enough. In general this cancellation token is not very useful. The only thing it cancels is writing to the channel. The way to do this is to cancel the actual send from the transport. I think it is quite a big change and should be done separately.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -99,13 +100,70 @@ public class HubConnectionTests
...
 
         [Fact]
+        public async Task CancelStartAsync()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            var tcs = new CancellationTokenSource();
+            tcs.Cancel();
+            var startTask = hubConnection.StartAsync(tcs.Token);
```
moozzyk: You can create a canceled token like this : `new CancellationToken(true)`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -99,13 +100,70 @@ public class HubConnectionTests
         }
 
         [Fact]
+        public async Task CancelStartAsync()
```
moozzyk: These tests are a good start but we need some test that cancel a task after successfully starting the operation. 

Also we should ensure that the connection is in the expected stated after cancellation.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -99,13 +100,70 @@ public class HubConnectionTests
...
+
+        [Fact]
+        public async Task CancelInvokeAsync()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            var tcs = new CancellationTokenSource();
+            tcs.Cancel();
+            await hubConnection.StartAsync();
```
moozzyk: use .OrTimeout() for all async calls.

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
```diff
@@ -39,10 +40,12 @@ public WebSocketsTransportTests(ServerFixture<Startup> serverFixture, ITestOutpu
                 var connectionToTransport = Channel.CreateUnbounded<SendMessage>();
                 var transportToConnection = Channel.CreateUnbounded<byte[]>();
                 var channelConnection = new ChannelConnection<SendMessage, byte[]>(connectionToTransport, transportToConnection);
+                var tcs = new CancellationTokenSource();
+
 
                 var webSocketsTransport = new WebSocketsTransport(loggerFactory);
                 await webSocketsTransport.StartAsync(new Uri(_serverFixture.WebSocketsUrl + "/echo"), channelConnection,
-                    TransferMode.Binary, connectionId: string.Empty).OrTimeout();
+                    TransferMode.Binary, connectionId: string.Empty, cancellationToken: tcs.Token).OrTimeout();
```
moozzyk: Why not `default`?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -277,9 +277,9 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
                     var cts = new CancellationTokenSource();
```
moozzyk: I am missing tests at the HttpConnection and transport (if possible)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -66,10 +66,11 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             _connection.Closed += Shutdown;
         }
 
-        public async Task StartAsync() => await StartAsyncCore().ForceAsync();
+        public async Task StartAsync(CancellationToken cancellationToken = default) => await StartAsyncCore(cancellationToken).ForceAsync();
 
-        private async Task StartAsyncCore()
+        private async Task StartAsyncCore(CancellationToken cancellationToken)
         {
+            cancellationToken.ThrowIfCancellationRequested();
```
moozzyk: I am actually thinking this is to early. Currently, the idea is that once you start the connection you cannot start it again. With throwing an exception here this principle is actually broken because since if you catch the exception you can start it again...

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -239,6 +241,7 @@ private Task InvokeCore(string methodName, InvocationRequest irq, object[] args,
                 _logger.SendInvocationFailed(hubMessage.InvocationId, ex);
                 irq.Fail(ex);
                 TryRemoveInvocation(hubMessage.InvocationId, out _);
+                throw;
```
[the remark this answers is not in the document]
moozzyk: (btw. I looked at the code and I think there is a separate issue here where SendAsync will not throw even if the connection is not running. For this change however adding the `throw` here is incorrect)

src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
```diff
@@ -10,9 +10,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public interface IConnection
     {
-        Task StartAsync();
-        Task SendAsync(byte[] data, CancellationToken cancellationToken);
-        Task DisposeAsync();
+        Task StartAsync(CancellationToken cancellationToken = default);
```
moozzyk: Does the interface have to have cancellationTokens set to default? Because it shouldn't

src/Microsoft.AspNetCore.Sockets.Client.Http/ITransport.cs
```diff
@@ -2,15 +2,16 @@
...
 using System.Threading.Tasks.Channels;
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
     public interface ITransport
     {
-        Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId);
-        Task StopAsync();
+        Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId, CancellationToken cancellationToken = default);
+        Task StopAsync(CancellationToken cancellationToken = default);
```
moozzyk: remove `= default`

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -23,7 +23,7 @@ public class LongPollingTransport : ITransport
         private Task _poller;
         private string _connectionId;
 
-        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+        private CancellationTokenSource _transportCts = new CancellationTokenSource();
```
moozzyk: remove this completely - it is not being used.

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -19,7 +19,7 @@ public class ServerSentEventsTransport : ITransport
     {
         private readonly HttpClient _httpClient;
         private readonly ILogger _logger;
-        private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
+        private CancellationTokenSource _transportCts = new CancellationTokenSource();
```
moozzyk: remove

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -256,7 +265,7 @@ public WebSocketsTransport(ILoggerFactory loggerFactory)
 
                     // We intentionally don't pass _transportCts.Token to CloseOutputAsync. The token can be cancelled
                     // for reasons not related to webSocket in which case we would not close the websocket gracefully.
-                    await _webSocket.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, null, CancellationToken.None);
+                    await _webSocket.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, statusDescription: null, cancellationToken: cancellationToken);
```
moozzyk: I think passing the token to CloseOutputAsync is incorrect. if you do that you might leave the connection is undefined state - the transport will be possibly still connected (because closing was cancelled) but the connection will transition to the `Disconnected` state and there will be no way to ever close the websocket (if you called DisposeAsync again it will no op because it is in the `Disconnected` state). Also, see the comment above.

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
```
moozzyk: remove

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -99,13 +100,68 @@ public class HubConnectionTests
...
+        [Fact]
+        public async Task CancelSendAsync()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            var token = new CancellationToken(true);
+
+            await hubConnection.StartAsync().OrTimeout(); ;
+            var cancelledsendTask = hubConnection.SendAsync("testMethod", token);
```
moozzyk: nit: casing

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -99,13 +100,68 @@ public class HubConnectionTests
...
+
+            await Assert.ThrowsAsync<TaskCanceledException>(async () => await cancelledsendTask);
```
moozzyk: nit: can be just `await Assert.ThrowsAsync<TaskCanceledException>(() => cancelledsendTask);`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -99,13 +100,68 @@ public class HubConnectionTests
...
+
+        [Fact]
+        public async Task CancelSendAsync()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            var token = new CancellationToken(true);
+
+            await hubConnection.StartAsync().OrTimeout(); ;
```
moozzyk: remove unneded `;`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -391,7 +391,7 @@ private void SetTransferMode(TransferMode transferMode)
             _logger.EndReceive(_connectionId);
         }
 
-        public async Task SendAsync(byte[] data, CancellationToken cancellationToken = default(CancellationToken)) =>
+        public async Task SendAsync(byte[] data, CancellationToken cancellationToken = default) =>
```
moozzyk: Are we going to flow the cancellation token to the transport as well?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -277,9 +277,9 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
                     cts.Cancel();
 
-                    var channel = await connection.StreamAsync<int>("Stream", 5, cts.Token).OrTimeout();
+                    var canceledStreamTask = connection.StreamAsync<int>("Stream", 5, cts.Token).OrTimeout();
```
moozzyk: Exceptions and cancellations should go through the channel. The previous behavior was correct.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -165,6 +165,15 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
         }
 
         [Fact]
+        public async Task CheckCancelledStart()
+        {
+            var connection = new HttpConnection(new Uri("http://fakeuri.org/"));
+            var token = new CancellationToken(true);
+            var exception = await Assert.ThrowsAsync<TaskCanceledException>(
```
moozzyk: Why do you need to store the exception?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -165,6 +165,15 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+                async () => await connection.StartAsync(token));
```
moozzyk: Make sure you cannot start the connection after cancelling a previous start?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
moozzyk: Another way to look at it is that the behavior is not deterministic. If cancellation happens after this line your connection will not be startable but if it happens on this line it will. Removing this line makes it not startable always.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -99,13 +100,68 @@ public class HubConnectionTests
...
+        public async Task CancelInvokeAsync()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            var token = new CancellationToken(true);
+            await hubConnection.StartAsync().OrTimeout();
+            var invokeTask = hubConnection.InvokeAsync<int>("testMethod", token);
+
+            await Assert.ThrowsAsync<TaskCanceledException>(async () => await invokeTask);
```
moozzyk: You should make sure that the connection is still usable after this happens. A single invocation was cancelled but it should not break the connection.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
moozzyk: This still has been addressed - I see new tests but they all pass an already cancelled CancellationToken. In real life often times the user will be using a CancellationTokenSource and call methods with non-cancelled token that might be canceled after the method was called later.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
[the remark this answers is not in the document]
moozzyk: Here is the fix for the SendAsync issue: https://github.com/aspnet/SignalR/pull/1084

src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
mikaelm12: Yeah, that did feel a bit weird to me.

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
mikaelm12: We actually don't want to be able to Cancel StartAsync in LongPolling though, right? The issue you described with outstanding polls

mikaelm12: We'd still need the _transportCts for the poll and send  but I think we just shouldn't create the linked cts with the Token passed into StartAsync?

moozzyk: Yes. Longpolling would ignore the `CancellationToken` passed in `StartAsync`.

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
mikaelm12: We need this for the Cts

mikaelm12: But looks like we should just be passing in `CancellationToken.None`  instead

PR #1030, opened 2017-10-17
Not very important code optimization

Gurgen:
I made same optimization changes of HubGroupList to HubConnectionList.


moozzyk:
Thanks @Gurgen !

PR #1031, opened 2017-10-18
Adding roundtripping test for CancelInvocation message

moozzyk:


test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/MessagePackHubProtocolTests.cs
```diff
@@ -43,6 +43,8 @@ public class MessagePackHubProtocolTests
             new object[] { new[] { new StreamItemMessage("xyz", new CustomObject()) } },
             new object[] { new[] { new StreamItemMessage("xyz", new[] { new CustomObject(), new CustomObject() }) } },
 
+            new object[] { new[] { new CancelInvocationMessage("xyz ") } },
```
BrennanConroy: space in id needed?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
```diff
@@ -47,6 +47,11 @@ private bool InvocationMessagesEqual(HubMessage x, HubMessage y)
                 left.NonBlocking == right.NonBlocking;
         }
 
+        private bool CancelInvocationMessagesEqual(HubMessage x, HubMessage y)
+        {
+            return x is CancelInvocationMessage && y is CancelInvocationMessage;
```
BrennanConroy: Shouldn't this check the invocation ids are the same?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/MessagePackHubProtocolTests.cs
moozzyk: this is just another character in the invocation id... 

image

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
moozzyk: `invocationId` is common to `HubMessage` and is being checked in `Equals`.

PR #1037, opened 2017-10-19
Enable reconnect for Redis connection by default

BrennanConroy:
For now if you try to send when the redis connection is down you'll get something like:
```
Error: No connection is available to service this operation:
PUBLISH ChatSample.Hubs.Chat; SocketFailure on 127.0.0.1:6379/Subscription,
origin: Error, input-buffer: 0, outstanding: 0, last-read: 1s ago, last-write: 1s ago,
unanswered-write: 21077s ago, keep-alive: 60s, pending: 0, state: Connecting,
last-heartbeat: never, last-mbeat: -1s ago, global: 1s ago
```
and we won't keep a queue or anything around.

We'll probably want to change the error returned in the future.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -51,7 +51,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             _ackHandler = new AckHandler();
 
             var writer = new LoggerTextWriter(logger);
-            _logger.ConnectingToEndpoints(string.Join(", ", options.Value.Options.EndPoints.Select(e => EndPointCollection.ToString(e))));
+            _logger.ConnectingToEndpoints(options.Value.Options.EndPoints);
             _redisServerConnection = _options.Connect(writer);
 
             _redisServerConnection.ConnectionRestored += (_, e) =>
@@ -72,7 +72,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             }
             else
             {
-                _logger.LogWarning("Not connected to redis.");
+                _logger.NotConnected();
             }
             _bus = _redisServerConnection.GetSubscriber();
 
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis
         private readonly AckHandler _ackHandler;
         private int _internalId;
         private State _connectionState;
-        private readonly SemaphoreSlim _redisConnectionEventLock = new SemaphoreSlim(1, 1);
+        private readonly object _redisConnectionEventLock = new object();
 
         // This serializer is ONLY use to transmit the data through redis, it has no connection to the serializer used on each connection.
         private readonly JsonSerializer _serializer = new JsonSerializer
@@ -58,9 +58,8 @@ namespace Microsoft.AspNetCore.SignalR.Redis
 
             _redisServerConnection.ConnectionRestored += (_, e) =>
             {
-                try
+                lock(_redisConnectionEventLock)
                 {
-                    _redisConnectionEventLock.Wait();
                     if (_connectionState == State.Connected)
                     {
                         return;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -31,6 +31,8 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
         private readonly string _serverName = Guid.NewGuid().ToString();
         private readonly AckHandler _ackHandler;
         private int _internalId;
+        private State _connectionState;
+        private readonly SemaphoreSlim _redisConnectionEventLock = new SemaphoreSlim(1, 1);
```
davidfowl: Why aren't we using a normal lock?

davidfowl:
the publish failure can be handled in user code right?

BrennanConroy:
Yeah, for example in the Chat sample we do
```
connection.invoke('Send', data).catch(err => appendLine(err, 'red'));
```
and it prints out the error I listed above

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -51,16 +53,46 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
             _ackHandler = new AckHandler();
 
             var writer = new LoggerTextWriter(logger);
-            _logger.ConnectingToEndpoints(string.Join(", ", options.Value.Options.EndPoints.Select(e => EndPointCollection.ToString(e))));
+            _logger.ConnectingToEndpoints(options.Value.Options.EndPoints);
             _redisServerConnection = _options.Connect(writer);
+
+            _redisServerConnection.ConnectionRestored += (_, e) =>
```
moozzyk: Shouldn't these events be wired before you call Connect? What if the ConnectionRestore event happens immediately after calling Connect and before the event is wired up?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -51,16 +53,46 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+            {
+                lock(_redisConnectionEventLock)
+                {
+                    if (_connectionState == State.Connected)
```
moozzyk: Why do we need it? Do we get an event per Subscription?

BrennanConroy: https://github.com/SignalR/SignalR/pull/3849/files#diff-193a338ff4a86e8ec19aa6cfa2954768L242

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
BrennanConroy: This is the earliest we get the connection

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
moozzyk: In old SignalR we also have an Interactive connection in addition to pub sub and this was the [source of duplicate](https://github.com/SignalR/SignalR/issues/3729#issuecomment-261760333) ConnectionRestored/ConnectionClosed events.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
moozzyk: Ah, yes. Now I remember. I never liked it.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
moozzyk: Also, because you only log this you could log all events and just provide the connection type or anything else that can differentiate. Might be even better/truer

BrennanConroy: I don't think it would be useful to see "Connection restored for Interactive connection" and "Connection restored for Subscription connection"?

moozzyk: OK. (Maybe it is moot point anyways because AFAIR we don't have anyInteractive connection)

src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs
```diff
@@ -10,7 +10,11 @@ namespace Microsoft.AspNetCore.SignalR.Redis
 {
     public class RedisOptions
     {
-        public ConfigurationOptions Options { get; set; } = new ConfigurationOptions();
+        public ConfigurationOptions Options { get; set; } = new ConfigurationOptions()
```
moozzyk: nit: remove `()`

PR #1040, opened 2017-10-20
[Redis] Adding same group to connection multiple times should NOP

BrennanConroy:
Added some more tests too

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
@@ -352,13 +352,69 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Tests
 
                 var connection = new HubConnectionContext(output, client.Connection);
 
-                await manager1.OnConnectedAsync(connection);
+                await manager1.OnConnectedAsync(connection).OrTimeout();
 
-                await manager2.AddGroupAsync(connection.ConnectionId, "name");
+                await manager2.AddGroupAsync(connection.ConnectionId, "name").OrTimeout();
 
-                await manager2.InvokeGroupAsync("name", "Hello", new object[] { "World" });
+                await manager2.InvokeGroupAsync("name", "Hello", new object[] { "World" }).OrTimeout();
 
                 AssertMessage(output);
+            }
+        }
+
+        [Fact]
+        public async Task AddGroupAsyncForLocalConnectionAlreadyInGroupDoesNothing()
+        {
+            var manager = new RedisHubLifetimeManager<MyHub>(new LoggerFactory().CreateLogger<RedisHubLifetimeManager<MyHub>>(), Options.Create(new RedisOptions()
+            {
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -334,6 +334,153 @@ public class RedisHubLifetimeManagerTests
...
+
+                await manager.OnConnectedAsync(connection).OrTimeout();
+
+                await manager.AddGroupAsync(connection.ConnectionId, "name").OrTimeout();
+                await manager.AddGroupAsync(connection.ConnectionId, "name").OrTimeout();
+
+                await manager.InvokeGroupAsync("name", "Hello", new object[] { "World" }).OrTimeout();
+
+                AssertMessage(output);
+                Assert.False(output.In.TryRead(out var item));
```
moozzyk: Shuoldn't this check if only one SubscribeAsync was send?

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -334,6 +334,153 @@ public class RedisHubLifetimeManagerTests
...
+
+                await manager1.OnConnectedAsync(connection).OrTimeout();
+
+                await manager1.AddGroupAsync(connection.ConnectionId, "name").OrTimeout();
+                await manager2.AddGroupAsync(connection.ConnectionId, "name").OrTimeout();
+
+                await manager2.InvokeGroupAsync("name", "Hello", new object[] { "World" }).OrTimeout();
+
+                AssertMessage(output);
+                Assert.False(output.In.TryRead(out var item));
```
moozzyk: Would this test fail before your change? Why?

BrennanConroy: Yes, because the bus would subscribe to the subscription again so publish would call the callback twice. Not sure what the behavior of the actual `ConnectionMultiplexer` is for subscribing to the same channel twice, but it is better to exit early anyways to avoid extra works and taking locks

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
BrennanConroy: I could expose the subscriptions list in the `TestConnectionMultiplexer`. Figured I didn't need to since its a naive implementation and if subscribe was called twice then the callback would be called twice and this `Assert.False` would fail

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
moozzyk: I did not realize that it would not be de-duped anywhere and the callback will be called multiple times. I agree that shortcircuiting is the right thing to do.

PR #1046, opened 2017-10-23
Default

mikaelm12:
default(CancellationToken ...) -> default

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
@@ -25,7 +25,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         public bool TryGetSetting<T>(string name, out T value)
         {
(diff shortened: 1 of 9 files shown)


davidfowl:
Is this all the places?

mikaelm12:
There are a few more. I was only looking in the HubConnectionExtensions. I can change the others here too as well

davidfowl:
Do them all at once. All.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
```diff
@@ -25,7 +25,7 @@ public void AddSetting<T>(string name, T value)
 
         public bool TryGetSetting<T>(string name, out T value)
         {
-            value = default(T);
+            value = default;
```
muratg: Does this work for all `T`s that we care about?

mikaelm12: It should work just the same.

PR #1051, opened 2017-10-24
[Redis] Add functional test support

BrennanConroy:
Recreating, wouldn't let me reopen
Requires #1037 to be in before merging.

### DIFF
--- a/test/Common/ServerFixture.cs
+++ b/test/Common/ServerFixture.cs
@@ -17,33 +17,33 @@ namespace Microsoft.AspNetCore.SignalR.Tests.Common
     public class ServerFixture<TStartup> : IDisposable
         where TStartup : class
     {
-        private ILoggerFactory _loggerFactory;
-        private ILogger _logger;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly ILogger _logger;
         private IWebHost _host;
         private IApplicationLifetime _lifetime;
         private readonly IDisposable _logToken;
 
-        public string BaseUrl { get; private set; }
+        public string WebSocketsUrl => Url.Replace("http", "ws");
 
-        public string WebSocketsUrl => BaseUrl.Replace("http", "ws");
+        public string Url { get; private set; }
 
         public ServerFixture()
         {
             var testLog = AssemblyTestLog.ForAssembly(typeof(ServerFixture<TStartup>).Assembly);
-            _logToken = testLog.StartTestLog(null, $"{nameof(ServerFixture<TStartup>)}_{typeof(TStartup).Name}" , out _loggerFactory, "ServerFixture");
+            _logToken = testLog.StartTestLog(null, $"{nameof(ServerFixture<TStartup>)}_{typeof(TStartup).Name}", out _loggerFactory, "ServerFixture");
             _logger = _loggerFactory.CreateLogger<ServerFixture<TStartup>>();
-            BaseUrl = "http://localhost:" + GetNextPort();
+            Url = "http://localhost:" + GetNextPort();
 
-            StartServer();
+            StartServer(Url);
         }
 
-        private void StartServer()
+        private void StartServer(string url)
         {
             _host = new WebHostBuilder()
                 .ConfigureLogging(builder => builder.AddProvider(new ForwardingLoggerProvider(_loggerFactory)))
                 .UseStartup(typeof(TStartup))
                 .UseKestrel()
-                .UseUrls(BaseUrl)
+                .UseUrls(url)
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .Build();
 
--- a/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
@@ -1,0 +1,108 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Redis.Tests
+{
+    public class Docker
+    {
+        private static readonly string _exeSuffix = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ".exe" : string.Empty;
+
+        private static readonly string _dockerContainerName = "redisTestContainer";
+        private static Lazy<Docker> _instance = new Lazy<Docker>(Create);
+
+        public static Docker Default => _instance.Value;
+
+        private readonly string _path;
+
+        public Docker(string path)
+        {
+            _path = path;
+        }
+
+        private static Docker Create()
+        {
+            // Currently Windows Server 2016 doesn't support linux containers which redis is.
+            if (string.Equals("True", Environment.GetEnvironmentVariable("APPVEYOR"), StringComparison.OrdinalIgnoreCase))
+            {
+                return null;
+            }
+
+            var location = GetDockerLocation();
+            return location == null ? null : new Docker(location);
+        }
+
+        private static string GetDockerLocation()
+        {
+            foreach (var dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
+            {
+                var candidate = Path.Combine(dir, "docker" + _exeSuffix);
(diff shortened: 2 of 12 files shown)


test/Microsoft.AspNetCore.SignalR.Redis.Tests/EchoHub.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class EchoHub : Hub
+    {
+        public override Task OnConnectedAsync()
+        {
+            return Groups.AddAsync(Context.ConnectionId, "Test");
+        }
+
+        public override Task OnDisconnectedAsync(Exception exception)
+        {
+            return Groups.RemoveAsync(Context.ConnectionId, "Test");
```
moozzyk: Do you **need** to do this or you **want** to do this? A connection should be removed from a group automatically when it disconnects

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisServerFixture.cs
```diff
@@ -0,0 +1,49 @@
...
+        public ServerFixture<TStartup> FirstServer { get; private set; }
+        public ServerFixture<TStartup> SecondServer { get; private set; }
+
+        private readonly ILogger _logger;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly IDisposable _logToken;
+
+        public RedisServerFixture()
+        {
+            if (Docker.Default == null)
```
moozzyk: Add some comment that it means that Docker is not installed and the tests will be skipped.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisServerFixture.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                return;
+            }
+
+            var testLog = AssemblyTestLog.ForAssembly(typeof(RedisServerFixture<TStartup>).Assembly);
+            _logToken = testLog.StartTestLog(null, $"{nameof(RedisServerFixture<TStartup>)}_{typeof(TStartup).Name}", out _loggerFactory, "RedisServerFixture");
+            _logger = _loggerFactory.CreateLogger<RedisServerFixture<TStartup>>();
+
+            FirstServer = new ServerFixture<TStartup>();
+            SecondServer = new ServerFixture<TStartup>("http://localhost:3123");
```
moozzyk: why you specify url only for one of the servers. Also what happens if the url is taken?

moozzyk: ServerFixture starts the server in the ctor - you may want to add some exception handling/logging so that it is easy to figure out what just happened.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/SkipIfDockerNotPresentAttribute.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Testing.xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Redis.Tests
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
```
moozzyk: Is it possible to make this attribute work on a class as well - I think this would be more useful

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs
```diff
@@ -0,0 +1,129 @@
...
+        [MemberData(nameof(TransportTypesAndProtocolTypes))]
+        public async Task HubConnectionCanSendAndReceiveMessages(TransportType transportType, Type protocolType)
+        {
+            using (StartLog(out var loggerFactory, testName:
+                $"{nameof(HubConnectionCanSendAndReceiveMessages)}_{transportType.ToString()}_{protocolType.Name}"))
+            {
+                var connection = CreateConnection(_serverFixture.FirstServer.Url + "/echo", transportType, protocolType, loggerFactory);
+
+                await connection.StartAsync().OrTimeout();
+                var str = await connection.InvokeAsync<string>("Echo", "Hello world").OrTimeout();
```
moozzyk: `Hello, World!` :D

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        public static IEnumerable<object[]> TransportTypesAndProtocolTypes
+        {
+            get
+            {
+                foreach (var transport in TransportTypes)
+                {
+                    yield return new object[] { transport[0], typeof(JsonHubProtocol) };
```
moozzyk: create Protocol instances instead of passing types to be able to use just `.WithProtocol()` when creating the client.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+            else if (protocolType == typeof(JsonHubProtocol))
+            {
+                builder.WithJsonProtocol();
+            }
+
+            return builder.Build();
+        }
+
+        public static IEnumerable<object[]> TransportTypes
```
moozzyk: You don't use TransportTypes as member data so you can just return IEnumerable<TransportType>
Also these can probably just private static and expression bodied methods

test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
```diff
@@ -0,0 +1,107 @@
...
+        private string _path;
+        public Docker(string path)
+        {
+            _path = path;
+        }
+
+        private static Docker Create()
+        {
+            // Currently Windows Server 2016 doesn't support linux containers which redis is.
+            if (string.Equals("True", Environment.GetEnvironmentVariable("APPVEYOR")))
```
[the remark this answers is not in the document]
BrennanConroy: These tests should be able to run on Windows, as long as they have linux docker capabilities. Right now appveyor uses Windows Server 2016 which doesn't have that support yet.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/SkipIfDockerNotPresentAttribute.cs
BrennanConroy: Unfortunately it doesn't seem like xunit has support for that

test/Common/ServerFixture.cs
```diff
@@ -15,32 +15,39 @@ namespace Microsoft.AspNetCore.SignalR.Tests.Common
...
-        private ILoggerFactory _loggerFactory;
-        private ILogger _logger;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly ILogger _logger;
         private IWebHost _host;
         private IApplicationLifetime _lifetime;
         private readonly IDisposable _logToken;
 
-        public string BaseUrl => "http://localhost:3000";
+        public static string BaseUrl => "http://localhost:3000";
```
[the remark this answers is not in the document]
BrennanConroy: For normal use cases where xunit creates the serverfixture and uses the default constructor.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/EchoHub.cs
BrennanConroy: Good point, I should check that. Also I'm probably going to add a hub method that assigns yourself to a group instead of doing it in `OnConnectedAsync`

test/Common/ServerFixture.cs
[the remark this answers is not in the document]
moozzyk: Shouldn't this be `private static readonly` or even `private const`? I am not sure why this needs to be exposed - we now have/use `Url`

BrennanConroy:
Did you want more tests in this PR?
I was thinking we could get this in then we can add more later.

moozzyk:
No, not in this PR. I thought about merging this and adding more tests after this initial merge.

BrennanConroy:
Ship it?

moozzyk:
Yes. I think it's good to go.

PR #1056, opened 2017-10-25
Fixing a bug where cancellation could result in HubException

moozzyk:
When the client cancels a streaming method the server would send an error completion. This was not correct because cancellation is not an error. We did not see this because our client ignores any messages for a given streaming invocation after sending a CancelInvokationMessage but other clients may want to drain messages before considering a streaming method canceled.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -46,7 +46,7 @@ namespace Microsoft.AspNetCore.SignalR
                            IHubContext<THub> hubContext,
                            IOptions<HubOptions> hubOptions,
                            ILogger<HubEndPoint<THub>> logger,
-                           IServiceScopeFactory serviceScopeFactory, 
+                           IServiceScopeFactory serviceScopeFactory,
                            IUserIdProvider userIdProvider)
         {
             _protocolResolver = protocolResolver;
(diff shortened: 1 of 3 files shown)


moozzyk:
This was also a root cause of this CI failure: https://travis-ci.org/aspnet/SignalR/jobs/292370029 - the CancelInvocationMessage was sent by the client but by the time the client removed the callback from the list the server  send a completion which was an error completion caused by the cancellation.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -476,6 +477,16 @@ private bool IsChannel(Type type, out Type payloadType)
             }
             catch (Exception ex)
             {
+                if (ex is OperationCanceledException)
+                {
+                    // If a streaming method was cancelled we don't want to send a HubException message - this is not an error case
```
BrennanConroy: canceled 😆

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1024,6 +1023,32 @@ public void HubsCannotHaveOverloadedMethods()
...
+            {
+                var endPointLifetime = endPoint.OnConnectedAsync(client.Connection);
+
+                await client.Connected.OrTimeout();
+
+                var invocationId = await client.SendInvocationAsync(nameof(StreamingHub.BlockingStream)).OrTimeout();
+                // cancel the Streaming method
+                await client.SendHubMessageAsync(new CancelInvocationMessage(invocationId));
+
+                var hubMessage = Assert.IsType<StreamCompletionMessage>(await client.ReadAsync());
```
BrennanConroy: OrTimeout()!

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -476,6 +477,16 @@ private bool IsChannel(Type type, out Type payloadType)
             }
             catch (Exception ex)
             {
+                if (ex is OperationCanceledException)
+                {
+                    // If a streaming method was canceled we don't want to send a HubException message - this is not an error case
+                    if (connection.ActiveRequestCancellationSources.TryGetValue(invocationId, out var cts) && cts.IsCancellationRequested)
```
davidfowl: It would be cleaner to just store an error message object and have a single code path that creates the StreamingCompletion.

moozzyk:
:up::date:

Issue #1054, opened 2017-10-25
Exceptions thrown in Redis events should be caught and logged


moozzyk:
If an exception is thrown in a Redis event handler it will just disappear without a trace. E.g. if deserializing [a message throws](https://github.com/aspnet/SignalR/blob/30e7422407760d0d56361f76b73c5930875a03d3/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs#L123) no one will be able to tell what happened.

BrennanConroy:
Fixed via https://github.com/aspnet/SignalR/commit/2419867dfcda3b11a83a30de6559f1c44b3da38c

PR #1069, opened 2017-10-31
Handle errors in Redis subscription callbacks

BrennanConroy:
https://github.com/aspnet/SignalR/issues/1054

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
@@ -8,7 +8,6 @@ using System.Threading.Tasks.Channels;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.AspNetCore.SignalR.Tests;
 using Microsoft.AspNetCore.SignalR.Tests.Common;
-using Microsoft.AspNetCore.Sockets;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Moq;
--- a/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
@@ -1,13 +1,17 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Channels;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.AspNetCore.SignalR.Tests;
 using Microsoft.AspNetCore.SignalR.Tests.Common;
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/DefaultHubLifetimeManagerTests.cs
```diff
@@ -127,6 +130,24 @@ public class DefaultHubLifetimeManagerTests
...
+                // Force an exception when writing to connection
+                var output = new Mock<Channel<HubMessage>>();
+                output.Setup(o => o.Out.WaitToWriteAsync(It.IsAny<CancellationToken>())).Throws(new Exception());
+
+                var manager = new DefaultHubLifetimeManager<MyHub>();
+                var connection = new HubConnectionContext(output.Object, client.Connection);
+
+                await manager.OnConnectedAsync(connection);
+
+                await Assert.ThrowsAsync<Exception>(() => manager.InvokeConnectionAsync(connection.ConnectionId, "Hello", new object[] { "World" }));
```
moozzyk: Assert it is the same exception

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -481,6 +484,88 @@ public class RedisHubLifetimeManagerTests
...
+            {
+                // Force an exception when writing to connection
+                var output = new Mock<Channel<HubMessage>>();
+                output.Setup(o => o.Out.WaitToWriteAsync(It.IsAny<CancellationToken>())).Throws(new Exception());
+
+                var connection = new HubConnectionContext(output.Object, client.Connection);
+
+                await manager2.OnConnectedAsync(connection).OrTimeout();
+
+                await manager1.InvokeConnectionAsync(connection.ConnectionId, "Hello", new object[] { "World" }).OrTimeout();
```
moozzyk: add a comment explaining why that it does not fail because WaitToWriteAsync is not even invoked - rather we publish to  Redis.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -522,13 +524,14 @@ public class RedisHubLifetimeManagerTests
...
-                output.Setup(o => o.Out.WaitToWriteAsync(It.IsAny<CancellationToken>())).Throws(new Exception());
+                output.Setup(o => o.Out.WaitToWriteAsync(It.IsAny<CancellationToken>())).Throws(new Exception("Message"));
 
                 var connection = new HubConnectionContext(output.Object, client.Connection);
 
                 await manager.OnConnectedAsync(connection).OrTimeout();
 
-                await Assert.ThrowsAsync<Exception>(() => manager.InvokeConnectionAsync(connection.ConnectionId, "Hello", new object[] { "World" }).OrTimeout());
+                var exception = await Assert.ThrowsAsync<Exception>(() => manager.InvokeConnectionAsync(connection.ConnectionId, "Hello", new object[] { "World" }).OrTimeout());
+                Assert.Equal("Message", exception.Message);
```
moozzyk: I was thinking more about just storing the exception in a variable and then using `Assert.Same()` but this is fine too.

PR #1071, opened 2017-10-31
Pin package and tool versions to make build more repeatable

natemcmaster:
Part of https://github.com/aspnet/Universe/issues/575 and https://github.com/aspnet/KoreBuild/issues/239

This removes floating versions of dependencies. This includes the korebuild version and versions used by PackageReference's. This solves issues with being unable to build source code when the packages on our myget feeds change.


NuGet.config
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <clear />
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
```
ryanbrandenburg: What do we need this source for?

natemcmaster: Internal.AspNetCore.Sdk. We didn't need it with package lineup because we required 'build /t:Pin', which added a feed with this package. By removing this, we now need to go back to using the tools myget feed so a clean solution opens in VS w/o restore errors.

natemcmaster:
Heads up other reviewers, I'm planning to merge this later this afternoon. Take a look now if you want to add feedback before then.

moozzyk:
Feel free to merge this change.

PR #1072, opened 2017-10-31
[Redis] Don't wait for group ack when you're the only server

BrennanConroy:


moozzyk:
I am not a fan of this change. It makes the code more complex without almost any benefits. First, running one server with scaleout should almost never happen. Second, if it happens ACKs should be very fast.

davidfowl:
Agreed. We thought about doing this with the original signalr as well and decided against special code paths.

BrennanConroy:
@moozzyk and I talked about this yesterday, and I have a more generic solution that I wanted to run by you.

You send the group add/remove and get back the number of servers sent to, then each server sends an ACK back, and a special ACK if they did the action. Then we can check the amount of ACKs received vs. the number from publish and finish waiting for the ACK or if we get the ACK saying the action was completed we can finish waiting for the ACK.

moozzyk:
Do you even need a special ACK? You just wait for everyone to ACK without needing to know whether they removed the user or not (unless the idea is to ignore other ACKs if you received the one that did the action)

BrennanConroy:
Yeah, the idea was we could ignore the other ACKs.

davidfowl:
@BrennanConroy Are you going to update this change?

BrennanConroy:
If we're okay with making the change to ACKs then I can

davidfowl:
Lets's not right now. We can optimize later.

davidfowl:
File an issue as a suggestion to improve the redis impl.

PR #1073, opened 2017-10-31
Replacing StreamCompletion with StreamInvocation

### LINKED ISSUE #1053: When a streaming method on the server side cannot be resolved the client receives an invalid message
When a client invokes a streaming method but the method cannot be resolved the server returns a Completion message because it can't tell that the target method is supposed to be a streaming method. Since the client used `StreamAsync` to invoke this method it throws an exception because Streaming methods are supposed to be invoked with a StreamCompletion message and not Completion message.

moozzyk:
Fixes #1053

Previously if the client used `StreamAsync` method to invoke a non-streaming hub method or `InvokeAsync` to invoke a streaming hub method it was the client who rejected the call based on the completion message kind it received. This did not work in some cases (e.g. #1053) where if the server could not resolve the method it could not tell whether the method the client wanted to call was a streaming or non-streaming so it would send just a regular completion message. If the client invoked the method that could not be resolved by the server with StreamAsync, it would then throw an error saying that non-streaming methods should be invoked with InvokeAsync hiding the actual problem which is that the method should not be resolved. A better way of doing this is to have separate Invocation messages - one for streaming and one for non-streaming methods. This way the server has all information to decide if the invocation request matches the method and is able to reject non-matching invocations with an error without even invoking the method (as it happened previously).


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -551,7 +551,7 @@ public override void WriteLine(string value)
             }
         }
 
-        public class RedisExcludeClientsMessage : InvocationMessage
+        public class RedisExcludeClientsMessage : NonStreamInvocationMessage
```
BrennanConroy: Unrelated to your change, but I just noticed this, could you change this class to private

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -402,7 +402,7 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
             }
         }
 
-        [Theory(Skip="https://github.com/aspnet/SignalR/issues/1053")]
+        [Theory()]
```
BrennanConroy: Remove parens

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -487,6 +487,88 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                finally
+                {
+                    await connection.DisposeAsync().OrTimeout();
+                }
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(HubProtocolsAndTransportsAndHubPaths))]
+        public async Task ServerThrowsHubExceptionIfBuildingAsyncEnumeratorIsNotPossible(IHubProtocol hubProtocol, TransportType transportType, string hubPath)
```
BrennanConroy: Add same test but for method that has an invalid return type

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestHubMessageEqualityComparer.cs
```diff
@@ -22,7 +22,7 @@ public bool Equals(HubMessage x, HubMessage y)
             }
 
             return InvocationMessagesEqual(x, y) || StreamItemMessagesEqual(x, y) || CompletionMessagesEqual(x, y)
-                || StreamCompletionMessagesEqual(x, y) || CancelInvocationMessagesEqual(x, y);
+                || StreamInvocationMessagesEqual(x, y)|| CancelInvocationMessagesEqual(x, y);
```
BrennanConroy: Add space before `||`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
```diff
@@ -58,12 +58,32 @@ public InvocationMessage(string invocationId, bool nonBlocking, string target, E
             Target = target;
             _arguments = arguments;
             _argumentBindingException = argumentBindingException;
+        }
+    }
+
+    public class NonStreamInvocationMessage : InvocationMessage
```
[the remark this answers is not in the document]
moozzyk: I decided to keep it in one file since you are pretty much always interested in the base class (since it has all the functionality) and use derived classes which are basically ctors.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
moozzyk: This is actually impossible - the server will reject the method as a non-streaming method invoked with StreamingInvocation here: https://github.com/aspnet/SignalR/pull/1073/files/84ba804d183b33243365ebe9b25dca75e9a0584c#diff-3f9cc189cbcb6b225bb8d6cb6ecdeb21R496

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/TestBinder.cs
```diff
@@ -17,7 +17,10 @@ public TestBinder(HubMessage expectedMessage)
         {
             switch (expectedMessage)
             {
-                case InvocationMessage i:
+                case StreamInvocationMessage i:
+                    _paramTypes = i.Arguments?.Select(a => a?.GetType() ?? typeof(object))?.ToArray();
```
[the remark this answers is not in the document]
moozzyk: it cannot because of pattern matching

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
```diff
@@ -70,11 +67,8 @@ internal static class SignalRClientLoggerExtensions
         private static readonly Action<ILogger, string, Exception> _cancelingInvocationCompletion =
             LoggerMessage.Define<string>(LogLevel.Trace, new EventId(19, nameof(CancelingInvocationCompletion)), "Canceling dispatch of Completion message for Invocation {invocationId}. The invocation was canceled.");
 
-        private static readonly Action<ILogger, string, Exception> _receivedStreamCompletion =
-            LoggerMessage.Define<string>(LogLevel.Trace, new EventId(20, nameof(ReceivedStreamCompletion)), "Received StreamCompletion for Invocation {invocationId}.");
-
-        private static readonly Action<ILogger, string, Exception> _cancelingStreamCompletion =
-            LoggerMessage.Define<string>(LogLevel.Trace, new EventId(21, nameof(CancelingStreamCompletion)), "Canceling dispatch of StreamCompletion message for Invocation {invocationId}. The invocation was canceled.");
+        private static readonly Action<ILogger, string, Exception> _cancelingCompletion =
```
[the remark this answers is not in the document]
moozzyk: OK. I just appended to not change ones that we already had but since we have not shipped it should be fine to re-number them.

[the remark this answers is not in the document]
muratg: Yeah, re-numbering should be totally fine for now.

src/Microsoft.AspNetCore.SignalR.Client.Core/InvocationRequest.cs
```diff
@@ -78,26 +77,21 @@ public Streaming(CancellationToken cancellationToken, Type resultType, string in
...
+                Logger.InvocationCompleted(InvocationId);
+                if (completionMessage.Result != null)
                 {
-                    Logger.ReceivedUnexpectedMessageTypeForStreamCompletion(InvocationId, message.GetType().Name);
-                    // This is not 100% accurate but it is the only case that can be encountered today when running end-to-end
-                    // and this is the most useful message to show to the user.
-                    Fail(new InvalidOperationException($"Streaming hub methods must be invoked with the '{nameof(HubConnection)}.{nameof(HubConnection.StreamAsync)}' method."));
-                    return;
+                    Logger.ReceivedUnexpectedComplete(InvocationId);
+                    _channel.Out.TryComplete(new InvalidOperationException("Server provided a result in a completion response to a streamed invocation."));
```
muratg: I had to read this twice :)

moozzyk: Yeah - it's pretty dense but I just took what we already [had](https://github.com/aspnet/SignalR/commit/8d1d6d03005ea55c646850832de2e0eda53c6ade#diff-d05d3cff4a15053230f359d7e95afe3eL86) before

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/InvocationMessage.cs
davidfowl: I don't like this class name. Lets change the hierarchy a bit. I'd like InvocationMessage to be the regular invocation and we can use InvocationMessageBase as the base type.

moozzyk: I originally had exactly that I did not like it because we use the base class in many places and InvocationMessage is a very good fit for the invocation message as concept. When I created NonStreamInvocationMessage I hoped someone will recommend a better name - I don't like it at all but it makes it clear what the difference between different kinds of IncocationMessages is (as opposed to InvocationMessage derived from InvocationMessageBase)

davidfowl: I don't like it because InvocationMessage is the primary type. NonStreamingInvocation makes it seems like StreamingInvocation is the primary type.

moozzyk: I understand that but I am more in favor of changing `NonStreamingInvocation` to something that does not make it sound secondary. How about:
- changing NonStreamingInvocationMessage to HubMethodInvocationMessage?
or changing NonStreamingInvocationMessage to HubMethodInvocationMessage and StreamingInvocationMessage to HubMethodInvocation
or changing the base class from InvocationMessage to HubMethodInvocationMessage, NonStreamingInvocationMessage to InvocationMessage and leave StreamingInvocationMessage as is? 
(I think I like the 3rd option the best)

davidfowl: 3rd option is best

moozzyk: l'll do that then.

moozzyk:
:up::date:

moozzyk:
:watch:

PR #1074, opened 2017-11-01
Pawelka/msgpack5

### LINKED ISSUE #945: Msgpack: byte[] decoded as string
Hub:
```C#
public Task<byte[]> GetPayload()
    => Task.FromResult(new byte[] {0, 1, 2});
```

Client:
```TS
connection.invoke('GetPayload')
            .then((p: number[]) => console.log(typeof(p)))
```

Prints `string`, expected `object`

Version: alpha1

moozzyk:
Fixes #945

### DIFF
--- a/client-ts/package.json
+++ b/client-ts/package.json
@@ -42,6 +42,6 @@
     "vinyl-source-stream": "^1.1.0"
   },
   "dependencies": {
-    "msgpack5": "^3.5.0"
+    "msgpack5": "^3.6.0"
   }
 }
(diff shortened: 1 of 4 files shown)


client-ts/package.json
```diff
@@ -35,13 +35,13 @@
     "gulp-babel": "^7.0.0",
     "gulp-rename": "^1.2.2",
     "gulp-typescript": "^3.1.3",
-    "jasmine": "^2.5.2",
+    "jasmine": "^2.8.0",
```
moozzyk: @Eilon - I updated versions of the msgpack5 and jasmine modules. The first contains a fix for bug #945. The second is needed because it has support for verifying typed arrays.

Eilon: Ok so just these 2 got updated? (jasmine + msgpack5)

moozzyk: Yes - they were updated to newest versions

PR #1080, opened 2017-11-02
Closed Task

mikaelm12:
New and improved PR
Replaces the Closed event with a task.
Part of #900


samples/ClientSample/HubSample.cs
```diff
@@ -48,12 +48,13 @@ internal static void Register(CommandLineApplication app)
                 // Set up handler
                 connection.On<string>("Send", Console.WriteLine);
 
-                connection.Closed += e =>
+                _ = connection.Closed.ContinueWith((task, state) =>
```
davidfowl: Update the sample to use a completely different pattern. Instead of using the cancellation token you can now us the task.

```C#
while (!connection.Closed.IsCompleted)
{
     var completedTask = await Task.WhenAny(Task.Run(() => Console.ReadLine()), connection.Closed);
     if (completedTask == connection.Closed)
     {
         break;
     }

     var line = await (Task<string>)completedTask;

     if (line == null)
     {
         break;
     }

     await connection.InvokeAsync<object>("Send", line);
}
```

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -62,7 +61,7 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
             _connection.OnReceived((data, state) => ((HubConnection)state).OnDataReceivedAsync(data), this);
-            _connection.Closed += Shutdown;
+            _connection.Closed.ContinueWith((task) => Shutdown(task.Exception));
```
davidfowl: This is interesting. Users that subscribe to the Closed callback will add a continuation to connection.Closed task. I'm not sure continuations for tasks run sequentially. It might be better to expose a separate task on HubConnection that runs after Shutdown runs instead of doing it like this.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -46,8 +46,13 @@ public class HttpConnection : IConnection
         public Uri Url { get; }
 
         public IFeatureCollection Features { get; } = new FeatureCollection();
+        private readonly TaskCompletionSource<object> _closedTcs = new TaskCompletionSource<object>();
 
-        public event Func<Exception, Task> Closed;
+        public Task Closed
```
davidfowl: For more @pranavkm points do:

`Closed => _closedTcs.Task`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -116,6 +121,10 @@ private Task StartAsyncCore()
                     if (t.IsFaulted)
                     {
                         _startTcs.SetException(t.Exception.InnerException);
+                        if (_closedTcs.Task.IsFaulted)
```
davidfowl: Why is there new code here that wasn't here before?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -463,6 +461,7 @@ private void SetTransferMode(TransferMode transferMode)
                 await _receiveLoopTask;
             }
 
+            _closedTcs.TrySetResult(null);
```
davidfowl: help me understand why we have this new code?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -204,18 +204,11 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-            connection.Closed += e =>
-            {
-                closedEventTcs.SetResult(e);
-                return Task.CompletedTask;
-            };
 
             await connection.StartAsync();
             await connection.DisposeAsync();
-
+            await connection.Closed;
```
davidfowl: OrTimeout()

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -644,21 +619,7 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
                     return Task.CompletedTask;
                 }, receiveTcs);
 
-                connection.Closed += e =>
```
davidfowl: Don't we still need this functionality? Not sure we can just delete it.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -707,19 +668,6 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
                     return Task.CompletedTask;
                 });
 
-                connection.Closed += e =>
```
davidfowl: Same here.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -770,19 +718,6 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-                {
```
davidfowl: Same here

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -98,26 +92,26 @@ public class HubConnectionTests
...
-            var hubConnection = new HubConnection(mockConnection.Object, Mock.Of<IHubProtocol>(), new LoggerFactory());
-
-            await hubConnection.StartAsync();
-            var invokeTask = hubConnection.InvokeAsync<int>("testMethod");
-            await hubConnection.DisposeAsync();
-
-            var thrown = await Assert.ThrowsAsync(exception.GetType(), async () => await invokeTask);
-            Assert.Same(exception, thrown);
-        }
+        //[Fact]
```
davidfowl: Dead code? Why is this commented out?

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -29,8 +29,11 @@ internal class TestConnection : IConnection
         private Task _receiveLoop;
 
         private TransferMode? _transferMode;
-
-        public event Func<Exception, Task> Closed;
+        private readonly TaskCompletionSource<object> _closeTcs = new TaskCompletionSource<object>();
+        public Task Closed
```
davidfowl: See @pranavkm points comment.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -165,7 +165,6 @@ public EndToEndTests(ServerFixture<Startup> serverFixture, ITestOutputHelper out
                 try
                 {
                     var receiveTcs = new TaskCompletionSource<string>();
-                    var closeTcs = new TaskCompletionSource<object>();
```
davidfowl: we need to make sure this tests the right things

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -192,18 +201,7 @@ private Task StartAsyncCore()
...
-                        try
-                        {
-                            await closedEventHandler.Invoke(t.IsFaulted ? t.Exception.InnerException : null);
-                        }
-                        catch (Exception ex)
-                        {
-                            _logger.ExceptionThrownFromCallback(_connectionId, nameof(Closed), ex);
-                        }
-                    }
+                   _closedTcs.TrySetException(t.IsFaulted ? t.Exception.InnerException : null);
```
BrennanConroy: Should be checking if `t` is faulted, and using `TrySetResult` if it isn't

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -393,11 +380,6 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
             var connection = new HttpConnection(new Uri("http://fakeuri.org/"), new TestTransportFactory(mockTransport.Object), loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
             connection.OnReceived(_ => blockReceiveCallbackTcs.Task);
 
-            connection.Closed += _ => {
```
moozzyk: if closedTcs is not used remove completely

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
moozzyk: I think users should get this task and not the original from _connection

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: @mikaelm12 can you add a comment there. I'm also not sure why we're checking the state of the close task.

samples/ClientSample/HubSample.cs
mikaelm12: We should still use the Cts though to watch for the CancelKeyPress though, right?

davidfowl: Hmm, maybe we can call `DisposeAsync` in CancelKeyPress

moozzyk: The you would not dispose if the connection was closed in a different manner (e.g. by the server or due to a network error)

davidfowl: I don't see why not.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Since closed is now a task, we're making sure to complete it in dispose

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
mikaelm12: Added this back. Not being marked as resolved though

mikaelm12:
⬆️ 📅

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -116,10 +117,12 @@ private Task StartAsyncCore()
                     if (t.IsFaulted)
                     {
                         _startTcs.SetException(t.Exception.InnerException);
+                        _closedTcs.TrySetException(t.Exception.InnerException);
                     }
                     else if (t.IsCanceled)
                     {
                         _startTcs.SetCanceled();
+                        _closedTcs.SetCanceled();
```
moozzyk: if the task is faulted we use the `Try` but not here. Why? Which one should be used?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -204,18 +204,11 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-            connection.Closed += e =>
-            {
-                closedEventTcs.SetResult(e);
-                return Task.CompletedTask;
-            };
-
-            await connection.StartAsync();
-            await connection.DisposeAsync();
 
+            await connection.StartAsync().OrTimeout();
```
moozzyk: :+1: for adding missing `OrTimeout`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -644,21 +617,20 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-                            receiveTcs.TrySetCanceled();
-                        }
-                        return Task.CompletedTask;
-                    };
+                        receiveTcs.TrySetCanceled();
+                    }
+                    return Task.CompletedTask;
+                });
 
                 await connection.StartAsync();
```
moozzyk: `OrTimeout()`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -813,20 +785,13 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
-                    closeTcs.TrySetResult(e);
-                    return Task.CompletedTask;
-                };
 
-                await connection.StartAsync();
+                await connection.StartAsync().OrTimeout();
 
                 // Exception in send should shutdown the connection
-                await closeTcs.Task.OrTimeout();
+                var ex = await Assert.ThrowsAsync<HttpRequestException>(() => connection.Closed.OrTimeout());
```
moozzyk: remove `ex` - it is not used at all.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -174,15 +168,15 @@ public class HubConnectionTests
             await Assert.ThrowsAsync<TaskCanceledException>(async () => await invokeTask);
         }
 
-        [Fact]
+        [Fact(Skip = "Will un-skip this")]
```
moozzyk: When?

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -29,8 +29,8 @@ internal class TestConnection : IConnection
         private Task _receiveLoop;
 
         private TransferMode? _transferMode;
-
-        public event Func<Exception, Task> Closed;
+        private readonly TaskCompletionSource<object> _closeTcs = new TaskCompletionSource<object>();
+        public Task Closed => _closeTcs.Task;
```
moozzyk: nit: group public properties together.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -174,15 +168,15 @@ public class HubConnectionTests
...
         public async Task PendingInvocationsAreTerminatedWithExceptionWhenConnectionClosesDueToError()
         {
             var exception = new InvalidOperationException();
             var mockConnection = new Mock<IConnection>();
             mockConnection.SetupGet(p => p.Features).Returns(new FeatureCollection());
             mockConnection
                 .Setup(m => m.DisposeAsync())
-                .Callback(() => mockConnection.Raise(c => c.Closed += null, exception))
+                //.Callback(() => mockConnection.Raise(c => c.Closed += null, exception))
```
moozzyk: remove?

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -225,8 +208,6 @@ public EndToEndTests(ServerFixture<Startup> serverFixture, ITestOutputHelper out
                     logger.LogInformation("Receiving message");
                     Assert.Equal(message, await receiveTcs.Task.OrTimeout());
                     logger.LogInformation("Completed receive");
-
-                    await closeTcs.Task.OrTimeout();
```
moozzyk: Shouldn't we await the `.Closed` task here?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -116,10 +117,12 @@ private Task StartAsyncCore()
                     if (t.IsFaulted)
                     {
                         _startTcs.SetException(t.Exception.InnerException);
+                        _closedTcs.TrySetException(t.Exception.InnerException);
```
moozzyk: do we have a test for this condition and (possibly) the next condition (not sure if `t.IsCanceled` is possible to hit)?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -191,18 +194,13 @@ private Task StartAsyncCore()
                     _httpClient?.Dispose();
 
                     _logger.RaiseClosed(_connectionId);
```
moozzyk: Probably needs rephrasing - we no longer have Closed event.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -191,18 +194,13 @@ private Task StartAsyncCore()
...
-
-                    var closedEventHandler = Closed;
-                    if (closedEventHandler != null)
+                    if (t.IsFaulted)
```
moozzyk: Do we have a test for this?

samples/ClientSample/RawSample.cs
```diff
@@ -39,25 +39,19 @@ internal static void Register(CommandLineApplication app)
...
-                    cts.Cancel();
-                    return Task.CompletedTask;
-                };
 
                 await connection.StartAsync();
 
                 Console.WriteLine($"Connected to {baseUrl}");
-
-                Console.CancelKeyPress += (sender, a) =>
+                var cts = new CancellationTokenSource();
```
moozzyk: shouldn't this be canceled inside `CancelKeyPress`?

samples/ClientSample/HubSample.cs
```diff
@@ -37,30 +37,21 @@ internal static void Register(CommandLineApplication app)
...
                 {
                     a.Cancel = true;
                     Console.WriteLine("Stopping loops...");
-                    cts.Cancel();
+                    await connection.DisposeAsync();
                 };
 
                 // Set up handler
                 connection.On<string>("Send", Console.WriteLine);
+                var cts = new CancellationTokenSource();
```
moozzyk: Is this being used at all?

mikaelm12: It's being passed into InvokeAsync here https://github.com/aspnet/SignalR/pull/1080/files/9cc475fc9fbf74d2b8b6778f0e979c52f812e80c#diff-c044366a1200519c9bd2ab30a0243a34R66 but it also needs to be canceled in CancelKeyPress

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: I'm not actually completely sure here. I figure we should use Try though. @davidfowl ?

moozzyk: You need to check how we complete this tcs and if the order is deterministic (I think it isn't).

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -121,8 +121,8 @@ private Task StartAsyncCore()
                     }
                     else if (t.IsCanceled)
                     {
-                        _startTcs.SetCanceled();
-                        _closedTcs.SetCanceled();
+                        _startTcs.TrySetCanceled();
```
moozzyk: This should not have been changed.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
```diff
@@ -111,8 +111,8 @@ internal static class SocketClientLoggerExtensions
         private static readonly Action<ILogger, DateTime, string, Exception> _drainEvents =
             LoggerMessage.Define<DateTime, string>(LogLevel.Debug, new EventId(5, nameof(DrainEvents)), "{time}: Connection Id {connectionId}: Draining event queue.");
 
-        private static readonly Action<ILogger, DateTime, string, Exception> _raiseClosed =
-            LoggerMessage.Define<DateTime, string>(LogLevel.Debug, new EventId(6, nameof(RaiseClosed)), "{time}: Connection Id {connectionId}: Raising Closed event.");
+        private static readonly Action<ILogger, DateTime, string, Exception> _completeClosed =
+            LoggerMessage.Define<DateTime, string>(LogLevel.Debug, new EventId(6, nameof(CompleteClosed)), "{time}: Connection Id {connectionId}: Raising Closed event.");
```
moozzyk: The description needs to be updated too.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -185,8 +183,7 @@ public class HubConnectionTests
             var invokeTask = hubConnection.InvokeAsync<int>("testMethod");
             await hubConnection.DisposeAsync();
 
-            var thrown = await Assert.ThrowsAsync(exception.GetType(), async () => await invokeTask);
-            Assert.Same(exception, thrown);
+            await Assert.ThrowsAsync<InvalidOperationException>(async () => await invokeTask);
```
moozzyk: Not sure what this exception is.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Some of the EndToEnd tests hit this path with a faulted task. 
This is where we had the issue with the wrapped faulted task

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: Hmm

PR #1081, opened 2017-11-02
[Redis] Reorder some code to make it less spaghetti

BrennanConroy:
- Pulled out subscriptions to separate functions
- Moved `OnConnectedAsync` and `OnDisconnectedAsync` to be right after the constructor
- WriteAsync is now static


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -544,6 +395,186 @@ private T DeserializeMessage<T>(RedisValue data)
             }
         }
 
+        private void SubscribeToBroadcast()
```
davidfowl: SubscribeToHub

Issue #349, opened 2017-03-28
Server needs perf counters 


moozzyk:

DamianEdwards:
We can look at `EventCounter` again. @kichalla even wrote a prototype cmd line viewer for them.

muratg:
The goal is to design and implement the infrastructure in preview1 and implement a few of the perf counters as well.

We'll finish up the rest of the perf counters in later releases.

muratg:
cc @anurse

anurse:
Using this to track adding EventSource/EventCounter to SignalR. See https://gist.github.com/anurse/af1859663ac91c6cf69c820cebe92303 for some guidance on adding EventSources and EventCounters to ASP.NET projects.

BrennanConroy:
Compiling list of Counters/Events we may want:

**Microsoft-AspNetCore-Sockets EventSource**
- [x] ConnectionsStarted - Counts started connections
- [x] ConnectionsStopped - Counts connections that stopped gracefully
- [x] ConnectionsTimedOut - Counts connections that timed out
- [ ] ConnectionsAborted - Counts connections that aborted abnormally
- [x] ConnectionDuration (EventCounter)
- [ ] BytesRead - Counts bytes read from connections
- [ ] BytesWritten - Counts bytes written to connections
- [ ] Connection counters per transport?

**Microsoft-AspNetCore-SignalR EventSource**
- [ ] MessagesReceived - Counts all messages received
- [ ] MessagesSent - Counts all messages sent
- [ ] InvocationsReceived - Counts invocation messages received
- [ ] CompletionsReceived - Counts all completion messages received
- [ ] ErrorsReceived - Counts completion messages received containing an error
- [ ] StreamItemsReceived - Counts stream items messages received
- [ ] InvocationsSent - Counts invocation messages sent
- [ ] CompletionsSent - Counts completion messages sent
- [ ] ErrorsSent - Counts completion messages sent containing an error
- [ ] StreamItemsSent - Counts stream items messages sent

Error counters: We need to review a little bit about where they belong in our system

Scale-out counters? Redis messages sent, received, groups etc.

anurse:
> Scale-out counters? Redis messages sent, received, groups etc.

Yep, sounds good :).

BrennanConroy:
Redis messages sent and received would probably be unnecessary since it would be counted in Microsoft-AspNetCore-SignalR

anurse:
[...]

PR #1083, opened 2017-11-02
Adding EventSource/EventCounters to Sockets

BrennanConroy:
Part of #349 
First time Eventing, lets see how it goes

@anurse

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
+++ b/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
@@ -58,6 +58,7 @@ namespace Microsoft.AspNetCore.Sockets
             var id = MakeNewConnectionId();
 
             _logger.CreatedNewConnection(id);
+            var connectionTimer = SocketEventSource.Log.ConnectionStart(id);
 
             var transportToApplication = Channel.CreateUnbounded<byte[]>();
             var applicationToTransport = Channel.CreateUnbounded<byte[]>();
@@ -66,6 +67,7 @@ namespace Microsoft.AspNetCore.Sockets
             var applicationSide = ChannelConnection.Create<byte[]>(transportToApplication, applicationToTransport);
 
             var connection = new DefaultConnectionContext(id, applicationSide, transportSide);
+            connection.ConnectionTimer = connectionTimer;
 
             _connections.TryAdd(id, connection);
             return connection;
@@ -73,9 +75,10 @@ namespace Microsoft.AspNetCore.Sockets
 
         public void RemoveConnection(string id)
         {
-            if (_connections.TryRemove(id, out _))
+            if (_connections.TryRemove(id, out var connection))
             {
                 // Remove the connection completely
+                SocketEventSource.Log.ConnectionStop(id, connection.ConnectionTimer);
                 _logger.RemovedConnection(id);
             }
         }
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Sockets/Internal/SocketEventSource.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Sockets.Internal
+{
+    [EventSource(Name = "Microsoft-AspNetCore-Sockets")]
+    internal class SocketEventSource : EventSource
+    {
+        public static readonly SocketEventSource Log = new SocketEventSource();
+
+        private readonly EventCounter _connectionsStartedCounter;
```
BrennanConroy: Counter is probably redundant

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -135,6 +138,7 @@ public void Scan()
                     // Once the decision has been made to dispose we don't check the status again
                     if (status == DefaultConnectionContext.ConnectionStatus.Inactive && (DateTimeOffset.UtcNow - lastSeenUtc).TotalSeconds > 5)
                     {
+                        SocketEventSource.Log.ConnectionTimedOut(c.Value.ConnectionId);
```
davidfowl: Where is the log for this?

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -66,16 +66,19 @@ public DefaultConnectionContext CreateConnection()
             var applicationSide = ChannelConnection.Create<byte[]>(transportToApplication, applicationToTransport);
 
             var connection = new DefaultConnectionContext(id, applicationSide, transportSide);
+            var connectionTimer = SocketEventSource.Log.ConnectionStart(id);
```
davidfowl: Where is the ILogger.Log for this?

BrennanConroy: On line 60, I'll move this to be next to it.

src/Microsoft.AspNetCore.Sockets/DefaultConnectionContext.cs
```diff
@@ -22,6 +23,7 @@ public class DefaultConnectionContext : ConnectionContext,
         // This tcs exists so that multiple calls to DisposeAsync all wait asynchronously
         // on the same task
         private TaskCompletionSource<object> _disposeTcs = new TaskCompletionSource<object>();
+        internal ValueStopwatch _connectionTimer;
```
anurse: Use a property. We generally avoid setting fields from outside the class. Keep it `internal` though.

src/Microsoft.AspNetCore.Sockets/Internal/SocketEventSource.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                if (IsEnabled(EventLevel.Informational, EventKeywords.None))
+                {
+                    ConnectionStop(connectionId);
+                }
+            }
+        }
+
+        [Event(eventId: 1, Level = EventLevel.Informational, Message = "Started connection '{0}'.")]
+        public ValueStopwatch ConnectionStart(string connectionId)
```
anurse: Any other values we could add to the payload? We don't know the transport yet, correct? Maybe include the request's TraceIdentifier if we have it?

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
anurse: Yeah, we aren't doing the "bundle the event source call in the logger extension method" thing, but in general, it should be abnormal to write an event and **not** write a log message. So if there isn't a log message for this, add one as well :).

src/Microsoft.AspNetCore.Sockets/Internal/SocketEventSource.cs
BrennanConroy: We don't have any info at this point

PR #1084, opened 2017-11-03
Actually throwing exceptions from SendAsync

moozzyk:
SendAsync was using InvokeCoreAsync code to send messages. In case of exception InvokeCoreAsync is blocking and returns a task to the user so they can await for the remote call to complete. Any exception thrown is caught and used to fail the task returned to the user. SendAsync does not return a special task to the user so re-using InvokeCore resulted in swallowing exceptions. While SendAsync is fire and forget it actually should throw if the message could not be send and it was not happening.

While adding tests it turned out we did not test cases where Invoke/SendAsync/StreamAsync were invoked before starting the connection and this resulted in a NullReferenceException. I also fixed that.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -154,7 +154,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         {
             if (!_startCalled)
             {
-                throw new InvalidOperationException($"The '{nameof(StreamAsync)}' method cannot be call before the connection has been started.");
+                throw new InvalidOperationException($"The '{nameof(StreamAsync)}' method cannot be called before the connection has been started.");
             }
 
             var invokeCts = new CancellationTokenSource();
@@ -196,7 +196,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         {
             if (!_startCalled)
             {
-                throw new InvalidOperationException($"The '{nameof(InvokeAsync)}' method cannot be call before the connection has been started.");
+                throw new InvalidOperationException($"The '{nameof(InvokeAsync)}' method cannot be called before the connection has been started.");
             }
 
             var irq = InvocationRequest.Invoke(cancellationToken, returnType, GetNextId(), _loggerFactory, this, out var task);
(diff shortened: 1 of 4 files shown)


davidfowl:
We should change the blocking non blocking terminology. It's extremely confusing 😄  (@anurse)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -55,6 +55,19 @@ public class HubConnectionTests
         }
 
         [Fact]
+        public async Task SendAsyncThrowsIfSerializingMessageFails()
```
davidfowl: Is this the case that would have swallowed the exception before?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -141,13 +152,18 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
 
         private async Task<ReadableChannel<object>> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken)
         {
+            if (!_startCalled)
+            {
+                throw new InvalidOperationException($"The '{nameof(StreamAsync)}' method cannot be call before the connection has been started.");
```
BrennanConroy: `cannot be **called**`

anurse:
> We should change the blocking non blocking terminology. It's extremely confusing

Yeah, it definitely is. Open for suggestions :)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
moozzyk: Yes this was one of the cases. `SendAsync` wouldn't throw at all. Connection could be closed or not started and it wouldn't throw either.

moozzyk:
`nonblocking` will be addressed separately from this PR (it will also be a bigger change, including changing the protocol). Some ideas for the name:
- `blocking`
- `waitForResult`
- `fireAndForget`

anurse:
The first two are good. I think `waitForResult` is clearest. `fireAndForget` is a colloquialism, so I'd avoid it.

Issue #932, opened 2017-09-23
Allow direct connect to the HubEndPoint without negotiation


davidfowl:
We should be able to expose a model where everything is fully specified via configuration on both server and client that would avoid any negotiation and would just directly connect using the specified transport and protocol.

This includes:
- [x] - Direct websocket connection without http negotiation for a connection id (#1035)
- [ ] - Specific protocol selection which avoids hub protocol negotiation.

Related to #854  

/cc @anurse @moozzyk

anurse:
To write down my thoughts from earlier in-person discussions:

I think a good way to represent this, architecturally, is to change our Hub Protocol abstraction to be a little more stateful and move negotiation into that layer. We'd build a `NegotiatedHubProtocol` which contains a list of Hub Protocols to negotiate. That component handles the first negotiation frame and then just proxies all data through to the appropriate Hub Protocol. Then, every Hub endpoint has one and only one instance of Hub Protocol associated with it (a `NegotiatedHubProtocol` containing the JSON and MsgPack Hub Protocols, by default), and users can reconfigure as needed. I think this would be better than a first-class "disable negotiation" feature.

Having said all that, we should consider making a simpler configuration model available (extension methods or something). A TCP endpoint that only wanted to speak MsgPack would simply change the Hub Protocol in the HubOptions from it's default value to `MsgPackHubProtocol`.

moozzyk:
We decided not to do the second part. Everything can work fine if things are aligned but if anything is misconfigured - and getting to this state is quite easy - it's a pit of failure because there is not a good way to detect this state and close the connection with a clear message. See comments in https://github.com/aspnet/SignalR/pull/1095 for more details.

PR #1095, opened 2017-11-07
Enabling starting connection without protocol negotiation

moozzyk:
#932

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
@@ -61,6 +61,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
             return hubConnectionBuilder.WithLoggerFactory(loggerFactory);
         }
 
+        public static IHubConnectionBuilder WithNoProtocolNegotiation(this IHubConnectionBuilder hubConnectionBuilder)
+        {
+            hubConnectionBuilder.AddSetting(HubConnectionBuilderDefaults.SkipNegotiateKey, true);
+            return hubConnectionBuilder;
+        }
+
         public static ILoggerFactory GetLoggerFactory(this IHubConnectionBuilder hubConnectionBuilder)
         {
             hubConnectionBuilder.TryGetSetting<ILoggerFactory>(HubConnectionBuilderDefaults.LoggerFactoryKey, out var loggerFactory);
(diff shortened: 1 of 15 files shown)


moozzyk:
This has questionable behavior in case of a client/server configuration mismatch - in some cases the server will just ignore the Negotiation message but in other cases it will fail with weird errors.

davidfowl:
I'm not a fan of the "skip negotiate flag on the client" (aka `WithNoProtocolNegotiation`). Picking a protocol on the server is enough to make it use that specific protocol without negotiation but it isn't enough on the client. 

We should clarify a few things:
- When a client is configured with a specific transport an a specific protocol, why does it need to negotiation (is it because of the text vs binary decision?).
- When the server specifies a specific protocol, what sort of error does the client get if the client doesnt' send a negotiate and connects with the wrong protocol?

PS: It should be WithoutProtocolNegotiation 😄 .

moozzyk:
> I'm not a fan of the "skip negotiate flag on the client"

We are on the same page. I initially tried adding a `skipNegotiate` parameter to `WithHubProtocol()` that the user could set but it is basically the same things only the number of overloads exploded. 
The main reason to have this flag is that if the server is not configured to use one protocol you have to send negotiate always to the server otherwise it will not know what protocol to use to handle your connection. 

If the server receives a wrong protocol it will not be able to parse the message and will throw `FormatException` which results in closing the connection. 

> It should be WithoutProtocolNegotiation

A very good point.

moozzyk:
So, actually when the server is using MsgPack and sends JSon the server will very likely not close the connection. It might read the first x bytes as a VarInt and then if the message is too short it will throw the contents. This needs more thought because we don't seem to have enough information on the server side to tell the protocols apart just by message contents (and this does not seem to be correct as well).

moozzyk:
Maybe could pass missing information in the transport-specific way. With websockets we could send it as a subprotocol, for SSE and longpolling we could use a header? HubConnection would set a feature on IConnection which the client transport would read and act accordingly. On the server side the server would read the value and set the feature when creating a connection. This feature could be then consumed by HubEndPoint....

davidfowl:
Or maybe this feature is just a bust. We can revisit later.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -72,5 +78,10 @@ public static IHubProtocol GetHubProtocol(this IHubConnectionBuilder hubConnecti
             hubConnectionBuilder.TryGetSetting<IHubProtocol>(HubConnectionBuilderDefaults.HubProtocolKey, out var hubProtocol);
             return hubProtocol;
         }
+
+        public static bool GetNoProtocolNegotiation(this IHubConnectionBuilder hubConnectionBuilder)
```
mikaelm12: I think this reads better as `GetWithoutProtocolNegotiation`

moozzyk:
This is a pit of failure. It is very easy to misconfigure things which may result in a range of issues - from abruptly closing the connection to connections that appear to be connected but throwing all the messages on the server side.

Issue #1018, opened 2017-10-13
Specify our pattern for JWT bearer auth. Write a sample at minimum and explore the API.


muratg:

Havret:
It would be very nice to have an opinionated approach. I was googling for a while to find out that the only way to pass JWT bearer is through qs.

moozzyk:
https://github.com/aspnet/SignalR/issues/1039

moozzyk:
0bafb304c2707739578e635bc9e00d0cd36420a7..4394b57143ad0517333ffbe88848fbb68264a614

PR #1101, opened 2017-11-10
Adding support for JWT in the C# client

moozzyk:
Fixes: #1018

(Bonus: also enabling passing headers)

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
@@ -27,16 +27,22 @@ export interface ITransport {
 
 export class WebSocketTransport implements ITransport {
     private readonly logger: ILogger;
+    private readonly jwtBearer: () => string;
     private webSocket: WebSocket;
 
-    constructor(logger: ILogger) {
+    constructor(jwtBearer: () => string, logger: ILogger) {
         this.logger = logger;
+        this.jwtBearer = jwtBearer;
     }
 
     connect(url: string, requestedTransferMode: TransferMode): Promise<TransferMode> {
 
         return new Promise<TransferMode>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
+            if (this.jwtBearer) {
+                let token = this.jwtBearer();
+                url += (url.indexOf("?") < 0 ? "?" : "&") + `signalRTokenHeader=${token}`;
+            }
 
             let webSocket = new WebSocket(url);
             if (requestedTransferMode == TransferMode.Binary) {
@@ -96,12 +102,14 @@ export class WebSocketTransport implements ITransport {
 
 export class ServerSentEventsTransport implements ITransport {
     private readonly httpClient: IHttpClient;
+    private readonly jwtBearer: () => string;
     private readonly logger: ILogger;
     private eventSource: EventSource;
     private url: string;
 
-    constructor(httpClient: IHttpClient, logger: ILogger) {
+    constructor(httpClient: IHttpClient, jwtBearer: () => string, logger: ILogger) {
         this.httpClient = httpClient;
+        this.jwtBearer = jwtBearer;
         this.logger = logger;
     }
 
@@ -109,10 +117,15 @@ export class ServerSentEventsTransport implements ITransport {
         if (typeof (EventSource) === "undefined") {
             Promise.reject("EventSource not supported by the browser.");
         }
-        this.url = url;
 
+        this.url = url;
         return new Promise<TransferMode>((resolve, reject) => {
-            let eventSource = new EventSource(this.url);
+            if (this.jwtBearer) {
+                let token = this.jwtBearer();
+                url += (url.indexOf("?") < 0 ? "?" : "&") + `signalRTokenHeader=${token}`;
+            }
+
+            let eventSource = new EventSource(url);
 
             try {
                 eventSource.onmessage = (e: MessageEvent) => {
@@ -152,7 +165,7 @@ export class ServerSentEventsTransport implements ITransport {
     }
 
     async send(data: any): Promise<void> {
-        return send(this.httpClient, this.url, data);
+        return send(this.httpClient, this.url, this.jwtBearer, data);
     }
 
     stop(): void {
@@ -168,14 +181,16 @@ export class ServerSentEventsTransport implements ITransport {
 
 export class LongPollingTransport implements ITransport {
     private readonly httpClient: IHttpClient;
+    private readonly jwtBearer: () => string;
     private readonly logger: ILogger;
 
     private url: string;
     private pollXhr: XMLHttpRequest;
     private shouldPoll: boolean;
 
-    constructor(httpClient: IHttpClient, logger: ILogger) {
+    constructor(httpClient: IHttpClient, jwtBearer: () => string, logger: ILogger) {
         this.httpClient = httpClient;
+        this.jwtBearer = jwtBearer;
         this.logger = logger;
     }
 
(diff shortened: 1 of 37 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/HttpOptions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+
+namespace Microsoft.AspNetCore.Sockets.Client.Http
+{
+    public class HttpOptions
+    {
+        public HttpMessageHandler HttpMessageHandler { get; set; }
+        public IReadOnlyCollection<KeyValuePair<string, string>> Headers { get; set; }
+        public Func<string> JwtBearerTokenFactory { get; set; }
```
davidfowl: This is too specific

moozzyk: In what sense it is too specific?

davidfowl: I'm not sold on having anything beyond headers on this API.

moozzyk: I started with that. However JWT token may change while the connection is open so the user will have to update the headers collection. How can they do that in a thread safe manner? Do you want to expose headers as ConcurrentDictionary?

moozzyk: Also, if this is treated like all other headers it will be on the user to format it correctly...

moozzyk:
Added also the first iteration for TS client. It's based on the same premise as the one for C# - the user provides a factory. We can iterate on it if we decide we want something different. Also, note that the client will create the Authorization header if possible but where it is not possible it will send the token as a url. The disadvantage is that the user needs to set the token manually. We should/could explore if this can be done for the user by SignalR when the user sets an option when mapping the endpoint.

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
```diff
@@ -107,5 +107,22 @@ internal static class SendUtils
...
+                foreach (var header in httpOptions.Headers)
+                {
+                    request.Headers.Add(header.Key, header.Value);
+                }
+            }
+            request.Headers.UserAgent.Add(Constants.UserAgentHeader);
+
+            if (httpOptions?.JwtBearerTokenFactory != null)
+            {
+                request.Headers.Add("Authorization", $"Bearer {httpOptions?.JwtBearerTokenFactory()}");
```
BrennanConroy: null check already happened

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -28,12 +29,25 @@ public class WebSocketsTransport : ITransport
...
+            if (httpOptions?.Headers != null)
+            {
+                foreach (var header in httpOptions.Headers)
+                {
+                    _webSocket.Options.SetRequestHeader(header.Key, header.Value);
+                }
+            }
+            if (httpOptions?.JwtBearerTokenFactory != null)
+            {
+                _webSocket.Options.SetRequestHeader("Authorization", $"Bearer {httpOptions?.JwtBearerTokenFactory()}");
```
BrennanConroy: null check not needed

moozzyk:
@Tratcher @davidfowl - added samples for C# and JS client that where the token is being refreshed pre-emptively. Let me know if I am completely off the rails...

client-ts/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
```diff
@@ -32,6 +79,24 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             app.UseSockets(options => options.MapEndPoint<EchoEndPoint>("echo"));
             app.UseSignalR(options => options.MapHub<TestHub>("testhub"));
             app.UseSignalR(options => options.MapHub<UncreatableHub>("uncreatable"));
+            app.UseSignalR(options => options.MapHub<HubWithAuthorization>("authorizedhub"));
+
+            app.Run(async (context) =>
```
Tratcher: You want this to be Use rather than Run, or else your app will never get 404s.

client-ts/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
```diff
@@ -18,6 +28,43 @@ public void ConfigureServices(IServiceCollection services)
                 // consistent casing makes it cleaner to verify results
                 options.JsonSerializerSettings.ContractResolver = new DefaultContractResolver();
             });
+
+            services.AddAuthorization(options =>
+            {
+                options.AddPolicy(JwtBearerDefaults.AuthenticationScheme, policy =>
```
Tratcher: This policy isn't being used anywhere. The Authorize attribute is referencing the authentication handler, not the policy.

samples/JwtSample/wwwroot/index.html
```diff
@@ -0,0 +1,104 @@
...
+        return get(tokenUrl)
+            .then(function (token) {
+                tokens[id] = token;
+            });
+    }
+
+    function runConnection(id, transportType) {
+        var connection;
+
+        refreshToken(id)
```
Tratcher: You're refreshing before every connection? How often is that likely to be? These tokens are meant to last at least an hour.

moozzyk: I need a token to start a connection. Then I refresh the token [every 20 secs](https://github.com/aspnet/SignalR/pull/1101/files/41cde6fc455d2ea87591ac1567ca4769368c0f2f#diff-2de970c6f616cd91f56bcb443e2e8367R84) because the server  creates tokens that [expire after 30 secs](https://github.com/aspnet/SignalR/pull/1101/files/41cde6fc455d2ea87591ac1567ca4769368c0f2f#diff-4179aa1ab03cf92976d053318cf3f436R79). 
I know that tokens generally live longer but this sample is supposed to show that the functionality being added is enough to update the token before it expires and hence prevent from SignalR connection failures caused by auth.

moozzyk: I think I need to change id to userId to make it cleaner - I open 3 connections.

client-ts/Microsoft.AspNetCore.SignalR.Test.Server/Startup.cs
```diff
@@ -18,6 +28,43 @@ public void ConfigureServices(IServiceCollection services)
...
+                    options.Events = new JwtBearerEvents
+                    {
+                        OnMessageReceived = context =>
+                        {
+                            var signalRTokenHeader = context.Request.Query["X-SignalR-Token"];
+
+                            if (!string.IsNullOrEmpty(signalRTokenHeader) &&
+                                (context.HttpContext.WebSockets.IsWebSocketRequest || context.Request.Headers["Accept"] == "text/event-stream"))
+                            {
+                                context.Token = context.Request.Query["X-SignalR-Token"];
```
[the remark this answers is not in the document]
moozzyk: Yeah - I am open to changing this name to something better.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -276,6 +294,12 @@ export class LongPollingTransport implements ITransport {
 
 const headers = new Map<string, string>();
```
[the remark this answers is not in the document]
moozzyk: + remove from send - it's an optional parmeter.

PR #1104, opened 2017-11-11
Hubconnection Disposables

MarkOMeara123:
Added ILoggerFactory and CancellationTokenSource to DisposeAsync
However ILoggerFactory should not be disposed if Hubconnectionbuilder was not used to instantiate Hubconnection.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -126,6 +126,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private async Task DisposeAsyncCore()
         {
             await _connection.DisposeAsync();
+            _connectionActive?.Dispose();
+            _loggerFactory?.Dispose();
         }
 
         // TODO: Client return values/tasks?


MarkOMeara123:
Perhaps a cleaner solution to ILoggerFactory dispose is to mode the Dispose method to HubConnectionBuilder instead. HubConnectionBuilder shall need to implement IDisposable interface and users shall need to be careful about when they dispose it.

moozzyk:
The test failure is legitimate - did you run build.cmd locally?
`HubConnectionBuilder` is only a convenience class that does not have to be used to create hub connections and as such can't be used to Dispose anything.

MarkOMeara123:
No did not run build.cmd, until now. It states : "Build Succeeded: 0 Warnings, 0 Errors". Perhaps I should update this PR now that I have run build.cmd?

As for where to put the ILoggerFactory dispose the case for disposing it in HubConnection is where it has been instanciated via dependency inject or via HubConnectionBuilder. Otherwise it should not be disposed (i.e. a manual instantiation of HubConnection passing through an ILoggerFactory instance that is used elswhere in code). The solution proposed here takes care of the former and will cause issues with the  latter.

anurse:
Hey @MarkOMeara123 , apologies that this seems to have been left behind a bit. I think we should take the CTS Dispose change, but I'm much less confident in disposing the `ILoggerFactory`. The general rule-of-thumb is to not dispose things that are provided to you in the constructor, so I don't think we should do that. As you suggest, we could look at doing that in `HubConnectionBuilder`, but again, only in the case where we create the `LoggerFactory` on your behalf. If a logger factory is passed in we should not dispose it.

If you want to rebase this and make some updates, I'd be happy to take a look at it. If you don't have time, that's OK too, we can close this and track it in an issue on our backlog.

davidfowl:
We need to revisit this design. This is problematic but @anurse 's point stands, we need to know if we created the logger factory in order to properly dispose it.

PR #1116, opened 2017-11-14
Fix couple flaky tests

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -98,6 +98,7 @@ public class HubConnectionTests
 
             await hubConnection.StartAsync();
             await hubConnection.DisposeAsync();
+            await hubConnection.Closed.OrTimeout();
```
moozzyk: this seems wrong - hubConnection has been disposed so the expectation that the `Closed` task is valid is not correct. The bug is somewhere else...

BrennanConroy: Yeah I wasn't sure about that. We could await the closed task in DisposeAsync?

moozzyk: Yes, it has to be something to this effect

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -334,7 +334,7 @@ private Task InvokeStreamCore(string methodName, InvocationRequest irq, object[]
             }
         }
 
-        private Task Shutdown(Exception ex = null)
+        private void Shutdown(Exception ex = null)
```
moozzyk: :+1:

PR #1119, opened 2017-11-14
First pass at Keep Alive

anurse:
**Important**: Right now this does **not** include the logic for automatically sending keep alive message. This is just the protocol changes necessary to support them. However, we can discuss how we're going to do keep alive here since it isn't totally worked out yet :).

**Open Question**: Where is keep alive really needed? In looking at the behavior of the transports, I'd think only SSE actually needs it, and only on the Server->Client channel. WebSockets has a pretty solid keep alive system. Long Polling is designed to support cases where the poll times out, and keeping it alive would actually require more network traffic. Having said that, SignalR needs to consider cases like TCP as well. On a raw TCP connection, you'll need full duplex Ping/Pong. One thought I had is to loosen the requirements I wrote in the spec even further and allow endpoints to ignore `Ping` frames if they want to. Then WebSockets and Long Polling can completely ignore it and SSE can send `Ping` frames (but the client can ignore them and not send `Pong` responses).

Big items:
* Defined `Ping` and `Pong` messages in the spec (used WebSockets as an inspiration here)
* Added support for `Ping`/`Pong` in JSON and MsgPack protocols in both C# and TypeScript code.
* Added support for sending a `Pong` in response to a `Ping` to all clients and server.
* Tests for all of the above

Small changes:
* Refactored some things in server/client hub protocol implementations to reduce duplication and risk of drift between client and server

I think we'll need a brief meeting to sync up on the needs for keep alive. But the bulk of this code will be useful for whatever model we build.


BrennanConroy:
SSE definitely needs it behind IIS because it times out the connection after ~100 seconds of inactivity .
I think I've seen cases with WebSockets where it becomes disconnected but doesn't react until you try to send through it.

moozzyk:
WS and SSE need keep alives. Long Polling should not have keep alives. 
Looking at the PR - I am not sure why we need ping/pong and if we need them whether they should be the same as keep alives. My thinking about keep alives was that we should make them as small as possible and as simple as possible. I thought about encoding them as `{}` in json and `0x01 0x91 0x06` in MessagePack. The only purpose of keep alives is to make sure a message can be sent/received. On top of that we could have a ServiceMessage if it turns out it is needed.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/PongMessage.cs
```diff
@@ -0,0 +1,12 @@
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
+{
+    public class PongMessage:HubMessage
```
davidfowl: I don't miss your awful formatting @anurse

davidfowl:
I agree with @moozzyk . We need keep alives for all transports. The basic idea is that we'll use these to avoid proxies killing the connection *and* to detect silent client disconnects like we did before. I'm not sure we need ping/pong, we just need server -> client messages.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/PongMessage.cs
anurse: I can happily stop writing code for you if you'd prefer ;)

anurse:
Sure, that's the gist of my comments in the PR description. I'm not sure I follow why we need it for WebSockets? In my experience, those do terminate quickly when the network drops (but I'm going to do some testing)

anurse:
Removing the payload is trivial, and I mostly agree, it seems unnecessary. I'm not even exactly sure why WebSockets needs it in theirs, but I left it in while I tried to research that. Came up with nothing and I know SignalR 2.x didn't have content, so I'm totally find removing that.

moozzyk:
@davidfowl - why do we need keep alives for long polling? If we send keep alives from server we will make the transport even less efficient because the client will have to potentially restart the poll request much often. We tried to add keep alives to SignalR 2.1.x and use chunked responses and it was a disaster - we removed the feature in the very next version.

anurse:
Agreed, we should only need it for WebSockets and SSE. We'll also need something in TCP/UDP if we ever decide to implement that. I do think we need the option for bi-directional keep alive (i.e. Client sending `Ping` to Server) for the same reasons (the client wants to know that the server is still there).

anurse:
> My thinking about keep alives was that we should make them as small as possible and as simple as possible. I thought about encoding them as {} in json and `0x02 0x91 0x06` in MessagePack.

I'm all for keeping them small but I think `{}` is overkill. Given that these messages will likely only be sent if the connection has been idle for a defined period of time, they'll probably end up in a single TCP frame. Size still matters somewhat there but if they're likely to be in their own frame, they don't need to be quite as tiny as two characters :). I'd prefer `{"type":6}` (where `6` is the type for `Ping`) in JSON. Your MsgPack suggestion is basically this already anyway :).

davidfowl:
Sure it's not needed for long polling since it already has timeouts.

anurse:
:up: :date: Removed payloads and the Pong message.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/MessagePackHubProtocol.spec.ts
```diff
@@ -97,4 +97,18 @@ describe("MessageHubProtocol", () => {
             } as CompletionMessage
         ]);
     });
+
+    it("can read ping message", () => {
+        let payload = [
+            0x02,
+            0x91, // message array length = 2 (fixarray)
```
moozzyk: lenght = 1

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/MessagePackHubProtocol.ts
```diff
@@ -30,15 +30,27 @@ export class MessagePackHubProtocol implements IHubProtocol {
...
+            case MessageType.Ping:
+                return this.createPingMessage(properties);
             default:
                 throw new Error("Invalid message type.");
         }
     }
 
+    private createPingMessage(properties: any[]): HubMessage {
+        if (properties.length != 1) {
+            throw new Error("Invalid payload for Ping/Pong message.");
```
moozzyk: nit: just Ping

specs/HubProtocol.md
```diff
@@ -97,7 +98,18 @@ If either endpoint commits a Protocol Error (see examples below), the other endp
 * It is a protocol error for a Caller to send a `Completion` message carrying both a result and an error.
 * It is a protocol error for an `Invocation` or `StreamInvocation` message to have an `Invocation ID` that has already been used by *that* endpoint. However, it is **not an error** for one endpoint to use an `Invocation ID` that was previously used by the other endpoint (allowing each endpoint to track it's own IDs).
 
-## Examples
+## Ping (aka "Keep Alive")
+
+The SignalR Hub protocol supports "Keep Alive" messages used to ensure that the underlying transport connection remains active. These messages help ensure:
+
+1. Proxies don't close the underlying connection during idle times (when few messages are being sent)
+1. If the underlying connection is dropped without being terminated gracefully, the application is informed as quickly as possible.
```
moozzyk: nit: 2.

specs/HubProtocol.md
```diff
@@ -97,7 +98,18 @@ If either endpoint commits a Protocol Error (see examples below), the other endp
...
+
+Keep alive behavior is achieved via the `Ping` message type. **Either endpoint** may send a `Ping` message at any time. The receiving endpoint is expected to simply ignore the message. No response is expected.
```
moozzyk: the receiving endpoint may actually update internal timeouts etc.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -391,5 +390,34 @@ public class HubConnectionProtocolTests
...
+        [Fact]
+        public async Task AcceptsPingMessages()
+        {
+            var connection = new TestConnection(TransferMode.Text);
+            var hubConnection = new HubConnection(connection,
+                new JsonHubProtocol(), new LoggerFactory());
+
+            try
+            {
+                // Receive the ping mid-invocation so we can see that the rest of the flow works fine
```
moozzyk: nit: move next to ping?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -391,5 +390,34 @@ public class HubConnectionProtocolTests
...
+                await hubConnection.StartAsync().OrTimeout();
+
+                // Ignore negotiate message
+                await connection.ReadSentTextMessageAsync().OrTimeout();
+
+                // Receive a ping
+                await connection.ReceiveJsonMessage(new { type = 6 }).OrTimeout();
+
+                // Send an invocation
+                var invokeTask = hubConnection.InvokeAsync("Foo");
```
moozzyk: should this call be awaited?

anurse: Yes, in fact it's in the wrong order. I'll fix that, not sure how it got messed up like this.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/PingMessage.cs
```diff
@@ -0,0 +1,11 @@
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
```
BrennanConroy: copyright

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -303,6 +303,10 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                                         }
                                         break;
 
+                                    case PingMessage pingMessage:
```
BrennanConroy: nit: _

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
moozzyk: Yeah, this test looked like it was not finished...

anurse: It was finished, and then I did some refactoring and it got unfinished :P

specs/HubProtocol.md
anurse: FYI, Markdown doesn't actually require that, it regenerates the numbers. I generally prefer to use `1.` so that it's not order-sensitive. Regardless, it's a trivial change and I'm happy to make it.

PR #1120, opened 2017-11-14
Fixing js end-to-end test after introducing streaminvocation

moozzyk:


moozzyk:
Also fixing (in the 2nd commit):
- 2 tests did not stop the connection and were causing hangs in the browser
- adding a parameter that allows using non-minified version of scripts to run tests

client-ts/Microsoft.AspNetCore.SignalR.Test.Server/wwwroot/connectionTests.html
```diff
@@ -18,9 +30,10 @@
         }
         else
         {
+            var minified = getParameterByName('debug') !== 'true' ? '.min' : '';
```
BrennanConroy: Shouldn't this be applied to ES5 too

moozzyk: OK (didn't do it because you would have to debug in IE which we typically don't do but it does not cost much and you may hit an IE specific issue...)

Issue #1111, opened 2017-11-13
Change negotiate from OPTIONS to use a "normal" HTTP method


anurse:
We've seen a few issues (#1107 for example) with using `OPTIONS` requests for connection negotiation. As a result, we should just switch to a "normal" HTTP method (like `POST` or `PUT`) and a URL endpoint

BrennanConroy:
https://github.com/aspnet/SignalR/commit/93cbf4dbef003b5101064978e954fed83095121e

PR #1122, opened 2017-11-15
Change negotiate to POST

BrennanConroy:
https://github.com/aspnet/SignalR/issues/1111

Another option would be to make `ExecuteAsync` take a flag and do all the logic as before with the `/path/negotiate` route passing in a "Negotiate" flag

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
@@ -212,7 +212,7 @@ describe("Connection", () => {
 
         it("preserves user query with / before query", async done =>
         {
-            connection = new HttpConnection("http://tempuri.org/endpoint/?q=myData", options);
+            connection = new HttpConnection("http://tempuri.org/endpoint/?q=my/Data", options);
 
             try {
                 await connection.start();
@@ -223,11 +223,11 @@ describe("Connection", () => {
                 done();
             }
 
-            expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate?q=myData");
+            expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate?q=my/Data");
         });
 
         it("preserves user query without / before query", async done => {
-            connection = new HttpConnection("http://tempuri.org/endpoint?q=myData", options);
+            connection = new HttpConnection("http://tempuri.org/endpoint?q=my/Data", options);
 
             try {
                 await connection.start();
@@ -238,7 +238,7 @@ describe("Connection", () => {
                 done();
             }
 
-            expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate?q=myData");
+            expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate?q=my/Data");
         });
 
         it("adds negotiate after path without /", async done => {
@@ -270,35 +270,6 @@ describe("Connection", () => {
 
             expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate");
         });
-    });
-
-    it("preserves users connection string in negotiate request", async done => {
-        let options: IHttpConnectionOptions = {
-            httpClient: <IHttpClient>{
-                post(url: string): Promise<string> {
-                    expect(url).toBe("http://tempuri.org/negotiate?q=myData");
-                    connection.stop();
-                    return Promise.resolve("{}");
-                },
-                get(url: string): Promise<string> {
-                    connection.stop();
-                    return Promise.resolve("");
-                }
-            },
-            logging: null
-        } as IHttpConnectionOptions;
-
-
-        let connection = new HttpConnection("http://tempuri.org?q=myData", options);
-
-        try {
-            await connection.start();
-            done();
-        }
-        catch (e) {
-            fail();
-            done();
-        }
     });
 
     eachTransport((requestedTransport: TransportType) => {
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
@@ -191,6 +191,116 @@ describe("Connection", () => {
         done();
     });
 
+    describe("negotiate request", () =>
+    {
+        let negotiateUrl: string;
+        let connection: HttpConnection;
+        let options: IHttpConnectionOptions = {
+            httpClient: <IHttpClient>{
+                post(url: string): Promise<string> {
+                    negotiateUrl = url;
+                    connection.stop();
+                    return Promise.resolve("{}");
+                },
+                get(url: string): Promise<string> {
+                    connection.stop();
+                    return Promise.resolve("");
+                }
+            },
+            logging: null
+        } as IHttpConnectionOptions;
+
+        it("preserves user query with / before query", async done =>
+        {
+            connection = new HttpConnection("http://tempuri.org/endpoint/?q=myData", options);
+
+            try {
+                await connection.start();
+                done();
+            }
+            catch (e) {
+                fail();
+                done();
+            }
+
+            expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate?q=myData");
+        });
+
+        it("preserves user query without / before query", async done => {
+            connection = new HttpConnection("http://tempuri.org/endpoint?q=myData", options);
+
+            try {
+                await connection.start();
+                done();
+            }
+            catch (e) {
+                fail();
+                done();
+            }
+
+            expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate?q=myData");
+        });
+
+        it("adds negotiate after path without /", async done => {
+            connection = new HttpConnection("http://tempuri.org/endpoint", options);
+
+            try {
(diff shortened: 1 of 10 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpConnection.ts
```diff
@@ -189,6 +189,17 @@ export class HttpConnection implements IConnection {
...
+        let index = url.indexOf("?");
+        let negotiateUrl = this.url.substring(0, index == -1 ? url.length : index);
+        if (negotiateUrl[negotiateUrl.length - 1] !== '/') {
+            negotiateUrl += '/';
+        }
+        negotiateUrl += "negotiate";
+        negotiateUrl += index == -1 ? "" : url.substring(index);
+        return negotiateUrl;
+    }
+
```
muratg: Should we have tests for this?

BrennanConroy: Probably

davidfowl:
Or a PUT

BrennanConroy:
> The difference between PUT and POST is that PUT is idempotent: calling it once or several times successively has the same effect (that is no side effect), where successive identical POST may have additional effects, like passing an order several times.

Negotiate sounds like POST

BrennanConroy:
@anurse convinced me that `/connection` or `/connections` would be better than `/negotiate`

anurse:
Reasoning: You're not creating a "negotiate", you're creating a "connection".

davidfowl:
Why would we change the name..... negotiate is what it's doing, it's familiar to what we had before, if we're going to change the URL, negotiate seems like what we'd pick. Seems like a change for change sake.

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpConnection.ts
anurse: Unfortunate that we have to do this :(. I like it better than splitting the query string off though. Looks fine, as long as it's well tested.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -63,6 +58,31 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
             }
         }
 
+        public async Task ExecuteNegotiateAsync(HttpContext context, HttpSocketOptions options)
+        {
+            // Create the log scope and attempt to pass the Connection ID to it so as many logs as possible contain
+            // the Connection ID metadata. If this is the negotiate request then the Connection ID for the scope will
```
anurse: This comment is a little different in this method since we know it definitely **is** the negotiate response :)

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HttpConnection.ts
```diff
@@ -189,6 +189,17 @@ export class HttpConnection implements IConnection {
         return normalizedUrl;
     }
 
+    private resolveNegotiateUrl(url: string): string {
+        let index = url.indexOf("?");
+        let negotiateUrl = this.url.substring(0, index == -1 ? url.length : index);
```
moozzyk: `===` ?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -220,7 +220,14 @@ private Task StartAsyncCore()
...
                 logger.EstablishingConnection(url);
-                using (var request = new HttpRequestMessage(HttpMethod.Options, url))
+                var urlBuilder = new UriBuilder(url);
+                if (!urlBuilder.Path.EndsWith("/"))
+                {
+                    urlBuilder.Path += "/";
+                }
+                urlBuilder.Path += "negotiate";
+
+                using (var request = new HttpRequestMessage(HttpMethod.Post, urlBuilder.Uri))
```
moozzyk: does this work fine with queryString? I think we now need a test where the url contains a queryString.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -926,5 +929,11 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
             Assert.NotNull(transferModeFeature);
             Assert.Equal(TransferMode.Binary, transferModeFeature.TransferMode);
         }
+
+        private bool IsNegotiateRequest(HttpRequestMessage request)
+        {
+            return request.Method == HttpMethod.Post &&
+                request.RequestUri.ToString().Contains("/negotiate");
```
moozzyk: Shouldn't be something like `Path.EndsWith("/negotiate")` instead of `Contains`?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
BrennanConroy: We definitely need tests now. I did manually verify that queryString is kept intact

BrennanConroy:
:up: :date:

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
```diff
@@ -191,6 +191,87 @@ describe("Connection", () => {
         done();
     });
 
+    describe("negotiate request", () =>
+    {
```
moozzyk: nit: should be on the previous line

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
```diff
@@ -191,6 +191,87 @@ describe("Connection", () => {
...
+            catch (e) {
+                fail();
+                done();
+            }
+
+            expect(negotiateUrl).toBe("http://tempuri.org/endpoint/negotiate");
+        });
+
+        it("adds negotiate after path with /", async done => {
+            connection = new HttpConnection("http://tempuri.org/endpoint/", options);
```
moozzyk: do we need a case for an url without trailing '/' ? (e.g. `http://tempuri.org/endpoint`)?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
```diff
@@ -191,6 +191,87 @@ describe("Connection", () => {
...
+                },
+                get(url: string): Promise<string> {
+                    connection.stop();
+                    return Promise.resolve("");
+                }
+            },
+            logging: null
+        } as IHttpConnectionOptions;
+
+        it("preserves user query with / before query", async done =>
```
moozzyk: make these tests theory-like? They are all the same except the input and output...

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -926,5 +929,36 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+                .Returns<HttpRequestMessage, CancellationToken>(async (request, cancellationToken) =>
+                {
+                    await Task.Yield();
+                    Assert.Equal(expectedNegotiate, request.RequestUri.ToString());
+                    return ResponseUtils.CreateResponse(HttpStatusCode.OK,
+                        ResponseUtils.CreateNegotiationResponse());
+                });
+
+            var connection = new HttpConnection(new Uri(requested), TransportType.LongPolling, loggerFactory: null, httpMessageHandler: mockHttpHandler.Object);
+            await connection.StartAsync();
```
moozzyk: `.OrTimeout()` (and below)

src/Microsoft.AspNetCore.Sockets.Http/SocketRouteBuilder.cs
```diff
@@ -28,6 +28,7 @@ public void MapSocket(string path, HttpSocketOptions options, Action<ISocketBuil
             socketConfig(socketBuilder);
             var socket = socketBuilder.Build();
             _routes.MapRoute(path, c => _dispatcher.ExecuteAsync(c, options, socket));
+            _routes.MapRoute(path + "/negotiate", c => _dispatcher.ExecuteNegotiateAsync(c, options));
```
davidfowl: Hmmm

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/Connection.spec.ts
BrennanConroy: Thats the test directly above

src/Microsoft.AspNetCore.Sockets.Http/SocketRouteBuilder.cs
BrennanConroy: What are your thoughts

anurse: The `HttpConnectionDispatcher` doesn't know the path the user asked to map the socket too, so this is the cleanest way to do this. I actually like it better anyway, the dispatcher shouldn't have to be doing routing things.

PR #1127, opened 2017-11-15
Switching to new base64 APIs

moozzyk:


src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -14,15 +16,22 @@ public byte[] Decode(byte[] payload)
             var buffer = new ReadOnlyMemory<byte>(payload);
             LengthPrefixedTextMessageParser.TryParseMessage(ref buffer, out var message);
 
-            return Convert.FromBase64String(Encoding.UTF8.GetString(message.ToArray()));
+            Span<byte> decoded = new byte[Base64.GetMaxDecodedFromUtf8Length(message.Length)];
+            var status = Base64.DecodeFromUtf8(message.Span, decoded, out _, out var written);
+            Debug.Assert(status == OperationStatus.Done);
+
+            return decoded.Slice(0, written).ToArray();
```
ahsonkhan: Is it necessary to copy the span (after decoded) into an array? What if the payload was `Span<byte>` instead of byte[] and the function returned a `Span<byte>`?

It looks like the only place Decode is called is here:
https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs#L24

You could change the TryParseMessage to take in a ReadOnlySpan<byte> instead of ReadOnlyMemory<byte>.
i.e. Change [this](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/IHubProtocol.cs#L16) to:
```C#
bool TryParseMessages(ReadOnlySpan<byte> input, IInvocationBinder binder, out IList<HubMessage> messages);
```

It looks like you get the span out of it anyway:
https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/TextMessageParser.cs#L14

Let me know if that would work or if I am missing something in how IHubProtocol and the TryParseMessage(s) methods are used.

cc @KrzysztofCwalina

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -14,15 +16,22 @@ public byte[] Decode(byte[] payload)
...
         {
-            var buffer = Encoding.UTF8.GetBytes(Convert.ToBase64String(payload));
+            Span<byte> buffer = new byte[Base64.GetMaxEncodedToUtf8Length(payload.Length)];
+
+            var status = Base64.EncodeToUtf8(payload, buffer, out _, out var written);
+            Debug.Assert(status == OperationStatus.Done);
             using (var stream = new MemoryStream())
             {
-                LengthPrefixedTextMessageWriter.WriteMessage(buffer, stream);
+                LengthPrefixedTextMessageWriter.WriteMessage(buffer.Slice(0, written), stream);
```
BrennanConroy: Probably not part of this PR, but we should look at removing the need for stream

moozzyk: This actually should be part of this PR. Fixing.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
moozzyk: I did not want this change to explode. I am trying it now though.

moozzyk:
:up:date:

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
moozzyk: Decided to do it separately...

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
KrzysztofCwalina: I agree with @ahsonkhan. The whole point of these APIs is to avoid allocating and copying data (as it happens in the call to ToArray).

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -2,27 +2,37 @@
...
-            var buffer = Encoding.UTF8.GetBytes(Convert.ToBase64String(payload));
+            Span<byte> buffer = new byte[Base64.GetMaxEncodedToUtf8Length(payload.Length)];
+
+            var status = Base64.EncodeToUtf8(payload, buffer, out _, out var written);
+            Debug.Assert(status == OperationStatus.Done);
+
             using (var stream = new MemoryStream())
             {
-                LengthPrefixedTextMessageWriter.WriteMessage(buffer, stream);
+                LengthPrefixedTextMessageWriter.WriteMessage(buffer.Slice(0, written), stream);
```
ahsonkhan: Similar to Decode, can we just return the `Span<byte> buffer`?

moozzyk: I don't think this gets us anything - the result is eventually passed to a method the expects `byte[]`. Also, this encoder is used only in very esoteric cases which are already known to be slow (e.g. require base64 encoding). The main encoder is the [pass through encoder](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/PassThroughEncoder.cs#L13-L16) that is no op if it takes and returns the byte[].

The long term plan is to take a comprehensive look at the stack to see how we can avoid copying because now we do it like crazy all over the place. This was postponed for now until we have a functionality with good tests in place and we switch to pipelines.

ahsonkhan:
Nice! 2x improvement :)

``` ini

BenchmarkDotNet=v0.10.9, OS=Windows 10 Redstone 2 (10.0.15063)
Processor=Intel Xeon CPU E5-2673 v4 2.30GHz, ProcessorCount=4
.NET Core SDK=2.1.0-preview1-007042
  [Host]     : .NET Core 2.0.3 (Framework 4.6.25815.02), 64bit RyuJIT
  Job-TRABFR : .NET Core 2.0.3 (Framework 4.6.25815.02), 64bit RyuJIT

RemoveOutliers=False  Runtime=Core  Server=True  
LaunchCount=3  RunStrategy=Throughput  TargetCount=10  
WarmupCount=5  

```
**BEFORE:**

 |              Method | ChunkSize | MessageLength |     Mean |     Error |    StdDev |         Op/s | Allocated |
 |-------------------- |---------- |-------------- |---------:|----------:|----------:|-------------:|----------:|
 | **SingleBinaryMessage** |        **32** |            **64** | **51.29 ns** |  **2.697 ns** |  **4.037 ns** | **19,498,377.5** |       **0 B** |
 |   SingleTextMessage |        32 |            64 | 68.24 ns |  7.703 ns | 11.530 ns | 14,655,052.9 |       0 B |
 | **SingleBinaryMessage** |        **32** |           **128** | **52.73 ns** |  **1.231 ns** |  **1.842 ns** | **18,963,258.0** |       **0 B** |
 |   SingleTextMessage |        32 |           128 | 77.85 ns | 14.517 ns | 21.728 ns | 12,844,851.5 |       0 B |
 | **SingleBinaryMessage** |        **64** |            **64** | **79.02 ns** |  **3.400 ns** |  **5.089 ns** | **12,655,765.7** |       **0 B** |
 |   SingleTextMessage |        64 |            64 | 94.65 ns |  7.637 ns | 11.430 ns | 10,564,702.5 |       0 B |
 | **SingleBinaryMessage** |        **64** |           **128** | **75.85 ns** |  **3.455 ns** |  **5.171 ns** | **13,184,063.8** |       **0 B** |
 |   SingleTextMessage |        64 |           128 | 88.38 ns |  5.542 ns |  8.294 ns | 11,314,405.0 |       0 B |


**AFTER:**

 |              Method | ChunkSize | MessageLength |     Mean |    Error |    StdDev |         Op/s | Allocated |
 |-------------------- |---------- |-------------- |---------:|---------:|----------:|-------------:|----------:|
 | **SingleBinaryMessage** |        **32** |            **64** | **25.92 ns** | **1.053 ns** |  **1.577 ns** | **38,575,094.5** |       **0 B** |
 |   SingleTextMessage |        32 |            64 | 39.73 ns | 5.846 ns |  8.750 ns | 25,167,756.2 |       0 B |
 | **SingleBinaryMessage** |        **32** |           **128** | **31.86 ns** | **1.329 ns** |  **1.989 ns** | **31,390,234.4** |       **0 B** |
 |   SingleTextMessage |        32 |           128 | 40.19 ns | 4.999 ns |  7.483 ns | 24,879,229.8 |       0 B |
 | **SingleBinaryMessage** |        **64** |            **64** | **26.82 ns** | **1.538 ns** |  **2.302 ns** | **37,288,000.5** |       **0 B** |
 |   SingleTextMessage |        64 |            64 | 35.74 ns | 3.981 ns |  5.959 ns | 27,979,007.2 |       0 B |
 | **SingleBinaryMessage** |        **64** |           **128** | **32.06 ns** | **2.111 ns** |  **3.160 ns** | **31,186,746.6** |       **0 B** |
 |   SingleTextMessage |        64 |           128 | 48.04 ns | 7.095 ns | 10.619 ns | 20,816,626.6 |       0 B |

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/IDataEncoder.cs
```diff
@@ -1,11 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+
 namespace Microsoft.AspNetCore.SignalR.Internal.Encoders
 {
     public interface IDataEncoder
```
KrzysztofCwalina: @davidfowl, @ahsonkhan, should this interface follow the IBufferOperation pattern? i.e. both input and output are in parameters. As designed today, the interface pretty much torpedoes all non allocating APIs. In case of pass through decoder, nothing is allocated, but the moment the decoder does something useful, it will have to allocate.

moozzyk: @KrzysztofCwalina - we need to look at at the entire parsing/writing stack and remove byte[] there. I think it will be required when switching to pipelines.

PR #1128, opened 2017-11-16
Avoiding serializing to MemoryStream

moozzyk:


moozzyk:
As per https://github.com/aspnet/SignalR/pull/1127/files/c193772d043c09bf13358aefa0a9cbf299f85690#r151273273

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -23,18 +23,29 @@ public ReadOnlySpan<byte> Decode(byte[] payload)
             return decoded.Slice(0, written);
         }
 
+        private const int Int32OverflowLength = 10;
+
         public byte[] Encode(byte[] payload)
         {
-            Span<byte> buffer = new byte[Base64.GetMaxEncodedToUtf8Length(payload.Length)];
+            //The format is: [{length}:{message};] so allocate enough to be able to write the entire message
+            Span<byte> buffer = new byte[Int32OverflowLength + 1 + Base64.GetMaxEncodedToUtf8Length(payload.Length) + 1];
```
ahsonkhan: **Question:** Could this result in integer overflow if `Base64.GetMaxEncodedToUtf8Length(payload.Length)` is close to int.MaxValue? Is the expectation that the payload is relatively small?

moozzyk: We expect the messages to be relatively small but in theory we support messages up to 2GB before encoding. In this particular case - when a message needs to be base64 encoded - even a smaller original message may not be able to be encoded. I guess I would just check if the message could potentially exceed 2GB and throw.

ahsonkhan: > We expect the messages to be relatively small but in theory we support messages up to 2GB before encoding. 

GetMaxEncodedToUtf8Length() already throws for any payload.Length larger than 1610612733.
https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Buffers/Text/Base64Encoder.cs#L107

Would that be a concern?

moozzyk: I know, I checked. I think it is fine. Any exception from this code path closes the connection and the stack trace should make it clear what happened. I don't expect anyone hitting this exception though - the biggest message someone tried to send was a few MBs and to hit this code path the user would have to use binary formats (e.g msgpack) over text transport (serverSentEvents) neither of which is the default.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -23,18 +22,37 @@ public ReadOnlySpan<byte> Decode(byte[] payload)
...
             {
-                LengthPrefixedTextMessageWriter.WriteMessage(buffer.Slice(0, written), stream);
-                return stream.ToArray();
+                buffer[Int32OverflowLength - 1 - prefixLength] = (byte)('0' + prefix % 10);
+                prefix /= 10;
+                prefixLength++;
             }
+            while (prefix > 0);
+
+            return buffer.Slice(Int32OverflowLength - prefixLength, prefixLength + 1 + written + 1).ToArray();
```
davidfowl: We have to change this interface from returning arrays

moozzyk: This was supposed to be done after/as part of changing to pipelines because it is just a tip of an iceberg.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -23,18 +22,37 @@ public ReadOnlySpan<byte> Decode(byte[] payload)
...
         }
 
+        private const int Int32OverflowLength = 10;
+
         public byte[] Encode(byte[] payload)
         {
-            Span<byte> buffer = new byte[Base64.GetMaxEncodedToUtf8Length(payload.Length)];
+            var maxEncodedLength = Base64.GetMaxEncodedToUtf8Length(payload.Length);
+
+            // Int32OverflowLength + 2 - Int32OverflowLength + length of separator (':') + length of terminator (';')
```
BrennanConroy: `Int32OverflowLength + 2 =` the `-` makes it look like subtract

BrennanConroy: Probably don't even need the `Int32OverflowLength + 2` part

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -23,18 +22,37 @@ public ReadOnlySpan<byte> Decode(byte[] payload)
...
+            if (int.MaxValue - maxEncodedLength < Int32OverflowLength + 2)
+            {
+                throw new FormatException("The encoded message exceeds maximum supported size.");
```
BrennanConroy: `exceeds the`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -23,18 +22,37 @@ public ReadOnlySpan<byte> Decode(byte[] payload)
...
+            //The format is: [{length}:{message};] so allocate enough to be able to write the entire message
+            Span<byte> buffer = new byte[Int32OverflowLength + 1 + maxEncodedLength + 1];
 
-            var status = Base64.EncodeToUtf8(payload, buffer, out _, out var written);
+            buffer[Int32OverflowLength] = (byte)':';
+            var status = Base64.EncodeToUtf8(payload, buffer.Slice(Int32OverflowLength + 1), out _, out var written);
             Debug.Assert(status == OperationStatus.Done);
 
-            using (var stream = new MemoryStream())
+            buffer.Slice(Int32OverflowLength + 1).Slice(written)[0] = (byte)';';
```
BrennanConroy: This is the same as `buffer.Slice(Int32OverflowLength + 1 + written)`?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Encoders/LengthPrefixedTextMessageFormatterTests.cs
```diff
@@ -42,5 +44,38 @@ public void WriteMessage(string encoded, string payload)
 
             Assert.Equal(encoded, Encoding.UTF8.GetString(output.ToArray()));
         }
+
+        private static class LengthPrefixedTextMessageWriter
```
BrennanConroy: ? Now you're testing test code

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
moozzyk: I removed slicing altogether.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Encoders/LengthPrefixedTextMessageFormatterTests.cs
moozzyk: I did not read the tests carefully. Rewriting the tests to test the encoder. Thanks for catching this.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Encoders/LengthPrefixedTextMessageFormatterTests.cs
```diff
@@ -1,46 +1,38 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
```
BrennanConroy: Change the file name too

BrennanConroy: You changed the wrong file name

moozzyk: Jeez. TGIF. Thanks for catching this.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Encoders/LengthPrefixedTextMessageFormatterTests.cs
```diff
@@ -1,46 +0,0 @@
...
-using System.Text;
-using Microsoft.AspNetCore.SignalR.Internal.Encoders;
-using Xunit;
-
-namespace Microsoft.AspNetCore.SignalR.Tests.Internal.Encoders
-{
-    public class LengthPrefixedTextMessageFormatterTests
-    {
-        [Fact]
-        public void WriteMultipleMessages()
```
BrennanConroy: Do we have a test like this somewhere else?

moozzyk: I don't think so because the test is trying to test a scenario that is/was not a thing. Before we would create a temporary stream for each message just to serialize to get the byte[]. We didn't serialize multiple messages to a stream. Now even the temporary stream is gone. I think this test was written when we had framing at the Sockets level where we would use this format for text frames. At that time it was possible to write multiple messages to a stream. The `LengthPrefixedTextMessageWriter` was then re-purposed to be used only for base64 encoding but the test was not deleted at that time.

PR #1131, opened 2017-11-16
Give redis docker image time to download

BrennanConroy:
Our CI was taking 5.9 seconds to download new redis images


davidfowl:
Isn't the image cached?

BrennanConroy:
Not when there is a new version, but otherwise I think it is

PR #1147, opened 2017-11-22
Making HttpConnection restartable (C#)

moozzyk:
~~(a.k.a. Opening Pandora box)~~

### DIFF
--- a/samples/ClientSample/HubSample.cs
+++ b/samples/ClientSample/HubSample.cs
@@ -30,11 +30,22 @@ namespace ClientSample
             baseUrl = string.IsNullOrEmpty(baseUrl) ? "http://localhost:5000/default" : baseUrl;
 
             Console.WriteLine("Connecting to {0}", baseUrl);
-            HubConnection connection = await ConnectAsync(baseUrl);
-            Console.WriteLine("Connected to {0}", baseUrl);
+            var connection = new HubConnectionBuilder()
+                .WithUrl(baseUrl)
+                .WithConsoleLogger(LogLevel.Trace)
+                .Build();
 
             try
             {
+                var closeTcs = new TaskCompletionSource<object>();
+                connection.Closed += e => closeTcs.SetResult(null);
+                // Set up handler
+                connection.On<string>("Send", Console.WriteLine);
+
+                await ConnectAsync(connection);
+
+                Console.WriteLine("Connected to {0}", baseUrl);
+
                 var sendCts = new CancellationTokenSource();
 
                 Console.CancelKeyPress += async (sender, a) =>
@@ -45,13 +56,10 @@ namespace ClientSample
                     await connection.DisposeAsync();
                 };
 
-                // Set up handler
-                connection.On<string>("Send", Console.WriteLine);
-
-                while (!connection.Closed.IsCompleted)
+                while (!closeTcs.Task.IsCompleted)
                 {
-                    var completedTask = await Task.WhenAny(Task.Run(() => Console.ReadLine()), connection.Closed);
-                    if (completedTask == connection.Closed)
+                    var completedTask = await Task.WhenAny(Task.Run(() => Console.ReadLine()), closeTcs.Task);
+                    if (completedTask == closeTcs.Task)
                     {
                         break;
                     }
@@ -79,19 +87,15 @@ namespace ClientSample
             return 0;
         }
 
-        private static async Task<HubConnection> ConnectAsync(string baseUrl)
+        private static async Task ConnectAsync(HubConnection connection)
         {
             // Keep trying to until we can start
             while (true)
             {
-                var connection = new HubConnectionBuilder()
-                                .WithUrl(baseUrl)
-                                .WithConsoleLogger(LogLevel.Trace)
-                                .Build();
+
                 try
                 {
                     await connection.StartAsync();
-                    return connection;
                 }
                 catch (Exception)
                 {
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -34,16 +34,16 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private HubProtocolReaderWriter _protocolReaderWriter;
 
         private readonly object _pendingCallsLock = new object();
-        private readonly CancellationTokenSource _connectionActive = new CancellationTokenSource();
         private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>();
         private readonly ConcurrentDictionary<string, List<InvocationHandler>> _handlers = new ConcurrentDictionary<string, List<InvocationHandler>>();
+        private CancellationTokenSource _connectionActive;
 
         private int _nextId = 0;
         private volatile bool _startCalled;
         private Timer _timeoutTimer;
         private bool _needKeepAlive;
 
-        public Task Closed { get; }
+        public event Action<Exception> Closed;
 
         /// <summary>
         /// Gets or sets the server timeout interval for the connection. Changes to this value
@@ -69,11 +69,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
             _connection.OnReceived((data, state) => ((HubConnection)state).OnDataReceivedAsync(data), this);
-            Closed = _connection.Closed.ContinueWith(task =>
-            {
-                Shutdown(task.Exception);
-                return task;
-            }).Unwrap();
+            _connection.Closed += e => Shutdown(e);
 
             // Create the timer for timeout, but disabled by default (we enable it when started).
             _timeoutTimer = new Timer(state => ((HubConnection)state).TimeoutElapsed(), this, Timeout.Infinite, Timeout.Infinite);
@@ -122,12 +118,14 @@ namespace Microsoft.AspNetCore.SignalR.Client
             transferModeFeature.TransferMode = requestedTransferMode;
             await _connection.StartAsync();
             _needKeepAlive = _connection.Features.Get<IConnectionInherentKeepAliveFeature>() == null;
+
             var actualTransferMode = transferModeFeature.TransferMode;
 
             _protocolReaderWriter = new HubProtocolReaderWriter(_protocol, GetDataEncoder(requestedTransferMode, actualTransferMode));
 
             _logger.HubProtocol(_protocol.Name);
 
+            _connectionActive = new CancellationTokenSource();
             using (var memoryStream = new MemoryStream())
             {
                 NegotiationProtocol.WriteMessage(new NegotiationMessage(_protocol.Name), memoryStream);
@@ -151,13 +149,16 @@ namespace Microsoft.AspNetCore.SignalR.Client
             return new PassThroughEncoder();
         }
 
+        public async Task StopAsync() => await StopAsyncCore().ForceAsync();
+
+        private Task StopAsyncCore() => _connection.StopAsync();
+
         public async Task DisposeAsync() => await DisposeAsyncCore().ForceAsync();
 
         private async Task DisposeAsyncCore()
         {
             _timeoutTimer.Dispose();
             await _connection.DisposeAsync();
-            await Closed;
         }
 
         // TODO: Client return values/tasks?
@@ -370,12 +371,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        private void Shutdown(Exception ex = null)
+        private void Shutdown(Exception exception = null)
         {
             _logger.ShutdownConnection();
-            if (ex != null)
+            if (exception != null)
             {
-                _logger.ShutdownWithError(ex);
+                _logger.ShutdownWithError(exception);
             }
 
             lock (_pendingCallsLock)
@@ -388,13 +389,22 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 foreach (var outstandingCall in _pendingCalls.Values)
                 {
                     _logger.RemoveInvocation(outstandingCall.InvocationId);
-                    if (ex != null)
+                    if (exception != null)
                     {
-                        outstandingCall.Fail(ex);
+                        outstandingCall.Fail(exception);
                     }
                     outstandingCall.Dispose();
                 }
                 _pendingCalls.Clear();
+            }
+
+            try
+            {
+                Closed?.Invoke(exception);
+            }
+            catch (Exception ex)
+            {
+                _logger.ErrorDuringClosedEvent(ex);
             }
         }
 
--- a/src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
@@ -12,12 +12,13 @@ namespace Microsoft.AspNetCore.Sockets.Client
     {
         Task StartAsync();
         Task SendAsync(byte[] data, CancellationToken cancellationToken);
+        Task StopAsync();
         Task DisposeAsync();
         Task AbortAsync(Exception ex);
 
         IDisposable OnReceived(Func<byte[], object, Task> callback, object state);
 
-        Task Closed { get; }
+        event Action<Exception> Closed;
 
         IFeatureCollection Features { get; }
     }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -28,17 +28,20 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
 
-        private volatile int _connectionState = ConnectionState.Initial;
+        private volatile ConnectionState _connectionState = ConnectionState.Disconnected;
+        private readonly object _stateChangeLock = new object();
+
         private volatile ChannelConnection<byte[], SendMessage> _transportChannel;
         private readonly HttpClient _httpClient;
         private readonly HttpOptions _httpOptions;
         private volatile ITransport _transport;
         private volatile Task _receiveLoopTask;
-        private TaskCompletionSource<object> _startTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
-        private readonly TaskCompletionSource<object> _closedTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
-        private TaskQueue _eventQueue = new TaskQueue();
+        private TaskCompletionSource<object> _startTcs;
+        private TaskCompletionSource<object> _closeTcs;
+        private TaskQueue _eventQueue;
         private readonly ITransportFactory _transportFactory;
         private string _connectionId;
+        private Exception _abortException;
         private readonly TimeSpan _eventQueueDrainTimeout = TimeSpan.FromSeconds(5);
         private ChannelReader<byte[]> Input => _transportChannel.Input;
         private ChannelWriter<SendMessage> Output => _transportChannel.Output;
@@ -49,7 +52,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         public IFeatureCollection Features { get; } = new FeatureCollection();
 
-        public Task Closed => _closedTcs.Task;
+        public event Action<Exception> Closed;
 
         public HttpConnection(Uri url)
             : this(url, TransportType.All)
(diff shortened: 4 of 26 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -32,14 +32,14 @@ public class HubConnection
...
-        private readonly CancellationTokenSource _connectionActive = new CancellationTokenSource();
         private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>();
         private readonly ConcurrentDictionary<string, List<InvocationHandler>> _handlers = new ConcurrentDictionary<string, List<InvocationHandler>>();
+        private CancellationTokenSource _connectionActive;
 
         private int _nextId = 0;
         private volatile bool _startCalled;
 
-        public Task Closed { get; }
+        public event Action<Exception> Closed;
```
BrennanConroy: Shouldn't this be a wrapper around `_connection.Closed`?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -103,25 +105,27 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
 
         private Task StartAsyncCore()
         {
-            if (Interlocked.CompareExchange(ref _connectionState, ConnectionState.Connecting, ConnectionState.Initial)
-                != ConnectionState.Initial)
+            if (ChangeState(from: ConnectionState.Disconnected, to: ConnectionState.Connecting) != ConnectionState.Disconnected)
             {
                 return Task.FromException(
-                    new InvalidOperationException("Cannot start a connection that is not in the Initial state."));
+                    new InvalidOperationException("Cannot start a connection that is not in the Disconnected state."));
```
BrennanConroy: `nameof(ConnectionState.Disconnected)`!

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -435,18 +445,28 @@ private void SetTransferMode(TransferMode transferMode)
...
 
-        public async Task DisposeAsync() => await DisposeAsyncCore().ForceAsync();
-
-        private async Task DisposeAsyncCore()
+        public async Task StopAsync()
         {
-            _logger.StoppingClient(_connectionId);
-
-            if (Interlocked.Exchange(ref _connectionState, ConnectionState.Disconnected) == ConnectionState.Initial)
+            lock(_stateChangeLock)
```
BrennanConroy: space

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -522,12 +561,36 @@ public void Dispose()
             }
         }
 
-        private class ConnectionState
+        private ConnectionState ChangeState(ConnectionState from, ConnectionState to)
+        {
+            lock(_stateChangeLock)
```
BrennanConroy: space

src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
```diff
@@ -12,11 +12,12 @@ public interface IConnection
     {
         Task StartAsync();
         Task SendAsync(byte[] data, CancellationToken cancellationToken);
+        Task StopAsync();
         Task DisposeAsync();
 
         IDisposable OnReceived(Func<byte[], object, Task> callback, object state);
 
-        Task Closed { get; }
+        event Action<Exception> Closed;
```
BrennanConroy: Used to be `Func<Exception, Task>`, any reason we don't go back to that?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -190,25 +196,29 @@ private Task StartAsyncCore()
                     await _eventQueue.Drain();
 
                     await Task.WhenAny(_eventQueue.Drain().NoThrow(), Task.Delay(_eventQueueDrainTimeout));
-                    _httpClient?.Dispose();
 
                     _logger.CompleteClosed(_connectionId);
+
+                    // At this point the connection can be either in the Connected or Disposed state. The state should be changed
```
BrennanConroy: Thanks for the comments 👍

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -190,25 +196,29 @@ private Task StartAsyncCore()
...
+                    // to the Disconnected state only if it was in the Connected state.
+                    ChangeState(from: ConnectionState.Connected, to: ConnectionState.Disconnected);
+
+                    _closeTcs.SetResult(null);
+
                     if (t.IsFaulted)
                     {
-                        _closedTcs.TrySetException(t.Exception.InnerException);
+                        Closed?.Invoke(t.Exception.InnerException);
```
BrennanConroy: We need to `try catch` around these and log `ExceptionThrownFromCallback`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -103,25 +105,27 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+            _eventQueue = new TaskQueue();
+
             StartAsyncInternal()
                 .ContinueWith(t =>
                 {
                     if (t.IsFaulted)
                     {
                         _startTcs.SetException(t.Exception.InnerException);
-                        _closedTcs.TrySetException(t.Exception.InnerException);
+                        Closed?.Invoke(t.Exception);
```
BrennanConroy: `InnerException`

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
```diff
@@ -153,6 +153,8 @@ internal static class SocketClientLoggerExtensions
         private static readonly Action<ILogger, DateTime, string, string, Exception> _exceptionThrownFromCallback =
             LoggerMessage.Define<DateTime, string, string>(LogLevel.Error, new EventId(19, nameof(ExceptionThrownFromCallback)), "{time}: Connection Id {connectionId}: An exception was thrown from the '{callback}' callback");
```
BrennanConroy: Unrelated to this PR, but could you add a `.` to this log?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -102,6 +102,78 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+            {
+                const string originalMessage = "SignalR";
+                var httpConnection = new HttpConnection(new Uri(_serverFixture.Url + path), transportType, loggerFactory);
+                var connection = new HubConnection(httpConnection, protocol, loggerFactory);
+                try
+                {
+                    await connection.StartAsync().OrTimeout();
+                    var result = await connection.InvokeAsync<string>("Echo", originalMessage).OrTimeout();
+                    Assert.Equal(originalMessage, result);
+                    await connection.StopAsync();
```
BrennanConroy: `OrTimeout`

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -102,6 +102,78 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                        restartTcs.SetResult(null);
+                    }
+                };
+
+                try
+                {
+                    await connection.StartAsync().OrTimeout();
+                    var result = await connection.InvokeAsync<string>("Echo", originalMessage).OrTimeout();
+                    Assert.Equal(originalMessage, result);
+                    await connection.StopAsync();
```
BrennanConroy: `OrTimeout`

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -102,6 +102,78 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+            }
+        }
+
+        [Fact]
+        public async Task CanStartConnectionFromClosedEvent()
+        {
+            using (StartLog(out var loggerFactory))
+            {
+                const string originalMessage = "SignalR";
+                var httpConnection = new HttpConnection(new Uri(_serverFixture.Url + "/default"));
```
BrennanConroy: add loggerFactory

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -76,7 +76,7 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
         }
 
         [Fact]
-        public async Task CannotStartStoppedConnection()
+        public async Task CannotStartConnectionDisposedAfterStarting()
```
BrennanConroy: `CannotStartDisposedConnection`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -159,6 +159,217 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+            {
+                await connection.StartAsync().OrTimeout();
+            }
+            catch { }
+            failNegotiate = false;
+            await connection.StartAsync().OrTimeout();
+            await connection.DisposeAsync().OrTimeout();
+        }
+
+
```
BrennanConroy: nit: newline

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -159,6 +159,217 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+                });
+
+            var connection = new HttpConnection(new Uri("http://fakeuri.org/"), TransportType.LongPolling, loggerFactory: null,
+                httpOptions: new HttpOptions { HttpMessageHandler = mockHttpHandler.Object });
+
+            var closeTcs = new TaskCompletionSource<object>();
+            connection.Closed += e => closeTcs.TrySetResult(null);
+
+            var startTask = connection.StartAsync();
+            await allowStopTcs.Task;
```
BrennanConroy: `OrTimeout`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -159,6 +159,217 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+
+            await Task.WhenAll(startTask.OrTimeout(), connection.StopAsync().OrTimeout());
```
BrennanConroy: Just `OrTimeout` the `WhenAll` call

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -159,6 +159,217 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+            await closeTcs.Task.OrTimeout();
+        }
+
+        [Fact]
+        public async Task CanStartConnectionStoppedWithError()
```
BrennanConroy: `CanStartConnectionAfterStoppedWithError`?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -159,6 +159,217 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+            var closeTcs = new TaskCompletionSource<object>();
+            connection.Closed += e => closeTcs.TrySetResult(null);
+
+            await connection.StartAsync().OrTimeout();
+            try
+            {
+                await connection.SendAsync(new byte[] { 0x42 }).OrTimeout();
+            }
+            catch { }
+            await closeTcs.Task.OrTimeout().OrTimeout();
```
BrennanConroy: Dupe `OrTimeout`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -159,6 +159,217 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+            await connection.StartAsync().OrTimeout();
+            await connection.DisposeAsync().OrTimeout();
+        }
+
+        [Fact]
+        public async Task CanStopStoppedConnection()
+        {
+            var mockHttpHandler = new Mock<HttpMessageHandler>();
```
BrennanConroy: unused

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -134,16 +138,16 @@ private byte[] FormatMessageToArray(byte[] message)
                         }
                     }
                 }
-                _closeTcs.TrySetResult(null);
+                Closed(null);
```
BrennanConroy: `Closed?.Invoke(...)`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
moozzyk: No. This way the only event that is hooked to HttpConnection is the one from hubConnection and the user subscribes to the object they are using.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: Fixed. Initially it was not possible because state was tracked as int not an enum.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: I actually removed it altogether - the connection has not been started so the Closed event should not be invoked. Note that the user will get exception from StartAsync.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
moozzyk: No. Exceptions should not be thrown from event handlers.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
moozzyk: Fixed.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
moozzyk: A test called `CannotStartDisposedConnection` already exists and this one tests a little different scenario.

src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
moozzyk: I am not sure we want it (and what we had was not implemented correctly). I think we need to try which pattern works better for restarting the connection from this event.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -364,6 +365,8 @@ private void Shutdown(Exception ex = null)
                 }
                 _pendingCalls.Clear();
             }
+
+            Closed?.Invoke(ex);
```
BrennanConroy: try catch, don't trust user code

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -190,25 +194,25 @@ private Task StartAsyncCore()
...
+                    _closeTcs.SetResult(null);
+
                     if (t.IsFaulted)
                     {
-                        _closedTcs.TrySetException(t.Exception.InnerException);
-                    }
-                    if (t.IsCanceled)
-                    {
-                        _closedTcs.TrySetCanceled();
+                        Closed?.Invoke(t.Exception.InnerException);
```
BrennanConroy: Can't trust user code, should wrap this

anurse:
This needs a bunch of rebasing and adjusting after merging #1165 . I'll do that and ping this when I'm done.

anurse:
Ok, it should be :up: to :date: now. I had to refactor some things to handle the new AbortAsync. Also, some tests seem to have been broken by the fact that Dispose waits on Start to finish, but I fixed that.

Finally, there was a test for Received not firing after Dispose is called which isn't really possible since the state changes that trigger that all occur after the channel is marked completed anyway. The test was broken so I just removed it (because it is no longer relevant).

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -355,46 +592,63 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
         [Fact]
         public async Task EventQueueTimeout()
```
anurse: Turns out this test wasn't actually working and it was hiding an issue in the EventQueue drain logic :)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -228,73 +463,37 @@ public void CannotStartConnectionWithInvalidTransportType(TransportType requeste
...
+                await Assert.ThrowsAsync<HttpRequestException>(() => closeTcs.Task.OrTimeout());
             }
             finally
             {
                 await connection.DisposeAsync();
             }
         }
 
-        [Fact]
-        public async Task ReceivedCallbackNotRaisedAfterConnectionIsDisposed()
```
anurse: This is the test that I mentioned in other comments. It is not really relevant any more as the channel is closed before the state changes to make this event no-op occur.

anurse:
Should be green after this next build. ping @BrennanConroy @mikaelm12 

This one is unfortunately a bit of a tricky one, and I'm thinking I might do some test refactoring after this PR to make them clearer as they're a bit messy now (due to how tests tend to grow organically)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -306,7 +320,11 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
                 await _transport.StartAsync(connectUrl, applicationSide, GetTransferMode(), _connectionId, this);
 
                 // actual transfer mode can differ from the one that was requested so set it on the feature
-                Debug.Assert(_transport.Mode.HasValue, "transfer mode not set after transport started");
+                if(!_transport.Mode.HasValue)
```
BrennanConroy: space

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -435,31 +453,42 @@ private void SetTransferMode(TransferMode transferMode)
...
+        //     before it was aborted
+        //  2. If the transport is closed gracefully and then AbortAsync is called before it captures the _abortException value
+        //     the graceful shutdown could be turned into an abort. However, again, this is an inherent race between two different conditions:
+        //     The transport shutting down because the server went away, and the user requesting the Abort
+        //  3. Finally, because this is an instance field, there is a possible race around accidentally re-using _abortException in the restarted
+        //     connection. The scenario here is: AbortAsync(someException); StartAsync(); CloseAsync(); Where the _abortException value from the
+        //     first AbortAsync call is still set at the time CloseAsync gets to calling the Closed event. However, this can't happen because the
+        //     StartAsync method can't be called until the connection state is changed to Disconnected, which happens AFTER the close code
+        //     captures and resets _abortException.
+        public async Task AbortAsync(Exception ex) => await StopAsyncCore(ex ?? throw new ArgumentNullException(nameof(ex))).ForceAsync();
```
BrennanConroy: super nit: public API we should name the argument `exception`

samples/ClientSample/HubSample.cs
```diff
@@ -79,19 +87,15 @@ internal static void Register(CommandLineApplication app)
...
-                var connection = new HubConnectionBuilder()
-                                .WithUrl(baseUrl)
-                                .WithConsoleLogger(LogLevel.Trace)
-                                .Build();
+
                 try
                 {
                     await connection.StartAsync();
-                    return connection;
                 }
```
fretje: Don't you need to return here? Otherwise this method will never return... or am I missing something?

BrennanConroy: The return value was changed from `Task<HubConnection>` to `Task`

acjh: What @fretje means is that this is an infinite loop:

```c#
while (true)
{
    try
    {
    }
    catch (Exception)
    {
    }
}
```

BrennanConroy: @acjh Thanks! Yes you guys are right, this is broken. Would either of you like to make a PR to fix it?

Issue #1160, opened 2017-11-27
Flaky Docker Redis tests


anurse:
Expanding this somewhat: I now have consistent failures on my desktop at home due to the docker container not starting properly. We should add some better diagnostics to these tests to report issues if the container fails to start properly so we can track down why they are failing on some (read: my) machines.

When your Docker for Windows instance is configured either for "Windows Container" or "Linux Containers on Windows" (LCOW), the Redis tests fail to run. In the first case, because Redis is a linux container and in the second case because of some bug in LCOW (it's experimental) that causes it to fail.

We should detect those cases and skip running the tests. You can detect this by the following command:

With Windows Containers or LCOW on (what we don't want):

```
› docker system info -f "{{.OSType}}"
windows
```

With it off (what we want):

```
› docker system info -f "{{.OSType}}"
linux
```

muratg:
These are running now. @ryanbrandenburg let us know if you hit this again.

PR #1161, opened 2017-11-27
Implement #1156 by having the server send Ping messages

### LINKED ISSUE #715: Create a common test project for test helpers and such
Common started with 1 file and so we just used `<compile include=""/>` now it is about to have 4 files. We should change it to a project.

### LINKED ISSUE #1156: [Keep-Alive] Server pinging
As part of keep-alive, we want the client to automatically consider a connection terminated if the server hasn't sent anything in a while. In order for that to work, the server needs to send [Ping messages](https://github.com/aspnet/SignalR/blob/dev/specs/HubProtocol.md#ping-aka-keep-alive) frequently. However, we only want to do this for the WebSockets and Server-Sent Events transports. For Long Polling, it would disrupt the "long poll" by ending it prematurely and is unnecessary since the transport already supports a "reconnection" mechanic for issues inherent to long polling.

Related: #1157

anurse:
This change teaches `HubEndpoint` to send `PingMessage.Instance` to the client at regular intervals, if no other messages have been sent. But only for SSE and WebSockets.

I found a sneaky and possibly ingenious (or possibly terrible) way to implement keep-alive. Since the HubEndpoint already creates a Channel that it uses to marshal writes from multiple writers, I just wrap it in a special `ChannelReader` that emits `PingMessage.Instance` if `WaitToReadAsync` blocks for more than the desired interval.

There is a feature, `IConnectionInherentKeepAliveFeature`, that indicates if a transport has "inherent keep-alive", like Long Polling (for which the poll itself serves as a way for the client to tell if the server is gone)

Notes:
* `package-lock.json` got changed, but I sense it will keep changing if we don't check in the changes, they're just ordering changes, no new packages.
* The Redis Docker tests are flaky on my home desktop (seems to be more than just #1160 after all @BrennanConroy) so I had to disable them

Fixes #1156

Also fixes #715


samples/SocketsSample/wwwroot/hubs.html
```diff
@@ -104,7 +104,7 @@ <h4>Private Message</h4>
     connectButton.disabled = true;
     disconnectButton.disabled = false;
     console.log('http://' + document.location.host + '/' + hubRoute);
-    connection = new signalR.HubConnection(hubRoute, logger, { transport: transportType, logger: logger });
+    connection = new signalR.HubConnection(hubRoute, { transport: transportType, logging: logger });
```
BrennanConroy: Copycat https://github.com/aspnet/SignalR/pull/1158/files#diff-c80b6b1153b8f1588d86994e37bb3c7fR107

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -8,6 +8,7 @@
 using System.Threading.Channels;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
+using Microsoft.AspNetCore.Sockets.Features;
```
BrennanConroy: nit: sort

samples/SocketsSample/wwwroot/hubs.html
anurse: Mine's better, I fixed all of them :P.

src/Microsoft.AspNetCore.Sockets.Http/HttpSocketOptions.cs
```diff
@@ -1,19 +1,27 @@
...
         public IList<IAuthorizeData> AuthorizationData { get; } = new List<IAuthorizeData>();
 
         public TransportType Transports { get; set; } = TransportType.All;
 
         public WebSocketOptions WebSockets { get; } = new WebSocketOptions();
 
         public LongPollingOptions LongPolling { get; } = new LongPollingOptions();
+
+        /// <summary>
+        /// The interval at which keep-alive messages should be sent. This will be
```
BrennanConroy: Exclamation point at the end `This will be!` :trollface:

src/Microsoft.AspNetCore.SignalR.Core/Internal/KeepAliveChannel.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+
+namespace Microsoft.AspNetCore.SignalR.Core.Internal
+{
+    internal class KeepAliveChannel : Channel<HubMessage>
```
davidfowl: I'm not too sure about this..

anurse: The thing I like about it is that it keeps the consumption code clean. Ideally, you really want to put keep-alive code in the `HubEndPoint.OnConnectionAsync.WriteToTransport` local function, since that's the central place messages appear. We could have a background task firing pings, but the goal of the ping is to send one if the timeout elapses while waiting for a message. A fixed interval kinda achieves that, but in a weird way and requires synchronization in order to know the time since the last message.

I tried putting this logic into the consumption side (in `WriteToTransport`) and it gets complicated fast, because KeepAlive isn't always enabled. Though I have another idea I'm going to try.

muratg: This class as a concept or name??

anurse: Probably the concept, it's a little sketchy and I'm not super happy with it myself.

anurse:
:up: :date: Tried a different approach to appease @davidfowl (and my own conscience)

BrennanConroy:
Spec update in this PR?

anurse:
No spec update. The only thing needed in the spec is the Ping frame. I'd argue that keep-alive behaviors are policy/config, not spec. I can look at throwing a sentence or two in though.

BrennanConroy:
It says "Either endpoint may send a Ping message at any time.", I don't see a good reason the client should be sending pings

anurse:
There are reasons for an implementation to want that. I'd rather keep the spec open. Just because it says an endpoint "may" do something doesn't mean our implementation has to do it.

BrennanConroy:
:'( easy to shoot yourself in the foot

anurse:
How so? There's no mechanism in our client to send a Ping message, and the server is tolerant of receiving them.

BrennanConroy:
If someone makes a client exactly like ours except it sends keep-alive, then that will cause the server to stop sending keep-alives and then the client could timeout since it doesn't receive a keep-alive. Not a big problem I guess

anurse:
Discussed with @BrennanConroy offline, he was slightly confused and we're all good 😃

anurse:
To clarify the behavior to avoid confusion: The keep-alive timer is only reset when a message is **sent** by the server. Receiving messages from the client does not reset the timer.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -76,27 +77,57 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
+                var serializedKeepAlive = protocolReaderWriter.WriteMessage(PingMessage.Instance);
+
                 try
                 {
-                    while (await output.Reader.WaitToReadAsync())
+                    while (true)
                     {
-                        while (output.Reader.TryRead(out var hubMessage))
+                        var keepAliveTask = needKeepAlive ?
+                            Task.Delay(_hubOptions.KeepAliveInterval) :
```
BrennanConroy: Should I be concerned that we new up a Task.Delay every send?

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/ConnectionInherentKeepAliveFeature.cs
```diff
@@ -0,0 +1,14 @@
+using System;
```
BrennanConroy: copyright

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionInherentKeepAliveFeature.cs
```diff
@@ -0,0 +1,21 @@
+using System;
```
BrennanConroy: copy

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -16,18 +17,25 @@ public class LongPollingTransport : IHttpTransport
...
+            _timeout = timeout;
             _application = application;
             _connectionId = connectionId;
             _logger = loggerFactory.CreateLogger<LongPollingTransport>();
         }
 
-        public async Task ProcessRequestAsync(HttpContext context, CancellationToken token)
+        public async Task ProcessRequestAsync(ConnectionContext connection, HttpContext context, CancellationToken token)
         {
+            if(connection.Features.Get<IConnectionInherentKeepAliveFeature>() == null)
```
BrennanConroy: space

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1333,78 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+
+            using (var client = new TestClient(false, new JsonHubProtocol()))
+            {
+                var endPointLifetime = endPoint.OnConnectedAsync(client.Connection).OrTimeout();
+
+                await client.Connected.OrTimeout();
+
+                // Echo a bunch of stuff, waiting 10ms between each, until 500ms have elapsed
+                DateTime start = DateTime.UtcNow;
+                while((DateTime.UtcNow - start).TotalMilliseconds <= 500.0)
```
BrennanConroy: space

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1333,78 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+
+                // Shut down
+                client.Dispose();
+
+                await endPointLifetime.OrTimeout();
+
+                // We shouldn't have any ping messages
+                HubMessage message;
+                var counter = 0;
+                while((message = await client.ReadAsync()) != null)
```
BrennanConroy: space

BrennanConroy: OrTimeout

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1333,78 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+                {
+                    counter += 1;
+                    Assert.IsNotType<PingMessage>(message);
+                }
+                Assert.InRange(counter, 1, Int32.MaxValue);
```
BrennanConroy: Maybe 50 would be better than max value

test/Microsoft.AspNetCore.Sockets.Tests/LongPollingTests.cs
```diff
@@ -11,6 +11,8 @@
 using Microsoft.AspNetCore.Sockets.Internal.Transports;
 using Microsoft.Extensions.Logging;
 using Xunit;
+using System;
```
BrennanConroy: sort

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
anurse: Not every send. Every time we need to go async to wait for the user to send. And you have to allocate a task for this no matter how you slice it.

anurse:
:up: :date:

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1333,78 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+
+                // We should have all pings
+                HubMessage message;
+                var counter = 0;
+                while ((message = await client.ReadAsync()) != null)
+                {
+                    counter += 1;
+                    Assert.Same(PingMessage.Instance, message);
+                }
+                Assert.InRange(counter, 1, Int32.MaxValue);
```
BrennanConroy: 1 - 5

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionInherentKeepAliveFeature.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The most common example of this feature is the Long Polling HTTP transport, which must (due to HTTP limitations) terminate
+    /// each poll within a particular interval and return a signal indicating "the server is still here, but there is no data yet".
+    /// This feature allows applications to add keep-alive functionality, but limit it only to transports that don't have some kind
+    /// of inherent keep-alive.
+    /// </remarks>
+    public interface IConnectionInherentKeepAliveFeature
+    {
+        TimeSpan KeepAliveInterval { get; }
```
BrennanConroy: Does this property provide any value? It isn't used currently

anurse: It exposes the duration of the inherent keep-alive (i.e. for Long Polling, the poll timeout value). To be honest, this is just used as a marker interface though, so we could drop it. It seemed like a useful value to have though.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: I'm going to do 1-10, because I want to make sure this doesn't flake out occasionally due to random delays here and there. Remember that `Task.Delay` only guarantees that you won't be woken up before the delay has elapsed; you might be sleeping for longer than the delay.

anurse:
:up: :date:

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionInherentKeepAliveFeature.cs
BrennanConroy: It would cleanup the LongPolling constructor a little :P

anurse: But marker interfaces are groooooooss

test/Microsoft.AspNetCore.Sockets.Tests/LongPollingTests.cs
```diff
@@ -19,28 +21,34 @@ public class LongPollingTests
...
+            var connection = new DefaultConnectionContext("foo", toTransport, toApplication);
+
             var timeoutToken = new CancellationToken(true);
-            var poll = new LongPollingTransport(timeoutToken, channel, connectionId: string.Empty, loggerFactory: new LoggerFactory());
+            var poll = new LongPollingTransport(timeoutToken, TimeSpan.Zero, toTransport.Reader, connectionId: string.Empty, loggerFactory: new LoggerFactory());
 
             using (var cts = CancellationTokenSource.CreateLinkedTokenSource(timeoutToken, context.RequestAborted))
             {
-                await poll.ProcessRequestAsync(context, cts.Token).OrTimeout();
+                await poll.ProcessRequestAsync(connection, context, cts.Token).OrTimeout().OrTimeout();
```
BrennanConroy: remove extra OrTimeout

anurse: What if the timeout times out?

anurse:
:up: :date: @davidfowl - This covers some of our in-person discussions. It needs a little clean-up, but the gist is what we talked about.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -136,15 +114,56 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
+
+                    // Serialize the keep-alive message in advance, it's a singleton and will always serialize the same.
+                    var serializedKeepAlive = protocolReaderWriter.WriteMessage(PingMessage.Instance);
+
+                    // Implements the keep-alive tick behavior
+                    // Each tick, we check if the time since the last send is larger than the keep alive duration (in ticks).
+                    // If it is, we send a ping frame, if not, we no-op on this tick. This means that in the worst case, the
+                    // true "ping rate" of the server could be (_hubOptions.KeepAliveInterval + HubEndPoint.KeepAliveTimerInterval),
+                    // because if the interval elapses right after the last tick of this timer, it won't be detected until the next tick.
+                    void KeepAliveTick(object state)
```
davidfowl: ugh, local functions. We should avoid capturing state.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -105,15 +114,56 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
                 }
             }
 
+            _lastSendTimestamp = Stopwatch.GetTimestamp();
             var writingOutputTask = WriteToTransport();
 
+            Timer keepAliveTimer = null;
```
davidfowl: We're going to have wayyyyy too many timers. We need a connection heartbeat design here. The only thing that has a connection list is the lifetime manager though. We need to think a bit more here.

davidfowl: I thought we were going to look at a single timer still model.

anurse: So did I until I realized how HubEndpoint works when reviewing it this morning :). We need to rethink this quite a bit. I think we need to go back to the slightly gross idea of having the connection manager tick the heartbeat via a feature. The problem there is the config gets gross. The ping rate is always going to be off by at most the tick rate and that gets weird.

anurse:
:up: :date: after discussions with @davidfowl, tests don't pass yet, but the pings are indeed sent.

anurse:
Pushed some test fixes, but I'm not happy with it. I think the HubLifetimeManager tests are too low-level and brittle. I want to rewrite them to use HubEndpoint but I don't have the time to do that today, I'll look at it tomorrow.

anurse:
I gave up trying to make the tests less brittle. It meant refactoring a huge amount of unit tests (because I needed to make TestClient changes to make it clean). I'm sure I could spend more energy and figure out an easier refactor, but I just don't have that kind of time :).

:up: :date: ready for review again. Tests pass, and we're now using a single heartbeat tick for all connections.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -4,43 +4,52 @@
...
-        private readonly ChannelWriter<HubMessage> _output;
         private readonly ConnectionContext _connectionContext;
+        private readonly ILogger _logger;
         private readonly CancellationTokenSource _connectionAbortedTokenSource = new CancellationTokenSource();
         private readonly TaskCompletionSource<object> _abortCompletedTcs = new TaskCompletionSource<object>();
+        private Task _writingTask = Task.CompletedTask;
 
-        public HubConnectionContext(ChannelWriter<HubMessage> output, ConnectionContext connectionContext)
+        private long _lastSendTimestamp = Stopwatch.GetTimestamp();
+        private long _keepAliveDuration;
```
davidfowl: nit: readonly

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -103,5 +244,16 @@ private static void AbortConnection(object state)
                 connection._abortCompletedTcs.TrySetException(ex);
             }
         }
+
+        public async Task WriteAsync(HubInvocationMessage message)
+        {
+            while(await Output.Writer.WaitToWriteAsync())
```
davidfowl: nit: formatting!

src/Microsoft.AspNetCore.SignalR.Core/Properties/AssemblyInfo.cs
```diff
@@ -3,5 +3,5 @@
 
 using System.Runtime.CompilerServices;
 
-
-[assembly: InternalsVisibleTo("Microsoft.AspNetCore.SignalR.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
\ No newline at end of file
+//[assembly: InternalsVisibleTo("Microsoft.AspNetCore.SignalR.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
davidfowl: Remove dead code

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionInherentKeepAliveFeature.cs
davidfowl: I'm with @BrennanConroy, I don't think this adds any value. Maybe this interface should have a bool so that keep alives can be manually turned off.

src/Microsoft.AspNetCore.Sockets/Internal/SocketEventSource.cs
```diff
@@ -74,5 +75,26 @@ public void ConnectionTimedOut(string connectionId)
                 }
             }
         }
+
+        [Event(eventId: 4, Level = EventLevel.Verbose, Message = "Scanning connections.")]
+        public void ScanningConnections()
+        {
+            if(IsEnabled() && IsEnabled(EventLevel.Verbose, EventKeywords.None))
```
davidfowl: nit: formatting!

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            _host = new WebHostBuilder()
+                .ConfigureLogging(builder => builder.AddProvider(new ForwardingLoggerProvider(_loggerFactory)))
+                .UseStartup(typeof(TStartup))
+                .UseKestrel()
+                .UseUrls(url)
+                .UseContentRoot(Directory.GetCurrentDirectory())
+                .Build();
+
+            var t = Task.Run(() => _host.Start());
```
davidfowl: I hate that this virus is spreading. We need to file a bug to fix this logic once and for all...

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1332,89 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+                // Wait 500 ms, but make sure to yield some time up to unblock concurrent threads
+                // This is useful on AppVeyor because it's slow enough to end up with no time
+                // being available for the endpoint to run.
+                for (var i = 0; i < 5; i += 1)
+                {
+                    // Tick the heartbeat feature manually, 10 times, every 10ms
+                    for(var j = 0; j < 10; j += 1)
+                    {
+                        client.Connection.TickHeartbeat();
+                        await Task.Yield();
```
davidfowl: What's this for?

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -62,50 +59,19 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
-            var output = Channel.CreateUnbounded<HubMessage>();
-
             // Set the hub feature before doing anything else. This stores
             // all the relevant state for a SignalR Hub connection.
             connection.Features.Set<IHubFeature>(new HubFeature());
 
-            var connectionContext = new HubConnectionContext(output, connection);
+            var connectionContext = new HubConnectionContext(connection, _hubOptions.KeepAliveInterval, _loggerFactory.CreateLogger<HubConnectionContext>());
```
davidfowl: Lets just pass the factory into the HubConnection.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -16,18 +17,25 @@ public class LongPollingTransport : IHttpTransport
...
+            _timeout = timeout;
             _application = application;
             _connectionId = connectionId;
             _logger = loggerFactory.CreateLogger<LongPollingTransport>();
         }
 
-        public async Task ProcessRequestAsync(HttpContext context, CancellationToken token)
+        public async Task ProcessRequestAsync(ConnectionContext connection, HttpContext context, CancellationToken token)
         {
+            if (connection.Features.Get<IConnectionInherentKeepAliveFeature>() == null)
```
davidfowl: This still makes me cringe. Do we call this before we call into the end point?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -4,43 +4,52 @@
...
 using Microsoft.AspNetCore.SignalR.Internal;
+using Microsoft.AspNetCore.SignalR.Internal.Encoders;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.AspNetCore.Sockets;
 using Microsoft.AspNetCore.Sockets.Features;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.SignalR
 {
     public class HubConnectionContext
```
davidfowl: General feedback for this class:
- Group method and properties by accessibility. public props then internal props (same for methods).
- The Input channel shouldn't be public. It would mean that any body can read from the HubConnection context which is fundamentally broken.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -62,50 +59,19 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
-            var output = Channel.CreateUnbounded<HubMessage>();
-
             // Set the hub feature before doing anything else. This stores
             // all the relevant state for a SignalR Hub connection.
             connection.Features.Set<IHubFeature>(new HubFeature());
```
davidfowl: This doesn't seem to be used anywhere anymore. Can we get rid of it altogether?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: Artifact from when the heartbeat was in a background timer. In my experience I'd seen cases where the timer fired but the async code it called didn't because xunit's sync context was blocking us. I'll remove it now though, should be unnecessary.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
anurse: Hrm. No we don't. Unfortunately I don't see an easy way to arrange this such that we do. It works today because of Negotiate and the way our client works, since the client always establishes the poll before sending the negotiate frame. As a result, `OnConnectionAsync` is still waiting for the negotiate message by the time the first poll comes in a sets the feature.

This could be a tricky one to figure out. I'm not sure the best way to rearrange this to make it work, but I'll take a look. Short of requiring the client to start the poll/SSE transport before even calling OnConnectedAsync, I'm not sure what to do.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
anurse: This is just copy-pasta, moving the "Common" code that was continuing to grow into a full-on test utilities assembly.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
anurse: Agreed, this class has needed some clean-up for a while

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
BrennanConroy: https://github.com/aspnet/SignalR/issues/715

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
anurse: What if the Channels along the path didn't "activate" until the transport was initialized? So `WaitForReadAsync` blocked until the first poll/event stream request came in. Then we check the feature after the first WaitForReadAsync?

src/Microsoft.AspNetCore.Sockets/Internal/SocketEventSource.cs
```diff
@@ -74,5 +75,26 @@ public void ConnectionTimedOut(string connectionId)
...
+            if (IsEnabled() && IsEnabled(EventLevel.Verbose, EventKeywords.None))
+            {
+                WriteEvent(4);
+            }
+        }
+
+        [Event(eventId: 5, Level = EventLevel.Verbose, Message = "Scanning connections in {0:0.00}ms.")]
+        private void ScannedConnections(double durationInMilliseconds) => WriteEvent(5, durationInMilliseconds);
+
+        [NonEvent]
```
BrennanConroy: `NonEvent` should be grouped at the top is what we talked about right?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1332,89 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+
+                // Shut down
+                client.Dispose();
+
+                await endPointLifetime.OrTimeout();
+
+                // We shouldn't have any ping messages
+                HubMessage message;
+                var counter = 0;
+                while ((message = await client.ReadAsync()) != null)
```
BrennanConroy: OrTimeout

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1332,89 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+
+            using (var client = new TestClient(false, new JsonHubProtocol()))
+            {
+                var endPointLifetime = endPoint.OnConnectedAsync(client.Connection).OrTimeout();
+                await client.Connected.OrTimeout();
+
+                // Wait 500 ms, but make sure to yield some time up to unblock concurrent threads
+                // This is useful on AppVeyor because it's slow enough to end up with no time
+                // being available for the endpoint to run.
+                for (var i = 0; i < 5; i += 1)
```
BrennanConroy: Can't this just be `for i < 50`?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1333,6 +1332,89 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
...
+
+                // Shut down
+                client.Dispose();
+
+                await endPointLifetime.OrTimeout();
+
+                // We should have all pings
+                HubMessage message;
+                var counter = 0;
+                while ((message = await client.ReadAsync()) != null)
```
BrennanConroy: OrTimeout

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: hush you

anurse: with your brain and your simple mathematics

src/Microsoft.AspNetCore.Sockets.Abstractions/Features/IConnectionHeartbeatFeature.cs
```diff
@@ -0,0 +1,9 @@
+using System;
```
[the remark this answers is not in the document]
anurse: ?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -71,7 +176,43 @@ public virtual void Abort()
...
+                // adding a Ping message when the transport is full is unnecessary since the
+                // transport is still in the process of sending frames.
+                if (_connectionContext.Transport.Writer.TryWrite(_pingMessage))
+                {
+                    _logger.SentPing();
+                }
+                else
+                {
+                    // This isn't necessarily an error, it just indicates that the transport is applying backpressure right now.
+                    _logger.TransportBufferFull();
```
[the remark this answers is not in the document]
anurse: I agree, but better to be prepared.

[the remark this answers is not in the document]
anurse: Since I'm very confident we'll change that

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/IHttpTransport.cs
```diff
@@ -12,9 +12,10 @@ public interface IHttpTransport
         /// <summary>
         /// Executes the transport
         /// </summary>
+        /// <param name="connection"></param>
         /// <param name="context"></param>
         /// <param name="token"></param>
         /// <returns>A <see cref="Task"/> that completes when the transport has finished processing</returns>
-        Task ProcessRequestAsync(HttpContext context, CancellationToken token);
+        Task ProcessRequestAsync(ConnectionContext connection, HttpContext context, CancellationToken token);
```
davidfowl: Undo

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
anurse: bonus

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ChannelExtensions.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
anurse: some of these never had them, good time to update them all :)

src/Microsoft.AspNetCore.Sockets/Internal/SocketEventSource.cs
```diff
@@ -74,5 +75,26 @@ public void ConnectionTimedOut(string connectionId)
...
+        [Event(eventId: 4, Level = EventLevel.Verbose, Message = "Scanning connections.")]
+        public void ScanningConnections()
+        {
+            if(IsEnabled() && IsEnabled(EventLevel.Verbose, EventKeywords.None))
+            {
+                WriteEvent(4);
+            }
+        }
+
+        [Event(eventId: 5, Level = EventLevel.Verbose, Message = "Scanning connections in {0:0.00}ms.")]
```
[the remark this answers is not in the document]
anurse: "Scanning" is also in the wrong tense. I'm rewording.

src/Microsoft.AspNetCore.SignalR.Core/Internal/SignalRCoreLoggerExtensions.cs
```diff
@@ -10,12 +10,6 @@ namespace Microsoft.AspNetCore.SignalR.Core.Internal
     internal static class SignalRCoreLoggerExtensions
     {
         // Category: HubEndPoint<THub>
-        private static readonly Action<ILogger, string, Exception> _usingHubProtocol =
-            LoggerMessage.Define<string>(LogLevel.Information, new EventId(0, nameof(UsingHubProtocol)), "Using HubProtocol '{protocol}'.");
-
-        private static readonly Action<ILogger, Exception> _negotiateCanceled =
-            LoggerMessage.Define(LogLevel.Debug, new EventId(1, nameof(NegotiateCanceled)), "Negotiate was canceled.");
-
         private static readonly Action<ILogger, Exception> _errorProcessingRequest =
```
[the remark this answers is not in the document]
anurse: fine for now, but this is a breaking change so we'll have to remember that when adjusting these post 1.0

davidfowl:
@anurse seems the tests are still failing. Do you know why?

src/Microsoft.AspNetCore.SignalR.Core/Properties/AssemblyInfo.cs
davidfowl: @anurse ping

BrennanConroy:
Not tests, compile error :D

anurse:
\*sigh* fixed

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -141,6 +143,9 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
             {
                 // GET /{path} maps to long polling
 
+                // Initialze Long Polling connection features
+
+
```
davidfowl: Nit remove whitespace

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -212,7 +217,7 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
                     context.Response.RegisterForDispose(timeoutSource);
                     context.Response.RegisterForDispose(tokenSource);
 
-                    var longPolling = new LongPollingTransport(timeoutSource.Token, connection.Application.Reader, connection.ConnectionId, _loggerFactory);
+                    var longPolling = new LongPollingTransport(timeoutSource.Token, options.LongPolling.PollTimeout, connection.Application.Reader, connection.ConnectionId, _loggerFactory);
```
davidfowl: Undo this ctor change

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -16,11 +17,13 @@ public class LongPollingTransport : IHttpTransport
         private readonly ChannelReader<byte[]> _application;
         private readonly ILogger _logger;
         private readonly CancellationToken _timeoutToken;
+        private readonly TimeSpan _timeout;
         private readonly string _connectionId;
 
-        public LongPollingTransport(CancellationToken timeoutToken, ChannelReader<byte[]> application, string connectionId, ILoggerFactory loggerFactory)
+        public LongPollingTransport(CancellationToken timeoutToken, TimeSpan timeout, ChannelReader<byte[]> application, string connectionId, ILoggerFactory loggerFactory)
```
davidfowl: Undo

src/Microsoft.AspNetCore.SignalR.Core/Properties/AssemblyInfo.cs
davidfowl: Remove this commented out code

anurse:
:up: :date:

PR #1165, opened 2017-11-29
Implement #1157 by adding client timeout for C# client

### LINKED ISSUE #1157: [Keep Alive] [C#] Client timeout on idle server
Once #1156 is implemented, the client can implement a "minimum data rate" timeout where if the server hasn't sent a message within a certain window, it will terminate the connection and report that back to the application. This allows applications to detect network failures without relying on OS-specific transport timeouts. This should be configurable, since it depends on how frequently the server is sending Ping messages.

This issue will track **C#** implementation of this behavior. #1162 tracks JavaScript implementation.

Related: #1156

anurse:
Add client timeout to C# client. It is configurable via an option and defaults to 30 seconds (the server ping rate defaults to 15 seconds, so this is effectively 2 server ping intervals).

NOTE: Should not be merged until #1161 is merged since it will immediately break any connection that goes idle for 30 seconds :).

Fixes #1157

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -24,6 +24,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
 {
     public class HubConnection
     {
+        public static readonly TimeSpan DefaultServerTimeout = TimeSpan.FromSeconds(30); // Server ping rate is 15 sec, this is 2 times that.
+
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
         private readonly IConnection _connection;
@@ -38,8 +40,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private int _nextId = 0;
         private volatile bool _startCalled;
+        private Timer _timeoutTimer;
+        private readonly bool _needKeepAlive;
 
         public Task Closed { get; }
+
+        public TimeSpan ServerTimeout { get; set; } = DefaultServerTimeout;
 
         public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFactory loggerFactory)
         {
@@ -59,6 +65,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
             _connection.OnReceived((data, state) => ((HubConnection)state).OnDataReceivedAsync(data), this);
+            _needKeepAlive = _connection.Features.Get<IConnectionInherentKeepAliveFeature>() == null;
+            _timeoutTimer = new Timer(state => ((HubConnection)state).TimeoutElapsed(), this, -1, -1);
             Closed = _connection.Closed.ContinueWith(task =>
             {
                 Shutdown(task.Exception);
@@ -75,6 +83,19 @@ namespace Microsoft.AspNetCore.SignalR.Client
             finally
             {
                 _startCalled = true;
+            }
+        }
+
+        private void TimeoutElapsed()
+        {
+            _connection.AbortAsync(new TimeoutException("Server timeout elapsed without receiving a message from the server."));
+        }
+
+        private void ResetTimeoutTimer()
+        {
+            if (_needKeepAlive)
+            {
+                _timeoutTimer.Change(ServerTimeout, Timeout.InfiniteTimeSpan);
             }
         }
 
@@ -105,6 +126,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 NegotiationProtocol.WriteMessage(new NegotiationMessage(_protocol.Name), memoryStream);
                 await _connection.SendAsync(memoryStream.ToArray(), _connectionActive.Token);
             }
+
+            ResetTimeoutTimer();
         }
 
         private IDataEncoder GetDataEncoder(TransferMode requestedTransferMode, TransferMode actualTransferMode)
@@ -125,6 +148,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private async Task DisposeAsyncCore()
         {
+            _timeoutTimer.Dispose();
             await _connection.DisposeAsync();
             await Closed;
         }
@@ -298,6 +322,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private async Task OnDataReceivedAsync(byte[] data)
         {
+            ResetTimeoutTimer();
             if (_protocolReaderWriter.ReadMessages(data, _binder, out var messages))
             {
                 foreach (var message in messages)
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -45,6 +45,10 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         public Task Closed { get; }
 
+        /// <summary>
+        /// Gets or sets the server timeout interval for the connection. Changes to this value
+        /// are ignored after <see cref="StartAsync"/> is called.
+        /// </summary>
         public TimeSpan ServerTimeout { get; set; } = DefaultServerTimeout;
 
         public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFactory loggerFactory)
@@ -66,12 +70,14 @@ namespace Microsoft.AspNetCore.SignalR.Client
             _logger = _loggerFactory.CreateLogger<HubConnection>();
             _connection.OnReceived((data, state) => ((HubConnection)state).OnDataReceivedAsync(data), this);
             _needKeepAlive = _connection.Features.Get<IConnectionInherentKeepAliveFeature>() == null;
-            _timeoutTimer = new Timer(state => ((HubConnection)state).TimeoutElapsed(), this, -1, -1);
             Closed = _connection.Closed.ContinueWith(task =>
             {
                 Shutdown(task.Exception);
                 return task;
             }).Unwrap();
+
+            // Create the timer for timeout, but disabled by default (we enable it when started).
+            _timeoutTimer = new Timer(state => ((HubConnection)state).TimeoutElapsed(), this, Timeout.Infinite, Timeout.Infinite);
         }
 
         public async Task StartAsync()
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -38,9 +40,17 @@ public class HubConnection
 
         private int _nextId = 0;
         private volatile bool _startCalled;
+        private Timer _timeoutTimer;
+        private readonly bool _needKeepAlive;
 
         public Task Closed { get; }
 
+        /// <summary>
+        /// Gets or sets the server timeout interval for the connection. Changes to this value
```
BrennanConroy: Looks like this variable is used after StartAsync via OnDataReceivedAsync

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -435,11 +435,25 @@ private void SetTransferMode(TransferMode transferMode)
...
-        private async Task DisposeAsyncCore()
+        private async Task DisposeAsyncCore(Exception ex = null)
         {
-            _logger.StoppingClient(_connectionId);
+            if (ex != null)
+            {
+                _logger.AbortingClient(ex, _connectionId);
+
+                // Immediately fault the close task. When the transport shuts down,
+                // it will trigger the close task to be completed, so we want to it be
```
BrennanConroy: "we want to it be" more gramrar

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
```diff
@@ -506,6 +509,14 @@ public static void SendingMessage(this ILogger logger, string connectionId)
             }
         }
 
+        public static void AbortingClient(this ILogger logger, Exception ex, string connectionId)
+        {
+            if (logger.IsEnabled(LogLevel.Information))
```
BrennanConroy: Error error ⚠️

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/SocketClientLoggerExtensions.cs
```diff
@@ -506,6 +509,14 @@ public static void SendingMessage(this ILogger logger, string connectionId)
             }
         }
 
+        public static void AbortingClient(this ILogger logger, Exception ex, string connectionId)
```
BrennanConroy: I prefer Exception and connectionId swtiched

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.AspNetCore.Sockets.Client.Internal;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.AspNetCore.Sockets.Features;
```
BrennanConroy: stro

src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
```diff
@@ -13,6 +13,7 @@ public interface IConnection
         Task StartAsync();
         Task SendAsync(byte[] data, CancellationToken cancellationToken);
         Task DisposeAsync();
+        Task AbortAsync(Exception ex);
```
davidfowl: This is a sign we should unify the client and sever connection objects. This will turn into `ConnectionContext`

src/Microsoft.AspNetCore.Sockets.Client.Http/ITransport.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public interface ITransport
     {
-        Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId);
+        Task StartAsync(Uri url, Channel<byte[], SendMessage> application, TransferMode requestedTransferMode, string connectionId, IConnection connection);
```
davidfowl: This isn't very clean. We shouldn't need to pass the IConnection here.

anurse: How else should transports add features? I actually think it makes sense, it's much like how it works on the server.

src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
anurse: Agreed, but I don't want to block this PR on that right now since I don't need to.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -622,6 +622,56 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                    loggerFactory.CreateLogger<HubConnectionTests>().LogError(ex, "Exception from test");
+                    throw;
+                }
+                finally
+                {
+                    try
+                    {
+                        await hubConnection.DisposeAsync().OrTimeout();
+                    }
+                    catch (TimeoutException) when (transportType != TransportType.LongPolling)
```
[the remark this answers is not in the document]
anurse: DisposeAsync also throws the exception because it re-awaits Closed.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -186,6 +188,67 @@ public class HubConnectionTests
...
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            hubConnection.ServerTimeout = TimeSpan.FromMilliseconds(100);
+
+            await hubConnection.StartAsync();
+
+            try
+            {
+                await connection.ReceiveJsonMessage(new { type = 6 });
```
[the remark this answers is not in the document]
anurse: No, but there are constants... It's internal but I suppose we could make them pubternal.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -78,6 +92,20 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
             }
         }
 
+        private void TimeoutElapsed()
+        {
+            _connection.AbortAsync(new TimeoutException("Server timeout elapsed without receiving a message from the server."));
```
muratg: Would it be useful to call out what the timeout was in the exception message?

src/Microsoft.AspNetCore.Sockets.Abstractions/IConnection.cs
muratg: https://github.com/aspnet/SignalR/issues/1177 to track that

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Most other systems don't. We could, I suppose.

anurse: A lot of those are simply converting error codes to strings though so they don't have the timeout value :). I should be able to add that in easily.

muratg: If this is not common elsewhere, let's not complicate this more then. :)

anurse: Meh, it's easy and done. It'd be more work to remove it now :P

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -185,6 +187,67 @@ public class HubConnectionTests
...
+
+        [Fact(Skip = "Not really able to do this reliably")]
+        public async Task ConnectionNotTerminatedIfServerKeepsSending()
+        {
+            var connection = new TestConnection();
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            hubConnection.ServerTimeout = TimeSpan.FromMilliseconds(100);
+
+            await hubConnection.StartAsync();
```
BrennanConroy: OrTimeout

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -185,6 +187,67 @@ public class HubConnectionTests
...
+
+            try
+            {
+                await connection.ReceiveJsonMessage(new { type = HubProtocolConstants.PingMessageType });
+                await Task.Delay(50);
```
BrennanConroy: OrTimeout :trollface:

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -185,6 +187,67 @@ public class HubConnectionTests
...
+        [Fact]
+        public async Task ServerTimeoutIgnoredIfTransportHasInherentKeepAlive()
+        {
+            var connection = new TestConnection();
+            connection.Features.Set<IConnectionInherentKeepAliveFeature>(new ConnectionInherentKeepAliveFeature(TimeSpan.FromSeconds(1)));
+            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+
+            hubConnection.ServerTimeout = TimeSpan.FromMilliseconds(100);
+
+            await hubConnection.StartAsync();
```
BrennanConroy: mOrTimeout

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -64,6 +74,9 @@ public HubConnection(IConnection connection, IHubProtocol protocol, ILoggerFacto
                 Shutdown(task.Exception);
                 return task;
             }).Unwrap();
+
+            // Create the timer for timeout, but disabled by default (we enable it when started).
+            _timeoutTimer = new Timer(state => ((HubConnection)state).TimeoutElapsed(), this, Timeout.Infinite, Timeout.Infinite);
```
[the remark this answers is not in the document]
anurse: yep

PR #1178, opened 2017-12-04
Use aspnetcore-dev

ryanbrandenburg:
Looks like this repo got missed when @pranavkm  was updating repos to use `aspnetcore-dev`.


pranavkm:
Nice catch. Must've missed this when I was updating.

PR #1182, opened 2017-12-05
Change log names in testing and assembly logs are from

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -28,7 +28,7 @@ public class ServerFixture<TStartup> : IDisposable
 
         public ServerFixture()
         {
-            var testLog = AssemblyTestLog.ForAssembly(typeof(ServerFixture<TStartup>).Assembly);
+            var testLog = AssemblyTestLog.ForAssembly(typeof(TStartup).Assembly);
```
BrennanConroy: This is the fix for "file locked" errors on CI

Issue #944, opened 2017-09-25
Expose IHttpConnectionFeature Properties 


zAfLu:
Hello my friends,

I'm currently struggling a bit with "how to get the caller ip". I thought the connection information’s where In _hub.Context.Connection_ but I cannot find any IPs in the whole Hub class. So maybe you can give me a hint :)

davidfowl:
We need to hoist this to a first class feature on the hub.Context.Connection but this should work:

```
Context.Connection.GetHttpContext().Connection.LocalIpAddress
```

zAfLu:
oh well ... i think i need to open my eyes next time. thanks mate :)
And yes hosting this as a first class feature would be nice.

mikaelm12:
Done in: 40e906f23a311d745230b8a3b48c919b25231f7c

PR #1186, opened 2017-12-07
Making CallerIP a first class feature on the HubConnectionContext

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/944

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -73,7 +73,7 @@ namespace Microsoft.AspNetCore.SignalR
         // Currently used only for streaming methods
         internal ConcurrentDictionary<string, CancellationTokenSource> ActiveRequestCancellationSources { get; } = new ConcurrentDictionary<string, CancellationTokenSource>();
 
-        public IPAddress ClientIP => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.RemoteIpAddress;
+        public IPAddress RemoteIpAddress => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.RemoteIpAddress;
 
         public IPAddress LocalIpAddress => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.LocalIpAddress;
 
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -73,7 +73,13 @@ namespace Microsoft.AspNetCore.SignalR
         // Currently used only for streaming methods
         internal ConcurrentDictionary<string, CancellationTokenSource> ActiveRequestCancellationSources { get; } = new ConcurrentDictionary<string, CancellationTokenSource>();
 
-        public ConnectionInfo HttpConnection => Features.Get<IHttpContextFeature>()?.HttpContext.Connection;
+        public IPAddress ClientIP => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.RemoteIpAddress;
+
+        public IPAddress LocalIpAddress => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.LocalIpAddress;
+
+        public int? RemotePort => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.RemotePort;
+
+        public int? LocalPort => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.LocalPort;
 
         public async Task WriteAsync(HubInvocationMessage message)
         {
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Net;
 using System.Runtime.ExceptionServices;
 using System.Security.Claims;
 using System.Threading;
@@ -18,7 +19,9 @@ using Microsoft.AspNetCore.SignalR.Internal.Encoders;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.AspNetCore.Sockets;
 using Microsoft.AspNetCore.Sockets.Features;
+using Microsoft.AspNetCore.Sockets.Http.Features;
 using Microsoft.Extensions.Logging;
+using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.SignalR
 {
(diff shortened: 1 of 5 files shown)


anurse:
I think we should actually hoist the whole `IHttpConnectionFeature` value up. The values are all relevant: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Features/IHttpConnectionFeature.cs

Let's hoist it to a property named `Connection`. So the pattern to get the ClientIP would be `Context.Connection.RemoteIPAddress`.

mikaelm12:
So the HubConnectionContext already lives in a property called Connection on the hub context. Should we call the IHttpContext HttpContext then..
Making getting the ClientIP now look like ```Context.Connection.HttpConnection.RemoteIPAddress```

anurse:
We shouldn't use `HttpConnection` because this data would also apply for a raw TCP connection (despite the feature's name, it's actually not HTTP specific)

How about we hoist all the properties then, rather than just Client IP (with the **exception** of `IHttpConnectionFeature.ConnectionId`, since that's actually a **different** id from the one we already have and would be confusing).

Also, let's use the names as they are in `IHttpConnectionFeature` rather than `ClientIP` (which is technically inaccurate since we don't know that it's the actual Client's IP address; if the client is going through a proxy it would be the proxy's address we see here)

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -70,6 +73,14 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
         // Currently used only for streaming methods
         internal ConcurrentDictionary<string, CancellationTokenSource> ActiveRequestCancellationSources { get; } = new ConcurrentDictionary<string, CancellationTokenSource>();
 
+        public IPAddress ClientIP => Features.Get<IHttpContextFeature>()?.HttpContext.Connection.RemoteIpAddress;
```
anurse: Use the name `RemoteIpAddress`

anurse:
I was thinking about tests but these are simply pass-through properties, so they wouldn't really do much. Maybe bash out a quick test that sets the feature and makes sure the properties are passed through correctly?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
mikaelm12: Oops

mikaelm12:
Added a test 😄

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
```diff
@@ -29,6 +29,26 @@ public class TestHub : Hub
             await Clients.Client(Context.ConnectionId).InvokeAsync("Echo", message);
         }
 
+        public int CheckLocalPort()
```
anurse: Use `Get` as in `GetLocalPort`. It fits with the other patterns (for headers/cookies, etc.)

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -31,7 +31,7 @@ public ServerFixture()
             var testLog = AssemblyTestLog.ForAssembly(typeof(TStartup).Assembly);
             _logToken = testLog.StartTestLog(null, $"{nameof(ServerFixture<TStartup>)}_{typeof(TStartup).Name}", out _loggerFactory, "ServerFixture");
             _logger = _loggerFactory.CreateLogger<ServerFixture<TStartup>>();
-            Url = "http://localhost:" + GetNextPort();
+            Url = "http://127.0.0.1:" + GetNextPort();
```
anurse: Put a comment here explaining why (because we want to ensure the RemoteIPAddress value is consistent)

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -31,6 +31,7 @@ public ServerFixture()
             var testLog = AssemblyTestLog.ForAssembly(typeof(TStartup).Assembly);
             _logToken = testLog.StartTestLog(null, $"{nameof(ServerFixture<TStartup>)}_{typeof(TStartup).Name}", out _loggerFactory, "ServerFixture");
             _logger = _loggerFactory.CreateLogger<ServerFixture<TStartup>>();
+            // Were using 127.0.0.1 instead of localhost to ensure that we use IPV4 across different OSes
```
anurse: µ-nit: "we're" :trollface:

mikaelm12: I'm disappointed in myself

anurse:
Filed https://github.com/aspnet/SignalR/issues/1214 for the flaky test. I'll try to look at it tomorrow.

src/Microsoft.AspNetCore.SignalR.Core/Microsoft.AspNetCore.SignalR.Core.csproj
```diff
@@ -9,6 +9,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Sockets.Abstractions\Microsoft.AspNetCore.Sockets.Abstractions.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.SignalR.Common\Microsoft.AspNetCore.SignalR.Common.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Sockets.Http\Microsoft.AspNetCore.Sockets.Http.csproj" />
```
davidfowl: There's a layering violation here. This entire feature shouldn't depend on http*. We should expose another feature for this.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
```diff
@@ -29,6 +29,26 @@ public class TestHub : Hub
             await Clients.Client(Context.ConnectionId).InvokeAsync("Echo", message);
         }
 
+        public int GetLocalPort()
```
davidfowl: A single method call with all of the data would be sufficient.

PS: This hub is becoming too large and we need to write more unit tests.

src/Microsoft.AspNetCore.SignalR.Core/Microsoft.AspNetCore.SignalR.Core.csproj
anurse: The problem is that right now we're re-using the `IHttpConnectionFeature`. We could introduce a new Connection Information feature I suppose.

davidfowl: We can use that feature without adding a direct dependency to Sockets.Http though. The IFeatureCollection dependency is already part of abstractions (https://github.com/aspnet/SignalR/blob/d2c1138429f491bb013b7e45bc7492676e18d67e/src/Microsoft.AspNetCore.Sockets.Abstractions/Microsoft.AspNetCore.Sockets.Abstractions.csproj#L10) which for better or worse, already means we have a dependency on these features https://github.com/aspnet/HttpAbstractions/blob/4fdeeb31311bb8cca8b62ed9ea4407c10b30518c/src/Microsoft.AspNetCore.Http.Features/IHttpConnectionFeature.cs. I'd suggesting looking for that specific feature in the connection feature collection instead of going through the HttpContext.

anurse: Yep, came to the same conclusions here. I think we should be OK with using Microsoft.AspNetCore.Http.Features in SignalR.Core. The problem was more about Sockets.Http.

mikaelm12:
This needs some discussion so I'm closing in the mean time

Issue #734, opened 2017-08-17
IHubClients Properties


mikaelm12:
Pulled from Big Rocks into its own issue. 

- [x] `IHubClients` missing properties (https://github.com/SignalR/SignalR/blob/b73cbabe69bc48c4cb8d768148e157712a264ce9/src/Microsoft.AspNet.SignalR.Core/Hubs/IHubConnectionContext.cs)
   -  [x] Caller (Hub only) **preview1**
   -  [x] Others **preview1** 
   -  [x] OthersInGroup **preview1** 
   -  [x] AllExcept
   -  [x] Groups 
    - [x] Clients
    - [x] Users

beastlike:
is there another way to get back to just the caller in the interim? 

examples I see from .net 4.6 use:
Clients.Caller.InvokeAsync("Send", message);

however all .net core examples I've seen using .net core signalr use:
Clients.All.InvokeAsync("Send", message);

Any workarounds for now? Thanks!

BrennanConroy:
`Clients.Client(Context.ConnectionId).InvokeAsync(...)`

beastlike:
Awesome thanks!

muratg:
For Others, see comment: https://github.com/aspnet/SignalR/pull/787#discussion_r135797550

BrennanConroy:
@muratg I think this should be preview1

mikaelm12:
Yeah, it is under preview one in the big rocks issue https://github.com/aspnet/SignalR/issues/394

muratg:
Right, brought it back, thanks!

mikaelm12:
All these are done ✔️

PR #1192, opened 2017-12-08
Support for Others

mikaelm12:
Want some inital feedback for this features. 
 - Needs better names 
 - Still need to add tests
 - Will add ```OthersInGroup``` later
 - Do we still want ```OthersInGroups```?

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
@@ -1,6 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System.Collections.Generic;
-using System.Text;
 
 namespace Microsoft.AspNetCore.SignalR
 {
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubClientsWrapperThing : IHubCallerClients
+    {
+        private string _connectionId;
+        private IHubClients _hubClients;
+
+        public HubClientsWrapperThing(IHubClients hubClients, string ConnectionId)
+        {
+            _connectionId = ConnectionId;
+            _hubClients = hubClients;
+        }
+        public IHubClients HubClients => _hubClients;
+
+        public IClientProxy Caller => _hubClients.Client(_connectionId);
+
+        public IClientProxy Others => _hubClients.AllExcept(new List<string> { _connectionId });
+
+        public IClientProxy All => _hubClients.All;
+
+        public IClientProxy AllExcept(IReadOnlyList<string> excludedIds)
+        {
+            return _hubClients.AllExcept(excludedIds);
+        }
+
+        public IClientProxy Client(string connectionId)
+        {
+            return _hubClients.Client(connectionId);
+        }
+
+        public IClientProxy Group(string groupName)
+        {
+            return _hubClients.Group(groupName);
+        }
+
+        public IClientProxy User(string userId)
+        {
+           return _hubClients.User(userId);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.SignalR.Core/IHubClients.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/IHubClients.cs
@@ -6,4 +6,6 @@ using System.Collections.Generic;
 namespace Microsoft.AspNetCore.SignalR
 {
     public interface IHubClients : IHubClients<IClientProxy> { }
+
+    public interface IHubCallerClients : IHubCallerClients<IClientProxy> { }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Core/IHubClients.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/IHubClients.cs
@@ -6,6 +6,4 @@ using System.Collections.Generic;
 namespace Microsoft.AspNetCore.SignalR
 {
     public interface IHubClients : IHubClients<IClientProxy> { }
-
-    public interface IHubCallerClients : IHubCallerClients<IClientProxy> { }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Core/IHubClients`T.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/IHubClients`T.cs
@@ -19,4 +19,12 @@ namespace Microsoft.AspNetCore.SignalR
 
         T User(string userId);
     }
+
+    public interface IHubCallerClients<T> : IHubClients<T>
+    {
+        T Caller { get; }
+
+        T Others { get; }
+    }
 }
+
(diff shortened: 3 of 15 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubClientsWrapperThing : IHubCallerClients
```
mikaelm12: Naming suggestions?

src/Microsoft.AspNetCore.SignalR.Core/IHubClients.cs
```diff
@@ -6,4 +6,6 @@
 namespace Microsoft.AspNetCore.SignalR
 {
     public interface IHubClients : IHubClients<IClientProxy> { }
+
+    public interface IHubCallerClients : IHubCallerClients<IClientProxy> { }
```
mikaelm12: Naming suggestions?

src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
BrennanConroy: `HubClientsWrapperThingy`

anurse: Just `HubCallerClients` would be fine actually.

src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        private string _connectionId;
+        private IHubClients _hubClients;
+
+        public HubClientsWrapperThing(IHubClients hubClients, string ConnectionId)
```
anurse: `connectionId` (camelCase)

src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            _connectionId = ConnectionId;
+            _hubClients = hubClients;
+        }
+        public IHubClients HubClients => _hubClients;
+
+        public IClientProxy Caller => _hubClients.Client(_connectionId);
+
+        public IClientProxy Others => _hubClients.AllExcept(new List<string> { _connectionId });
```
anurse: Initialize this list in the constructor so we don't create it every time you call `Others`.

anurse: Also, `string[]` implements `IReadOnlyList<string>`, so just use an array.

src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        private string _connectionId;
+        private IHubClients _hubClients;
+
+        public HubClientsWrapperThing(IHubClients hubClients, string ConnectionId)
+        {
+            _connectionId = ConnectionId;
+            _hubClients = hubClients;
+        }
+        public IHubClients HubClients => _hubClients;
```
anurse: No need to make this public, or even have it at all I think

src/Microsoft.AspNetCore.SignalR.Core/IHubClients`T.cs
```diff
@@ -19,4 +19,12 @@ public interface IHubClients<T>
 
         T User(string userId);
     }
+
+    public interface IHubCallerClients<T> : IHubClients<T>
+    {
+        T Caller { get; }
+
+        T Others { get; }
```
anurse: Do we have an issue tracking `OthersInGroup`? It's fine to not do that in this PR but it should be tracked :)

src/Microsoft.AspNetCore.SignalR.Core/HubClientsWrapperThing.cs
anurse: Maybe this should be pubternal? I could maybe see reasons for users to want to create custom ones (when using HubContext directly for example) but I don't think we want to commit this to public API.

src/Microsoft.AspNetCore.SignalR.Core/IHubClients`T.cs
mikaelm12: Yup https://github.com/aspnet/SignalR/issues/734

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -826,6 +826,68 @@ public void HubsCannotHaveOverloadedMethods()
...
+                Assert.Equal("To others", invocation.Arguments[0]);
+
+                var firstClientResult = await firstClient.ReadAsync().OrTimeout();
+                var completionMessage = Assert.IsType<CompletionMessage>(firstClientResult);
+
+                // kill the connections
+                firstClient.Dispose();
+                secondClient.Dispose();
+
+                await Task.WhenAll(firstEndPointTask, secondEndPointTask).OrTimeout();
```
anurse: My only minor concern is that a message could have been sent to `firstClient` and the test wouldn't fail, as far as I can tell. We should check that `.TryRead` returns false (or only returns Ping messages) **after** the endpoint tasks have concluded.

mikaelm12: I broadcast again after the others invocation and assert that that message is the first invocation message to the first client

anurse: Ah, I see that now. Looks good then, thanks for clarifying.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -826,6 +826,76 @@ public void HubsCannotHaveOverloadedMethods()
...
+                Task firstEndPointTask = endPoint.OnConnectedAsync(firstClient.Connection);
+
+                await firstClient.Connected;
+
+                await firstClient.SendInvocationAsync("SendToCaller", "To caller").OrTimeout();
+
+                var firstClientResult = await firstClient.ReadAsync().OrTimeout();
+                var invocation = Assert.IsType<InvocationMessage>(firstClientResult);
+                Assert.Equal("Send", invocation.Target);
+                Assert.Equal("To caller", invocation.Arguments[0]);
```
anurse: Should do a similar trick as in the previous test to ensure that this isn't just being Broadcast and passing because you only connect a single client. Connect another client and assert that the other client doesn't receive the message.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -826,6 +826,84 @@ public void HubsCannotHaveOverloadedMethods()
...
+
+                await firstClient.SendInvocationAsync("BroadcastMethod", "To everyone").OrTimeout();
+                var secondClientResult = await secondClient.ReadAsync().OrTimeout();
+                invocation = Assert.IsType<InvocationMessage>(secondClientResult);
+                Assert.Equal("Broadcast", invocation.Target);
+                Assert.Equal("To everyone", invocation.Arguments[0]);
```
anurse: :+1: thanks!

PR #1204, opened 2017-12-12
AllExcept for Groups - GroupExcept

mikaelm12:
This is in preparation for `OthersInGroup`. `GroupExcept` is analogous to `AllExcept `so that we can make `OthersInGroup `work similarly to Others
Still working on this but was looking for early feedback. Hanging some tests btw... any insight there would be appreciated. 
Haven't added new tests for this feature yet as well.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -129,6 +129,25 @@ namespace Microsoft.AspNetCore.SignalR
             return Task.CompletedTask;
         }
 
+        public override Task InvokeGroupExceptAsync(string groupName, string methodName, object[] args, IReadOnlyList<string> excludedIds)
+        {
+            if (groupName == null)
+            {
+                throw new ArgumentNullException(nameof(groupName));
+            }
+
+            var group = _groups[groupName];
+            if (group != null)
+            {
+                var message = CreateInvocationMessage(methodName, args);
+                var tasks = group.Values.Where(connection => !excludedIds.Contains(connection.ConnectionId))
+                    .Select(c => c.WriteAsync(message));
+                return Task.WhenAll(tasks);
+            }
+
+            return Task.CompletedTask;
+        }
+
         private InvocationMessage CreateInvocationMessage(string methodName, object[] args)
         {
             return new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, argumentBindingException: null, arguments: args);
--- a/src/Microsoft.AspNetCore.SignalR.Core/Proxies.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Proxies.cs
@@ -40,6 +40,25 @@ namespace Microsoft.AspNetCore.SignalR
         }
     }
 
+    public class GroupExceptProxy<THub> : IClientProxy
+    {
+        private readonly string _groupName;
+        private readonly HubLifetimeManager<THub> _lifetimeManager;
+        private IReadOnlyList<string> _excludedIds;
+
+        public GroupExceptProxy(HubLifetimeManager<THub> lifetimeManager, string groupName, IReadOnlyList<string> excludedIds)
+        {
+            _lifetimeManager = lifetimeManager;
+            _groupName = groupName;
+            _excludedIds = excludedIds;
+        }
+
+        public Task InvokeAsync(string method, params object[] args)
+        {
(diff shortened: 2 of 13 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -190,7 +190,19 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
...
+        }
+
+        public override Task InvokeGroupExceptAsync(string groupName, string methodName, object[] args, IReadOnlyList<string> excludedIds)
+        {
+            if (groupName == null)
+            {
+                throw new ArgumentNullException(nameof(groupName));
+            }
+
+            var message = new RedisExcludeClientsMessage(GetInvocationId(), nonBlocking: true, target: methodName, excludedIds: excludedIds, arguments: args);
```
anurse: I like this. It uses an existing message and we just publish it to a different Redis key. 👍

anurse:
> Hanging some tests btw... any insight there would be appreciated

Story of my week ;). We need to get a handle on hanging and flaky tests. There are a couple of options to help diagnose them. Once #1147 is merged we'll get long-running test notifications which will tell you which tests are still running after 5 seconds.

mikaelm12:
Fixed the hang so I need to add test now 😄

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -129,6 +129,25 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
...
+            }
+
+            var group = _groups[groupName];
+            if (group != null)
+            {
+                var message = CreateInvocationMessage(methodName, args);
+                var tasks = group.Values.Where(connection => !excludedIds.Contains(connection.ConnectionId))
+                    .Select(c => c.WriteAsync(message));
+                return Task.WhenAll(tasks);
+            }
```
muratg: Is this a reasonable place to add logging or should we first finalize our logging strategy? @BrennanConroy @anurse

davidfowl:
I have a growing concern that we should look at addressing and it's about the explosion of methods on the HubLifetimeManager as a result of these slices of connections. I want us to investigate a named filter concept to see if we can avoid some of the explosion.

src/Microsoft.AspNetCore.SignalR.Core/Proxies.cs
```diff
@@ -40,6 +40,25 @@ public Task InvokeAsync(string method, params object[] args)
         }
     }
 
+    public class GroupExceptProxy<THub> : IClientProxy
+    {
+        private readonly string _groupName;
+        private readonly HubLifetimeManager<THub> _lifetimeManager;
+        private IReadOnlyList<string> _excludedIds;
```
BrennanConroy: readonly

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -102,6 +103,37 @@ public class RedisHubLifetimeManagerTests
...
+                var connection1 = HubConnectionContextUtils.Create(client1.Connection);
+                var connection2 = HubConnectionContextUtils.Create(client2.Connection);
+
+                await manager.OnConnectedAsync(connection1).OrTimeout();
+                await manager.OnConnectedAsync(connection2).OrTimeout();
+
+                await manager.AddGroupAsync(connection1.ConnectionId, "gunit").OrTimeout();
+                await manager.AddGroupAsync(connection2.ConnectionId, "gunit").OrTimeout();
+
+                var excludedIds = new List<string>{client2.Connection.ConnectionId };
```
BrennanConroy: nit: space after `{`

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -102,6 +103,37 @@ public class RedisHubLifetimeManagerTests
...
+                await manager.InvokeGroupExceptAsync("gunit", "Hello", new object[] { "World" }, excludedIds).OrTimeout();
+
+                await AssertMessageAsync(client1);
+
+                await connection1.DisposeAsync().OrTimeout();
+                await connection2.DisposeAsync().OrTimeout();
+                Assert.Null(client2.TryRead());
```
BrennanConroy: Do this before disposing

PR #1208, opened 2017-12-12
clean up HttpConnectionTests

anurse:
No changes to product code at all, just to test code.

* There was a lot of duplication in these tests, so I introduced a few helpers (mostly just the ones I had already done with the AbortAsync tests in #1147)
* The `try { ... } finally { DisposeAsync() }` pattern gets noisy fast so I added `WithConnectionAsync()` which runs a provided delegate in the body of that try-finally pattern
* Lots of Moq usage that just ends up doing the same thing. So I introduced `TestTransport` and `TestHttpMessageHandler` code to remove most of the duplication.
* Lots of TCSes everywhere. Many of them are doing the job that the `SyncPoint` helper I introduced in #1147 does (having a pair of TCSes to allow test code to wait for code-under-test to reach a certain point and then release it to continue after a condition is met). So I refactored to use that.
* One big giant test file was getting unwieldy so I split it into other files and used nested classes to group tests to allow you to easily run each group separately.
* A couple tests were a bit weird and got refactored/renamed:
  * `EventsAreNotRunningOnMainLoop` was depending upon a race. It was testing that when `TryWrite` on the channel was called, a `TryRead` immediately following would fail because the `HttpConnection` receive loop already read the message. This depends on the way inline-continuations work in Channels and may break with other configurations. The refactored test waits for `WaitToReadAsync` to block, which should be more reliable
  * `EventQueueTimeoutWithException` just seemed to be checking that exceptions in OnReceive are handled, which other tests already do. So I just removed it.


davidfowl:
wtf is a SyncPoint...

anurse:
Read the code and find out. It's a pair of TCSes that replace a bunch of repeated code in our tests. `WaitToContinue` triggers the TCS attached to `WaitForSyncPoint` and `Continue` triggers the TCS attached to `WaitToContinue`. We were implementing this with pairs of TCSes in each test and it was super noisy.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Helpers.cs
```diff
@@ -0,0 +1,128 @@
+using System;
```
muratg: Nit: (c) notice.   </TroyMode>

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Negotiate.cs
```diff
@@ -0,0 +1,91 @@
+using System;
```
muratg: (c)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -0,0 +1,366 @@
+using System;
```
BrennanConroy: copy

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.OnReceived.cs
```diff
@@ -0,0 +1,104 @@
+using System;
```
BrennanConroy: copy

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.SendAsync.cs
```diff
@@ -0,0 +1,119 @@
+using System;
```
BrennanConroy: copy

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestHttpMessageHandler.cs
```diff
@@ -9,17 +9,128 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
...
+        public static HttpMessageHandler CreateDefault()
+        {
+            var testHttpMessageHandler = new TestHttpMessageHandler();
+
+            testHttpMessageHandler.OnSocketSend((_, __) => ResponseUtils.CreateResponse(HttpStatusCode.Accepted));
+            testHttpMessageHandler.OnLongPoll(async cancellationToken =>
             {
-                return Task.FromException<HttpResponseMessage>(new InvalidOperationException($"Http endpoint not implemented: {request.RequestUri}"));
-            }
+                // Just block until cancelled
```
BrennanConroy: canceled

anurse:
I'm going to take a look at some of the test flakiness after this. I don't like how hard it is to get the checks to go green.

PR #1218, opened 2017-12-14
Fix #1170 by removing invocationId from non-blocking calls

### LINKED ISSUE #1170: Review Invocation ID format
We want these to be per connection, but serialization could be tricky. 

cc @anurse

anurse:
When talking about #1170 we were talking about how it's going to be a performance problem that we serialize Invocations once for each connection involved in a broadcast. However, if we serialize once, we have to update Invocation IDs as they are supposed to be local to the connection. That means that to serialize once we have to use globally-unique IDs (GUIDs).

However, the messages sent from Server to Client are always non-blocking. In a non-blocking message, the sender **never** expects to receive a response. This means that there's no reason for an invocation ID at all. So this PR changes how non-blocking messages are rendered. Instead of using a boolean flag, the presence of `invocationId` will indicate if a message is "blocking" and expects a response. This is similar to how [JSON-RPC notifications](http://www.jsonrpc.org/specification#notification) work.

In JSON this is represented by simply omitting the `invocationId` parameter. In MsgPack, the `invocationId` value is written as a MsgPack `nil` value. Only the Invocation message is affected as StreamInvocations are always blocking by design.

Spec update is coming shortly in a commit to this PR.


davidfowl:
Lets also get rid of the blocking/non-blocking nomenclature in the spec, it's extremely confusing to talk about in the same breath as async (which all of the methods are). We might consider using something related to ACKs (ACKed invocations) or something similar.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -566,12 +566,12 @@ public class HubEndpointTests
...
                 await client.SendInvocationAsync(methodName, nonBlocking: true).OrTimeout();
 
-                // Nothing should have been written
-                Assert.False(client.Application.Reader.TryRead(out var buffer));
-
                 // kill the connection
                 client.Dispose();
 
+                // Nothing should have been written
+                Assert.False(client.Application.Reader.TryRead(out var buffer));
```
davidfowl: Why this change?

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/HubProtocolBenchmark.cs
```diff
@@ -36,16 +36,16 @@ public void GlobalSetup()
             switch (Input)
             {
                 case Message.NoArguments:
-                    _hubMessage = new InvocationMessage("123", true, "Target", null);
+                    _hubMessage = new InvocationMessage(invocationId: null, target: "Target", argumentBindingException: null);
```
davidfowl: Why not add an ctor overload of InvocationMessage that doesn't take the invocationId?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMethodInvocationMessage.cs
```diff
@@ -63,23 +56,27 @@ public HubMethodInvocationMessage(string invocationId, string target, ExceptionD
 
     public class InvocationMessage : HubMethodInvocationMessage
```
davidfowl: Add a constructor that doesn't take invocationId

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -77,18 +77,25 @@ private static HubMessage ParseMessage(Stream input, IInvocationBinder binder)
         private static InvocationMessage CreateInvocationMessage(Unpacker unpacker, IInvocationBinder binder)
         {
             var invocationId = ReadInvocationId(unpacker);
-            var nonBlocking = ReadBoolean(unpacker, "nonBlocking");
+
+            // For MsgPack, we represent an empty invocation ID as an empty string,
+            // so we need to normalize that to "null", which is what indicates a non-blocking invocation.
+            if(string.IsNullOrEmpty(invocationId))
```
davidfowl: nit: f touching ( 😠

anurse:
> Lets also get rid of the blocking/non-blocking nomenclature in the spec

Got distracted but I do plan to update the spec in this PR before final approval/merging.

anurse:
Updated the spec. I also removed the ProtoBuf stuff rather than updating it because there is no plan to have an official ProtoBuf encoding in v1. We can easily bring it back from history (and it wasn't exactly complex)

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: It was never actually asserting that nothing was sent.

anurse: As a test, I commented out https://github.com/aspnet/SignalR/blob/1a842e92eccd1a54104b0e723abcc505dd20201b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs#L361 and the test still passed despite the fact that we now erroneously sent a Completion Message. This check makes it fail (as it should) when we send a response to the non-blocking invocation.

anurse:
Unfortunately, JSON.NET gets grumpy when you deserialize objects with read-only properties and more than one constructor. Adding an overloaded ctor to `InvocationMessage` that doesn't take an `invocationId` caused this problem in the Redis lifetime manager. So we have to do one of:

* Mark the ctor that contains `invocationId` with the `[JsonConstructor]` attribute (downside: couples to JSON.NET attributes)
* Remove the extra ctor (downside: lots of `invocationId: null` when you want to create non-blocking Invocation Message objects)
* Make the properties read-write (downside: the type really doesn't need to be mutable; and I think we'd also have to add a default ctor to allow JSON.NET to do it's thing)
* Use a different serialization technique in redis (surrogate types, JsonWriter, etc.)

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -150,7 +150,7 @@ public override Task InvokeGroupExceptAsync(string groupName, string methodName,
 
         private InvocationMessage CreateInvocationMessage(string methodName, object[] args)
         {
-            return new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, argumentBindingException: null, arguments: args);
+            return new InvocationMessage(null, target: methodName, argumentBindingException: null, arguments: args);
```
davidfowl: Why do you need to pass null? Also can you remove the GetInvocationId method?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -170,7 +170,7 @@ public override Task InvokeConnectionAsync(string connectionId, string methodNam
                 throw new ArgumentNullException(nameof(connectionId));
             }
 
-            var message = new InvocationMessage(GetInvocationId(), nonBlocking: true, target: methodName, argumentBindingException: null, arguments: args);
+            var message = new InvocationMessage(null, target: methodName, argumentBindingException: null, arguments: args);
```
davidfowl: Why pass null?

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
anurse: Just missed updating this. Though the above comment means we need to rethink the plan for this a little.

anurse:
Decided to take the last option above. We didn't really need to be serializing full HubMessage objects anyway.

PR #1219, opened 2017-12-14
[WIP] Deflaking Tests

anurse:
It's only flaky in travis so I'm opening a PR to play around with some things.


davidfowl:
Why is travis failing on every PR

anurse:
That's the point of this PR, start reducing that :). Hopefully there are useful logs, but I bet this failure is a different test than I've been chasing (just because flaky tests are a pain)

anurse:
FYI, there's no change here yet, this is just logging and using a PR to trigger Travis so I can see and debug the flaky test.

anurse:
Think I found a way to resolve some of the flakiness around WebSockets while also finding a product bug. We didn't handle the close timeout very well and relied on `Task.Delay` to time us out. I think `Task.Delay` is less reliable than `CancellationTokenSource.CancelAfter` after looking at the code a little. Might look at applying this elsewhere.

mikaelm12:
@anurse Gonna go ahead and close this for you

PR #1222, opened 2017-12-15
WIP: Allocate the HubCallerContext and HubCallerClients per connection

davidfowl:
- It's actually only less overall allocations if we're doing invocations. If not then we have
2 extra long lived objects.
- I also don't like keeping this internal but I don't want it to be exposed in the hub


anurse:
Could these just be structs? Aren't they attached to the hub and simple wrapper around existing reference types?

davidfowl:
Possibly, I’ll look into that. I was looking at reducing the array allocation for connection ids for “others”

anurse:
Yeah, that would be harder to avoid

PR #1230, opened 2017-12-18
Others in group

mikaelm12:
The name says it all. (Now based on dev)


src/Microsoft.AspNetCore.SignalR.Core/IHubCallerClients`T.cs
```diff
@@ -1,12 +1,16 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
```
anurse: Unintentional?

mikaelm12: Oops, yeah. Good catch

PR #1239, opened 2017-12-20
Initial typescript tidy up

anurse:
This is a first pass at some of the tidy up items:

* HttpConnection (and options)
  * jwt bearer rename -> bearer token (got C# while I was at it)
  * ~~onreceive and onclose should be methods and camelCase~~ see note below
* HubConnection (and options)
  * logging rename -> logger
  * Ctor needs more overloads
  * stop should be a promise.
  * ~~camelCase onclose~~ see note below
* Transports
  * ~~onclose and onreceive should be camelCase~~ see note below
  * Stop should be a promise

NOTE: We were going to camelCase the events. However, the events on `WebSocket` and `EventSource` are **not** camelCased (see https://developer.mozilla.org/en-US/docs/Web/API/WebSocket for example). Given that, I've changed my mind a bit and think it would be more consistent for us to not camelCase them. Open to discussion if people disagree, but either way we should just pick a style and stick with it.

Going through the TypeScript code, I have some larger concerns that we may need to address prior to Preview 1. The testing is very minimal and relies heavily on specific errors occurring to indicate that things are "working" because we don't have sufficient abstractions for WebSocket, EventSource or XMLHttpRequest in the transports. I'm going to work on some more significant changes when I've got these small tidy-up items done.


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Transports.ts
```diff
@@ -103,14 +104,14 @@ export class WebSocketTransport implements ITransport {
 
 export class ServerSentEventsTransport implements ITransport {
     private readonly httpClient: IHttpClient;
-    private readonly jwtBearer: () => string;
+    private readonly accessToken: () => string;
```
muratg: :+1:

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
```diff
@@ -56,7 +56,7 @@ describe("HubConnection", () => {
         it("sends an invocation", async () => {
             let connection = new TestConnection();
 
-            let hubConnection = new HubConnection(connection);
+            let hubConnection = new HubConnection(connection, { logger: null });
```
mikaelm12: Why do we bother passing in the null logger? We don't need a HubOptionsObject here, right?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -34,10 +33,12 @@ export class HubConnection {
     private timeoutHandle: NodeJS.Timer;
     private serverTimeoutInMilliseconds: number;
 
+    constructor(url: string, options?: IHubConnectionOptions);
+    constructor(connection: IConnection, options?: IHubConnectionOptions);
     constructor(urlOrConnection: string | IConnection, options: IHubConnectionOptions = {}) {
         options = options || {};
 
-        this.serverTimeoutInMilliseconds = options.serverTimeoutInMilliseconds || DEFAULT_SERVER_TIMEOUT_IN_MS;
+        this.serverTimeoutInMilliseconds = options.timeoutInMilliseconds || DEFAULT_SERVER_TIMEOUT_IN_MS;
```
mikaelm12: Did we also want to rename this `serverTimeoutInMiliseconds`?

mikaelm12: Maybe the constant as well

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HubConnection.spec.ts
anurse: If the property isn't present, it uses ConsoleLogger by default. In tests, we want no logging at all (because it messes with the test output).

PR #1242, opened 2017-12-20
Invoke Multiple Connections

mikaelm12:
Need suggestions with some of the naming 😄

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -582,18 +582,20 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             }
             var publishTasks = new List<Task>(connectionIds.Count);
             var message = new RedisInvocationMessage(target: methodName, arguments: args);
-
-            // If the connection is local we can skip sending the message through the bus since we require sticky connections.
-            // This also saves serializing and deserializing the message!
+            
             foreach(string connectionId in connectionIds)
             {
                 var connection = _connections[connectionId];
+                // If the connection is local we can skip sending the message through the bus since we require sticky connections.
+                // This also saves serializing and deserializing the message!
                 if (connection != null)
                 {
-                    return connection.WriteAsync(message.CreateInvocation());
+                     publishTasks.Add(connection.WriteAsync(message.CreateInvocation()));
                 }
-
-                publishTasks.Add(PublishAsync(_channelNamePrefix + "." + connectionId, message));
+                else
+                {
+                    publishTasks.Add(PublishAsync(_channelNamePrefix + "." + connectionId, message));
+                }
             }
 
             return Task.WhenAll(publishTasks);
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -574,6 +574,31 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             });
         }
 
+        public override Task InvokeConnectionsAsync(IReadOnlyList<string> connectionIds, string methodName, object[] args)
+        {
+            if (connectionIds == null)
+            {
+                throw new ArgumentNullException(nameof(connectionIds));
+            }
+            var publishTasks = new List<Task>(connectionIds.Count);
+            var message = new RedisInvocationMessage(target: methodName, arguments: args);
+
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -574,6 +574,31 @@ private Task SubscribeToGroup(string groupChannel, GroupData group)
...
+            var message = new RedisInvocationMessage(target: methodName, arguments: args);
+
+            // If the connection is local we can skip sending the message through the bus since we require sticky connections.
+            // This also saves serializing and deserializing the message!
+            foreach(string connectionId in connectionIds)
+            {
+                var connection = _connections[connectionId];
+                if (connection != null)
+                {
+                    return connection.WriteAsync(message.CreateInvocation());
```
anurse: This is going to exit from the whole method when you find a local connection, skipping all the remaining connections in the list. Instead, we should put the `PublishAsync` into an `else` clause and remove this `return`.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -574,6 +574,31 @@ private Task SubscribeToGroup(string groupChannel, GroupData group)
...
+        public override Task InvokeConnectionsAsync(IReadOnlyList<string> connectionIds, string methodName, object[] args)
+        {
+            if (connectionIds == null)
+            {
+                throw new ArgumentNullException(nameof(connectionIds));
+            }
+            var publishTasks = new List<Task>(connectionIds.Count);
+            var message = new RedisInvocationMessage(target: methodName, arguments: args);
+
+            // If the connection is local we can skip sending the message through the bus since we require sticky connections.
```
anurse: This comment belongs down on the `if` inside the loop

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -937,6 +937,51 @@ public void HubsCannotHaveOverloadedMethods()
...
+            using (var secondClient = new TestClient())
+            using (var thirdClient = new TestClient())
+            {
+                Task firstEndPointTask = endPoint.OnConnectedAsync(firstClient.Connection);
+                Task secondEndPointTask = endPoint.OnConnectedAsync(secondClient.Connection);
+                Task thirdEndPointTask = endPoint.OnConnectedAsync(thirdClient.Connection);
+
+                await Task.WhenAll(firstClient.Connected, secondClient.Connected, thirdClient.Connected).OrTimeout();
+
+                var secondAndThirdClients = new HashSet<string> {secondClient.Connection.ConnectionId, thirdClient.Connection.ConnectionId };
```
anurse: split this into multiple lines

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: Oops... Copy pasta error

anurse: That's what I figured :)

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -937,6 +937,53 @@ public void HubsCannotHaveOverloadedMethods()
...
+        [MemberData(nameof(HubTypes))]
+        public async Task SendToMultipleClients(Type hubType)
+        {
+            dynamic endPoint = HubEndPointTestUtils.GetHubEndpoint(hubType);
+
+            using (var firstClient = new TestClient())
+            using (var secondClient = new TestClient())
+            using (var thirdClient = new TestClient())
+            {
+                Task firstEndPointTask = endPoint.OnConnectedAsync(firstClient.Connection);
```
muratg: nit: var :trollface:

mikaelm12: Turns out these actually can't be var because we need `endPoint` to be dynamic. The endpoint tasks have to be explicitly typed as tasks

anurse: You're not wrong, I just dislike anything `dynamic` :trollface:

anurse: Might be worth refactoring these into separate tests with a common helper and using real generics (instead of Theories) sometime.

PR #1243, opened 2017-12-21
Refactor HttpClient and use it in LongPollingTransport

anurse:
This refactors the HttpClient abstraction a bit and changes LongPollingTransport to use it. This will allow us to add better unit tests of the LongPolling transport. Next I'll look at a similar abstraction for WebSocket and EventSource so we can test all the transports.

Note: I haven't actually added any new tests for LongPolling, but I added some tests for additional things.

To handle cancellation of Http requests, I whipped up my own version of [`AbortController` and friends](https://developer.mozilla.org/en-US/docs/Web/API/AbortController), an recently-specified JavaScript API that's similar to `CancellationTokenSource` and friends. Rather than doing it as a polyfill, since it's so simple, I don't even bother looking for the real one. It turns out that's a little tricky to do from within your own TypeScript module (since you have to mess with globals).

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/AbortSignal.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/AbortSignal.ts
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 import { asyncit as it } from "./Utils";
 import { AbortController } from "../Microsoft.AspNetCore.SignalR.Client.TS/AbortController";
 
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/AbortSignal.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/AbortSignal.ts
@@ -1,0 +1,28 @@
+import { asyncit as it } from "./Utils";
+import { AbortController } from "../Microsoft.AspNetCore.SignalR.Client.TS/AbortController";
+
+describe("AbortSignal", () => {
+    describe("aborted", () => {
+        it("is false on initialization", () => {
+            expect(new AbortController().signal.aborted).toBe(false);
+        });
+
+        it("is true when aborted", () => {
+            let controller = new AbortController();
+            let signal = controller.signal;
+            controller.abort();
+            expect(signal.aborted).toBe(true);
+        })
+    });
+
+    describe("onabort", () => {
+        it("is called when abort is called", () => {
+            let controller = new AbortController();
+            let signal = controller.signal;
+            let abortCalled = false;
+            signal.onabort = () => abortCalled = true;
+            controller.abort();
+            expect(abortCalled).toBe(true);
+        })
+    })
+});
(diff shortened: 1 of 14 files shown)


anurse:
All the JS functional tests pass after this change, FYI.

anurse:
I also set the logging level for SocketsSample up to Trace because it's really useful to see trace logging in the samples (and they aren't designed to be super fast or anything ;)).

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/AbortSignal.ts
```diff
@@ -0,0 +1,28 @@
+import { asyncit as it } from "./Utils";
```
mikaelm12: nit:Header

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HttpClient.spec.ts
```diff
@@ -0,0 +1,83 @@
+import { asyncit as it } from "./Utils"
```
mikaelm12: nit: Header

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HttpConnection.spec.ts
```diff
@@ -1,14 +1,15 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-import { IHttpClient } from "../Microsoft.AspNetCore.SignalR.Client.TS/HttpClient"
+import { TestHttpClient } from "./TestHttpClient"
```
mikaelm12: Super nit: Do we in general not care about the order of imports? Not a big deal

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/TestHttpClient.ts
```diff
@@ -0,0 +1,87 @@
+import { HttpClient, HttpRequest, HttpResponse } from "../Microsoft.AspNetCore.SignalR.Client.TS/HttpClient"
```
mikaelm12: nit: Header

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/AbortController.ts
```diff
@@ -0,0 +1,30 @@
+// Rough polyfill of https://developer.mozilla.org/en-US/docs/Web/API/AbortController
```
mikaelm12: nit: Headers

client-ts/Microsoft.AspNetCore.SignalR.Client.TS.Tests/HttpConnection.spec.ts
anurse: We should order them. It's just harder because we don't have a typescript auto-formatter to help us :)

anurse: Also, I'm not sure what we should order by here. The `from` clause? Or the list of imported symbols. It's worth thinking about, but we're pretty rough with it throughout the code right now.

anurse:
:up: :date:

Issue #1203, opened 2017-12-12
Don't use FormatException for negotiate response errors


anurse:
If the negotiate response from the server is invalid, the client throws `FormatException`. This is not an appropriate use of that exception, as it's supposed to be used when arguments that the caller provides are invalid. We should use either a custom exception type or something like `InvalidOperationException`.

anurse:
The fix for this will lie in https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/NegotiationProtocol.cs, and corresponding tests in https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs

We need to discuss what exception to use, or if we should create a custom one. That should happen in this issue before starting on a PR :)

**NOTE**: In #1208, the HttpConnectionTests are being significantly rewritten and reordered. It would be worth waiting for that PR to finish before writing tests.

jackjwilliams:
This looks like an easy one. I see no reason for a custom exception type for one class if InvalidOperationException fits the bill. Thoughts?

anurse:
I'm not sure `InvalidOperationException` fits the bill though. The error cases we want to express here are when the server sends a malformed Negotiate response. I see `InvalidOperationException` as kinda like an HTTP 400 error, it indicates the caller of the method did something wrong. But the caller didn't do anything wrong, the server did something wrong. I think maybe [`InvalidDataException`](https://msdn.microsoft.com/en-us/library/system.io.invaliddataexception(v=vs.110).aspx) fits better.

If you'd like to send a PR for this, please feel free! Let me know if you have any questions or concerns (you can reach me directly at `anurse@microsoft.com` if you'd like some 1:1 assistance, I'm very happy to help a new contributor out)

jackjwilliams:
Thanks! I've sent an introductory email. Yes, InvalidDataException does seem to fit better. It looks like we TryParseMessage the data, resulting in a payload (or a FormatException). Then attempt to read the Token, throwing a FormatException again if not valid.

Shouldn't be too difficult to change the exception type to InvalidDataException, keeping the error messages the same and update the test expectations.

ankitvijay:
This looks to be an easy fix and good start for my first PR. I can take this up.

jackjwilliams:
@ankitvijay, I already created a PR for this at https://github.com/aspnet/SignalR/issues/1246

[...]

PR #1246, opened 2017-12-21
Throw InvalidDataException instead of FormatException in NegotationProtocol

jackjwilliams:
Throw InvalidDataException instead of FormatException in NegotationProtocol

 - Modify NegotiationProtocol to throw InvalidDataException
 - Update NegotiationProtocolTests expectations
 - Remove test case for InlineData "Missing required property 'protocol'"

Addresses #1203

@anurse It took me a minute to figure out why one case failed for `[InlineData("{}\u001e", "Missing required property 'protocol'.")]`, it was because the test was expecting a FormatException from JsonUtils.GetRequiredProperty (which is still a valid case). 

I've removed it for now, but removing test cases is usually frowned upon. So I can account for the FormatException from JsonUtils.GetRequiredProperty and re-work the test some if you want.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
@@ -93,7 +93,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
 
                     if (token == null || token.Type != JTokenType.Object)
                     {
-                        throw new FormatException($"Unexpected JSON Token Type '{token?.Type}'. Expected a JSON Object.");
+                        throw new InvalidDataException($"Unexpected JSON Token Type '{token?.Type}'. Expected a JSON Object.");
                     }
 
                     var json = (JObject)token;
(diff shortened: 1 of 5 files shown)


anurse:
> it was because the test was expecting a FormatException from JsonUtils.GetRequiredProperty (which is still a valid case).

Actually, go ahead and change the exceptions in `JsonUtils` to `InvalidDataException` as well, which should allow you to restore the test case (it might mean some other test cases need to be updated too). We shouldn't really be using `FormatException` anywhere since it's specifically designed for `string.Format` and related APIs.

Looking good so far! The Travis failure is a flaky test that we've seen fail off-and-on, so it's not related to your change. If it fails again when you update the PR, I can rerun it until it passes (which may be the definition of insanity, but hey, it usually does the trick)

anurse:
Thanks! Change looks good so far, we'll see how the build shakes out (some of the tests are a bit flaky, but I can re-run them if they fail)

BrennanConroy:
Ping @anurse 
Good to merge?

anurse:
Yep. Thanks for the contribution @jackjwilliams !

jackjwilliams:
@anurse No problem, was a good learning experience!

Issue #1191, opened 2017-12-08
Socket Sample - disable buttons when not connected


mikaelm12:
When the client isn't connected all the buttons that require the client to be connected should be disabled.

anurse:
This is just a fix in the sample and should be fairly straightforward JavaScript code in the SocketsSample html pages. Marking up for grabs.

PR #1247, opened 2017-12-21
Update SocketsSample html buttons based on connectivity status

jackjwilliams:
- Update hubs.html buttons disabled prop for load/connect/disconnect/exception cases

Addresses #1191

Because why not?

Also changed connect/disconnect button to enable / disable on a connection.start error.

### DIFF
--- a/samples/SocketsSample/wwwroot/hubs.html
+++ b/samples/SocketsSample/wwwroot/hubs.html
@@ -108,12 +108,32 @@
 
     let connectButton = document.getElementById('connect');
     let disconnectButton = document.getElementById('disconnect');
-    disconnectButton.disabled = true;
+    let broadcastButton = document.getElementById('broadcast');
+    let broadcastExceptMeButton = document.getElementById('broadcast-exceptme');
+    let sendToConnectionButton = document.getElementById('connection-send');
+    let joinGroupButton = document.getElementById('join-group');
+    let leaveGroupButton = document.getElementById('leave-group');
+    let groupMsgButton = document.getElementById('groupmsg');
+    let othersGroupMsgButton = document.getElementById('others-groupmsg');
+    let sendButton = document.getElementById('send');
+
+    function updateButtonState(isConnected) {
+        broadcastButton.disabled = !isConnected;
+        broadcastExceptMeButton.disabled = !isConnected;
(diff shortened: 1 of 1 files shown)


jackjwilliams:
@anurse Yeah I'm not sure why I left those out. The strangest thing started happening when I attempted to reload my project: All the sample projects fail to load with this error:

The SDK 'Microsoft.NET.Sdk.Web' specified could not be found

dotnet --version shows 2.2.0-preview1-007522

global.json in SignalR/ is 2.2.0-preview1-007522

Not sure what I did. I'll have to look more closely later.

anurse:
Try closing all VS instances, running `./build.cmd /t:Restore` and then re-open VS. The `Restore` target *should* be installing the necessary SDKs for you, but something may have been messed up.

Also, open VS from the command prompt in which you run `build.cmd`. That script sets some things in your `PATH` environment variable that VS needs to inherit or it may not be able to find the appropriate SDKs

jackjwilliams:
@anurse Sorry I've had some life things come up, when is code freeze expected for the 2.1 release?

anurse:
No problem at all :). We've got a preview release coming up soon but we haven't branched for that yet. Don't worry about the timelines though, let's just get the PR ready to go and we'll merge it into whatever release is next. This is a sample app, so it's generally not subject to code freezes/escrow periods during our releases.

jackjwilliams:
@anurse Alright I got it running, if anyone else stumbles upon this, your solution of running from the command line works, or swap your stable / pre-release PATH variables around then re-run normally. 

Pre-release
image

Stable
image

anurse:
Ah yes! We've seen that just recently as well. Glad you got it working. Unfortunately (as described in https://github.com/aspnet/SignalR/issues/1298#issuecomment-358035319) we have to use that directory to put pre-release .NET SDKs on the PATH :(.

Thanks for the update!

anurse:
Hey @jackjwilliams . This got left behind a bit as we got busy with some other things, but I've gotten it rebased and it works great. I'm going to open a new PR with the updated branch just to verify that our CI still passes. You'll still get credit for the commit though! Thanks for the contribution!

jackjwilliams:
No problem, thank you!

PR #1250, opened 2017-12-22
React to pipeline changes

pakrym:


build/sources.props
```diff
@@ -7,6 +7,8 @@
       $(RestoreSources);
       https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
+      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
```
muratg: @natemcmaster FYI in case you have an alternative.

pakrym: It's temporary

natemcmaster: This is fine for local builds and temporary workarounds, but would break in Universe if you don't also update that to pull the new corefxlab packages.

PR #1251, opened 2017-12-22
Use new common benchmark config

pakrym:


benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/CoreConfig.cs
```diff
@@ -1,34 +0,0 @@
-﻿using BenchmarkDotNet.Columns;
-using BenchmarkDotNet.Configs;
-using BenchmarkDotNet.Diagnosers;
-using BenchmarkDotNet.Engines;
-using BenchmarkDotNet.Jobs;
-using BenchmarkDotNet.Validators;
-
-namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
-{
-    public class CoreConfig : ManualConfig
```
muratg: Is this configuration no longer required, or did the logic move somewhere else?

pakrym: Moved to common: https://github.com/aspnet/Common/pull/303

pakrym:
Travis failure is due to unrelated flakiness

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/AssemblyInfo.cs
```diff
@@ -0,0 +1 @@
+[assembly: BenchmarkDotNet.Attributes.AspNetCoreBenchmark]
```
natemcmaster: You don't need to change this...i'm just curious. Would does this work if you put it into the csproj file?
```xml
<ItemGroup>
   <AssemblyAttribute Include="BenchmarkDotNet.Attributes.AspNetCoreBenchmark" />
</ItemGroup>
```
If so, you could add this piece of MSBuild into the Benchmark.Sources package so it's automatically applied to any benchmark project that uses it.

pakrym: I don't want things that implicit. There might be a case where configs need to be overridden and having source file is much easier than trying to find some packages target that brought attribute and figure out how to change it.

pakrym: I could've just dropped assembly info into source package too.

PR #1254, opened 2017-12-28
Invoke Multiple Groups

mikaelm12:
Part of https://github.com/aspnet/SignalR/issues/734

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -131,8 +131,9 @@ namespace Microsoft.AspNetCore.SignalR
         {
             // Each task represents the list of tasks for each of the writes within a group
             var tasks = new List<Task>();
+            var message = CreateInvocationMessage(methodName, args);
 
-            foreach (string groupName in groupNames)
+            foreach (var groupName in groupNames)
             {
                 if (groupName == null)
                 {
@@ -142,7 +143,6 @@ namespace Microsoft.AspNetCore.SignalR
                 var group = _groups[groupName];
                 if (group != null)
                 {
-                    var message = CreateInvocationMessage(methodName, args);
                     tasks.Add(Task.WhenAll(group.Values.Select(c => c.WriteAsync(message))));
                 }
             }
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -135,9 +135,9 @@ namespace Microsoft.AspNetCore.SignalR
 
             foreach (var groupName in groupNames)
             {
-                if (groupName == null)
+                if (string.IsNullOrEmpty(groupName))
                 {
-                    throw new ArgumentNullException(nameof(groupName));
+                    throw new ArgumentException(nameof(groupName));
                 }
 
                 var group = _groups[groupName];
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -127,6 +127,29 @@ namespace Microsoft.AspNetCore.SignalR
             return Task.CompletedTask;
         }
 
+        public override Task InvokeGroupsAsync(IReadOnlyList<string> groupNames, string methodName, object[] args)
+        {
+            // Each task represents the list of tasks for each of the writes within a group
+            var tasks = new List<Task>();
+
+            foreach (string groupName in groupNames)
+            {
+                if (groupName == null)
+                {
+                    throw new ArgumentNullException(nameof(groupName));
+                }
+
+                var group = _groups[groupName];
+                if (group != null)
+                {
+                    var message = CreateInvocationMessage(methodName, args);
+                    tasks.Add(Task.WhenAll(group.Values.Select(c => c.WriteAsync(message))));
+                }
+            }
+
+            return Task.WhenAll(tasks);
+        }
+
         public override Task InvokeGroupExceptAsync(string groupName, string methodName, object[] args, IReadOnlyList<string> excludedIds)
         {
             if (groupName == null)
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubCallerClients.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubCallerClients.cs
@@ -38,6 +38,7 @@ namespace Microsoft.AspNetCore.SignalR
         {
             return _hubClients.Group(groupName);
         }
+
         public IClientProxy Groups(IReadOnlyList<string> groupNames)
         {
             return _hubClients.Groups(groupNames);
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubCallerClients.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubCallerClients.cs
@@ -38,9 +38,9 @@ namespace Microsoft.AspNetCore.SignalR
         {
             return _hubClients.Group(groupName);
         }
-        public IClientProxy MultipleGroups(IReadOnlyList<string> groupNames)
+        public IClientProxy Groups(IReadOnlyList<string> groupNames)
         {
-            return _hubClients.MultipleGroups(groupNames);
+            return _hubClients.Groups(groupNames);
         }
 
         public IClientProxy OthersInGroup(string groupName)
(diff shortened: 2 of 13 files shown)


src/Microsoft.AspNetCore.SignalR.Core/IHubClients`T.cs
```diff
@@ -17,6 +17,8 @@ public interface IHubClients<T>
 
         T Group(string groupName);
 
+        T MultipleGroups(IReadOnlyList<string> groupNames);
```
davidfowl: This should be called Groups

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -127,6 +127,29 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
...
+            {
+                if (groupName == null)
+                {
+                    throw new ArgumentNullException(nameof(groupName));
+                }
+
+                var group = _groups[groupName];
+                if (group != null)
+                {
+                    var message = CreateInvocationMessage(methodName, args);
```
anurse: Looks like the Invocation can be created outside the `foreach` loop. That would save some allocations.

src/Microsoft.AspNetCore.SignalR.Core/HubCallerClients.cs
```diff
@@ -38,6 +38,10 @@ public IClientProxy Group(string groupName)
         {
             return _hubClients.Group(groupName);
         }
+        public IClientProxy MultipleGroups(IReadOnlyList<string> groupNames)
```
anurse: newline needed

src/Microsoft.AspNetCore.SignalR.Core/IHubClients`T.cs
anurse: Not a huge fan of `Groups` either because it's only one character off, but I think it's more appropriate than `MultipleGroups`.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1123,6 +1123,50 @@ public void HubsCannotHaveOverloadedMethods()
...
+
+            using (var firstClient = new TestClient())
+            using (var secondClient = new TestClient())
+            using (var thirdClient = new TestClient())
+            {
+                Task firstEndPointTask = endPoint.OnConnectedAsync(firstClient.Connection);
+                Task secondEndPointTask = endPoint.OnConnectedAsync(secondClient.Connection);
+                Task thirdEndPointTask = endPoint.OnConnectedAsync(thirdClient.Connection);
+
+
```
anurse: Extra newline here.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -601,6 +601,23 @@ public override Task InvokeConnectionsAsync(IReadOnlyList<string> connectionIds,
...
+        public override Task InvokeGroupsAsync(IReadOnlyList<string> groupNames, string methodName, object[] args)
+        {
+            if (groupNames == null)
+            {
+                throw new ArgumentNullException(nameof(groupNames));
+            }
+            var publishTasks = new List<Task>(groupNames.Count);
+            var message = new RedisInvocationMessage(target: methodName, arguments: args);
+
+            foreach (string groupName in groupNames)
```
BrennanConroy: var

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1123,6 +1123,50 @@ public void HubsCannotHaveOverloadedMethods()
...
 
+        [Theory]
+        [MemberData(nameof(HubTypes))]
+        public async Task InvokeMultipleGroups(Type hubType)
+        {
+            dynamic endPoint = HubEndPointTestUtils.GetHubEndpoint(hubType);
+
+            using (var firstClient = new TestClient())
+            using (var secondClient = new TestClient())
+            using (var thirdClient = new TestClient())
```
BrennanConroy: nit: should be able to use 2 clients only

src/Microsoft.AspNetCore.SignalR.Core/IHubClients`T.cs
mikaelm12: I put multiple groups because the HubContext already refers to the `IGroupManager` as groups. Should we just rename that instead and use Groups here?

anurse: Hrm, why does `HubContext` implement `IHubClients`? SignalR 2.x didn't do that.

anurse: Maybe that's the fix here. HubContext shouldn't need to implement `IHubClients`, the clients can be accessed via `.Clients`.

anurse: That does mean we'd need to introduce a type to actually implement `Clients` (since HubContext just returns itself for that property right now)

mikaelm12: Okay, we can do that.

mikaelm12: Hmm, thinking that I might put this on hold to do that change and then renames

davidfowl: We should call it Groups

mikaelm12:
Updated. Re based and renamed

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -127,6 +127,29 @@ public override Task InvokeGroupAsync(string groupName, string methodName, objec
...
 
+        public override Task InvokeGroupsAsync(IReadOnlyList<string> groupNames, string methodName, object[] args)
+        {
+            // Each task represents the list of tasks for each of the writes within a group
+            var tasks = new List<Task>();
+            var message = CreateInvocationMessage(methodName, args);
+
+            foreach (var groupName in groupNames)
+            {
+                if (groupName == null)
```
BrennanConroy: `string.IsNullOrEmpty`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -601,6 +601,23 @@ public override Task InvokeConnectionsAsync(IReadOnlyList<string> connectionIds,
...
+        {
+            if (groupNames == null)
+            {
+                throw new ArgumentNullException(nameof(groupNames));
+            }
+            var publishTasks = new List<Task>(groupNames.Count);
+            var message = new RedisInvocationMessage(target: methodName, arguments: args);
+
+            foreach (var groupName in groupNames)
+            {
```
BrennanConroy: `string.IsNullOrEmpty` check here

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1123,6 +1123,46 @@ public void HubsCannotHaveOverloadedMethods()
...
+                Task firstEndPointTask = endPoint.OnConnectedAsync(firstClient.Connection);
+                Task secondEndPointTask = endPoint.OnConnectedAsync(secondClient.Connection);
+
+                await Task.WhenAll(firstClient.Connected, secondClient.Connected).OrTimeout();
+
+                await secondClient.InvokeAsync(nameof(MethodHub.GroupAddMethod), "GroupA").OrTimeout();
+                await firstClient.InvokeAsync(nameof(MethodHub.GroupAddMethod), "GroupB").OrTimeout(); ;
+
+                var groupNames = new List<string> { "GroupA", "GroupB" };
+                await firstClient.SendInvocationAsync("SendToMultipleGroups", "test", groupNames).OrTimeout();
```
BrennanConroy: nameof for "SendToMultipleGroups"

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1123,6 +1123,46 @@ public void HubsCannotHaveOverloadedMethods()
             }
         }
 
+        [Theory]
+        [MemberData(nameof(HubTypes))]
+        public async Task InvokeMultipleGroups(Type hubType)
```
BrennanConroy: Redis version of this test?

Issue #1236, opened 2017-12-19
Does Chatsample work with core 2.0?


mortenbirkelund:
I have tried to implement the example in order to get the presence part working in my core 2.0 project. But after implementation I get an error in PresenceHubLifetimeManager.cs on line 114

`hub.Clients = new HubCallerClients(_hubContext.Clients, connection.ConnectionId);`

> The type or namespace name 'HubCallerClients' could not be found (are you missing a using directive or an assembly reference?)

Will the sample only work in asp.net core 2.1 where signalr core is also released?

Thanks in advance for your help.

mikaelm12:
Are you using alpha 2 packages? HubCallerClients was added very recently so you'd need to be using the our most recent stuff. We are planning to release a preview1 soon which will have these changes and more.

You can go to the `rel/1.0.0-alpha2` branch to see how the ChatSample worked before if you're interested.

anurse:
If you want to try out the latest code in `dev` you can use the MyGet nightly feed, but beware that these bits have not received any verification aside from automated tests and can often have major bugs. See https://github.com/aspnet/Home#builds for some information on how to do that.

mortenbirkelund:
@mikaelm12  I was by mistake using the `dev`. I will try with the correct branch, `rel/1.0.0-alpha2`. Thanks

 @anurse Thanks, but it was by a mistake i used dev. I will try again with the correct branch. `

mortenbirkelund:
I works fine when using the correct branch. Silly me. 
However, whenever the function `UsersLeft` or `UsersJoined` are invoked, I get an error in firebug, stating that `TypeError: users.forEach is not a function`. 
However, when `SetUsersOnline` is invoked, it works fine, without any error, even though it also uses forEach to iterate. I cant seem to find any difference in the `users` object and the `usersOnline` object. 

`    // User online/offline
    connection.on('SetUsersOnline', usersOnline => {
        usersOnline.forEach(user => addUserOnline(user));
        
    });

    connection.on('UsersJoined', users => {
        users.forEach(user => {
            appendLine(user.name + " ankom til chatten");
            
            if (user.userId != myId) { addUserOnline(user) }
        });
    });


    connection.on('UsersLeft', users => {
        users.forEach(user => {           
            appendLine(user.name + " forlod chatten");
            removeUserOnline(user);
        });
    });`

mortenbirkelund:
@mikaelm12 @anurse ?

acjh:
[...]

PR #1261, opened 2018-01-02
Fix ChatSample to send array correctly

BrennanConroy:
#1236


anurse:
TODO: Consider renaming `InvokeAsync(string, object[])`? It's easy to fall in to this trap :(

PR #1262, opened 2018-01-02
Refactor HubContext

mikaelm12:
Changing HubContext to not implement IHubClients. Instead we'll introduce another type that will be exposed through the `Clients` property.
This allows us to rename some of the IHubClients properties (`MultipleConnections` -> `Clients`, `MultipleGroups` -> `Groups` etc).

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubContext.cs
@@ -5,51 +5,18 @@ using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public class HubContext<THub> : IHubContext<THub>, IHubClients where THub : Hub
+    public class HubContext<THub> : IHubContext<THub> where THub : Hub
     {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubContext.cs
```diff
@@ -5,51 +5,18 @@
...
         {
             _lifetimeManager = lifetimeManager;
-            All = new AllClientProxy<THub>(_lifetimeManager);
             Groups = new GroupManager<THub>(lifetimeManager);
         }
 
-        public IHubClients Clients => this;
-
-        public virtual IClientProxy All { get; }
+        public IHubClients Clients => new HubClients<THub>(_lifetimeManager);
```
BrennanConroy: Cache this

mikaelm12:
Updated

mikaelm12:
Made analogous changes in HubContext`T.
This adds another file `HubClients`T`. We can get rid of this though if we're okay with just closing the generic in `HubContext` by using `HubClients<THub, IClientProxy>`
Thoughts?

mikaelm12:
Ready for a final look

PR #1263, opened 2018-01-02
Fix race in Restart during Close task

BrennanConroy:


BrennanConroy:
Needs a little bit of work still, the race just moved/got smaller

anurse:
Hang on, doesn't this basically solve the race? We have start wait for a previous stop to complete before resetting the Close TCS.

BrennanConroy:
Stop isn't the only thing that can cause a reconnect :(

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -102,9 +108,239 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+                        {
+                            _logger.ExceptionThrownFromCallback(nameof(OnReceived), ex);
+                        }
+                    }
+                };
+                _connection.Closed += ex =>
+                {
+                    ChangeState(from: ConnectionState.Connected, to: ConnectionState.Disconnected);
+                    var closed = _closed.ToArray();
+                    foreach (var callback in closed)
```
mikaelm12: New line

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -27,32 +27,38 @@ public class HttpConnection : IConnection
...
+
+        private HttpConnectionOneOff _connection;
+        private List<Action<Exception>> _closed = new List<Action<Exception>>();
 
         public Uri Url { get; }
 
         public IFeatureCollection Features { get; } = new FeatureCollection();
 
-        public event Action<Exception> Closed;
+        public event Action<Exception> Closed
```
mikaelm12: Can we not just change _closed to be a Thread Safe type?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -102,9 +108,239 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+        public async Task StopAsync() => await StopAsyncCore().ForceAsync();
+
+        private async Task StopAsyncCore()
+        {
+            HttpConnectionOneOff connection;
+            lock (_stateChangeLock)
+            {
+                if (!(_connectionState == ConnectionState.Connecting || _connectionState == ConnectionState.Connected))
+                {
+                    //_logger.SkippingStop(_connectionId);
```
mikaelm12: A bunch of commented out logging lines like this

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -102,9 +108,239 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+                }
+
+                _connectionState = ConnectionState.Disconnected;
+                connection = _connection;
+            }
+
+            await connection.StopAsync();
+        }
+
+        private ConnectionState ChangeState(ConnectionState from, ConnectionState to)
```
mikaelm12: Consider changing from - to to something like startState, transitionState

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -102,9 +108,239 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+                {
+                    ChangeState(from: ConnectionState.Connected, to: ConnectionState.Disconnected);
+                    var closed = _closed.ToArray();
+                    foreach (var callback in closed)
+                    {
+                        try
+                        {
+                            callback(ex);
+                        }
+                        catch (Exception)
```
mikaelm12: nit: Make this one line

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -102,9 +108,239 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+                {
+                    _connectionState = to;
+                }
+
+                //_logger.ConnectionStateChanged(_connectionId, state, to);
+                return state;
+            }
+        }
+
+        private ConnectionState ChangeState(ConnectionState to)
```
mikaelm12: Change param name

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
BrennanConroy: Accessing shouldn't be frequent, in fact only 1 access should happen at a time in our code.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -102,9 +108,239 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+        internal enum ConnectionState
+        {
+            Disconnected,
+            Connecting,
+            Connected,
+            Disposed
+        }
+    }
+
+    internal class HttpConnectionOneOff : IConnection
```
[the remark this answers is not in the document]
BrennanConroy: `HttpConnectionCore`
`HttpConnectionSession`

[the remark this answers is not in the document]
anurse: Let's go with `Session`.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/HttpConnectionSession.cs
```diff
@@ -0,0 +1,509 @@
...
+                            _closedEvent.Invoke(abortException);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        // Suppress (but log) the exception, this is user code
+                        _logger.ErrorDuringClosedEvent(ex);
+                    }
+
+                    _closeTcs.SetResult(null);
```
BrennanConroy: FYI, I moved this to be after `_closedEvent` is called, this means closed callbacks will finish before `StopAsync` returns

BrennanConroy:
New attempt

BrennanConroy:
Ping
Review please

anurse:
Still a bit nervous about the race that occurs when the transport closes itself. If the transport closes (triggering `Input.Completion`) and the user calls `StopAsync` there's a race. If the `Completion` callback runs to the end before `StopAsync` captures the `closeTcs`, it gets the `closeTcs` for the previous call.

We need to talk about this today with @davidfowl. I want to get a handle on these races and the behavior we're actually looking for.

BrennanConroy:
Yeah we should definitely talk, but I'm pretty sure I fixed that race...

anurse:
No, it can still happen any time between the start of `StopAsync` and where you capture the `closeTcs`

BrennanConroy:
Thats why I added the check inside the lock and return

anurse:
Ah, ok, we might be good there. Let's walk through it though.

PR #1264, opened 2018-01-02
Lock around heartbeat handler

BrennanConroy:


src/Microsoft.AspNetCore.Sockets/DefaultConnectionContext.cs
```diff
@@ -75,14 +75,20 @@ public DefaultConnectionContext(string id, Channel<byte[]> transport, Channel<by
 
         public void OnHeartbeat(Action<object> action, object state)
         {
-            _heartbeatHandlers.Add((action, state));
```
mikaelm12: Why isn't this a concurrent dictionary?

anurse: It could be a `ConcurrentBag`. Or, as I suggested originally, an `ImmutableList` :trollface: 

It's rarely updated though. For the most part, handlers are added at the start and then run, so the lock is pretty low impact. Concurrent collections have their own trade-offs but tbh I'm not sure which is more appropriate now.

anurse:
I just remembered that this was in fact the place @davidfowl said not to worry about locking ;). Where was the contention appearing @BrennanConroy ?

BrennanConroy:
When I was running tests. Scan would run and a connection would be created at the same time

anurse:
Ok, well let's start with the lock and see what we find. The lock is the safest solution anyway.

Issue #898, opened 2017-09-16
Support cancelling Streaming results without client disconnection


davidfowl:
Today cancelling streaming results requires the connection to go away or the server side to sever the close it. There's no way for the client to independently cancel a stream without closing the connection. Support this requires a protocol change and a new message type to indicate a cancelled stream.

/cc @anurse @moozzyk

BrennanConroy:
Our current observable implementation on the TS client has a closed bool that is unused.
https://github.com/aspnet/SignalR/blob/e42f6980b643f6474c144fc7fdd36aedb0bac19f/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Observable.ts#L6-L11

I hacked together using it and clients can locally unsubscribe by setting close to true.
We could combine that with sending a new message type to the server that will unsubscribe it from the server side.

davidfowl:
@BrennanConroy look at RxJS to see how un subscription is done there.

BrennanConroy:
Looks like they check the closed value like I was. As well as splicing the subscribers array to remove the closed subscription.

muratg:
We should probably consider supporting non-streaming stops as well, perhaps via CancellationTokens.

davidfowl:
@BrennanConroy I assigned this to you

BrennanConroy:
When you call stream get back an observable
When you call unsubscribe on the observable send "complete" to server
When you call unsubscribe on the observer, remove from observable
When all observers unsubscribe, send "complete" to server
When server completes stream, complete all observers and clean up observable

```
var subject = connection.stream("streamMethod");
var subscriber = subject.subscribe({ next: x => console.log(x) });

// will cause subject to tell server to complete
subscriber.unsubscribe();
// or
// will complete subscriptions and send complete to server
subject.close();
```

@davidfowl @anurse @moozzyk 
Did I miss anything, or misinterpret anything?

anurse:
> When you call unsubscribe on the observable send "complete" to server

You can call unsubscribe on an observable? I'm not sure I understand what that means...

davidfowl:
Yea that doesn't look right. I think the code looks like this:

```
var observable = connection.stream("streamMethod");
var subscription = observable.subscribe({ next: x => console.log(x) });

// will cause subject to tell server to complete
subscription.unsubscribe();
```

BrennanConroy:
[...]

PR #1275, opened 2018-01-05
Support stream cancellation on TS client

BrennanConroy:
https://github.com/aspnet/SignalR/issues/898

### DIFF
--- a/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
+++ b/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
@@ -6,7 +6,7 @@ import { IConnection } from "./IConnection"
 import { HttpConnection } from "./HttpConnection"
 import { TransportType, TransferMode } from "./Transports"
 import { Subject, Observable } from "./Observable"
-import { IHubProtocol, ProtocolType, MessageType, HubMessage, CompletionMessage, ResultMessage, InvocationMessage, StreamInvocationMessage, NegotiationMessage } from "./IHubProtocol";
+import { IHubProtocol, ProtocolType, MessageType, HubMessage, CompletionMessage, ResultMessage, InvocationMessage, StreamInvocationMessage, NegotiationMessage, CancelInvocation } from "./IHubProtocol";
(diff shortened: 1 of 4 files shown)


client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Observable.ts
```diff
@@ -10,16 +10,39 @@ export interface Observer<T> {
...
+    observer: Observer<T>;
+
+    constructor(subject: Subject<T>, observer: Observer<T>) {
+        this.subject = subject;
+        this.observer = observer;
+    }
+
+    public dispose(): void {
+        let index: number = this.subject.observers.indexOf(this.observer);
+        if (index > -1) {
```
mikaelm12: Would an equality check be faster?

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Observable.ts
```diff
@@ -10,16 +10,39 @@ export interface Observer<T> {
     complete?: () => void;
 }
 
+export class ObserverDisposable<T> {
```
mikaelm12: Should this just be called Subscription?

mikaelm12: I think it makes more sense as Subscription. It's the result of calling subscribe. It's the Rx convention as well I believe

BrennanConroy:
🆙 📅

mikaelm12:
Looks pretty good to me but I'll let someone more comfortable with typescript do the final sign off 😄

client-ts/Microsoft.AspNetCore.SignalR.Client.TS/HubConnection.ts
```diff
@@ -310,4 +317,11 @@ export class HubConnection {
             arguments: args,
         };
     }
+
+    private createCancelInvocation(id: string): CancelInvocation {
+        return <CancelInvocation>{
```
anurse: Do you need the cast? I thought TypeScript would be smart enough to infer this. If it's necessary though, keep it.

PR #1279, opened 2018-01-09
Fix flaky SSE test

BrennanConroy:
There is technically a bug here, however it requires https://github.com/dotnet/corefxlab/issues/1824 to fix and in practice will never happen because the SSE stream is getting data from the network so it will be async and not hit the issue.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -82,7 +82,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
             var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
 
-            using (var stream = await response.Content.ReadAsStreamAsync())
+            var stream = await response.Content.ReadAsStreamAsync();
             {
                 var pipelineReader = StreamPipeConnection.CreateReader(PipeOptions.Default, stream);
                 var readCancellationRegistration = cancellationToken.Register(
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -83,59 +83,64 @@ namespace Microsoft.AspNetCore.Sockets.Client
             var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
 
             var stream = await response.Content.ReadAsStreamAsync();
+            var pipelineReader = StreamPipeConnection.CreateReader(PipeOptions.Default, stream);
+
+            var readCancellationRegistration = cancellationToken.Register(
+                reader => ((PipeReader)reader).CancelPendingRead(), pipelineReader);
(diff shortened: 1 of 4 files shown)


anurse:
I'm gonna need some background on why the test was flaking out and how this fixes it :).

davidfowl:
Yea, this needs some explanation.

Eilon:
Also, is this meant as a temporary thing? If so, please make sure there's a new bug logged to revert this.

BrennanConroy:
?

Eilon:
Sorry didn't realize this is a month old. But it still wasn't clear to me what was said about "this is technically a bug" and it depends on a CoreFXLab change that hasn't been made.

BrennanConroy:
Since they weren't shipping that code yet, we copied it into our project and modified it, we can file a bug to use them again when they ship

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -82,60 +82,65 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferMode requestedT
...
+            finally
+            {
+                readCancellationRegistration.Dispose();
+                _transportCts.Cancel();
+                try
                 {
-                    readCancellationRegistration.Dispose();
-                    _transportCts.Cancel();
-                    _logger.ReceiveStopped();
+                    stream.Dispose();
```
davidfowl: I was going to ask you to ifdef the try catch by meh.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -260,7 +260,7 @@ public static IEnumerable<object[]> MessageSizesData
         [MemberData(nameof(MessageSizesData))]
         public async Task ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport(string message)
         {
-            using (StartLog(out var loggerFactory, LogLevel.Trace, testName: $"ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport_{message.Length}"))
+            using (StartLog(out var loggerFactory, LogLevel.Information, testName: $"ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport_{message.Length}"))
```
davidfowl: Leave it as Trace

test/Microsoft.AspNetCore.SignalR.Client.Tests/ServerSentEventsTransportTests.cs
```diff
@@ -81,9 +80,9 @@ public class ServerSentEventsTransportTests
                         {
                             await Task.Yield();
                             var buffer = Encoding.ASCII.GetBytes("data: 3:abc\r\n\r\n");
-                            while (!eventStreamCts.IsCancellationRequested)
+                            while (!t.IsCancellationRequested)
```
davidfowl: rename this to Token.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
BrennanConroy: It floods the logs with the large message

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
BrennanConroy: ifdef requires cross-compiling

PR #1282, opened 2018-01-10
Updating SocketsSample

mikaelm12:
Updating SocketsSample with new IHubClients and IHubCallerClients APIs
There was some missing functionality that we can now add easily.

### DIFF
--- a/samples/SocketsSample/Hubs/HubTChat.cs
+++ b/samples/SocketsSample/Hubs/HubTChat.cs
@@ -16,7 +16,7 @@ namespace SocketsSample.Hubs
 
         public override async Task OnDisconnectedAsync(Exception ex)
         {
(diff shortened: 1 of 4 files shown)


samples/SocketsSample/Hubs/HubTChat.cs
```diff
@@ -38,14 +48,14 @@ public Task SendToGroup(string groupName, string message)
 
         public async Task LeaveGroup(string groupName)
         {
-            await Groups.RemoveAsync(Context.ConnectionId, groupName);
```
davidfowl: Why did you change this order?

mikaelm12: So that you see the "left group" message when you leave.

PR #1288, opened 2018-01-11
remove System.IO.Pipelines.Extensions dependency and internalize code

anurse:


davidfowl:
@anurse you should copy the minimal code required to make the scenario work. I think you can delete a bunch of those functions.

anurse:
@davidfowl This **is** after deleting functions.

PR #1289, opened 2018-01-11
InvokeFirstWhere

mikaelm12:
In the `DefaultHubLifeTimeManagers ` `InvokeUserAsync` we use `InvokeAllWhere `which scans every connnection every time even though we only care about one. `InvokeFirstWhere `stops after we find the connection we want.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -90,6 +90,25 @@ namespace Microsoft.AspNetCore.SignalR
             return Task.WhenAll(tasks);
         }
 
+        private Task InvokeFirstWhere(string methodName, object[] args, Func<HubConnectionContext, bool> include)
+        {
+            if (_connections.Count == 0)
+            {
+                return Task.CompletedTask;
+            }
+
+            foreach (var connection in _connections)
+            {
+                if (include(connection))
+                {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -176,7 +195,7 @@ private InvocationMessage CreateInvocationMessage(string methodName, object[] ar
 
         public override Task InvokeUserAsync(string userId, string methodName, object[] args)
         {
-            return InvokeAllWhere(methodName, args, connection =>
+            return InvokeFirstWhere(methodName, args, connection =>
```
BrennanConroy: I don't think we want this. You can have the same user connected from multiple locations and with this change only one would get the message, but all of them should get the message.

mikaelm12: Hmmm, good point. Didn't think of that.

mikaelm12: Should there be a better way to may User Identifies to connection ids though?

davidfowl:
What's the scenario for this?

mikaelm12:
It was to avoid using InvokeAllWhere to implement InvokeUserAsync but as @BrennanConroy pointed out we actually should be using InvokeAllWhere

Issue #1300, opened 2018-01-13
Consider renaming `InvokeAsync` on server to `SendAsync`


anurse:
The client has `SendAsync` which is fire-and-forget and `InvokeAsync` which waits for a completion message from the server.

The server has `InvokeAsync`, which is fire-and-forget and thus out of sync with the client APIs.

I propose renaming the server's `InvokeAsync` to `SendAsync` to match the client. It is a break from the SignalR 2.x API, but a) it's a simple rename and b) most people were using the `dynamic` API which isn't present by default anyway so it's not like porting relied on this API being the same name.

It also allows us to open the door to Client Results by using the `InvokeAsync` name on the server.

I think we should do this in Preview 1 if we're going to do it.

cc @davidfowl @DamianEdwards @muratg

muratg:
@anurse makes sense to me, let's do it unless @davidfowl or @DamianEdwards reacts in the meantime.

davidfowl:
Ugh, I was hoping we would fudge this but the purist in me says its right

anurse:
![https://media1.tenor.com/images/ee9682d61532ff2b111775f22e4d7077/tenor.gif?itemid=5340126](https://media1.tenor.com/images/ee9682d61532ff2b111775f22e4d7077/tenor.gif?itemid=5340126)

DamianEdwards:
Agreed

grahamehorner:
@anurse @DamianEdwards @davidfowl IMHO InvokeAsync seems correct as SignalR is in essence doing a RPC (just my two pennies worth) but whichever I'm look forward to SignalR got some interesting ideas for a SignalR mesh 👯‍♂️

anurse:
@grahamehorner You're correct, which is why the Client has `InvokeAsync`. RPC implies that the call doesn't return until the method invocation has completed (which is how the Client's version of `InvokeAsync` behaves). However, the server does not currently support receiving a result from the client. All server-to-client invocations are fire-and-forget (as it was in SignalR 2.x), so it's more appropriate to use `SendAsync` to match the behavior of the client's version of `SendAsync`. We do have some thoughts about enabling results from server-to-client invocations as well, so we'll want to reserve `InvokeAsync` for that purpose.

anurse:
It sounded like @BrennanConroy was looking for Preview 1 work at stand-up this morning ;) /cc @muratg

BrennanConroy:
[...]

PR #1312, opened 2018-01-17
Rename InvokeAsync to SendAsync on the server

BrennanConroy:
#1300

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/IClientProxy.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/IClientProxy.cs
@@ -8,7 +8,7 @@ namespace Microsoft.AspNetCore.SignalR
     public interface IClientProxy
     {
         /// <summary>
-        /// Invokes a method on the connection(s) represented by the <see cref="IClientProxy"/> instance.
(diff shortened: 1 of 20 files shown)


src/Microsoft.AspNetCore.SignalR.Core/IClientProxy.cs
```diff
@@ -8,11 +8,11 @@ namespace Microsoft.AspNetCore.SignalR
     public interface IClientProxy
     {
         /// <summary>
-        /// Invokes a method on the connection(s) represented by the <see cref="IClientProxy"/> instance.
+        /// Fire-and-forget a method on the connection(s) represented by the <see cref="IClientProxy"/> instance.
```
mikaelm12: I do agree that this should be changed but this doesn't read well to me.

davidfowl: I'd change it back. It does invoke a method. Just mention that it doesn't wait for the other side to send a reply.

davidfowl:
1. Why is Travis failing?
1. Are we going to change the HubLifetimeManager methods?

BrennanConroy:
1. One of our flaky tests
2. We probably should for consistency

Issue #1310, opened 2018-01-17
SignalR social weather sample is giving 1006 websocket error for both C# and json client


yackmurali:
Both C# and json client were not able to connect to a server in same machine. It is giving the error 1006

Status: WebSocket closed. Reason: 1006
WebSocket connection to 'ws://localhost:59619/weather/ws?formatType=json' failed: Error during WebSocket handshake: Unexpected response code: 404
document.addEventListener @ (index):71
:59619/favicon.ico Failed to load resource: the server responded with a status of 404 (Not Found)

But the same client got connected to socialweather server in azure site and getting response.

what we need to do to correct the problem.

BrennanConroy:
What OS are you running the server on? Is it behind a proxy like IIS?

yackmurali:
I am running on windows 10.
How can i make sure it is not running behind the proxy.

yackmurali:
Also socket sample is working without any problem

BrennanConroy:
Looked into it, and the sample is out of date, it no longer works.

BrennanConroy:
Fixed via https://github.com/aspnet/SignalR/commit/4203540cb07d446ff7a32d0973846b32e2ce450b

PR #1313, opened 2018-01-17
Fix SocialWeather sample

BrennanConroy:
#1310
* Was pointing to old /ws endpoint
* Wasn't using format sent via query-string

Lua and CSharp clients which use the other formats are located https://github.com/moozzyk/SocialWeatherClients/tree/master/src
Not sure what to do about those


moozzyk:
Send a PR (or I can fix them 😄)

moozzyk:
@BrennanConroy - So, I took a look and Protobuf does not work because websockets are always using text mode. [This](https://github.com/aspnet/SignalR/pull/1362) fixes it. An alternative (which I have not tried) could be to set the transfer mode for websockets always to binary.

PR #1319, opened 2018-01-18
Major TypeScript module refactoring

anurse:
This is a bit large, but it should be fairly straightforward, there's no functional code change, just some restructuring.

1. Split @aspnet/signalr-client into two packages: @aspnet/signalr (base) and @aspnet/signalr-protocol-msgpack (msgpack extension)
    * Also split tests!
2. Remove MSBuild projects for npm modules. It was a weird hybrid. I might try to reintroduce them if people really want them. The net effect is that you can't use full VS to edit them easily, but VSCode works great.
3. Restructure package.json devDependencies to avoid duplication
4. Remove gulp, just use NPM scripts to automate
5. Build four outputs for each package:
    * [ECMAScript Module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) using ES2016 in `dist/esm` (future-proofing "best practice" used by many packages)
    * [CommonJS Module](http://wiki.commonjs.org/wiki/Modules/1.1.1) using ES5 in `dist/cjs` (for NodeJS)
    * [UMD Module](https://github.com/umdjs/umd) using ES5 in `dist/browser` (for Browser), including minified version (using uglify-js)
    * TypeScript Definitions
6. Use [Rollup](https://rollupjs.org/) to produce the UMD module, it's like browserify but with tree-shaking to reduce unused polyfill code
7. Only produce the ES5 version for the browser. Most people seem to want to use that, the ES2016 version was extraneous. If you're targetting only ES2016 browsers, use the `dist/esm` files natively or with something like webpack/rollup. We can always revisit this.
8. Integrate new NPM packages into the build.
9. Convert the functional tests to use TypeScript (it was a pain that they were the only things that didn't use that ;))

~~TODO:~~ Done!
* [x] Clean up git ignores and reverify samples (some of the code the FunctionalTests use is still in there, will fix shortly)
* [x] Finish reintegrating npm packaging with the build
* [x] Verify functional tests on IE (the least-compatible browser ;))
* [x] Review need for `msgpackshim` func.


client-ts/package.json
```diff
@@ -1,47 +1,24 @@
...
   "scripts": {
-    "gulp": "gulp",
-    "pretest": "tsc -p Microsoft.AspNetCore.SignalR.Client.TS.Tests",
-    "test": "jasmine JASMINE_CONFIG_PATH=Microsoft.AspNetCore.SignalR.Client.TS.Tests/jasmine.json"
-  },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/aspnet/SignalR.git"
+    "build": "cd ./signalr && npm run build && cd ../signalr-protocol-msgpack && npm run build && cd ../FunctionalTests && npm run build",
+    "test": "cd ./signalr && npm run test && cd ../signalr-protocol-msgpack && npm run test"
```
anurse: My kingdom for JSON comments.

These scripts are just for convenience. You can run `npm run build` or `npm test` in the `client-ts` folder and it will build/test all the subfolders.

client-ts/rollup-base.js
```diff
@@ -0,0 +1,55 @@
+import path from 'path';
```
anurse: https://rollupjs.org/guide/en#configuration-files

client-ts/rollup-base.js
```diff
@@ -0,0 +1,55 @@
...
+            format: "umd",
+            name: pkg.umd_name,
+            sourcemap: true,
+            banner: "/* @license\r\n" +
+                " * Copyright (c) .NET Foundation. All rights reserved.\r\n" +
+                " * Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.\r\n" +
+                "*/",
+            globals: moduleGlobals,
+        },
+        external: function (m) {
```
anurse: This function ensures that things in `node_modules` aren't "rolled up" into the browser file. **EXCEPT** for the polyfills we explicitly allow, listed above.

anurse: More context: This function is called with the right-side of every `import {} from "foo"`, and given the **resolved** path for `"foo"`. So for things within our own source, we return `false`, indicating it is NOT external and should be rolled-up. For things in `node_modules`, we check against the polyfill list.

client-ts/signalr-protocol-msgpack/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "main": "./dist/cjs/index.js",
+  "browser": "./dist/browser/signalr-protocol-msgpack.js",
+  "module": "./dist/esm/index.js",
+  "typings": "./dist/esm/index.d.ts",
+  "umd_name": "signalR.protocols.msgpack",
+  "directories": {
+    "test": "spec"
+  },
+  "scripts": {
+    "clean": "node ../node_modules/rimraf/bin.js ./dist",
```
anurse: The `node ../node_modules/.../` syntax is needed because we actually install the tools into the parent directory's `package.json`. If we installed them in this `package.json` we could use `rimraf ../dist`, **but** we'd have copies of them in both the `signalr` and `signalr-protocol-msgpack` folders :(

client-ts/signalr-protocol-msgpack/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "build": "npm run clean && npm run build:esm && npm run build:cjs && npm run build:browser && npm run build:uglify",
+    "build:esm": "node ../node_modules/typescript/bin/tsc --project ./tsconfig.json --module es2015 --outDir ./dist/esm --target ES2015 -d",
+    "build:cjs": "node ../node_modules/typescript/bin/tsc --project ./tsconfig.json --module commonjs --outDir ./dist/cjs --target ES5",
```
anurse: Have to build twice to get the right ESM and CJS builds. I looked at using babel to translate from ESM to CJS (rather than compiling the TS files again) but a) the perf is basically the same and b) it wasn't perfect, the helpers that TS injects get funky

client-ts/signalr-protocol-msgpack/spec/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "compilerOptions": {
+        "module": "commonjs",
+        "target": "es5",
+        "moduleResolution": "node",
+        "outDir": "./obj",
+        "lib": [ "es2015", "dom" ],
+        "baseUrl": ".",
+        "paths": {
+            "@aspnet/signalr": [ "../../signalr/src/index" ]
+        }
```
anurse: This allows us to use `import {} from "@aspnet/signalr"` even though we don't actually use the `@aspnet/signalr` NPM package, we want to use it from source in the sibling directory.

client-ts/signalr-protocol-msgpack/src/MessagePackHubProtocol.ts
```diff
@@ -1,9 +1,16 @@
...
+import { IHubProtocol, ProtocolType, MessageType, HubMessage, InvocationMessage, ResultMessage, CompletionMessage, StreamInvocationMessage } from "@aspnet/signalr";
+import { BinaryMessageFormat } from "./BinaryMessageFormat"
+import { Buffer } from 'buffer';
+import * as msgpack5 from "msgpack5";
+
+// This shim is needed to handle the fact that msgpack is exported like a CommonJS module
+// but we're building as an ECMAScript module.
+function msgpackshim(): msgpack5.MessagePack {
+    return <msgpack5.MessagePack>msgpack5.call({});
+}
```
anurse: \*sigh* msgpack is goofy. I'm not quite sure why this was "working" before. It may not have been fully working. I get typescript warnings in VSCode with the original code. This adds some forced casting to clear those out. I'll re-review that before merge.

client-ts/signalr/build/process-dts.js
```diff
@@ -0,0 +1,8 @@
+// Edit the index.d.ts file to add the UMD export
+const fs = require('fs');
+const path = require('path');
+
+const target = path.resolve(__dirname, "..", "dist", "esm", "index.d.ts");
+
+let content = fs.readFileSync(target);
+fs.writeFileSync(target, content + "\r\nexport as namespace signalR;");
```
anurse: This syntax (`export as namespace signalR`) allows the dts file to be used both when using it from a module (`import { HubConnection } from "@aspnet/signalr"`) and when using it from a "script" (a file with no `imports`, via `let c = new signalR.HubConnection`).

client-ts/umd-test-page.html
```diff
@@ -0,0 +1,103 @@
+<html>
```
anurse: This was a test HTML page I added to do some quick testing to make sure the browser scripts "export" themselves properly to global variables. I figured it didn't hurt to keep it in.

client-ts/signalr-protocol-msgpack/src/third-party-notices.txt
```diff
@@ -8,6 +8,85 @@ bring it to our attention. Post an issue or email us:
 
 The attached notices are provided for information only.
```
anurse: Paging @Eilon

client-ts/signalr/src/HubConnection.ts
```diff
@@ -3,22 +3,22 @@
...
 import { ConnectionClosed } from "./Common"
 import { IConnection } from "./IConnection"
-import { HttpConnection } from "./HttpConnection"
+import { HttpConnection, IHttpConnectionOptions } from "./HttpConnection"
 import { TransportType, TransferMode } from "./Transports"
 import { Subject, Observable } from "./Observable"
 import { IHubProtocol, ProtocolType, MessageType, HubMessage, CompletionMessage, ResultMessage, InvocationMessage, StreamInvocationMessage, NegotiationMessage, CancelInvocation } from "./IHubProtocol";
-import { JsonHubProtocol } from "./JsonHubProtocol";
-import { TextMessageFormat } from "./Formatters"
+import { JsonHubProtocol, JSON_HUB_PROTOCOL_NAME } from "./JsonHubProtocol";
```
BrennanConroy: Explain?

client-ts/README.md
```diff
@@ -1,34 +0,0 @@
-JavaScript and TypeScript clients for SignalR for ASP.NET Core
```
BrennanConroy: We need a readme for both packages for npmjs.com

client-ts/signalr/src/third-party-notices.txt
```diff
@@ -0,0 +1,32 @@
+.NET Core uses third-party libraries or other resources that may be
```
anurse: Also paging @Eilon (just to make sure this looks right)

client-ts/signalr/src/HubConnection.ts
anurse: Old code, didn't end up needing it.

anurse:
It should now be properly producing NPM packages that will be published to MyGet during the CI build. Hoping to get this in today :)

client-ts/signalr-protocol-msgpack/src/third-party-notices.txt
Eilon: This looks good.

client-ts/signalr/src/third-party-notices.txt
Eilon: Looks good.

client-ts/signalr-protocol-msgpack/src/third-party-notices.txt
anurse: And just to clarify, the `src` directory **is** included in the final package (it's useful for source maps, debugging, etc.). If we decide to remove it, we'll need to ensure this file remains somewhere in the final package.

Eilon: Yup.

client-ts/signalr/README.md
```diff
@@ -20,7 +20,7 @@ The following polyfills are required to use the client in Node.js applications:
 - `EventSource` - to use the ServerSentEvents transport
 - `btoa/atob` - to use binary protocols (e.g. MessagePack) over text transports (ServerSentEvents)
```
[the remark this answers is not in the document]
anurse: Actually no. The base64 logic is in the core. It's used to adapt a transport from Text mode to Binary mode, which actually lives in the transport layer.

PR #1322, opened 2018-01-19
Expose IHttpConnectionFeature Properties

mikaelm12:
Issue https://github.com/aspnet/SignalR/issues/944

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -460,7 +460,7 @@ namespace Microsoft.AspNetCore.Sockets
                 return false;
             }
 
-            //Set the IHttpConnectionFeature now that we can access it.
+            // Set the IHttpConnectionFeature now that we can access it.
             connection.Features.Set(context.Features.Get<IHttpConnectionFeature>());
 
             var transport = (TransportType?)connection.Metadata[ConnectionMetadataNames.Transport];
@@ -541,7 +541,6 @@ namespace Microsoft.AspNetCore.Sockets
                 return null;
             }
 
-            connection.Features.Set(context.Features.Get<IHttpConnectionFeature>());
             return connection;
         }
     }
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -8,6 +8,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Sockets.Features;
 using Microsoft.AspNetCore.Sockets.Internal;
 using Microsoft.AspNetCore.Sockets.Internal.Transports;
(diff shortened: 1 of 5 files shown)


BrennanConroy:
Tests?

mikaelm12:
+ Tests

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
```diff
@@ -44,6 +44,26 @@ public string GetCookieValue(string cookieName)
         {
             return Context.Connection.GetHttpContext().Request.Cookies[cookieName];
         }
+
+        public int GetLocalPort()
```
davidfowl: Change to a single method that returns all of the properties

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -363,6 +364,9 @@ private Task ProcessNegotiate(HttpContext context, HttpSocketOptions options, Co
             // Establish the connection
             var connection = _manager.CreateConnection();
 
+            //Set the IHttpConnectionFeature now that we can access it.
+            connection.Features.Set(context.Features.Get<IHttpConnectionFeature>());
```
davidfowl: Why do it here? User code never accessing the connection at this point right?

mikaelm12: I put it here because it was the earliest place we could. Where do you think would be a better spot?

davidfowl: I think it isn't required here at all.

mikaelm12: It is. The feature isn't set otherwise and we get null refs in the tests.

davidfowl: Are the tests broken? I think we should be able to move this to EnsureConnectionState

mikaelm12: Yeah, the issue was human error. Working now 😄

mikaelm12:
Updated

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -459,6 +460,9 @@ private static string GetNegotiatePayload(string connectionId, HttpSocketOptions
                 return false;
             }
 
+            //Set the IHttpConnectionFeature now that we can access it.
```
BrennanConroy: nit: space after `//`

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -537,6 +541,7 @@ private static string GetNegotiatePayload(string connectionId, HttpSocketOptions
                 return null;
             }
 
+            connection.Features.Set(context.Features.Get<IHttpConnectionFeature>());
```
BrennanConroy: I don't think this one is needed anymore

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -757,6 +757,45 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
...
+                    loggerFactory.CreateLogger<HubConnectionTests>().LogError(ex, "Exception from test");
+                    throw;
+                }
+                finally
+                {
+                    await hubConnection.DisposeAsync().OrTimeout();
+                }
+            }
+        }
+
```
BrennanConroy: remove newline

BrennanConroy:
Can close this now ;)

PR #1331, opened 2018-01-22
Return stream method error to client

BrennanConroy:
Before this change `Observable.Throw` would cause a `ChannelClosedException` and the client would get an error stating the "channel has closed", now the client will get the message from `Observable.Throw`

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -400,6 +400,11 @@ namespace Microsoft.AspNetCore.SignalR
                     await SendMessageAsync(connection, new StreamItemMessage(invocationId, enumerator.Current));
                 }
             }
+            catch (ChannelClosedException ex)
+            {
+                // If the channel closes from an exception in the streaming method, grab the innerException for the error from the streaming method
+                error = ex.InnerException == null ? ex.Message : ex.InnerException.Message;
+            }
             catch (Exception ex)
             {
                 // If the streaming method was canceled we don't want to send a HubException message - this is not an error case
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1386,6 +1386,31 @@ public void HubsCannotHaveOverloadedMethods()
...
+                await client.Connected.OrTimeout();
+
+                var messages = await client.StreamAsync(nameof(StreamingHub.ThrowStream));
+
+                Assert.Equal(1, messages.Count);
+                var completion = messages[0] as CompletionMessage;
+                Assert.NotNull(completion);
+                Assert.Equal("Exception from observable", completion.Error);
+
+                client.Dispose();
```
mikaelm12: Do you need to dispose here? You have the using statement

BrennanConroy: Needed because `endPointLifetime` is blocked until the client goes away

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -400,6 +400,11 @@ private static bool IsChannel(Type type, out Type payloadType)
                     await SendMessageAsync(connection, new StreamItemMessage(invocationId, enumerator.Current));
                 }
             }
+            catch (ChannelClosedException ex)
```
anurse: :( Not a huge fan of this, but I don't see a better way to do this via IAsyncEnumerator. We could use the channel APIs directly I guess, and use `WaitToRead` to detect this case (and then `.Completion` to get the exception).

BrennanConroy:
@muratg May I merge to preview1 branch?

Roonerelli:
@BrennanConroy @anurse @muratg 
this doesnt appear to be fixed
e.g if I have a hub method that return the following:

public IObservable<MyClass> LiveStream()
        {

            return Observable.Create(
                async (IObserver<FxDeal> observer) =>
                {
                    for (int i = 0; i < 100; i++)
                    {
                        if (i == 5)
                        {
                            observer.OnError(new Exception("Test."));
                        }
                        observer.OnNext(new MyClass("id", i) );
                        await Task.Delay(1000);
                    }
                })

Then the exception message isn't propagated to the client/browser

If I return the error as follows, it is

Subject<MyClass> rateChanged = new Subject<MyClass>();
            rateChanged.OnError(new Exception("Test."));
            return rateChanged.AsObservable();

Any ideas what is wrong? is it a known issue?

thanks

Issue #1169, opened 2017-11-30
Stop using EventId of 0 for named events and tidy up logger extension methods.


anurse:
The EventID "0" tends to be reserved in our system for "unnamed" events. Note that other Logging Extension methods in other repos don't use it: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MvcViewFeaturesLoggerExtensions.cs#L39

We should avoid using it and start known events from `1`. The Event ID `0` should remain reserved for unnamed events. We should definitely do this before preview 1, but it's a simple quick pass over our extension methods. We should also look at ways to ensure they don't get mixed up between categories. There are a couple ways to do this:

1. @rynowak has used this pattern: Add a static nested class called `Log` to the component itself that has these "extension" methods on them (they can't be true extension methods because of that). Then each of these classes has only the log messages relevant to that logger category. See https://github.com/aspnet/HttpClientFactory/blob/9a393fa4ce189fee3eaefe12ba2ad8a3dc06c752/src/Microsoft.Extensions.Http/Logging/LoggingHttpMessageHandler.cs#L47
2. Hang the extension methods off of `ILogger<T>` instead of `ILogger`. This doesn't stop you from mixing up event IDs though since the static field is "global".

anurse:
One other crazy idea:

```csharp
public class SomeComponent
{
    private SomeComponentLogger _logger;
    public SomeComponent(ILogger<SomeComponent> logger) {
        _logger = new SomeComponentLogger(logger);
    }

    public void DoSomething() {
        _logger.SomeEvent(...);
    }

    private struct SomeComponentLogger {
        private static readonly Action<ILogger, ...> _someEvent = LoggerMessage.Define<...>(...);

        private ILogger _logger;
        public SomeComponentLogger(ILogger logger) {
            _logger = logger;
        }

        public void SomeEvent(...) {
            _someEvent(_logger, ...);
        }
    }
}
```

muratg:
Moving to preview2, but if you feel like this can (and should) be done quickly in preview1, feel free to bring it back!

BrennanConroy:
Prototype PR for when we do the actual change across the board: https://github.com/aspnet/SignalR/pull/1342

davidfowl:
We did this AFAIK

PR #1342, opened 2018-01-25
Move logging into class it's associated with

BrennanConroy:
Gathering feedback on this pattern
https://github.com/aspnet/SignalR/issues/1169

In `HttpClientFactory` they did it slightly differently

```c#
class HttpClientFactory
{
    private ILogger _logger;

    public HttpClientFactory(ILogger logger)
    {
        _logger = logger;
    }

    public void CreateThingy()
    {
        Log.LogCreateMessage(_logger);
    }

    private static class Log
    {
        private static readonly Action<ILogger, Exception> _logCreateMessage = LoggerMessage.Define(LogLevel.Error, 1, "Creating thingy");

        public static void LogCreateMessage(ILogger logger)
        {
            _logCreateMessage(logger, null);
        }
    }
}
```


anurse:
@rynowak This might be an unnecessary complication on the pattern you did in HttpClientFactory, but it looked like it made things slightly closer to what we do with Extension methods today. Thoughts?

@BrennanConroy Let's do the same thing in both HttpClientFactory and here. So if we're not agreed on the struct thing, let's just do the static Log class.

rynowak:
Yeah, I just made stuff static because I didn't want the allocation. I don't see a reason why you guys couldn't do it this way though. Shoot for the moon! Even if you miss, you will land among the stars.

anurse:
> because I didn't want the allocation

That's why we use a `struct`. `struct`s are **always** free. **ALWAYS**

BrennanConroy:
Alright, changing to static class and passing logger everywhere

anurse:
Yeah, I think consistency is better here. And we don't want to just bikeshed forever on the "cleanest" pattern.

BrennanConroy:
Yeah, I'll probably do it over a few days otherwise I'll go crazy

davidfowl:
This doesn't feel like it improves much tbh.

anurse:
The point is to remove the global `ILogger` extensions to avoid confusion about which methods map to which class. It's fairly minimal and safe, so I don't see a problem doing it.

davidfowl:
I dislike how it bloats the file.

davidfowl:
Maybe a partial class would be better

anurse:
> I dislike how it bloats the file.

Disagree hard. I'd rather have the log messages in the same file that they are used in. I don't care how it affects the size. I'd rather we could make the `DefineMessage` pattern smaller, but until that happens I'd rather have it in the file.

davidfowl:
I do care and I find it is extrmrly noisy especially for classes that have a lot of log messages. I’d like to see what the partial class looks like. I don’t see why it need to be in the same file as the class. It can be nested I just think it distracts from the actual content of the original file.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -605,5 +605,240 @@ private class TransferModeFeature : ITransferModeFeature
         {
             public TransferMode TransferMode { get; set; }
         }
+
+        private static class Log
+        {
+            private static readonly Action<ILogger, string, int, Exception> _preparingNonBlockingInvocation =
+                LoggerMessage.Define<string, int>(LogLevel.Trace, new EventId(1, nameof(PreparingNonBlockingInvocation)), "Preparing non-blocking invocation of '{target}', with {argumentCount} argument(s).");
```
anurse: Something @rynowak pointed out in another PR is that we shouldn't actually use `nameof` because this name is a public API of sorts. Using `nameof` means that when we rename the private function that triggers this message we could accidentally rename this public API name.

anurse:
Fine, but if I keep having to switch between a bunch of different files to add log messages I'll just spend less time spacing out my `if` statements for you :)

anurse:
> just think it distracts from the actual content of the original file.

Still super disagree with this. I spent a bunch of time going through a number of our repos trying to figure out where they were lacking in logging and lot of it came down to people not bothering to put log messages in because they had to do this dance among a bunch of files to do it "right". It makes the barrier for diagnostics higher than it should be and leaves us with very few data points.

We need to decide if we want this pattern to even be the norm. Maybe we just use the normal string-based logging and optimize only when we see problems or something, or we change our attitude around productivity and code review to say that you have to add the logging right away, even if it takes longer (see @BrennanConroy 's comments about just doing the refactoring driving him crazy), or we invest in some templates or automation, but something has to be made better or we aren't going to get anywhere near the diagnostics we need.

anurse:
This is what I'm referring to regarding removing the public method btw: https://gist.github.com/anurse/9619d59bddc7e3843d93cfb96aff6f1c

davidfowl:
@anurse I think both patterns work well when the amount of logs are small.  The above class has 24 logs and there's a tipping point where (IMO) it feels like a bunch of noise to add to the bottom of the file.

That said, lets decide on this tomorrow and move forward so we don't help @BrennanConroy 's PR hostage.

anurse:
If we're saying we only put it in a partial when it goes over a certain threshold, I can take that compromise. I just think that we obsess too much on class size as a metric but not developer experience/comprehension, which is the scenario behind that metric.

rynowak:
>  I think both patterns work well when the amount of logs are small

It sounds like you agree then that this is an improvement over what MVC does, which is total bedlam https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreLoggerExtensions.cs I haven't gone so far as to recommend that we change MVC to follow this pattern, but I'm close to doing so. 

Another crazy idea. We could use Xaml to generate the code. #ComeAtMeBro

anurse:
> We could use Xaml to generate the code.

I love it. But can we use a JSON encoding of the Xaml? Jaml? (pronounced "yeah-mul")

anurse:
Rebase?

davidfowl:
@BrennanConroy are you going to continue this?

davidfowl:
@BrennanConroy is this a real PR or is it the prototype to become the real PR?

BrennanConroy:
This was the prototype

davidfowl:
Ok, lets close it, I think we landed somewhere. We can prioritize this amongst the other things but I think this should be deferred until we do other important things.

PR #1343, opened 2018-01-25
fix #1155 by renaming signalRTokenHeader to access_token

### LINKED ISSUE #1155: Make the new bearer authentication support more standard compliant
Having built-in bearer authentication in the C#/TS clients is great but there are two annoying things in the current implementation:

  - The query string parameter name used by the TS client is non-standard and non-replaceable (the standard name is "access_token", not "signalRTokenHeader").

  - The `jwtBearer` properties/methods used in both clients have a name that assumes the token is a JWT, which is wrong as tokens are supposed to be opaque for clients.

/cc @Tratcher @davidfowl @moozzyk

anurse:
I actually started this on `release/2.1` because to be honest, it's pretty simple and I feel comfortable taking it in preview 1 if @muratg is OK with that. It's low-hanging fruit, and it's "public" API surface (kinda) so front-loading it would be nice. Also, I was reviewing the OAuth spec and noticed that it calls out how to pass access tokens via the query string in the spec (https://tools.ietf.org/html/rfc6750#section-2.3), including the query string name to use, `access_token` (as @PinpointTownes referenced in the issue #1155).

Browser functional tests have been updated and run (next up on my plate is to automate those ;)). JwtSample has also been updated and verified. The C# client doesn't ever use the query string, so the only product changes are in the Jwt sample (since we haven't decided what to "build in" for this yet) and the TS client.

Anyway, this PR is on `dev` right now. If @muratg signs off on pulling it up to preview 1, I'll retarget to `release/2.1`

### DIFF
--- a/client-ts/signalr/spec/HttpConnection.spec.ts
+++ b/client-ts/signalr/spec/HttpConnection.spec.ts
@@ -9,25 +9,29 @@ import { ITransport, TransportType, TransferMode } from "../src/Transports"
 import { eachTransport, eachEndpointUrl } from "./Common";
 import { HttpResponse } from "../src/index";
 
+const commonOptions: IHttpConnectionOptions = {
+    logger: null
+};
+
 describe("HttpConnection", () => {
     it("cannot be created with relative url if document object is not present", () => {
-        expect(() => new HttpConnection("/test"))
+        expect(() => new HttpConnection("/test", commonOptions))
             .toThrow(new Error("Cannot resolve '/test'."));
     });
 
     it("cannot be created with relative url if window object is not present", () => {
         (<any>global).window = {};
-        expect(() => new HttpConnection("/test"))
+        expect(() => new HttpConnection("/test", commonOptions))
             .toThrow(new Error("Cannot resolve '/test'."));
         delete (<any>global).window;
     });
 
     it("starting connection fails if getting id fails", async (done) => {
         let options: IHttpConnectionOptions = {
+            ...commonOptions,
             httpClient: new TestHttpClient()
                 .on("POST", r => Promise.reject("error"))
                 .on("GET", r => ""),
-            logger: null
         } as IHttpConnectionOptions;
 
         let connection = new HttpConnection("http://tempuri.org", options);
@@ -45,6 +49,7 @@ describe("HttpConnection", () => {
 
     it("cannot start a running connection", async (done) => {
         let options: IHttpConnectionOptions = {
+            ...commonOptions,
             httpClient: new TestHttpClient()
                 .on("POST", r => {
                     connection.start()
@@ -58,7 +63,6 @@ describe("HttpConnection", () => {
                         });
                     return Promise.reject("error");
                 }),
-            logger: null
         } as IHttpConnectionOptions;
 
         let connection = new HttpConnection("http://tempuri.org", options);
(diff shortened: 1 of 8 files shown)


client-ts/signalr/src/HttpConnection.ts
```diff
@@ -68,9 +68,9 @@ export class HttpConnection implements IConnection {
             }
             else {
                 let headers;
-                if (this.options.accessToken) {
+                if (this.options.accessTokenFactory) {
```
PinpointTownes: Curious: is `accessTokenFactory` allowed to return a null/empty token? Should a non-null token check be added?

client-ts/signalr/src/Transports.ts
```diff
@@ -29,21 +29,21 @@ export interface ITransport {
...
     connect(url: string, requestedTransferMode: TransferMode, connection: IConnection): Promise<TransferMode> {
 
         return new Promise<TransferMode>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
-            if (this.accessToken) {
-                let token = this.accessToken();
-                url += (url.indexOf("?") < 0 ? "?" : "&") + `signalRTokenHeader=${token}`;
+            if (this.accessTokenFactory) {
+                let token = this.accessTokenFactory();
+                url += (url.indexOf("?") < 0 ? "?" : "&") + `access_token=${token}`;
```
PinpointTownes: Same question (also, should you URLencode the token query string parameter, just in case?)

PinpointTownes:
Thanks for fixing that :+1:

client-ts/signalr/src/HttpConnection.ts
anurse: I guess it could, and I don't see why we couldn't "allow" it (and just not provide the header).

client-ts/signalr/spec/HttpConnection.spec.ts
```diff
@@ -9,25 +9,29 @@ import { ITransport, TransportType, TransferMode } from "../src/Transports"
 import { eachTransport, eachEndpointUrl } from "./Common";
 import { HttpResponse } from "../src/index";
 
+const commonOptions: IHttpConnectionOptions = {
+    logger: null
```
mikaelm12: What other options do we plan on including here in the future?

anurse:
Base changed to `release/2.1`, feedback addressed. Functional Tests actually weren't quite working, fixed that now. Also verified the JwtSample and Functional Tests with the latest version, after addressing feedback.

anurse:
FYI I addressed @PinpointTownes 's comment about the factory returning null by changing the logic a little. The constructors that accept the factories now ensure they are **always** non-null by using a "default" factory that returns null all the time if the factory itself is null. Then the logic just shifts to only writing the header/qs if the factory returns a truthy value.

mikaelm12:
Yeah, until we get some automation running, the first comment for any typescript changes should be have you run the browser tests

client-ts/signalr/src/HttpConnection.ts
```diff
@@ -42,7 +42,10 @@ export class HttpConnection implements IConnection {
     constructor(url: string, options: IHttpConnectionOptions = {}) {
         this.logger = LoggerFactory.createLogger(options.logger);
         this.baseUrl = this.resolveUrl(url);
+
         options = options || {};
+        options.accessTokenFactory = options.accessTokenFactory || (() => null);
```
mikaelm12: Would it be easier to just do a null check on the `accessTokenFactory` before creating and checking the token. It gets rid of some of the logic

mikaelm12: In a couple of places

davidfowl:
Looks mechanical

anurse:
Rebased so Travis and AppVeyor could see it. Merging when I get some green.

PR #1347, opened 2018-01-25
Fix incorrect content-type

BrennanConroy:


BrennanConroy:
@muratg Super small, preview1?

PR #1359, opened 2018-01-26
fix #1155 by renaming signalRTokenHeader to access_token (#1343)

### LINKED ISSUE #1155: Make the new bearer authentication support more standard compliant

anurse:
Doin the Pranav

### DIFF
--- a/client-ts/FunctionalTests/Startup.cs
+++ b/client-ts/FunctionalTests/Startup.cs
@@ -58,12 +58,12 @@ namespace FunctionalTests
                     {
                         OnMessageReceived = context =>
                         {
-                            var signalRTokenHeader = context.Request.Query["signalRTokenHeader"];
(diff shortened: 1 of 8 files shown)


davidfowl:
@anurse whats left here?

anurse:
Getting green builds, but I think it's just a flaky test.

davidfowl:
We need to fix our flaky tests. A bunch of our PRs are still red.

BrennanConroy:
Lets have a quick talk on Monday

PR #1365, opened 2018-01-29
React to pipeline changes

pakrym:


src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsMessageParser.cs
```diff
@@ -35,7 +36,7 @@ public ParseResult ParseMessage(ReadOnlyBuffer buffer, out Position consumed, ou
 
             while (buffer.Length > 0)
             {
-                if (ReadOnlyBuffer.Seek(start, end, out var lineEnd, ByteLF) == -1)
+                if (!(buffer.PositionOf(ByteLF) is SequencePosition lineEnd))
```
anurse: Maybe flip the conditional now that it's an `is` check?

pakrym: It makes a lot of code more nested

anurse: Ok, fair enough. Just thought that if it were quick we should do it. Go ahead and merge without changing that.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/StreamExtensions.cs
```diff
@@ -38,22 +38,22 @@ public static Task CopyToAsync(this Stream stream, IPipeWriter writer, Cancellat
         }
 
         /// <summary>
-        /// Copies a <see cref="ReadOnlyBuffer"/> to a <see cref="Stream"/> asynchronously
+        /// Copies a <see cref="ReadOnlyBuffer{T}"/> to a <see cref="Stream"/> asynchronously
         /// </summary>
-        /// <param name="buffer">The <see cref="ReadOnlyBuffer"/> to copy</param>
+        /// <param name="buffer">The <see cref="ReadOnlyBuffer{T}"/> to copy</param>
```
BrennanConroy: T?

BrennanConroy:
Don't forget to merge back to dev

Issue #1357, opened 2018-01-26
Suggestion - Change default PollTimeout for LongPolling


gary-holland:
For the LongPolling transport I've noticed the default timeout is set to 110 seconds.  

The issue I had is that CloudFlare have a default timeout of ~100 seconds ([details](https://support.cloudflare.com/hc/en-us/articles/200171926-Error-524-A-timeout-occurred)) which was causing my client to error intermittently due to CloudFlares timeout.  

I am aware that this can be overridden in the configure section using the following (which has resolved issue for me).

```
app.UseSignalR(routes =>
	{
		routes.MapHub<RemoteAgentHub>("remoteagent", options =>
		{
			options.LongPolling.PollTimeout = TimeSpan.FromSeconds(60);
		});
	});
```

My suggestion is that you drop the default from 110 seconds to something below 100 seconds, as the current configuration means that everyone hosting a site behind Cloudflare will encounter this issue with the default configuration.

anurse:
I'm not sure what drove the decision to use 120 seconds, it may have been what we used for the previous versions of SignalR, so we'll need to look in to the reasons behind that, but I agree, it would be good if we can pick a default that doesn't cause CloudFlare to timeout, since it is so common. Thanks for the report!

mikaelm12:
> I'm not sure what drove the decision to use 110 seconds

Wasn't it because of the default timeout in IIS being like 120 seconds? I remember something like that

anurse:
Ah, ok, that makes sense. I think 90 or 100 seconds would be reasonable based on CloudFlare's limit. CloudFlare is a fairly common proxy to use.

davidfowl:
Sounds fine.

mikaelm12:
Done: e214d5cdfac12475f0d5c7a1b6a5e063a61870e1
Thanks for the suggestion @gary-holland

PR #538, opened 2017-06-07
Add support for timing out poll requests

davidfowl:
- Default poll request is 110 seconds (like in previous versions of SignalR)
- Changed the "stop polling" status code to 205 and the time out
status code to 204
- Added LongPollingOptions to HttpSocketOptions

#500

PR #1368, opened 2018-01-29
Reduce LongPolling timeout to accommodate Cloudflare's timeout

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1357
Cloudflare has a timeout of only 100 seconds after which they close the connection. Detailed here: https://support.cloudflare.com/hc/en-us/articles/200171926-Error-524-A-timeout-occurred

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/LongPollingOptions.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/LongPollingOptions.cs
@@ -6,6 +6,6 @@ namespace Microsoft.AspNetCore.Sockets
 {
     public class LongPollingOptions
     {
-        public TimeSpan PollTimeout { get; set; } = TimeSpan.FromSeconds(110);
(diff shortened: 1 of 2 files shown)


anurse:
Hrm. A little concerned that we don't have a test for the actual timeout behavior. The test in here is a little contrived...

davidfowl:
That’s unrelated to this change. We should have to test to very that timeouts work  (I’m surprised we don’t tbh)

anurse:
Yeah, I agree. Just pointing that out.

davidfowl:
We do have a test:

https://github.com/aspnet/SignalR/pull/538/files#diff-feef691b2b3ce98f9972dd0ab49dea5eR316

anurse:
Yeah, and I suppose it isn't really feasible to test the exact timeout because of how our unit tests run. Ok, I'm good with what we've got so far.

PR #1384, opened 2018-01-31
Workaround flakiness in compiling the ChatSample

natemcmaster:
Compilation fails sometimes because it's trying to bundle/minify twice at the same time.

> C:\b\w\4ae60d3b5c30e105\modules\SignalR\samples\ChatSample\wwwroot\css\site.min.css : Bundler & Minifier error 0: The process cannot access the file 'C:\b\w\4ae60d3b5c30e105\modules\SignalR\samples\ChatSample\wwwroot\css\site.min.css' because it is being used by another process.


anurse:
Is there somewhere to track this in the bundler/minifier itself as well? I'm not even sure where this package came from, but it probably shouldn't have this kind of race...

natemcmaster:
https://github.com/madskristensen/BundlerMinifier

The repo doesn't look like it's very active, so I'm favoring working around for now - cc @madskristensen

Issue #1385, opened 2018-01-31
Flaky tests: HubConnectionCanSendAndReceiveMessages


ryanbrandenburg:
I set `HubConnectionCanSendAndReceiveMessages` and `HubConnectionCanSendAndReceiveGroupMessages` to ignore for now because they [fail](http://aspnetci/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=397066&_focus=84625) on Ubuntu sometimes. It should be investigated and re-enabled once fixed.

ryanbrandenburg:
I talked this over with Brennan and we proved that it's a CI agent problem that we think we know how to solve.

PR #1386, opened 2018-01-31
Skip flaky tests

ryanbrandenburg:
aspnet/SignalR#1385.


ryanbrandenburg:
I talked this over with Brennan and we proved that it's a CI agent problem that we think we know how to solve.

PR #1393, opened 2018-02-01
fix #1387 by removing our hacky EventSource typings

### LINKED ISSUE #1387: Typescript 2.7 breaks our build
Was getting an error that `EventSource` is already defined when building "client-ts/signalr"

anurse:
Thanks to https://github.com/Microsoft/TSJS-lib-generator/pull/341 we don't need this anymore if we update to the latest TypeScript.


BrennanConroy:
:shipit:

PR #1397, opened 2018-02-01
Fix flaky test

BrennanConroy:
Registering `Closed` callback after connection started would be a race between the timeout and the register


davidfowl:
Is this the only place we do this?

BrennanConroy:
It looks like it after a quick search

PR #1402, opened 2018-02-02
Remove some test flakiness and add logging

BrennanConroy:


test/xunit.runner.json
```diff
@@ -1,4 +1,5 @@
 {
   "longRunningTestSeconds": 30,
-  "diagnosticMessages": true
+  "diagnosticMessages": true,
+  "maxParallelThreads": -1
```
anurse: Turns out this isn't actually disabling the sync context :(.

PR #1403, opened 2018-02-02
Flake Hunters Part 2: Functional Test startup and more.

anurse:
image

* Use port `0` for functional test server. We are running in-proc, so why encourage a race on the port number
* Lots of logging to see if we can track down more flaky test problems
* Also removed the "ForwardingLogProvider" because it wasn't really working. Will have to revisit this because we really want to try and get the server logs for a particular test invocation.


anurse:
This isn't getting anywhere, closing for now, will start up a new investigation.

PR #1409, opened 2018-02-05
Increase ServerFixture startup timeout

anurse:
Also adds some code to capture logs during startup and add them to the exception message if the startup times out.

It looks like it just sometimes takes longer than 5 seconds to start up on AppVeyor. Given my own tests on a 2 core Azure VM, this isn't terribly surprising.

Also, switched to using port `0` when creating the server to auto-allocate a port.

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
@@ -61,7 +61,9 @@ namespace Microsoft.AspNetCore.SignalR.Tests
             var t = Task.Run(() => _host.Start());
             _logger.LogInformation("Starting test server...");
             _lifetime = _host.Services.GetRequiredService<IApplicationLifetime>();
-            if (!_lifetime.ApplicationStarted.WaitHandle.WaitOne(TimeSpan.FromSeconds(5)))
+
+            // This only happens once per fixture, so we can afford to wait a little bit on it.
+            if (!_lifetime.ApplicationStarted.WaitHandle.WaitOne(TimeSpan.FromSeconds(20)))
             {
                 // t probably faulted
                 if (t.IsFaulted)
--- a/test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
@@ -51,7 +51,9 @@ namespace Microsoft.AspNetCore.SignalR.Tests
             var url = "http://127.0.0.1:0";
 
             _host = new WebHostBuilder()
-                .ConfigureLogging(builder => builder.AddProvider(_logSinkProvider))
+                .ConfigureLogging(builder => builder
+                    .AddProvider(_logSinkProvider)
+                    .AddProvider(new ForwardingLoggerProvider(_loggerFactory)))
                 .UseStartup(typeof(TStartup))
                 .UseKestrel()
                 .UseUrls(url)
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -61,118 +61,114 @@ private void StartServer(string url)
...
 
             public ILogger CreateLogger(string categoryName)
             {
-                return new AsyncLocalForwardingLogger(categoryName, _localLogger);
+                return new LogSinkLogger(categoryName, this);
             }
 
             public void Dispose()
             {
+                throw new NotImplementedException();
```
BrennanConroy: ?

test/Microsoft.AspNetCore.SignalR.Tests/Startup.cs
```diff
@@ -16,7 +18,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddEndPoint<EchoEndPoint>();
         }
 
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILogger<Startup> logger)
```
BrennanConroy: Artifact? Remove?

test/Microsoft.AspNetCore.SignalR.Tests/Startup.cs
```diff
@@ -1,9 +1,11 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Threading;
```
BrennanConroy: Need these?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -23,35 +26,34 @@ public class ServerFixture<TStartup> : IDisposable
...
-        }
+            // We're using 127.0.0.1 instead of localhost to ensure that we use IPV4 across different OSes
+            var url = "http://127.0.0.1:0";
 
-        private void StartServer(string url)
-        {
             _host = new WebHostBuilder()
-                .ConfigureLogging(builder => builder.AddProvider(new ForwardingLoggerProvider(_loggerFactory)))
+                .ConfigureLogging(builder => builder
+                    .AddProvider(_logSinkProvider)
```
BrennanConroy: Doesn't this provider get added twice now?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -61,119 +63,133 @@ private void StartServer(string url)
...
             {
                 // t probably faulted
                 if (t.IsFaulted)
                 {
                     throw t.Exception.InnerException;
                 }
-                throw new TimeoutException("Timed out waiting for application to start.");
+
+                var logs = _logSinkProvider.GetLogs();
+                throw new TimeoutException($"Timed out waiting for application to start. {Environment.NewLine}Startup Logs:{Environment.NewLine}{RenderLogs(logs)}");
```
BrennanConroy: super nit: Don't need the space after "start."

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
anurse: Yeah, but it actually isn't needed above. I'll remove it there.

PR #1410, opened 2018-02-06
Unskip tests

mikaelm12:
@BrennanConroy Added some logging and other fixes so we want to see the results on TeamCity


ryanbrandenburg:
Sounds good to me. The Travis failure contains my next project though.

anurse:
The Travis failure is the thing we're trying to get an xunit fix in for.

PR #1411, opened 2018-02-06
Publish test logs to AppVeyor Artifacts

anurse:
https://www.appveyor.com/docs/packaging-artifacts/

screen shot 2018-02-06 at 10 02 29 am

screen shot 2018-02-06 at 10 03 54 am

This gives us test logs as artifacts in AppVeyor which makes it way easier to review flaky test failures.

TODO:
* [x] The `.testlogs` directory means that the zip contains a `.testlogs` directory, which is a pain on Unix. I'm going to remove the leading `.` and verify that it still works fine.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,6 +10,11 @@ branches:
     - /^rel\/.*/
 build_script:
   - ps: .\run.ps1 default-build
+  - ps: '7z a testlogs.zip .testlogs'
+artifacts:
+  - path: testlogs.zip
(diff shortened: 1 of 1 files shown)


ryanbrandenburg:
Awesome! We'll probably do something similar to all other repos the next time we mass-update all repos.

anurse:
Pretty sure this won't upload test logs on failure. Testing now and I should be able to find a way to make it work.

PR #1413, opened 2018-02-06
remove bundler/minifier from ChatSample

anurse:
It doesn't like multi-targeted projects and causes file locking issues on build, making it flaky.

I also removed the `less` file since we don't really need to use it in a sample. Similarly, I removed reference to the `.min.css` file in the Layout page.

**NOTE** The diff got a little borked. All I did was **delete** the old `site.css` (generated from the `less` file) and rename `site.less` to `site.css`, removing the leading `@import` statement. We had two copies of bootstrap because we were bringing it in via a `<link>` tag on the view **and** we were embedding it in our `css` file via the less `@import` command.


anurse:
Travis failure is definitely timeout issues. AppVeyor is not one we've seen before but it does create a bunch of different tasks and depends upon them completing in a timely fashion, so it could be the same issue. Rerunning CIs.

davidfowl:
@anurse Merge this

PR #1414, opened 2018-02-06
Move AppVeyor artifact publishing to on_finish to ensure it runs on failed builds too

anurse:
I think the `7z` command won't run when the `run.ps1` script fails, so the test logs won't be packaged when the tests fail. I think we can fix this, so I'm opening this PR to verify and fix this.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,6 +10,7 @@ branches:
     - /^rel\/.*/
 build_script:
   - ps: .\run.ps1 default-build
+after_build:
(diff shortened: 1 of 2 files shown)


anurse:
I was right, and have now fixed it. Adding reviewers and removing WIP label.

anurse:
Only failure is flaky timeout tests, which we're waiting on a fix for. Pinged the xunit PR to see if we can get it merged so we could update SignalR to use the preview build of that just to pilot it out :)

PR #1425, opened 2018-02-09
Exit receive loop faster when connection closed

BrennanConroy:
The issue is, under high load, the client would have a large queue of items in the Channel and it would take a long time to drain the items on close. There is no need to drain (that I know of), we can just exit and let GC clean it up

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -422,6 +422,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 }
 
                 await Input.Completion;
+                _transportChannel = null;
             }
             catch (Exception ex)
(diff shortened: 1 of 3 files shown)


moozzyk:
AFAIR it was related to `Input.Completion` not being completed until you drain so double check this because all hinges on this.

BrennanConroy:
Oh dang, that's unfortunate, will have to use a cancellation token instead
Thanks @moozzyk !

davidfowl:
@BrennanConroy a test should have failed here.

BrennanConroy:
This is difficult to write a test for, we'd need to inject our own transport and have a custom channel implementation for it. I'll look at writing one when I correctly fix it

davidfowl:
Does it work?

BrennanConroy:
I'm not sure, we can talk about it at work

davidfowl:
Working on removing the event queue completely.

PR #1426, opened 2018-02-09
Fixed exception during deserialization of a message with Redis

Dona278:
Changed wrong message type used during deserialization in "SubscribeToUser"

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -533,9 +533,9 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             {
                 try
                 {
-                    var message = DeserializeMessage<HubInvocationMessage>(data);
+                    var message = DeserializeMessage<RedisInvocationMessage>(data);
 
-                    await connection.WriteAsync(message);
+                    await connection.WriteAsync(message.CreateInvocation());
                 }
                 catch (Exception ex)
                 {


davidfowl:
Why aren’t there any test updates?

BrennanConroy:
@Dona278 Good catch! Could you add a test here https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs please?

Dona278:
@BrennanConroy Unfortunately SendUser needs a UserIdentifier to work ([code](https://github.com/aspnet/SignalR/blob/0435b6dc6e0ecae6920d1aca6cd7117dfe7cd307/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs#L106)) which is set only if an HubEndpoint calls NegotiateAsync ([code](https://github.com/aspnet/SignalR/blob/94155b0e89888298d2abc62a00a424d604599211/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs#L178)).  
There are no redis tests that use HubEndpoint in RedisHubLifetimeManagerTests, is it the correct file to add them?  
Is there a test file I can copy from? The RedisEndToEndTests seems the right class for endpoint testing.

BrennanConroy:
You can just set the `UserIdentifier` on the connection for testing
```
var connection1 = HubConnectionContextUtils.Create(client1.Connection);
connection1.UserIdentifier = "Bob";
```

Dona278:
The property is private set and is not mockable in ([HubConnectionContext.cs:72](https://github.com/aspnet/SignalR/blob/94155b0e89888298d2abc62a00a424d604599211/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs#L72))

BrennanConroy:
Adding a test in `RedisEndToEndTests` seems best

Dona278:
Ok, i tried to add the test into RedisEndToEndTests and here I only have a HubConnection got through `CreateConnection()` present at the end of the file but the instance of `RedisHubLifetimeManager` (that i need to test the `SendUser` and `SendUsers` behaviours) needs an `HubConnectionContext` in his `OnConnectedAsync` in which subscribings are performed and seems there are no relations between HubConnection and HubConnectionContext. So, in `RedisHubLifetimeManagerTests` I cannot use `manager.SendUser` because the user connection hasn't the UserIdentifier set and the property has private set, in `RedisEndToEndTests` I cannot use `RedisHubLifetimeManager` because I have different connection types, I'm wrong? Or this is the reasons why only User tests are missing?

Thanks

BrennanConroy:
Oh, yeah that's unfortunate. Makes sense why we didn't add a test. I think I'll merge this PR and build a new test suite that enables testing this feature

anurse:
@BrennanConroy give this a manual smoke-test and then merge it. Make sure an issue is filed for adding an automated test to cover this scenario.

BrennanConroy:
#1447 filed

BrennanConroy:
Thanks @Dona278

buvinghausen:
Hey I know this is probably not the place to ask a road-map question but this PR contains the fix we need in order to use SignalR in our app.  Is there any timeline on when the next public preview/RC for .NET Core 2.1 will drop? I'd prefer not to go into the daily build matrix but we really would love to start using SignalR to get the events back out to the browsers from our message queues.

anurse:
@buvinghausen there is no public timeline for the next preview, but it is coming soon! The next release will be a preview release (Preview 2)

buvinghausen:
Thanks Andrew! Looking forward to it!

PR #1427, opened 2018-02-10
Run tests outside of xunit synccontext

BrennanConroy:


davidfowl:
@BrennanConroy seems like this made things worse. I'm seeing lots of failures locally.

BrennanConroy:
>Works on my machine

anurse:
🎉

PR #1433, opened 2018-02-11
Don't allocate all the memory up front when receiving incoming sends

davidfowl:


davidfowl:
I was thinking the same thing. I don't do all of it though, the biggest cost will still be serialization

PR #1435, opened 2018-02-12
Client pipelines

davidfowl:
- Reworked the Client to be based on pipelines instead of Channels
- SendAsync no longer fails if the http request itself fails but the connection is closed as a result.
- Updated tests

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
@@ -10,16 +10,17 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Encoders
 {
     public class Base64Encoder : IDataEncoder
     {
-        public ReadOnlySpan<byte> Decode(byte[] payload)
+        public bool TryDecode(ref ReadOnlySpan<byte> buffer, out ReadOnlySpan<byte> data)
         {
-            ReadOnlySpan<byte> buffer = payload;
-            LengthPrefixedTextMessageParser.TryParseMessage(ref buffer, out var message);
-
-            Span<byte> decoded = new byte[Base64.GetMaxDecodedFromUtf8Length(message.Length)];
-            var status = Base64.DecodeFromUtf8(message, decoded, out _, out var written);
-            Debug.Assert(status == OperationStatus.Done);
-
-            return decoded.Slice(0, written);
+            if (LengthPrefixedTextMessageParser.TryParseMessage(ref buffer, out var message))
+            {
+                Span<byte> decoded = new byte[Base64.GetMaxDecodedFromUtf8Length(message.Length)];
+                var status = Base64.DecodeFromUtf8(message, decoded, out _, out var written);
+                Debug.Assert(status == OperationStatus.Done);
+                data = decoded.Slice(0, written);
+                return true;
+            }
+            return false;
         }
 
         private const int Int32OverflowLength = 10;
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -3,12 +3,10 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Threading.Channels;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Sockets.Client.Http;
 using Microsoft.AspNetCore.Sockets.Client.Internal;
@@ -18,6 +16,7 @@ using Microsoft.AspNetCore.Sockets.Internal;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Newtonsoft.Json;
+using System.IO.Pipelines;
 
 namespace Microsoft.AspNetCore.Sockets.Client
 {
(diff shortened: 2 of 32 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -83,67 +81,67 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
...
+                    _logger.ReceiveStopped();
+                }
             }
         }
 
         public async Task StopAsync()
         {
             _logger.TransportStopping();
             _transportCts.Cancel();
-            _application.Writer.TryComplete();
```
BrennanConroy: Why is this gone?

test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -166,13 +160,19 @@ public class LongPollingTransportTests
...
                     var exception =
-                        await Assert.ThrowsAsync<HttpRequestException>(async () => await transportToConnection.Reader.Completion.OrTimeout());
+                        await Assert.ThrowsAsync<HttpRequestException>(async () =>
+                        {
+                            async Task ReadAsync()
+                            {
+                                await pair.Transport.Input.ReadAsync();
+                            }
+
+                            await ReadAsync().OrTimeout();
```
BrennanConroy: Looks like overkill o.O

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -237,7 +236,8 @@ private Task StartAsyncCore()
                         // Suppress (but log) the exception, this is user code
                         _logger.ErrorDuringClosedEvent(ex);
                     }
-                });
+
+                }, null);
```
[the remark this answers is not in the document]
davidfowl: Yes but the code was always capturing. I didn't want the diff to be big so I left it as is. Will change it in another PR (lots of low hanging fruit 😄 )

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -122,17 +123,11 @@ public Task StartAsync(Uri url, Channel<byte[], SendMessage> application, Transf
...
                         if (payload.Length > 0)
                         {
-                            while (!_application.Writer.TryWrite(payload))
-                            {
-                                if (cancellationToken.IsCancellationRequested || !await _application.Writer.WaitToWriteAsync(cancellationToken))
-                                {
-                                    return;
-                                }
-                            }
+                            await _application.Output.WriteAsync(payload);
```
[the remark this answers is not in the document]
davidfowl: https://github.com/dotnet/corefx/pull/27007#discussion_r167422800

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
davidfowl: It isn't required. All completion happens in here now https://github.com/aspnet/SignalR/pull/1435/files/c34463436737ee897219649a4e61c56e8f646346#diff-579fedec848db9fecbd95f172d7654e7R66. Both tasks should always end and we should clean up in the same place.

test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
davidfowl: This was just to get a `Task` from the pipe awaiter.

davidfowl: I could use one of the helpers I used in the other tests.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Encoders/Base64Encoder.cs
```diff
@@ -10,16 +10,17 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Encoders
 {
     public class Base64Encoder : IDataEncoder
     {
-        public ReadOnlySpan<byte> Decode(byte[] payload)
+        public bool TryDecode(ref ReadOnlySpan<byte> buffer, out ReadOnlySpan<byte> data)
```
BrennanConroy: Add a test, and also fix the compilation errors :P

davidfowl: Yea doing this in VS code 😄. I'll spend more time looking at this soon.

davidfowl:
ZOMG green!

BrennanConroy:
But will it blend? I mean, will it pass if you run it again?

davidfowl:
Lets see.

davidfowl:
@BrennanConroy we need to add logging to a few key places to help diagnose issues:
- SSE Transport and LongPolling transport
- HubConnection needs some logs when it's processing messages (how many it parsed)
- Should also show print the message when tracing.

BrennanConroy:
I also found an issue in a couple of our logs at Trace level

davidfowl:
Argggg, OSX!!! 😄

davidfowl:
This test, failed twice now.

- ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport

@BrennanConroy can we get artifacts from travis?

BrennanConroy:
Nope, they only support pushing to an AWS subscription you own

davidfowl:
Well I'm going to give this a few more goes then rage merge. Things need to be cleaned up. All of the parsers need to be rewritten.

PR #1443, opened 2018-02-13
Remove unneeded dependencies

davidfowl:


src/Microsoft.AspNetCore.Sockets.Abstractions/ChannelConnection.cs
```diff
@@ -1,63 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-using System.Threading.Channels;
-
-namespace Microsoft.AspNetCore.Sockets.Internal
-{
-    public static class ChannelConnection
```
anurse: I presume we weren't actually using this anywhere anymore?

anurse:
Who approved this?

src/Microsoft.AspNetCore.Sockets.Abstractions/ChannelConnection.cs
davidfowl: The build wouldn't have passed right 😉

PR #1446, opened 2018-02-14
Change the IConnection contract to be an IDuplexPipe

davidfowl:
This is a design change that we considered in the past but moved away from while we were discussiong messaging vs streams. We're firmly in the streams camp now so I think we should make this change (it also cleans and a bunch of things and introduces problems 😄 ).

- The IConnection implements IDuplexPipe, it's a very simple contract, the IConnection is responsible for writing to the IDuplex pipe that is exposed. 
- We already changed the HttpConnection to use pipelines internal so the change there was deleting code. 
- This change is also required to support SignalR on top of TCP.
- The server and client dispatching loop look more similar.
- We were able to remove all callback handling from this layer.
- The event queue is gone, the pipe is single producer and single consumer. The hub connection layer is the only one that has to implement callbacks.
- With this change we don't allow running client callbacks in parallel (not sure why we even bothered before) so we were able to remove the event queue.

Quirks:
- Completing the pipe is messy, do we still need StopAsync and DisposeAsync? 
- Who is responsible for completing the PipeReader and PipeWriter exposed by the connection? At a minimum the IConnection impl needs to complete the application sides.
- ~No errors if the you write to the connection before it starts or after it is disposed. This is possible but requires wrapping the PipeReader and PipeWriter returned.~ (solved)
- ~The current implementation gives you a null pipe if you you don't call Start first, this can be fixed by making a wrapping PipeWriter and PipeReader.~ (solved)
- IConnection.DisposeAsync doesn't own the receive loop anymore so it returns before the loop ended. This might be fixable if we hook the right events. This also makes it hard to reuse pipes.(not solved)
- Today we have an ill defined contract as to which part of the client is reusable, is that part of the IConnection contract? This might make that a bit tricker.


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -207,19 +206,21 @@ public class HubConnectionTests
...
-            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
-            await hubConnection.StartAsync().OrTimeout();
-            await hubConnection.DisposeAsync().OrTimeout();
-
-            // Fire callbacks, they shouldn't fail
-            foreach (var registration in connection.Callbacks)
-            {
-                await registration.InvokeAsync(new byte[0]);
-            }
+            var connection = new Mock<IConnection>();
```
davidfowl: /cc @pakrym this is what it looks like to mock a pipe 😄

pakrym: You can implement completed IPipeAwaiter very easy, it would make this mock much cleaner.

Maybe we would end up with a complete mock pipe in a bit.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -31,24 +32,24 @@ public class HttpConnection : IConnection
         private readonly object _stateChangeLock = new object();
 
         private volatile IDuplexPipe _transportChannel;
```
mikaelm12: Should we not rename away from channel?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.PipeReader.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
```
mikaelm12: Header

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.PipeWriter.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System;
```
mikaelm12: Header

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -341,47 +353,82 @@ private Task InvokeStreamCore(string methodName, InvocationRequest irq, object[]
...
+                            if (_protocolReaderWriter.ReadMessages(buffer, _binder, out var messages, out consumed, out examined))
                             {
-                                _logger.DropStreamMessage(streamItem.InvocationId);
-                                return;
+                                foreach (var message in messages)
+                                {
+                                    InvocationRequest irq;
+                                    switch (message)
+                                    {
+                                        case InvocationMessage invocation:
```
mikaelm12: nit: Can we get some new lines in here

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
```diff
@@ -22,11 +23,11 @@ internal static class SignalRClientLoggerExtensions
...
 
-        private static readonly Action<ILogger, string, Exception> _sendInvocation =
-            LoggerMessage.Define<string>(LogLevel.Debug, new EventId(5, nameof(SendInvocation)), "Sending Invocation '{invocationId}'.");
+        private static readonly Action<ILogger, string, HubMessage, Exception> _sendInvocation =
+            LoggerMessage.Define<string, HubMessage>(LogLevel.Debug, new EventId(5, nameof(SendInvocation)), "Sending Invocation '{invocationId}'. {payload}");
 
-        private static readonly Action<ILogger, string, Exception> _sendInvocationCompleted =
-            LoggerMessage.Define<string>(LogLevel.Debug, new EventId(6, nameof(SendInvocationCompleted)), "Sending Invocation '{invocationId}' completed.");
+        private static readonly Action<ILogger, string, HubMessage, Exception> _sendInvocationCompleted =
+            LoggerMessage.Define<string, HubMessage>(LogLevel.Debug, new EventId(6, nameof(SendInvocationCompleted)), "Sending Invocation '{invocationId}' completed. {payload}");
```
anurse: Still super unsure about putting message payloads in logs. They potentially have sensitive information in them.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.PipeReader.cs
```diff
@@ -0,0 +1,55 @@
...
+            private readonly HttpConnection _connection;
+
+            public HttpConnectionPipeReader(HttpConnection connection)
+            {
+                _connection = connection;
+            }
+
+            public override void AdvanceTo(SequencePosition consumed)
+            {
+                _connection._transportChannel.Input.AdvanceTo(consumed);
```
anurse: This will NRE if the connection hasn't been started. We should put in a check that provides a better error message.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
anurse: [](https://en.wikipedia.org/wiki/The_Treachery_of_Images)

anurse:
> Today we have an ill defined contract as to which part of the client is reusable, is that part of the IConnection contract? This might make that a bit tricker.

I wonder if maybe `IConnection` should be non-restartable and we move the restart logic to `HubConnection`. After all, the main reason is to keep Hub callbacks bound. It would mean `HubConnection` needs an `IConnection`-factory rather than an `IConnection` but I think that's cleaner. It also means the pipelines, transports, etc. all just get dumped when a reconnect happens.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -341,47 +363,82 @@ private Task InvokeStreamCore(string methodName, InvocationRequest irq, object[]
...
 
-        private async Task OnDataReceivedAsync(byte[] data)
+        private async Task StartReading()
         {
-            ResetTimeoutTimer();
-            if (_protocolReaderWriter.ReadMessages(data, _binder, out var messages))
+            try
             {
-                foreach (var message in messages)
+                while (true)
```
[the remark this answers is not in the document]
BrennanConroy: We should add a couple cancellation checks in the below code as well, maybe after every message

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.PipeReader.cs
davidfowl: I don’t want to put checks in all of the methods here since some of performance critical. I’ll see what makes sense and update.

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
davidfowl: Yea the current logging isn’t great but that’s a good point

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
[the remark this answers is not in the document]
davidfowl: Nah. I’m going to do a pass removing all of the tokens. I hate the fact that it throws an exception for expected behavior.

davidfowl:
Agree @anurse but not as part of this change. It also means your paying for negotiation each time. But maybe that’s ok.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -653,6 +496,20 @@ private ConnectionState ChangeState(ConnectionState to)
...
+            if (_connectionState != ConnectionState.Connected)
+            {
+                throw new InvalidOperationException(
+                "Cannot send messages when the connection is not in the Connected state.");
+            }
+        }
+
+        void IDisposable.Dispose()
+        {
+            // We want to hide this since it noops, it's temporary until we remove IDisposable from IDuplexPipe
```
mikaelm12: File this somewhere so it doesn't get overlooked later

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.SendAsync.cs
```diff
@@ -86,11 +87,12 @@ public Task SendThrowsIfConnectionIsDisposed()
                         await connection.DisposeAsync().OrTimeout();
 
                         var exception = await Assert.ThrowsAsync<InvalidOperationException>(
-                            () => connection.SendAsync(new byte[0]).OrTimeout());
+                            () => connection.Output.WriteAsync(new byte[0]).OrTimeout());
                         Assert.Equal("Cannot send messages when the connection is not in the Connected state.", exception.Message);
                     });
             }
 
+
```
mikaelm12: nit: Remove new line

anurse:
> It also means your paying for negotiation each time

Only for WebSockets, for the other transports you have to pay that cost on reconnect anyway. In theory we could try to detect that WebSockets worked and just do a bare WebSocket connection on reconnect, but that would make transport fallback on reconnect tricky. I think the cost is minimal and acceptable.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.PipeReader.cs
anurse: `catch(NullReferenceException)` :trollface:

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/SignalRClientLoggerExtensions.cs
anurse: We do it in the JS client too. I'm filing an issue for that because I think this it's generally bad. Let's not add it here though. We can add in attribute **about** the message (type, probably even method name) but not the argument values.

anurse: https://github.com/aspnet/SignalR/issues/1449

anurse:
https://github.com/aspnet/SignalR/issues/1450

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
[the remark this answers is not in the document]
davidfowl: Most of the pipelines loops look like this FWIW

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: This is literally copied from what was here before. Where do you want the newlines and does our other code look like that?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: https://github.com/dotnet/corefx/issues/27154

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -140,9 +141,13 @@ private void ResetTimeoutTimer()
             using (var memoryStream = new MemoryStream())
             {
                 NegotiationProtocol.WriteMessage(new NegotiationMessage(_protocol.Name), memoryStream);
-                await _connection.SendAsync(memoryStream.ToArray(), _connectionActive.Token);
+
+                // TODO: Pass the token when that's available
```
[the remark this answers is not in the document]
davidfowl: https://github.com/aspnet/SignalR/issues/1451

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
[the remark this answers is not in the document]
anurse: > I hate the fact that it throws an exception for expected behavior.

Only if you use `ThrowIfCancellationRequested`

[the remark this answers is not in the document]
davidfowl: That is the contract for cancellation tokens. The operation is cancelled via an exception. Pipelines has an alternative model that I plan to change to once all the code is moved.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -162,14 +167,24 @@ private IDataEncoder GetDataEncoder(TransferMode requestedTransferMode, Transfer
 
         public async Task StopAsync() => await StopAsyncCore().ForceAsync();
 
-        private Task StopAsyncCore() => _connection.StopAsync();
+        private async Task StopAsyncCore()
+        {
+            await _connection.StopAsync();
+
+            if (_readingTask != null)
```
BrennanConroy: I think you have to capture `_readingTask` before calling `_connection.StopAsync` otherwise someone could start a new connection in the closed callback and set `_readingTask` for the new connection

davidfowl: Yea, this part of the code is racy. I took a look last night and since the receive loop moved out of the place that had all of the locks and synchronization it's a bit messy.

davidfowl:
I'll be back with something else

PR #1455, opened 2018-02-15
First pass fallback. 

mikaelm12:
The idea is to only fallback when there is an issue in start on the HttpConnection. 
I exposed the Transport type so that as transports fail we can remove them from the considered transports. 
Wanted to get some early thoughts. 
Of course needs tests eventually, but it's still WIP

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -209,7 +209,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 
                 if (_transport == null)
                 {
-                    throw new InvalidOperationException("No transport was created.");
+                    throw new InvalidOperationException("Unable to connect to the server with any of the available transports.");
                 }
             }
 
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -47,6 +47,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private PipeWriter Output => _transportChannel.Output;
         private readonly List<ReceiveCallback> _callbacks = new List<ReceiveCallback>();
         private readonly TransportType _requestedTransportType = TransportType.All;
+        private TransportType _serverTransports = TransportType.All;
+        // The order of the transports here is the order determines the fallback order.
+        private static readonly TransportType[] AllTransports = new[]{ TransportType.WebSockets, TransportType.ServerSentEvents, TransportType.LongPolling };
         private readonly ConnectionLogScope _logScope;
         private readonly IDisposable _scopeDisposable;
 
@@ -142,6 +145,14 @@ namespace Microsoft.AspNetCore.Sockets.Client
             return _startTcs.Task;
         }
 
+        private async Task<NegotiationResponse> GetNegotiationResponse()
+        {
+            var negotiationResponse = await Negotiate(Url, _httpClient, _logger);
+            _connectionId = negotiationResponse.ConnectionId;
+            _logScope.ConnectionId = _connectionId;
+            return negotiationResponse;
+        }
+
         private async Task StartAsyncInternal()
         {
             Log.HttpConnectionStarting(_logger);
@@ -151,13 +162,13 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 var connectUrl = Url;
                 if (_requestedTransportType == TransportType.WebSockets)
                 {
-                    _transport = _transportFactory.CreateTransport(TransportType.WebSockets);
+                    Log.StartingTransport(_logger, _requestedTransportType, connectUrl);
+                    _logger.StartingTransport(_requestedTransportType, connectUrl);
+                    await StartTransport(connectUrl, _requestedTransportType);
                 }
                 else
                 {
-                    var negotiationResponse = await Negotiate(Url, _httpClient, _logger);
-                    _connectionId = negotiationResponse.ConnectionId;
-                    _logScope.ConnectionId = _connectionId;
+                    var negotiationResponse = await GetNegotiationResponse();
 
                     // Connection is being disposed while start was in progress
                     if (_connectionState == ConnectionState.Disposed)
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -164,13 +165,33 @@ private Task StartAsyncCore()
...
-
-                    _transport = _transportFactory.CreateTransport(GetAvailableServerTransports(negotiationResponse));
+                    _serverTransports = GetAvailableServerTransports(negotiationResponse);
+                    _transport = _transportFactory.CreateTransport(_serverTransports);
                     connectUrl = CreateConnectUrl(Url, negotiationResponse);
                 }
 
                 _logger.StartingTransport(_transport, connectUrl);
-                await StartTransport(connectUrl);
+                var validTransports = _requestedTransportType & _serverTransports;
```
anurse: The bit-twiddling is clever, but I think I'd prefer something a little more clear-cut. My thought is that we have an ordered list of transport types internally, and we iterate through that. For each transport type we check it against the `_requestedTransportType` and `_serverTransports` masks to see if we should try it. If it passes that mask, we try it and if it fails we move on to the next one.

Something like:

```csharp
var allowedTransports = _requestedTransportType & _serverTransports;
foreach(var transport in KnownTransports)
{
    if ((transport & allowedTransports) != 0) {
        try {
            await StartTransport(connectUri, transport);
            // We succeeded, exit the loop
            break;
        } catch(Exception ex) {
            // Log and just allow the loop to continue.
        }
    }
}

// If we haven't started a transport yet, we fail here because we're out of transports to try.
```

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -164,13 +165,33 @@ private Task StartAsyncCore()
...
+                while (validTransports > 0)
+                {
+                    try
+                    {
+                        await StartTransport(connectUrl);
```
anurse: We should pass the transport we want to try in here directly rather than assuming it will pick the next applicable transport.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
mikaelm12: Much cleaner. This saves us from having to expose the transport type on ITransport as well.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -136,13 +137,10 @@ Task OnTransportStart()
                             transport: new TestTransport(onTransportStart: OnTransportStart)),
                         async (connection, closed) =>
                     {
-                        var actual = await Assert.ThrowsAsync<Exception>(() => connection.StartAsync());
-                        Assert.Same(expected, actual);
+                        Assert.True(shouldFail);
+                        await connection.StartAsync();
+                        Assert.False(shouldFail);
```
anurse: Stick a call counter in `OnTransportStart` and verify that too. Otherwise looks good

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -164,13 +166,27 @@ private Task StartAsyncCore()
...
                     connectUrl = CreateConnectUrl(Url, negotiationResponse);
                 }
 
                 _logger.StartingTransport(_transport, connectUrl);
-                await StartTransport(connectUrl);
+                foreach (var transport in _allTransports)
+                {
+                    try
+                    {
+                        if ((transport & _serverTransports) != 0)
```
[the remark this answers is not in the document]
mikaelm12: This happens internally in he DefaultTransportFactory https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/DefaultTransportFactory.cs#L39
Only thing I can think of is if a user passes a custom implementation of ITransportFactory then it might not and we should consider doing the mask ourselves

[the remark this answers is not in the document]
mikaelm12: We should actually do the mask here as well. We'll avoid unnecessary calls into `StartTransport` that way

mikaelm12:
Updated. Also do we still need to discuss any more design on this feature? @anurse @davidfowl

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -165,13 +170,41 @@ private Task StartAsyncCore()
...
-                _logger.StartingTransport(_transport, connectUrl);
-                await StartTransport(connectUrl);
+                foreach (var transport in AllTransports)
+                {
+                    var createNewConnectionId = false;
+                    try
+                    {
+                        if ((transport & _serverTransports & _requestedTransportType) != 0)
+                        {
+                            if (createNewConnectionId)
```
anurse: Two things: the `createNewConnectionId` value is defined inside the loop, so this won't really work. Also, you can simplify this by just doing the "re-negotiation" at the end of the loop, in the catch block.

mikaelm12: Smh

> Also, you can simplify this by just doing the "re-negotiation" at the end of the loop, in the catch block.

I wanted to add the logic with flag so that we can avoid re-negotiating after the last transport has failed

anurse: Hrm, yeah... I'm not sure I like either way of dealing with that... At the very least, fix the variable for now ;)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -165,13 +175,43 @@ private Task StartAsyncCore()
...
-                await StartTransport(connectUrl);
+                                await StartTransport(connectUrl, transport);
+                                _logger.StartingTransport(_transport, connectUrl);
+                                break;
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            // Try the next transport
+                            _logger.TransportFailedToStart(nameof(transport), ex);
```
davidfowl: Why are we using nameof(transport), won't that just end up being "transport"? You want the transport.ToString() right?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -345,6 +387,7 @@ private static Uri CreateConnectUrl(Uri url, NegotiationResponse negotiationResp
             catch (Exception ex)
             {
                 _logger.ErrorStartingTransport(_transport, ex);
```
davidfowl: Seems like we log transport failures already, do we need to log them twice?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -136,14 +138,36 @@ Task OnTransportStart()
...
+                        await connection.StartAsync();
+                        Assert.Equal(passThreshold, startCounter);
+                    });
+                }
+            }
 
-                        // Should succeed this time
-                        shouldFail = false;
+            [Fact]
+            public async Task StartThrowsAfterAllTransportsFail()
```
davidfowl: Can you add a test to verify the server behavior as well? Verify that a new connection id is created for each attempt and that all of the events run that you expect to be run.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -141,6 +144,14 @@ private Task StartAsyncCore()
             return _startTcs.Task;
         }
 
+        private async Task<NegotiationResponse> GetNewNegotiationResponse()
```
anurse: ~~New~~

It doesn't know it's "new" ;). The method just gets a negotiation response

mikaelm12:
@anurse Should that go in this PR or can that be put in a new one? This could conceptually be just for WebSockets failing to start initially. Then I can add the changes to LongPolling and SSE.

anurse:
> Should that go in this PR or can that be put in a new one

I'm OK with that as long as we file a bug (I'll do that) and it is the priority after this is merged. We don't want to think we've gotten transport fallback working and then be surprised :).

anurse:
Filed https://github.com/aspnet/SignalR/issues/1554

Making one last review pass now

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -166,13 +176,43 @@ private Task StartAsyncCore()
...
+                            // Try the next transport
+                            // Clear the negotiation response so we know to re-negotiate.
+                            negotiationResponse = null;
+                        }
+                    }
+                }
+                
+                if (_transport == null)
+                {
+                    throw new InvalidOperationException("No transport was created.");
```
anurse: Let's finesse this a bit: "Unable to connect to the server with any of the available transports".

Issue #928, opened 2017-09-22
Flaky Test: ServerSentEventsTransportTests.SSETransportStopsSendAndReceiveLoopsWhenTransportStopped


jkotalik:
Stack trace:

```
Failed   Microsoft.AspNetCore.SignalR.Client.Tests.ServerSentEventsTransportTests.SSETransportStopsSendAndReceiveLoopsWhenTransportStopped
Error Message:
 System.TimeoutException : Operation in SSETransportStopsSendAndReceiveLoopsWhenTransportStopped timed out at /Users/travis/build/aspnet/SignalR/test/Microsoft.AspNetCore.SignalR.Client.Tests/ServerSentEventsTransportTests.cs:117
Stack Trace:
   at Microsoft.AspNetCore.SignalR.Tests.Common.TaskExtensions.<OrTimeout>d__2.MoveNext() in /Users/travis/build/aspnet/SignalR/test/Common/TaskExtensions.cs:line 29
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.SignalR.Client.Tests.ServerSentEventsTransportTests.<SSETransportStopsSendAndReceiveLoopsWhenTransportStopped>d__1.MoveNext() in /Users/travis/build/aspnet/SignalR/test/Microsoft.AspNetCore.SignalR.Client.Tests/ServerSentEventsTransportTests.cs:line 117
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

moozzyk:
[...]

PR #1460, opened 2018-02-16
Remove the Channel<HubMessage> from the HubConnectionContext

davidfowl:
- Replace the channel with a single lock around the pipewriter. Since writes are always synchronous, the lock is held for a very short time.
- We were only using them in this scenario for handling multiple producers (the hub output, the keep alive ping and the broadcast).
- Handle the scenario where there's back pressure (when we use pipes that are bounded) and give callers a single task representing when back pressure is released.


davidfowl:
@BrennanConroy found a bug in the RedisHubLifetimeManager that was making my test fail. We don't handle synchronous exceptions properly. It only worked before because the method was async.

BrennanConroy:
I thought about adding exceptions there but thought it was fine, didn't think about it running synchronously! Nice catch!

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -41,6 +41,12 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
             _serverFixture = serverFixture;
         }
 
+        public IDisposable StartLog2(out ILoggerFactory loggerFactory, LogLevel logLevel = LogLevel.Trace, string testName = null)
```
anurse: `StartLog2` WAT? Let's get a better name there :)

test/Microsoft.AspNetCore.SignalR.Client.Tests/ServerSentEventsTransportTests.cs
```diff
@@ -65,7 +65,7 @@ public class ServerSentEventsTransportTests
             }
         }
 
-        [Fact]
+        [Fact(Skip = "Flaky tests keep failing")]
```
anurse: #928 ? Add a link to the skip message if so

anurse: Also, re-open it, because it's supposedly closed.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Startup.cs
```diff
@@ -44,8 +46,17 @@ public void ConfigureServices(IServiceCollection services)
             });
         }
 
-        public void Configure(IApplicationBuilder app)
+        public void Configure(IApplicationBuilder app, TestLoggerFactoryManager testLoggerFactory)
         {
+            app.Use(async (context, next) =>
```
anurse: Could we put this in a startup filter and put it in `ServerFixture`? So everybody gets it for free?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Startup.cs
```diff
@@ -44,8 +46,17 @@ public void ConfigureServices(IServiceCollection services)
...
+            {
+                var testName = context.Request.Query["testName"];
+
+                testLoggerFactory.SetCurrent(testName);
+
+                await next();
```
anurse: Doesn't need to be async, does it? Tail-await ;)

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -86,33 +88,56 @@ public class HubConnectionContext
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private Task FlushAsync(PipeWriter pipeWriter, CancellationToken cancellationToken)
+        {
+            var awaitable = pipeWriter.FlushAsync(cancellationToken);
+            if (awaitable.IsCompleted)
             {
-                throw new OperationCanceledException("Outbound channel was closed while trying to write hub message");
+                // The flush task can't fail today
```
anurse: micro-nit: Don't put temporal references in comments, it'll just get out of date ;P.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -86,33 +88,56 @@ public class HubConnectionContext
...
+        private async Task FlushAsyncAwaited(ValueAwaiter<FlushResult> awaitable, CancellationToken cancellationToken)
         {
-            // Nothing should be writing to the HubConnectionContext
-            _output.Writer.TryComplete();
+            // Since the flush awaitable doesn't currently support multiple awaiters
+            // we need to use a task to track the callbacks.
+            // All awaiters get the same task
+            lock (_flushLock)
+            {
+                if (_flushTcs == null || _flushTcs.Task.IsCompleted)
```
anurse: If the TCS isn't null or complete, we just tack another wait on it, but we have a different awaitable than the one that the `_flushTcs` is tracking. If I call our `FlushAsync` twice simultaneously doesn't that mean I call the Pipe `FlushAsync` twice but both tasks only end up waiting on the first one?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
davidfowl: I'm not checking this in

test/Microsoft.AspNetCore.SignalR.Client.Tests/ServerSentEventsTransportTests.cs
davidfowl: I'm not checking this in

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Startup.cs
davidfowl: I'm not checking this in, I'm just trying to find the bugs

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
davidfowl: Yes, thats by design. The there's only a single thing to wait on and that's backpressure was released. You don't get a task for your individual write.

anurse: Ok, is that what L123 is trying to say? It wasn't entirely clear to me at first reading.

davidfowl: > but we have a different awaitable than the one that the _flushTcs is tracking.

This might be where there's a misunderstanding? We're ignoring the returned awaitable while the flush task is incomplete. So it looks like,

FlushAsync() -> complete
FlushAsync() -> complete
FlushAsync() -> back pressure (waiting on flush tcs ignoring the returned awaitable)
FlushAsync() -> back pressure (waiting on flush tcs ignoring the returned awaitable)
FlushAsync() -> back pressure (waiting on flush tcs ignoring the returned awaitable)

Since the pipe is single producer single consumer, the awaitable returned represents the state of the PipeWriter itself, not the individual write operation. You can't have multiple callers waiting on the same FlushAsync call, as such, it isn't even possible to FlushAsync twice because the second one will throw.

PR #1466, opened 2018-02-17
Missed a couple workaround places

BrennanConroy:


BrennanConroy:
Nope, I'm going to revert everything and keep it skipped

PR #1467, opened 2018-02-17
Pass the array segment offset to WriteAsync

davidfowl:
Found it!


anurse:
Wait, how was this even working at all before? :) Post-merge :shipit:

davidfowl:
Must have been 0 most of the time

PR #1468, opened 2018-02-17
Remove the Channel<HubMessage> from the HubConnectionContext

davidfowl:
- Replace the channel with a single lock around the pipewriter. Since writes are always synchronous, the lock is held for a very short time.
- We were only using them in this scenario for handling multiple producers (the hub output, the keep alive ping and the broadcast).
- Handle the scenario where there's back pressure (when we use pipes that are bounded) and give callers a single task representing when back pressure is released.


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -86,33 +83,21 @@ public class HubConnectionContext
...
-            // Nothing should be writing to the HubConnectionContext
-            _output.Writer.TryComplete();
+            Interlocked.Exchange(ref _lastSendTimestamp, Stopwatch.GetTimestamp());
 
-            // This should unwind once we complete the output
-            await _writingTask;
-        }
+            await _connectionContext.Transport.Output.FlushAsync(CancellationToken.None);
 
+            _writeLock.Release();
```
BrennanConroy: Shouldn't this be in a `finally`

PR #1470, opened 2018-02-19
Tackling some low hanging performance fruit

davidfowl:
- Use native Memory/Span APIs on Stream in .NET Core 2.0
- Remove double copying in formatters
- Implemented custom HttpContent over ReadOnlyBuffer<byte>


src/Common/WebSocketExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    internal static class WebSocketExtensions
+    {
+        public static async Task SendAsync(this WebSocket webSocket, ReadOnlyBuffer<byte> buffer, WebSocketMessageType webSocketMessageType, CancellationToken cancellationToken = default)
+        {
+            // TODO: Consider chunking writes here if we get a multi segment buffer
+#if NETCOREAPP2_1
+            if (buffer.IsSingleSegment)
+            {
+                await webSocket.SendAsync(buffer.First, webSocketMessageType, endOfMessage: true, CancellationToken.None);
```
BrennanConroy: Pass in `cancellationToken`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
```diff
@@ -2,21 +2,23 @@
...
+        public static void WriteLengthPrefix(long length, Stream output)
         {
             // This code writes length prefix of the message as a VarInt. Read the comment in
             // the BinaryMessageParser.TryParseMessage for details.
 
-            var lenBuffer = stackalloc byte[5];
+#if NETCOREAPP2_1
+            Span<byte> lenBuffer = stackalloc byte[5];
+#else
+            var lenBuffer = new byte[5];
```
BrennanConroy: Why isn't this stackalloc?

src/Common/StreamExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Buffers;
```
BrennanConroy: Missed a few copyrights

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
benaadams: Unsafe?

src/Common/StreamExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+namespace System.IO
+{
+    internal static class StreamExtensions
+    {
+        public static async Task WriteAsync(this Stream stream, ReadOnlyBuffer<byte> buffer)
+        {
+            // REVIEW: Should we special case IsSingleSegment here?
```
anurse: To remove the state machine? Let's wait until we see it show up in perf data.

src/Common/WebSocketExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Net.WebSockets
+{
+    internal static class WebSocketExtensions
+    {
+        public static async Task SendAsync(this WebSocket webSocket, ReadOnlyBuffer<byte> buffer, WebSocketMessageType webSocketMessageType, CancellationToken cancellationToken = default)
```
anurse: Looks like these are all tail-`awaits`, no need for `async`?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageFormatter.cs
anurse: Also, does `Stream` have a `WriteAsync` that takes `byte*`? We used to copy to a pooled array before, which seems unnecessary.

PR #1471, opened 2018-02-20
Remove timeout from receive tcs.

davidfowl:
- This test echos a large message and it fails sometimes before the entire thing is delivered. Just drop the timeout.
- Avoid creating timers for already completed tasks


test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -290,7 +290,7 @@ public static IEnumerable<object[]> MessageSizesData
                     logger.LogInformation("Sent message", bytes.Length);
 
                     logger.LogInformation("Receiving message");
-                    var receivedData = await receiveTcs.Task.OrTimeout();
+                    var receivedData = await receiveTcs.Task;
```
BrennanConroy: Add a comment so we don't accidentally add it back

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -290,7 +290,8 @@ public static IEnumerable<object[]> MessageSizesData
                     logger.LogInformation("Sent message", bytes.Length);
 
                     logger.LogInformation("Receiving message");
-                    var receivedData = await receiveTcs.Task.OrTimeout();
+                    // No timeout here because it can take a while to receive all the bytes
```
anurse: What's a while? Is this a useful test if it takes more than 1 minute? 5 minutes? 10 minutes?

Issue #1207, opened 2017-12-12
Allow Pipes used in SignalR to be bounded


anurse:
Server-side Unbounded Channels are a Bad Idea(TM). We should make sure to expose the ability to configure Bounded Channels for all of our channels.

davidfowl:
This is an implementation detail. We need to figure out the implications of what we're doing right now and see if we need to expose anything or just pick a default (though, when has that ever worked without making it configurable).

anurse:
Agreed, just want this tracked. Exposing it seems pretty much necessary for properly adapting to client load.

anurse:
This should be done in Preview 2. I'm nervous it'll get missed. We'll need to look at how many Channels go away with the pipelines change, but if any are left they should be bounded.

davidfowl:
So there are several things here:
- So we need to expose the thresholds.
- `PipeOptions` needs a size for  PauseWriterThreshold and ResumeWriterThresold  here https://github.com/aspnet/SignalR/blob/1bd37cabf0675dd24689e03c8724fd8277a8a9c3/src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs#L59
- We need to pass the options in so we don't hardcode them.

mikaelm12:
Done: 0e4731e480ba2d44963d43dfb99451ee468f5b77

PR #1473, opened 2018-02-20
Expose PauseWriter and ResumeWriter Thresholds

mikaelm12:
Exposing pipe thresholds in HttpSocketOptions
Issue : https://github.com/aspnet/SignalR/issues/1207

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -365,8 +365,9 @@ namespace Microsoft.AspNetCore.Sockets
             context.Response.ContentType = "application/json";
 
             // Establish the connection
-            var pipeOptions = new PipeOptions(pauseWriterThreshold: options.PauseWriterThreshold, resumeWriterThreshold: options.ResumeWriterThreshold);
-            var connection = _manager.CreateConnection(pipeOptions);
+            var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportPauseWriterThreshold, resumeWriterThreshold: options.TransportPauseWriterThreshold/2);
+            var appPipeOptions = new PipeOptions(pauseWriterThreshold: options.ApplicationPauseWriterThreshold, resumeWriterThreshold: options.ApplicationPauseWriterThreshold/2);
+            var connection = _manager.CreateConnection(transportPipeOptions, appPipeOptions);
 
             // Set the Connection ID on the logging scope so that logs from now on will have the
             // Connection ID metadata set.
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -126,7 +126,7 @@ namespace Microsoft.AspNetCore.Sockets
             else if (context.WebSockets.IsWebSocketRequest)
             {
                 // Connection can be established lazily
-                var connection = await GetOrCreateConnectionAsync(context);
+                var connection = await GetOrCreateConnectionAsync(context, options);
                 if (connection == null)
                 {
                     // No such connection, GetOrCreateConnection already set the response status code
@@ -365,8 +365,8 @@ namespace Microsoft.AspNetCore.Sockets
             context.Response.ContentType = "application/json";
 
             // Establish the connection
-            var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportPauseWriterThreshold, resumeWriterThreshold: options.TransportPauseWriterThreshold/2);
-            var appPipeOptions = new PipeOptions(pauseWriterThreshold: options.ApplicationPauseWriterThreshold, resumeWriterThreshold: options.ApplicationPauseWriterThreshold/2);
+            var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportMaxBufferSize, resumeWriterThreshold: options.TransportMaxBufferSize / 2);
+            var appPipeOptions = new PipeOptions(pauseWriterThreshold: options.ApplicationMaxBufferSize, resumeWriterThreshold: options.ApplicationMaxBufferSize/ 2);
             var connection = _manager.CreateConnection(transportPipeOptions, appPipeOptions);
 
             // Set the Connection ID on the logging scope so that logs from now on will have the
@@ -518,7 +518,7 @@ namespace Microsoft.AspNetCore.Sockets
             return connection;
         }
 
-        private async Task<DefaultConnectionContext> GetOrCreateConnectionAsync(HttpContext context)
+        private async Task<DefaultConnectionContext> GetOrCreateConnectionAsync(HttpContext context, HttpSocketOptions options)
         {
             var connectionId = GetConnectionId(context);
             DefaultConnectionContext connection;
@@ -526,7 +526,9 @@ namespace Microsoft.AspNetCore.Sockets
             // There's no connection id so this is a brand new connection
             if (StringValues.IsNullOrEmpty(connectionId))
             {
-                connection = _manager.CreateConnection();
+                var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportMaxBufferSize, resumeWriterThreshold: options.TransportMaxBufferSize / 2);
+                var appPipeOptions = new PipeOptions(pauseWriterThreshold: options.ApplicationMaxBufferSize, resumeWriterThreshold: options.ApplicationMaxBufferSize / 2);
+                connection = _manager.CreateConnection(transportPipeOptions, appPipeOptions);
             }
             else if (!_manager.TryGetConnection(connectionId, out connection))
             {
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -365,7 +365,7 @@ namespace Microsoft.AspNetCore.Sockets
             context.Response.ContentType = "application/json";
 
             // Establish the connection
-            var connection = _manager.CreateConnection();
+            var connection = _manager.CreateConnection(options.PauseWriterThreshold, options.ResumeWriterThreshold);
 
             // Set the Connection ID on the logging scope so that logs from now on will have the
             // Connection ID metadata set.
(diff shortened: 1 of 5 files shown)


anurse:
Trying to think how we can test this. The simplest way would be to test that the data flows to the right place. We could also consider building a test that checks that FlushAsync will block appropriately, but that seems like a flake-farm.

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -56,22 +56,27 @@ public bool TryGetConnection(string id, out DefaultConnectionContext connection)
             return _connections.TryGetValue(id, out connection);
         }
 
-        public DefaultConnectionContext CreateConnection()
+        public DefaultConnectionContext CreateConnection(long pauseWriterThreshold, long resumeWriterThreshold)
```
davidfowl: This should take a PipeOptions in directly, not the explicit thresholds

davidfowl:
We should have a test for it to make sure that it's applied and that we get the appropriate backpressure.

mikaelm12:
Hmm, what should that look like?

anurse:
I'm not sure we have the pipelines exposed enough for a test to simulate that, but maybe @davidfowl has some idea?

src/Microsoft.AspNetCore.Sockets/ConnectionManager.cs
```diff
@@ -56,22 +56,26 @@ public bool TryGetConnection(string id, out DefaultConnectionContext connection)
...
-            var pair = DuplexPipe.CreateConnectionPair(PipeOptions.Default, PipeOptions.Default);
-            
             var connection = new DefaultConnectionContext(id, pair.Application, pair.Transport);
             connection.ConnectionTimer = connectionTimer;
 
             _connections.TryAdd(id, connection);
             return connection;
         }
 
+        public DefaultConnectionContext CreateConnection()
```
anurse: Do we need this overload? It's an internal type, so I think we don't need it.

davidfowl: I believe the tests use it.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -365,7 +365,8 @@ private Task ProcessNegotiate(HttpContext context, HttpSocketOptions options, Co
             context.Response.ContentType = "application/json";
 
             // Establish the connection
-            var connection = _manager.CreateConnection();
+            var pipeOptions = new PipeOptions(pauseWriterThreshold: options.PauseWriterThreshold, resumeWriterThreshold: options.ResumeWriterThreshold);
+            var connection = _manager.CreateConnection(pipeOptions);
```
davidfowl: What about the other code path where we connect via a websocket directly without negotiating? This is why we need tests 😄

davidfowl:
So testing this will be a bit tricky but doable. We want to make sure that when the limits are applied and are effective in all cases. We can discuss it tomorrow

src/Microsoft.AspNetCore.Sockets.Http/HttpSocketOptions.cs
```diff
@@ -18,8 +18,8 @@ public class HttpSocketOptions
 
         public LongPollingOptions LongPolling { get; } = new LongPollingOptions();
 
-        public long PauseWriterThreshold { get; set; } = PipeOptions.Default.PauseWriterThreshold;
+        public long TransportPauseWriterThreshold { get; set; } = PipeOptions.Default.PauseWriterThreshold;
 
-        public long ResumeWriterThreshold { get; set; } = PipeOptions.Default.ResumeWriterThreshold;
+        public long ApplicationPauseWriterThreshold{ get; set; } = PipeOptions.Default.ResumeWriterThreshold;
```
anurse: Initialize these both to `0` (and initializing this to the default `ResumeWriterThreshold` would be wrong anyway ;)). We'll explore defaults ourselves later.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/PipeReaderExtensions.cs
```diff
@@ -51,6 +51,19 @@ public static class PipeReaderExtensions
             }
         }
 
+
+        public static async Task ReadAsync(this PipeReader pipeReader, int numBytes)
```
anurse: Rename to `SkipAsync` since it's not returning the values it reads

test/Microsoft.AspNetCore.SignalR.Tests.Utils/PipeReaderExtensions.cs
```diff
@@ -51,6 +51,19 @@ public static class PipeReaderExtensions
...
+        {
+            var result = await pipeReader.ReadAsync();
+            if (result.Buffer.Length < numBytes)
+            {
+                pipeReader.AdvanceTo(result.Buffer.Start, result.Buffer.End);
```
anurse: This should loop until you're actually able to read the requested number of bytes.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -57,6 +57,40 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
             }
         }
 
+        [Fact]
+        public async Task CheckThatThresholdValuesAreEnforced()
+        {
+            using (StartLog(out var loggerFactory, LogLevel.Debug))
+            {
+                while (!System.Diagnostics.Debugger.IsAttached) { }
```
anurse: Leftover I presume ;)

test/Microsoft.AspNetCore.SignalR.Tests.Utils/PipeReaderExtensions.cs
davidfowl: ConsumeAsync?

src/Microsoft.AspNetCore.Sockets.Http/HttpSocketOptions.cs
davidfowl: These names aren't great, can we call them TransportMaxBufferSize and ApplicationMaxBufferSize, and do what @anurse said and make them 0.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -365,7 +365,9 @@ private Task ProcessNegotiate(HttpContext context, HttpSocketOptions options, Co
             context.Response.ContentType = "application/json";
 
             // Establish the connection
-            var connection = _manager.CreateConnection();
+            var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportPauseWriterThreshold, resumeWriterThreshold: options.TransportPauseWriterThreshold/2);
```
davidfowl: 2 things.
- Format the document
- You're still not setting the pipe options for direct websockets...

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
mikaelm12: ...

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -516,15 +518,17 @@ private static string GetNegotiatePayload(string connectionId, HttpSocketOptions
...
+        private async Task<DefaultConnectionContext> GetOrCreateConnectionAsync(HttpContext context, HttpSocketOptions options)
         {
             var connectionId = GetConnectionId(context);
             DefaultConnectionContext connection;
 
             // There's no connection id so this is a brand new connection
             if (StringValues.IsNullOrEmpty(connectionId))
             {
-                connection = _manager.CreateConnection();
+                var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportMaxBufferSize, resumeWriterThreshold: options.TransportMaxBufferSize / 2);
```
davidfowl: Move this duplicated logic into a method.

src/Microsoft.AspNetCore.Sockets.Http/HttpSocketOptions.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Authorization;
+using System.IO.Pipelines;
```
davidfowl: Sort usings

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -57,6 +57,39 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
...
+
+                // This write should complete immediately but it exceeds the writer threshold
+                var writeTask = connection.Application.Output.WriteAsync(new byte[] { (byte)'b', (byte)'y', (byte)'t', (byte)'e', (byte)'s' });
+
+                Assert.False(writeTask.IsCompleted);
+
+                // Reading here puts us below the threshold 
+                await connection.Transport.Input.ConsumeAsync(5);
+
+                await writeTask;
```
davidfowl: OrTimeout

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -516,15 +516,22 @@ private static string GetNegotiatePayload(string connectionId, HttpSocketOptions
             return connection;
         }
 
-        private async Task<DefaultConnectionContext> GetOrCreateConnectionAsync(HttpContext context)
+        private DefaultConnectionContext CreateConnectionInternal(HttpSocketOptions options)
+        {
+            var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportMaxBufferSize, resumeWriterThreshold: options.TransportMaxBufferSize / 2);
+            var appPipeOptions = new PipeOptions(pauseWriterThreshold: options.ApplicationMaxBufferSize, resumeWriterThreshold: options.ApplicationMaxBufferSize / 2);
+             return _manager.CreateConnection(transportPipeOptions, appPipeOptions);
```
davidfowl: nit: formatting is off here. Do you have the same settings as @anurse ? :trollface:

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -57,6 +57,39 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
             }
         }
 
+        [Fact]
+        public async Task CheckThatThresholdValuesAreEnforced()
```
davidfowl: There should also be a test that sends data via /send to make sure it doesn't return until back pressure is released.

mikaelm12:
Updated

PR #1479, opened 2018-02-22
Fix #924 by introducing HubDispatcher abstraction

### LINKED ISSUE #924: Add support to HubLifetimeManager for handling incoming invocations before the hub method is called
This is basically slowly bringing back the hub pipeline. We'll eventually end up with a decorator pattern again for wrapping all hub incoming and outgoing invocations

anurse:
We originally talked about using HubLifetimeManager for this, but I have concerns about that. It makes HubLifetimeManager into a God Object that has lots of different unrelated behaviors. For example, the RedisHubLifetimeManager would have to provided implementations of the dispatching-related methods that have the same implementation as the DefaultHubLifetimeManager. Even if we provide a common type or base-class to do so, it feels wrong. I have some thoughts on how to factor things a little better among these types but they're a little too big to do in one go and need some discussion first. I tried a few attempts today but it got tricky fast.

So this is the keep-it-simple first pass at doing this refactoring. There is a new service, `HubDispatcher<THub>` which contains all the logic needed to dispatch `OnConnectedAsync`, `OnDisconnectedAsync` and `HubMessage`s to the Hub. Almost all the code is just literally extracted from HubEndpoint.

Interestingly, no test changes were needed. I like this because it tells me that the tests weren't coupled to the specific architecture of the system 😄, also because it tells me I didn't break things. The HubEndpointTests actually use `.AddSignalR` to build a DI container, so everything Just Worked(TM).

Also, while I was there, I did the `Log` nested class refactoring (since I was moving log messages around already)


samples/JwtSample/Properties/launchSettings.json
```diff
@@ -3,7 +3,7 @@
     "windowsAuthentication": false,
     "anonymousAuthentication": true,
     "iisExpress": {
-      "applicationUrl": "http://localhost:57780/",
+      "applicationUrl": "http://localhost:8719/",
```
anurse: I don't know what I did to make VS do this...

image

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -78,85 +61,38 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
+                await connection.AbortAsync();
             }
             catch (Exception ex)
             {
-                _logger.ErrorInvokingHubMethod("OnDisconnectedAsync", ex);
-                throw;
+                Log.AbortFailed(_logger, ex);
             }
+
+            await _dispatcher.OnDisconnectedAsync(connection, exception);
```
davidfowl: Why did you move this out of the try catch block?

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubDispatcher.cs
```diff
@@ -0,0 +1,461 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    public partial class HubDispatcher<THub> : IInvocationBinder where THub : Hub
```
davidfowl: Is this implementing IInvocationBinder an implementation requirement?

anurse:
> I think this needs to be an interface or a true abstract base class

Sure, I had that originally but it seemed a little overkill. I guess it makes sense though, you usually want to replace all or none of the behaviors.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubDispatcher.cs
anurse: It is right now. The Invocation Binder and the Dispatcher both need to share the metadata about the methods in the Hub. We could do that in two ways: Extract "Discovery" into a separate component and have both the Binder and Dispatcher use it, or make the Binder and Dispatcher the same component and do discovery internally. Until we decide what to do with Discovery, this seemed like the most straightforward way to do it.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
anurse: Because... uh... shut up. That's why.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubDispatcher.Log.cs
```diff
@@ -0,0 +1,143 @@
...
+                LoggerMessage.Define<string, string>(LogLevel.Trace, new EventId(10, "HubMethodBound"), "'{hubName}' hub method '{hubMethod}' is bound.");
+
+            private static readonly Action<ILogger, string, Exception> _cancelStream =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(11, "CancelStream"), "Canceling stream for invocation {invocationId}.");
+
+            private static readonly Action<ILogger, Exception> _unexpectedCancel =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(12, "UnexpectedCancel"), "CancelInvocationMessage received unexpectedly.");
+
+            private static readonly Action<ILogger, StreamInvocationMessage, Exception> _receivedStreamHubInvocation =
+                LoggerMessage.Define<StreamInvocationMessage>(LogLevel.Debug, new EventId(14, "ReceivedStreamHubInvocation"), "Received stream hub invocation: {invocationMessage}.");
```
BrennanConroy: Skipped 13 for some reason

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -41,11 +39,31 @@ public IHubProtocol GetProtocol(string protocolName, HubConnectionContext connec
...
             }
 
             throw new NotSupportedException($"The protocol '{protocolName ?? "(null)"}' is not supported.");
         }
+
+        private static class Log
+        {
+            // Category: DefaultHubProtocolResolver
+            private static readonly Action<ILogger, string, Type, Exception> _registeredSignalRProtocol =
+                LoggerMessage.Define<string, Type>(LogLevel.Debug, new EventId(1, nameof(RegisteredSignalRProtocol)), "Registered SignalR Protocol: {ProtocolName}, implemented by {ImplementationType}.");
```
BrennanConroy: remove nameof

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -244,345 +141,23 @@ public class HubEndPoint<THub> : IInvocationBinder where THub : Hub
...
-
-        private void InitializeHub(THub hub, HubConnectionContext connection)
+        private static class Log
         {
-            hub.Clients = new HubCallerClients(_hubContext.Clients, connection.ConnectionId);
-            hub.Context = new HubCallerContext(connection);
-            hub.Groups = _hubContext.Groups;
-        }
+            private static readonly Action<ILogger, Exception> _errorProcessingRequest =
+                LoggerMessage.Define(LogLevel.Error, new EventId(1, nameof(ErrorProcessingRequest)), "Error when processing requests.");
```
BrennanConroy: remove nameof

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubDispatcher.cs
```diff
@@ -0,0 +1,461 @@
+using System;
```
BrennanConroy: Copyright

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubDispatcher.Log.cs
anurse: Removed it, forgot to renumber :).

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
anurse: Actually, the try catch for that code is now in the dispatcher. It was always a log-and-rethrow catch, so it's fine there.

anurse:
:up: :date:

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -0,0 +1,464 @@
...
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    public partial class DefaultHubDispatcher<THub> : HubDispatcher<THub> where THub : Hub
+    {
+        private readonly Dictionary<string, HubMethodDescriptor> _methods = new Dictionary<string, HubMethodDescriptor>(StringComparer.OrdinalIgnoreCase);
+        private readonly IServiceScopeFactory _serviceScopeFactory;
+        private readonly IHubContext<THub> _hubContext;
+        private readonly ILogger<HubDispatcher<THub>> _logger;
+
+        public DefaultHubDispatcher(IServiceScopeFactory serviceScopeFactory, IHubContext<THub> hubContext, ILogger<HubDispatcher<THub>> logger)
```
BrennanConroy: `ILogger<DefaultHubDispatcher>`

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -236,5 +235,42 @@ private static void AbortConnection(object state)
                 connection._abortCompletedTcs.TrySetException(ex);
             }
         }
+
+        private static class Log
+        {
+            // Category: HubConnectionContext
+            private static readonly Action<ILogger, string, Exception> _usingHubProtocol =
+                LoggerMessage.Define<string>(LogLevel.Information, new EventId(1, nameof(UsingHubProtocol)), "Using HubProtocol '{protocol}'.");
```
BrennanConroy: nmaefo

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
davidfowl: We should try catch here now that we can't trust what it does

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -2,57 +2,40 @@
...
         private readonly ILogger<HubEndPoint<THub>> _logger;
-        private readonly IServiceScopeFactory _serviceScopeFactory;
         private readonly IHubProtocolResolver _protocolResolver;
         private readonly HubOptions _hubOptions;
         private readonly IUserIdProvider _userIdProvider;
+        private readonly HubDispatcher<THub> _dispatcher;
 
         public HubEndPoint(HubLifetimeManager<THub> lifetimeManager,
                            IHubProtocolResolver protocolResolver,
                            IHubContext<THub> hubContext,
```
xscript: minor: `hubContext` is not required any more.

anurse:
Alright, final :up: :date:. I'll be merging when builds are green.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubDispatcher.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    public abstract class HubDispatcher<THub> : IInvocationBinder where THub : Hub
+    {
+        public abstract Task OnConnectedAsync(HubConnectionContext connection);
+        public abstract Task OnDisconnectedAsync(HubConnectionContext connection, Exception exception);
+        public abstract Task DispatchMessageAsync(HubConnectionContext connection, HubMessage hubMessage);
+        public abstract IReadOnlyList<Type> GetParameterTypes(string methodName);
```
davidfowl: We should discuss separating this. It really shouldn’t be coupled to the dispatcher

anurse: I don't disagree, and I have some ideas. Though it means making Discovery more of A Thing(TM).

Issue #1449, opened 2018-02-15
Don't log message content in the JavaScript client


anurse:
We generally avoid logging things that may have sensitive data in them, we should avoid logging the message content in the JavaScript client. I'd certainly expect users to ship logs from the client over to a telemetry service, so they should be treated just as sensitively as server-side logs.

PR #1491, opened 2018-02-25
Don't log message data

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1449
We don't want to log message content on the client.

### DIFF
--- a/client-ts/signalr/src/HubConnection.ts
+++ b/client-ts/signalr/src/HubConnection.ts
@@ -85,7 +85,7 @@ export class HubConnection {
                     // Don't care about pings
                     break;
                 default:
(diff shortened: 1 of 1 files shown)


BrennanConroy:
Pretty sure there are more locations that we need to change.

davidfowl:
Also tests?

davidfowl:
It comes down to understanding how important this change is. If another change regresses this I have extremely low confidence that we'll catch it in a code review.

davidfowl:
@anurse and @mikaelm12 can you talk to @anpete and @ajcvickers about how EF handles logging of parameter values?

mikaelm12:
Not working on the atm.

Issue #854, opened 2017-09-12
Add ability to restrict protocol on the server side


davidfowl:
Today we have defaults that let the hub protocol be negotiated over the connection. We should support a direct connection mode as well where the transport is specific as part of a specific hub configuration.

```C#
app.UseSignalR(routes =>
{
    routes.MapHub<Cache>("cache", o =>
    {
        o.Transports = TransportType.WebSockets;

        // Remove all protocols
        o.SupportedProtocols.Clear();

        // Only support msg pack
        o.SupportedProtocols.Add(DefaultProtocols.MsgPackProtocol);
    });
});
```

mikaelm12:
There's actually a layering issue with the example above. The `SupportedProtocols` property shouldn't live in the `HttpSocketOptions` but in `HubOptions`.
Setting would look like 

```C#
            services.AddSignalR(options =>
            {
                // Remove all protocols
                options.SupportedProtocols.Clear();

                // Only support msg pack
                options.SupportedProtocols.Add(DefaultProtocols.MsgPackProtocol);
            });
```

mikaelm12:
Done: 1b9313287bad3c11d5532f6f4315485ad12f21c5

PR #1492, opened 2018-02-26
Restrict HubProtocol on Server + HubOptions<THub>

mikaelm12:
Adding the ability to restrict the HubProtocol on the server
Issue: https://github.com/aspnet/SignalR/issues/854

### DIFF
--- a/samples/SocketsSample/Startup.cs
+++ b/samples/SocketsSample/Startup.cs
@@ -20,8 +20,7 @@ namespace SocketsSample
         {
             services.AddSockets();
 
-            services.AddSignalR(options =>
-            {
+            services.AddSignalR(options => {
                 // Faster pings for testing
                 options.KeepAliveInterval = TimeSpan.FromSeconds(5);
             })
--- a/samples/SocketsSample/Startup.cs
+++ b/samples/SocketsSample/Startup.cs
@@ -20,7 +20,8 @@ namespace SocketsSample
         {
             services.AddSockets();
 
-            services.AddSignalR(options => {
+            services.AddSignalR(options =>
+            {
                 // Faster pings for testing
                 options.KeepAliveInterval = TimeSpan.FromSeconds(5);
             })
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -17,13 +17,13 @@ namespace Microsoft.AspNetCore.SignalR
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger<HubEndPoint<THub>> _logger;
         private readonly IHubProtocolResolver _protocolResolver;
-        private readonly HubOptions _hubOptions;
+        private readonly HubOptions<THub> _hubOptions;
         private readonly IUserIdProvider _userIdProvider;
         private readonly HubDispatcher<THub> _dispatcher;
 
         public HubEndPoint(HubLifetimeManager<THub> lifetimeManager,
                            IHubProtocolResolver protocolResolver,
-                           IOptions<HubOptions> hubOptions,
+                           IOptions<HubOptions<THub>> hubOptions,
                            ILoggerFactory loggerFactory,
                            IUserIdProvider userIdProvider,
                            HubDispatcher<THub> dispatcher)
@@ -39,9 +39,14 @@ namespace Microsoft.AspNetCore.SignalR
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
+            if (_hubOptions.SupportedProtocols != null && _hubOptions.SupportedProtocols.Count == 0)
+            {
+                throw new InvalidOperationException("There are no supported protocols");
+            }
+
             var connectionContext = new HubConnectionContext(connection, _hubOptions.KeepAliveInterval, _loggerFactory);
 
-            if (!await connectionContext.NegotiateAsync(_hubOptions.NegotiateTimeout, _protocolResolver, _userIdProvider))
+            if (!await connectionContext.NegotiateAsync(_hubOptions.NegotiateTimeout, _hubOptions.SupportedProtocols, _protocolResolver, _userIdProvider))
             {
                 return;
             }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -43,8 +43,9 @@ namespace Microsoft.AspNetCore.SignalR
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
             // We check to see if HubOptions<THub> are set because those take precedence over global hub options.
-            var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval;
-            var negotiateTimeout = _hubOptions.NegotiateTimeout ?? _globalHubOptions.NegotiateTimeout;
+            // Then set the keepAlive and negotiateTimeout values to the defaults in HubOptionsSetup incase they were explicitly set to null.
+            var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval ?? HubOptionsSetup.DefaultKeepAliveInterval;
+            var negotiateTimeout = _hubOptions.NegotiateTimeout ?? _globalHubOptions.NegotiateTimeout ?? HubOptionsSetup.DefaultNegotiateTimeout;
             var supportedProtocols = _hubOptions.SupportedProtocols;
             if (supportedProtocols == null && _globalHubOptions.SupportedProtocols != null)
             {
@@ -56,9 +57,9 @@ namespace Microsoft.AspNetCore.SignalR
                 throw new InvalidOperationException("There are no supported protocols");
             }
 
-            var connectionContext = new HubConnectionContext(connection, keepAlive.Value, _loggerFactory);
+            var connectionContext = new HubConnectionContext(connection, keepAlive, _loggerFactory);
 
-            if (!await connectionContext.NegotiateAsync(negotiateTimeout.Value, supportedProtocols, _protocolResolver, _userIdProvider))
+            if (!await connectionContext.NegotiateAsync(negotiateTimeout, supportedProtocols, _protocolResolver, _userIdProvider))
             {
                 return;
             }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -46,11 +46,7 @@ namespace Microsoft.AspNetCore.SignalR
             // Then set the keepAlive and negotiateTimeout values to the defaults in HubOptionsSetup incase they were explicitly set to null.
             var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval ?? HubOptionsSetup.DefaultKeepAliveInterval;
             var negotiateTimeout = _hubOptions.NegotiateTimeout ?? _globalHubOptions.NegotiateTimeout ?? HubOptionsSetup.DefaultNegotiateTimeout;
-            var supportedProtocols = _hubOptions.SupportedProtocols;
-            if (supportedProtocols == null && _globalHubOptions.SupportedProtocols != null)
-            {
-                supportedProtocols = _globalHubOptions.SupportedProtocols;
-            }
+            var supportedProtocols = _hubOptions.SupportedProtocols ?? _globalHubOptions.SupportedProtocols;
 
             if (supportedProtocols != null && supportedProtocols.Count == 0)
             {
@@ -74,7 +70,6 @@ namespace Microsoft.AspNetCore.SignalR
                 await _lifetimeManager.OnDisconnectedAsync(connectionContext);
             }
         }
-
 
         private async Task RunHubAsync(HubConnectionContext connection)
         {
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -42,6 +42,7 @@ namespace Microsoft.AspNetCore.SignalR
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
+            // We check to see if HubOptions<THub> are set because those take precedence over global hub options.
             var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval;
             var negotiateTimeout = _hubOptions.NegotiateTimeout ?? _globalHubOptions.NegotiateTimeout;
             var supportedProtocols = _hubOptions.SupportedProtocols;
@@ -55,9 +56,9 @@ namespace Microsoft.AspNetCore.SignalR
                 throw new InvalidOperationException("There are no supported protocols");
             }
 
-            var connectionContext = new HubConnectionContext(connection, keepAlive, _loggerFactory);
+            var connectionContext = new HubConnectionContext(connection, keepAlive.Value, _loggerFactory);
 
-            if (!await connectionContext.NegotiateAsync(negotiateTimeout, supportedProtocols, _protocolResolver, _userIdProvider))
+            if (!await connectionContext.NegotiateAsync(negotiateTimeout.Value, supportedProtocols, _protocolResolver, _userIdProvider))
             {
                 return;
             }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
@@ -18,11 +18,13 @@ namespace Microsoft.AspNetCore.SignalR
         private readonly ILogger<HubEndPoint<THub>> _logger;
         private readonly IHubProtocolResolver _protocolResolver;
         private readonly HubOptions<THub> _hubOptions;
+        private readonly HubOptions _globalHubOptions;
         private readonly IUserIdProvider _userIdProvider;
         private readonly HubDispatcher<THub> _dispatcher;
 
         public HubEndPoint(HubLifetimeManager<THub> lifetimeManager,
                            IHubProtocolResolver protocolResolver,
+                           IOptions<HubOptions> globalHubOptions,
                            IOptions<HubOptions<THub>> hubOptions,
                            ILoggerFactory loggerFactory,
                            IUserIdProvider userIdProvider,
@@ -32,6 +34,7 @@ namespace Microsoft.AspNetCore.SignalR
             _lifetimeManager = lifetimeManager;
             _loggerFactory = loggerFactory;
             _hubOptions = hubOptions.Value;
+            _globalHubOptions = globalHubOptions.Value;
             _logger = loggerFactory.CreateLogger<HubEndPoint<THub>>();
             _userIdProvider = userIdProvider;
             _dispatcher = dispatcher;
@@ -39,14 +42,22 @@ namespace Microsoft.AspNetCore.SignalR
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
-            if (_hubOptions.SupportedProtocols != null && _hubOptions.SupportedProtocols.Count == 0)
+            var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval;
+            var negotiateTimeout = _hubOptions.NegotiateTimeout ?? _globalHubOptions.NegotiateTimeout;
+            var supportedProtocols = _hubOptions.SupportedProtocols;
+            if (supportedProtocols == null && _globalHubOptions.SupportedProtocols != null)
+            {
+                supportedProtocols = _globalHubOptions.SupportedProtocols;
+            }
+
+            if (supportedProtocols != null && supportedProtocols.Count == 0)
             {
                 throw new InvalidOperationException("There are no supported protocols");
             }
 
-            var connectionContext = new HubConnectionContext(connection, _hubOptions.KeepAliveInterval, _loggerFactory);
+            var connectionContext = new HubConnectionContext(connection, keepAlive, _loggerFactory);
 
-            if (!await connectionContext.NegotiateAsync(_hubOptions.NegotiateTimeout, _hubOptions.SupportedProtocols, _protocolResolver, _userIdProvider))
+            if (!await connectionContext.NegotiateAsync(negotiateTimeout, supportedProtocols, _protocolResolver, _userIdProvider))
             {
                 return;
             }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
@@ -8,23 +8,9 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class HubOptions
     {
-        /// <summary>
-        /// The default keep-alive interval. This is set to exactly half of the default client timeout window,
-        /// to ensure a ping can arrive in time to satisfy the client timeout.
-        /// </summary>
-        public static readonly TimeSpan DefaultKeepAliveInterval = TimeSpan.FromSeconds(15);
+        public TimeSpan? NegotiateTimeout { get; set; } = null;
 
-        public TimeSpan NegotiateTimeout { get; set; } = TimeSpan.FromSeconds(5);
-
-        /// <summary>
-        /// The interval at which keep-alive messages should be sent. The default interval
-        /// is 15 seconds.
-        /// </summary>
-        /// <remarks>
-        /// This interval is not used by the Long Polling transport as it has inherent keep-alive
-        /// functionality because of the polling mechanism. 
-        /// </remarks>
-        public TimeSpan KeepAliveInterval { get; set; } = DefaultKeepAliveInterval;
+        public TimeSpan? KeepAliveInterval { get; set; } = null;
 
         public List<string> SupportedProtocols { get; set; } = null;
     }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.SignalR
 {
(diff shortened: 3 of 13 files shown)


davidfowl:
I think this is way too global (I was going to update the issue but this is fine). It needs to be per hub end point not per signalr instance. We might need HubOptions<THub> to support this properly.

src/Microsoft.AspNetCore.SignalR.Common/DefaultProtocols.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public static class DefaultProtocols
+    {
+        public static readonly string JSONProtocol = "json";
```
anurse: Use `Json`. We generally treat it like a "Word" in our code, even though it is technically an acronym ;)

mikaelm12: I made it all lowercase to match the name property on the JsonHubProtocol https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs#L31
I can change that too if we want.

davidfowl: The only default protocol we have is JSON really. MSGpack is an addon.

mikaelm12: Okay, I can remove that from here. Was following what the issue laid out 😄

davidfowl:
We'll remove the HubOptions, in the AddSignalR overload and instead make them `HubOptions<THub>`. We'll revisit global options later.

`IHubProtocol.GetProtocol` now takes 3 arguments:

```C#
GetProtocol(IReadOnlyList<string> supportedProtocols,
string requestedProtocols,
HubConnectionContext context); 
```

We also need to add an extension method AddHubOptions on ISignalRBuilder to enable configuring this:

```C#
services.AddSignalR()
              .AddHubOptions<Chat>(o =>
              {
                   o.KeepAliveTimeout = TimeSpan.FromSeconds(5);
                   o.SupportedProtocols.Add("msgpack");
              });
```

mikaelm12:
Okay, replaced `HubOptions` with `HubOptions<THub>`. HubOptions are no longer global. This probably could have logically been split into two PRs but oh well. 
The internals are quite different now btw

src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
```diff
@@ -24,5 +25,7 @@ public class HubOptions
         /// functionality because of the polling mechanism. 
         /// </remarks>
         public TimeSpan KeepAliveInterval { get; set; } = DefaultKeepAliveInterval;
+
+        public List<string> SupportedProtocols { get; set; } = new List<string> { DefaultProtocols.JSONProtocol, DefaultProtocols.MsgPackProtocol};
```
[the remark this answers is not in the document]
davidfowl: I was waiting for this to come up. I think by default you support all available transports. That is, not specified means all. There also needs to be validation like what if you clear all? What does that mean?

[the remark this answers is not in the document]
BrennanConroy: Could we somehow grab all available transports when HubOptions is created and populate the list before running the users action on it?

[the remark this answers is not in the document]
davidfowl: We have nested generics so I'm not sure how well it will work. You'd need to get all the configured IHubProtocol implementations inside of an `IConfigureOptions<HubOptions<THub>>` which makes it a bit difficult as there's no explicit call to register a hub.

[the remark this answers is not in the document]
anurse: Other fun things: MsgPack is in a completely separate assembly that isn't available in `AspNetCore.All` currently.

samples/SocketsSample/Startup.cs
```diff
@@ -20,11 +20,25 @@ public void ConfigureServices(IServiceCollection services)
...
-            services.AddSignalR(options =>
+            services.AddSignalR()
+
+            .AddHubOptions<Chat>(options =>
+            {
+                options.KeepAliveInterval = TimeSpan.FromSeconds(5);
+            })
+            .AddHubOptions<DynamicChat>(options =>
             {
-                // Faster pings for testing
```
mikaelm12: Don't know why I removed this. I'll bring it back

src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
```diff
@@ -15,10 +18,49 @@ public static ISignalRBuilder AddSignalR(this IServiceCollection services)
                 .AddJsonProtocol();
         }
 
+        public static ISignalRBuilder AddGlobalHubOptions(this ISignalRBuilder signalRBuilder, Action<HubOptions> options)
```
davidfowl: This should be an overload of AddSignalR. There should be no global options call.

src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
```diff
@@ -15,10 +18,49 @@ public static ISignalRBuilder AddSignalR(this IServiceCollection services)
...
+        }
+
         public static ISignalRBuilder AddHubOptions<THub>(this ISignalRBuilder signalrBuilder, Action<HubOptions<THub>> options) where THub : Hub
         {
             signalrBuilder.Services.Configure(options);
             return signalrBuilder;
         }
     }
+
+    public class HubOptionsSetup : IConfigureOptions<HubOptions>
```
davidfowl: Put these in separate files.

src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
mikaelm12: Makes sense

mikaelm12:
Okay, this is working now and is ready for review

samples/SocketsSample/Startup.cs
```diff
@@ -20,11 +20,11 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddSockets();
 
-            services.AddSignalR(options =>
-            {
+            services.AddSignalR(options => {
                 // Faster pings for testing
                 options.KeepAliveInterval = TimeSpan.FromSeconds(5);
             })
+
```
davidfowl: nit: remove space

src/Microsoft.AspNetCore.SignalR.Core/HubOptionsSetup.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            foreach (IHubProtocol hubProtocol in protocols)
+            {
+                _protocols.Add(hubProtocol.Name);
+            }
+        }
+
+        public void Configure(HubOptions options)
+        {
+            if( options.SupportedProtocols == null)
```
davidfowl: nit: formatting

src/Microsoft.AspNetCore.SignalR.Core/HubOptionsSetup`T.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        private readonly HubOptions _hubOptions;
+        public HubOptionsSetup(IOptions<HubOptions> options)
+        {
+            _hubOptions = options.Value;
+        }
+
+        public void Configure(HubOptions<THub> options)
+        {
+            if(_hubOptions.SupportedProtocols == null)
```
davidfowl: Can this ever be null?

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -33,11 +31,11 @@ public DefaultHubProtocolResolver(IOptions<HubOptions> options, IEnumerable<IHub
             }
         }
 
-        public IHubProtocol GetProtocol(string protocolName, HubConnectionContext connection)
+        public IHubProtocol GetProtocol(string protocolName, List<string> supportedProtocols, HubConnectionContext connection)
         {
             protocolName = protocolName ?? throw new ArgumentNullException(nameof(protocolName));
 
-            if (_availableProtocols.TryGetValue(protocolName, out var protocol))
+            if (_availableProtocols.TryGetValue(protocolName, out var protocol) && (supportedProtocols ==  null || supportedProtocols.Contains(protocolName)))
```
davidfowl: What about casing? Do the case rules match the dictionary? It should be insensitive.

src/Microsoft.AspNetCore.SignalR.Core/SignalRDependencyInjectionExtensions.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.AspNetCore.SignalR.Core;
 using Microsoft.AspNetCore.SignalR.Internal;
+using Microsoft.Extensions.Options;
```
davidfowl: Revert this file?

src/Microsoft.AspNetCore.SignalR/HubOptionsDependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class HubOptionsDependencyInjectionExtensions
+    {
+
```
davidfowl: nit: remove empty space.

src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
```diff
@@ -3,22 +3,25 @@
...
-        {
-            services.Configure(configure);
             services.AddSockets();
             return services.AddSignalRCore()
                 .AddJsonProtocol();
         }
+
+        public static ISignalRBuilder AddSignalR(this IServiceCollection services, Action<HubOptions> options)
+        {
+            services.AddSingleton<IConfigureOptions<HubOptions>, HubOptionsSetup>()
```
davidfowl: This should be in the parameterless AddSignalR

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -274,7 +274,7 @@ public class HubEndpointTests
         {
             var serviceProvider = HubEndPointTestUtils.CreateServiceProvider(services =>
             {
-                services.Configure<HubOptions>(hubOptions =>
+                services.Configure<HubOptions<SimpleHub>>(hubOptions =>
```
davidfowl: Why isn't this just the same as before? why did you change it?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1325,10 +1325,15 @@ public void HubsCannotHaveOverloadedMethods()
         [MemberData(nameof(StreamingMethodAndHubProtocols))]
         public async Task HubsCanStreamResponses(string method, IHubProtocol protocol)
         {
-            var serviceProvider = HubEndPointTestUtils.CreateServiceProvider();
+            var serviceProvider = HubEndPointTestUtils.CreateServiceProvider(services =>
+            {
+                services.AddSignalR(options =>
```
davidfowl: Why did you add this?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1673,7 +1678,7 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
         public async Task DoesNotWritePingMessagesIfSufficientOtherMessagesAreSent()
         {
             var serviceProvider = HubEndPointTestUtils.CreateServiceProvider(services =>
-                services.Configure<HubOptions>(options =>
+                services.Configure<HubOptions<MethodHub>>(options =>
```
davidfowl: Why did you change these?

test/Microsoft.AspNetCore.SignalR.Tests/Internal/DefaultHubProtocolResolverTests.cs
```diff
@@ -22,18 +22,48 @@ public class DefaultHubProtocolResolverTests
             new MessagePackHubProtocol()
         };
 
+        private static readonly List<string> AllProtocolNames = new List<string> { "json", "messagepack" };
```
davidfowl: Why do you have this?

src/Microsoft.AspNetCore.SignalR.Core/HubOptionsSetup`T.cs
mikaelm12: It shouldn't be. I'll remove this check

mikaelm12:
Okay. The tests weren't actually testing the correct thing before. I fixed them to ensure that the global hub options are being configured.

I changed the HubOptions<THub> class values to be null by default to add logic to determine if we should use the global options. We have to do this because configure on HubOptionsSetup<THub> doesn't run unless we explicitly call `addHubOptions` for a specific THub.

samples/SocketsSample/Startup.cs
```diff
@@ -20,8 +20,7 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddSockets();
 
-            services.AddSignalR(options =>
-            {
+            services.AddSignalR(options => {
```
BrennanConroy: revert

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -32,16 +34,30 @@ public class HubEndPoint<THub> where THub : Hub
...
+            _globalHubOptions = globalHubOptions.Value;
             _logger = loggerFactory.CreateLogger<HubEndPoint<THub>>();
             _userIdProvider = userIdProvider;
             _dispatcher = dispatcher;
         }
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
-            var connectionContext = new HubConnectionContext(connection, _hubOptions.KeepAliveInterval, _loggerFactory);
+            var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval;
```
BrennanConroy: Add a comment saying `HubOptions<T>` is preferred over `HubOptions`

src/Microsoft.AspNetCore.SignalR.Core/HubOptionsSetup.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubOptionsSetup : IConfigureOptions<HubOptions>
+    {
+        private readonly List<string> _protocols = new List<string>();
+
+        public HubOptionsSetup(IEnumerable<IHubProtocol> protocols)
+        {
+            foreach (IHubProtocol hubProtocol in protocols)
```
BrennanConroy: var

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -33,11 +32,11 @@ public DefaultHubProtocolResolver(IOptions<HubOptions> options, IEnumerable<IHub
             }
         }
 
-        public IHubProtocol GetProtocol(string protocolName, HubConnectionContext connection)
+        public IHubProtocol GetProtocol(string protocolName, List<string> supportedProtocols, HubConnectionContext connection)
         {
             protocolName = protocolName ?? throw new ArgumentNullException(nameof(protocolName));
 
-            if (_availableProtocols.TryGetValue(protocolName, out var protocol))
+            if (_availableProtocols.TryGetValue(protocolName, out var protocol) && (supportedProtocols ==  null || supportedProtocols.Contains(protocolName, StringComparer.OrdinalIgnoreCase)))
```
BrennanConroy: nit: remove extra space

src/Microsoft.AspNetCore.SignalR.Core/Internal/IHubProtocolResolver.cs
```diff
@@ -2,11 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using System.Collections.Generic;
```
BrennanConroy: nit: sort

test/Microsoft.AspNetCore.SignalR.Tests/Internal/DefaultHubProtocolResolverTests.cs
```diff
@@ -22,18 +22,47 @@ public class DefaultHubProtocolResolverTests
...
```
davidfowl: Why are you hardcode this?

davidfowl: move to top of the file?

src/Microsoft.AspNetCore.SignalR.Core/HubOptions`T.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubOptions<THub> : HubOptions where THub : Hub
+    {
+        public new TimeSpan? NegotiateTimeout { get; set; } = null;
```
JamesNK: It feels really weird to me that these properties are defined again and are using new to hide the original properties. Is it because you want NegotiateTimeout and KeepAliveInterval to be nullable here? What scenario is this for?

davidfowl: Yea, maybe they should just be null in the base class.

mikaelm12: Good question. 
The properties are null in this class and not the base class because we want the default values from the base class, but we need to be able to tell if a specific `HubOptions<THub>` has been configured (aka the options are not null) 
It allows us to differentiate between the scenarios in which there are only global options defined, global and local so we know which settings to enforce. 

That being said, yeah we could just make them null in the base class but I do like the idea of default values.

davidfowl: We can just have defaults come from the IConfigureOptions<HubOptions> instead of being on the HubOptions class. It's a bit cleaner.

src/Microsoft.AspNetCore.SignalR.Core/HubOptionsSetup.cs
```diff
@@ -26,6 +27,18 @@ public void Configure(HubOptions options)
...
             }
+
+            if (options.KeepAliveInterval == null)
+            {
+                // The default keep - alive interval.This is set to exactly half of the default client timeout window,
+                // to ensure a ping can arrive in time to satisfy the client timeout.
+                options.KeepAliveInterval = TimeSpan.FromSeconds(15);
+            }
+
+            if(options.NegotiateTimeout == null)
```
davidfowl: nit: formatting

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -32,16 +34,31 @@ public class HubEndPoint<THub> where THub : Hub
...
+
+            if (supportedProtocols != null && supportedProtocols.Count == 0)
+            {
+                throw new InvalidOperationException("There are no supported protocols");
+            }
+
+            var connectionContext = new HubConnectionContext(connection, keepAlive.Value, _loggerFactory);
 
-            if (!await connectionContext.NegotiateAsync(_hubOptions.NegotiateTimeout, _protocolResolver, _userIdProvider))
+            if (!await connectionContext.NegotiateAsync(negotiateTimeout.Value, supportedProtocols, _protocolResolver, _userIdProvider))
```
davidfowl: This can throw if negotiateTimeout is null...

mikaelm12: It shouldn't be null since it gets set in the HubOptionsSetup

davidfowl: What if I explicitly set it to null?

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -32,16 +34,31 @@ public class HubEndPoint<THub> where THub : Hub
...
             _userIdProvider = userIdProvider;
             _dispatcher = dispatcher;
         }
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
-            var connectionContext = new HubConnectionContext(connection, _hubOptions.KeepAliveInterval, _loggerFactory);
+            // We check to see if HubOptions<THub> are set because those take precedence over global hub options.
+            var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval;
+            var negotiateTimeout = _hubOptions.NegotiateTimeout ?? _globalHubOptions.NegotiateTimeout;
```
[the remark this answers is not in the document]
mikaelm12: I think this makes sense.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -32,16 +34,32 @@ public class HubEndPoint<THub> where THub : Hub
...
 
         public async Task OnConnectedAsync(ConnectionContext connection)
         {
-            var connectionContext = new HubConnectionContext(connection, _hubOptions.KeepAliveInterval, _loggerFactory);
+            // We check to see if HubOptions<THub> are set because those take precedence over global hub options.
+            // Then set the keepAlive and negotiateTimeout values to the defaults in HubOptionsSetup incase they were explicitly set to null.
+            var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval ?? HubOptionsSetup.DefaultKeepAliveInterval;
+            var negotiateTimeout = _hubOptions.NegotiateTimeout ?? _globalHubOptions.NegotiateTimeout ?? HubOptionsSetup.DefaultNegotiateTimeout;
+            var supportedProtocols = _hubOptions.SupportedProtocols;
+            if (supportedProtocols == null && _globalHubOptions.SupportedProtocols != null)
```
JamesNK: You could replace this if statement with a null coalesce on the line above. It doesn't matter if `_globalHubOptions.SupportedProtocols` protocols is null based on the check line 55.

src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
```diff
@@ -2,27 +2,16 @@
...
     {
-        /// <summary>
-        /// The default keep-alive interval. This is set to exactly half of the default client timeout window,
-        /// to ensure a ping can arrive in time to satisfy the client timeout.
-        /// </summary>
-        public static readonly TimeSpan DefaultKeepAliveInterval = TimeSpan.FromSeconds(15);
+        public TimeSpan? NegotiateTimeout { get; set; } = null;
 
-        public TimeSpan NegotiateTimeout { get; set; } = TimeSpan.FromSeconds(5);
+        public TimeSpan? KeepAliveInterval { get; set; } = null;
```
anurse: Bring the docs back, and add a note about `null` indicating the default interval (15 seconds).

src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
```diff
@@ -2,27 +2,16 @@
...
-        /// <summary>
-        /// The interval at which keep-alive messages should be sent. The default interval
-        /// is 15 seconds.
-        /// </summary>
-        /// <remarks>
-        /// This interval is not used by the Long Polling transport as it has inherent keep-alive
-        /// functionality because of the polling mechanism. 
-        /// </remarks>
-        public TimeSpan KeepAliveInterval { get; set; } = DefaultKeepAliveInterval;
+        public List<string> SupportedProtocols { get; set; } = null;
```
anurse: Make this `IList<string>`

We're saying `null` is "any protocol" right? Let's throw a doc comment on this explaining it as well.

src/Microsoft.AspNetCore.SignalR.Core/Internal/IHubProtocolResolver.cs
```diff
@@ -1,12 +1,13 @@
...
 
+using System.Collections.Generic;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 
 namespace Microsoft.AspNetCore.SignalR.Internal
 {
     public interface IHubProtocolResolver
     {
-        IHubProtocol GetProtocol(string protocolName, HubConnectionContext connection);
+        IHubProtocol GetProtocol(string protocolName, List<string> supportedProtocols, HubConnectionContext connection);
```
anurse: `List<T>` in a public API!



Use `IReadOnlyList<string>`. The method only needs to be able to read from the list, and doesn't care what kind of list you give it.

mikaelm12: This should also just be an `IList<string>`, right? It get passed the supportedProtocol list from the HubEndPoint which is an IList

src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
JamesNK: The comment is a bit more complicated than that because `null` could indicate the default, or it could indicate that the global option should be used

src/Microsoft.AspNetCore.SignalR.Core/Internal/IHubProtocolResolver.cs
JamesNK: Should it not be `List<T>` even if it is in the internal namespace? I wasn't sure about that when I looked at the code.

anurse: In this case, the API is likely to be made properly public at some point. Also, "pubternal" (public in the Internal namespace) types should still try to keep to our public API standards.

src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
anurse: Ugh, yeah. Ok, save it for now, we'll need to doc it at some point though.

PR #1498, opened 2018-02-27
Serialize once per format to increase perf in broadcasting scenarios

BrennanConroy:
Also found an issue where we registered KeepAliveTick too soon and fixed it.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
@@ -59,7 +59,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
                 return false;
             }
 
-            return _dataEncoder == readerWriter._dataEncoder && _hubProtocol == readerWriter._hubProtocol;
+            return ReferenceEquals(_dataEncoder, readerWriter._dataEncoder) && _hubProtocol.Equals(readerWriter._hubProtocol);
         }
 
         public override int GetHashCode()
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
@@ -16,8 +16,6 @@ namespace Microsoft.AspNetCore.SignalR.Internal
         public readonly IHubProtocol HubProtocol;
         public readonly IDataEncoder DataEncoder;
 
-        public string ProtocolName => _hubProtocol.Name;
-
         public HubProtocolReaderWriter(IHubProtocol hubProtocol, IDataEncoder dataEncoder)
         {
             HubProtocol = hubProtocol;
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
@@ -53,13 +53,15 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
         public override bool Equals(object obj)
         {
-            HubProtocolReaderWriter readerWriter = obj as HubProtocolReaderWriter;
+            var readerWriter = obj as HubProtocolReaderWriter;
             if (readerWriter == null)
             {
                 return false;
             }
 
-            return ReferenceEquals(_dataEncoder, readerWriter._dataEncoder) && _hubProtocol.Equals(readerWriter._hubProtocol);
+            // Note: ReferenceEquals on HubProtocol works for our implementation of IHubProtocolResolver because we use Singletons from DI
+            // However if someone replaces the implementation and returns a new ProtocolResolver for every connection they wont get the perf benefits
+            return ReferenceEquals(_dataEncoder, readerWriter._dataEncoder) && ReferenceEquals(_hubProtocol, readerWriter._hubProtocol);
         }
 
         public override int GetHashCode()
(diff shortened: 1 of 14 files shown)


BrennanConroy:
Before:
127k - Json
145k - MsgPack

After:
170k - Json
425k - MsgPack

src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
```diff
@@ -50,5 +50,23 @@ public byte[] WriteMessage(HubMessage hubMessage)
...
+        public override bool Equals(object obj)
+        {
+            var readerWriter = obj as HubProtocolReaderWriter;
+            if (readerWriter == null)
+            {
+                return false;
+            }
+
+            // Note: ReferenceEquals on HubProtocol works for our implementation of IHubProtocolResolver because we use Singletons from DI
+            // However if someone replaces the implementation and returns a new ProtocolResolver for every connection they wont get the perf benefits
```
anurse: TBH, still nervous about this. It's not even that they won't get the perf benefits, they'll get significantly more memory usage (and `O(n)` memory growth where `n` is the number of connections).

src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
```diff
@@ -50,5 +50,23 @@ public byte[] WriteMessage(HubMessage hubMessage)
...
+            return ReferenceEquals(_dataEncoder, readerWriter._dataEncoder) && ReferenceEquals(_hubProtocol, readerWriter._hubProtocol);
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
anurse: Comment as to why this is OK (we never want to put this in a hash-table)

src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
anurse: Also, the client is in control of how much this memory grows.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMessage.cs
```diff
@@ -1,12 +1,45 @@
...
+            for (var i = 0; i < _serializedMessages.Count; i++)
+            {
+                if (_serializedMessages[i].ProtocolReaderWriter.Equals(protocolReaderWriter))
+                {
+                    return _serializedMessages[i].Message;
+                }
+            }
+
+            var bytes = protocolReaderWriter.WriteMessage(this);
+            _serializedMessages.Add(new SerializedMessage(protocolReaderWriter, bytes));
```
anurse: If we capped the cache size, that would relieve some of my concerns about client-controlled linear memory growth from the cache.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMessage.cs
```diff
@@ -1,12 +1,45 @@
...
     public abstract class HubMessage
     {
         protected HubMessage()
         {
         }
+
+        // Initialize with capacity 4 for the 2 built in protocols and 2 data encoders
+        private readonly List<SerializedMessage> _serializedMessages = new List<SerializedMessage>(4);
+
+        public byte[] WriteMessage(HubProtocolReaderWriter protocolReaderWriter)
```
anurse: Still convinced it's going to bite us that this is on the `HubMessage` type itself.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -84,17 +80,41 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+            // Don't wait for the lock, if it returns false that means someone wrote to the connection
+            // and we don't need to send a ping anymore
+            if (!await _writeLock.WaitAsync(0))
+            {
+                return;
+            }
+
+            try
+            {
+                _connectionContext.Transport.Output.Write(_cachedPingMessage);
```
anurse: Assert `_cachedPingMessage != null`. It depends upon the caller initializing properly so we should assert it.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMessage.cs
anurse: Having this on here is going to mess with the singleton `PingMessage.Instance`... I know you special case that elsewhere but it still makes me nervous dangit!

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -154,11 +174,18 @@ public virtual void Abort()
...
                                     ProtocolReaderWriter = new HubProtocolReaderWriter(protocol, dataEncoder);
+                                    _cachedPingMessage = ProtocolReaderWriter.WriteMessage(PingMessage.Instance);
 
                                     Log.UsingHubProtocol(_logger, protocol.Name);
 
                                     UserIdentifier = userIdProvider.GetUserId(this);
 
+                                    if (Features.Get<IConnectionInherentKeepAliveFeature>() == null)
+                                    {
+                                        // Only register KeepAlive after protocol negotiated otherwise KeepAliveTick could try to write without having a ProtocolReaderWriter
```
anurse: Good comment

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMessage.cs
davidfowl: I don’t really understand the concern. Are you worrried about a misconfigured application? We can cap it at 10 if that makes you feel better.

src/Microsoft.AspNetCore.SignalR.Common/Internal/HubProtocolReaderWriter.cs
davidfowl: How is it client controlled? The server is the one that configures the amount of protocols supported? Are you taking about in theory you can return a different instance of a protocol per connection? Again how is that a real concern?

BrennanConroy:
Changed the backpressure on the pipes and JSON jumped to ~250k now. We'll have to talk about actual default values, not for this PR though.

BrennanConroy:
:up: 📅

PR #1501, opened 2018-02-27
Pipelines in corefx

pakrym:
Disregard source.props

### DIFF
--- a/samples/SocketsSample/EndPoints/MessagesEndPoint.cs
+++ b/samples/SocketsSample/EndPoints/MessagesEndPoint.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Buffers;
 using System.Collections.Generic;
 using System.IO.Pipelines;
 using System.Text;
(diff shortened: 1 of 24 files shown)


samples/SocketsSample/EndPoints/MessagesEndPoint.cs
```diff
@@ -64,10 +65,11 @@ private Task Broadcast(string text)
         private Task Broadcast(byte[] payload)
         {
             var tasks = new List<Task>(Connections.Count);
+            async Task<FlushResult> ToTask(PipeAwaiter<FlushResult> awaiter) => await awaiter;
 
             foreach (var c in Connections)
             {
-                tasks.Add(c.Transport.Output.WriteAsync(payload));
+                tasks.Add(ToTask(c.Transport.Output.WriteAsync(payload)));
```
mikaelm12: This is a little weird. Why exactly is this happening

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -93,7 +93,7 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferMode requestedT
                     {
                         var result = await pipelineReader.ReadAsync();
                         var input = result.Buffer;
-                        if (result.IsCancelled || (input.IsEmpty && result.IsCompleted))
+                        if (result.IsCanceled || (input.IsEmpty && result.IsCompleted))
```
mikaelm12: There was talk in the demo meeting about a new API for something like HasData (I don't remember what was actually said). Coming later?

pakrym: Definitely not in this iteration.

samples/SocketsSample/EndPoints/MessagesEndPoint.cs
pakrym: WriteAsync returns custom Awaiter object, not Task that the list requires.

PR #1510, opened 2018-02-28
Add some more logging

BrennanConroy:


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -139,6 +139,7 @@ private void ResetTimeoutTimer()
             _connectionActive = new CancellationTokenSource();
             using (var memoryStream = new MemoryStream())
             {
+                _logger.SendingHubNegotiate();
```
davidfowl: Do we have a log after it succeeds or fails?

BrennanConroy:
Mostly too lazy last night

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -344,8 +345,10 @@ private Task InvokeStreamCore(string methodName, InvocationRequest irq, object[]
         private async Task OnDataReceivedAsync(byte[] data)
         {
             ResetTimeoutTimer();
+            _logger.ParsingMessages(data.Length);
             if (_protocolReaderWriter.ReadMessages(data, _binder, out var messages))
```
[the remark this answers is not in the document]
BrennanConroy: Would that be a warning, since it shouldn't happen?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
BrennanConroy: If it succeeds StartAsync returns successfully, if it fails you'll get an exception from StartAsync. Would a log still be useful?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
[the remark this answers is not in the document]
davidfowl: Yep

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: No, that's fine.

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
```diff
@@ -119,5 +118,64 @@ protected override bool TryComputeLength(out long length)
...
+                LoggerMessage.Define<long, Uri>(LogLevel.Debug, new EventId(10, "SendingMessages"), "Sending {count} bytes to the server using url: {url}.");
+
+            private static readonly Action<ILogger, Exception> _sentSuccessfully =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(11, "SentSuccessfully"), "Message(s) sent successfully.");
+
+            private static readonly Action<ILogger, Exception> _noMessages =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(12, "NoMessages"), "No messages in batch to send.");
+
+            private static readonly Action<ILogger, Uri, Exception> _errorSending =
+                LoggerMessage.Define<Uri>(LogLevel.Error, new EventId(13, "ErrorSending"), "Error while sending to '{url}'.");
```
anurse: Put a comment here similar to the ones in the other Log classes indicating that if you add a new EventID you need to reconcile it with the other loggers.

anurse: Also, start them all at `100` so we don't have to deal with pushing event IDs around if we add a new one.

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.Log.cs
```diff
@@ -0,0 +1,88 @@
...
+            private static readonly Action<ILogger, Exception> _closingConnection =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(7, "ClosingConnection"), "The server is closing the connection.");
+
+            private static readonly Action<ILogger, Exception> _receivedMessages =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(8, "ReceivedMessages"), "Received messages from the server.");
+
+            private static readonly Action<ILogger, Uri, Exception> _errorPolling =
+                LoggerMessage.Define<Uri>(LogLevel.Error, new EventId(9, "ErrorPolling"), "Error while polling '{pollUrl}'.");
+
+            // EventId's 100 - 106 used in SendUtils
```
anurse: nit: `EventIds` (it's plural, not possessive)

src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs
```diff
@@ -119,5 +118,66 @@ protected override bool TryComputeLength(out long length)
...
+            private static readonly Action<ILogger, Exception> _sentSuccessfully =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(104, "SentSuccessfully"), "Message(s) sent successfully.");
+
+            private static readonly Action<ILogger, Exception> _noMessages =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(105, "NoMessages"), "No messages in batch to send.");
+
+            private static readonly Action<ILogger, Uri, Exception> _errorSending =
+                LoggerMessage.Define<Uri>(LogLevel.Error, new EventId(106, "ErrorSending"), "Error while sending to '{url}'.");
+
+            // When adding a new log message make sure to check with the other locations that share these logs to not have conflicting event ids
```
anurse: List the other locations. Also a note about why `100` range is used (to avoid conflicts since these events are used by multiple logger categories)

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.Log.cs
```diff
@@ -0,0 +1,88 @@
...
+            private static readonly Action<ILogger, int, Exception> _messageToApp =
+                LoggerMessage.Define<int>(LogLevel.Debug, new EventId(7, "MessageToApp"), "Passing message to application. Payload size: {count}.");
+
+            private static readonly Action<ILogger, Exception> _eventStreamEnded =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(8, "EventStreamEnded"), "Server-Sent Event Stream ended.");
+
+            private static readonly Action<ILogger, long, Exception> _parsingSSE =
+                LoggerMessage.Define<long>(LogLevel.Debug, new EventId(9, "ParsingSSE"), "Received {count} bytes. Parsing SSE frame.");
+
+            // EventId's 100 - 106 used in SendUtils
```
anurse: same nit as above: `EventIds`

PR #1513, opened 2018-02-28
Emit warnings after off is called

olofd:
If a handler has been registered, and then gets unregistered, the key/methodname still exists in the map and signalr does not warn about a missing handler. Might be by design, but I like it this way.

### DIFF
--- a/client-ts/signalr/src/HubConnection.ts
+++ b/client-ts/signalr/src/HubConnection.ts
@@ -267,6 +267,9 @@ export class HubConnection {
         const removeIdx = handlers.indexOf(method);
         if (removeIdx !== -1) {
             handlers.splice(removeIdx, 1);
+            if(handles.length === 0) {
+                handles.delete(methodName);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


client-ts/signalr/src/HubConnection.ts
```diff
@@ -267,6 +267,9 @@ export class HubConnection {
         const removeIdx = handlers.indexOf(method);
         if (removeIdx !== -1) {
             handlers.splice(removeIdx, 1);
+            if(handles.length === 0) {
+                handles.delete(methodName);
```
davidfowl: Isn't handlers an array? What does this do?

olofd: Fixed.

client-ts/signalr/src/HubConnection.ts
```diff
@@ -267,6 +267,9 @@ export class HubConnection {
         const removeIdx = handlers.indexOf(method);
         if (removeIdx !== -1) {
             handlers.splice(removeIdx, 1);
+            if(handlers.length === 0) {
```
davidfowl: nit: add a space between the if and the (

olofd: (Note to self: Stop doing PR's on the phone)

davidfowl:
@olofd can you add a test for this? If not I can, just let me know.

Issue #1490, opened 2018-02-25
Object reference not set to an instance of an object


SabirHossain:
Object reference not set to an instance of an object. signalr-client-1.0.0-alpha2-final.js:494:32

BrennanConroy:
Could you provide any other information?

SabirHossain:
If anything fails on the Hub and if you forget to catch the fail then the
error will appear on the client side console.. Ex. I was trying to save
some data on the database using EF Core just for test I didn't use any
'try' and 'catch' and after complete the database action I was also calling
a client side method but when I was debugging I saw after the database
changes fail it don't execute the client side method as usual but then why
the client side error happened?  [aspnet/SignalR] Object reference not set
to an instance of an object (#1490)

BrennanConroy:
The "Object reference not set to an instance of an object" is an error thrown from your Hub. We could probably clean up the client side log a little to make that more clear

SabirHossain:
Yes you are right, I'm sorry for inconvenience, the problem was in database
updating in the Hub. We need a better reference for .net core 2.0 and
signalr-alpha2. the old documention have some deprecated coeds. Thank you.

davidfowl:
We should fix how we expose errors on the client side.

alexzaytsev-newsroomly:
@davidfowl the current way is not so bad, i actually liked being able to see server-side errors on the client while debugging; what would be nice if the console would indicate that this happened on the server; further, it would be great if we could elect to opt-out of this for prod builds.

davidfowl:
That's what this bug is about, improving that error experience. It should be wrapped in a specific exception type. That might be enough to distinguish.

JamesNK:
In the .NET client the error comes back as a string on `CompletionMessage`.

https://github.com/aspnet/SignalR/blob/1790d6bdf088bab120cd14912d6d625d983fb077/test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs#L452

What we could do is prepend some text to it to give it some more context, i.e. `An error occurred invoking method "MethodThatThrows" on the server: BOOM!`

BrennanConroy:
Yeah, I think all we need is a little info wrapped around the error message. It's the same in the js client

JamesNK:
Fixed https://github.com/aspnet/SignalR/commit/846432c9ac4599cbd24ceac6e756627d3f1859c7

PR #1532, opened 2018-03-06
Improve unexpected server error message to client

JamesNK:
I've included the method name to keep it consistent with the existing unauthorized error message

https://github.com/aspnet/SignalR/blob/6a2d41cc9ffc6e64af6a3dbc6d3cd0ed2bc44f52/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs#L160-L166

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
@@ -195,18 +195,23 @@ namespace Microsoft.AspNetCore.SignalR.Internal
                 catch (TargetInvocationException ex)
                 {
                     Log.FailedInvokingHubMethod(_logger, hubMethodInvocationMessage.Target, ex);
-                    await SendInvocationError(hubMethodInvocationMessage, connection, ex.InnerException.Message);
+                    await SendInvocationError(hubMethodInvocationMessage, connection, BuildUnexpectedErrorMessage(hubMethodInvocationMessage.Target, ex.InnerException));
                 }
                 catch (Exception ex)
                 {
                     Log.FailedInvokingHubMethod(_logger, hubMethodInvocationMessage.Target, ex);
-                    await SendInvocationError(hubMethodInvocationMessage, connection, ex.Message);
+                    await SendInvocationError(hubMethodInvocationMessage, connection, BuildUnexpectedErrorMessage(hubMethodInvocationMessage.Target, ex));
                 }
                 finally
                 {
                     hubActivator.Release(hub);
                 }
             }
+        }
+
+        private string BuildUnexpectedErrorMessage(string methodName, Exception exception)
+        {
+            return $"An unexpected error occurred invoking '{methodName}' on the server. {exception.GetType().Name}: {exception.Message}";
         }
 
         private async Task StreamResultsAsync(string invocationId, HubConnectionContext connection, IAsyncEnumerator<object> enumerator)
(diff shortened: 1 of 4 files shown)


JamesNK:
Fixs https://github.com/aspnet/SignalR/issues/1490

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -209,6 +209,11 @@ public override IReadOnlyList<Type> GetParameterTypes(string methodName)
             }
         }
 
+        private string BuildUnexpectedErrorMessage(string methodName, Exception exception)
+        {
+            return $"An unexpected error occurred invoking '{methodName}' on the server. {exception.GetType().Name}: {exception.Message}";
```
jkotalik: null check on exception?

anurse: It can't really be `null`. It's passed a value from a `catch` block. At most it would be an `Assert` and we don't really do those kind of asserts.

anurse: Wasn't 100% positive of that so I verified :): https://sharplab.io/#v2:C4LgTgrgdgPgAgBgARwIwG4CwAoOBmFAJiQGEkBvHJalAtANhQBYVUAOACgEoBIS7GoKTAwATwpUh1HsAAWYAPYB3JFAgAbdVgFSkAXyQBjAIbBDsjgFEAHoYCmAB2ABLBVCR3rXCTt000AJwcAESeSM4Azqoa6gD8SMFIANRIHGEAvOnRmlxc2n56ktSF2HpAA=

Throwing `null` appears to yield a `NullReferenceException`, so the exception value wouldn't be null.

JamesNK:
Separated expected vs unexpected error messages. Fixed broken tests. JS tests todo

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
JamesNK: You can throw non-Exceptions via IL but we're explicitly catching `Exception`

JamesNK:
client-ts functional tests fixed (apart from failing tests that were already failing prior to my changes)

davidfowl:
We're going to add a hub option that determines whether to send the exception back to the client just like we did in previous SignalR.

anurse:
> apart from failing tests that were already failing prior to my changes

Which tests were failing? I'm not aware of any broken tests right now. In general, we'll want to the functional tests to be reliable so if there are some that are flaking out on your environment, we should track them down ;).

anurse:
> We're going to add a hub option that determines whether to send the exception back to the client just like we did in previous SignalR.

After https://github.com/aspnet/SignalR/pull/1492 is merged, we can put that on `HubOptions`/`HubOptions<THub>`

JamesNK:
Attached (docx to keep formatting)

[functionalerrors.docx](https://github.com/aspnet/SignalR/files/1783502/functionalerrors.docx)

davidfowl:
@JamesNK did you attach a word document to a PR... 😄

anurse:
Ok, I think I've seen those and might know the solution. No need to do anything to fix them, just wanted to see what issues you were seeing. Thanks!

JamesNK:
@davidfowl Would you prefer a pdf? 😇

davidfowl:
The hang happened again https://travis-ci.org/aspnet/SignalR/jobs/349628966. I need to add the timeout back. /cc @BrennanConroy

davidfowl:
> Did you restart the job? That reuses the url.

Yea, I did

BrennanConroy:
Was it the large websocket message?

davidfowl:
> Was it the large websocket message?

Yea, the same test that always fails, the 4 MB one.

BrennanConroy:
We need to figure out the cause :(

PR #1520, opened 2018-03-02
Fix incorrect package ID reference in README

anurse:

PR #1533, opened 2018-03-06
Fix UMD module version of @aspnet/signalr-protocol-msgpack

### LINKED ISSUE #1530: Incorrect dependencies in browser version of signalr-protocol-msgpack 1.0.0-preview1-final
Trying to bundle it using webpack produces:
>ERROR in ./node_modules/@aspnet/signalr-protocol-msgpack/dist/browser/signalr-protocol-msgpack.js
>Module not found : error : Can't resolve '    ode_modulesmsgpack5index.js' in 'D:\Programs\node_modules\\@aspnet\signalr-protocol-msgpack\dist\browser'
     @ ./node_modules/@aspnet/signalr-protocol-msgpack/dist/browser/signalr-protocol-msgpack.js 6:138-270
     @ dll vendor
ERROR in ./node_modules/@aspnet/signalr-protocol-msgpack/dist/browser/signalr-protocol-msgpack.js
Module not found : error : Can't resolve '../../node_modules/msgpack5/index.js' in 'D:\Programs\node_modules\\@aspnet\signalr-protocol-msgpack\dist\browser'
     @ ./node_modules/@aspnet/signalr-protocol-msgpack/dist/browser/signalr-protocol-msgpack.js 6:89-136
     @ dll vendor

webpack 3.10.0
node 8.9.3

anurse:
The problem was that the `rollup` was injecting `require("../../node_modules/msgpack5/index.js")` into the UMD build. It looks like it **was** working when used as a browser `<script>` reference. It was only when used as an AMD module that it failed because of the broken `require`. There's a separate issue here which is that WebPack was choosing to use the `dist/browser` script instead of the `dist/cjs` script it's supposed to use. I'll file that issue for later, but as long as this change makes it work in `webpack` (which I will test before merging), then we can ship this and #1520 as `1.0.0-preview1-update1` of the respective packages.

TODO:
* [x] Verify this fixes webpack (and thus will Fix #1530)

### DIFF
--- a/client-ts/rollup-base.js
+++ b/client-ts/rollup-base.js
@@ -22,31 +22,13 @@ export default function(rootDir, moduleGlobals) {
             banner: "/* @license\r\n" +
                 " * Copyright (c) .NET Foundation. All rights reserved.\r\n" +
                 " * Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.\r\n" +
-                "*/",
+                " */",
             globals: moduleGlobals,
         },
-        external: function (m) {
-            let match = m.match(/node_modules/);
-            if (match) {
-                let moduleName = m.substring(match.index + "node_modules".length + 1);
-                let slashIndex = moduleName.indexOf('/');
-                if(slashIndex < 0) {
-                    slashIndex = moduleName.indexOf('\\');
-                }
-                moduleName = moduleName.substring(0, slashIndex);
 
-                if(polyfills.indexOf(moduleName) >= 0) {
-                    // This is a polyfill
-                    console.log(`Importing polyfill: ${m}`);
-                    return false;
-                }
-                else if(!allowed_externals.indexOf(moduleName)) {
-                    console.log(`WARNING: External module '${m}' in use.`);
-                }
-                return true;
-            }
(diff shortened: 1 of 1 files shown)


anurse:
Confirmed that webpack failed to build with the old version with:

```
ERROR in ./node_modules/@aspnet/signalr-protocol-msgpack/dist/browser/signalr-protocol-msgpack.js
Module not found: Error: Can't resolve 'commonjs-proxy:Cwdfc1cae7b2a38modulesSignalRclient-tssignalr-protocol-msgpack
ode_modulesmsgpack5index.js' in '/private/var/folders/18/xkl3z2tx7yq1wb3lllyvhn9m0000gn/T/tmp.TcVIUHxW/node_modules/@aspnet/signalr-protocol-msgpack/dist/browser'
 @ ./node_modules/@aspnet/signalr-protocol-msgpack/dist/browser/signalr-protocol-msgpack.js 6:138-270
 @ ./index.js
```

Now succeeds and I verified that the bundled script can generate msgpack messages.

anurse:
Opening a new combine PR just to cover all the things approved for the brief preview1 update :)

PR #1535, opened 2018-03-06
Update korebuild to fix osx problem

davidfowl:


anurse:
Although.... 

screen shot 2018-03-06 at 8 57 19 am

davidfowl:
Yea, so @pakrym told me that we fixed the crash during dotnet restore but now it repros during tests... I see the same on kestrel.

anurse:
Progress! I guess...

PR #1540, opened 2018-03-06
Fixes for NPM packages in preview1

anurse:
* Fix the README for `@aspnet/signalr` (was approved in #1520)
* Fix packaging of the `@aspnet/signalr-protocol-msgpack` UMD build to ensure the resulting bundled file references `msgpack5` properly. (not yet approved, originated in #1533)

The quicker this is approved, the sooner I can release the NPM packages ;).


client-ts/signalr/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aspnet/signalr",
-  "version": "1.0.0-preview1-t000",
+  "version": "1.0.0-preview1-update1",
```
anurse: I'm going to undo these changes and just do them locally when building. The CI will be replacing the version with a build number anyway.

anurse:
Yeah, I figured most of the JavaScript stuff would be beyond most people's experience ;)

client-ts/typings/EventSource.d.ts
```diff
@@ -0,0 +1,31 @@
...
+
+interface EventSourceOptions {
+    withcredentials: boolean
+}
+
+declare class EventSource extends EventTarget {
+    constructor(url: string);
+    constructor(url: string, configuration: EventSourceOptions);
+
+    readonly CLOSED: number;
```
JamesNK: Move constants to an enum? https://www.typescriptlang.org/docs/handbook/enums.html

anurse: This is a declaration file for an existing browser API, not our own code :). It's actually not needed in `dev` because we updated TypeScript to a version that includes this declaration (it's the client API for Server-Sent Events in the browser)

anurse:
We fixed most of these flaky tests in `dev`, but the fixes aren't here :(. This change doesn't affect any of the tests that are failing, it's JS only. So I'm rage-merging.

Issue #1534, opened 2018-03-06
method with return type Task<IObservable<T>> is not recogniced as Streaming method


JanEggers:
im having a gateway that analyses the request and based on the parameters forwards the request to some backend service.

im using the signalr client in my gateway to achive that proxy behavior. what i dont like is that i have to 

`hubConnection.StreamAsync<TResult>("Method").Result;`

because if i await the call and change my hubmethod to return a task then signalr does not recognice it as streaming method.

BrennanConroy:
Just return `IObservable<T>` from your hub

JanEggers:
how can i do that? if i use your signalr client to subscribe to a stream it is always async!

https://github.com/aspnet/SignalR/blob/00da533f109a0f82cf2d9d8af26d0f39666bf0a4/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L194

davidfowl:
@JamesNK can you take a look?  

@JanEggers You're invoking StreamAsync from the client right?

JanEggers:
@davidfowl i have a js client that uses this:

https://github.com/aspnet/SignalR/blob/846432c9ac4599cbd24ceac6e756627d3f1859c7/client-ts/signalr/src/HubConnection.ts#L163

that connects to a gateway server with a hub. that gateway hub uses 

https://github.com/aspnet/SignalR/blob/00da533f109a0f82cf2d9d8af26d0f39666bf0a4/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L194

to forward the request to the backend.

and that backend hub returns an `IObservable<T>`

JanEggers:
I used signalr client in the server because i have to call different backends depending on parameters otherwise i would have used some websocket proxy

davidfowl:
Great. We have a bug here that we need to fix. Does it throw when you have a `Task<IObservable<T>>` or `Task<ChannelReader<T>>` saying that the invocation isn't streaming?

As a workaround in the meantime you could do something like this (if using RX):

```
public IObservable<int> Proxy()
{
     var subject = new Subject<int>();
     _ =StreamAsync(hubConnection, subject);
     return subject;
}

private async Task StreamAsync(HubConnection conn, Subject<int> subject)
{
    var reader = await conn.StreamAsync("SomeBackendMethod");
    while (reader.WaitToReadAsync())
    {
        while(reader.TryRead(out var item))
        {  
              subject.OnNext(item);
        }
[...]

PR #1544, opened 2018-03-06
Fix streaming hub methods combined with async

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/1534

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
@@ -372,6 +372,9 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
         private static bool IsStreamed(Type resultType)
         {
+            // TODO: cache reflection for performance, on HubMethodDescriptor maybe?
+            resultType = UnwrapTask(resultType);
+
             var observableInterface = IsIObservable(resultType) ?
                 resultType :
                 resultType.GetInterfaces().FirstOrDefault(IsIObservable);
@@ -393,6 +396,9 @@ namespace Microsoft.AspNetCore.SignalR.Internal
         {
             if (result != null)
             {
+                // TODO: cache reflection for performance, on HubMethodDescriptor maybe?
+                resultType = UnwrapTask(resultType);
+
                 var observableInterface = IsIObservable(resultType) ?
                     resultType :
                     resultType.GetInterfaces().FirstOrDefault(IsIObservable);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -393,6 +396,9 @@ private bool TryGetStreamingEnumerator(HubConnectionContext connection, string i
         {
             if (result != null)
             {
+                // TODO: cache reflection for performance, on HubMethodDescriptor maybe?
+                resultType = UnwrapTask(resultType);
```
anurse: I spent a few minutes wondering if it was really this simple and thinking you forgot to `await` the result to get the `IObservable` before realizing that you **do** already have the unwrapped `IObservable` in `result`, it's just that `resultType` is the defined return type obtained via reflection rather than the actual type of `result`. That makes me wonder if we should just use `result.GetType()` for all the detection logic instead of `resultType`.

JamesNK: What should happen if there is a hub method that is typed to return `object`, and at runtime logic sometimes returns `List<string>` and other times returns `IObservable<string>`. Should it be streamable depending on what is returned?

By basing logic on the methods type definition you are able to cache exactly what you expect.

JamesNK: > Should it be streamable depending on what is returned?

Never mind, `ValidateInvocationMode` already enforces based on the method's type definition.

IMO for perf type info should be cached for `ValidateInvocationMode` and we might as well reuse the cache type info when resolving the streaming enumerator.

anurse: > What should happen if there is a hub method that is typed to return object, and at runtime logic sometimes returns List<string> and other times returns IObservable<string>

Yeah, that was my concern too. I suppose this is still a better way to do this.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTestUtils/Hubs.cs
```diff
@@ -529,6 +534,11 @@ public ChannelReader<string> CounterChannel(int count)
             return channel.Reader;
         }
 
+        public Task<ChannelReader<string>> CounterChannelAsync(int count)
+        {
```
davidfowl: Make one of these truly async just so we have coverage for it.

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -372,6 +372,9 @@ private static bool IsChannel(Type type, out Type payloadType)
 
         private static bool IsStreamed(Type resultType)
         {
+            // TODO: cache reflection for performance, on HubMethodDescriptor maybe?
```
davidfowl: File issues for these now. Maybe we can cache at discovery time.

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
davidfowl: Yea this is fine. I think mvc uses the runtime type but this is a bit cleaner

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
anurse: He already filed #1546 before submitting the PR :)

Issue #763, opened 2017-08-25
Version in the user-agent header should be generated automatically


moozzyk:
We should use https://github.com/aspnet/SignalR/blob/7b0be52c545bc20f7402ed1b58c5ff8f99c2030f/version.xml#L6 to generate the User-Agent header. Currently the version is hardcoded - it is fragile and will very likely be missed when we move to a newer version/version suffix.

(We already use this to generate version for the npm module for the client: https://github.com/aspnet/SignalR/blob/cdbb13de66b068015d068e5fab67e4e6ab42d3da/client-ts/Microsoft.AspNetCore.SignalR.Client.TS/Microsoft.AspNetCore.SignalR.Client.TS.csproj#L42)

anurse:
All we need to do here is make sure the AssemblyInformationalVersion ends up in the User Agent from the client.

PR #1551, opened 2018-03-07
Client sends user agent with version based on assembly version

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/763

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Http.Headers;
 using System.Reflection;
@@ -19,6 +20,8 @@ namespace Microsoft.AspNetCore.Sockets.Client.Http
                 .Assembly
                 .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
                 .FirstOrDefault();
+
+            Debug.Assert(assemblyVersion != null);
 
             // assembly version attribute should always be present
             // but in case it isn't then don't include version in user-agent
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
@@ -1,13 +1,25 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
 using System.Net.Http.Headers;
+using System.Reflection;
 
 namespace Microsoft.AspNetCore.Sockets.Client.Http
 {
-    public class Constants
+    public static class Constants
     {
-        private static readonly string UserAgent = "Microsoft.AspNetCore.Sockets.Client.Http/1.0.0-alpha";
-        public static readonly ProductInfoHeaderValue UserAgentHeader = ProductInfoHeaderValue.Parse(UserAgent);
+        public static readonly ProductInfoHeaderValue UserAgentHeader;
+
+        static Constants()
+        {
+            var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+                .Assembly
+                .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
+                .Single();
+
+            var userAgent = "Microsoft.AspNetCore.Sockets.Client.Http/" + assemblyVersion.InformationalVersion;
+            UserAgentHeader = ProductInfoHeaderValue.Parse(userAgent);
+        }
     }
 }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
@@ -16,9 +16,9 @@ namespace Microsoft.AspNetCore.Sockets.Client.Http
         {
             var userAgent = "Microsoft.AspNetCore.Sockets.Client.Http";
 
-            var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+            var assemblyVersion = typeof(Constants)
                 .Assembly
-                .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
+                .GetCustomAttributes<AssemblyInformationalVersionAttribute>()
                 .FirstOrDefault();
 
             Debug.Assert(assemblyVersion != null);
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
@@ -13,12 +13,20 @@ namespace Microsoft.AspNetCore.Sockets.Client.Http
 
         static Constants()
         {
+            var userAgent = "Microsoft.AspNetCore.Sockets.Client.Http";
+
             var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
                 .Assembly
                 .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
-                .Single();
+                .FirstOrDefault();
 
-            var userAgent = "Microsoft.AspNetCore.Sockets.Client.Http/" + assemblyVersion.InformationalVersion;
+            // assembly version attribute should always be present
+            // but in case it isn't then don't include version in user-agent
+            if (assemblyVersion != null)
+            {
+                userAgent += "/" + assemblyVersion.InformationalVersion;
+            }
+
             UserAgentHeader = ProductInfoHeaderValue.Parse(userAgent);
         }
     }
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
@@ -4,8 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.IO.Pipelines;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Reflection;
 using System.Text;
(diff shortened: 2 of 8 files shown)


JamesNK:
With this change the user-agent is now `Microsoft.AspNetCore.Sockets.Client.Http/1.0.0-preview2-t000` (until assembly version is modified)

src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
```diff
@@ -1,13 +1,25 @@
...
-        private static readonly string UserAgent = "Microsoft.AspNetCore.Sockets.Client.Http/1.0.0-alpha";
-        public static readonly ProductInfoHeaderValue UserAgentHeader = ProductInfoHeaderValue.Parse(UserAgent);
+        public static readonly ProductInfoHeaderValue UserAgentHeader;
+
+        static Constants()
+        {
+            var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+                .Assembly
+                .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
+                .Single();
```
davidfowl: SingleOrDefault? I guess this should never fail but this sorta thing makes me paranoid

JamesNK: I'm not sure of a world where that could happen. Plus the unit tests are doing the same check.

What should the version be if the attribute couldn't be found?

JamesNK: `Microsoft.AspNetCore.Sockets.Client.Http/unknown`?

JamesNK: https://tools.ietf.org/html/rfc7231#section-5.5.3

Spec says version in user-agent is optional. I'll not include it if for some crazy reason the attribute can't be found.

davidfowl: Add a Debug.Assert that it's never null. That should be fine.

src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
```diff
@@ -1,13 +1,33 @@
...
+            var userAgent = "Microsoft.AspNetCore.Sockets.Client.Http";
+
+            var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+                .Assembly
+                .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
+                .FirstOrDefault();
+
+            // assembly version attribute should always be present
+            // but in case it isn't then don't include version in user-agent
+            if (assemblyVersion != null)
```
davidfowl: Add a Debug.Assert

davidfowl:
Are we using this header for websockets as well?

JamesNK:
User agent is only used here for HTTP requests

https://github.com/aspnet/SignalR/blob/1c44e8febfd57a1038715d6c74f030cebdcd14f2/src/Microsoft.AspNetCore.Sockets.Client.Http/SendUtils.cs#L84-L99

davidfowl:
Err, we should fix that. I think we should set it as a default header in our options. I believe those flow to all transports.

JamesNK:
IMO it is kind of weird that the web socket transport uses `HttpOptions`

https://github.com/aspnet/SignalR/blob/1c44e8febfd57a1038715d6c74f030cebdcd14f2/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs#L33

davidfowl:
> IMO it is kind of weird that the web socket transport uses HttpOptions

Why? WebSockets is an Http protocol 😄

JamesNK:
https://en.wikipedia.org/wiki/WebSocket

> WebSocket is a different TCP protocol from HTTP. Both protocols are located at layer 7 in the OSI model and, as such, depend on TCP at layer 4. Although they are different, RFC 6455 states that WebSocket "is designed to work over HTTP ports 80 and 443 as well as to support HTTP proxies and intermediaries" thus making it compatible with the HTTP protocol. To achieve compatibility, the WebSocket handshake uses the HTTP Upgrade header[1] to change from the HTTP protocol to the WebSocket protocol.

Ah, it is designed to be compatible. I really don't know anything about web sockets 😇

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
```diff
@@ -45,6 +50,36 @@ public WebSocketsTransportTests(ServerFixture<Startup> serverFixture, ITestOutpu
...
+                    TransferMode.Binary, connection: Mock.Of<IConnection>()).OrTimeout();
+
+                await pair.Transport.Output.WriteAsync(Encoding.UTF8.GetBytes("User-Agent"));
+
+                // The HTTP header endpoint closes the connection immediately after sending response which should stop the transport
+                await webSocketsTransport.Running.OrTimeout();
+
+                Assert.True(pair.Transport.Input.TryRead(out var result));
+
+                string userAgent = Encoding.UTF8.GetString(result.Buffer.ToArray());
```
JamesNK: I couldn't see a hook to test the user agent on the client with websockets so I made an endpoint that sends a requested header back from the server where you have access to the `HttpContext` 🤷‍♂️

anurse: We've done that in other places, it's a reasonable workaround :)

src/Microsoft.AspNetCore.Sockets.Client.Http/Constants.cs
```diff
@@ -1,13 +1,36 @@
...
-        public static readonly ProductInfoHeaderValue UserAgentHeader = ProductInfoHeaderValue.Parse(UserAgent);
+        public static readonly ProductInfoHeaderValue UserAgentHeader;
+
+        static Constants()
+        {
+            var userAgent = "Microsoft.AspNetCore.Sockets.Client.Http";
+
+            var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+                .Assembly
+                .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
```
anurse: [`GetCustomAttributes<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.customattributeextensions.getcustomattributes?view=netstandard-2.0#System_Reflection_CustomAttributeExtensions_GetCustomAttributes__1_System_Reflection_Assembly_) ?

test/Microsoft.AspNetCore.SignalR.Client.Tests/ServerSentEventsTransportTests.cs
```diff
@@ -299,6 +302,43 @@ public class ServerSentEventsTransportTests
...
+            }
+
+            Assert.NotNull(userAgentHeaderCollection);
+            var userAgentHeader = Assert.Single(userAgentHeaderCollection);
+            Assert.Equal("Microsoft.AspNetCore.Sockets.Client.Http", userAgentHeader.Product.Name);
+
+            // user agent version should come from version embedded in assembly metadata
+            var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+                .Assembly
+                .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
```
anurse: Can use `GetCustomAttribute<T>` here rather than casting and using `.Single()`

test/Microsoft.AspNetCore.SignalR.Client.Tests/LongPollingTransportTests.cs
```diff
@@ -388,6 +392,51 @@ public class LongPollingTransportTests
...
+            }
+
+            Assert.NotNull(userAgentHeaderCollection);
+            var userAgentHeader = Assert.Single(userAgentHeaderCollection);
+            Assert.Equal("Microsoft.AspNetCore.Sockets.Client.Http", userAgentHeader.Product.Name);
+
+            // user agent version should come from version embedded in assembly metadata
+            var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+                .Assembly
+                .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
```
anurse: Can use `GetCustomAttribute<T>` here rather than casting and using `.Single()`

test/Microsoft.AspNetCore.SignalR.Tests/EchoEndPoint.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.SignalR.Tests
 {
     public class EchoEndPoint : EndPoint
     {
-        public async override Task OnConnectedAsync(ConnectionContext connection)
+        public override async Task OnConnectedAsync(ConnectionContext connection)
```
anurse: Resharper or something? It's fine (better, even), just curious :)

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
```diff
@@ -45,6 +50,36 @@ public WebSocketsTransportTests(ServerFixture<Startup> serverFixture, ITestOutpu
...
+
+                // user agent version should come from version embedded in assembly metadata
+                var assemblyVersion = (AssemblyInformationalVersionAttribute)typeof(Constants)
+                    .Assembly
+                    .GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute))
```
anurse: `GetCustomAttribute` again ;)

JamesNK:
So it turns out web sockets in .NET don't like having a user agent set: https://github.com/dotnet/corefx/issues/26627. For some reason this error doesn't show up in the unit tests but does show up in the functional tests. Are some of the functional tests .NET Framework instead of core perhaps?

test/Microsoft.AspNetCore.SignalR.Tests/EchoEndPoint.cs
JamesNK: Yes. Do most MS devs not use it? I see a lot of RS suggestions throughout signalr code.

anurse: Plenty of folks at MS use it (EF team uses it quite a bit). No one on the SignalR team uses it that I know of, but there's no problem with it. We can talk about trying to resolve some of the suggestions. It's certainly OK to throw these small fixes in as you're modifying a file.

PR #1557, opened 2018-03-08
Add additional properties to HttpOptions

JamesNK:


JamesNK:
Application logic is done, unit tests to come. Review that what is there is ok, I'll add tests tomorrow.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpOptions.cs
```diff
@@ -3,17 +3,24 @@
...
     public class HttpOptions
     {
-        public HttpMessageHandler HttpMessageHandler { get; set; }
+        public Func<HttpMessageHandler, HttpMessageHandler> HttpMessageHandler { get; set; }
         public IReadOnlyCollection<KeyValuePair<string, string>> Headers { get; set; }
         public Func<string> AccessTokenFactory { get; set; }
         public TimeSpan CloseTimeout { get; set; } = TimeSpan.FromSeconds(5);
+        public ICredentials Credentials { get; set; }
+        public X509CertificateCollection ClientCertificates { get; set; }
+        public CookieContainer Cookies { get; set; }
```
JamesNK: `CookieContainer` or `IList<Cookie>`? Likewise `X509CertificateCollection` or `IList<X509Certificate>`?

JamesNK: CookieContainer isn't a traditional collection data structure - https://msdn.microsoft.com/en-us/library/system.net.cookiecontainer%28v=vs.110%29.aspx

anurse: I think we should use the same type as ClientWebSocketOptions and HttpClientHandler, even though they are concrete. Though we may want to pre-initialize them, which is what [`ClientWebSocketOptions` does](https://source.dot.net/#System.Net.WebSockets.Client/System/Net/WebSockets/ClientWebSocketOptions.cs,91) for example.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -62,7 +74,7 @@ public static IHubConnectionBuilder WithTransport(this IHubConnectionBuilder hub
             return hubConnectionBuilder;
         }
 
-        public static IHubConnectionBuilder WithMessageHandler(this IHubConnectionBuilder hubConnectionBuilder, HttpMessageHandler httpMessageHandler)
+        public static IHubConnectionBuilder WithMessageHandler(this IHubConnectionBuilder hubConnectionBuilder, Func<HttpMessageHandler, HttpMessageHandler> httpMessageHandler)
```
anurse: Was chatting with @davidfowl and @Tratcher about this. It sounds like we have some precedent for the model where we give you the individual properties (`Cookies`, `Headers`, etc.) **and** a property to replace the `HttpMessageHandler` itself and if you set the latter, we **ignore** the individual properties (it's the "take my life into my own hands" option). I think we should stick with that for now.

The tricky thing here is that when we people do replace the `HttpMessageHandler`, they basically have to replace the `ClientWebSocketOptions` as well.

The error messages seem a little unpleasant for this too, since the user is providing a delegate and could return null in the delegate (so we can't just have an ArgumentNullException here).

anurse: Whichever we choose, this could benefit from ye-olde-`///` comments describing the behavior a little bit.

JamesNK: `HttpMessageHandler`s compose. A func is useful because devs don't necessarily want to completely replace the handler, they just want to wrap the inner `HttpMessageHandler` - which has been initialized with the other options on `HttpOptions` - with a handler that logs or something and then calls the inner handler.

I should also add that the func turned out to be useful for unit testing the new settings https://github.com/aspnet/SignalR/pull/1557/commits/3e8604670a4a3ea12ab12ddf4735ad78008a468b#diff-b5ac650137bfa6943ef9d5d0cfb05460R169

JamesNK: Docs added

anurse: Fair enough, let's go with it. We've got some time to get feedback anyway :)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -86,23 +86,60 @@ public HttpConnection(Uri url, TransportType transportType, ILoggerFactory logge
...
+                {
+                    httpClientHandler.UseDefaultCredentials = _httpOptions.UseDefaultCredentials.Value;
+                }
+                httpMessageHandler = httpClientHandler;
+                if (_httpOptions.HttpMessageHandler != null)
+                {
+                    httpMessageHandler = _httpOptions.HttpMessageHandler(httpClientHandler);
+                    if (httpMessageHandler == null)
+                    {
+                        throw new InvalidOperationException("Configured HttpMessageHandler did not return a value.");
```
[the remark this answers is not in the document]
davidfowl: I don't think we should throw here. Just ignore the return value.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpOptions.cs
```diff
@@ -3,17 +3,29 @@
...
 namespace Microsoft.AspNetCore.Sockets.Client.Http
 {
     public class HttpOptions
     {
-        public HttpMessageHandler HttpMessageHandler { get; set; }
+        /// <summary>
+        /// Gets or sets a delegate for wrapping or replacing the <see cref="HttpMessageHandler"/>
+        /// that will make HTTP requests the server.
+        /// </summary>
+        public Func<HttpMessageHandler, HttpMessageHandler> HttpMessageHandler { get; set; }
```
davidfowl: HttpMessageHandlerCallback? HttpMessageHandlerFactory? It's not a message handler so lets call it something else.

anurse: Yeah, I got nervous about `HttpMessageHandlerFactory` because it's getting close to an `HttpMessageHandlerFactoryStrategyObserverBuilder` but I do think it needs to be a `Factory`.

Tratcher: ConfigureMessageHandler

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketsTransportTests.cs
```diff
@@ -35,6 +40,33 @@ public WebSocketsTransportTests(ServerFixture<Startup> serverFixture, ITestOutpu
             _serverFixture = serverFixture;
         }
 
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win2008R2, SkipReason = "No WebSockets Client for this platform")]
```
anurse: I'm going to file a bug for this (not worth doing in this PR) but we're going to need a new skip attribute for WebSockets because we actually **do** want to run on Windows 7 but only with `netcoreapp2.1`. They switched `ClientWebSocket` to use the Managed implementation on all OSes in .NET Core 2.1 which means it now supports Windows 7.

anurse: Filed https://github.com/aspnet/SignalR/issues/1568 to track.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpOptions.cs
anurse: Ooh, we use the `Configure` prefix for `Action<SomeOptionsType>` delegates, I do like it here.

anurse: I like it most in the extension method name on `HubConnectionBuilder`, not sure how I feel about it on the `HttpOptions` property name. Factory might still make sense here, but we could use `.ConfigureHttpMessageHandler` as the extension method (which is what most users interact with)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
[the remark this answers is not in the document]
JamesNK: IMO if a dev doesn't want to use this func then they won't set or set it to null. And if they want to use the original handler then they can return it. Returning null would be a dev mistake which is why it should throw.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpOptions.cs
JamesNK: I just called it `HttpMessageHandler` on `HttpOptions` because there is just a 'WebSocketOptions' action on the same class. Should they both be prefixed with Configure?

Also it is just `WithWebSocketOptions` on the `HubConnectionBuilderHttpExtensions`, although the parameter is called `configureWebSocketOptions`.

anurse: Let's make them both `Configure`: `ConfigureHttpMessageHandler` and `ConfigureWebSocketOptions`.

PR #1561, opened 2018-03-08
Update to new corefx

pakrym:


src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -143,7 +144,7 @@ public WebSocketsTransport(WebSocketOptions options, IDuplexPipe application, De
 #if NETCOREAPP2_1
                     var receiveResult = await socket.ReceiveAsync(memory, CancellationToken.None);
 #else
-                    var isArray = memory.TryGetArray(out var arraySegment);
+                    var isArray = MemoryMarshal.TryGetArray<byte>(memory, out var arraySegment);
```
anurse: Is the `<byte>` not inferrable?

pakrym: The first parameter type is `ReadOnlyMemory`, it's either inference or implicit conversion, unfortunately.

pakrym:
I hope so, it contains the fix for the issue OSX we reported. We'll see if it works :)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -367,7 +367,7 @@ private static Uri CreateConnectUrl(Uri url, string connectionId)
         private async Task StartTransport(Uri connectUrl, TransportType transportType)
         {
 
-            var options = new PipeOptions(readerScheduler: PipeScheduler.ThreadPool);
+            var options = new PipeOptions(writerScheduler:PipeScheduler.Inline, readerScheduler: PipeScheduler.ThreadPool, useSynchronizationContext: false);
```
BrennanConroy: nit: space after `:`

src/Common/PipeWriterStream.cs
```diff
@@ -63,11 +63,11 @@ public override Task WriteAsync(byte[] buffer, int offset, int count, Cancellati
         }
 
 #if NETCOREAPP2_1
-        public override Task WriteAsync(ReadOnlyMemory<byte> source, CancellationToken cancellationToken = default)
+        public override ValueTask WriteAsync(ReadOnlyMemory<byte> source, CancellationToken cancellationToken = default)
         {
             _pipeWriter.Write(source.Span);
             _length += source.Length;
-            return Task.CompletedTask;
+            return new ValueTask(Task.CompletedTask);
```
davidfowl: return default;

src/Common/WebSocketExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static Task SendAsync(this WebSocket webSocket, ReadOnlySequence<byte> bu
 #if NETCOREAPP2_1
             if (buffer.IsSingleSegment)
             {
-                return webSocket.SendAsync(buffer.First, webSocketMessageType, endOfMessage: true, cancellationToken);
+                return webSocket.SendAsync(buffer.First, webSocketMessageType, endOfMessage: true, cancellationToken).AsTask();
```
davidfowl: Make this method return ValueTask.

davidfowl:
Super small set of changes.

PR #1566, opened 2018-03-09
Detect Successful Start in LongPolling and SSE

mikaelm12:
So this is hacky and doesn't work yet but I wanted to get some feedback.
Currently storing a first request flag in the HttpContext which is probably 👎 

Basically wanted some help with the LongPolling transport code. Not setting the ContentLength or the Status code if flush has already been called.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -23,7 +23,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private IDuplexPipe _application;
         private Task _sender;
         private Task _poller;
-        private HttpResponseMessage _response;
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
 
         public Task Running { get; private set; } = Task.CompletedTask;
@@ -76,7 +75,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             try
             {
-                _response?.Dispose();
                 await Running;
             }
             catch
@@ -94,6 +92,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 {
                     var request = new HttpRequestMessage(HttpMethod.Get, pollUrl);
                     SendUtils.PrepareHttpRequest(request, _httpOptions);
+                    HttpResponseMessage _response;
 
                     try
                     {
@@ -122,7 +121,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         Log.ReceivedMessages(_logger);
 
                         var stream = new PipeWriterStream(_application.Output);
-                        await _response.Content.CopyToAsync(stream);
+                        using (cancellationToken.Register(response => { ((HttpResponseMessage)response).Dispose(); }, _response))
+                        {
+                            await _response.Content.CopyToAsync(stream);
+                        }
                         await _application.Output.FlushAsync();
                     }
                 }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -52,7 +52,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             Log.StartTransport(_logger, transferFormat);
 
             // Start sending and polling (ask for binary if the server supports it)
-            var startTcs = new TaskCompletionSource<object>();
+            var startTcs = new TaskCompletionSource<object>(TaskContinuationOptions.RunContinuationsAsynchronously);
             _poller = Poll(url, startTcs, _transportCts.Token);
             _sender = SendUtils.SendMessages(url, _application, _httpClient, _httpOptions, _transportCts, _logger);
 
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -23,6 +23,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private IDuplexPipe _application;
         private Task _sender;
         private Task _poller;
+        private HttpResponseMessage _response;
 
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
 
@@ -53,7 +54,8 @@ namespace Microsoft.AspNetCore.Sockets.Client
             Log.StartTransport(_logger, transferFormat);
 
             // Start sending and polling (ask for binary if the server supports it)
-            _poller = Poll(url, _transportCts.Token);
+            var startTcs = new TaskCompletionSource<object>();
+            _poller = Poll(url, startTcs, _transportCts.Token);
             _sender = SendUtils.SendMessages(url, _application, _httpClient, _httpOptions, _transportCts, _logger);
 
             Running = Task.WhenAll(_sender, _poller).ContinueWith(t =>
@@ -64,7 +66,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 return t;
             }).Unwrap();
 
-            return Task.CompletedTask;
+            return startTcs.Task;
         }
 
         public async Task StopAsync()
@@ -75,6 +77,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             try
             {
+                _response?.Dispose();
                 await Running;
             }
             catch
@@ -83,7 +86,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
         }
 
-        private async Task Poll(Uri pollUrl, CancellationToken cancellationToken)
+        private async Task Poll(Uri pollUrl, TaskCompletionSource<object> startTcs,CancellationToken cancellationToken)
         {
             Log.StartReceive(_logger);
             try
@@ -93,11 +96,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
                     var request = new HttpRequestMessage(HttpMethod.Get, pollUrl);
                     SendUtils.PrepareHttpRequest(request, _httpOptions);
 
-                    HttpResponseMessage response;
-
                     try
                     {
-                        response = await _httpClient.SendAsync(request, cancellationToken);
+                        _response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
                     }
                     catch (OperationCanceledException)
                     {
@@ -107,9 +108,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         continue;
                     }
 
-                    response.EnsureSuccessStatusCode();
+                    _response.EnsureSuccessStatusCode();
+                    startTcs.TrySetResult(null);
 
-                    if (response.StatusCode == HttpStatusCode.NoContent || cancellationToken.IsCancellationRequested)
+                    if (_response.StatusCode == HttpStatusCode.NoContent || cancellationToken.IsCancellationRequested)
                     {
                         Log.ClosingConnection(_logger);
 
@@ -121,18 +123,24 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         Log.ReceivedMessages(_logger);
 
                         var stream = new PipeWriterStream(_application.Output);
-                        await response.Content.CopyToAsync(stream);
+                        try
+                        {
+                            await _response.Content.CopyToAsync(stream);
+                        }
+                        catch { /*First request*/};
                         await _application.Output.FlushAsync();
                     }
                 }
             }
-            catch (OperationCanceledException)
+            catch (OperationCanceledException ex)
             {
                 // transport is being closed
+                startTcs.TrySetException(ex);
                 Log.ReceiveCanceled(_logger);
             }
             catch (Exception ex)
             {
+                startTcs.TrySetException(ex);
                 Log.ErrorPolling(_logger, pollUrl, ex);
                 throw;
             }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -24,7 +24,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
         private Task _sender;
         private Task _poller;
         private HttpResponseMessage _response;
-
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
 
         public Task Running { get; private set; } = Task.CompletedTask;
@@ -123,11 +122,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         Log.ReceivedMessages(_logger);
 
                         var stream = new PipeWriterStream(_application.Output);
-                        try
-                        {
-                            await _response.Content.CopyToAsync(stream);
-                        }
-                        catch { /*First request*/};
+                        await _response.Content.CopyToAsync(stream);
                         await _application.Output.FlushAsync();
                     }
                 }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -86,7 +86,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
             }
         }
 
-        private async Task Poll(Uri pollUrl, TaskCompletionSource<object> startTcs,CancellationToken cancellationToken)
+        private async Task Poll(Uri pollUrl, TaskCompletionSource<object> startTcs, CancellationToken cancellationToken)
         {
             Log.StartReceive(_logger);
             try
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
@@ -92,11 +92,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 {
                     var request = new HttpRequestMessage(HttpMethod.Get, pollUrl);
                     SendUtils.PrepareHttpRequest(request, _httpOptions);
-                    HttpResponseMessage _response;
+                    HttpResponseMessage response;
 
                     try
                     {
-                        _response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+                        response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
                     }
                     catch (OperationCanceledException)
                     {
@@ -106,10 +106,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         continue;
                     }
 
-                    _response.EnsureSuccessStatusCode();
+                    response.EnsureSuccessStatusCode();
                     startTcs.TrySetResult(null);
 
-                    if (_response.StatusCode == HttpStatusCode.NoContent || cancellationToken.IsCancellationRequested)
+                    if (response.StatusCode == HttpStatusCode.NoContent || cancellationToken.IsCancellationRequested)
                     {
                         Log.ClosingConnection(_logger);
 
@@ -121,9 +121,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
                         Log.ReceivedMessages(_logger);
 
                         var stream = new PipeWriterStream(_application.Output);
-                        using (cancellationToken.Register(response => { ((HttpResponseMessage)response).Dispose(); }, _response))
+                        using (cancellationToken.Register(res => { ((HttpResponseMessage)res).Dispose(); }, response))
                         {
-                            await _response.Content.CopyToAsync(stream);
+                            await response.Content.CopyToAsync(stream);
                         }
                         await _application.Output.FlushAsync();
                     }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -50,12 +50,13 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 throw new ArgumentException($"The '{transferFormat}' transfer format is not supported by this transport.", nameof(transferFormat));
             }
 
+            var startTcs = new TaskCompletionSource<object>();
             _application = application;
 
             Log.StartTransport(_logger, transferFormat);
 
             var sendTask = SendUtils.SendMessages(url, _application, _httpClient, _httpOptions, _transportCts, _logger);
-            var receiveTask = OpenConnection(_application, url, _transportCts.Token);
+            var receiveTask = OpenConnection(_application, url, startTcs, _transportCts.Token);
 
             Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
             {
@@ -66,10 +67,10 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 return t;
             }).Unwrap();
 
-            return Task.CompletedTask;
+            return startTcs.Task;
         }
 
-        private async Task OpenConnection(IDuplexPipe application, Uri url, CancellationToken cancellationToken)
+        private async Task OpenConnection(IDuplexPipe application, Uri url, TaskCompletionSource<object> tcs, CancellationToken cancellationToken)
         {
             Log.StartReceive(_logger);
 
@@ -77,6 +78,18 @@ namespace Microsoft.AspNetCore.Sockets.Client
             SendUtils.PrepareHttpRequest(request, _httpOptions);
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
             var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+
+            try
+            {
+                response.EnsureSuccessStatusCode();
+                tcs.TrySetResult(null);
+            }
+            catch (Exception ex)
+            {
+                Log.TransportStopping(_logger);
+                tcs.TrySetException(ex);
+                return;
+            }
 
             using (var stream = await response.Content.ReadAsStreamAsync())
             {
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -77,10 +77,11 @@ namespace Microsoft.AspNetCore.Sockets.Client
             var request = new HttpRequestMessage(HttpMethod.Get, url);
             SendUtils.PrepareHttpRequest(request, _httpOptions);
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
-            var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
 
+            HttpResponseMessage response;
             try
             {
+                response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
                 response.EnsureSuccessStatusCode();
                 tcs.TrySetResult(null);
             }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -50,7 +50,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 throw new ArgumentException($"The '{transferFormat}' transfer format is not supported by this transport.", nameof(transferFormat));
             }
 
-            var startTcs = new TaskCompletionSource<object>();
+            var startTcs = new TaskCompletionSource<object>(TaskContinuationOptions.RunContinuationsAsynchronously);
             _application = application;
 
             Log.StartTransport(_logger, transferFormat);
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -30,10 +30,18 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
         {
             try
             {
+                var headersFlushed = false;
+                if (context.Items.TryGetValue("FirstRequest", out var firstRequest))
```
davidfowl: Why can't you just pass it into the constructor? The transport itself isn't long lived.

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -68,17 +69,24 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferMode requestedT
...
-        private async Task OpenConnection(IDuplexPipe application, Uri url, CancellationToken cancellationToken)
+        private async Task OpenConnection(IDuplexPipe application, Uri url, TaskCompletionSource<object> tcs, CancellationToken cancellationToken)
         {
             Log.StartReceive(_logger);
 
             var request = new HttpRequestMessage(HttpMethod.Get, url);
             SendUtils.PrepareHttpRequest(request, _httpOptions);
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
             var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+            if (!response.IsSuccessStatusCode)
```
davidfowl: Why not use `response.EnsureSuccessStatusCode()` like the long polling transport does?

anurse: You want the exception that is thrown by `EnsureSuccessStatusCode` to manifest on the TCS.

Having said that, I actually think `EnsureSuccessStatusCode` coupled with a `try`-`catch` that puts the exception into the TCS (like in Long Polling) would be good as it would capture the stack properly.

anurse: Also, we need the `try`-`catch` anyway since not all errors will manifest as a failed response, some will manifest as exceptions thrown by `SendAsync` (socket errors such as not being able to reach the server will manifest that way)

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
anurse: Yeah, that works. We were dealing with the restriction of the `IHttpTransport` interface, but you're right, the ctor is invoked for each incoming poll.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -30,10 +30,18 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
...
+                    if ((bool)firstRequest) {
+                        await context.Response.Body.FlushAsync();
+                        headersFlushed = true;
+                    }
+                }
                 var result = await _application.ReadAsync(token);
                 var buffer = result.Buffer;
 
-                if (buffer.IsEmpty && result.IsCompleted)
+                if (buffer.IsEmpty && result.IsCompleted && !headersFlushed)
```
anurse: Put a comment here saying that if we get an empty&completed result in the first request, we'll have to send a 200 because we've already flushed the headers, but that's OK since the client will poll again and immediately get the 204.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -192,6 +192,11 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
                             Log.PollCanceled(_logger, connection.ConnectionId, existing.TraceIdentifier);
                         }
                     }
+                    else
+                    {
+                        // Hack to indicate that we're handling the first poll.
+                        context.Items.Add("FirstRequest", true);
```
JamesNK: I can't comment on whether this is a good idea but `"FirstRequest"` should be a const field somewhere

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
JamesNK: Use const

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -499,6 +499,7 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
         {
             using (StartLog(out var loggerFactory, LogLevel.Debug))
             {
+                //while (!System.Diagnostics.Debugger.IsAttached) { }
```
JamesNK: remove

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -753,6 +757,8 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
...
```
JamesNK: remove

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -773,9 +779,18 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
                 // Write to the application
                 await connection.Application.Output.WriteAsync(buffer);
 
-                await task;
+                // The fir
```
JamesNK: remove

mikaelm12:
Still very WIP. Working through more tests

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
anurse: We're changing it to a constructor argument to LongPollingTransport (since it's created on each request), which means no more magic strings.

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -85,7 +86,7 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferMode requestedT
             }
         }
 
-        private async Task Poll(Uri pollUrl, CancellationToken cancellationToken)
+        private async Task Poll(Uri pollUrl, TaskCompletionSource<object> startTcs,CancellationToken cancellationToken)
```
anurse: nit: spacing after comma

Finally! My turn to remark on spacing :P

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -80,6 +81,18 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferMode requestedT
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
             var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
 
+            try
```
anurse: Move the `try` up to encompass the `SendAsync` too. That will throw if there's a socket error.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -17,23 +17,31 @@ public class LongPollingTransport : IHttpTransport
...
+            _isFirstRequest = isFirstRequest;
         }
 
         public async Task ProcessRequestAsync(HttpContext context, CancellationToken token)
         {
             try
             {
+                if (_isFirstRequest)
+                {
+                        await context.Response.Body.FlushAsync();
```
anurse: nit: indent

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -513,6 +514,9 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
                 var app = builder.Build();
                 await dispatcher.ExecuteAsync(context, new HttpSocketOptions(), app);
 
+                Assert.Equal(StatusCodes.Status200OK, context.Response.StatusCode);
+
+                await dispatcher.ExecuteAsync(context, new HttpSocketOptions(), app);
```
anurse: Put a comment explaining why two requests are needed

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -135,6 +137,7 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferMode requestedT
             }
             catch (Exception ex)
             {
+                startTcs.TrySetException(ex);
                 Log.ErrorPolling(_logger, pollUrl, ex);
                 throw;
```
JamesNK: Make sure a unit test hits this catch

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -80,6 +81,18 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferMode requestedT
...
+            {
+                response.EnsureSuccessStatusCode();
+                tcs.SetResult(null);
+            }
+            catch (Exception ex)
+            {
+                Log.TransportStopping(_logger);
+                tcs.TrySetException(ex);
+                return;
```
JamesNK: Make sure a unit test hits this catch

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -121,18 +123,24 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
                         Log.ReceivedMessages(_logger);
 
                         var stream = new PipeWriterStream(_application.Output);
-                        await response.Content.CopyToAsync(stream);
+                        try
+                        {
+                            await _response.Content.CopyToAsync(stream);
+                        }
+                        catch { /*First request*/};
```
anurse: Why would this throw? If it's an expected condition, can we check it before calling `CopyToAsync`?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -73,20 +92,22 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
...
                     context.Response.ContentType = "text/plain";
                     context.Response.StatusCode = StatusCodes.Status200OK;
                 }
                 else
                 {
                     // Case 3
-                    Log.LongPolling204(_logger);
-                    context.Response.ContentType = "text/plain";
-                    context.Response.StatusCode = StatusCodes.Status204NoContent;
+                    if (!_isFirstRequest)
```
anurse: You can merge this with the `else` above into an `else if`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -237,17 +237,7 @@ Task OnTransportStart()
...
-                    var longPollResult = new TaskCompletionSource<HttpResponseMessage>();
-                    httpHandler.OnLongPoll(cancellationToken =>
-                    {
-                        cancellationToken.Register(() =>
-                        {
-                            longPollResult.TrySetResult(ResponseUtils.CreateResponse(HttpStatusCode.NoContent));
-                        });
-                        return longPollResult.Task;
-                    });
+                    var httpHandler = TestHttpMessageHandler.CreateDefault(handleSend:false);
```
anurse: spacing

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.SendAsync.cs
```diff
@@ -96,16 +95,7 @@ public Task SendThrowsIfConnectionIsDisposed()
...
-
-                var longPollTcs = new TaskCompletionSource<HttpResponseMessage>(TaskCreationOptions.RunContinuationsAsynchronously);
-
-                testHttpHandler.OnLongPoll(cancellationToken =>
-                {
-                    cancellationToken.Register(() => longPollTcs.TrySetResult(null));
-
-                    return longPollTcs.Task;
-                });
+                var testHttpHandler = TestHttpMessageHandler.CreateDefault(handleSend:false);
```
anurse: Dr. Spaceman!

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestHttpMessageHandler.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
     public class TestHttpMessageHandler : HttpMessageHandler
     {
         private Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessage>> _handler;
+        public bool IsFirstRequest = true;
```
anurse: `IsFirstLongPoll` or something. There are various kinds of requests handled by this handler :)

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestHttpMessageHandler.cs
```diff
@@ -28,20 +29,28 @@ public TestHttpMessageHandler(bool autoNegotiate = true)
...
+                    // Just block until canceled
+                    var tcs = new TaskCompletionSource<object>();
+                    using (cancellationToken.Register(() => tcs.TrySetResult(null)))
+                    {
+                        await tcs.Task;
+                    }
+                    return ResponseUtils.CreateResponse(HttpStatusCode.NoContent);
                 }
-                return ResponseUtils.CreateResponse(HttpStatusCode.NoContent);
+                testHttpMessageHandler.IsFirstRequest = false;
```
anurse: Comment: "Otherwise, if this is the first request, just toggle the first request flag and send an empty response"

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -512,8 +512,13 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
                 var builder = new ConnectionBuilder(services.BuildServiceProvider());
                 builder.UseEndPoint<ImmediatelyCompleteEndPoint>();
                 var app = builder.Build();
+
+                // We need two requests because te first just establishes the transport.
```
anurse: nit: `the`

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
mikaelm12: This call to copytoasync was throwing on the first request. We should be keeping track of the first request on the client side transport as well and add that check here.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -17,38 +17,66 @@ public class LongPollingTransport : IHttpTransport
...
+                {
+                    await context.Response.Body.FlushAsync();
+                }
+
+                ReadResult result;
+                try
+                {
+                    result = await _application.ReadAsync(token);
+                }
+                catch (Exception) when (_isFirstRequest)
```
mikaelm12: We can't send a 500 on the first request since the headers have already been sent so we catch here. The next poll will throw

mikaelm12:
This is ready for review

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
anurse: Put a note in the comment below about how the next poll should throw the same exception, and the client will get the error in that poll.

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -23,7 +23,7 @@ public partial class LongPollingTransport : ITransport
         private IDuplexPipe _application;
         private Task _sender;
         private Task _poller;
-
+        private HttpResponseMessage _response;
```
davidfowl: Why is this being stored?

mikaelm12: Because we needed to dispose the response in StopAsync. This is irrelevant now though as per Andrew's suggestion above

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
davidfowl: What's the point in reading after flushing?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -17,38 +17,67 @@ public class LongPollingTransport : IHttpTransport
         private readonly ILogger _logger;
         private readonly CancellationToken _timeoutToken;
         private readonly string _connectionId;
+        private readonly bool _isFirstRequest;
```
davidfowl: This entire change should be reverted and we should just flush from the outside on first request and avoid calling into the long polling transport altogether.

mikaelm12: awesome

anurse: Wait, so just flush a blank `200` for the first request? The change as it stands still writes the first message to the first request if there is one. The only time we have to flush a blank `200` is when there's no data in the pipe. It would make things simpler but all Long Polling connections would require two requests to initialize.

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -92,12 +92,11 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
...
                     var request = new HttpRequestMessage(HttpMethod.Get, pollUrl);
                     SendUtils.PrepareHttpRequest(request, _httpOptions);
-
-                    HttpResponseMessage response;
+                    HttpResponseMessage _response;
 
                     try
                     {
-                        response = await _httpClient.SendAsync(request, cancellationToken);
+                        _response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
```
anurse: You don't declare `_response` any more :)

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
davidfowl: Ah I misunderstood this change a bit. Still I hate the complexity of the control flow here. Can we expose 2 methods instead of passing the bool?

anurse: Not sure I see how that would help, it would have to either a) Call into a single method with a bool (thus defeating the purpose) or b) Duplicate quite a bit of logic.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
anurse: As mentioned earlier: because we still try to write out the first batch of messages in this request, we just do it via chunked encoding.

mikaelm12:
Reverted the changes to `ProcessRequestAsync` and move the first request logic to `ProcessFirstRequestAsync` as per @davidfowl's request

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -58,26 +89,21 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
                     // Complete the ReadAsync by indicating we haven't processed any data yet
                     // (of course, there is no data because the buffer is empty, but we have to 
                     // call AdvanceTo after every ReadAsync even if we don't have any data)
-                    _application.AdvanceTo(buffer.Start);
+                    //_application.AdvanceTo(buffer.Start);
```
davidfowl: ? Remove dead code?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -18,23 +18,79 @@ public class LongPollingTransport : IHttpTransport
         private readonly CancellationToken _timeoutToken;
         private readonly string _connectionId;
 
-        public LongPollingTransport(CancellationToken timeoutToken, PipeReader application, string connectionId, ILoggerFactory loggerFactory)
+        public LongPollingTransport(CancellationToken timeoutToken, PipeReader application, string connectionId, ILoggerFactory loggerFactory, bool isFirstRequest = false)
```
davidfowl: Remove the  bool isFirstRequest argument.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -26,15 +26,70 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
             _logger = loggerFactory.CreateLogger<LongPollingTransport>();
         }
 
+        public async Task ProcessFirstRequestAsync(HttpContext context, CancellationToken token)
+        {
+            if (!token.IsCancellationRequested)
```
davidfowl: Why are we checking the token here?

davidfowl: If we do decide to keep it, here's a style thing.

if (token.IsCancellationRequested)
{
    return;
}

mikaelm12: I thought we needed it for one of the test edge cases but turns out I can remove it from the method completely

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -26,15 +26,70 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
...
+                try
+                {
+                    await context.Response.Body.FlushAsync();
+
+                    ReadResult result;
+                    try
+                    {
+                        result = await _application.ReadAsync(token);
+                    }
+                    catch (Exception)
```
davidfowl: Remove the  (Exception) we aren't doing anything with it. catch will suffice.

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -53,7 +52,8 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
             Log.StartTransport(_logger, transferFormat);
 
             // Start sending and polling (ask for binary if the server supports it)
-            _poller = Poll(url, _transportCts.Token);
+            var startTcs = new TaskCompletionSource<object>();
```
davidfowl: Ah almost forgot this, this should look like:

```C#
new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously)
```

We don't want to run user code on our threads.

src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -50,12 +50,13 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
                 throw new ArgumentException($"The '{transferFormat}' transfer format is not supported by this transport.", nameof(transferFormat));
             }
 
+            var startTcs = new TaskCompletionSource<object>();
```
davidfowl: Same here see https://github.com/aspnet/SignalR/pull/1566/files#r175983362

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -26,15 +26,67 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
...
                 var buffer = result.Buffer;
 
+                // IF the buffer is empty and the read result is completed on the first request
+                // we still have to send a 200 because the headers have already been flushed.
                 if (buffer.IsEmpty && result.IsCompleted)
                 {
+                    // Complete the ReadAsync by indicating we haven't processed any data yet
+                    // (of course, there is no data because the buffer is empty, but we have to 
+                    // call AdvanceTo after every ReadAsync even if we don't have any data)
+                    //_application.AdvanceTo(buffer.Start);
```
davidfowl: Why is this commented out?

mikaelm12: We needed this before when we were handling the first request in this method

mikaelm12: Actually, I think we still need this..

davidfowl: Why?

mikaelm12: Technically we don't need it because in this branch when the 204 is returned the connection would get disposed. https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs#L250-L253
I was correct before when I said that we needed it when we handled the first request in this method. If we didn't call advance an exception would be thrown. Unnecessary now

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -26,13 +26,60 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
...
+
         public async Task ProcessRequestAsync(HttpContext context, CancellationToken token)
         {
             try
             {
                 var result = await _application.ReadAsync(token);
                 var buffer = result.Buffer;
 
+                // IF the buffer is empty and the read result is completed on the first request
+                // we still have to send a 200 because the headers have already been flushed.
```
davidfowl: This comment is stale.

src/Microsoft.AspNetCore.Sockets.Client.Http/LongPollingTransport.cs
```diff
@@ -121,18 +121,23 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
                         Log.ReceivedMessages(_logger);
 
                         var stream = new PipeWriterStream(_application.Output);
-                        await response.Content.CopyToAsync(stream);
+                        using (cancellationToken.Register(res => { ((HttpResponseMessage)res).Dispose(); }, response))
```
davidfowl: nit: Remove surrounding braces

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestHttpMessageHandler.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
     public class TestHttpMessageHandler : HttpMessageHandler
     {
         private Func<HttpRequestMessage, CancellationToken, Task<HttpResponseMessage>> _handler;
+        public bool IsFirstLongPoll = true;
```
davidfowl: nit: property?

davidfowl: Why is this public?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/LongPollingTransport.cs
```diff
@@ -26,13 +26,60 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
             _logger = loggerFactory.CreateLogger<LongPollingTransport>();
         }
 
+        public async Task ProcessFirstRequestAsync(HttpContext context, CancellationToken token)
```
anurse: When we talked yesterday, the plan was to just remove all the Long Polling logic (set both Client and Server back to their behavior before this PR). Long Polling will be dumb, and the client will just assume `StartAsync` works as before (and thus the server doesn't need to do any flushing). This PR is still relevant for the Server-Sent Events changes, but the Long Polling changes can just be reverted entirely

Issue #1562, opened 2018-03-08
Add "Service Guard" to SignalR


anurse:
We need something that detects if you haven't called `.AddSignalR`. MVC does something similar, we can take code from that.

mikaelm12:
Done here: 0b816588500beee58711b779ba0583fc24bb7e01

PR #1567, opened 2018-03-09
Check for SignalR service

mikaelm12:
Adding a SignalR marker service to endure that the service has been added.
Issue: https://github.com/aspnet/SignalR/issues/1562

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.Builder
 {
@@ -10,6 +11,12 @@ namespace Microsoft.AspNetCore.Builder
     {
         public static IApplicationBuilder UseSignalR(this IApplicationBuilder app, Action<HubRouteBuilder> configure)
         {
+            var marker = app.ApplicationServices.GetService(typeof(SignalRMarkerService));
+            if (marker == null)
+            {
+                throw new InvalidOperationException("The SignalR service needs to be registered");
+            }
+
             app.UseSockets(routes =>
             {
                 configure(new HubRouteBuilder(routes));
--- a/src/Microsoft.AspNetCore.SignalR/SignalRMarkerService.cs
+++ b/src/Microsoft.AspNetCore.SignalR/SignalRMarkerService.cs
@@ -1,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    class SignalRMarkerService
+    {
+    }
+}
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.SignalR/SignalRMarkerService.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    class SignalRMarkerService
```
davidfowl: Mark internal

src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
```diff
@@ -3,13 +3,20 @@
...
 namespace Microsoft.AspNetCore.Builder
 {
     public static class SignalRAppBuilderExtensions
     {
         public static IApplicationBuilder UseSignalR(this IApplicationBuilder app, Action<HubRouteBuilder> configure)
         {
+            var marker = app.ApplicationServices.GetService(typeof(SignalRMarkerService));
+            if (marker == null)
+            {
+                throw new InvalidOperationException("The SignalR service needs to be registered");
```
davidfowl: Say something about adding SignalR? Doesn’t mvc do that?

mikaelm12: Yeah, I'll make this more similar to what they have

test/Microsoft.AspNetCore.SignalR.Tests/MapSignalRTests.cs
```diff
@@ -107,5 +121,16 @@ private IWebHost BuildWebHost(Action<HubRouteBuilder> configure)
                 })
                 .Build();
         }
+
+        private IWebHost BuildWebHostWithOutSignalR(Action<HubRouteBuilder> configure)
```
anurse: Does this need to be a helper method? Do we do it anywhere else?

test/Microsoft.AspNetCore.SignalR.Tests/MapSignalRTests.cs
```diff
@@ -23,6 +23,20 @@ public void MapSignalRFailsForInvalidHub()
             Assert.Equal("Duplicate definitions of 'OverloadedMethod'. Overloading is not supported.", ex.Message);
         }
 
+        [Fact]
+        public void NotAddingSignalRServiceThrows()
+        {
+            var ex = Assert.Throws<InvalidOperationException>(() =>
```
anurse: Wrap the `Assert.Throws` directly around the `UseSignalR` call, since that's what we expect to throw.

src/Microsoft.AspNetCore.SignalR/SignalRAppBuilderExtensions.cs
JamesNK: Full stop at end of exception message.

mikaelm12: "Full stop" 😄

JamesNK: Here in New Zealand we speak English as the Queen intended https://en.wikipedia.org/wiki/Full_stop

PR #1572, opened 2018-03-09
Fixed streaming hub method with ValueTask

JamesNK:
Updates dispatcher so streaming hub methods that use ValueTask are called correctly.


src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -392,12 +393,14 @@ private static bool IsStreamed(Type resultType)
...
         }
 
-        private bool TryGetStreamingEnumerator(HubConnectionContext connection, string invocationId, ObjectMethodExecutor methodExecutor, object result, Type resultType, out IAsyncEnumerator<object> enumerator)
+        private bool TryGetStreamingEnumerator(HubConnectionContext connection, string invocationId, ObjectMethodExecutor methodExecutor, object result, out IAsyncEnumerator<object> enumerator)
         {
             if (result != null)
             {
                 // TODO: cache reflection for performance, on HubMethodDescriptor maybe?
-                resultType = UnwrapTask(resultType);
+                var resultType = (!methodExecutor.IsMethodAsync)
```
davidfowl: !method.IsMethodAsync?Why is it inverted?

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -392,13 +393,15 @@ private static bool IsStreamed(Type resultType)
...
             if (result != null)
             {
-                // TODO: cache reflection for performance, on HubMethodDescriptor maybe?
-                resultType = UnwrapTask(resultType);
+                var resultType = (methodExecutor.IsMethodAsync)
+                    ? methodExecutor.AsyncResultType
+                    : methodExecutor.MethodReturnType;
 
+                // TODO: cache reflection for performance, on HubMethodDescriptor maybe?
                 var observableInterface = IsIObservable(resultType) ?
```
davidfowl: Can you do this next? 😄

PR #1574, opened 2018-03-10
Hub method reflection caching and invocation benchmarks

JamesNK:
Benchmarks in preparation for caching invocation reflection

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -1,10 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.IO.Pipelines;
 using System.Reactive.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Attributes.Jobs;
+using BenchmarkDotNet.Engines;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Protocols;
 using Microsoft.AspNetCore.SignalR.Internal;
@@ -13,10 +16,12 @@ using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using Moq;
 using DefaultConnectionContext = Microsoft.AspNetCore.Sockets.DefaultConnectionContext;
 
 namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
 {
+    //[SimpleJob(RunStrategy.Throughput, launchCount: 1, invocationCount: 10)]
     public class DefaultHubDispatcherBenchmark
     {
         private DefaultHubDispatcher<TestHub> _dispatcher;
@@ -41,12 +46,32 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
             var pair = DuplexPipe.CreateConnectionPair(options, options);
             var connection = new DefaultConnectionContext(Guid.NewGuid().ToString(), pair.Transport, pair.Application);
 
-            _connectionContext = new HubConnectionContext(connection, TimeSpan.Zero, NullLoggerFactory.Instance);
+            _connectionContext = new NoErrorHubConnectionContext(connection, TimeSpan.Zero, NullLoggerFactory.Instance);
 
-            var hubProtocol = Moq.Mock.Of<IHubProtocol>();
-            var dataEncoder = Moq.Mock.Of<IDataEncoder>();
+            var hubProtocolMock = new Mock<IHubProtocol>();
+            var dataEncoder = Mock.Of<IDataEncoder>();
 
-            _connectionContext.ProtocolReaderWriter = new HubProtocolReaderWriter(hubProtocol, dataEncoder);
+            _connectionContext.ProtocolReaderWriter = new HubProtocolReaderWriter(hubProtocolMock.Object, dataEncoder);
+        }
+
+        public class NoErrorHubConnectionContext : HubConnectionContext
+        {
+            public NoErrorHubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory) : base(connectionContext, keepAliveInterval, loggerFactory)
+            {
+            }
+
+            public override Task WriteAsync(HubMessage message)
+            {
+                if (message is CompletionMessage completionMessage)
+                {
+                    if (!string.IsNullOrEmpty(completionMessage.Error))
+                    {
+                        throw new Exception("Error invoking hub method: " + completionMessage.Error);
+                    }
+                }
+
+                return Task.CompletedTask;
+            }
         }
 
         public class TestHub : Hub
@@ -55,9 +80,9 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
             {
             }
 
-            public async Task InvocationAsync()
+            public Task InvocationAsync()
             {
-                await Task.Yield();
+                return Task.CompletedTask;
             }
 
             public int InvocationReturnValue()
@@ -65,16 +90,14 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
                 return 1;
             }
 
-            public async Task<int> InvocationReturnAsync()
+            public Task<int> InvocationReturnAsync()
             {
-                await Task.Yield();
-                return 1;
+                return Task.FromResult(1);
             }
 
-            public async ValueTask<int> InvocationValueTaskAsync()
+            public ValueTask<int> InvocationValueTaskAsync()
             {
-                await Task.Yield();
-                return 1;
+                return new ValueTask<int>(1);
             }
 
             public IObservable<int> SteamObservable()
@@ -82,16 +105,14 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
                 return Observable.Empty<int>();
             }
 
-            public async Task<IObservable<int>> SteamObservableAsync()
+            public Task<IObservable<int>> SteamObservableAsync()
             {
-                await Task.Yield();
-                return Observable.Empty<int>();
+                return Task.FromResult(Observable.Empty<int>());
             }
 
-            public async ValueTask<IObservable<int>> SteamObservableValueTaskAsync()
+            public ValueTask<IObservable<int>> SteamObservableValueTaskAsync()
             {
-                await Task.Yield();
-                return Observable.Empty<int>();
+                return new ValueTask<IObservable<int>>(Observable.Empty<int>());
             }
         }
 
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -1,22 +1,53 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Reactive.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Protocols;
 using Microsoft.AspNetCore.SignalR.Internal;
+using Microsoft.AspNetCore.SignalR.Internal.Encoders;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using DefaultConnectionContext = Microsoft.AspNetCore.Sockets.DefaultConnectionContext;
 
 namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
 {
     public class DefaultHubDispatcherBenchmark
     {
-        private readonly DefaultHubDispatcher<TestHub> _dispatcher;
-        private readonly HubConnectionContext _connectionContext;
+        private DefaultHubDispatcher<TestHub> _dispatcher;
+        private HubConnectionContext _connectionContext;
+
+        [GlobalSetup]
+        public void GlobalSetup()
+        {
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSignalRCore();
+
+            var provider = serviceCollection.BuildServiceProvider();
+
+            var serviceScopeFactory = provider.GetService<IServiceScopeFactory>();
+
+            _dispatcher = new DefaultHubDispatcher<TestHub>(
+                serviceScopeFactory,
+                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>()),
+                new Logger<DefaultHubDispatcher<TestHub>>(NullLoggerFactory.Instance));
+
+            var options = new PipeOptions();
+            var pair = DuplexPipe.CreateConnectionPair(options, options);
+            var connection = new DefaultConnectionContext(Guid.NewGuid().ToString(), pair.Transport, pair.Application);
+
+            _connectionContext = new HubConnectionContext(connection, TimeSpan.Zero, NullLoggerFactory.Instance);
+
+            var hubProtocol = Moq.Mock.Of<IHubProtocol>();
+            var dataEncoder = Moq.Mock.Of<IDataEncoder>();
+
+            _connectionContext.ProtocolReaderWriter = new HubProtocolReaderWriter(hubProtocol, dataEncoder);
+        }
 
         public class TestHub : Hub
         {
@@ -28,6 +59,40 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
             {
                 await Task.Yield();
             }
+
+            public int InvocationReturnValue()
+            {
+                return 1;
+            }
+
+            public async Task<int> InvocationReturnAsync()
+            {
+                await Task.Yield();
+                return 1;
+            }
+
+            public async ValueTask<int> InvocationValueTaskAsync()
+            {
+                await Task.Yield();
+                return 1;
+            }
+
+            public IObservable<int> SteamObservable()
+            {
+                return Observable.Empty<int>();
+            }
+
+            public async Task<IObservable<int>> SteamObservableAsync()
+            {
+                await Task.Yield();
+                return Observable.Empty<int>();
+            }
+
+            public async ValueTask<IObservable<int>> SteamObservableValueTaskAsync()
+            {
+                await Task.Yield();
+                return Observable.Empty<int>();
+            }
         }
 
         public class TestServiceScopeFactory : IServiceScopeFactory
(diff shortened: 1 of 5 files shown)


JamesNK:
TIL streaming is slooow

```
                           Method |       Mean |     Error |    StdDev |      Op/s | Allocated |
--------------------------------- |-----------:|----------:|----------:|----------:|----------:|
                       Invocation |   3.176 us | 0.3600 us | 0.2381 us | 314,834.9 |   1.23 KB |
                  InvocationAsync |   3.584 us | 0.4009 us | 0.2652 us | 279,052.3 |   1.34 KB |
            InvocationReturnValue |   3.626 us | 0.4230 us | 0.2798 us | 275,801.6 |    1.4 KB |
            InvocationReturnAsync |   4.468 us | 0.3385 us | 0.2239 us | 223,837.3 |   1.58 KB |
         InvocationValueTaskAsync |   4.365 us | 0.6386 us | 0.4224 us | 229,107.7 |   1.48 KB |
                  SteamObservable | 377.178 us | 4.9030 us | 3.2431 us |   2,651.3 |   6.34 KB |
             SteamObservableAsync | 381.484 us | 3.2692 us | 2.1623 us |   2,621.3 |   6.44 KB |
    SteamObservableValueTaskAsync | 214.774 us | 2.1665 us | 1.4330 us |   4,656.1 |   4.04 KB |
               SteamChannelReader | 367.572 us | 2.3842 us | 1.5770 us |   2,720.6 |   6.34 KB |
          SteamChannelReaderAsync | 378.704 us | 2.4904 us | 1.6473 us |   2,640.6 |   6.44 KB |
 SteamChannelReaderValueTaskAsync | 210.917 us | 3.3284 us | 2.2015 us |   4,741.2 |   4.04 KB |
```

BrennanConroy:
Nice, how long does the run take?

JamesNK:
Forever with default config. I turned down iterations a lot locally. With settings below it takes 100 secs on my computer. I need to figure out how to pass this stuff via command line to benchmarkdotnet

```
Add(Job.Core
    .With(CsProjCoreToolchain.From(NetCoreAppSettings.NetCoreApp21))
    .WithRemoveOutliers(false)
    .With(new GcMode { Server = true })
    .With(RunStrategy.Throughput)
    .WithLaunchCount(1)
    .WithWarmupCount(10)
    .WithTargetCount(10)
    .WithInvocationCount(256));

```

BrennanConroy:
We use a shared benchmarks package for a lot of our repos, you can look at the source here https://github.com/aspnet/Common/tree/dev/shared/Microsoft.AspNetCore.BenchmarkRunner.Sources

JamesNK:
Yeah that's what I modified. Those settings take forever to run. Do you know if they can be override from the command line to speed things along? Waiting 10 minutes for benchmarks to run to test a minor perf change isn't productive

BrennanConroy:
I don't know, @pakrym

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                return Task.FromResult(1);
+            }
+
+            public ValueTask<int> InvocationValueTaskAsync()
+            {
+                return new ValueTask<int>(1);
+            }
+
+            public IObservable<int> SteamObservable()
```
BrennanConroy: Stream
Also, for all the other ones

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
```diff
@@ -0,0 +1,203 @@
+﻿using System;
```
BrennanConroy: Copyright too

JamesNK:
Streaming was throwing an error which is why it was so bad. Halfway through updates:

``` ini

BenchmarkDotNet=v0.10.11, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.248)
Processor=AMD Ryzen 7 1700 Eight-Core Processor, ProcessorCount=16
Frequency=3703983 Hz, Resolution=269.9796 ns, Timer=TSC
.NET Core SDK=2.1.300-preview2-008251
  [Host]     : .NET Core 2.0.5 (Framework 4.6.26020.03), 64bit RyuJIT
  Job-MUNCIE : .NET Core 2.1.0-preview2-26308-01 (Framework 4.6.26307.01), 64bit RyuJIT

RemoveOutliers=False  Runtime=Core  Server=True  
Toolchain=.NET Core 2.1  InvocationCount=256  LaunchCount=1  
RunStrategy=Throughput  TargetCount=10  WarmupCount=10  

```
|                           Method |      Mean |     Error |    StdDev |      Op/s | Allocated |
|--------------------------------- |----------:|----------:|----------:|----------:|----------:|
|                       Invocation |  2.425 us | 0.2874 us | 0.1901 us | 412,386.0 |   1.09 KB |
|                  InvocationAsync |  2.660 us | 0.6844 us | 0.4527 us | 375,889.8 |   1.09 KB |
|            InvocationReturnValue |  2.728 us | 0.3799 us | 0.2513 us | 366,568.0 |   1.18 KB |
|            InvocationReturnAsync |  2.878 us | 0.5330 us | 0.3525 us | 347,454.1 |   1.27 KB |
|         InvocationValueTaskAsync |  2.689 us | 0.2763 us | 0.1828 us | 371,901.9 |   1.24 KB |
|                  SteamObservable | 15.465 us | 0.4398 us | 0.2909 us |  64,664.0 |   6.33 KB |
|             SteamObservableAsync | 15.522 us | 0.5942 us | 0.3930 us |  64,425.4 |   6.42 KB |
|    SteamObservableValueTaskAsync | 16.164 us | 0.9328 us | 0.6170 us |  61,867.2 |   6.41 KB |
|               SteamChannelReader | 15.419 us | 0.6392 us | 0.4228 us |  64,854.8 |   6.33 KB |
|          SteamChannelReaderAsync | 16.141 us | 0.6359 us | 0.4206 us |  61,955.3 |   6.42 KB |
| SteamChannelReaderValueTaskAsync | 16.416 us | 0.6506 us | 0.4303 us |  60,914.5 |   6.41 KB |

JamesNK:
Creating AsyncEnumerator with compiled expression. Streaming calls 15us to 10us.

``` ini

BenchmarkDotNet=v0.10.11, OS=Windows 10 Redstone 3 [1709, Fall Creators Update] (10.0.16299.248)
Processor=AMD Ryzen 7 1700 Eight-Core Processor, ProcessorCount=16
Frequency=3703983 Hz, Resolution=269.9796 ns, Timer=TSC
.NET Core SDK=2.1.300-preview2-008251
  [Host]     : .NET Core 2.0.5 (Framework 4.6.26020.03), 64bit RyuJIT
  Job-SMGCYT : .NET Core 2.1.0-preview2-26308-01 (Framework 4.6.26307.01), 64bit RyuJIT

RemoveOutliers=False  Runtime=Core  Server=True  
Toolchain=.NET Core 2.1  InvocationCount=256  LaunchCount=1  
RunStrategy=Throughput  TargetCount=10  WarmupCount=10  

```
|                           Method |      Mean |     Error |    StdDev |      Op/s | Allocated |
|--------------------------------- |----------:|----------:|----------:|----------:|----------:|
|                       Invocation |  2.496 us | 0.3826 us | 0.2531 us | 400,608.2 |   1.09 KB |
|                  InvocationAsync |  2.629 us | 0.7065 us | 0.4673 us | 380,337.6 |   1.09 KB |
|            InvocationReturnValue |  2.505 us | 0.3366 us | 0.2226 us | 399,250.4 |   1.18 KB |
|            InvocationReturnAsync |  2.792 us | 0.5051 us | 0.3341 us | 358,149.9 |   1.27 KB |
|         InvocationValueTaskAsync |  2.628 us | 0.3667 us | 0.2425 us | 380,505.5 |   1.24 KB |
|                  SteamObservable | 10.800 us | 0.5220 us | 0.3453 us |  92,595.1 |    5.8 KB |
|             SteamObservableAsync | 11.182 us | 1.6867 us | 1.1156 us |  89,429.8 |   5.89 KB |
|    SteamObservableValueTaskAsync | 10.481 us | 0.5168 us | 0.3418 us |  95,406.8 |   5.88 KB |
|               SteamChannelReader | 10.195 us | 0.5454 us | 0.3608 us |  98,084.2 |    5.8 KB |
|          SteamChannelReaderAsync | 10.601 us | 0.5601 us | 0.3705 us |  94,331.9 |   5.89 KB |
| SteamChannelReaderValueTaskAsync | 10.586 us | 0.7765 us | 0.5136 us |  94,463.5 |   5.88 KB |

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -267,8 +267,7 @@ private string BuildUnexpectedErrorMessage(string methodName, Exception exceptio
 
         private static async Task<object> ExecuteHubMethod(ObjectMethodExecutor methodExecutor, THub hub, object[] arguments)
         {
-            // ReadableChannel is awaitable but we don't want to await it.
-            if (methodExecutor.IsMethodAsync && !IsChannel(methodExecutor.MethodReturnType, out _))
```
JamesNK: I couldn't find any reference of `ReadableChannel` and no tests failed when I removed this...

davidfowl: They must have removed that feature from Channels.

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -340,10 +324,10 @@ private static bool IsChannel(Type type, out Type payloadType)
             return authorizationResult.Succeeded;
         }
 
-        private async Task<bool> ValidateInvocationMode(ObjectMethodExecutor methodExecutor, bool isStreamedInvocation,
+        private async Task<bool> ValidateInvocationMode(HubMethodDescriptor hubMethodDescriptor, bool isStreamedInvocation,
             HubMethodInvocationMessage hubMethodInvocationMessage, HubConnectionContext connection)
         {
-            var isStreamedResult = IsStreamed(methodExecutor);
+            var isStreamedResult = hubMethodDescriptor.IsObservable || hubMethodDescriptor.IsChannel;
```
davidfowl: Add hubMethodDescriptor.IsStreamed

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Channels;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    // REVIEW: We can decide to move this out of here if we want pluggable hub discovery
```
davidfowl: Remove this comment

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        private Func<object, CancellationToken, IAsyncEnumerator<object>> CreateConversionMethod(MethodInfo methodInfo)
+        {
+            var methodParameters = methodInfo.GetParameters();
+
+            var targetParameter = Expression.Parameter(typeof(object), "arg1");
+            var parametersParameter = Expression.Parameter(typeof(CancellationToken), "arg2");
+
+            var parameters = new List<Expression>
+                {
```
davidfowl: nit: fix formatting

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            if (_convertToEnumerator == null)
+            {
+                _convertToEnumerator = CreateConversionMethod(_getAsyncEnumeratorMethod.MakeGenericMethod(StreamReturnType));
+            }
+
+            return _convertToEnumerator.Invoke(channel, cancellationToken);
+        }
+
+        private Func<object, CancellationToken, IAsyncEnumerator<object>> CreateConversionMethod(MethodInfo methodInfo)
```
davidfowl: Add some comments to this method detailing the final codegen. Something like this https://github.com/aspnet/HttpAbstractions/blob/9119433ab778d7adf04c4601917809f2f4582a93/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UseMiddlewareExtensions.cs#L147-L171

davidfowl:
@JamesNK when you paste results from BDN the smaller table looks much better. Don't bother with the markdown table.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            if (_convertToEnumerator == null)
+            {
+                _convertToEnumerator = CompileConvertToEnumerator(GetAsyncEnumeratorMethod);
+            }
+
+            return _convertToEnumerator.Invoke(channel, cancellationToken);
+        }
+
+        private Func<object, CancellationToken, IAsyncEnumerator<object>> CompileConvertToEnumerator(MethodInfo methodInfo)
```
davidfowl: Can this be static?

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                _convertToEnumerator = CompileConvertToEnumerator(FromObservableMethod);
+            }
+
+            return _convertToEnumerator.Invoke(observable, cancellationToken);
+        }
+
+        public IAsyncEnumerator<object> FromChannel(object channel, CancellationToken cancellationToken)
+        {
+            if (_convertToEnumerator == null)
```
davidfowl: Everything else on this type is immutable, do we want to compile this delegate up front or lazily? Are there any thread safety concerns or will it just do more work.

JamesNK: No thread safety issues. Just more work if there are multiple requests for a streamable method on startup.

I think it should be lazy. Decreases startup time and overall memory usage by only compiling methods that are used.

One optimization could be to place them in a concurrent dictionary and share compiled methods between method descriptors. Each method that returns `IObservable<int>` will compile its own copy right now.

JamesNK: `ObjectMethodExecutor` isn't lazy and it also doesn't share compiled methods.

I'd be curious to see how MVC uses `ObjectMethodExecutor` - whether it creates instances for it up front for every action method or whether creates them on first invoke. Right now SignalR creates them up front.

JamesNK: `ControllerActionInvokerCache.GetCachedResult` creates `ObjectMethodExecutor` the first time an action is called. IMO SignalR should do the same. Compiling a lot of methods unnecessarily will hurt startup time and memory usage.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
```diff
@@ -0,0 +1,200 @@
...
+                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>()),
+                new Logger<DefaultHubDispatcher<TestHub>>(NullLoggerFactory.Instance));
+
+            var options = new PipeOptions();
+            var pair = DuplexPipe.CreateConnectionPair(options, options);
+            var connection = new DefaultConnectionContext(Guid.NewGuid().ToString(), pair.Transport, pair.Application);
+
+            _connectionContext = new NoErrorHubConnectionContext(connection, TimeSpan.Zero, NullLoggerFactory.Instance);
+
+            var hubProtocolMock = new Mock<IHubProtocol>();
```
pakrym: I wouldn't use Moq in a benchmark. In theory it should return empty method implementations by default but it might call into Moq infrastructure to record invocations etc.

pakrym:
As to benchmarks taking long time I'll do two things:
1. Update to newer BDN as there are some perf improvements made since 0.10.11
2. Add another configuration with less warmup and main iterations for large benchmarks.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
```diff
@@ -0,0 +1,200 @@
...
+        [Benchmark]
+        public Task InvocationValueTaskAsync()
+        {
+            return _dispatcher.DispatchMessageAsync(_connectionContext, new InvocationMessage("123", "InvocationValueTaskAsync", null));
+        }
+
+        [Benchmark]
+        public Task StreamObservable()
+        {
+            return _dispatcher.DispatchMessageAsync(_connectionContext, new StreamInvocationMessage("123", "StreamObservable", null));
```
pakrym: This is exactly the same as `StreamChannelReader`

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
```diff
@@ -0,0 +1,200 @@
...
+            public ValueTask<ChannelReader<int>> StreamChannelReaderValueTaskAsync()
+            {
+                return new ValueTask<ChannelReader<int>>(ChannelReaderInstance);
+            }
+        }
+
+        [Benchmark]
+        public Task Invocation()
+        {
+            return _dispatcher.DispatchMessageAsync(_connectionContext, new InvocationMessage("123", "Invocation", null));
```
pakrym: Benchmarks differ only in one string. Params(https://github.com/dotnet/BenchmarkDotNet/blob/master/docs/guide/Advanced/Params.md) can be used for this purpose.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks.csproj
```diff
@@ -14,6 +14,9 @@
     <PackageReference Include="Microsoft.AspNetCore.BenchmarkRunner.Sources" Version="$(MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion)" />
     <PackageReference Include="System.Threading.Channels" Version="$(SystemThreadingChannelsPackageVersion)" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="$(SystemThreadingTasksExtensionsPackageVersion)" />
+    <PackageReference Include="System.Reactive.Linq" Version="$(SystemReactiveLinqPackageVersion)" />
```
pakrym: Sort.

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
anurse: It's [`ChannelReader`](https://github.com/dotnet/corefx/blob/master/src/System.Threading.Channels/src/System/Threading/Channels/ChannelReader.cs) now and they did indeed remove the awaitable feature.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                return iface.IsGenericType && iface.GetGenericTypeDefinition() == typeof(IObservable<>);
+            }
+        }
+
+        public IAsyncEnumerator<object> FromObservable(object observable, CancellationToken cancellationToken)
+        {
+            if (_convertToEnumerator == null)
+            {
+                _convertToEnumerator = CompileConvertToEnumerator(FromObservableMethod);
```
anurse: Let's comment on the known race here (and that it's harmless) just to avoid someone getting confused next time around.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
anurse: `methodInfo` should perhaps be `adapterMethod` because I initially confused it with the hub method.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
```diff
@@ -0,0 +1,149 @@
...
+            // AsyncEnumeratorAdapters.FromObservable<T>(observable, cancellationToken);
+            //
+            // ChannelReader<T>
+            // AsyncEnumeratorAdapters.GetAsyncEnumerator<T>(channelReader, cancellationToken);
+
+            var genericMethodInfo = methodInfo.MakeGenericMethod(StreamReturnType);
+
+            var methodParameters = genericMethodInfo.GetParameters();
+
+            var targetParameter = Expression.Parameter(typeof(object), "arg1");
```
anurse: names for the arguments? I don't know if they show up in callstacks or something but if `Expression.Parameter` takes a name, perhaps it should be a more useful name than `arg1`. That's assuming compiling to `Func` doesn't mandate the argument names. Even if they get wiped out by compiling to `Func`, it doesn't hurt, if nothing else it serves as commentary in this code

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
anurse: Agreed, we should do the same as MVC here.

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubMethodDescriptor.cs
JamesNK: `arg1` and `arg2` is coming from the parameter names of `Func<T1, T2, TReturn1>`. Need to use those names to get a reference to them and pass them to our own method. I'll improve docs to be clearer

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
JamesNK: I thought about that but the don't differ by just one string - there are invocation messages and stream messages which have a different message type.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
JamesNK: No, that's a bug. Thanks

JamesNK:
@pakrym Sounds good. I think the current configuration is a little too thorough for most usage. The default should be quicker. Launch count of 1 alone will cut the time by 66%.

Ideally it would be nice if a config could be selected from the command line but I don't know whether that is possible with BDN.

JamesNK:
@pakrym Random question: what is the best way to debug a BDN benchmark? Is there a way to run it with a debugger attached?

pakrym:
Easy way it just to add `-validate *` to command line arguments in VS and run it.

Harder way: if you look at source referenced Program.cs it has `BeforeMain` partial method, usually I add another Program2.cs that overrides the partial method and call benchmark methods manually.

pakrym:
> Ideally it would be nice if a config could be selected from the command line but I don't know whether that is possible with BDN. 

It is with our harness, but we just use it for validation - running a single iteration of every benchmark to make sure it works during repo build

PR #1577, opened 2018-03-12
Remove threshold from client SSE pipe

BrennanConroy:


BrennanConroy:
Virtual @davidfowl approval

PR #1588, opened 2018-03-13
add AddRedis overload that takes connection string

anurse:
This has been annoying me for a while, and I kept forgetting to file it so I just decided to spend the 10 mins fixing it :). It's a pain to add Azure Redis as a redis cache because it gives you this convenient connection string and we don't provide an easy way to just ~~paste it in~~ read it in from config.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisDependencyInjectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisDependencyInjectionExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.AspNetCore.SignalR.Redis;
+using StackExchange.Redis;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
@@ -12,6 +13,14 @@ namespace Microsoft.Extensions.DependencyInjection
         public static ISignalRBuilder AddRedis(this ISignalRBuilder builder)
         {
             return AddRedis(builder, o => { });
+        }
+
+        public static ISignalRBuilder AddRedis(this ISignalRBuilder builder, string redisConnectionString)
+        {
+            return AddRedis(builder, o =>
+            {
+                o.Options = ConfigurationOptions.Parse(redisConnectionString);
+            });
         }
 
         public static ISignalRBuilder AddRedis(this ISignalRBuilder builder, Action<RedisOptions> configure)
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisDependencyInjectionExtensionsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        // No need to go too deep with these tests, or we're just testing StackExchange.Redis again :). It's the one doing the parsing.
+        [Theory]
+        [InlineData("testredis.example.com", "testredis.example.com", 0, null, false)]
+        [InlineData("testredis.example.com:6380,ssl=True", "testredis.example.com", 6380, null, true)]
+        [InlineData("testredis.example.com:6380,password=hunter2,ssl=True", "testredis.example.com", 6380, "hunter2", true)]
+        public void AddRedisWithConnectionStringProperlyParsesOptions(string connectionString, string host, int port, string password, bool useSsl)
+        {
+            var services = new ServiceCollection();
+            services.AddSignalR().AddRedis(connectionString);
```
JamesNK: Do we need to worry about what happens if `AddRedis` is called multiple times? Ignore it? Throw error?

anurse: Our general pattern is to allow that to happen. The way Options works is that each delegate that's registered to the container is run in the order of registration. So if I did:

```
.AddRedis(o => o.Options.EndPoints.Add("myredis.example.com"))
.AddRedis("adifferentredis.example.com")
.AddRedis(o => o.Options.Ssl = true);
```

The first line would run and add the endpoint, but the second line would overwrite the config (because we replace the `ConfigurationOptions` and StackExchange.Redis doesn't provide a way to merge them). The third line would **augment** the configuration to enable SSL.

You'd end up with `adifferentredis.example.com` and SSL enabled.

JamesNK: ok!

src/Microsoft.AspNetCore.SignalR.Redis/RedisDependencyInjectionExtensions.cs
```diff
@@ -14,6 +15,14 @@ public static ISignalRBuilder AddRedis(this ISignalRBuilder builder)
             return AddRedis(builder, o => { });
         }
 
+        public static ISignalRBuilder AddRedis(this ISignalRBuilder builder, string redisConnectionString)
```
davidfowl: YESSS

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisDependencyInjectionExtensionsTests.cs
anurse: Calling these `.AddFoo` methods multiple times is also a common pattern when you want to override our default options (see how we call `.AddJsonProtocol` for you in [`.AddSignalR`](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs#L12) but you can call it again to change your serializer settings).

Issue #1555, opened 2018-03-07
If WriteAsync throws other connections might not be sent to


BrennanConroy:
`WriteAsync` can throw, and if it throws before going async then other connections after the faulty one won't get the Send.
https://github.com/aspnet/SignalR/blob/1c44e8febfd57a1038715d6c74f030cebdcd14f2/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs#L86

Also the Hub will get the error, and I think we decided it wasn't an interesting failure for the Hub to see

anurse:
We should take this in preview2

anurse:
In Redis as well

PR #1589, opened 2018-03-13
Ignore writeasync failures when sending to multiple connections

BrennanConroy:
#1555

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
@@ -26,7 +26,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
         [GlobalSetup]
         public void GlobalSetup()
         {
-            _hubLifetimeManager = new DefaultHubLifetimeManager<Hub>(new Logger<DefaultHubLifetimeManager<Hub>>(NullLoggerFactory.Instance));
+            _hubLifetimeManager = new DefaultHubLifetimeManager<Hub>(NullLogger<DefaultHubLifetimeManager<Hub>>.Instance);
 
 
             IHubProtocol protocol;
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
@@ -7,6 +7,7 @@ using BenchmarkDotNet.Attributes;
 using Microsoft.AspNetCore.SignalR.Internal;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 using Microsoft.AspNetCore.Sockets;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
@@ -25,7 +26,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
         [GlobalSetup]
         public void GlobalSetup()
         {
-            _hubLifetimeManager = new DefaultHubLifetimeManager<Hub>();
+            _hubLifetimeManager = new DefaultHubLifetimeManager<Hub>(new Logger<DefaultHubLifetimeManager<Hub>>(NullLoggerFactory.Instance));
 
 
             IHubProtocol protocol;
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -36,7 +36,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
 
             _dispatcher = new DefaultHubDispatcher<TestHub>(
                 serviceScopeFactory,
-                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>(new Logger<DefaultHubLifetimeManager<TestHub>>(NullLoggerFactory.Instance))),
+                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>(NullLogger<DefaultHubLifetimeManager<TestHub>>.Instance)),
                 new Logger<DefaultHubDispatcher<TestHub>>(NullLoggerFactory.Instance));
 
             var pair = DuplexPipe.CreateConnectionPair(PipeOptions.Default, PipeOptions.Default);
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -36,7 +36,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
 
             _dispatcher = new DefaultHubDispatcher<TestHub>(
                 serviceScopeFactory,
-                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>()),
+                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>(new Logger<DefaultHubLifetimeManager<TestHub>>(NullLoggerFactory.Instance))),
                 new Logger<DefaultHubDispatcher<TestHub>>(NullLoggerFactory.Instance));
 
             var pair = DuplexPipe.CreateConnectionPair(PipeOptions.Default, PipeOptions.Default);
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -83,11 +83,7 @@ namespace Microsoft.AspNetCore.SignalR
                     continue;
                 }
 
-                try
-                {
-                    tasks.Add(connection.WriteAsync(message));
-                }
-                catch { }
+                tasks.Add(SafeWriteAsync(connection, message));
             }
 
             return Task.WhenAll(tasks);
@@ -125,11 +121,7 @@ namespace Microsoft.AspNetCore.SignalR
                 var message = CreateInvocationMessage(methodName, args);
                 var tasks = group.Values.Select(async c =>
                 {
-                    try
-                    {
-                        await c.WriteAsync(message);
-                    }
-                    catch { }
+                    await SafeWriteAsync(c, message);
                 });
                 return Task.WhenAll(tasks);
             }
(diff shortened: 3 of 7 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/DefaultHubLifetimeManagerTests.cs
```diff
@@ -128,7 +128,37 @@ public class DefaultHubLifetimeManagerTests
         }
 
         [Fact]
-        public async Task InvokeConnectionAsyncOnNonExistentConnectionNoops()
+        public async Task SendAllAsyncSendsToAllConnectionsEvenWhenSomeFailToSend()
```
BrennanConroy: Should probably add a test for each method I changed

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -83,7 +83,11 @@ private Task SendAllWhere(string methodName, object[] args, Func<HubConnectionCo
                     continue;
                 }
 
-                tasks.Add(connection.WriteAsync(message));
+                try
+                {
+                    tasks.Add(connection.WriteAsync(message));
+                }
+                catch { }
```
BrennanConroy: Do we want to log?

JamesNK: Isn't `WriteAsync` returning a `Task` and here you're simply adding a `Task` to a list. Does that need a try/catch? Isn't it the await that needs a try/catch?

anurse: It looks like `WriteAsync` starts with and `await`, so it's not likely to be an issue in practice. This would be more of a defense-in-depth kind of thing (so that if it's refactored to do argument validation or serialization synchronously, it doesn't have a risk of killing the whole loop).

Perhaps we should be doing `tasks.Add(Task.FromException(ex))`, so that the caller still gets the error? Or maybe this is overkill.

JamesNK: The `async` keyword on `WriteAsync` should generate a state machine that will always catch exceptions and return them as a faulted `Task` instead of throwing out of the method, at least that is my understanding of async/await.

davidfowl: That's correct but this shouldn't start failing if we remove the state machine from WriteAsync (which we'll very likely do).

anurse:
> Should we log when a connection fails?

If it fails inside the state machine, the exception bubbles up to the caller via the `WhenAll`. We should do the same.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
JamesNK: When that happens shouldn't the try/catch should be inside WriteAsync and return and `Task.FromException(ex)`?

anurse: That would make the calling code consistent

davidfowl: > When that happens shouldn't the try/catch should be inside WriteAsync and return and Task.FromException(ex)?

Ugh, I dislike that but I see why it's an attractive option. try catch is fine..... 😄. I don't know if there's a best practice here tbh. For e.g. if you're doing argument validation, should it throw async or sync?

JamesNK: Argument validation should throw sync. `JsonTextWriter.WritePropertyNameAsync(null)` for example will throw a `NullArgumentException` instead of returning a faulted task, but any internal errors like the underlying stream being disposed will return a faulted task.

Similar to `Task.Delay(-1000)` throwing an argument exception instead of returning a faulted task.

davidfowl: Which is why I like having a try catch

JamesNK: We control the method so we can choose not to do that.

What about a `SafeWriteAsync` helper method that wraps WriteAsync and has the try/catch? If we need it wrapped for safety then it would be nicer that it was in one place.

davidfowl: That sounds good.

test/Microsoft.AspNetCore.SignalR.Tests/DefaultHubLifetimeManagerTests.cs
anurse: Yeah, and unfortunately it doesn't really work with a Theory.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -215,5 +215,17 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
                 return userIds.Contains(connection.UserIdentifier);
             });
         }
+
+        private Task SafeWriteAsync(HubConnectionContext connection, InvocationMessage message)
```
JamesNK: Add a comment with what this method is protecting against

davidfowl:
After the build is fixed...

test/Microsoft.AspNetCore.SignalR.Tests/DefaultHubLifetimeManagerTests.cs
```diff
@@ -128,7 +128,38 @@ public class DefaultHubLifetimeManagerTests
...
+                // Force an exception when writing to connection
+                connectionMock.Setup(m => m.WriteAsync(It.IsAny<HubMessage>())).Callback(() => tcs.TrySetResult(null)).Throws(new Exception("Message"));
+                connectionMock2.Setup(m => m.WriteAsync(It.IsAny<HubMessage>())).Callback(() => tcs2.TrySetResult(null)).Throws(new Exception("Message"));
+                var connection = connectionMock.Object;
+                var connection2 = connectionMock2.Object;
+
+                await manager.OnConnectedAsync(connection).OrTimeout();
+                await manager.OnConnectedAsync(connection2).OrTimeout();
+
+                var exception = await Assert.ThrowsAsync<Exception>(() => manager.SendAllAsync("Hello", new object[] { "World" }).OrTimeout());
```
BrennanConroy: Not a huge fan of this, you get an exception from your hub method if a connection failed to send to

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
JamesNK: Can be static?

JamesNK: Also, can it be static?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -662,6 +627,19 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
             return Task.CompletedTask;
         }
 
+        private Task SafeWriteAsync(HubConnectionContext connection, InvocationMessage message)
```
JamesNK: Comment explaining what it is protecting against

test/Microsoft.AspNetCore.SignalR.Tests/DefaultHubLifetimeManagerTests.cs
anurse: It's always been that way (at least in ASP.NET Core). Not sure what SignalR 2.x did. I think you do want a chance to know about this in the application, so I do like that it's sent. Though of course, in the Redis scenario, you don't get any notification if another server fails to send to a client. Maybe logging and dumping is sufficient here.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -215,5 +215,17 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+        {
+            try
+            {
+                return connection.WriteAsync(message);
```
anurse: What if we `await`ed here (to get all the exceptions), logged the exception and ignored it? Is there actually any value in sending this exception back to the client? What are they going to do about it?

Plus, things are worse in the Redis case. If **your** server fails to send the message you get an exception but if a **different** one fails to send to one of it's clients you get nothing.

BrennanConroy:
🆙 📅

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -215,5 +215,16 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+
+        // This method is to protect against connections throwing synchronously when writing to them and preventing other connections from being written to
+        private static async Task SafeWriteAsync(HubConnectionContext connection, InvocationMessage message)
+        {
+            try
+            {
+                await connection.WriteAsync(message);
+            }
+            // This exception isn't interesting to users
+            catch { }
```
anurse: Add a logger here and log this.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
```diff
@@ -25,7 +26,7 @@ public class BroadcastBenchmark
         [GlobalSetup]
         public void GlobalSetup()
         {
-            _hubLifetimeManager = new DefaultHubLifetimeManager<Hub>();
+            _hubLifetimeManager = new DefaultHubLifetimeManager<Hub>(new Logger<DefaultHubLifetimeManager<Hub>>(NullLoggerFactory.Instance));
```
anurse: `NullLogger<DefaultHubLifetimeManager<THub>>.Instance`?

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
```diff
@@ -36,7 +36,7 @@ public void GlobalSetup()
 
             _dispatcher = new DefaultHubDispatcher<TestHub>(
                 serviceScopeFactory,
-                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>()),
+                new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>(new Logger<DefaultHubLifetimeManager<TestHub>>(NullLoggerFactory.Instance))),
```
anurse: ditto

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -215,5 +222,30 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+            catch (Exception ex)
+            {
+                Log.FailedWritingMessage(_logger, ex);
+            }
+        }
+
+        private static class Log
+        {
+            private static readonly Action<ILogger, Exception> _failedWritingMessage =
+                LoggerMessage.Define(LogLevel.Warning, new EventId(1, "FailedWritingMessage"), "Failed writing message.");
```
davidfowl: This should be trace.

PR #1590, opened 2018-03-13
[WIP] change our default casing for serialization to include dictionary keys

### LINKED ISSUE #1550: JSON Serialization is different between DefaultHubLifetimeManager and RedisHubLifetimeManager
The JSON serializer in the RedisHubLifetimeManager doesn't camelCase the properties of a serialized object where the JSONHubProtocol does if using the DefaultHubLifetimeManager.  This results in breaking changes based on the default settings between the two hubs when doing initial implementation using DefaultHubLifetimeManager then switching to Redis.

anurse:
Fixes #1550

I didn't have a chance to write the tests before I had to leave for the evening but wanted to start a discussion on this :). Tests will be in RedisHubLifetimeManagerTests and will consist of calling SendAllAsync with a PascalCased object and ensuring it becomes camelCased.

It's not a totally benign change since it means that if I write the following:

```csharp
Clients.All.SendAsync("SomeMethod", new Dictionary<string, string>() {
  { "FooBar", "Baz" }
});
```

It will be serialized as:

```json
{
  "fooBar": "Baz"
}
```

That may not be desirable, since I gave you a dictionary with strings (indicating I knew what I wanted to serialize).

I think the fact that it's configurable makes up for that though. Other fixes for this are a little hairy. We could serialize using the payload serializer before putting the value in Redis, but we don't easily have access to that in the lifetime manager today since we don't know the protocols in use.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
@@ -377,7 +377,16 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
 
         internal static JsonSerializerSettings CreateDefaultSerializerSettings()
         {
-            return new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() };
+            return new JsonSerializerSettings
+            {
+                ContractResolver = new DefaultContractResolver()
+                {
+                    NamingStrategy = new CamelCaseNamingStrategy()
+                    {
+                        ProcessDictionaryKeys = true
+                    }
+                }
+            };
         }
     }
 }


JamesNK:
So the problem is that the message pack library camel cases property names AND dictionary keys? And because there is no way to only camel case properties names then JSON also needs to camel case both to match?

anurse:
Well, no, the problem is that when we send the message through Redis, it goes from being a .NET Object (when it was serialized to JSON by our internal serializer) into being a Dictionary (when it's deserialized by the other servers using the internal serializer). Then, when those servers go to serialize the value to their clients (using the **user-customizable** serializer settings), it's a Dictionary and JSON.NET no longer camel-cases it.

It's actually not related to MessagePack at all. This would happen if all clients were connected via JSON. It's also magnified because in some cases, all clients (even the ones on the same server as the sender) get the value after it's been filtered to and from JSON.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -377,7 +377,16 @@ private object[] BindArguments(JArray args, IReadOnlyList<Type> paramTypes)
...
         internal static JsonSerializerSettings CreateDefaultSerializerSettings()
         {
-            return new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() };
+            return new JsonSerializerSettings
+            {
+                ContractResolver = new DefaultContractResolver()
+                {
+                    NamingStrategy = new CamelCaseNamingStrategy()
+                    {
+                        ProcessDictionaryKeys = true
```
davidfowl: Does MVC do this?

anurse: Probably not, but MVC doesn't have the intermediate bus in the way. I'm not sure this is the best way to solve this yet, but we need to do something to resolve the underlying issue :)

JamesNK: No, MVC doesn't. People complain that it doesn't, but people would also complain if it did.

anurse:
Chatted with @davidfowl . We're going to go with this solution, it's not the best, but it'll get it working for Preview 2. The downside is that our default serialization logic is a little different from MVC, but it'll be consistent at least.

anurse:
Never mind, it looks like this doesn't even fix the issue... I'll have to take a deeper look later.

anurse:
This didn't work :(

Issue #1508, opened 2018-02-27
Make Parsers Incremental and Consistent


anurse:
We should make the parsers incremental (so we don't have to buffer) and make the APIs consistent (instead of Reading multiple but writing a single one, etc.). We'll still keep them stateless. cc @davidfowl

anurse:
This should include #1618

PR #1591, opened 2018-03-14
Rename negotiate to handshake, add response, add close frame

JamesNK:


test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -283,11 +284,12 @@ public class HubEndpointTests
 
             using (var client = new TestClient())
             {
-                // TestClient automatically writes negotiate, for this test we want to assume negotiate never gets sent
-                client.Connection.Transport.Input.TryRead(out var item);
-                client.Connection.Transport.Input.AdvanceTo(item.Buffer.End);
+                Task endPointTask = await client.ConnectAsync(endPoint, false);
```
JamesNK: This test doesn't work, I don't think it has ever worked as intended (`NegotiateTimeout` doesn't do anything in the test), and I'm not sure how to fix it.

Instead of waiting on the server for content and being cancelled by the cancellation token, the test has always exited straight away.

JamesNK: Checking for `NegotiationResponseMessage` below is why it doesn't work now. If negotiate was properly timing out then you would get a response saying so. Instead it doesn't send anything.

JamesNK:
`HubConnectionTests.CheckHttpConnectionFeatures` also fail and I'm not sure why. There are 3 tests because of the parameters and if you run them one at a time and they each pass, but if you run them together then the first passes and the second and third always fail with a timeout.

anurse:
> but if you run them together then the first passes and the second and third always fail with a timeout.

That smells like there's some shared global state getting messed up...

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/NegotiationProtocol.cs
```diff
@@ -31,7 +32,70 @@ public static void WriteMessage(NegotiationMessage negotiationMessage, Stream ou
...
+                    var error = JsonUtils.GetOptionalProperty<string>(negotiationJObject, ErrorPropertyName);
+                    negotiationMessage = new NegotiationResponseMessage(error);
+                }
+            }
+            return true;
+        }
+
+        public static bool TryParseResponseMessage(ReadOnlySequence<byte> buffer, out NegotiationResponseMessage negotiationMessage, out SequencePosition consumed, out SequencePosition examined)
+        {
+            var separator = buffer.PositionOf(TextMessageFormatter.RecordSeparator);
```
anurse: This code being identical means we can probably make this a helper. `TryReadTextMessageIntoSingleSpan` or something.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/NegotiationProtocol.cs
```diff
@@ -31,7 +32,70 @@ public static void WriteMessage(NegotiationMessage negotiationMessage, Stream ou
...
+        }
+
+        public static bool TryParseResponseMessage(ReadOnlySpan<byte> input, out NegotiationResponseMessage negotiationMessage)
+        {
+            if (!TextMessageParser.TryParseMessage(ref input, out var payload))
+            {
+                throw new InvalidDataException("Unable to parse payload as a negotiation message.");
+            }
+
+            using (var memoryStream = new MemoryStream(payload.ToArray()))
```
anurse: There's some code in here in common with `TryParseRequestMessage` as well, can we get that into a private helper?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -257,8 +276,8 @@ private static void AbortConnection(object state)
         private static class Log
         {
             // Category: HubConnectionContext
-            private static readonly Action<ILogger, string, Exception> _usingHubProtocol =
-                LoggerMessage.Define<string>(LogLevel.Information, new EventId(1, "UsingHubProtocol"), "Using HubProtocol '{protocol}'.");
+            private static readonly Action<ILogger, string, Exception> _negotiateComplete =
+                LoggerMessage.Define<string>(LogLevel.Information, new EventId(1, "NegotiateComplete"), "Successfully negotiated connection. Using HubProtocol '{protocol}'.");
```
anurse: :+1:

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
```diff
@@ -5,6 +5,10 @@
     <RootNamespace>Microsoft.AspNetCore.SignalR.Tests</RootNamespace>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="Microsoft.CSharp" Version="$(MicrosoftCSharpPackageVersion)" />
```
anurse: This looks suspicious. Accidentally added somehow?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/TestClient.cs
```diff
@@ -47,11 +48,24 @@ public TestClient(bool synchronousCallbacks = false, IHubProtocol protocol = nul
 
             _cts = new CancellationTokenSource();
 
-            using (var memoryStream = new MemoryStream())
+        }
+
+        public async Task<Task> ConnectAsync(dynamic endPoint, bool sendNegotiateMessage = true)
```
anurse: Can we make this `ConnectAsync<THub>(HubEndpoint<THub> endPoint, ...)` ? Though I have a vague feeling I may have tried this very refactoring in a spike and ended up swallowing my pride and using `dynamic` as well...

anurse: Ugh, I remember now, `HubEndPointTestUtils` returns `dynamic`... Ok, let's leave fixing that for another day :)

specs/HubProtocol.md
```diff
@@ -20,13 +20,17 @@ This document describes two encodings of the SignalR protocol: [JSON](http://www
...
-* `StreamItem` Message - Indicates individual items of streamed response data from a previous Invocation message.
-* `Completion` Message - Indicates a previous Invocation or StreamInvocation has completed. Contains an error if the invocation concluded with an error or the result of a non-streaming method invocation. The result will be absent for `void` methods. In case of streaming invocations no further `StreamItem` messages will be received
-* `CancelInvocation` Message - Sent by the client to cancel a streaming invocation on the server.
-* `Ping` Message - Sent by either party to check if the connection is active.
+| Message Name          | Sender         | Description                                                                                                                    |
+| ------------------    | -------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `NegotiationRequest`  | Caller         | Sent by the client to negotiate the message format.                                                                            |
```
anurse: Let's say `Client` here, since it does actually have to be the Client that does this.

specs/HubProtocol.md
```diff
@@ -20,13 +20,17 @@ This document describes two encodings of the SignalR protocol: [JSON](http://www
...
+| `NegotiationResponse` | Callee         | Sent by the server as an acknowledgment of the previous `NegotiationRequest` message. Contains an error if negotiation failed. |
```
anurse: Similarly, this is the `Server`.

specs/HubProtocol.md
```diff
@@ -20,13 +20,17 @@ This document describes two encodings of the SignalR protocol: [JSON](http://www
...
+| `Close`               | Callee         | Sent by the server when a connection is closed. Contains an error if the connection was closed because of an error.            |
```
anurse: This should be `Callee, Caller` since it could be either party that decides it has to close the connection.

Hey @davidfowl , this mean we kinda have our `/abort` endpoint from SignalR 2.x back, it's just in the SignalR layer, not the underlying transports. The client can send `Close` and the server can simply tear down the connection immediately, allowing cooperative clients to clean up Long Polling resources without having to wait for the poll interval to elapse.

specs/HubProtocol.md
```diff
@@ -4,8 +4,8 @@ The SignalR Protocol is a protocol for two-way RPC over any Message-based transp
 
 ## Terms
 
-* Caller - The node that is issuing an `Negotiation`, `Invocation`, `StreamInvocation`, `CancelInvocation` messages and receiving `Completion` and `StreamItem` messages (a node can be both Caller and Callee for different invocations simultaneously)
-* Callee - The node that is receiving an `Negotiation`, `Invocation`, `StreamInvocation`, `CancelInvocation` messages and issuing `Completion` and `StreamItem` messages (a node can be both Callee and Caller for different invocations simultaneously)
+* Caller - The node that is issuing an `Negotiation`, `Invocation`, `StreamInvocation`, `CancelInvocation`, `Ping` messages and receiving `Completion`, `StreamItem` and `Ping` messages (a node can be both Caller and Callee for different invocations simultaneously)
+* Callee - The node that is receiving an `Negotiation`, `Invocation`, `StreamInvocation`, `CancelInvocation`, `Ping` messages and issuing `Completion`, `StreamItem` and `Ping` messages (a node can be both Callee and Caller for different invocations simultaneously)
```
anurse: \*sigh\* these terms seemed much more useful when we had a simpler protocol. We could go back to Client/Server, but I can look at that in a separate PR.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -118,6 +119,15 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
             }
         }
 
+        private async Task WriteNegotiateResponseAsync(NegotiationResponseMessage message)
+        {
+            MemoryStream ms = new MemoryStream();
+            NegotiationProtocol.WriteResponseMessage(message, ms);
+
+            _connectionContext.Transport.Output.Write(ms.ToArray());
+            await _connectionContext.Transport.Output.FlushAsync();
```
davidfowl: You can call WriteAsync

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -170,22 +180,25 @@ public virtual void Abort()
...
 
                                     if (Features.Get<IConnectionInherentKeepAliveFeature>() == null)
                                     {
                                         // Only register KeepAlive after protocol negotiated otherwise KeepAliveTick could try to write without having a ProtocolReaderWriter
-                                        Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext)state).KeepAliveTick(), this);
+                                        Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext) state).KeepAliveTick(), this);
                                     }
 
+                                    Log.NegotiateComplete(_logger, Protocol.Name);
+                                    await WriteNegotiateResponseAsync(new NegotiationResponseMessage(null));
```
davidfowl: null?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: The behavior we're testing is that the client never sends us a Negotiate request right? Couldn't we just call `OnConnectedAsync` directly for that rather than using our `ConnectAsync` helper? We should get a response after the timeout elapses containing the error message, right?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -170,22 +180,25 @@ public virtual void Abort()
...
                                     return true;
                                 }
                             }
                             else if (result.IsCompleted)
                             {
-                                break;
+                                // connection was closed before we every received a response
+                                // can't send a negotiate response because there is no longer a connection
+                                Log.NegotiateFailed(_logger, null);
```
davidfowl: Don't we log this on the outside?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -606,7 +651,7 @@ public class HubEndpointTests
 
             using (var client = new TestClient())
             {
-                var endPointTask = endPoint.OnConnectedAsync(client.Connection);
+                Task endPointTask = await client.ConnectAsync(endPoint);
```
davidfowl: Help me understand all of these test changes.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
JamesNK: For the error. Because negotiate succeeds here then it is null. Want a static helper method to create NegotiateResponseMessage with error defaulting to null?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
JamesNK: No. TestClient takes dynamic for one of its parameters because the tests are creating hub from type and dynamic allows TestClient to still invoke OnConnected on the hub.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
JamesNK: `ConnectAsync` does 3 things:
1. Sends negotiate request message (previously done in TestClient ctor 💩 )
2. Calls `OnConnectedAsync`
3. Reads negotiate response message

test/Microsoft.AspNetCore.SignalR.Tests.Utils/TestClient.cs
JamesNK: If there was a non-generic base class or interface for `HubEndpoint<T>` then we could pass that around instead of `dynamic`.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
JamesNK: The problem I'm observing is the server doesn't wait for a request message and so doesn't timeout. It immediately accepts that there is no content for the connection and exits.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
JamesNK: There is nothing in `OnConnectedAsync` which is the only caller of this method. Where else would it be?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: That seems wrong, it should be looping until either the pipeline completes, or there's enough information in the buffer to parse a negotiate message. In the case that `ReadAsync` never returns, the token should be firing. @davidfowl thoughts?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
anurse: Oh yeah, and `dynamic` needs `Microsoft.CSharp`, but it's built in for the `netcoreapp` tfm.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
davidfowl: Where is the endpoint task? Do we still return it (unawaited?)

anurse: Yes, `ConnectAsync` is `Task<Task>` and the inner task is the endpoint task.

JamesNK: I'd never do it on a public API, it is too easy to not `await` and get the task result, but it works here for testing.

specs/HubProtocol.md
```diff
@@ -20,13 +20,17 @@ This document describes two encodings of the SignalR protocol: [JSON](http://www
...
-* `StreamItem` Message - Indicates individual items of streamed response data from a previous Invocation message.
-* `Completion` Message - Indicates a previous Invocation or StreamInvocation has completed. Contains an error if the invocation concluded with an error or the result of a non-streaming method invocation. The result will be absent for `void` methods. In case of streaming invocations no further `StreamItem` messages will be received
-* `CancelInvocation` Message - Sent by the client to cancel a streaming invocation on the server.
-* `Ping` Message - Sent by either party to check if the connection is active.
+| Message Name          | Sender         | Description                                                                                                                    |
+| ------------------    | -------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `NegotiationRequest`  | Client         | Sent by the client to negotiate the message format.                                                                            |
```
anurse: `s/Negotiation/Handshake/g`

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/CloseMessage.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
+{
+    public class CloseMessage : HubMessage
+    {
+        private static readonly CloseMessage _empty = new CloseMessage(null);
```
anurse: Could this just be a public field instead of using the method?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/CloseMessage.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public string Error { get; }
+
+        public CloseMessage(string error)
+        {
+            Error = error;
+        }
+
+        // Static factory methods. Don't want to use constructor overloading because it will break down
+        // if you need to send a payload statically-typed as a string. And because a static factory is clearer here
```
anurse: Looks like Copy-pasta from `CompletionMessage` ;)

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (!TryReadMessageIntoSingleSpan(buffer, out consumed, out examined, out var span))
+            {
+                responseMessage = null;
+                return false;
+            }
+
+            return TryParseResponseMessage(span, out responseMessage);
+        }
+
+        public static bool TryParseRequestMessage(ReadOnlySpan<byte> input, out HandshakeRequestMessage requestMessage)
```
anurse: Now that we have a `TryReadMessageIntoSingleSpan` method, I think we can actually just inline this into the version that takes `ReadOnlySequence<byte>`. Eventually we'll be building a `Stream` over `ReadOnlySequence` and passing that directly into `JsonTextReader` anyway.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeResponseMessage.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public string Error { get; }
+
+        public HandshakeResponseMessage(string error)
+        {
+            Error = error;
+        }
+
+        // Static factory methods. Don't want to use constructor overloading because it will break down
+        // if you need to send a payload statically-typed as a string. And because a static factory is clearer here
```
anurse: More comment copy-pasta

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeResponseMessage.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
+{
+    public class HandshakeResponseMessage : HubMessage
+    {
+        private static readonly HandshakeResponseMessage _empty = new HandshakeResponseMessage(null);
```
anurse: As with `CloseMessage`, could this just be a public field? That's what we've done in other situations like this.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeResponseMessage.cs
anurse: Also, I don't think we really need static factory methods for this one (or for `CloseMessage`). Especially if we have an `Empty` instance field.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/HandshakeProtocolTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Text;
+using Microsoft.AspNetCore.SignalR.Internal.Protocol;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Common.Tests.Internal.Protocol
+{
+    public class HandshakeProtocolTests
+    {
+        [Fact]
+        public void CanRoundtripHandshakeRequest()
```
anurse: I know the old tests for this did the same thing, but I'm not a fan of "roundtrip" tests for our protocols since we have a JS client as well (and possibly other clients in the future). Can we make these proper "given this specific JSON I get this parsed object" and "given this specific object I get this JSON" tests rather than using roundtripping? There's not much variety here, so it should be pretty straightforward.

Round-tripping tests don't break if we break the protocol in both the reader and the writer.

specs/HubProtocol.md
```diff
@@ -20,13 +20,17 @@ This document describes two encodings of the SignalR protocol: [JSON](http://www
...
+| `NegotiationResponse` | Server         | Sent by the server as an acknowledgment of the previous `NegotiationRequest` message. Contains an error if negotiation failed. |
+| `Close`               | Callee, Caller | Sent by the server when a connection is closed. Contains an error if the connection was closed because of an error.            |
```
anurse: We'll also need to add information to the encoding specifics further down the spec for the new message types.

BrennanConroy: Close is Callee and Caller?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
```diff
@@ -0,0 +1,136 @@
...
+                }
+                writer.WriteEndObject();
+            }
+
+            TextMessageFormatter.WriteRecordSeparator(output);
+        }
+
+        private static JsonTextWriter CreateJsonTextWriter(Stream output)
+        {
+            return new JsonTextWriter(new StreamWriter(output, _utf8NoBom, 1024, leaveOpen: true));
```
BrennanConroy: nit: can we make it smaller than 1024?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -118,6 +119,14 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
             }
         }
 
+        private async Task WriteHandshakeResponseAsync(HandshakeResponseMessage message)
+        {
+            MemoryStream ms = new MemoryStream();
```
BrennanConroy: `var` and `using`

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -170,22 +179,25 @@ public virtual void Abort()
...
-                                    Log.UsingHubProtocol(_logger, Protocol.Name);
-
                                     UserIdentifier = userIdProvider.GetUserId(this);
 
                                     if (Features.Get<IConnectionInherentKeepAliveFeature>() == null)
                                     {
-                                        // Only register KeepAlive after protocol negotiated otherwise KeepAliveTick could try to write without having a ProtocolReaderWriter
-                                        Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext)state).KeepAliveTick(), this);
+                                        // Only register KeepAlive after protocol handshake otherwise KeepAliveTick could try to write without having a ProtocolReaderWriter
+                                        Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext) state).KeepAliveTick(), this);
```
BrennanConroy: nit: revert

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -67,6 +67,12 @@ public class HubEndPoint<THub> where THub : Hub
             {
                 await _lifetimeManager.OnConnectedAsync(connectionContext);
                 await RunHubAsync(connectionContext);
+
+                await connectionContext.WriteAsync(CloseMessage.Empty());
```
BrennanConroy: Do these sends actually get sent successfully? Because at the end of `RunHubAsync` we try to abort the connection https://github.com/aspnet/SignalR/pull/1591/files#diff-3f9cc189cbcb6b225bb8d6cb6ecdeb21R117

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -346,6 +357,20 @@ private void WriteCancelInvocationMessage(CancelInvocationMessage message, Packe
             packer.PackString(message.InvocationId);
         }
 
+        private void WriteCloseMessage(CloseMessage message, Packer packer)
+        {
+            packer.PackArrayHeader(3);
```
BrennanConroy: This should be 2 right?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -36,7 +37,7 @@ public class HubEndpointTests
...
```
BrennanConroy: Does var not work here?

specs/HubProtocol.md
anurse: Correct. It can be sent by either node. Since there's no invocation context, it would probably be more accurate to say "Client, Server". This is one of many reasons I'm no longer in love with "Caller/Callee" :)

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
JamesNK: My VS has been broken on this branch so I haven't tested yet.

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
JamesNK: Thanks. I didn't know what it did

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
anurse: `dynamic` is just the worst.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
anurse: These streams are going to get a bit of an overhaul when we get to #1508 . I wouldn't optimize too heavily here until we know how that's going to land.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -162,30 +178,35 @@ public virtual void Abort()
...
 
+                                    Log.HandshakeComplete(_logger, Protocol.Name);
+                                    await WriteHandshakeResponseAsync(HandshakeResponseMessage.Empty);
                                     return true;
                                 }
                             }
                             else if (result.IsCompleted)
                             {
-                                break;
+                                // connection was closed before we every received a response
```
anurse: nit: "ever received"

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
```diff
@@ -4,7 +4,14 @@
     <TargetFrameworks>$(StandardTestTfms)</TargetFrameworks>
     <RootNamespace>Microsoft.AspNetCore.SignalR.Tests</RootNamespace>
   </PropertyGroup>
-  
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="Microsoft.CSharp" Version="$(MicrosoftCSharpPackageVersion)" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
```
anurse: I think these conditions can be grouped. Also, if we need it for `netcoreapp2.0` and `net461` I'd be surprised if we don't need it for `netcoreapp2.1` so the condition probably isn't needed at all.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -1769,34 +1840,86 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
                 // Shut down
                 client.Dispose();
 
-                await endPointLifetime.OrTimeout();
+                await endPointTask.OrTimeout();
 
                 client.Connection.Transport.Output.Complete();
 
                 // We should have all pings
```
anurse: and now, a close message too! :)

specs/HubProtocol.md
```diff
@@ -42,6 +46,18 @@ Example:
...
+* `error` - the optional error message if the server does not support the request protocol
+
+Example:
+
+```json
+{
+    "error": "Requested protocol 'messagepack' is not available."
+}
+```
+
```
anurse: We'll need to get the Close frame encodings in both JSON and MsgPack in here as well. It's a bit of a pain, but we should keep this spec accurate.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
JamesNK: 1024 is the StreamWriter default. Lets leave it for now and come back when analyzing perf

specs/HubProtocol.md
JamesNK: Done

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
JamesNK: Figured this test out

JamesNK:
I haven't been able to figure this one out 😒 

---

`HubConnectionTests.CheckHttpConnectionFeatures` also fail and I'm not sure why. There are 3 tests because of the parameters and if you run them one at a time and they each pass, but if you run them together then the first passes and the second and third always fail with a timeout.

JamesNK:
I'm going to need to catch up about what needs to happen in clients when getting handshake response messages and close messages.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
davidfowl: Yes, I want to get to these changes too but not as part of this PR.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -131,7 +140,7 @@ public virtual void Abort()
             Task.Factory.StartNew(_abortedCallback, this);
         }
 
-        internal async Task<bool> NegotiateAsync(TimeSpan timeout, IList<string> supportedProtocols, IHubProtocolResolver protocolResolver, IUserIdProvider userIdProvider)
+        internal async Task<bool> HandshakeAsync(TimeSpan timeout, IList<string> supportedProtocols, IHubProtocolResolver protocolResolver, IUserIdProvider userIdProvider)
```
davidfowl: ShakeHandsAsync? I'm kidding :trollface:

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -126,6 +132,22 @@ public class HubEndPoint<THub> where THub : Hub
             }
         }
 
+        private async Task SendCloseAsync(HubConnectionContext connection, Exception exception)
+        {
+            CloseMessage closeMessage = exception == null
+                ? CloseMessage.Empty
+                : new CloseMessage($"Connection closed with an error. {exception.GetType().Name}: {exception.Message}");
```
davidfowl: We need to be careful sending exception details from the server to the client by default.

BrennanConroy: https://github.com/aspnet/SignalR/issues/1614

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -635,10 +635,14 @@ public class HubEndpointTests
...
                 client.Dispose();
 
-                // Nothing should have been written
-                bool checkForContent = client.Connection.Application.Input.TryRead(out var buffer);
-
-                Assert.False(checkForContent, "Expected no content but got: " + Encoding.UTF8.GetString(buffer.Buffer.ToArray()));
+                // this test will differ slightly depending on the speed of the computer it is run on
+                // may or may not receive a close message so wait a second to check before exiting
+                var readTask = client.ReadAsync();
+                if ((await Task.WhenAny(readTask, Task.Delay(TimeSpan.FromSeconds(1)))) == readTask)
```
davidfowl: 😢

JamesNK: False alarm, it is consistent. The test pre-PR was incorrect because it didn't await at all to check for content.

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionMetadata.cs
```diff
@@ -3,14 +3,15 @@
...
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Sockets
 {
     internal class ConnectionMetadata : IDictionary<object, object>
     {
         public ConnectionMetadata()
-            : this(new Dictionary<object, object>())
+            : this(new ConcurrentDictionary<object, object>())
```
JamesNK: Change is because of https://github.com/aspnet/SignalR/issues/1636

Functional tests were sporadically failing because of the error there

JamesNK:
Functionally complete! 😅

client-ts/signalr/src/HubConnection.ts
```diff
@@ -63,36 +64,61 @@ export class HubConnection {
...
         this.configureTimeout();
     }
 
+    private processHandshakeResponse(data: any) {
+        if (data instanceof ArrayBuffer) {
+            // Format is binary but still need to read JSON text from handshake response
+            data = String.fromCharCode.apply(null, new Uint8Array(data));
+        }
+        const messages = TextMessageFormat.parse(data);
+        const responseMessage: HandshakeResponseMessage = JSON.parse(messages[0]);
```
BrennanConroy: Need to handle multiple messages like we do in C# client

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -570,6 +583,12 @@ private object[] BindArguments(JsonTextReader reader, IReadOnlyList<Type> paramT
             throw new JsonReaderException("Unexpected end when reading JSON");
         }
 
+        private CloseMessage BindCloseMessage(string error)
+        {
+            var message = new CloseMessage(error);
```
BrennanConroy: If error is empty we can return the `CloseMessage.Empty` instance

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -102,6 +105,9 @@ public class HubEndPoint<THub> where THub : Hub
 
         private async Task HubOnDisconnectedAsync(HubConnectionContext connection, Exception exception)
         {
+            // send close message before aborting the connection
+            await SendCloseAsync(connection, exception);
```
davidfowl: Do we always send a close message?

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -33,7 +33,7 @@ public HttpConnectionDispatcher(ConnectionManager manager, ILoggerFactory logger
             _logger = _loggerFactory.CreateLogger<HttpConnectionDispatcher>();
         }
 
-        public async Task ExecuteAsync(HttpContext context, HttpSocketOptions options, ConnectionDelegate ConnectionDelegate)
+        public async Task ExecuteAsync(HttpContext context, HttpSocketOptions options, ConnectionDelegate connectionDelegate)
```
davidfowl: Woah, who Pascal cased this thing...

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
```diff
@@ -37,7 +37,20 @@ public class TestHub : Hub
 
         public IEnumerable<string> GetHeaderValues(string[] headerNames)
         {
-            var headers = Context.Connection.GetHttpContext().Request.Headers;
+            var context = Context.Connection.GetHttpContext();
+
+            if (context == null)
+            {
+                throw new InvalidOperationException("Unable to get HttpContext from request.");
```
davidfowl: Why would this ever happen in our tests?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
```diff
@@ -37,7 +37,20 @@ public class TestHub : Hub
...
+            }
+
+            var headers = context.Request.Headers;
+
+            if (headers == null)
+            {
+                throw new InvalidOperationException("Unable to get headers from context.");
```
davidfowl: When would this ever happen? Did you make a mock?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
```diff
@@ -4,7 +4,11 @@
     <TargetFrameworks>$(StandardTestTfms)</TargetFrameworks>
     <RootNamespace>Microsoft.AspNetCore.SignalR.Tests</RootNamespace>
   </PropertyGroup>
-  
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="Microsoft.CSharp" Version="$(MicrosoftCSharpPackageVersion)" />
```
davidfowl: dynamic!

test/Microsoft.AspNetCore.SignalR.Tests.Utils/TestClient.cs
```diff
@@ -20,11 +21,12 @@ public class TestClient : IDisposable
...
         private readonly IHubProtocol _protocol;
         private readonly IInvocationBinder _invocationBinder;
-        private CancellationTokenSource _cts;
-        private Queue<HubMessage> _messages = new Queue<HubMessage>();
+        private readonly CancellationTokenSource _cts;
+        private readonly Queue<HubMessage> _messages = new Queue<HubMessage>();
 
         public DefaultConnectionContext Connection { get; }
         public Task Connected => ((TaskCompletionSource<bool>)Connection.Items["ConnectedTask"]).Task;
+        public HandshakeResponseMessage HandshakeResponseMessage { get; private set; }
```
davidfowl: You could return `Task<Task<HandshakeResponseMessage>>` from ConnectAsync

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
JamesNK: Because I can't figure out why this test fails. Only long polling, only on OSX https://travis-ci.org/aspnet/SignalR/jobs/355210173#L1309 😢

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
davidfowl: Yea, we can fix that by making HubEndPoint\<T\> derive from  EndPoint

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Hubs.cs
JamesNK: https://travis-ci.org/aspnet/SignalR/jobs/355210173#L1309

You tell me?! 😕

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -358,8 +392,47 @@ private Task InvokeStreamCore(string methodName, InvocationRequest irq, object[]
...
+                    throw new InvalidDataException("Unable to parse payload as a handshake response message.");
+                }
+
+                message = HandshakeProtocol.ParseResponseMessage(payload.Span);
+            }
+            catch (Exception ex)
+            {
+                // shutdown if we're unable to read handshake
+                Log.ErrorReceivingHandshakeResponse(_logger, ex);
+                Shutdown(ex);
```
davidfowl: What state are we in when we call this from the OnDataReceived callback? Does this fire the closed event inline?

client-ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -345,20 +345,14 @@ describe("hubConnection", () => {
...
-
                 const hubConnection = new HubConnection("http://" + document.location.host + "/uncreatable", {
                     logger: LogLevel.Trace,
                     protocol,
                     transport: transportType,
                 });
 
                 hubConnection.onclose((error) => {
-                    expect(error.message).toMatch(errorRegex[TransportType[transportType]]);
+                    expect(error.message).toEqual("Server returned an error on close: Connection closed with an error. InvalidOperationException: Unable to resolve service for type 'System.Object' while attempting to activate 'FunctionalTests.UncreatableHub'.");
```
anurse: :tada:

JamesNK: Don't celebrate too soon. This works on my computer but on the build server the message is sometimes the old error. I'll need to talk to you about it.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
JamesNK: Yes please. Or a non-generic abstract base class for HubEndPoint. All of the abstract members on it don't use the generic type.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
JamesNK: The spec of the close message is to always send a message on close. An error message tells the client if the connection closed from the server because of an error. Talk with @anurse if you don't like it.

JamesNK:
Tests passing. Last call for reviews

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
```diff
@@ -83,7 +83,11 @@ public class HubEndPoint<THub> where THub : Hub
             catch (Exception ex)
             {
                 Log.ErrorDispatchingHubEvent(_logger, "OnConnectedAsync", ex);
-                throw;
+
+                await SendCloseAsync(connection, ex);
+
+                // return instead of throw to let close message send successfully
```
anurse: FYI @davidfowl this is needed because if we rethrow, it ends up completing the transport pipe with the exception. The Pipe then throws from `ReadAsync` and we can't drain remaining data (i.e. the close frame)

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
anurse: It's generally good form to send a close frame when a graceful disconnect occurs. It tells the client that the server intended to close the connection, rather than a mid-stream proxy killing it. For example, our client auto-reconnect logic should **not** attempt to reconnect if it sees a close frame with no error.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -283,8 +283,8 @@ public static IEnumerable<object[]> MessageSizesData
                     logger.LogInformation("Sent message", bytes.Length);
 
                     logger.LogInformation("Receiving message");
-                    // No timeout here because it can take a while to receive all the bytes
-                    var receivedData = await receiveTcs.Task;
+                    // Big timeout here because it can take a while to receive all the bytes
```
anurse: 👍 👍 👍

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: We should consider `Task.Run`ing the Closed event or something, but I don't think that's a problem isolated to this PR.

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
davidfowl: This just means there’s a hang still...

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: Are we usually running user code here?

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
JamesNK: I just changed it so that you know there is a problem after 1 minute instead of hanging the CI server until it kills the entire build.

davidfowl: yeap

client-ts/signalr/spec/HubConnection.spec.ts
```diff
@@ -562,17 +724,53 @@ class TestConnection implements IConnection {
...
+    public receiveHandshakeResponse(error?: string): void {
+        this.receive({error: error});
+    }
+
     public receive(data: any): void {
         const payload = JSON.stringify(data);
         this.onreceive(TextMessageFormat.write(payload));
     }
 
+    public receiveText(data: string) {
```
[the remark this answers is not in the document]
JamesNK: I like it being explicit. Makes the tests a touch easier to read.

src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs
davidfowl: Something about these changes to the HubEndPoint rub me the wrong way. I hate that we send the close message from so many different places. I really want this logic to be in a single place and I want various pieces of code to communicate via an exception. We have too many ways to send the close message right now.

davidfowl: Oh wait, the handshake response is different from the close message...

client-ts/signalr/src/HubConnection.ts
```diff
@@ -63,36 +64,100 @@ export class HubConnection {
...
+                        this.logger.log(LogLevel.Warning, "Invalid message type: " + data);
+                        break;
+                }
             }
         }
 
         this.configureTimeout();
     }
 
+    private processHandshakeResponse(data: any): any {
```
davidfowl: ZOMG I didn't see this change...

PR #1596, opened 2018-03-14
Further making Microsoft.AspNetCore.Sockets "not a thing"

davidfowl:
- This PR attempts to move things where they are needed instead of where they
happened to be used. As a result we should now have Sockets.Abstractions and
Sockets down to the minimal set of things required to make them run.
Sockets.Abstractions should go away in favor of Protocol.Abstractions and
Sockets contains the EndPoint abstraction and related types.
- Moved ConnectionManager, DefaultConnection context and friends to
Sockets.Http.
- Added a TestConnectionContext for unit tests and bench marking

@anurse @DamianEdwards It's looking like we should decide to kill `EndPoint` as a concept or move it down into Protocol.Abstractions. The only reason I didn't do that in the past was because I wanted to remove the DI dependency (which isn't really required if we don't care about an AddEndPoint method).


test/Microsoft.AspNetCore.SignalR.Tests.Utils/TestConnectionContext.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Protocols;
+using Microsoft.AspNetCore.Protocols.Features;
+
+namespace Microsoft.AspNetCore.Sockets
+{
+    public class TestConnectionContext : ConnectionContext,
```
anurse: So this is just a copy of `HttpConnectionContext` renamed?

davidfowl: And with the ValueStopwatchTimer removed.

anurse: But is that worth having a near duplicate set of code used for testing?

davidfowl: I can move this to common but I don't want it to be a thing. Each "dispatcher" should have their own implementation.

PR #1598, opened 2018-03-14
Copy the InvocationHandler if list changed not when invoked

FlorianGrimm:


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -438,14 +432,7 @@ private void Shutdown(Exception exception = null)
...
 
-            // TODO: Optimize this!
-            // Copying the callbacks to avoid concurrency issues
-            InvocationHandler[] copiedHandlers;
-            lock (handlers)
-            {
-                copiedHandlers = new InvocationHandler[handlers.Count];
-                handlers.CopyTo(copiedHandlers);
-            }
+            InvocationHandler[] copiedHandlers = handlers.CopiedHandlers;
```
anurse: Our code style is to use `var` in all possible cases.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -622,6 +606,39 @@ public Task InvokeAsync(object[] parameters)
...
         }
 
+        private class InvocationHandlerList
+        {
+            private readonly List<InvocationHandler> _invocationHandlers;
+            internal InvocationHandler[] CopiedHandlers;
+
+            internal InvocationHandlerList(InvocationHandler handler)
+            {
+                this._invocationHandlers = new List<InvocationHandler>() { handler};
```
anurse: Remove `this.` prefix.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -594,9 +578,9 @@ public IReadOnlyList<Type> GetParameterTypes(string methodName)
                 // We use the parameter types of the first handler
                 lock (handlers)
                 {
-                    if (handlers.Count > 0)
+                    if (handlers.CopiedHandlers.Length > 0)
```
anurse: Put this code (including the `lock` above) in a method on `InvocationHandlerList` and make sure it locks on `_invocationHandlers`. Because it locks on `handlers` here, it will actually be locking on a **different** object than `.Add` and `.Remove` (which lock on `handlers._invocationHandlers`), so it's not thread-safe

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -622,6 +606,39 @@ public Task InvokeAsync(object[] parameters)
...
+                return this;
+            }
+
+            internal void Remove(InvocationHandler handler)
+            {
+                lock (_invocationHandlers)
+                {
+                    if (_invocationHandlers.Remove(handler))
+                    {
+                        this.CopiedHandlers = _invocationHandlers.ToArray();
```
anurse: Remove `this.` prefix.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Also, a space is needed after `handler` and before `}`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -622,6 +606,39 @@ public Task InvokeAsync(object[] parameters)
...
 
+        private class InvocationHandlerList
+        {
+            private readonly List<InvocationHandler> _invocationHandlers;
+            internal InvocationHandler[] CopiedHandlers;
+
+            internal InvocationHandlerList(InvocationHandler handler)
+            {
+                this._invocationHandlers = new List<InvocationHandler>() { handler};
+                CopiedHandlers = _invocationHandlers.ToArray();
```
BrennanConroy: We should lazy initialize this

FlorianGrimm: if we don't do this lazy, we dont have to lock

anurse: The list isn't generally modified after you start invoking handlers (since generally the `.On` calls all occur at startup, before `StartAsync` is called). It'll be rare to have contention on this lock since the collection isn't usually modified once the `CopiedHandlers` value is requested.

Also, it strikes me that the 90% use-case for this is to have only a single handler. We could probably avoid the allocation of the list altogether by having a field on this list represent the "first" item, and only allocate the list when a second handler is added. That way, in the 90% case, we don't have an extra list hanging around.

FlorianGrimm:
sorry was my first rebase

FlorianGrimm:
i don't know what happend. Git strikes me down. should i close this?

davidfowl:
@FlorianGrimm yes, close and reopen a PR on top of dev.

davidfowl:
Closing this out

PR #1599, opened 2018-03-14
Further making Microsoft.AspNetCore.Sockets "not a thing"

davidfowl:
- This PR attempts to move things where they are needed instead of where they
happened to be used. As a result we should now have Sockets.Abstractions and
Sockets down to the minimal set of things required to make them run.
Sockets.Abstractions should go away in favor of Protocol.Abstractions and
Sockets contains the EndPoint abstraction and related types.
- Moved ConnectionManager and friends to
Sockets.Http.
-Removed Sockets and moved everything into Sockets.Abstractions.
- Moved DefaultConnection and put it in Sockets.Abstractions.


davidfowl:
🆙 📅

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -22,7 +22,7 @@ public class ConnectionManager
         // TODO: Consider making this configurable? At least for testing?
         private static readonly TimeSpan _heartbeatTickRate = TimeSpan.FromSeconds(1);
 
-        private readonly ConcurrentDictionary<string, DefaultConnectionContext> _connections = new ConcurrentDictionary<string, DefaultConnectionContext>();
+        private readonly ConcurrentDictionary<string, (DefaultConnectionContext Connection, ValueStopwatch Timer)> _connections = new ConcurrentDictionary<string, (DefaultConnectionContext Connection, ValueStopwatch Timer)>();
```
anurse: 🎉

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -53,7 +53,14 @@ public void Start()
 
         public bool TryGetConnection(string id, out DefaultConnectionContext connection)
         {
-            return _connections.TryGetValue(id, out connection);
+            connection = null;
+
+            if (_connections.TryGetValue(id, out var pair))
```
anurse: Would `return _connections.TryGetValue(id, out (connection, _));` work? Destructure all the things?!

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -78,10 +84,10 @@ public DefaultConnectionContext CreateConnection()
 
         public void RemoveConnection(string id)
         {
-            if (_connections.TryRemove(id, out var connection))
+            if (_connections.TryRemove(id, out var pair))
```
anurse: As above, `if (_connections.TryRemove(id, out (_, timer)))`?

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -128,33 +134,34 @@ public void Scan()
                 {
                     var status = DefaultConnectionContext.ConnectionStatus.Inactive;
                     var lastSeenUtc = DateTimeOffset.UtcNow;
+                    var connection = c.Value.Connection;
```
anurse: Destructure in the for loop! `foreach (var (connection, _) in _connections)`

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
davidfowl: why?

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
davidfowl: Ugh,

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
davidfowl: ....

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionBuilderExtensions.cs
```diff
@@ -1,16 +1,21 @@
...
-using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.Sockets
 {
     public static class ConnectionBuilderExtensions
     {
         public static IConnectionBuilder UseEndPoint<TEndPoint>(this IConnectionBuilder connectionBuilder) where TEndPoint : EndPoint
         {
-            var endpoint = connectionBuilder.ApplicationServices.GetRequiredService<TEndPoint>();
+            var endpoint = (TEndPoint)connectionBuilder.ApplicationServices.GetService(typeof(TEndPoint));
```
JamesNK: Where is no generic `GetService`?

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionBuilderExtensions.cs
```diff
@@ -1,16 +1,21 @@
...
+
+            if (endpoint == null)
+            {
+                throw new InvalidOperationException($"{nameof(EndPoint)} type {typeof(TEndPoint)} is not registered.");
```
JamesNK: Is there a test that throws this error and checks for this error message?

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
JamesNK: Destructure looks sweet but IMO not as easy to parse as a reader.

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionBuilderExtensions.cs
davidfowl: I didn't want to add a dependency on dependency injection 😄

src/Microsoft.AspNetCore.Sockets.Abstractions/ConnectionBuilderExtensions.cs
davidfowl: No. But now we're debating removing end points altogether. I think I might just do that..

davidfowl: This is a bigger change than I'd want to make in this PR.

PR #1600, opened 2018-03-14
fix the umd build of msgpack

anurse:
This was breaking the SocketsSample. Unfortunately, the UMD build isn't what the functional tests test, so it wasn't caught by them.


anurse:
I verified that this fixes the sample.

JamesNK:
:shipit:

PR #1601, opened 2018-03-14
[WIP] Use CancellationToken on async write and read calls

JamesNK:
To start with I've placed cancellation token everywhere I could see async methods that took a token and no token was being used.

Review and say where it is not appropriate, or a better test can be made (i.e. checking whether cancellation has been required and exiting instead of waiting for a callee to throw).

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class HubConnectionContext
     {
-        private static Action<object> _abortedCallback = AbortConnection;
+        private static readonly Action<object> _abortedCallback = AbortConnection;
 
         private readonly ConnectionContext _connectionContext;
         private readonly ILogger _logger;
@@ -75,7 +75,7 @@ namespace Microsoft.AspNetCore.SignalR
 
         public virtual async Task WriteAsync(HubMessage message)
         {
-            await _writeLock.WaitAsync();
+            await _writeLock.WaitAsync(ConnectionAbortedToken);
 
             try
             {
@@ -86,7 +86,7 @@ namespace Microsoft.AspNetCore.SignalR
 
                 Interlocked.Exchange(ref _lastSendTimestamp, Stopwatch.GetTimestamp());
 
-                await _connectionContext.Transport.Output.FlushAsync();
+                await _connectionContext.Transport.Output.FlushAsync(ConnectionAbortedToken);
             }
             finally
             {
@@ -110,7 +110,7 @@ namespace Microsoft.AspNetCore.SignalR
 
                 Interlocked.Exchange(ref _lastSendTimestamp, Stopwatch.GetTimestamp());
 
-                await _connectionContext.Transport.Output.FlushAsync();
+                await _connectionContext.Transport.Output.FlushAsync(ConnectionAbortedToken);
             }
             finally
             {
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -75,6 +75,11 @@ namespace Microsoft.AspNetCore.SignalR
 
         public virtual async Task WriteAsync(HubMessage message)
         {
+            if (IsConnectionAborted(message))
+            {
+                return;
+            }
+
             await _writeLock.WaitAsync(ConnectionAbortedToken);
 
             try
@@ -96,6 +101,11 @@ namespace Microsoft.AspNetCore.SignalR
 
         private async Task TryWritePingAsync()
         {
+            if (IsConnectionAborted(PingMessage.Instance))
+            {
+                return;
+            }
+
             // Don't wait for the lock, if it returns false that means someone wrote to the connection
             // and we don't need to send a ping anymore
             if (!await _writeLock.WaitAsync(0, ConnectionAbortedToken))
@@ -254,6 +264,18 @@ namespace Microsoft.AspNetCore.SignalR
             }
         }
 
+        private bool IsConnectionAborted(HubMessage message)
+        {
+            if (ConnectionAbortedToken.IsCancellationRequested)
+            {
+                // log that the message was not able to be sent because the connection was aborted
+                Log.ConnectionAborted(_logger, message.GetType().Name);
+                return true;
+            }
+
+            return false;
+        }
+
         private static class Log
         {
             // Category: HubConnectionContext
@@ -268,6 +290,9 @@ namespace Microsoft.AspNetCore.SignalR
 
             private static readonly Action<ILogger, Exception> _transportBufferFull =
                 LoggerMessage.Define(LogLevel.Debug, new EventId(4, "TransportBufferFull"), "Unable to send Ping message to client, the transport buffer is full.");
+
+            private static readonly Action<ILogger, string, Exception> _connectionAborted =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(5, "ConnectionAborted"), "Unable to write message '{messageType}', the connection has been aborted.");
 
             public static void UsingHubProtocol(ILogger logger, string hubProtocol)
             {
@@ -288,7 +313,11 @@ namespace Microsoft.AspNetCore.SignalR
             {
                 _transportBufferFull(logger, null);
             }
-        }
 
+            public static void ConnectionAborted(ILogger logger, string messageType)
+            {
+                _connectionAborted(logger, messageType, null);
+            }
+        }
     }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -98,7 +98,7 @@ namespace Microsoft.AspNetCore.SignalR
         {
             // Don't wait for the lock, if it returns false that means someone wrote to the connection
             // and we don't need to send a ping anymore
-            if (!await _writeLock.WaitAsync(0))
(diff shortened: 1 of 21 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -75,7 +75,7 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
 
         public virtual async Task WriteAsync(HubMessage message)
         {
-            await _writeLock.WaitAsync();
+            await _writeLock.WaitAsync(ConnectionAbortedToken);
```
JamesNK: If a connection is aborted inside a hub method then an error will be thrown which will trigger a `CompleteMessage` which will hit this line and then error because the connection is aborted. Check whether token has been cancelled and exit the method first?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -86,7 +86,7 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
 
                 Interlocked.Exchange(ref _lastSendTimestamp, Stopwatch.GetTimestamp());
 
-                await _connectionContext.Transport.Output.FlushAsync();
+                await _connectionContext.Transport.Output.FlushAsync(ConnectionAbortedToken);
```
JamesNK: On `HubConnectionContext` there is a virtual `ConnectionAbortedToken` token, but also a private `_connectionAbortedTokenSource` that is used in multiple places. What is `ConnectionAbortedToken` for and why is it virtual?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -217,7 +217,7 @@ public override Task SendUserAsync(string userId, string methodName, object[] ar
             using (var writer = new JsonTextWriter(new StreamWriter(stream)))
             {
                 _serializer.Serialize(writer, message);
-                await writer.FlushAsync();
+                writer.Flush();
```
JamesNK: No point to `FlushAsync` to a `MemoryStream`.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -64,11 +64,11 @@ public WebSocketsTransport(WebSocketOptions options, IDuplexPipe application, De
             }
         }
 
-        public async Task ProcessSocketAsync(WebSocket socket)
+        public async Task ProcessSocketAsync(WebSocket socket, CancellationToken cancellationToken)
```
JamesNK: Should this be public? It is only used by this class except for unit tests.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
JamesNK: Or should aborting the connection still allow a complete and close message messages to go through?

BrennanConroy: If the connection is aborted we can't send to it, so we should try not to send a `CompletionMessage`

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
```diff
@@ -40,7 +40,7 @@ public WebSocketsTests(ITestOutputHelper output)
                     var ws = new WebSocketsTransport(new WebSocketOptions(), connection.Application, connectionContext, loggerFactory);
 
                     // Give the server socket to the transport and run it
-                    var transport = ws.ProcessSocketAsync(await feature.AcceptAsync());
+                    var transport = ws.ProcessSocketAsync(await feature.AcceptAsync(), CancellationToken.None);
```
anurse: We've generally been using the new [`default` literals](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-1#default-literal-expressions) for this, assuming it works in this position. I do like making it explicit though, so I'm for the change in general :)

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
anurse: It's because of the tests that it's public, since it's "pubternal". I'm fine with making it `InternalsVisibleTo` but @davidfowl might complain (not necessarily something to be avoided ;)).

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
JamesNK: If it is a true public API then we should default it, but if it is internal then not. Since this will become internal then I'll leave it without default.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
JamesNK: I also like `InternalsVisibleTo` over making things public for testing

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
anurse: Nah, I just meant using `default` in the caller. It's just a small style thing to be consistent with how we use do it elsewhere.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -336,8 +336,17 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
                     await channel.WaitToReadAsync().AsTask().OrTimeout();
                     cts.Cancel();
 
-                    var results = await channel.ReadAllAsync().OrTimeout();
+                    List<int> results = new List<int>();
+                    try
+                    {
+                        await channel.ReadAllIntoListAsync(results).OrTimeout();
+                    }
+                    catch (OperationCanceledException)
```
JamesNK: This fixes the unit test but I don't know if it is correct. Is reading from a cancelled channel throwing a `OperationCanceledException` correct?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -68,7 +68,7 @@ public class HubEndpointTests
             }
         }
 
-        [Fact]
+        [Fact(Skip = "Timing out")]
```
JamesNK: I'm going to need a hand with this one. In `AbortFromHubMethodForcesClientDisconnect` the server connection is aborting but the client keeps listening forever.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
JamesNK: Attempting to send a message after cancel is what causes a `OperationCanceledException` when the connected has been aborted.

It is a simple matter to check the cancellation token and exit from `WriteAsync(HubMessage message)` if the connection has been aborted. Thoughts @davidfowl @anurse

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
JamesNK: Thoughts on this @anurse @davidfowl

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -338,9 +337,11 @@ private Task StartAsyncCore(TransferFormat transferFormat)
                         Log.ErrorDuringClosedEvent(_logger, ex);
                     }
 
+                    // signal that close has completed
+                    _closeTcs.SetResult(null);
```
JamesNK: The `Closed` event was running after this close TCS was being set as complete, which was causing `HubConnectionTests.CanStopAndStartConnection` to fail because a connection was being started up again while it was still shutting down by `Closed`.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
anurse: Yeah, generally if it's only for testing, we also lean to `InternalsVisibleTo`. Pubternal is supposed to be for those things that the super advanced user might actually want to tweak. Let's `IVT` it up.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
anurse: My understanding is yes, it is. A channel can be completed with an exception and it's expected that the channel will throw that exception after reading the last item written prior to completion.

anurse: Looks like `ReadAllIntoListAsync` explicitly manifests the `Completion` task's error, which would be expected to yield this exception.

test/Microsoft.AspNetCore.Sockets.Tests/WebSocketsTests.cs
anurse: Meh, I don't really care either way though ;). This is the place where that may be a little odd since it's a little like passing in `null`. We normally used named arguments when passing in uninformative literals like `null`, so this would be `cancellationToken: default`, at which point it might as well be `CancellationToken.None` :)

anurse:
@davidfowl So do you want to close this and approach it in RC? You told me to close the issue in triage sooo...?

Issue #1582, opened 2018-03-13
Reduce memory usage of idle websockets with bufferless reads


davidfowl:
A common technique used to reduce memory usage of idle websockets is to issue a read with 0 bytes then issue the actual read with a buffer when data is ready. We should do this and measure the memory and throughput differences.

We may also need to make adjustments to the websocket middleware so that we can accept the websocket and configure the minimum buffer size (internally buffered by the websocket).

davidfowl:
Example here https://github.com/davidfowl/PlatformLevelTechempower/blob/944597f9ad196d3a333b837fb8c62a7304d5f9d2/ServerWithKestrel2.1/ChatConnection.cs#L15-L45

anurse:
Blocked on https://github.com/dotnet/corefx/issues/28138

There's already a PR, we just want the corefx fix to be able to apply it to the send buffer too.

davidfowl:
We may have to punt this to RC and make the fix to corefx so receives aren't super inefficient.

davidfowl:
We should be unblocked now, @BrennanConroy before you commit this change I'd like to get a bare bones test of some number of concurrent connections and the memory it takes up so we can measure before and after.

PR #1603, opened 2018-03-15
Use 0 byte reads in WebSockets to lower Idle Connection memory usage

BrennanConroy:
#1582

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
@@ -18,7 +18,6 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public partial class WebSocketsTransport : ITransport
     {
-        private static readonly ArraySegment<byte> _emptySegment = new ArraySegment<byte>(new byte[0]);
         private readonly ClientWebSocket _webSocket;
         private IDuplexPipe _application;
         private WebSocketMessageType _webSocketMessageType;
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
@@ -18,6 +18,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public partial class WebSocketsTransport : ITransport
     {
+        private static readonly ArraySegment<byte> _emptySegment = new ArraySegment<byte>(new byte[0]);
         private readonly ClientWebSocket _webSocket;
         private IDuplexPipe _application;
         private WebSocketMessageType _webSocketMessageType;
@@ -180,18 +181,12 @@ namespace Microsoft.AspNetCore.Sockets.Client
             {
                 while (true)
                 {
-                    var memory = _application.Output.GetMemory();
-
 #if NETCOREAPP2_1
-                    var receiveResult = await socket.ReceiveAsync(memory, CancellationToken.None);
+                    var result = await socket.ReceiveAsync(Memory<byte>.Empty, CancellationToken.None);
 #else
-                    var isArray = MemoryMarshal.TryGetArray<byte>(memory, out var arraySegment);
-                    Debug.Assert(isArray);
-
-                    // Exceptions are handled above where the send and receive tasks are being run.
-                    var receiveResult = await socket.ReceiveAsync(arraySegment, CancellationToken.None);
+                    var result = await socket.ReceiveAsync(_emptySegment, CancellationToken.None);
 #endif
-                    if (receiveResult.MessageType == WebSocketMessageType.Close)
+                    if (result.MessageType == WebSocketMessageType.Close)
                     {
                         Log.WebSocketClosed(_logger, _webSocket.CloseStatus);
 
@@ -202,20 +197,33 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
                         return;
                     }
-
-                    Log.MessageReceived(_logger, receiveResult.MessageType, receiveResult.Count, receiveResult.EndOfMessage);
-
-                    _application.Output.Advance(receiveResult.Count);
-
-                    if (receiveResult.EndOfMessage)
+                    else
                     {
-                        var flushResult = await _application.Output.FlushAsync();
+                        var memory = _application.Output.GetMemory();
+#if NETCOREAPP2_1
+                        var receiveResult = await socket.ReceiveAsync(memory, CancellationToken.None);
+#else
+                        var isArray = MemoryMarshal.TryGetArray<byte>(memory, out var arraySegment);
+                        Debug.Assert(isArray);
 
-                        // We canceled in the middle of applying back pressure
-                        // or if the consumer is done
-                        if (flushResult.IsCanceled || flushResult.IsCompleted)
+                        // Exceptions are handled above where the send and receive tasks are being run.
+                        var receiveResult = await socket.ReceiveAsync(arraySegment, CancellationToken.None);
+#endif
+
+                        Log.MessageReceived(_logger, receiveResult.MessageType, receiveResult.Count, receiveResult.EndOfMessage);
+
+                        _application.Output.Advance(receiveResult.Count);
+
+                        if (receiveResult.EndOfMessage)
                         {
-                            break;
+                            var flushResult = await _application.Output.FlushAsync();
+
+                            // We canceled in the middle of applying back pressure
+                            // or if the consumer is done
+                            if (flushResult.IsCanceled || flushResult.IsCompleted)
+                            {
+                                break;
+                            }
                         }
                     }
                 }
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
@@ -182,14 +182,12 @@ namespace Microsoft.AspNetCore.Sockets.Client
                 var endOfMessage = true;
                 while (true)
                 {
+#if NETCOREAPP2_1
                     // If there was a read that had a 'false' EndOfMessage then we should skip the 0 byte read since there is an in progress frame
                     if (endOfMessage)
                     {
-#if NETCOREAPP2_1
                         var result = await socket.ReceiveAsync(Memory<byte>.Empty, CancellationToken.None);
-#else
-                        var result = await socket.ReceiveAsync(_emptySegment, CancellationToken.None);
-#endif
+
                         if (result.MessageType == WebSocketMessageType.Close)
                         {
                             Log.WebSocketClosed(_logger, _webSocket.CloseStatus);
@@ -202,7 +200,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
                             return;
                         }
                     }
-
+#endif
                     var memory = _application.Output.GetMemory();
 #if NETCOREAPP2_1
                     var receiveResult = await socket.ReceiveAsync(memory, CancellationToken.None);
@@ -212,6 +210,18 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
                     // Exceptions are handled above where the send and receive tasks are being run.
                     var receiveResult = await socket.ReceiveAsync(arraySegment, CancellationToken.None);
+
+                    if (receiveResult.MessageType == WebSocketMessageType.Close)
+                    {
+                        Log.WebSocketClosed(_logger, _webSocket.CloseStatus);
+
+                        if (_webSocket.CloseStatus != WebSocketCloseStatus.NormalClosure)
+                        {
+                            throw new InvalidOperationException($"Websocket closed with error: {_webSocket.CloseStatus}.");
+                        }
+
+                        return;
+                    }
 #endif
                     endOfMessage = receiveResult.EndOfMessage;
                     Log.MessageReceived(_logger, receiveResult.MessageType, receiveResult.Count, endOfMessage);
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.AspNetCore.Sockets.Client
 {
     public partial class WebSocketsTransport : ITransport
     {
+        private static readonly ArraySegment<byte> _emptySegment = new ArraySegment<byte>(new byte[0]);
```
BrennanConroy: remove

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
```diff
@@ -107,23 +108,45 @@ public override void Dispose()
             {
                 try
                 {
-                    await _input.WaitToReadAsync();
-
-                    if (_input.TryRead(out var message))
+                    if (_internalBuffer.Buffer == null || _internalBuffer.Buffer.Length == 0)
```
BrennanConroy: Daivd suggested trying a channel/pipe here, I'll try it out

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -138,8 +139,21 @@ public WebSocketsTransport(WebSocketOptions options, IDuplexPipe application, De
         {
             try
             {
+                var endOfMessage = true;
                 while (true)
                 {
+#if NETCOREAPP2_1
+                    // If there was a read that had a 'false' EndOfMessage then we should skip the 0 byte read since there is an in progress frame
+                    if (endOfMessage)
```
BrennanConroy: Not sure how to cleanly not copy this code from below

davidfowl:
Lets also add AcceptWebSocket and pass in the buffer size for the web socket middleware and make it 14 bytes which is the max header size for a websocket.

davidfowl:
Is also like to see memory and throughput differences

src/Microsoft.AspNetCore.Sockets.Http/SocketsAppBuilderExtensions.cs
```diff
@@ -18,7 +18,10 @@ public static IApplicationBuilder UseSockets(this IApplicationBuilder app, Actio
 
             callback(new SocketRouteBuilder(routes, dispatcher));
 
-            app.UseWebSockets();
+            app.UseWebSockets(new WebSocketOptions()
+            {
+                ReceiveBufferSize = 14
```
JamesNK: Make this a private const and add a comment explaining why it is 14

davidfowl: Yes, please...  Point to this https://github.com/dotnet/corefx/blob/1df4a4866a90f22f861156b8ff496c24103d39cc/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs#L61 and this https://github.com/dotnet/corefx/blob/1df4a4866a90f22f861156b8ff496c24103d39cc/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs#L188

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -178,10 +178,29 @@ public WebSocketsTransport(HttpOptions httpOptions, ILoggerFactory loggerFactory
...
             {
+                var endOfMessage = true;
                 while (true)
                 {
-                    var memory = _application.Output.GetMemory();
+#if NETCOREAPP2_1
+                    // If there was a read that had a 'false' EndOfMessage then we should skip the 0 byte read since there is an in progress frame
+                    if (endOfMessage)
+                    {
+                        var result = await socket.ReceiveAsync(Memory<byte>.Empty, CancellationToken.None);
```
anurse: Comment about why we're doing a `0`-byte read (to wait for data without allocating any buffers)

anurse: Also, couldn't we do this with `Array.Empty<byte>` on all platforms? Still wouldn't allocate any buffers. I suppose we wouldn't be using the `struct`-returning overload though.

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -178,10 +178,29 @@ public WebSocketsTransport(HttpOptions httpOptions, ILoggerFactory loggerFactory
         {
             try
             {
+                var endOfMessage = true;
                 while (true)
                 {
-                    var memory = _application.Output.GetMemory();
+#if NETCOREAPP2_1
+                    // If there was a read that had a 'false' EndOfMessage then we should skip the 0 byte read since there is an in progress frame
```
anurse: mega-nit: `in-progress`

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
anurse: If we can unify more of the connection primitives, we may be able to unify the transports a bit. Let's not worry too much about this right now, as long as we have testing in place for behavior changes and automated benchmarks for performance changes like this to catch regression.

src/Microsoft.AspNetCore.Sockets.Http/SocketsAppBuilderExtensions.cs
anurse: Or just to the spec: https://tools.ietf.org/html/rfc6455#section-5.2

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
```diff
@@ -107,23 +108,45 @@ public override void Dispose()
...
+                                _closeStatus = message.CloseStatus;
+                                _closeStatusDescription = message.CloseStatusDescription;
+                                return new WebSocketReceiveResult(0, WebSocketMessageType.Close, true, message.CloseStatus, message.CloseStatusDescription);
+                            }
+
+                            _internalBuffer = message;
                         }
+                    }
 
-                        // REVIEW: This assumes the buffer passed in is > the buffer received
```
anurse: Just to clarify, it's because of this that we have to update this class right? Because we now pass in a `0`-byte buffer?

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
BrennanConroy: I should probably add a comment saying that 0 byte reads don't seem to be supported for other platforms

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
BrennanConroy: Yeah :(

davidfowl:
When 0 byte reads aren’t supported maybe we can use a 1 byte buffer?

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
anurse: Ah, ok.

src/Microsoft.AspNetCore.Sockets.Http/SocketsAppBuilderExtensions.cs
```diff
@@ -18,7 +23,10 @@ public static IApplicationBuilder UseSockets(this IApplicationBuilder app, Actio
 
             callback(new SocketRouteBuilder(routes, dispatcher));
 
-            app.UseWebSockets();
+            app.UseWebSockets(new WebSocketOptions()
```
JamesNK: nit: remove `()` with ctor initializer

src/Microsoft.AspNetCore.Sockets.Http/SocketsAppBuilderExtensions.cs
```diff
@@ -10,6 +10,11 @@ namespace Microsoft.AspNetCore.Builder
 {
     public static class SocketsAppBuilderExtensions
     {
+        // 14 is the maximum websocket frame header size
+        // See https://github.com/dotnet/corefx/blob/1df4a4866a90f22f861156b8ff496c24103d39cc/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs#L61
+        // And where it is used https://github.com/dotnet/corefx/blob/1df4a4866a90f22f861156b8ff496c24103d39cc/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs#L188
+        private const int ReceiveBufferSize = 14;
```
JamesNK: 👍

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
davidfowl: @BrennanConroy does this code work? If it's tested enough then it's fine.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -138,8 +138,22 @@ public WebSocketsTransport(WebSocketOptions options, IDuplexPipe application, De
         {
             try
             {
+                var endOfMessage = true;
```
davidfowl: Why does this start as true? Doesn’t it allocate on the first read?

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
BrennanConroy: All our non-functional websockets tests use it

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
BrennanConroy: I guess since we'll read the handshake it should be false

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
anurse: Does it work? DOES IT WORK? Of course it works.

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
BrennanConroy: Although, if not using Hubs then there isn't a guaranteed read

davidfowl: So it should be false right?

BrennanConroy: If we want to optimize for the Hub usage, yeah

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
```diff
@@ -107,23 +108,45 @@ public override void Dispose()
...
-                        Buffer.BlockCopy(message.Buffer, 0, buffer.Array, buffer.Offset, message.Buffer.Length);
+                    var length = _internalBuffer.Buffer.Length;
```
davidfowl: You might be able to clean up this code a little bit if you use `Memory<byte>` instead of an array.

test/Microsoft.AspNetCore.Sockets.Tests/TestWebSocketConnectionFeature.cs
```diff
@@ -107,23 +108,45 @@ public override void Dispose()
...
+                    if (buffer.Count - buffer.Offset < _internalBuffer.Buffer.Length)
```
davidfowl: Do you have to subtract the offset here?

davidfowl:
We need to wait on this, see https://github.com/dotnet/corefx/blob/23165cee77ae4dd7a7645b7927872948bb27a73e/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs#L698

And this comment https://github.com/dotnet/corefx/issues/28138#issuecomment-373778144

src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs
```diff
@@ -178,10 +178,31 @@ public WebSocketsTransport(HttpOptions httpOptions, ILoggerFactory loggerFactory
         {
             try
             {
+                // Skip the 0 byte read at first because we will receive the Hub handshake response shortly
+                var endOfMessage = false;
                 while (true)
                 {
-                    var memory = _application.Output.GetMemory();
+#if NETCOREAPP2_1
+                    // If there was a read that had a 'false' EndOfMessage then we should skip the 0 byte read since there is an in-progress frame
```
anurse: Having a previous `EndOfMessage = false` frame doesn't guarantee that there's data available to be read from the socket though. The server may not have sent the next fragment yet. Is it worth saving a few `0`-byte reads?

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -138,8 +138,23 @@ public WebSocketsTransport(WebSocketOptions options, IDuplexPipe application, De
         {
             try
             {
+                // Skip the 0 byte read at first because we will receive the Hub handshake shortly
```
anurse: We don't know this in the Sockets layer, and shouldn't rely on it.

anurse:
This is blocked and not happening for Preview 2, right? Can we close?

davidfowl:
Yep

davidfowl:
We'll revisit this

PR #1604, opened 2018-03-15
ConnectionMetadata is now ConnectionItems

davidfowl:
Reacting to https://github.com/aspnet/KestrelHttpServer/pull/2395

### DIFF
--- a/samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
+++ b/samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
@@ -23,9 +23,9 @@ namespace SocialWeather
 
         public void OnConnectedAsync(ConnectionContext connection)
         {
-            connection.Features.Get<IConnectionMetadataFeature>().Metadata["groups"] = new HashSet<string>();
+            connection.Items["groups"] = new HashSet<string>();
             var format = connection.GetHttpContext().Request.Query["formatType"].ToString();
-            connection.Features.Get<IConnectionMetadataFeature>().Metadata["format"] = format;
+            connection.Items["format"] = format;
             if (string.Equals(format, "protobuf", StringComparison.OrdinalIgnoreCase))
             {
                 var transferFormatFeature = connection.Features.Get<ITransferFormatFeature>();
@@ -47,7 +47,7 @@ namespace SocialWeather
             foreach (var connection in _connectionList)
             {
                 var context = connection.GetHttpContext();
-                var formatter = _formatterResolver.GetFormatter<T>((string)connection.Features.Get<IConnectionMetadataFeature>().Metadata["format"]);
+                var formatter = _formatterResolver.GetFormatter<T>((string)connection.Items["format"]);
                 var ms = new MemoryStream();
                 await formatter.WriteAsync(data, ms);
 
@@ -72,7 +72,7 @@ namespace SocialWeather
 
         public void AddGroupAsync(ConnectionContext connection, string groupName)
         {
-            var groups = (HashSet<string>)connection.Features.Get<IConnectionMetadataFeature>().Metadata["groups"];
+            var groups = (HashSet<string>)connection.Items["groups"];
             lock (groups)
             {
                 groups.Add(groupName);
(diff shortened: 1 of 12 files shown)


samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
```diff
@@ -23,9 +23,9 @@ public PersistentConnectionLifeTimeManager(FormatterResolver formatterResolver)
 
         public void OnConnectedAsync(ConnectionContext connection)
         {
-            connection.Features.Get<IConnectionMetadataFeature>().Metadata["groups"] = new HashSet<string>();
+            connection.Items["groups"] = new HashSet<string>();
```
JamesNK: Can Items be shared with other pieces of code? Should there be a prefix on it so make it less likely to conflict with someone else who might add data with a `"groups"` key? e.g. `"socialweather-groups"`.

Also these keys should be `const` fields. It's just a sample but try to encourage good practice though leadership :)

samples/SocketsSample/EndPoints/MessagesEndPoint.cs
```diff
@@ -19,7 +19,7 @@ public class MessagesEndPoint : EndPoint
         {
             Connections.Add(connection);
 
-            await Broadcast($"{connection.ConnectionId} connected ({connection.Features.Get<IConnectionMetadataFeature>().Metadata[ConnectionMetadataNames.Transport]})");
+            await Broadcast($"{connection.ConnectionId} connected ({connection.Items[ConnectionMetadataNames.Transport]})");
```
JamesNK: So much nicer 👼

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
```diff
@@ -66,9 +68,9 @@ public DefaultConnectionContext(string id, IDuplexPipe transport, IDuplexPipe ap
 
         public ClaimsPrincipal User { get; set; }
 
-        public IDictionary<object, object> Metadata { get; set; } = new ConnectionMetadata();
+        public override IDictionary<object, object> Items { get; set; } = new ConnectionMetadata();
```
JamesNK: Still need `ConnectionMetadata`?

davidfowl: That's correct. The base class does nothing.

samples/SocialWeather/PersistentConnectionLifeTimeManager.cs
davidfowl: In general yea, but this is just a crappy sample 😄

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
JamesNK: I see `ConnectionMetadata` makes it safe to get a value that doesn't exist.

```
// Replace the indexer with one that returns null for missing values
object IDictionary<object, object>.this[object key]
{
    get
    {
        if (Items.TryGetValue(key, out var value))
        {
            return value;
        }
        return null;
    }
    set { Items[key] = value; }
}
```

If we're doing it here, should this be the default value of `Items` in Kestrel?

At the very least I don't think we need to override Items on DefaultConnectionContext.

davidfowl: That sounds like a fine change to make but right now it's abstract in kestrel.

JamesNK: Ah ok, I didn't realize it was an abstract class. Carry on then.

davidfowl: You're talking about making it non virtual right?

JamesNK: No, I just didn't realize `ConnectionContext` was abstract.

Issue #1268, opened 2018-01-03
Consider better ways to send array in InvokeAsync


BrennanConroy:
A very easy mistake to make is passing an array to `InvokeAsync` which will treat the array as a list of arguments instead of a single argument that is an array.

```
public async Task Send(byte[] data)
{
    await Clients.All.InvokeAsync("Send", data); // wrong
    await Clients.All.InvokeAsync("Send", new [] { data }); // correct
}
```

We should consider ways of either making this more obvious or changing APIs to prevent this issue.

anurse:
Unfortunately, the only thing I can think of right now is to rename the one that takes the array to something obnoxious like `SendInvocationAsync` or something. We can't justify making it `internal` because it's a useful API.

davidfowl:
On the client we did this with overloads.

anurse:
The client was protected by the fact that it had an additional parameter in the argument array (the return type). The `SendAsync` helpers on the client have the same issue.

anurse:
Overloads don't help, because an `object[]` binds to the overload that takes an array, not the one that takes `object`

davidfowl:
That's true, I guess the only thing we could do is rename the array overload, hmmmmm.

simeyla:
This is especially confusing (new user from SignalR 2 > 3) because when you hit F12 on `InvokeAsync` it takes you to the list of extension method signatures for `IClientProxy` and you don't even see that a signature with `[]` exists!

cvanem:
I ran into this as well.  Using new [] { data } is acceptable, but won't that cause some performance concerns for large arrays as it is making a copy of the data with the new operator?  I am using this solution instead:

public async Task Send(byte[] data)
{
    await Clients.All.InvokeAsync("Send", data as object); 
}

Renaming the overload that takes the arguments array seems acceptable to me.  I don't think it is necessary, but it would prevent others from encountering this issue.

anurse:
The array is being created no matter what. `params object[]` will always allocate an array. If you are sending a message with a single argument that is a large array though, no copy of the large array occurs. The array that is allocated will contain a **reference** to the large array, not a copy of it (since only `struct`s are copied by value)

muratg:
Current POR: Rename the method which takes in an Array. (No good candidates yet.)

BrennanConroy:
[...]

PR #1605, opened 2018-03-15
Removed params from proxy's SendAsync

JamesNK:
https://github.com/aspnet/SignalR/issues/1268

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -9,7 +9,7 @@ using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public class DefaultHubLifetimeManager<THub> : HubLifetimeManager<THub>
+    public class DefaultHubLifetimeManager<THub> : HubLifetimeManager<THub> where THub : Hub
     {
         private readonly HubConnectionList _connections = new HubConnectionList();
         private readonly HubGroupList _groups = new HubGroupList();
--- a/src/Microsoft.AspNetCore.SignalR.Core/Proxies.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Proxies.cs
@@ -17,7 +17,7 @@ namespace Microsoft.AspNetCore.SignalR
             _userId = userId;
         }
 
-        public Task SendAsync(string method, object[] args)
+        public Task SendCoreAsync(string method, object[] args)
         {
             return _lifetimeManager.SendUserAsync(_userId, method, args);
         }
@@ -34,7 +34,7 @@ namespace Microsoft.AspNetCore.SignalR
             _userIds = userIds;
         }
 
-        public Task SendAsync(string method, object[] args)
+        public Task SendCoreAsync(string method, object[] args)
         {
             return _lifetimeManager.SendUsersAsync(_userIds, method, args);
         }
@@ -51,7 +51,7 @@ namespace Microsoft.AspNetCore.SignalR
             _groupName = groupName;
         }
 
-        public Task SendAsync(string method, object[] args)
+        public Task SendCoreAsync(string method, object[] args)
         {
             return _lifetimeManager.SendGroupAsync(_groupName, method, args);
         }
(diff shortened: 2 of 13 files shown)


src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public class DefaultHubLifetimeManager<THub> : HubLifetimeManager<THub>
+    public class DefaultHubLifetimeManager<THub> : HubLifetimeManager<THub> where THub : Hub
```
JamesNK: I noticed there were some classes with generic arguments for a hub but had no type constraint. I've added them but was it intentional? If so then let me know and I'll remove and add comments with a reason

src/Microsoft.AspNetCore.SignalR.Core/Proxies.cs
```diff
@@ -17,13 +17,13 @@ public UserProxy(HubLifetimeManager<THub> lifetimeManager, string userId)
             _userId = userId;
         }
 
-        public Task SendAsync(string method, params object[] args)
```
JamesNK: I noticed that when the reference was `IClientProxy` then a byte array would always be the first argument but with these typed proxies the `params` would take precedence. Just removed `params` on all the `IClientProxy` implementations.

Note that `client.SendAsync((object[])bytes);` will still expand out but that seems like a strange thing to do.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
davidfowl: That's good, we just forgot them.

davidfowl:
I only have one question, should we change this to IReadOnlyList or is that an endless rathole?

JamesNK:
@BrennanConroy Can you verify whether this fixes the original issue or not?

BrennanConroy:
I thought there was a problem with this approach. I'll take a quick look and see if I can find any issues

BrennanConroy:
Oh wait, if you pass `new string[] { "test", "this", "function" }` into `SendAsync` then you get 3 argument values on the client instead of one `"arguments":["test","this","function"]` but it should be a single array argument on the client?

JamesNK:
Hmmm, so a struct array doesn't choose the `params` overload but a reference type array does. Weird behavior.

We could have `IClientProxy.SendCoreAsync(string name, object[] args)` and then all the extension methods are called `SendAsync`.

davidfowl:
Yes, lets do the rename for the array overload. SendArrayAsync?

JamesNK:
It isn't sending a value that is an array, it is sending an array of arguments.

I like `SendCoreAsync` since it is the core method that is doing the sending that all the other overloads call.

davidfowl:
> I like SendCoreAsync since it is the core method that is doing the sending that all the other overloads call.

Usually, the *Core* method is an internal/private method that all other public methods delegate to. It's rarely also itself a public method and that's why I hesitate.

JamesNK:
True, but other than being public it fulfills the rest of the core *vibe*. It is the method that isn't intended to be called by most users and is delegated to by all the `SendAsync` overloads.

davidfowl:
Lets just rename it then. It should be rare enough.

JamesNK:
Ok

How is this for a crazy idea though:
1. Still have `IClientProxy.SendAsyncCore(string methodName, object[] args)`
2. Abstract base class called `ClientProxy` that explicitly implements `IClientProxy.CoreSendAsync` so it isn't public
3. Instead of exposing `IClientProxy` everywhere we expose `ClientProxy`
4. `IClientProxy` extension methods should still work and because `CoreSendAsync` was explicitly implemented then it isn't visible.

Personally I don't mind SendAsyncCore being public but if you don't want it visible in intellisense...

anurse:
Honestly I think this is one of the rare appropriate cases for `EditorBrowsableAttribute`... I do like your crazy idea, but I think just hiding it from IntelliSense would be sufficient rather than actually trying to hide it (plus there may be legitimate reasons for users to want to have access to `SendAsyncCore`)

davidfowl:
I'd rather expose SendCoreAsync. I'm not *that* impractical.

Issue #1523, opened 2018-03-02
HMAC the Connection ID


anurse:
For additional security, we should HMAC the connection ID before sending it to the client. When we receive it, we must verify the HMAC.

We should use the [HMACSHA256](https://msdn.microsoft.com/en-us/library/system.security.cryptography.hmacsha256(v=vs.110).aspx) class. Since we require sticky sessions, the key can be random (which the default constructor automatically generates for us)

PR #1606, opened 2018-03-15
Create connectionIds using RNGCrypto

BrennanConroy:
https://github.com/aspnet/SignalR/issues/1523

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
@@ -24,15 +24,24 @@ namespace Microsoft.AspNetCore.Sockets
     {
         private object _heartBeatLock = new object();
         private List<(Action<object> handler, object state)> _heartbeatHandlers;
+        private PipeOptions _transportPipeOptions;
+        private PipeOptions _applicationPipeOptions;
+        private IDuplexPipe _transportPipe;
+        private IDuplexPipe _applicationPipe;
 
         // This tcs exists so that multiple calls to DisposeAsync all wait asynchronously
         // on the same task
         private TaskCompletionSource<object> _disposeTcs = new TaskCompletionSource<object>();
 
-        public DefaultConnectionContext(string id, IDuplexPipe transport, IDuplexPipe application)
+        public DefaultConnectionContext(string id)
+            : this(id, PipeOptions.Default, PipeOptions.Default)
         {
-            Transport = transport;
-            Application = application;
+        }
+
+        public DefaultConnectionContext(string id, PipeOptions transportPipeOptions, PipeOptions applicationPipeOptions)
+        {
+            _transportPipeOptions = transportPipeOptions;
+            _applicationPipeOptions = applicationPipeOptions;
             ConnectionId = id;
             LastSeenUtc = DateTime.UtcNow;
 
@@ -71,9 +80,53 @@ namespace Microsoft.AspNetCore.Sockets
 
         public override IDictionary<object, object> Items { get; set; } = new ConnectionMetadata();
 
-        public IDuplexPipe Application { get; set; }
+        public IDuplexPipe Application
+        {
+            get
+            {
+                if (_transportPipe == null)
+                {
+                    lock (_heartBeatLock)
+                    {
+                        if (_transportPipe == null)
+                        {
+                            var duplexPipe = DuplexPipe.CreateConnectionPair(_transportPipeOptions, _applicationPipeOptions);
+                            _transportPipe = duplexPipe.Application;
+                            _applicationPipe = duplexPipe.Transport;
+                        }
+                    }
+                }
+                return _transportPipe;
+            }
+            set
+            {
+                _transportPipe = value;
+            }
+        }
 
-        public override IDuplexPipe Transport { get; set; }
+        public override IDuplexPipe Transport
+        {
+            get
+            {
+                if (_applicationPipe == null)
+                {
+                    lock (_heartBeatLock)
+                    {
+                        if (_applicationPipe == null)
+                        {
+                            var duplexPipe = DuplexPipe.CreateConnectionPair(_transportPipeOptions, _applicationPipeOptions);
+                            _transportPipe = duplexPipe.Application;
+                            _applicationPipe = duplexPipe.Transport;
+                        }
+                    }
+                }
+                return _applicationPipe;
+            }
+            set
+            {
+                _applicationPipe = value;
+            }
+        }
 
         public TransferFormat SupportedFormats { get; set; }
 
--- a/src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
@@ -22,7 +22,8 @@ namespace Microsoft.AspNetCore.Sockets
                                             IConnectionHeartbeatFeature,
                                             ITransferFormatFeature
     {
-        private List<(Action<object> handler, object state)> _heartbeatHandlers = new List<(Action<object> handler, object state)>();
+        private object _heartBeatLock = new object();
+        private List<(Action<object> handler, object state)> _heartbeatHandlers;
 
         // This tcs exists so that multiple calls to DisposeAsync all wait asynchronously
         // on the same task
@@ -80,16 +81,25 @@ namespace Microsoft.AspNetCore.Sockets
 
         public void OnHeartbeat(Action<object> action, object state)
         {
-            lock (_heartbeatHandlers)
+            lock (_heartBeatLock)
             {
+                if (_heartbeatHandlers == null)
+                {
+                    _heartbeatHandlers = new List<(Action<object> handler, object state)>();
+                }
                 _heartbeatHandlers.Add((action, state));
             }
         }
 
         public void TickHeartbeat()
         {
-            lock (_heartbeatHandlers)
+            lock (_heartBeatLock)
             {
+                if (_heartbeatHandlers == null)
+                {
+                    return;
+                }
+
                 foreach (var (handler, state) in _heartbeatHandlers)
                 {
                     handler(state);
--- a/src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
@@ -24,24 +24,13 @@ namespace Microsoft.AspNetCore.Sockets
     {
         private object _heartBeatLock = new object();
         private List<(Action<object> handler, object state)> _heartbeatHandlers;
-        private PipeOptions _transportPipeOptions;
-        private PipeOptions _applicationPipeOptions;
-        private IDuplexPipe _transportPipe;
-        private IDuplexPipe _applicationPipe;
 
         // This tcs exists so that multiple calls to DisposeAsync all wait asynchronously
         // on the same task
         private TaskCompletionSource<object> _disposeTcs = new TaskCompletionSource<object>();
 
         public DefaultConnectionContext(string id)
-            : this(id, PipeOptions.Default, PipeOptions.Default)
         {
-        }
-
-        public DefaultConnectionContext(string id, PipeOptions transportPipeOptions, PipeOptions applicationPipeOptions)
-        {
-            _transportPipeOptions = transportPipeOptions;
-            _applicationPipeOptions = applicationPipeOptions;
             ConnectionId = id;
             LastSeenUtc = DateTime.UtcNow;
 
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -71,6 +78,7 @@ public DefaultConnectionContext CreateConnection(PipeOptions transportPipeOption
             var connectionTimer = SocketEventSource.Log.ConnectionStart(id);
             var pair = DuplexPipe.CreateConnectionPair(transportPipeOptions, appPipeOptions);
 
+            // TODO: We can possibly delay creating connections until we get the first request and check the connectionId
```
davidfowl: Why is this a todo? If we're not doing this I'm not sure what the point is?

JamesNK:
The issue talks about using an HMAC to valid the ID, but the PR is using HMAC as the ID. Issue change in offline discussion?

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -94,8 +102,10 @@ public void RemoveConnection(string id)
 
         private static string MakeNewConnectionId()
         {
-            // TODO: We need to sign and encyrpt this
-            return Guid.NewGuid().ToString();
+            var id = Interlocked.Increment(ref _id);
+            // HMAC the id because we want a cryptographically random id, which GUID is not
+            var hashed = _hmacHash.ComputeHash(Encoding.UTF8.GetBytes(id.ToString()));
```
davidfowl: Maybe this whole thing makes no sense and we should just use rngcryptoserviceprovider

BrennanConroy:
Yeah, lets talk offline today

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
anurse: That's what I've been saying. If the only outcome is to generate a more-different random number, why not just do that? If the goal is to validate something, we need a different approach.

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
```diff
@@ -22,16 +22,26 @@ public class DefaultConnectionContext : ConnectionContext,
...
+        private PipeOptions _applicationPipeOptions;
+        private IDuplexPipe _transportPipe;
+        private IDuplexPipe _applicationPipe;
 
         // This tcs exists so that multiple calls to DisposeAsync all wait asynchronously
         // on the same task
         private TaskCompletionSource<object> _disposeTcs = new TaskCompletionSource<object>();
 
-        public DefaultConnectionContext(string id, IDuplexPipe transport, IDuplexPipe application)
+        public DefaultConnectionContext(string id)
```
davidfowl: I'm not a fan of these changes. I think you should just set the pipes on the outside. The options shouldn't be part of this class at all.

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -94,8 +97,19 @@ public void RemoveConnection(string id)
...
         private static string MakeNewConnectionId()
         {
-            // TODO: We need to sign and encyrpt this
-            return Guid.NewGuid().ToString();
+#if NETCOREAPP2_1
+            // 128 bit buffer / 8 bits per byte = 16 bytes
+            Span<byte> buffer = stackalloc byte[16];
+            _keyGen.GetBytes(buffer);
+            // Generate the id with RNGCrypto because we want a cryptographically random id, which GUID is not
+            return WebEncoders.Base64UrlEncode(buffer.ToArray());
```
davidfowl: Convert to hex, don't base64 encode.

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
BrennanConroy: Do we care about people using this layer having to do that themselves?

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
anurse: Y u hate base64? It looks so much cleaner, and no ugly `==` because it's the URL encode variant.

Plus, the hex encoding code looks gross (`BitConverter.ToString(buffer.ToArray()).Replace("-", "")`)

```
Base64URL: MTIzNDU2NzgxMjM0NTY3OA
Hex:       31323334353637383132333435363738

**Comment by anurse** [comment]
To close the loop on the discussions, we decided that the impact of HMACing or encrypting the key was heavy and would only gain us the ability to remove the initial allocation of the entry in the connection list. Keeping track of the IDs we've issued (which we do now) is equally secure, but we can reduce allocations by avoiding allocating the full `DefaultConnectionContext` on the `/negotiate` endpoint, and only allocate it when the client returns for the poll/SSE/WebSocket/send endpoints.

File: src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs (hunk 4 above)
Comment (anurse) [reply src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs]: So much cleaner, looks more encrypty (it's "Base64 Encrypted" after all ;))

Comment (anurse) [reply src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs]: fite me irl

File: src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs [hunk 5]
```diff
@@ -94,8 +97,19 @@ public void RemoveConnection(string id)
...
+#else
+            // 128 bit buffer / 8 bits per byte = 16 bytes
+            var buffer = new byte[16];
```
anurse: Since this buffer is so short-lived and limited in scope, it it worth using `ArrayPool<byte>.Shared` to rent it? Legit not sure exactly what the use cases are for that. @davidfowl ?

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
anurse: I think it's OK to require that callers have to set the pipe themselves. User applications (Endpoints) will get a fully configured `ConnectionContext`, so we're the only consumers of this during initialization.

davidfowl: We can keep the other ctor to avoid changing all of the callers. We're just adding a new optional overload.

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
davidfowl: Then we need to update the Base64 encoder to support taking a span and we need to update it to use the new base64 decoder.

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
davidfowl: No I think this is stackalloc worthy but only if we update `WebEncoders.Base64UrlEncode(buffer)` to take a `Span<byte>`

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -22,6 +24,8 @@ public class ConnectionManager
         // TODO: Consider making this configurable? At least for testing?
         private static readonly TimeSpan _heartbeatTickRate = TimeSpan.FromSeconds(1);
 
+        private static readonly RNGCryptoServiceProvider _keyGen = new RNGCryptoServiceProvider();
```
davidfowl: `_keyGenerator`

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
BrennanConroy: https://github.com/aspnet/Home/issues/2966

davidfowl: Until then, lets remove the stackalloc since that just forces us to allocate more.

test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
```diff
@@ -23,7 +24,6 @@ public void NewConnectionsHaveConnectionId()
             Assert.Null(connection.TransportTask);
             Assert.Null(connection.Cancellation);
             Assert.NotEqual(default(DateTime), connection.LastSeenUtc);
-            Assert.NotNull(connection.Transport);
```
davidfowl: Assert.Null

test/Microsoft.AspNetCore.Sockets.Tests/ConnectionManagerTests.cs
```diff
@@ -180,10 +181,9 @@ public void RemoveConnection()
         public async Task DisposeInactiveConnection()
         {
             var connectionManager = CreateConnectionManager();
-            var connection = connectionManager.CreateConnection();;
+            var connection = connectionManager.CreateConnection(PipeOptions.Default, PipeOptions.Default);
 
             Assert.NotNull(connection.ConnectionId);
-            Assert.NotNull(connection.Transport);
```
davidfowl: Add this back

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -555,6 +558,15 @@ private DefaultConnectionContext CreateConnectionInternal(HttpSocketOptions opti
                 return null;
             }
 
+            if (connection.Transport == null)
```
davidfowl: Add a comment here.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -527,16 +527,19 @@ private static void WriteTransport(JsonWriter writer, string transportName, Tran
...
+            {
+                var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportMaxBufferSize, resumeWriterThreshold: options.TransportMaxBufferSize / 2, readerScheduler: PipeScheduler.ThreadPool, useSynchronizationContext: false);
+                var appPipeOptions = new PipeOptions(pauseWriterThreshold: options.ApplicationMaxBufferSize, resumeWriterThreshold: options.ApplicationMaxBufferSize / 2, readerScheduler: PipeScheduler.ThreadPool, useSynchronizationContext: false);
+                var pair = DuplexPipe.CreateConnectionPair(transportPipeOptions, appPipeOptions);
+                connection.Transport = pair.Application;
+                connection.Application = pair.Transport;
+            }
 
-        private DefaultConnectionContext CreateConnectionInternal(HttpSocketOptions options)
```
davidfowl: Why not bring this back and just check the transport here internally?

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
```diff
@@ -22,16 +22,20 @@ public class DefaultConnectionContext : ConnectionContext,
                                             IConnectionHeartbeatFeature,
                                             ITransferFormatFeature
     {
-        private List<(Action<object> handler, object state)> _heartbeatHandlers = new List<(Action<object> handler, object state)>();
+        private object _heartBeatLock = new object();
```
anurse: nit: `_heartbeatLock` (lower case `b`)

src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs
```diff
@@ -22,16 +22,20 @@ public class DefaultConnectionContext : ConnectionContext,
...
+        private List<(Action<object> handler, object state)> _heartbeatHandlers;
 
         // This tcs exists so that multiple calls to DisposeAsync all wait asynchronously
         // on the same task
         private TaskCompletionSource<object> _disposeTcs = new TaskCompletionSource<object>();
 
-        public DefaultConnectionContext(string id, IDuplexPipe transport, IDuplexPipe application)
+        /// <summary>
+        /// Creates the DefaultConnectionContext without Pipes to avoid upfront allocations.
+        /// You'll need to add Transport and Application pipes on your own.
```
anurse: ~~You'll need to add Transport and Application pipes on your own~~

`The caller is expected to set the <see cref="Transport"/> and <see cref="Application"/> pipes manually`

src/Microsoft.AspNetCore.Sockets.Http/ConnectionManager.cs
```diff
@@ -94,8 +106,12 @@ public void RemoveConnection(string id)
 
         private static string MakeNewConnectionId()
         {
-            // TODO: We need to sign and encyrpt this
-            return Guid.NewGuid().ToString();
+            // TODO: Use Span when WebEncoders implements Span methods
```
anurse: File a bug

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -527,16 +527,25 @@ private static void WriteTransport(JsonWriter writer, string transportName, Tran
...
             return connection;
         }
 
-        private DefaultConnectionContext CreateConnectionInternal(HttpSocketOptions options)
+        private void EnsureConnectionStateInternal(DefaultConnectionContext connection, HttpSocketOptions options)
         {
-            var transportPipeOptions = new PipeOptions(pauseWriterThreshold: options.TransportMaxBufferSize, resumeWriterThreshold: options.TransportMaxBufferSize / 2, readerScheduler: PipeScheduler.ThreadPool, useSynchronizationContext: false);
-            var appPipeOptions = new PipeOptions(pauseWriterThreshold: options.ApplicationMaxBufferSize, resumeWriterThreshold: options.ApplicationMaxBufferSize / 2, readerScheduler: PipeScheduler.ThreadPool, useSynchronizationContext: false);
-            return _manager.CreateConnection(transportPipeOptions, appPipeOptions);
+            // If the connection doesn't have a pipe yet then create one, we lazily create the pipe to save on allocations until the connection actually connects
```
anurse: `until the client actually connects`

Issue #1087, opened 2017-11-03
Use PascalCasing for parameter names in Logger templates


anurse:
It looks like that's the common pattern here: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MvcViewFeaturesLoggerExtensions.cs

cc @muratg @BrennanConroy

PR #1615, opened 2018-03-16
Pascal case formatted log parameter names

BrennanConroy:
#1087


davidfowl:
Can you use a tool and tell me if ConnectionId clashes with the Kestrel ConnectionId

BrennanConroy:
Tomorrow, sure

davidfowl:
> Presuming ConnectionId doesn't clash or you rename it if it does: LGTM

Yep. I think it does. Last time I used Seq I believe there was ConnectionId from Kestrel and connectionId from SignalR.

BrennanConroy:
Renaming to `SocketsConnectionId`

Issue #1129, opened 2017-11-16
Consider the following optimization to TryParseMessage


ahsonkhan:
1) Only set `payload = default` when necessary.
2) Do not use _numBitsToShift and just use numBytes * 7.

![image](https://user-images.githubusercontent.com/6527137/32870291-11d52984-ca30-11e7-8dda-f8e8c4d8c572.png)

https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs#L13
```C#
public static bool TryParseMessage(ref ReadOnlySpan<byte> buffer, out ReadOnlySpan<byte> payload)
{
    var index = buffer.IndexOf(TextMessageFormatter.RecordSeparator);
    if (index == -1)
    {
        payload = default;
        return false;
    }

    payload = buffer.Slice(0, index);

    // Skip record separator
    buffer = buffer.Slice(index + 1);

    return true;
}
```

https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs#L13
```C#
public static bool TryParseMessage(ref ReadOnlySpan<byte> buffer, out ReadOnlySpan<byte> payload)
{
    if (buffer.IsEmpty) goto FalseExit;
...
    do
    {
        byteRead = lengthPrefixBuffer[numBytes];
        length = length | (((uint)(byteRead & 0x7f)) << (numBytes * 7) );
        numBytes++;
    }
    while (numBytes < lengthPrefixBuffer.Length && ((byteRead & 0x80) != 0));
    // size bytes are missing
    if ((byteRead & 0x80) != 0 && (numBytes < MaxLengthPrefixSize)) goto FalseExit;
...
    // We don't have enough data
    if (buffer.Length < length + numBytes) goto FalseExit;

    // Get the payload
    payload = buffer.Slice(numBytes, (int)length);

    // Skip the payload
    buffer = buffer.Slice(numBytes + (int)length);
    return true;

    FalseExit:
    payload = default;
    return false;
}
```

cc @moozzyk, @KrzysztofCwalina

moozzyk:
@davidfowl https://github.com/aspnet/SignalR/pull/948#discussion_r141091410  LOL

davidfowl:
LOL, lookup table for the fail 😄 . I bet the shift is faster 😄

ahsonkhan:
[...]

PR #1616, opened 2018-03-16
Perf in MessageParsers

BrennanConroy:
https://github.com/aspnet/SignalR/issues/1129

I'll run micro and macro tomorrow. Don't expect anything from macro until they're more stable though


src/Microsoft.AspNetCore.SignalR.Common/Internal/Formatters/BinaryMessageParser.cs
```diff
@@ -39,14 +37,15 @@ public static bool TryParseMessage(ref ReadOnlySpan<byte> buffer, out ReadOnlySp
             do
             {
                 byteRead = lengthPrefixBuffer[numBytes];
-                length = length | (((uint)(byteRead & 0x7f)) << _numBitsToShift[numBytes]);
+                length = length | (((uint)(byteRead & 0x7f)) << (numBytes * 7));
```
anurse: YAS. I had this in my initial incremental parsing spike. Nice.

PR #1617, opened 2018-03-16
Small optimizations

davidfowl:
- Return ValueTask instead of Task from WriteAsync helpers
- Use TryGet instead of foreach to avoid enumerator (though it's just a stack allocation here)

### DIFF
--- a/src/Common/WebSocketExtensions.cs
+++ b/src/Common/WebSocketExtensions.cs
@@ -1,12 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Buffers;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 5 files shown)


src/Common/WebSocketExtensions.cs
```diff
@@ -1,42 +1,60 @@
...
+                Debug.Assert(isArray);
+                await webSocket.SendAsync(arraySegment, webSocketMessageType, endOfMessage: false, cancellationToken);
+#endif
+            }
+
+            // Empty end of message frame
+#if NETCOREAPP2_1
+            await webSocket.SendAsync(Memory<byte>.Empty, webSocketMessageType, endOfMessage: true, cancellationToken);
+#else
+            await webSocket.SendAsync(new ArraySegment<byte>(Array.Empty<byte>()), webSocketMessageType, endOfMessage: true, cancellationToken);
```
BrennanConroy: Cache this?

BrennanConroy:
Don't forget to change the branch base to release/2.1

src/Common/WebSocketExtensions.cs
JamesNK: It's a struct so I don't think there is any benefit

davidfowl:
@JamesNK it would be but we don't have anything that would show this..

PR #1625, opened 2018-03-16
fix #1542 by removing browser field

### LINKED ISSUE #1542: Don't specify the 'browser' field in package.json
It turns out webpack tries to use that when packing even though it is more appropriate for it use the CommonJS build (currently in `main`). We just need to remove the `browser` field, I misunderstood it :)

anurse:
WebPack is using it, but it's supposed to use the "Main" value.


anurse:
Travis failure was a network issue. Restarted

PR #1627, opened 2018-03-16
Upgrade Microsoft.NET.Test.Sdk to 15.6.1

mikeharder:
* Should fix race condition which occasionally causes tests to fail (https://github.com/Microsoft/vstest/issues/1435)


anurse:
Actually, can we put this into `release/2.1`? We're going to be churning that branch quite a bit still, so it'd be good to have it there.

mikeharder:
Yes, I will rebase.  Travis/Linux is green, but other builds have been queued for hours.  Is just Travis/Linux good enough?

anurse:
Yeah, that's OK with me. Can't wait for the new PR CI :(

PR #1630, opened 2018-03-17
Reduce the per message allocations

### LINKED ISSUE #1618: Reduce per message reading allocations in JsonHubProtocol
Today in the JsonHubProtocol to read a single message, we allocate
 - A new byte[] with the data sliced from the original buffer
 - A MemoryStream wrapping that byte[]
 - A StreamReader wrapping that MemoryStream (which itself has a 1K byte[] and 2K char[])
 - A JsonTextReader that allocates internally a 1K buffer from an optional array pool (which we don't currently use) 

https://github.com/aspnet/SignalR/blob/b0e5483a5c9452446169664f525cc3279ae4d0ad/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs#L52-L58

https://github.com/aspnet/SignalR/blob/8b8c7107a8e0b4a2123f8a8178525f1da8aaeb16/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs#L72

The protocol parsers themselves are stateless and singleton which makes caching hard unless we use an object pool on the server. On the client, the protocol is allocated per client connection (which makes sense). The other complication is that the exact same code is shared on client and server so that makes it hard to specialize for the sever though these optimization should apply to the client as well.

davidfowl:
- Introduce Utf8BufferTextReader which is a simple TextReader implemented over a utf8 buffer. It encodes directly into the buffer allocated by JSON.NET.
- Use JSON.NET's IArrayPool feature and back it with the shared array pool.

## Baseline

image

```
             Method |          Input | HubProtocol |         Mean |       Error |       StdDev |       Median |        Op/s |   Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------ |-------------:|------------:|-------------:|-------------:|------------:|--------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |        Json |   6,748.6 ns |   385.87 ns |  1,107.14 ns |   6,381.7 ns |   148,179.5 |  0.7172 | 0.0076 |    8728 B |
  ReadSingleMessage | LargeArguments |        Json | 137,546.7 ns | 4,607.99 ns | 13,514.45 ns | 136,188.9 ns |     7,270.3 | 18.0664 | 4.6387 |  217408 B | 
  ReadSingleMessage |  ManyArguments |        Json |   8,444.5 ns |   165.59 ns |    154.89 ns |   8,435.7 ns |   118,420.3 |  0.8698 | 0.0153 |   10600 B |
  ReadSingleMessage |    NoArguments |        Json |   4,381.1 ns |   211.40 ns |    613.31 ns |   4,278.5 ns |   228,254.2 |  0.6332 |      - |    7648 B |
```

## This PR

image

```
             Method |          Input | HubProtocol |         Mean |        Error |       StdDev |       Median |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------ |-------------:|-------------:|-------------:|-------------:|------------:|-------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |        Json |   5,576.3 ns |   127.120 ns |   197.911 ns |   5,510.1 ns |   179,331.1 | 0.2594 |      - |    3264 B |
  ReadSingleMessage | LargeArguments |        Json | 109,498.0 ns | 3,127.507 ns | 8,973.404 ns | 107,669.3 ns |     9,132.6 | 4.8828 | 0.4883 |   60960 B |
  ReadSingleMessage |  ManyArguments |        Json |   8,756.9 ns |   323.938 ns |   934.636 ns |   8,294.2 ns |   114,196.0 | 0.4120 |      - |    5096 B |
  ReadSingleMessage |    NoArguments |        Json |   3,712.5 ns |   162.313 ns |   470.900 ns |   3,570.0 ns |   269,359.2 | 0.1793 |      - |    2192 B |
 ```

Fixes #1618

### CODE REVIEW

File: src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs [hunk 1]
```diff
@@ -47,15 +48,13 @@ public JsonHubProtocol(IOptions<JsonHubProtocolOptions> options)
...
             while (TextMessageParser.TryParseMessage(ref input, out var payload))
             {
-                // TODO: Need a span-native JSON parser!
-                using (var memoryStream = new MemoryStream(payload.ToArray()))
-                {
-                    messages.Add(ParseMessage(memoryStream, binder));
-                }
+                var textReader = new Utf8BufferTextReader(payload);
+                messages.Add(ParseMessage(textReader, binder));
+
```
JamesNK: nit: remove line

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -67,10 +66,12 @@ public void WriteMessage(HubMessage message, Stream output)
             TextMessageFormatter.WriteRecordSeparator(output);
         }
 
-        private HubMessage ParseMessage(Stream input, IInvocationBinder binder)
+        private HubMessage ParseMessage(TextReader input, IInvocationBinder binder)
         {
-            using (var reader = new JsonTextReader(new StreamReader(input)))
+            using (var reader = new JsonTextReader(input))
```
JamesNK: Add an internal helper method for creating a `JsonTextReader` with shared array pool?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextReader.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
+{
+    internal class Utf8BufferTextReader : TextReader
+    {
+        private ReadOnlyMemory<byte> _utf8Buffer;
+
+        public Utf8BufferTextReader(ReadOnlyMemory<byte> utf8Buffer)
```
JamesNK: Every time I read `HubEndPoint` and see the `ToArray` line 150 I don't like it. See below.

What about if `HubProtocol` took a sequence, which was then passed to this text reader. The logic in `Utf8BufferTextReader` would stay largely the same expect it could read over each memory block in the sequence instead of just one. We can avoid copying the sequence in its entirety into a new array.

https://github.com/aspnet/SignalR/blob/8b8c7107a8e0b4a2123f8a8178525f1da8aaeb16/src/Microsoft.AspNetCore.SignalR.Core/HubEndPoint.cs#L138-L168

davidfowl: This change will be made after @anurse 's refactoring. The problem is that right now the client and server are asymmetrical and they don't both parse `ReadOnlySequence<byte>`. Client uses a byte[] and the server uses `ReadOnlySequence<byte>`.  This will be done as part of https://github.com/aspnet/SignalR/issues/1508.

JamesNK: Could we make the server and `Utf8BufferTextReader` work with `ReadOnlySequence<byte>` now? On the client we can just wrap the `byte[]` in a `ReadOnlySequence<byte>` with little to no overhead before parsing, no?

davidfowl: I don't really see the point.

davidfowl: I also hate ToArray() with a passion but there is a master plan here and it's going to be glorious when it comes together.

davidfowl:
Closing in favor of https://github.com/aspnet/SignalR/pull/1635

client-ts/FunctionalTests/package-lock.json
```diff
@@ -7,13 +7,13 @@
     "@std/esm": {
       "version": "0.18.0",
       "resolved": "https://registry.npmjs.org/@std/esm/-/esm-0.18.0.tgz",
-      "integrity": "sha512-oeHSSVp/WxC08ngpKgyYR4LcI0+EBwZiJcB58jvIqyJnOGxudSkxTgAQKsVfpNsMXfOoILgu9PWhuzIZ8GQEjw==",
+      "integrity": "sha1-4hK1Zcdl+Tsp7FIqSToZLOeuK6Y=",
```
anurse: Upgrade your NPM bro

PR #1631, opened 2018-03-17
WIP TS Fallback

mikaelm12:
Needs tests and some todos still

### DIFF
--- a/client-ts/signalr/src/HttpConnection.ts
+++ b/client-ts/signalr/src/HttpConnection.ts
@@ -42,6 +42,8 @@ export class HttpConnection implements IConnection {
     private transport: ITransport;
     private connectionId: string;
     private startPromise: Promise<void>;
+    private negotiateResponse: INegotiateResponse;
+    private allTransports: TransportType[] = [TransportType.WebSockets, TransportType.ServerSentEvents, TransportType.LongPolling];
 
     public readonly features: any = {};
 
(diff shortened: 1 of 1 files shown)


client-ts/signalr/src/HttpConnection.ts
```diff
@@ -87,41 +92,61 @@ export class HttpConnection implements IConnection {
...
 
                 // the user tries to stop the the connection when it is being started
                 if (this.connectionState === ConnectionState.Disconnected) {
                     return;
                 }
 
-                if (this.connectionId) {
-                    this.url = this.baseUrl + (this.baseUrl.indexOf("?") === -1 ? "?" : "&") + `id=${this.connectionId}`;
-                    this.transport = this.createTransport(this.options.transport, negotiateResponse.availableTransports, transferFormat);
+                for (const transport of this.allTransports) {
```
mikaelm12: Should be doing something more similar to https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs#L228-L234

davidfowl: Yes, why do we hardcode the list?

client-ts/signalr/src/HttpConnection.ts
```diff
@@ -42,6 +42,8 @@ export class HttpConnection implements IConnection {
     private transport: ITransport;
     private connectionId: string;
     private startPromise: Promise<void>;
+    private negotiateResponse: INegotiateResponse;
```
anurse: This can just be a local in `startInternal`, right?

client-ts/signalr/src/HttpConnection.ts
anurse: `createTransport` and `resolveTransport` already does this looping. We should move this logic into `createTransport`. That code is already using `resolveTransport` to determine if the client can support the transport, all we should need is to move the `try { connect(); break; } catch { /* continue */ }` logic into the loop in `createTransport`

Issue #1545, opened 2018-03-07
Add better logic around deserializing headers


BrennanConroy:
Currently when deserializing messages we grab the header value and check if there are any headers to write. This causes an allocation per send that we could easily avoid
https://github.com/aspnet/SignalR/blob/0eb2b96c45964be461394e73b960dd634fca405f/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubInvocationMessage.cs#L16

![dict](https://user-images.githubusercontent.com/7574801/37066103-ab192ef4-2158-11e8-8218-3233d8cead80.PNG)

davidfowl:
@anurse tsk 😄

anurse:
I just wanted to leave some low-hanging performance fruit around ;). Boost morale and such XD

PR #1632, opened 2018-03-17
[WIP]: Rewriting JsonHubProtocol to use JsonTextReader directly

### LINKED ISSUE #1549: Improve JsonHubProtocol parsing of JSON message
Today when parsing incoming messages `JsonHubProtocol` loads the whole JSON document into a `JObject` and then pulls data out of it as needed. This has the overhead of allocating many JObject/JArray/JValue instances and their own internal data structures (lists, dictionaries, box primitives, etc). Also arguments are parsed once into JTokens, and then those JTokens will be read again by the serializer.

Changing parse message to use `JsonTextReader` directly would be more efficient - both in allocations and amount of work done - but poses some challenges. The reader is forward only, and JSON objects are unordered.

Parse message will need to handle data coming in any order, e.g. arguments might be read before the method name, or the message type might come last. Ideally we would want arguments deserialized directly from the reader, but in cases where we don't know the method name yet - and therefor don't know the argument types - we'll need to temporarily cached them in a JArray as a fallback.

https://github.com/aspnet/SignalR/blob/0eb2b96c45964be461394e73b960dd634fca405f/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs#L69-L111

davidfowl:
The parsing works but the tests don't pass yet because @anurse's error messages from the original impl were very specific and I need to either delete those tests or try to get all of the same failures (Will probably do a mix of both). This PR builds on top of https://github.com/aspnet/SignalR/pull/1630.

I want to get early feedback from @JamesNK on the pattern I used here. Here are preliminary results. TL;DR everything allocates less and is faster except large arguments, and I'm looking at that (I'm not sure why it's worse but maybe something obvious sticks out):

## Baseline

image

```
             Method |          Input | HubProtocol |         Mean |        Error |       StdDev |       Median |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------ |-------------:|-------------:|-------------:|-------------:|------------:|-------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |        Json |   5,576.3 ns |   127.120 ns |   197.911 ns |   5,510.1 ns |   179,331.1 | 0.2594 |      - |    3264 B |
  ReadSingleMessage | LargeArguments |        Json | 109,498.0 ns | 3,127.507 ns | 8,973.404 ns | 107,669.3 ns |     9,132.6 | 4.8828 | 0.4883 |   60960 B |
  ReadSingleMessage |  ManyArguments |        Json |   8,756.9 ns |   323.938 ns |   934.636 ns |   8,294.2 ns |   114,196.0 | 0.4120 |      - |    5096 B |
  ReadSingleMessage |    NoArguments |        Json |   3,712.5 ns |   162.313 ns |   470.900 ns |   3,570.0 ns |   269,359.2 | 0.1793 |      - |    2192 B |
 ```

## This PR

image

```
             Method |          Input | HubProtocol |         Mean |        Error |       StdDev |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------ |-------------:|-------------:|-------------:|------------:|-------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |        Json |   3,970.2 ns |    45.860 ns |    42.898 ns |   251,877.8 | 0.1450 |      - |    1880 B |
  ReadSingleMessage | LargeArguments |        Json | 148,925.5 ns | 2,226.241 ns | 1,859.012 ns |     6,714.8 | 8.7891 | 0.4883 |  107584 B |
  ReadSingleMessage |  ManyArguments |        Json |   5,926.6 ns |    71.372 ns |    63.269 ns |   168,730.3 | 0.2060 |      - |    2600 B |
  ReadSingleMessage |    NoArguments |        Json |   2,397.2 ns |    29.460 ns |    27.557 ns |   417,151.5 | 0.1221 |      - |    1480 B |
```

## Updated

image

```
             Method |          Input | HubProtocol |        Mean |       Error |      StdDev |      Median |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------ |------------:|------------:|------------:|------------:|------------:|-------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |        Json |  2,284.4 ns |    45.48 ns |    37.98 ns |  2,269.2 ns |   437,746.6 | 0.0992 |      - |    1216 B |
  ReadSingleMessage | LargeArguments |        Json | 99,503.4 ns | 2,332.69 ns | 6,692.91 ns | 99,054.2 ns |    10,049.9 | 4.7607 | 0.2441 |   59144 B |
  ReadSingleMessage |  ManyArguments |        Json |  3,798.7 ns |    73.69 ns |    68.93 ns |  3,807.6 ns |   263,251.4 | 0.1450 |      - |    1776 B |
  ReadSingleMessage |    NoArguments |        Json |  1,346.2 ns |    16.38 ns |    15.32 ns |  1,351.4 ns |   742,818.6 | 0.0744 |      - |     896 B |
 ```

## Final PR

```
             Method |          Input | HubProtocol |        Mean |        Error |       StdDev |      Median |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------ |------------:|-------------:|-------------:|------------:|------------:|-------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |        Json |  2,295.2 ns |    68.669 ns |    60.873 ns |  2,273.0 ns |   435,693.4 | 0.0992 |      - |    1216 B |
  ReadSingleMessage | LargeArguments |        Json | 91,082.5 ns | 1,701.028 ns | 1,820.081 ns | 91,296.7 ns |    10,979.1 | 4.7607 | 0.2441 |   59144 B |
  ReadSingleMessage |  ManyArguments |        Json |  3,626.3 ns |    68.961 ns |    64.506 ns |  3,599.4 ns |   275,760.7 | 0.1450 |      - |    1776 B |
  ReadSingleMessage |    NoArguments |        Json |  1,260.0 ns |    24.713 ns |    27.469 ns |  1,254.7 ns |   793,670.2 | 0.0744 |      - |     896 B |
```

Fixes #1549 and #1545


samples/SocketsSample/Program.cs
```diff
@@ -14,35 +18,38 @@ public class Program
     {
         public static void Main(string[] args)
         {
-            var config = new ConfigurationBuilder()
```
davidfowl: Will undo this before merging. This was me doing memory measurements.

samples/SocketsSample/Properties/launchSettings.json
```diff
@@ -17,7 +17,6 @@
     },
     "SocketsSample": {
       "commandName": "Project",
-      "launchBrowser": true,
```
davidfowl: Will revert this too.

samples/SocketsSample/SocketsSample.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp2.1;net461</TargetFrameworks>
```
davidfowl: And this.

davidfowl:
Seems like we're double allocating strings with the current reader approach, here's a memory profile for the large argument before and after:

## Before
image

## After

image

We do allocate the JsonTextReader twice and do 2 passes but I'm wondering why that allocates strings.

image

davidfowl:
I'm going to try a hyrbid approach. Will update in a couple of hours.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/HubMessageHelpers.cs
```diff
@@ -11,6 +11,11 @@ internal static class HubMessageHelpers
         // This lets you add headers to a hub message and return it, in a single expression.
         public static HubMessage AddHeaders(IDictionary<string, string> headers, HubInvocationMessage hubMessage)
         {
+            if (hubMessage.Headers == null)
+            {
+                hubMessage.Headers = new Dictionary<string, string>();
```
JamesNK: Skip if there are no values in `headers`?

davidfowl: This is a test 😄

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -66,70 +65,183 @@ public void WriteMessage(HubMessage message, Stream output)
...
-                            return BindCancelInvocationMessage(json);
-                        case HubProtocolConstants.PingMessageType:
-                            return PingMessage.Instance;
-                        default:
-                            throw new InvalidDataException($"Unknown message type: {type}");
+                        switch (reader.TokenType)
+                        {
+                            case JsonToken.PropertyName:
+                                // We only care about parsing top level properties
+                                if (reader.Depth == 1)
```
JamesNK: I'm not a fan of this. Better to `reader.Skip()` if we don't care about the content in a property. Then we know we're always at the root object here and the only two token types to worry about are properties and the end object.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -66,70 +65,183 @@ public void WriteMessage(HubMessage message, Stream output)
...
+                                {
+                                    string memberName = reader.Value.ToString();
+
+                                    if (string.Equals(memberName, TypePropertyName, StringComparison.Ordinal))
```
JamesNK: Is this faster than a switch on `memberName`?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -66,70 +65,183 @@ public void WriteMessage(HubMessage message, Stream output)
...
+                                            catch (Exception ex)
+                                            {
+                                                argumentBindingException = ExceptionDispatchInfo.Capture(ex);
+                                            }
+                                        }
+                                    }
+                                    else if (string.Equals(memberName, HeadersPropertyName, StringComparison.Ordinal))
+                                    {
+                                        reader.Read();
+                                        headers = _jsonSerializer.Deserialize<Dictionary<string, string>>(reader);
```
JamesNK: This will be faster manually read with `JsonReader`.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -66,70 +65,183 @@ public void WriteMessage(HubMessage message, Stream output)
...
+                                        }
+                                    }
+                                    else if (string.Equals(memberName, ItemPropertyName, StringComparison.Ordinal))
+                                    {
+                                        reader.Read();
+
+                                        if (string.IsNullOrEmpty(invocationId))
+                                        {
+                                            // If we don't have an invocation id then we need to store it as a JToken so we can parse it later
+                                            itemToken = JToken.Load(reader);
```
JamesNK: Maybe share the resultToken/result and itemToken/item variables. A message can't have both I believe.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
davidfowl: I bet but headers aren't in our main scenario and I didn't want to think about it. I guess it isn't that bad since we're just parsing string ->string. Still I don't think it's worth it.

JamesNK:
I like it. The hybrid approach is exactly what I was thinking of 👍

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
davidfowl: Yea, I just wanted to be explicit. I'll have to check if item is JToken in the methods that consume them.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
davidfowl: I can try it

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
davidfowl: Can you clarify? What exactly would I change the code to?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -261,95 +373,139 @@ private static void WriteMessageType(JsonTextWriter writer, int type)
...
+            {
+                throw new InvalidDataException($"Missing required property '{InvocationIdPropertyName}'.");
+            }
 
-            var args = JsonUtils.GetRequiredProperty<JArray>(json, ArgumentsPropertyName, JTokenType.Array);
+            return new CancelInvocationMessage(invocationId);
+        }
 
-            var paramTypes = binder.GetParameterTypes(target);
+        private HubMessage BindCompletionMessage(string invocationId, string error, object result, JToken resultToken, IInvocationBinder binder)
```
JamesNK: Right now you're always passing both processed result and raw result to these bind methods which is fine but I have a different approach to float.

What do you think about processing the raw result, if it exists, when we read the invocationId? e.g. when the invocationId is read we check to see if resultToken has a value and turn the resultToken into the result then. When we come to the bind we don't need to pass both in.

I don't have a huge preference between the two.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
davidfowl: Is it basically the default case?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
JamesNK: A sample I prepared earlier https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Converters/ExpandoObjectConverter.cs#L113-L141

Change the dictionary type line 115 and the change the `ReadValue` line 129 to a check that the token type is a string and then set the `reader.Value` into the dictionary.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
JamesNK: Example of simply reading an object here https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Converters/ExpandoObjectConverter.cs#L113-L141

The important thing is if you don't care about a property you need to `reader.Skip()` to completely read past its value. Then the next read will either be the next property name or the end object.

The code I linked to is safer than what you're doing. Your current code wouldn't error if the object suddenly cut off with no end object, e.g.

```
{
   "type":1,
```

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
davidfowl: Will do.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
JamesNK: ok!

davidfowl: I changed it 😄. Will see if it affects perf.

davidfowl: Changing it back

davidfowl:
@JamesNK updated based on your feedback

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -66,71 +64,273 @@ public void WriteMessage(HubMessage message, Stream output)
...
+                                    case TargetPropertyName:
+                                        target = reader.ReadAsString();
+                                        break;
+                                    case ErrorPropertyName:
+                                        error = reader.ReadAsString();
+                                        break;
+                                    case ResultPropertyName:
+                                        if (!reader.Read())
+                                        {
+                                            throw new JsonReaderException("Unexpected end when reading JSON");
```
JamesNK: Wrap this reader + check for content in a helper method called CheckRead and reuse

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -66,71 +64,273 @@ public void WriteMessage(HubMessage message, Stream output)
...
+                        {
+                            case JsonToken.PropertyName:
+                                string memberName = reader.Value.ToString();
+
+                                switch (memberName)
+                                {
+                                    case TypePropertyName:
+                                        var messageType = reader.ReadAsInt32();
+
+                                        if (messageType == null)
```
JamesNK: This test is already be done at the end. Just assign result to `type`

JamesNK:
Looks good to me. I think the tests could do with some more scenarios, e.g. unfinished content, incorrect content, content out of order, but other than that it looks good to me.

davidfowl:
Closed in favor of https://github.com/aspnet/SignalR/pull/1635

davidfowl:
Thanks for the feedback @JamesNK

PR #1641, opened 2018-03-19
The payload property was never used

davidfowl:


davidfowl:
I hate when we branch 😄

PR #1643, opened 2018-03-19
Use ActivatorUtlities.CreateFactory instead of CreateInstance

### LINKED ISSUE #1570: Use ActivatorUtilities.CreateFactory instead of CreateInstance
@pakrym made me aware of the `CreateFactory` method which should be faster that `CreateInstance` especially when creating instances many times (hub creation per invocation). And of course we can cache the factory method.

https://github.com/aspnet/SignalR/blob/4a4efe0f39477dff01bdff5006df477498e5e6f2/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubActivator.cs#L28

davidfowl:
- Turns out CreateFactory is much much faster
- Added a benchmark for hub activation

## Before

```
 Method |     Mean |    Error |   StdDev |        Op/s |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|------------:|-------:|----------:|
 Create | 710.7 ns | 14.40 ns | 39.89 ns | 1,406,988.1 | 0.0153 |     224 B |
```

## After

```
  Method |     Mean |    Error |   StdDev |         Op/s |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------------:|-------:|----------:|
 Create | 76.34 ns | 3.347 ns | 9.386 ns | 13,099,035.7 | 0.0037 |      48 B |
```

## Lazy

```
 Method |     Mean |    Error |   StdDev |   Median |         Op/s |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|---------:|-------------:|-------:|----------:|
 Create | 65.39 ns | 1.480 ns | 4.102 ns | 63.73 ns | 15,292,741.1 | 0.0037 |      48 B |
```

Fixes #1570

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubActivator.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubActivator.cs
@@ -9,8 +9,8 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class DefaultHubActivator<THub> : IHubActivator<THub> where THub: Hub
     {
-        // Object factory for all this Hub type
-        private static ObjectFactory _objectFactory = ActivatorUtilities.CreateFactory(typeof(THub), Type.EmptyTypes);
+        // Object factory for THub instances
+        private static Lazy<ObjectFactory> _objectFactory = new Lazy<ObjectFactory>(() => ActivatorUtilities.CreateFactory(typeof(THub), Type.EmptyTypes));
         private readonly IServiceProvider _serviceProvider;
         private bool? _created;
 
@@ -27,7 +27,7 @@ namespace Microsoft.AspNetCore.SignalR
             var hub = _serviceProvider.GetService<THub>();
             if (hub == null)
             {
-                hub = (THub)_objectFactory(_serviceProvider, Array.Empty<object>());
+                hub = (THub)_objectFactory.Value(_serviceProvider, Array.Empty<object>());
                 _created = true;
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Core/DefaultHubActivator.cs
```diff
@@ -9,6 +9,8 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class DefaultHubActivator<THub> : IHubActivator<THub> where THub: Hub
     {
+        // Object factory for all this Hub type
+        private static ObjectFactory _objectFactory = ActivatorUtilities.CreateFactory(typeof(THub), Type.EmptyTypes);
```
JamesNK: This is a pretty heavy operation to have in a static constructor. Also if the hub type is registered with the service provider then we'll never even use it. IMO create it when it is needed

davidfowl: I looked at that, I can make it lazy.

davidfowl:
🆙 📅

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubActivator.cs
```diff
@@ -9,6 +9,8 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class DefaultHubActivator<THub> : IHubActivator<THub> where THub: Hub
     {
+        // Object factory for THub instances
+        private static Lazy<ObjectFactory> _objectFactory = new Lazy<ObjectFactory>(() => ActivatorUtilities.CreateFactory(typeof(THub), Type.EmptyTypes));
```
JamesNK: readonly

pakrym:
Why did you need full on `Lazy<T>`? Null check would be just fine.

davidfowl:
> Why did you need full on Lazy<T>? Null check would be just fine.

You're right. I guess worst case we would have compiled a few times. Meh.

Issue #1475, opened 2018-02-21
Low level IConnection refactoring


davidfowl:
So far the client's internal implementation has been changed to use pipelines but we want to move to a model where the underlying connection is just a raw stream and some metadata instead of an evented reusable connection. Connection reusability will be implemented at the hub layer instead of the protocol layer. As part of this refactoring we'd also like to find a way to unify the client and server low level connection objects to enable more complex scenarios like server to server, peer to peer or proxying with extremely low overhead. This also makes it easier to unify the parsing and removes the event queue and other things from the client. 

The work was started here https://github.com/aspnet/SignalR/pull/1446 but was closed due to connection reusability problems.

Here are some related issues

- https://github.com/aspnet/SignalR/issues/1453
- https://github.com/aspnet/SignalR/issues/816
- https://github.com/aspnet/SignalR/issues/828
- https://github.com/aspnet/SignalR/issues/1450

JamesNK:
re: https://github.com/aspnet/SignalR/issues/1633

IConnection currently exposes `SendCore(byte[] data, CancellationToken cancellationToken)`. To optimize some sending scenarios we want to write to `Output` directly. Also want to think about writing `Memory<byte>` instead of an array so we can pass a section of an existing array to be written (although if `Output` was exposed we could just use that).

davidfowl:
@JamesNK we're going to be exposing pipes directly from the IConnection.

anurse:
This is done!

PR #1651, opened 2018-03-19
Make HubConnection re-create the HttpConnection instead of restarting it

anurse:
Part of #1475 

With this change, `HttpConnection` is left mostly unchanged (it is restartable), however `HubConnection` does not use that logic. Instead it dumps and recreates the `HttpConnection` each time. This is part of moving the restart logic to `HubConnection`.

A few behavioral changes:
* We used to silently complete streams that were cancelled, I changed that to throw OperationCancelledException (or `TaskCancelledException` in the Completion Task of the Channel))
* We used to have `HubConnection.StopAsync` wait until the `HttpConnection.Closed` event fired. Now, we dispose the `HttpConnection` and just null it out. We no longer care about it, so we can just ignore the `Closed` event when it fires. Note that **if** the `Closed` event fires before `StopAsync` (because the server disconnected), we still proceed to shut down. We do this by including the `IConnection` instance that triggered the close in the `Closed` event. This is mostly temporary until we pipelinify the client, at which point we'll have a Completion task to check and `ReadAsync` calls to wait on.

This is **not** complete and I expect to close it and continue iterating, but I want to get some code out to look at it. This component of the change is basically finished.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -449,8 +449,9 @@ namespace Microsoft.AspNetCore.SignalR.Client
             if (!_receivedHandshakeResponse)
             {
                 // process handshake and return left over data to parse additional messages
-                if (!ProcessHandshakeResponse(ref currentData))
+                if (!ProcessHandshakeResponse(ref currentData, out var exception))
                 {
+                    await OnClosed(_connection, exception);
                     return;
                 }
 
@@ -494,15 +495,20 @@ namespace Microsoft.AspNetCore.SignalR.Client
                             DispatchInvocationStreamItemAsync(streamItem, irq);
                             break;
                         case CloseMessage close:
+                            // In theory, there's a race here. If a close message arrives, but the user
+                            // restarts the connection, BEFORE it is processed here, we could close the
+                            // new connection when calling OnClosed. I think that's OK since we'll be fixing
+                            // this in later phases
+
                             if (string.IsNullOrEmpty(close.Error))
                             {
                                 Log.ReceivedClose(_logger);
-                                Shutdown();
+                                await OnClosed(_connection, ex: null);
                             }
                             else
                             {
                                 Log.ReceivedCloseWithError(_logger, close.Error);
-                                Shutdown(new InvalidOperationException(close.Error));
+                                await OnClosed(_connection, new InvalidOperationException(close.Error));
                             }
                             break;
                         case PingMessage _:
@@ -521,7 +527,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
-        private bool ProcessHandshakeResponse(ref ReadOnlyMemory<byte> data)
+        private bool ProcessHandshakeResponse(ref ReadOnlyMemory<byte> data, out Exception exception)
         {
             HandshakeResponseMessage message;
 
@@ -539,15 +545,16 @@ namespace Microsoft.AspNetCore.SignalR.Client
             {
                 // shutdown if we're unable to read handshake
                 Log.ErrorReceivingHandshakeResponse(_logger, ex);
-                Shutdown(ex);
+                exception = ex;
                 return false;
             }
+
+            exception = null;
 
             if (!string.IsNullOrEmpty(message.Error))
             {
                 // shutdown if handshake returns an error
                 Log.HandshakeServerError(_logger, message.Error);
-                Shutdown();
                 return false;
             }
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -31,8 +31,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private readonly Func<IConnection> _connectionFactory;
         private readonly HubBinder _binder;
 
-        // All the below fields are protected by _connectionLock
-        private SemaphoreSlim _connectionLock = new SemaphoreSlim(1, 1);
+        // This lock protects the connection and the disposed flag to ensure changes to them are seen at the same time.
+        private readonly SemaphoreSlim _connectionLock = new SemaphoreSlim(1, 1);
         private bool _disposed;
         private IConnection _connection;
 
@@ -83,17 +83,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         public async Task StartAsync()
         {
             CheckDisposed();
-
-            await _connectionLock.WaitAsync();
-
-            try
-            {
-                await StartAsyncCore().ForceAsync();
-            }
-            finally
-            {
-                _connectionLock.Release();
-            }
+            await StartAsyncCore().ForceAsync();
         }
 
         private async Task TimeoutElapsed()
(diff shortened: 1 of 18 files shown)


anurse:
Rebased on the latest `release/2.1` (including @JamesNK 's fixes). Not as conflicty as I'd feared. Going to start on making HttpConnection non-restartable today. Please do take a look @davidfowl and let me know if this is mostly matching how you expected this to go :)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -25,18 +27,22 @@ public partial class HubConnection
 
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
-        private readonly IConnection _connection;
         private readonly IHubProtocol _protocol;
+        private readonly Func<IConnection> _connectionFactory;
         private readonly HubBinder _binder;
 
+        // All the below fields are protected by _connectionLock
+        private SemaphoreSlim _connectionLock = new SemaphoreSlim(1, 1);
```
JamesNK: readonly

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -48,45 +54,62 @@ public partial class HubConnection
...
             // Create the timer for timeout, but disabled by default (we enable it when started).
-            _timeoutTimer = new Timer(state => ((HubConnection)state).TimeoutElapsed(), this, Timeout.Infinite, Timeout.Infinite);
+            _timeoutTimer = new Timer(state => _ = ((HubConnection)state).TimeoutElapsed(), this, Timeout.Infinite, Timeout.Infinite);
         }
 
         public async Task StartAsync()
         {
+            CheckDisposed();
+
+            await _connectionLock.WaitAsync();
```
JamesNK: This await will resume back on the sync context. ConfigureAwait(false)? Or move connection lock inside StartAsyncCore.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -111,6 +134,20 @@ private void ResetTimeoutTimer()
...
+
+            if (_connection != null)
+            {
+                // We have an existing connection!
+                throw new InvalidOperationException($"The '{nameof(StartAsync)}' method cannot be called if the connection has already been started.");
+            }
+
+            _connection = _connectionFactory();
+            _connection.OnReceived((data, state) => ((HubConnection)state).OnDataReceivedAsync(data), this);
+            _connection.Closed += (c, e) => _ = OnClosed(c, e);
```
JamesNK: What is underscore here? Maybe place a comment with what is going on for the uninformed reader.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -128,23 +165,85 @@ private void ResetTimeoutTimer()
...
+        /// <summary>
+        /// Requests that the connection be stopped. This task completes when the stop request has been processed,
+        /// the connection is not fully stopped until the <see cref="Closed"/> event is raised.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> that completes when connection shutdown has begun.</returns>
+        public async Task StopAsync()
+        {
+            CheckDisposed();
+
+            await _connectionLock.WaitAsync();
```
JamesNK: Sync context thing again. Same comment for all public methods with await

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -161,66 +260,80 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
         {
+            CheckDisposed();
+
             return await StreamAsChannelAsyncCore(methodName, returnType, args, cancellationToken).ForceAsync();
         }
 
         private async Task<ChannelReader<object>> StreamAsChannelAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken)
         {
-            if (!_startCalled)
+            await _connectionLock.WaitAsync();
```
JamesNK: Does this mean only one thread can be inside stream/invoke/etc at a time? Won't this negatively effect performance?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -320,8 +449,9 @@ private Task InvokeStreamCore(string methodName, InvocationRequest irq, object[]
             if (!_receivedHandshakeResponse)
             {
                 // process handshake and return left over data to parse additional messages
-                if (!ProcessHandshakeResponse(ref currentData))
+                if (!ProcessHandshakeResponse(ref currentData, out var exception))
                 {
+                    await OnClosed(_connection, exception);
```
JamesNK: Is this to keep closing more centralized?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -542,6 +711,9 @@ private void ThrowIfConnectionTerminated(string invocationId)
 
         private void AddInvocation(InvocationRequest irq)
         {
+            SafeAssert(_connectionLock.CurrentCount == 0, "We're not in the Connection Lock!");
```
JamesNK: Should we have a helper method that wraps SafeAssert for checking the connection lock? This is copy/pasted 9 times.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -74,101 +48,128 @@ public class HubConnectionTests
...
+            await hubConnection.StopAsync().OrTimeout();
             Assert.Null(await closedEventTcs.Task);
         }
 
         [Fact]
         public async Task CannotCallInvokeOnNotStartedHubConnection()
         {
             var connection = new TestConnection();
-            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+            var hubConnection = new HubConnection(() => connection, new JsonHubProtocol(), new LoggerFactory());
```
JamesNK: Create a helper method for creating a HubConnection with defaulted args and reuse when creating a HubConnection in unit tests. That means if the HubConnection ctor changes then we don't need to update dozens of unit tests.

test/Microsoft.AspNetCore.SignalR.Client.Tests/SyncPoint.cs
```diff
@@ -0,0 +1,76 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+// Possibly useful as a general-purpose async testing helper?
+public class SyncPoint
+        {
+            private TaskCompletionSource<object> _atSyncPoint = new TaskCompletionSource<object>();
```
JamesNK: Formatting

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Yeah, I'll move the locks inside the ForceAsync logic.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: We already serialize writes further down. Pipelines are single-producer (they aren't thread-safe). But because of how Pipelines work, this will only go async when the server is applying backpressure to us (i.e. our local buffers are full), so we want to reduce the throughput there anyway.

test/Microsoft.AspNetCore.SignalR.Client.Tests/SyncPoint.cs
anurse: C#'s "Move type to file" refactoring derped it up :)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Yeah, I was seeing that too. I'll probably wrap it up in a helper.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Adding the comment now. We generally do this pattern instead of `async void` for fire-and-forget tasks.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Yeah, and because `OnClosed` is async and `ProcessHandshakeResponse` has a `ref` param ;)

anurse:
@davidfowl and I decided to make this a mergable PR. He still needs to take a look, but we're not going to wait on getting it all in one PR. The tests should all be green here (I see build failures atm, which I'll look in to, but it passes on my machine)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -48,47 +54,54 @@ public partial class HubConnection
...
             _binder = new HubBinder(this);
             _protocol = protocol;
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
-            _connection.OnReceived((data, state) => ((HubConnection)state).OnDataReceivedAsync(data), this);
-            _connection.Closed += Shutdown;
 
             // Create the timer for timeout, but disabled by default (we enable it when started).
-            _timeoutTimer = new Timer(state => ((HubConnection)state).TimeoutElapsed(), this, Timeout.Infinite, Timeout.Infinite);
+            _timeoutTimer = new Timer(state => _ = ((HubConnection)state).TimeoutElapsed(), this, Timeout.Infinite, Timeout.Infinite);
```
davidfowl: ?

davidfowl: Oh, wow, thats ugly 😄

anurse: It's that or `async void` :)

davidfowl: yeah

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -410,35 +544,77 @@ private bool ProcessHandshakeResponse(ref ReadOnlyMemory<byte> data)
...
+
+            // Dispose the timer AFTER shutting down the connection.
+            _timeoutTimer.Dispose();
+
+            CancelOutstandingInvocations(exception);
+
+            // Fire our closed event.
+            try
+            {
+                Closed?.Invoke(exception);
```
davidfowl: Run this outside of the lock.

anurse:
Flip-flopping again. As I'm working on the next phase, this code is churning enough that it seems like this PR isn't really necessary. I'll send a new one out with the bulk of the work in it.

PR #1653, opened 2018-03-20
Cleanup rogue timeout in TS client

BrennanConroy:


client-ts/signalr/src/HubConnection.ts
```diff
@@ -141,6 +139,8 @@ export class HubConnection {
 
         this.logger.log(LogLevel.Information, `Using HubProtocol '${this.protocol.name}'.`);
 
+        // defensively cleanup timeout in case we receive a message from the server before we finish start
+        this.cleanupTimeout();
```
anurse: So, just to clarify what was happening: We were timing out because a message was available before we even started the timeout, we processed that message, and then set the timeout (in the receive callback), but then we immediately overwrite the handle here without checking if it is set here, thus never calling `clearTimeout` on that first handle and causing the callback to fire. Right?

BrennanConroy: Correct, thus the test

davidfowl:
Nice catch!

client-ts/signalr/src/HubConnection.ts
anurse: Cool, just wanted to make sure I understood. Appreciate the test.

PR #1654, opened 2018-03-20
Fix LongPolling retry after poll timeout

BrennanConroy:
An alternative, we could remove the `extends Error` part

### DIFF
--- a/client-ts/signalr/src/Errors.ts
+++ b/client-ts/signalr/src/Errors.ts
@@ -2,15 +2,29 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 export class HttpError extends Error {
+    // tslint:disable-next-line:variable-name
+    private __proto__: Error;
     public statusCode: number;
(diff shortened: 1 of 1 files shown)


client-ts/signalr/src/Errors.ts
```diff
@@ -2,15 +2,29 @@
...
+    private __proto__: Error;
     public statusCode: number;
     constructor(errorMessage: string, statusCode: number) {
+        const trueProto = new.target.prototype;
         super(errorMessage);
         this.statusCode = statusCode;
+
+        // Workaround issue in Typescript compiler
+        // https://github.com/Microsoft/TypeScript/issues/13965#issuecomment-278570200
+        this.__proto__ = trueProto;
```
anurse: Ugh, looks ugly but I think you're right, this is the right approach. We should continue extending `Error`.

davidfowl: wtf is this.

BrennanConroy: I linked the issue

PR #1658, opened 2018-03-20
Made HubEndPoint derive from EndPoint

davidfowl:


test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -314,7 +314,7 @@ public class HubEndpointTests
             var hubProtocolMock = new Mock<IHubProtocol>();
             hubProtocolMock.Setup(m => m.Name).Returns("CustomProtocol");
 
-            dynamic endPoint = HubEndPointTestUtils.GetHubEndpoint(typeof(HubT));
+            EndPoint endPoint = HubEndPointTestUtils.GetHubEndpoint(typeof(HubT));
```
JamesNK: Can these be var now?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
```diff
@@ -5,10 +5,6 @@
     <RootNamespace>Microsoft.AspNetCore.SignalR.Tests</RootNamespace>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
-    <PackageReference Include="Microsoft.CSharp" Version="$(MicrosoftCSharpPackageVersion)" />
```
JamesNK: Can it also be removed from Microsoft.AspNetCore.SignalR.Tests.csproj?

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
JamesNK: Also the variable results from ConnectAsync can be var instead of task.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
davidfowl: Maybe, we do have things that test the dynamic hubs so it may be required.

davidfowl: Yea, I'm going to leave that one in https://github.com/aspnet/SignalR/blob/8d4945df8870e74e1ffa722f28215dcd077da665/test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTestUtils/Hubs.cs#L171

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
davidfowl: Done!

davidfowl:
> I always wondered why you removed it :).

Minimal Dependencies

anurse:
> Minimal Dependencies

PR #1660, opened 2018-03-20
Improved allocations and throughput for broadcast scenarios

davidfowl:
- Don't allocate when enumerating connections
- Don't allocate tasks unless we truly go async
- Don't get the timestamp, just write the pings always (if there's no ongoing write)
- Track the time since last keep alive write instead of the last write (this needs some reviewing)

## Old

image

```
       Method | Connections | Protocol |         Mean |       Error |        StdDev |      Op/s |   Gen 0 |  Gen 1 | Allocated |
------------- |------------ |--------- |-------------:|------------:|--------------:|----------:|--------:|-------:|----------:|
 SendAsyncAll |           1 |     json |     7.545 us |   0.4048 us |     1.1936 us | 132,532.1 |  0.3204 |      - |   6.62 KB |
 SendAsyncAll |           1 |  msgpack |     5.492 us |   0.2845 us |     0.8343 us | 182,092.0 |  0.0610 |      - |   1.53 KB |
 SendAsyncAll |          10 |     json |    24.614 us |   1.3042 us |     3.7421 us |  40,626.6 |  0.2747 |      - |   7.27 KB |
 SendAsyncAll |          10 |  msgpack |    17.706 us |   0.3166 us |     0.3769 us |  56,476.9 |  0.0916 |      - |   2.17 KB |
 SendAsyncAll |        1000 |     json | 3,620.429 us | 371.4354 us | 1,095.1858 us |     276.2 | 11.7188 | 3.9063 |  593.6 KB |
 SendAsyncAll |        1000 |  msgpack | 2,705.518 us | 103.2645 us |   291.2594 us |     369.6 | 15.6250 | 3.9063 | 516.87 KB |
```

## New

image

```
        Method | Connections | Protocol |         Mean |       Error |      StdDev |       Median |      Op/s |   Gen 0 |  Gen 1 | Allocated |
------------- |------------ |--------- |-------------:|------------:|------------:|-------------:|----------:|--------:|-------:|----------:|
 SendAsyncAll |           1 |     json |     6.614 us |   0.5787 us |   1.7064 us |     6.025 us | 151,184.7 |  0.3052 |      - |   6.33 KB |
 SendAsyncAll |           1 |  msgpack |     3.705 us |   0.1580 us |   0.4585 us |     3.576 us | 269,920.6 |  0.0496 |      - |   1.24 KB |
 SendAsyncAll |          10 |     json |    16.866 us |   0.9062 us |   2.6292 us |    17.083 us |  59,291.6 |  0.2899 | 0.0153 |   6.73 KB |
 SendAsyncAll |          10 |  msgpack |    14.836 us |   0.5637 us |   1.6443 us |    14.572 us |  67,403.7 |  0.0458 |      - |   1.36 KB |
 SendAsyncAll |        1000 |     json | 2,291.156 us |  75.0624 us | 212.9395 us | 2,282.967 us |     436.5 | 11.7188 | 3.9063 | 404.08 KB |
 SendAsyncAll |        1000 |  msgpack | 2,310.744 us | 116.0950 us | 333.0982 us | 2,209.508 us |     432.8 |  7.8125 |      - | 262.68 KB |
```

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
@@ -48,6 +48,8 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
                 var hubConnection = new HubConnectionContext(connection, Timeout.InfiniteTimeSpan, NullLoggerFactory.Instance);
                 hubConnection.Protocol = protocol;
                 _hubLifetimeManager.OnConnectedAsync(hubConnection).Wait();
+
+                _ = ConsumeAsync(connection.Application);
             }
 
             _hubContext = new HubContext<Hub>(_hubLifetimeManager);
@@ -57,6 +59,27 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
         public Task SendAsyncAll()
         {
             return _hubContext.Clients.All.SendAsync("Method");
+        }
+
+        // Consume the data written to the transport
+        private static async Task ConsumeAsync(IDuplexPipe application)
+        {
+            while (true)
+            {
+                var result = await application.Input.ReadAsync();
+                var buffer = result.Buffer;
+
+                if (!buffer.IsEmpty)
+                {
+                    application.Input.AdvanceTo(buffer.End);
+                }
+                else if (result.IsCompleted)
+                {
+                    break;
+                }
+            }
+
+            application.Input.Complete();
         }
     }
 }
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
@@ -40,9 +40,10 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
                 protocol = new MessagePackHubProtocol();
             }
 
+            var options = new PipeOptions();
             for (var i = 0; i < Connections; ++i)
             {
-                var pair = DuplexPipe.CreateConnectionPair(PipeOptions.Default, PipeOptions.Default);
+                var pair = DuplexPipe.CreateConnectionPair(options, options);
                 var connection = new DefaultConnectionContext(Guid.NewGuid().ToString(), pair.Application, pair.Transport);
                 var hubConnection = new HubConnectionContext(connection, Timeout.InfiniteTimeSpan, NullLoggerFactory.Instance);
                 hubConnection.Protocol = protocol;
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
@@ -69,18 +69,37 @@ namespace Microsoft.AspNetCore.SignalR
 
         public override Task SendAllAsync(string methodName, object[] args)
         {
-            return SendAllWhere(methodName, args, c => true);
-        }
+            List<Task> tasks = null;
+            var message = CreateInvocationMessage(methodName, args);
 
-        private Task SendAllWhere(string methodName, object[] args, Func<HubConnectionContext, bool> include)
-        {
-            var count = _connections.Count;
-            if (count == 0)
+            foreach (var connection in _connections)
+            {
+                var task = connection.WriteAsync(message);
+
+                if (!task.IsCompleted)
+                {
+                    if (tasks == null)
+                    {
+                        tasks = new List<Task>();
+                    }
+
+                    tasks.Add(task);
+                }
+            }
+
+            // No async
+            if (tasks == null)
             {
                 return Task.CompletedTask;
             }
 
-            var tasks = new List<Task>(count);
+            // Some connections are slow
+            return Task.WhenAll(tasks);
+        }
+
+        private Task SendAllWhere(string methodName, object[] args, Func<HubConnectionContext, bool> include)
+        {
+            List<Task> tasks = null;
             var message = CreateInvocationMessage(methodName, args);
 
             foreach (var connection in _connections)
@@ -90,9 +109,25 @@ namespace Microsoft.AspNetCore.SignalR
                     continue;
                 }
 
-                tasks.Add(SafeWriteAsync(connection, message));
+                var task = connection.WriteAsync(message);
+
+                if (!task.IsCompleted)
+                {
+                    if (tasks == null)
+                    {
+                        tasks = new List<Task>();
+                    }
+
+                    tasks.Add(task);
+                }
             }
 
+            if (tasks == null)
+            {
+                return Task.CompletedTask;
+            }
+
+            // Some connections are slow
             return Task.WhenAll(tasks);
         }
 
@@ -112,7 +147,7 @@ namespace Microsoft.AspNetCore.SignalR
 
             var message = CreateInvocationMessage(methodName, args);
 
-            return SafeWriteAsync(connection, message);
+            return connection.WriteAsync(message);
         }
 
         public override Task SendGroupAsync(string groupName, string methodName, object[] args)
@@ -126,7 +161,7 @@ namespace Microsoft.AspNetCore.SignalR
             if (group != null)
             {
                 var message = CreateInvocationMessage(methodName, args);
-                var tasks = group.Values.Select(c => SafeWriteAsync(c, message));
+                var tasks = group.Values.Select(c => c.WriteAsync(message));
                 return Task.WhenAll(tasks);
             }
 
@@ -149,7 +184,7 @@ namespace Microsoft.AspNetCore.SignalR
                 var group = _groups[groupName];
                 if (group != null)
                 {
-                    tasks.Add(Task.WhenAll(group.Values.Select(c =>  SafeWriteAsync(c, message))));
+                    tasks.Add(Task.WhenAll(group.Values.Select(c => c.WriteAsync(message))));
                 }
             }
 
@@ -168,7 +203,7 @@ namespace Microsoft.AspNetCore.SignalR
             {
                 var message = CreateInvocationMessage(methodName, args);
                 var tasks = group.Values.Where(connection => !excludedIds.Contains(connection.ConnectionId))
-                    .Select(c => SafeWriteAsync(c, message));
+                    .Select(c => c.WriteAsync(message));
                 return Task.WhenAll(tasks);
             }
 
@@ -221,31 +256,6 @@ namespace Microsoft.AspNetCore.SignalR
             {
                 return userIds.Contains(connection.UserIdentifier);
             });
-        }
-
-        // This method is to protect against connections throwing synchronously when writing to them and preventing other connections from being written to
-        private async Task SafeWriteAsync(HubConnectionContext connection, InvocationMessage message)
-        {
-            try
-            {
-                await connection.WriteAsync(message);
-            }
-            // This exception isn't interesting to users
-            catch (Exception ex)
-            {
-                Log.FailedWritingMessage(_logger, ex);
-            }
-        }
-
-        private static class Log
-        {
-            private static readonly Action<ILogger, Exception> _failedWritingMessage =
-                LoggerMessage.Define(LogLevel.Warning, new EventId(1, "FailedWritingMessage"), "Failed writing message.");
-
-            public static void FailedWritingMessage(ILogger logger, Exception exception)
-            {
-                _failedWritingMessage(logger, exception);
-            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -74,20 +74,71 @@ namespace Microsoft.AspNetCore.SignalR
 
         public int? LocalPort => Features.Get<IHttpConnectionFeature>()?.LocalPort;
 
-        public virtual async Task WriteAsync(HubMessage message)
+        public virtual Task WriteAsync(HubMessage message)
         {
-            await _writeLock.WaitAsync();
+            // We were unable to get the lock so take the slow async path of waiting for the semaphore
+            if (!_writeLock.Wait(0))
+            {
+                return WriteSlowAsync(message);
+            }
 
+            // This method should never throw synchronously
+            var task = WriteCore(message);
+
+            // The write didn't complete synchronously so await completion
+            if (!task.IsCompleted)
+            {
+                return CompleteWriteAsync(task);
+            }
+
+            // The write failed so observe the exception
+            if (task.IsFaulted)
+            {
+                try
+                {
+                    task.GetAwaiter().GetResult();
+                }
+                catch (Exception ex)
+                {
+                    Log.FailedWritingMessage(_logger, ex);
+                }
+            }
+
+            // Otherwise, release the lock
+            _writeLock.Release();
+
+            return Task.CompletedTask;
+        }
+
+        private ValueTask<FlushResult> WriteCore(HubMessage message)
+        {
             try
             {
-                // This will internally cache the buffer for each unique HubProtocol/DataEncoder combination
+                // This will internally cache the buffer for each unique HubProtocol
                 // So that we don't serialize the HubMessage for every single connection
                 var buffer = message.WriteMessage(Protocol);
+
                 _connectionContext.Transport.Output.Write(buffer);
 
-                Interlocked.Exchange(ref _lastSendTimestamp, Stopwatch.GetTimestamp());
+                return _connectionContext.Transport.Output.FlushAsync();
+            }
+            catch (Exception ex)
+            {
+                Log.FailedWritingMessage(_logger, ex);
 
-                await _connectionContext.Transport.Output.FlushAsync();
+                return new ValueTask<FlushResult>(new FlushResult(isCanceled: false, isCompleted: true));
+            }
+        }
+
+        private async Task CompleteWriteAsync(ValueTask<FlushResult> task)
+        {
+            try
+            {
+                await task;
+            }
+            catch (Exception ex)
+            {
+                Log.FailedWritingMessage(_logger, ex);
             }
             finally
             {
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -69,18 +69,37 @@ public override Task RemoveGroupAsync(string connectionId, string groupName)
...
 
-        private Task SendAllWhere(string methodName, object[] args, Func<HubConnectionContext, bool> include)
-        {
-            var count = _connections.Count;
-            if (count == 0)
+            foreach (var connection in _connections)
+            {
+                var task = connection.WriteAsync(message);
+
+                if (!task.IsCompleted)
```
JamesNK: I think you want `IsCompletedSuccessfully`. It was added in netcoreapp2.0

Helper method from Json.NET:
```
        public static bool IsCompletedSucessfully(this Task task)
        {
            // IsCompletedSucessfully is the faster method, but only currently exposed on .NET Core 2.0
#if NETCOREAPP2_0
            return task.IsCompletedSucessfully;
#else
            return task.Status == TaskStatus.RanToCompletion;
#endif
        }
```

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -90,9 +109,25 @@ private Task SendAllWhere(string methodName, object[] args, Func<HubConnectionCo
                     continue;
                 }
 
-                tasks.Add(SafeWriteAsync(connection, message));
+                var task = connection.WriteAsync(message);
+
+                if (!task.IsCompleted)
```
JamesNK: IsCompletedSuccessfully?

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -126,7 +161,7 @@ public override Task SendGroupAsync(string groupName, string methodName, object[
             if (group != null)
             {
                 var message = CreateInvocationMessage(methodName, args);
-                var tasks = group.Values.Select(c => SafeWriteAsync(c, message));
+                var tasks = group.Values.Select(c => c.WriteAsync(message));
```
JamesNK: Write a helper method for calling WriteAsync on a group of connection contexts? Can avoid the overhead of LINQ and closure.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -149,7 +184,7 @@ public override Task SendGroupsAsync(IReadOnlyList<string> groupNames, string me
                 var group = _groups[groupName];
                 if (group != null)
                 {
-                    tasks.Add(Task.WhenAll(group.Values.Select(c =>  SafeWriteAsync(c, message))));
+                    tasks.Add(Task.WhenAll(group.Values.Select(c => c.WriteAsync(message))));
```
JamesNK: Use collection helper method above here.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
```diff
@@ -168,7 +203,7 @@ public override Task SendGroupExceptAsync(string groupName, string methodName, o
             {
                 var message = CreateInvocationMessage(methodName, args);
                 var tasks = group.Values.Where(connection => !excludedIds.Contains(connection.ConnectionId))
-                    .Select(c => SafeWriteAsync(c, message));
+                    .Select(c => c.WriteAsync(message));
```
JamesNK: Reuse helper method++ (and it probably take an `IEnumerable<HubConnectionContext>` to support consuming the LINQ result here.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -74,44 +74,123 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+            if (!_writeLock.Wait(0))
+            {
+                return WriteSlowAsync(message);
+            }
+
+            // This method should never throw synchronously
+            var task = WriteCore(message);
+
+            // The write didn't complete synchronously so await completion
+            if (!task.IsCompleted)
```
JamesNK: IsCompletedSuccessfully

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -74,44 +74,123 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+            {
+                return CompleteWriteAsync(task);
+            }
+
+            // The write failed so observe the exception
+            if (task.IsFaulted)
```
JamesNK: I think if you use `IsCompletedSuccessfully` above then you'll get the exception observed automatically when the task is passed to `CompleteWriteAsync` and you don't need this. It is slightly slower but it is an exceptional situation so perf doesn't matter as much.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionList.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public class HubConnectionList : IReadOnlyCollection<HubConnectionContext>
+    public class HubConnectionList
```
JamesNK: Having a public class that is called a list but doesn't implement a collection interface feels wrong.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionList.cs
```diff
@@ -32,14 +32,29 @@ public void Remove(HubConnectionContext connection)
...
         {
-            return _connections.Values.GetEnumerator();
+            return new Enumerator(this);
         }
 
-        IEnumerator IEnumerable.GetEnumerator()
+        public struct Enumerator : IEnumerator<HubConnectionContext>
         {
-            return GetEnumerator();
+            private IEnumerator<KeyValuePair<string, HubConnectionContext>> _enumerator;
```
JamesNK: readonly? Also can the struct be readonly?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -308,6 +386,9 @@ private static class Log
             private static readonly Action<ILogger, Exception> _handshakeFailed =
                 LoggerMessage.Define(LogLevel.Error, new EventId(5, "HandshakeFailed"), "Failed connection handshake.");
 
+            private static readonly Action<ILogger, Exception> _failedWritingMessage =
+                LoggerMessage.Define(LogLevel.Trace, new EventId(6, "FailedWritingMessage"), "Failed writing message.");
```
JamesNK: Include the message type in the log message? Also the level should probably be higher than LogLevel.Trace.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
davidfowl: I'm not fixing LINQ in this change. We need to add benchmarks for groups and the other overloads. I almost went down this rabbit hole but resisted as part of this change.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
davidfowl: I think it needs to be trace. Nobody should care about write failures. We saw lots of feedback from kestrel and dialed this down. It's something that will happen all the time

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionList.cs
davidfowl: Meh.

src/Microsoft.AspNetCore.SignalR.Core/DefaultHubLifetimeManager.cs
davidfowl: See comment about fixing the LINQ.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionList.cs
davidfowl: struct based enumerators are usually mutable structs but this *might* since it's just wrapping the dictionary enumerator.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -74,44 +74,123 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+            _writeLock.Release();
+
+            return Task.CompletedTask;
+        }
 
+        private ValueTask<FlushResult> WriteCore(HubMessage message)
+        {
             try
             {
-                // This will internally cache the buffer for each unique HubProtocol/DataEncoder combination
```
anurse: RIP DataEncoder

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -74,44 +74,123 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+
+        private async Task TryWritePingSlowAsync()
+        {
             try
             {
                 Debug.Assert(_cachedPingMessage != null);
-                _connectionContext.Transport.Output.Write(_cachedPingMessage);
 
-                Interlocked.Exchange(ref _lastSendTimestamp, Stopwatch.GetTimestamp());
+                _connectionContext.Transport.Output.Write(_cachedPingMessage);
```
anurse: Yeah, in reviewing the WebSockets code, this is basically how it's pings work. It sends them at the ping interval every time, regardless of if another message was sent. But it won't wait on the send lock.

A side effect of this, is that Long Poll requests will **always** complete in at least `KeepAliveInterval`. We could lower the timeout on the client to something nearer to this value in theory.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
anurse: I think this is `Debug`-worthy, but I don't have the context from Kestrel. My general rules have been:

* `Information` - Always useful to every user
* `Debug` - Sometimes useful to a human user investigating an issue
* `Trace` - Only ever useful to automation

This message seems like it would be useful to a human user to diagnose why messages aren't arriving on a client (it's an exceptional case, we just don't consider it a full-on error).

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -259,16 +338,15 @@ private void KeepAliveTick()
...
             {
                 // Haven't sent a message for the entire keep-alive duration, so send a ping.
                 // If the transport channel is full, this will fail, but that's OK because
                 // adding a Ping message when the transport is full is unnecessary since the
                 // transport is still in the process of sending frames.
-
                 _ = TryWritePingAsync();
-                Log.SentPing(_logger);
+
+                Interlocked.Exchange(ref _lastSendTimestamp, Stopwatch.GetTimestamp());
```
BrennanConroy: You still kept the second `GetTimestamp`! Call it once before the if statement and store it in a variable

davidfowl: Woops 😄

davidfowl:
I think we should go through and ValueTask-ify these APIs and re-measure.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
```diff
@@ -40,13 +40,16 @@ public void GlobalSetup()
...
             for (var i = 0; i < Connections; ++i)
             {
-                var pair = DuplexPipe.CreateConnectionPair(PipeOptions.Default, PipeOptions.Default);
+                var pair = DuplexPipe.CreateConnectionPair(options, options);
                 var connection = new DefaultConnectionContext(Guid.NewGuid().ToString(), pair.Application, pair.Transport);
                 var hubConnection = new HubConnectionContext(connection, Timeout.InfiniteTimeSpan, NullLoggerFactory.Instance);
                 hubConnection.Protocol = protocol;
                 _hubLifetimeManager.OnConnectedAsync(hubConnection).Wait();
+
+                _ = ConsumeAsync(connection.Application);
```
BrennanConroy: Did you run the before and after with this?

davidfowl: Yes.

test/Microsoft.AspNetCore.SignalR.Tests/DefaultHubLifetimeManagerTests.cs
```diff
@@ -110,54 +110,6 @@ public class DefaultHubLifetimeManagerTests
...
-                var connection = connectionMock.Object;
-
-                await manager.OnConnectedAsync(connection).OrTimeout();
-
-                await manager.SendConnectionAsync(connection.ConnectionId, "Hello", new object[] { "World" }).OrTimeout();
-            }
-        }
-
-        [Fact]
-        public async Task SendAllAsyncSendsToAllConnectionsEvenWhenSomeFailToSend()
```
BrennanConroy: ???

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
anurse: Never mind, I forgot my own code. Keep-Alives aren't sent for long polling because of my favo**u**rite feature: `IConnectionInherentKeepAliveFeature`

test/Microsoft.AspNetCore.SignalR.Tests/DefaultHubLifetimeManagerTests.cs
davidfowl: The contract is that HubConnectionContext.WriteAsync never throws.

BrennanConroy: Which is what the test was testing

BrennanConroy: We've already regressed this behavior before, would be nice if we had a test to make sure we don't do it again

anurse: Nobody **ever** breaks a contract. That's like a solemn vow between engineers.

davidfowl: I need to update Redis as well but I left it as is for now

davidfowl: > We've already regressed this behavior before, would be nice if we had a test to make sure we don't do it again

What do you suggest? I can make HubConnectionContext.WriteAsync non virtual.

BrennanConroy:
MessagePack allocations went up slightly in the latest, how?

davidfowl:
> MessagePack allocations went up slightly in the latest, how?

I doubt it's consistent. I've seen strange things with these broadcast tests seems like it's possible to allocate more because more work is happening. If you look at the two memory profiles the `byte[]` allocated in each of the cases (exactly 1000 messages) are slightly different.

davidfowl:
Anything else? I think this is ready to go.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -74,44 +74,110 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+            {
+                await task;
+            }
+            catch (Exception ex)
+            {
+                Log.FailedWritingMessage(_logger, ex);
+            }
+            finally
+            {
+                _writeLock.Release();
```
JamesNK: Where is this writeLock acquired? There is no `WaitAsync` in this method, I don't see one in `WriteAsync` or `WriteCore`.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -74,44 +74,110 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+            var task = WriteCore(message);
+
+            // The write didn't complete synchronously so await completion
+            if (!task.IsCompletedSuccessfully)
+            {
+                return new ValueTask(CompleteWriteAsync(task));
+            }
+
+            // Otherwise, release the lock
+            _writeLock.Release();
```
JamesNK: Where is this writeLock acquired?

JamesNK: Ah, it is the `Wait` on line 80. Could you add a little bit of documentation, e.g.

`// Otherwise, release the lock acquired when entering WriteAsync`

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
JamesNK: Ah, it is the `Wait` on line 80. Could you add a little bit of documentation, e.g.

`// Release the lock acquired when entering WriteAsync`

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -74,44 +74,110 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
...
+            {
+                _writeLock.Release();
+            }
+        }
+
+        private async Task WriteSlowAsync(HubMessage message)
+        {
+            try
+            {
+                await _writeLock.WaitAsync();
```
JamesNK: Add a little bit of documentation

`// Failed to get the lock immediately when entering WriteAsync so await until it is available`

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionList.cs
JamesNK: Was it not `readonly`able?

davidfowl: Works

Issue #1648, opened 2018-03-19
Flaky tests: HubConnectionCanSendAndReceiveGroupMessages et al


ryanbrandenburg:
These tests [fail](http://aspnetci/viewLog.html?buildId=422519&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR) with the following message:
```
System.AggregateException : One or more errors occurred. (One or more errors occurred. (Connection refused)) (The following constructor parameters did not have matching fixture data: RedisServerFixture`1 serverFixture)
---- System.AggregateException : One or more errors occurred. (Connection refused)
-------- System.Net.Sockets.SocketException : Connection refused
---- The following constructor parameters did not have matching fixture data: RedisServerFixture`1 serverFixture

----- Inner Stack Trace #1 (System.AggregateException) -----
   at Microsoft.AspNetCore.SignalR.Tests.ServerFixture`1.StartServer() in /_/test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs:line 75
   at Microsoft.AspNetCore.SignalR.Tests.ServerFixture`1..ctor() in /_/test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs:line 45
   at Microsoft.AspNetCore.SignalR.Redis.Tests.RedisServerFixture`1.StartServer() in /_/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisServerFixture.cs:line 44
   at Microsoft.AspNetCore.SignalR.Redis.Tests.RedisServerFixture`1..ctor() in /_/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisServerFixture.cs:line 36
----- Inner Stack Trace -----
   at System.Net.Sockets.Socket.BeginConnectEx(EndPoint remoteEP, Boolean flowContext, AsyncCallback callback, Object state)
   at System.Net.Sockets.Socket.UnsafeBeginConnect(EndPoint remoteEP, AsyncCallback callback, Object state, Boolean flowContext)
   at System.Net.Sockets.Socket.BeginConnect(EndPoint remoteEP, AsyncCallback callback, Object state)
   at System.Net.Sockets.Socket.ConnectAsync(EndPoint remoteEP)
[...]

PR #1667, opened 2018-03-20
Properly wait for process exit for docker helper and log errors

BrennanConroy:
#1648

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
@@ -63,8 +63,8 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Tests
             // create and run docker container, remove automatically when stopped, map 6379 from the container to 6379 localhost
             // use static name 'redisTestContainer' so if the container doesn't get removed we don't keep adding more
             // use redis base docker image
-            // 10 second timeout to allow redis image to be downloaded
-            RunProcessAndThrowIfFailed(_path, $"run --rm -p 6379:6379 --name {_dockerContainerName} -d redis", logger, TimeSpan.FromSeconds(10));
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
```diff
@@ -125,7 +125,11 @@ private static int RunProcess(string fileName, string arguments, ILogger logger,
             process.BeginErrorReadLine();
             process.BeginOutputReadLine();
 
-            process.WaitForExit((int)timeout.TotalMilliseconds);
+            if (!process.WaitForExit((int)timeout.TotalMilliseconds))
+            {
+                process.Close();
+                logger.LogError("Closing process '{processName}' because it is running longer than the configured timeout.", fileName);
```
JamesNK: Will fileName be used to replace {processName} by the logging framework? Just wondering if the string was suppose to be interpolated.

BrennanConroy: Logging does in order replacement, and it's bad practice to interpolate strings passed to logging

anurse: Yeah, it's intentionally not an interpolated string. Logging uses the template to generate structure log data: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?tabs=aspnetcore2x#log-message-template

JamesNK: Ok. I recall a PR recently to make all the names in brackets pascal case. Should this log message follow that standard?

BrennanConroy: Sure, although it doesn't really matter for test code

PR #1671, opened 2018-03-21
Fix #1575 by removing use of Map

### LINKED ISSUE #1575: TypeScript version (esm) should be built with ES5
As you state that IE11 is supported I find it inconsistent that the TypeScript version is not built with ES5 as the target.

When I try to use @aspnet/signalr through 
`import { HubConnection } from "@aspnet/signalr";`
in get this error:

`ERROR in [at-loader] ./node_modules/@aspnet/signalr/dist/esm/HttpClient.d.ts:6:15
    TS2304: Cannot find name 'Map'.`

I am using webpack and my tsconfig.json is set to "target": "es5" as I need to support IE11.

anurse:
Turns out there was a runtime issue in #1575, not just a problem with TypeScript definitions. We use `Map`, which is not present in ECMAScript 5. We build for ES5, but TypeScript doesn't polyfill types, it only rewrites syntax. That means TypeScript was just emitting use of the `Map` object, even into the ES5 build. We have two options:

1. Add a `Map` polyfill to our build (yuck, more dependencies)
2. Don't use `Map`

This PR is option 2.

We don't really _need_ `Map`, it's just a slightly more "dictionary"-like API instead of just using JavaScript Objects.

[I had some spare cycles to do some work, but had only my Mac and it hasn't been cooperating when doing C# stuff, so I decided to grab this one because it was easy]

### DIFF
--- a/client-ts/signalr/src/HttpConnection.ts
+++ b/client-ts/signalr/src/HttpConnection.ts
@@ -83,8 +83,9 @@ export class HttpConnection implements IConnection {
                 let headers;
                 const token = this.options.accessTokenFactory();
                 if (token) {
-                    headers = new Map<string, string>();
-                    headers.set("Authorization", `Bearer ${token}`);
+                    headers = {
+                        ["Authorization"]: `Bearer ${token}`,
+                    };
                 }
 
                 const negotiatePayload = await this.httpClient.post(this.resolveNegotiateUrl(this.baseUrl), {
--- a/client-ts/tsconfig-base.json
+++ b/client-ts/tsconfig-base.json
@@ -13,6 +13,6 @@
         "noImplicitAny": true,
         "suppressImplicitAnyIndexErrors": true,
         "noEmitOnError": true,
-        "lib": [ "es2016", "dom" ]
+        "lib": [ "es5", "es2015.promise", "es2015.iterable", "dom" ]
     }
 }
--- a/client-ts/signalr-protocol-msgpack/package.json
+++ b/client-ts/signalr-protocol-msgpack/package.json
@@ -4,7 +4,7 @@
   "description": "MsgPack Protocol support for ASP.NET Core SignalR",
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm/index.js",
-  "typings": "./dist/esm/index.d.ts",
+  "typings": "./dist/cjs/index.d.ts",
   "umd": "./dist/browser/signalr-protocol-msgpack.js",
   "umd_name": "signalR.protocols.msgpack",
   "directories": {
--- a/client-ts/signalr-protocol-msgpack/package.json
+++ b/client-ts/signalr-protocol-msgpack/package.json
@@ -4,7 +4,7 @@
   "description": "MsgPack Protocol support for ASP.NET Core SignalR",
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm/index.js",
-  "typings": "./dist/cjs/index.d.ts",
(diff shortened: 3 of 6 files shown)


client-ts/tsconfig-base.json
```diff
@@ -13,6 +13,6 @@
         "noImplicitAny": true,
         "suppressImplicitAnyIndexErrors": true,
         "noEmitOnError": true,
-        "lib": [ "es2016", "dom" ]
+        "lib": [ "es5", "es2015.promise", "es2015.iterable", "dom" ]
```
anurse: We want to use the `es5` lib, but:
* We **do** Polyfill `Promise`, so we want to keep that
* `es2015.iterable` enables `for (const item of list)` syntax, and TypeScript **will** transpile this down to a standard `for(let i = 0; i < list.length; i+=1)` loop, so we still want to use it.

davidfowl: What's the side effect of this?

client-ts/signalr/src/HttpConnection.ts
```diff
@@ -83,8 +83,9 @@ export class HttpConnection implements IConnection {
                 let headers;
                 const token = this.options.accessTokenFactory();
                 if (token) {
-                    headers = new Map<string, string>();
-                    headers.set("Authorization", `Bearer ${token}`);
+                    headers = {
```
davidfowl: What is this crazy syntax? inline dictionary literals? Looks like C#

anurse: There is no such thing as a "Dictionary" in JavaScript. This syntax is just a way of setting a property with a name that isn't a valid identifier. This code is functionally equivalent to the following:

```javascript
headers = {
    Authorization: `Bearer ${token}`,
}
```

However, since we're treating this object like a "dictionary" I wanted to use the `[]` syntax to make it clear. The `[]` syntax would only really be needed if we wanted to set a key with a name like `Foo Bar` (with spaces) or something.

anurse: See [New notations in ECMAScript 2015](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer)

client-ts/tsconfig-base.json
anurse: 1. We can only use types available on ES5 browsers.
2. Exception to the above: We can use `Promise` (and we include a polyfill for that already)
3. Exception to the above: The ECMAScript 2015 [Iteration protocols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) are available, meaning we can use `for (const item of array)` syntax (`foreach`-like) on the built-in iterable types (`Array`, `[Type]Array`, and also `Set` and `Map` if we could access those types)

anurse: The `lib` property basically sets standard TypeScript declarations we want auto-imported into our TypeScript files.

client-ts/signalr/src/HttpConnection.ts
davidfowl: object literal.

client-ts/tsconfig-base.json
davidfowl: Does this make our js library bigger than it was before?

anurse:
> LGTM just want to know if we have more bloat because of the polyfill

We have no additional polyfills in this PR. We already had the `Promise` polyfill. This PR is to avoid having a `Map` polyfill too.

PR #1674, opened 2018-03-21
Don't expose HubConnectionContext on the Hub

### LINKED ISSUE #1369: Reduce the surface area on HubConnectionContext
Today we expose the HubConnectionContext on the Hub itself via Context.Connection. This is also the same connection used by HubLifetimeManagers to communicate directly with the socket via the Output channel. We should decouple these 2 things so that we expose less to the Hub itself (for example, it makes little sense for the hub to be writing to the output directly). 

We'll likely need a smaller object that wraps the connection and exposes:

- Items
- User
- ConnectionId
- UserIdentifier
- Abort/AbortToken
- Maybe IP Address

davidfowl:
- Made HubCallerContext an abstract class
- Made DefaultHubCallerContext that gets data from the HubConnectionContext.
- Removed IP address
- Removed Connection property

Fixes #1369

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubCallerContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubCallerContext.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AspNetCore.SignalR
 
         public override IFeatureCollection Features => _connection.Features;
 
-        public override CancellationToken ConnectionAbortedToken => _connection.ConnectionAbortedToken;
+        public override CancellationToken ConnectionAborted => _connection.ConnectionAborted;
 
         public override void Abort() => _connection.Abort();
     }
--- a/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubCallerContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/DefaultHubCallerContext.cs
@@ -1,0 +1,36 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Threading;
+using Microsoft.AspNetCore.Http.Features;
+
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNetCore.SignalR.Core/DefaultHubCallerContext.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public override string UserIdentifier => _connection.UserIdentifier;
+
+        public override ClaimsPrincipal User => _connection.User;
+
+        public override IDictionary<object, object> Items => _connection.Items;
+
+        public override IFeatureCollection Features => _connection.Features;
+
+        public override CancellationToken ConnectionAbortedToken => _connection.ConnectionAbortedToken;
```
anurse: It's `RequestAborted` not `RequestAbortedToken`, so this should be `ConnectionAborted`

https://github.com/aspnet/HttpAbstractions/blob/b059bcc426ea7593ceeed2df1b6ce14fd3a626ba/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs#L158-L162

anurse: I realize you didn't add this property, but the comment is still relevant and a simple fix ;)

davidfowl: Will do

davidfowl:
```
[xUnit.net 00:00:14.7830100]     Microsoft.AspNetCore.SignalR.Tests.HubEndpointTests.HubMethodDoesNotSendResultWhenInvocationIsNonBlocking [FAIL]
Failed   Microsoft.AspNetCore.SignalR.Tests.HubEndpointTests.HubMethodDoesNotSendResultWhenInvocationIsNonBlocking
Error Message:
 Assert.Null() Failure
Expected: (null)
Actual:   CloseMessage { Error = null }
Stack Trace:
   at Microsoft.AspNetCore.SignalR.Tests.HubEndpointTests.HubMethodDoesNotSendResultWhenInvocationIsNonBlocking() in /_/test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs:line 545
```

/cc @JamesNK

PR #1681, opened 2018-03-21
Check for actual start in SSE

mikaelm12:
Closed #1566  in favor of this.
We're just gonna leave LP alone since it's the last transport in the fallback loop and don't actually need to detect successful start up.
That means a lot less churn on the client and sever and in the tests.
You'll see that this change is much smaller.

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
@@ -54,8 +54,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
             Log.StartTransport(_logger, transferFormat);
 
+            var startTcs = new TaskCompletionSource<object>(TaskContinuationOptions.RunContinuationsAsynchronously);
             var sendTask = SendUtils.SendMessages(url, _application, _httpClient, _httpOptions, _transportCts, _logger);
-            var receiveTask = OpenConnection(_application, url, _transportCts.Token);
+            var receiveTask = OpenConnection(_application, url, startTcs, _transportCts.Token);
 
             Running = Task.WhenAll(sendTask, receiveTask).ContinueWith(t =>
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/ServerSentEventsTransport.cs
```diff
@@ -66,17 +67,30 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
...
 
             var request = new HttpRequestMessage(HttpMethod.Get, url);
             SendUtils.PrepareHttpRequest(request, _httpOptions);
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/event-stream"));
-            var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+
+            HttpResponseMessage response;
+            try
+            {
+                response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
```
anurse: We should be able to add a test for this using the TestHttpMessageHandler. Can you put a test for both the success and the failure case? Just making sure that `StartAsync` doesn't complete until the TestHttpMessageHandler returns the response.

mikaelm12: Yeah, was thinking about that. Gonna add a test after lunch!

mikaelm12:
Added some tests. Ready for review

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -370,6 +370,57 @@ public Task ClosedEventNotRaisedWhenTheClientIsStoppedButWasNeverStarted()
...
+                        return Task.FromResult(ResponseUtils.CreateResponse(HttpStatusCode.Accepted));
+                    });
+
+                    var sse = new ServerSentEventsTransport(new HttpClient(httpHandler));
+
+                    await WithConnectionAsync(
+                        CreateConnection(httpHandler, loggerFactory: loggerFactory, url: null, transport: sse),
+                        async (connection, closed) =>
+                        {
+                            Assert.False(sseStartTcs.Task.IsCompleted);
```
anurse: Actually, we want this the other way around. Have the `OnGet` handler above _wait_ on a TCS, then call `var startTask = connection.StartAsync(TransferFormat.Text).OrTimeout();`, assert that `startTask.IsCompleted` is `false`, then release the TCS and `await startTask`

We want to test that the `StartAsync` task is blocked waiting for the HTTP request to complete.

Issue #1317, opened 2018-01-18
Detect availability of transport dependencies on client and server and disable them automatically


anurse:
We should consider how to detect WebSocket availability on the server and automatically disable the transport when the server can't support WebSockets.

This would depend on aspnet/Hosting#1241 since we can't tell if the server will support it from the application without detecting the specific name of the server (and OS version, etc.)

We should also do similar detection on the TS client, detecting `WebSocket` and `EventSource` types. In the .NET client we should detect Windows 7 (non-`netcoreapp2.1`) where `ClientWebSocket` will fail.

muratg:
We need to check IHttpWebSocketFeature when generating negotiate response.

anurse:
I'm expanding this issue to cover the client too. Editing the description and renaming. The necessary changes for the server are starting to land

anurse:
* On JS Client this is just about checking for `typeof WebSocket !== "undefined"` (and the same for `EventSource`) to determine if the transports are available and skip them if they aren't
* On the .NET Server side we just need to look for `IHttpWebSocketFeature` when generating the `/negotiate` response, if it's not there, just don't include WebSockets.

PR #1682, opened 2018-03-21
Detect availability of web sockets on client and server

JamesNK:
https://github.com/aspnet/SignalR/issues/1317

### DIFF
--- a/client-ts/signalr/spec/HttpConnection.spec.ts
+++ b/client-ts/signalr/spec/HttpConnection.spec.ts
@@ -269,7 +269,30 @@ describe("HttpConnection", () => {
         }
     });
 
-    it("does not select web socket transport when not available in environment", async (done) => {
+    it("does not select ServerSentEvents transport when not available in environment", async (done) => {
+        const serverSentEventsTransport = { transport: "ServerSentEvents", transferFormats: [ "Text" ] };
+
+        const options: IHttpConnectionOptions = {
+            ...commonOptions,
+            httpClient: new TestHttpClient()
+                .on("POST", (r) => ({ connectionId: "42", availableTransports: [serverSentEventsTransport] })),
+        } as IHttpConnectionOptions;
+
+        const connection = new HttpConnection("http://tempuri.org", options);
+
+        try {
+            await connection.start(TransferFormat.Text);
+            fail();
+            done();
+        } catch (e) {
+            // ServerSentEvents is only transport returned from server but is not selected
+            // because there is no support in the environment, leading to the following error
+            expect(e.message).toBe("Unable to initialize any of the available transports.");
+            done();
+        }
+    });
+
+    it("does not select WebSockets transport when not available in environment", async (done) => {
         const webSocketsTransport = { transport: "WebSockets", transferFormats: [ "Text" ] };
 
         const options: IHttpConnectionOptions = {
@@ -285,7 +308,7 @@ describe("HttpConnection", () => {
             fail();
             done();
         } catch (e) {
-            // WebSocket is only transport returned from server but is not selected
+            // WebSockets is only transport returned from server but is not selected
             // because there is no support in the environment, leading to the following error
             expect(e.message).toBe("Unable to initialize any of the available transports.");
             done();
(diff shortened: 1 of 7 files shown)


client-ts/signalr/src/HttpConnection.ts
```diff
@@ -160,8 +160,12 @@ export class HttpConnection implements IConnection {
             const transferFormats = endpoint.transferFormats.map((s) => TransferFormat[s]);
             if (!requestedTransport || transport === requestedTransport) {
                 if (transferFormats.indexOf(requestedTransferFormat) >= 0) {
-                    this.logger.log(LogLevel.Trace, `Selecting transport '${TransportType[transport]}'`);
-                    return transport;
+                    if (transport === TransportType.WebSockets && typeof WebSocket === "undefined") {
```
anurse: Need the same check for `transport === TransportType.ServerSentEvents && typeof EventSource === "undefined"`, but only in TypeScript (because we have to use the `EventSource` type to implement it, we can't use `XMLHTTPRequest` to implement SSE)

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.Log.cs
```diff
@@ -98,6 +98,9 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _transportFailed =
                 LoggerMessage.Define<string>(LogLevel.Debug, new EventId(29, "TransportFailed"), "Skipping transport {TransportName} because it failed to initialize.");
 
+            private static readonly Action<ILogger, Exception> _webSocketsNotSupportedByOS =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(30, "WebSocketsNotSupportedByOS"), "Skipping web sockets because they are not supported by the OS.");
```
anurse: Expand OS to "Operating System" in the message

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -711,6 +720,31 @@ private ConnectionState ChangeState(ConnectionState to)
...
+                }
+                else
+                {
+                    // Window 8 and above has web sockets
+                    _hasWebSocketsSupport = Environment.OSVersion.Version > new Version(6, 1);
+                }
+#endif
+            }
+
+            return _hasWebSocketsSupport.Value;
```
anurse: I think we can skip the static. In the first place, `Nullable<bool>` is a two-field struct so it could shear, plus the code required to calculate it is basically just checking some already-cached values ([`Environment.OSVersion`](https://source.dot.net/#System.Runtime.Extensions/System/Environment.cs,137), [`RuntimeInformation.IsOSPlatform`](https://source.dot.net/#System.Runtime.InteropServices.RuntimeInformation/System/Runtime/InteropServices/RuntimeInformation/RuntimeInformation.Windows.cs,17))

src/Microsoft.AspNetCore.Sockets.Http/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -47,7 +47,7 @@ public WebSocketsTransport(WebSocketOptions options, IDuplexPipe application, De
 
         public async Task ProcessRequestAsync(HttpContext context, CancellationToken token)
         {
-            Debug.Assert(context.WebSockets.IsWebSocketRequest, "Not a websocket request");
+            Debug.Assert(context.WebSockets.IsWebSocketRequest, "Not a web socket request");
```
anurse: Leave this as WebSocket

client-ts/signalr/spec/HttpConnection.spec.ts
```diff
@@ -269,6 +269,29 @@ describe("HttpConnection", () => {
         }
     });
 
+    it("does not select web socket transport when not available in environment", async (done) => {
```
anurse: I believe "WebSocket" (one word) is the term in general use, and is used in the WebSocket Protocol spec.

https://tools.ietf.org/html/rfc6455

PR #1684, opened 2018-03-22
Copy HttpContext properties for long polling transport 

### LINKED ISSUE #1644: Null HttpContext with long polling
A hub method that accesses the `HttpContext` can get null instead if the method was invoked while the poll request ended.

https://github.com/aspnet/SignalR/blob/0e38ee3e63f4fb7ed08b3626fa85ec5f33952e09/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs#L279

This has shown up with `HubConnectionTests.ClientCanSendHeaders(transportType: LongPolling)` on OSX build.

https://travis-ci.org/aspnet/SignalR/jobs/355210173#L1309

```
[xUnit.net 00:00:06.1150930]     Microsoft.AspNetCore.SignalR.Client.FunctionalTests.HubConnectionTests.ClientCanSendHeaders(transportType: LongPolling) [FAIL]
Failed   Microsoft.AspNetCore.SignalR.Client.FunctionalTests.HubConnectionTests.ClientCanSendHeaders(transportType: LongPolling)
Error Message:
 Microsoft.AspNetCore.SignalR.Client.HubException : An unexpected error occurred invoking 'GetHeaderValues' on the server. InvalidOperationException: Unable to get HttpContext from request.
Stack Trace:
```

davidfowl:
- The long polling transport simulates a persistent connection
over multiple http requests. In order to expose common http request
properties, we need to copy them to a fake http context on the first poll
and set that as the HttpContext exposed via the IHttpContextFeature.

Fixes #1644

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -2,14 +2,17 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Pipelines;
+using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Http.Internal;
 using Microsoft.AspNetCore.Protocols;
 using Microsoft.AspNetCore.Protocols.Features;
 using Microsoft.AspNetCore.Sockets.Internal;
@@ -276,8 +279,6 @@ namespace Microsoft.AspNetCore.Sockets
 
                             connection.Status = DefaultConnectionContext.ConnectionStatus.Inactive;
 
-                            connection.SetHttpContext(null);
-
                             // Dispose the cancellation token
                             connection.Cancellation.Dispose();
 
@@ -488,21 +489,88 @@ namespace Microsoft.AspNetCore.Sockets
                 return false;
             }
 
+            // Setup the connection state from the http context
+            connection.User = context.User;
+
             // Configure transport-specific features.
             if (transportType == TransportType.LongPolling)
             {
                 connection.Features.Set<IConnectionInherentKeepAliveFeature>(new ConnectionInherentKeepAliveFeature(options.LongPolling.PollTimeout));
-            }
 
-            // Setup the connection state from the http context
-            connection.User = context.User;
-            connection.SetHttpContext(context);
+                // For long polling, the requests come and go but the connection is still alive.
+                // to make the IHttpContextFeature work well, we make a copy of the relevant properties
+                // to a new HttpContext. This means that it's impossible to affect the context
+                // with subsequent requests.
+                if (connection.GetHttpContext() == null)
+                {
+                    var httpContext = CloneHttpContext(context);
+                    connection.SetHttpContext(httpContext);
+                }
+            }
+            else
+            {
+                connection.SetHttpContext(context);
+            }
 
             // Set the Connection ID on the logging scope so that logs from now on will have the
             // Connection ID metadata set.
             logScope.ConnectionId = connection.ConnectionId;
 
             return true;
+        }
+
+        private static HttpContext CloneHttpContext(HttpContext context)
+        {
+            // The reason we're copying the base features instead of the HttpContext properties is
+            // so that we can get all of the logic built into DefaultHttpContext to extract higher level
+            // structure from the low level properties
+            var existingRequestFeature = context.Features.Get<IHttpRequestFeature>();
+
+            var requestFeature = new HttpRequestFeature();
+            requestFeature.Protocol = existingRequestFeature.Protocol;
+            requestFeature.Method = existingRequestFeature.Method;
+            requestFeature.Scheme = existingRequestFeature.Scheme;
+            requestFeature.Path = existingRequestFeature.Path;
+            requestFeature.PathBase = existingRequestFeature.PathBase;
+            requestFeature.QueryString = existingRequestFeature.QueryString;
+            requestFeature.RawTarget = existingRequestFeature.RawTarget;
+            var requestHeaders = new Dictionary<string, StringValues>(existingRequestFeature.Headers.Count);
+            foreach (var header in existingRequestFeature.Headers)
+            {
+                requestHeaders[header.Key] = header.Value;
+            }
+            requestFeature.Headers = new HeaderDictionary(requestHeaders);
+
+            var existingConnectionFeature = context.Features.Get<IHttpConnectionFeature>();
+            var connectionFeature = new HttpConnectionFeature();
+            connectionFeature.ConnectionId = existingConnectionFeature.ConnectionId;
+            connectionFeature.LocalIpAddress = existingConnectionFeature.LocalIpAddress;
+            connectionFeature.LocalPort = existingConnectionFeature.LocalPort;
+            connectionFeature.RemoteIpAddress = existingConnectionFeature.RemoteIpAddress;
+            connectionFeature.RemotePort = existingConnectionFeature.RemotePort;
+
+            // The response is a dud, you can't do anything with it anyways
+            var responseFeature = new HttpResponseFeature();
+
+            var features = new FeatureCollection();
+            features.Set<IHttpRequestFeature>(requestFeature);
+            features.Set<IHttpResponseFeature>(responseFeature);
+            features.Set<IHttpConnectionFeature>(connectionFeature);
+
+            // REVIEW: We could strategically look at adding other features but it might be better
+            // if the expose a callback that would allow the user to preserve HttpContext properties.
+
+            var newHttpContext = new DefaultHttpContext(features);
+            newHttpContext.TraceIdentifier = context.TraceIdentifier;
+            newHttpContext.User = context.User;
+
+            // Making request services function property could be tricky and expensive as it would require
+            // DI scope per connection.
+            newHttpContext.RequestServices = EmptyServiceProvider.Instance;
+
+            // REVIEW: This extends the lifetime of anything that got put into HttpContext.Items
+            newHttpContext.Items = context.Items.ToDictionary(p => p.Key, p => p.Value);
+            return newHttpContext;
         }
 
         private async Task<DefaultConnectionContext> GetConnectionAsync(HttpContext context, HttpSocketOptions options)
@@ -567,6 +635,12 @@ namespace Microsoft.AspNetCore.Sockets
             EnsureConnectionStateInternal(connection, options);
 
             return connection;
+        }
+
+        private class EmptyServiceProvider : IServiceProvider
+        {
+            public static EmptyServiceProvider Instance { get; } = new EmptyServiceProvider();
+            public object GetService(Type serviceType) => null;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
@@ -568,7 +568,7 @@ namespace Microsoft.AspNetCore.Sockets
             features.Set<IHttpConnectionFeature>(connectionFeature);
 
             // REVIEW: We could strategically look at adding other features but it might be better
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -505,6 +525,64 @@ private static void WriteTransport(JsonWriter writer, string transportName, Tran
...
+            // The response is a dud, you can't do anything with it anyways
+            var responseFeature = new HttpResponseFeature();
+
+            var features = new FeatureCollection();
+            features.Set<IHttpRequestFeature>(requestFeature);
+            features.Set<IHttpResponseFeature>(responseFeature);
+            features.Set<IHttpConnectionFeature>(connectionFeature);
+
+            // REVIEW: We could strategically look at adding other features but it might be better
+            // if the expose a callback that would allow the user to preserve HttpContext properties.
```
mikaelm12: if *we* expose a callback...

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -505,6 +525,64 @@ private static void WriteTransport(JsonWriter writer, string transportName, Tran
...
+            features.Set<IHttpConnectionFeature>(connectionFeature);
+
+            // REVIEW: We could strategically look at adding other features but it might be better
+            // if we expose a callback that would allow the user to preserve HttpContext properties.
+
+            var newHttpContext = new DefaultHttpContext(features);
+            newHttpContext.TraceIdentifier = context.TraceIdentifier;
+            newHttpContext.User = context.User;
+
+            // Making request services function property could be tricky and expensive as it would require
```
mikaelm12: Hmm. Not sure I understand.
Is this something that should be investigated further later on?

davidfowl: Yah, I can remove the comment but I wanted to make it visible in the review.

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -540,7 +540,17 @@ public class HubEndpointTests
                 client.Dispose();
 
                 // Ensure the client channel is empty
-                Assert.Null(client.TryRead());
+                var message = client.TryRead();
```
davidfowl: I had to "fix" this test since it kept failing with the close message thing.

/cc @JamesNK

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
mikaelm12: The comment is fine. Maybe log an issue and link to it

JamesNK: How do DI scopes work with signalr? Why would it be tricky/expensive?

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -505,6 +525,64 @@ private static void WriteTransport(JsonWriter writer, string transportName, Tran
...
+            // DI scope per connection.
+            newHttpContext.RequestServices = EmptyServiceProvider.Instance;
+
+            // REVIEW: This extends the lifetime of anything that got put into HttpContext.Items
+            newHttpContext.Items = context.Items.ToDictionary(p => p.Key, p => p.Value);
```
JamesNK: Hmmm. To be honest I'm not 100% sure how `HttpContext` works in websockets land. Is there middleware that would place values in `Items`?

JamesNK: Also, would this work to copy the dictionary:

`newHttpContext.Items = new Dictionary<object, object>(context.Items);`

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -336,6 +338,90 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
...
+                    Assert.Equal("h2", connectionHttpContext.Request.Headers["header2"]);
+                    Assert.Equal("h3", connectionHttpContext.Request.Headers["header3"]);
+                    Assert.Equal("requestid", connectionHttpContext.TraceIdentifier);
+                    Assert.Equal("claimValue", connectionHttpContext.User.Claims.FirstOrDefault().Value);
+                    Assert.Equal("connectionid", connectionHttpContext.Connection.Id);
+                    Assert.Equal(IPAddress.Loopback, connectionHttpContext.Connection.LocalIpAddress);
+                    Assert.Equal(4563, connectionHttpContext.Connection.LocalPort);
+                    Assert.Equal(IPAddress.IPv6Any, connectionHttpContext.Connection.RemoteIpAddress);
+                    Assert.Equal(43456, connectionHttpContext.Connection.RemotePort);
+                    Assert.NotNull(connectionHttpContext.RequestServices);
```
JamesNK: Also check that doing stuff with the response has no effect?

davidfowl: I can assert the properties are null.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
davidfowl: > Hmmm. To be honest I'm not 100% sure how HttpContext works in websockets land. Is there middleware that would place values in Items?

The HttpContext is fine in server sent events and websockets, it's in tact because the request doesn't complete until the connection is dead.

> Also, would this work to copy the dictionary:

I'll do that not sure why I LINQified this thing 😄

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
```diff
@@ -540,7 +540,17 @@ public class HubEndpointTests
...
+                switch (message)
+                {
+                    case CloseMessage close:
+                        break;
+                    case null:
```
mikaelm12: Can you just remove this case since it would be handled by the default case

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
davidfowl: SignalR makes a scope per hub invocation which works great. The RequestServices on the other hand would exist for the lifetime of the connection. We'd then need to dispose the scope when the connection is disposed, that in itself isn't too bad. Where it gets wonky is when somebody tries to get the same instance of a service in middleware and then in a Hub. It'll be different (which is by design and most likely fine). The strange thing about it is that RequestServices would have been replaced.

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -336,6 +338,90 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
             }
         }
 
+        [Fact]
+        public async Task HttpContextFeatureForLongpollingWorksBetweenPolls()
```
mikaelm12: I'm sure @moozzyk would love how long this test is 😄

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTests.cs
davidfowl: True. Will do.

davidfowl:
🆙 📅

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -568,5 +646,11 @@ private void EnsureConnectionStateInternal(DefaultConnectionContext connection,
 
             return connection;
         }
+
+        private class EmptyServiceProvider : IServiceProvider
+        {
+            public static EmptyServiceProvider Instance { get; } = new EmptyServiceProvider();
+            public object GetService(Type serviceType) => null;
```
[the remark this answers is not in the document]
davidfowl: This will at least give a proper exception message when somebody is asking for a required service. That's why I didn't leave it as null. Push come to shove, we can implement something where we make a scoped service provider. I just wanted to see what people think of this first.

davidfowl:
> One more thing. I don't know all of gotchas of long polling so wait for more approves before merge

np

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -488,15 +488,34 @@ private static void WriteTransport(JsonWriter writer, string transportName, Tran
...
+                var existing = connection.GetHttpContext();
+                if (existing == null)
+                {
+                    var httpContext = CloneHttpContext(context);
+                    connection.SetHttpContext(httpContext);
+                }
+                else
+                {
+                    // Set the request trace identifier to the current http request handling the poll
+                    existing.TraceIdentifier = context.TraceIdentifier;
```
anurse: :+1:

Any other properties we want to update? Connection info (IP/Port/etc; In case a client changes networks, etc.)?

Of course, this is run in parallel with existing invocations so the data could "shear" (but as long as each set is atomic, it won't break anything).

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
```diff
@@ -505,6 +524,65 @@ private static void WriteTransport(JsonWriter writer, string transportName, Tran
...
+            newHttpContext.TraceIdentifier = context.TraceIdentifier;
+            newHttpContext.User = context.User;
+
+            // Making request services function property could be tricky and expensive as it would require
+            // DI scope per connection. It would also mean that services resolved in middleware leading up to here
+            // wouldn't be the same instance (but maybe that's fine). For now, we just return an empty service provider
+            newHttpContext.RequestServices = EmptyServiceProvider.Instance;
+
+            // REVIEW: This extends the lifetime of anything that got put into HttpContext.Items
+            newHttpContext.Items = new Dictionary<object, object>(context.Items);
```
anurse: This worries me a bit... people use this to cache per-request stuff and we don't want to make those items live longer than the actual request.

davidfowl: > This worries me a bit... people use this to cache per-request stuff and we don't want to make those items live longer than the actual request.

Yea, thats why I commented.

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
davidfowl: Extremely unlikely this would matter in real life. The reason I changed the trace identifier is because we log it to tell the user what the previous connection id is.

BrennanConroy: Should we similarly set the `existing.User` value? Although we use `connection.User` for the Hub method auth

davidfowl: Sure

src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs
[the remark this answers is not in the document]
JamesNK: What is the error message? I was thinking that calling GetService would throw an error explaining why it isn't supported rather than return null and lead to a hard to debug NRE.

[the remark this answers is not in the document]
davidfowl: The usage pattern is that the caller does GetService in order to get a null and we have a GetRequiredService that throws if GetService returns null. If anyone is calling GetService today and not expecting null then they already have an NRE

test/Microsoft.AspNetCore.Sockets.Tests/HttpConnectionDispatcherTests.cs
moozzyk: :D

PR #1694, opened 2018-03-23
Remove message content from JS client logging

JamesNK:
https://github.com/aspnet/SignalR/issues/1449

### DIFF
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -95,7 +95,7 @@ export class HubConnection {
                         this.connection.stop(message.error ? new Error("Server returned an error on close: " + message.error) : null);
                         break;
                     default:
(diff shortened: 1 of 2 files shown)


davidfowl:
> Looks good to me. @davidfowl if you want to introduce a flag to re-enable this, file an issue for that and we can track it later. This change is required to ship though, so I'd rather get it in ASAP.

Ummm, naw. See the problem is when you use binary you can't really use any other tools to look at the data, so I think the flag *has* to be added along with the feature.

JamesNK:
Where does the flag exist? How does a dev set it?

davidfowl:
> Where should the flag exist? How does a dev set it?

All of our ts flags are currrently on an options object passed into the hubConnection constructor. I imagine it would go there.

anurse:
> See the problem is when you use binary you can't really use any other tools

What do you expect our logging to output? Base64? Hex?

PR #1695, opened 2018-03-23
Fallback for TS client

mikaelm12:
Still needs tests.

### DIFF
--- a/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -567,7 +567,6 @@ describe("hubConnection", () => {
         } catch (e) {
             fail(e);
         }
-
     });
 
     function getJwtToken(url): Promise<string> {
--- a/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -560,8 +560,9 @@ describe("hubConnection", () => {
         try {
             await hubConnection.start();
 
-            // Make sure that we connect with SSE.
-            expect(await hubConnection.invoke("GetActiveTransportName")).toEqual("ServerSentEvents" || "LongPolling");
+            // Make sure that we connect with SSE or LongPolling after Websockets fail
+            const transportName = await hubConnection.invoke("GetActiveTransportName");
+            expect(transportName === "ServerSentEvents" || transportName === "LongPolling").toBe(true);
             (window as any).WebSocket = oldWebSocket;
             done();
         } catch (e) {
--- a/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -563,10 +563,11 @@ describe("hubConnection", () => {
             // Make sure that we connect with SSE or LongPolling after Websockets fail
             const transportName = await hubConnection.invoke("GetActiveTransportName");
             expect(transportName === "ServerSentEvents" || transportName === "LongPolling").toBe(true);
-            (window as any).WebSocket = oldWebSocket;
-            done();
         } catch (e) {
             fail(e);
+        } finally {
+            (window as any).WebSocket = oldWebSocket;
+            done();
         }
     });
 
--- a/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -544,7 +544,7 @@ describe("hubConnection", () => {
         }
     });
 
-    it("transport falls back from WebSockets to SSE", async (done) => {
+    it("transport falls back from WebSockets to SSE or LongPolling", async (done) => {
         // Replace Websockets with a function that just
         // throws to force fallback.
         const oldWebSocket = (window as any).WebSocket;
@@ -561,7 +561,7 @@ describe("hubConnection", () => {
             await hubConnection.start();
 
             // Make sure that we connect with SSE.
-            expect(await hubConnection.invoke("GetActiveTransportName")).toEqual("ServerSentEvents");
+            expect(await hubConnection.invoke("GetActiveTransportName")).toEqual("ServerSentEvents" || "LongPolling");
             (window as any).WebSocket = oldWebSocket;
             done();
         } catch (e) {
--- a/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -544,6 +544,32 @@ describe("hubConnection", () => {
         }
     });
 
+    it("transport falls back from WebSockets to SSE", async (done) => {
+        // Replace Websockets with a function that just
+        // throws to force fallback.
+        const oldWebSocket = (window as any).WebSocket;
+        (window as any).WebSocket = () => {
+            throw new Error("Kick rocks");
+        };
+
+        const hubConnection = new HubConnection(TESTHUBENDPOINT_URL, {
+            logger: LogLevel.Trace,
+            protocol: new JsonHubProtocol(),
+        });
+
+        try {
+            await hubConnection.start();
+
+            // Make sure that we connect with SSE.
+            expect(await hubConnection.invoke("GetActiveTransportName")).toEqual("ServerSentEvents");
+            (window as any).WebSocket = oldWebSocket;
+            done();
+        } catch (e) {
+            fail(e);
+        }
+
+    });
+
     function getJwtToken(url): Promise<string> {
         return new Promise((resolve, reject) => {
             const xhr = new XMLHttpRequest();
--- a/clients/ts/signalr/src/HttpConnection.ts
+++ b/clients/ts/signalr/src/HttpConnection.ts
@@ -91,7 +91,7 @@ export class HttpConnection implements IConnection {
                     };
                 }
 
-                const negotiateResponse = await this.setNegotiationResponse(headers);
+                const negotiateResponse = await this.getNegotiationResponse(headers);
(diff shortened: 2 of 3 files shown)


clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -79,39 +81,35 @@ export class HttpConnection implements IConnection {
...
                 this.url = this.baseUrl;
                 this.transport = this.constructTransport(TransportType.WebSockets);
+                // We should just call connect directly in this case.
+                // No fallback or negotiate in this case.
+                await this.transport.connect(this.url, transferFormat, this);
             } else {
                 let headers;
                 const token = this.options.accessTokenFactory();
                 if (token) {
                     headers = {
```
JamesNK: You can get rid of this variable and just assign directly to `this.headers` and use that.

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -79,39 +81,35 @@ export class HttpConnection implements IConnection {
...
                         ["Authorization"]: `Bearer ${token}`,
                     };
+                    this.headers = headers;
                 }
 
                 const negotiatePayload = await this.httpClient.post(this.resolveNegotiateUrl(this.baseUrl), {
```
JamesNK: What is this for compared to the same call in `setNegotiationResponse`?

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -123,16 +121,48 @@ export class HttpConnection implements IConnection {
...
         }
 
-        for (const endpoint of availableTransports) {
+        const transports = this.negotiationResponse.availableTransports;
+        for (const endpoint of transports) {
             const transport = this.resolveTransport(endpoint, requestedTransport, requestedTransferFormat);
             if (typeof transport === "number") {
-                return this.constructTransport(transport);
+                this.transport = this.constructTransport(transport);
+                if (this.negotiationResponse.connectionId == null) {
```
JamesNK: `===` or face the wrath of the linter gods.

clients/ts/signalr/src/HttpConnection.ts
mikaelm12: Missed this

clients/ts/signalr/src/HttpConnection.ts
mikaelm12: Yeah. Was hitting scoping issues before. I'll clean this up

clients/ts/signalr/src/HttpConnection.ts
anurse: `===` FOR THE EQUALITY GODS

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -79,39 +81,28 @@ export class HttpConnection implements IConnection {
...
                 // the user tries to stop the the connection when it is being started
                 if (this.connectionState === ConnectionState.Disconnected) {
                     return;
                 }
-
-                if (this.connectionId) {
-                    this.url = this.baseUrl + (this.baseUrl.indexOf("?") === -1 ? "?" : "&") + `id=${this.connectionId}`;
-                    this.transport = this.createTransport(this.options.transport, negotiateResponse.availableTransports, transferFormat);
-                }
+                return this.createTransport(this.options.transport, transferFormat);
```
anurse: You need to `await this.createTransport(...` instead of `return`ing it. When you have an `async` function in TypeScript, you don't return a `Promise<T>`, you return a `T` and the compiler wraps everything up into a `Promise` for you. Just like in C# (partly because C# was the inspiration for async/await in TypeScript, and later ECMAScript)

anurse: If you do this, what you're actually going to end up returning is `Promise<Promise<void>>`. The problem is that JavaScript allows any object to be returned by any function, even if it's labelled `void`, so it's not an error to do this, it's just weird behavior.

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -123,16 +113,50 @@ export class HttpConnection implements IConnection {
         }
     }
 
-    private createTransport(requestedTransport: TransportType | ITransport, availableTransports: IAvailableTransport[], requestedTransferFormat: TransferFormat): ITransport {
+    private async setNegotiationResponse(headers: any): Promise<INegotiateResponse> {
```
anurse: nit: this is now `getNegotiationResponse` 🙃

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -123,16 +113,50 @@ export class HttpConnection implements IConnection {
...
         if (this.isITransport(requestedTransport)) {
             this.logger.log(LogLevel.Trace, "Connection was provided an instance of ITransport, using that directly.");
-            return requestedTransport;
+            this.transport = requestedTransport;
+            await this.transport.connect(this.url, requestedTransferFormat, this);
+
+            // only change the state if we were connecting to not overwrite
+            // the state if the connection is already marked as Disconnected
+            this.changeState(ConnectionState.Connecting, ConnectionState.Connected);
+            return Promise.resolve();
```
anurse: `async` function means just use `return` here. TypeScript manages the outer `Promise`

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -123,16 +113,50 @@ export class HttpConnection implements IConnection {
...
-                return this.constructTransport(transport);
+                this.transport = this.constructTransport(transport);
+                if (negotiateResponse.connectionId === null) {
+                    negotiateResponse = await this.setNegotiationResponse(headers);
+                    this.updateConnectionId(negotiateResponse);
+                }
+                try {
+                    await this.transport.connect(this.url, requestedTransferFormat, this);
+                    this.changeState(ConnectionState.Connecting, ConnectionState.Connected);
+                    return Promise.resolve();
```
anurse: Just use `return`, no `Promise.resolve()` needed

mikaelm12:
Added tests with @anurse's "Crazy pants" idea 😄

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -544,6 +544,31 @@ describe("hubConnection", () => {
...
+
+        const hubConnection = new HubConnection(TESTHUBENDPOINT_URL, {
+            logger: LogLevel.Trace,
+            protocol: new JsonHubProtocol(),
+        });
+
+        try {
+            await hubConnection.start();
+
+            // Make sure that we connect with SSE.
```
anurse: This won't be true on IE when we test there. Let's check that it's either ServerSentEvents OR LongPolling.

mikaelm12: Good catch 👍

clients/ts/signalr/src/HttpConnection.ts
moozzyk: @mikaelm12 _All animals are equal, but some animals are more equal than others_

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -544,6 +544,31 @@ describe("hubConnection", () => {
...
+            expect(await hubConnection.invoke("GetActiveTransportName")).toEqual("ServerSentEvents" || "LongPolling");
```
anurse: I don't think this works, I think you'd have to do something like

```typescript
let transportName = await hubConnection.invoke("GetActiveTransportName");
expect(transportName === "ServerSentEvents" || transportName === "LongPolling").toBe(true)
```

Truthy and falsey (the weirdest concepts in JavaScript) strike again, because `x || y` means "if `x` is truthy, it is the value of the expression, otherwise `y` is". Any non-empty string is truthy, so `"ServerSentEvents" || "LongPolling" === "ServerSentEvents"`.

JavaScript is super weird and confusing... this stuff gets everyone confused.

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -544,6 +544,31 @@ describe("hubConnection", () => {
         }
     });
 
+    it("transport falls back from WebSockets to SSE or LongPolling", async (done) => {
+        // Replace Websockets with a function that just
+        // throws to force fallback.
+        const oldWebSocket = (window as any).WebSocket;
+        (window as any).WebSocket = () => {
+            throw new Error("Kick rocks");
```
davidfowl: 🤣

sergeiTrice:
I am having difficulty actually understanding any of this. Could someone please clarify what's going on?

mikaelm12:
> I am having difficulty actually understanding any of this. Could someone please clarify what's going on?

@sergeiTrice 

SignalR can communicate using one of 3 transports. Websockets, Server Sents Events or Long Polling. You can chose a specific transport to use or default to using any in the order that I listed them. If a connection fails to start with a transport it will attempt to connect again with the next transport.

mikaelm12:
The functional test I added now passes in Edge and IE

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -544,6 +544,32 @@ describe("hubConnection", () => {
...
+            protocol: new JsonHubProtocol(),
+        });
+
+        try {
+            await hubConnection.start();
+
+            // Make sure that we connect with SSE or LongPolling after Websockets fail
+            const transportName = await hubConnection.invoke("GetActiveTransportName");
+            expect(transportName === "ServerSentEvents" || transportName === "LongPolling").toBe(true);
+            (window as any).WebSocket = oldWebSocket;
```
anurse: Put this in a `finally` block. If the test throws, we don't want to leave `window.WebSocket` in a broken state.

Issue #1690, opened 2018-03-22
Change to MessagePack-CSharp


anurse:
https://github.com/neuecc/MessagePack-CSharp/

* Remove the structural MessagePack tests, they're nice but if they don't work easily let's just get this in
* Verify that case-insensitive binding works and if not, file an issue and try to get that moving.

moozzyk:
Make sure this kinds of scenarios work https://github.com/neuecc/MessagePack-CSharp/issues/61. When I originally looked at MessagePack-CSharp it was not [mature enough](https://github.com/aspnet/SignalR/issues/543#issuecomment-313100131).

BrennanConroy:
neuecc/MessagePack-CSharp#223 to track ContractlessReflectionObjectResolver

PR #1698, opened 2018-03-23
[WIP] Switch to MessagePack-CSharp

BrennanConroy:
#1690 

## Before
```
             Method |          Input | HubProtocol |        Mean |      Error |     StdDev |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------ |------------:|-----------:|-----------:|------------:|-------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |     MsgPack |  1,454.6 ns |  28.547 ns |  28.037 ns |   687,493.0 | 0.0095 |      - |     928 B |
 WriteSingleMessage |   FewArguments |     MsgPack |  1,465.8 ns |  24.796 ns |  21.981 ns |   682,224.6 | 0.0134 |      - |    1224 B |
  ReadSingleMessage | LargeArguments |     MsgPack |  9,990.2 ns | 197.219 ns | 193.695 ns |   100,098.2 | 0.5035 |      - |   41784 B |
 WriteSingleMessage | LargeArguments |     MsgPack | 19,843.1 ns | 375.902 ns | 638.311 ns |    50,395.4 | 1.2512 | 0.0305 |  113944 B |
  ReadSingleMessage |  ManyArguments |     MsgPack |  2,937.8 ns |  58.649 ns |  54.860 ns |   340,389.1 | 0.0153 |      - |    1392 B |
 WriteSingleMessage |  ManyArguments |     MsgPack |  2,806.6 ns |  67.961 ns |  93.026 ns |   356,301.9 | 0.0153 |      - |    1496 B |
  ReadSingleMessage |    NoArguments |     MsgPack |    487.7 ns |   9.532 ns |   9.788 ns | 2,050,537.8 | 0.0057 |      - |     536 B |
 WriteSingleMessage |    NoArguments |     MsgPack |    686.0 ns |  12.392 ns |  11.591 ns | 1,457,775.1 | 0.0105 |      - |     976 B |    
```

## After
```
             Method |          Input | HubProtocol |        Mean |     Error |     StdDev |        Op/s |  Gen 0 | Allocated |
------------------- |--------------- |------------ |------------:|----------:|-----------:|------------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |     MsgPack |    616.6 ns |  12.10 ns |  15.305 ns | 1,621,868.1 | 0.0048 |     456 B |
 WriteSingleMessage |   FewArguments |     MsgPack |    748.6 ns |  10.59 ns |   8.840 ns | 1,335,854.4 | 0.0086 |     752 B |
  ReadSingleMessage | LargeArguments |     MsgPack |  7,929.0 ns | 155.05 ns | 196.089 ns |   126,119.9 | 0.3815 |   31144 B |
 WriteSingleMessage | LargeArguments |     MsgPack | 13,051.2 ns | 259.98 ns | 455.326 ns |    76,621.4 | 0.7324 |   62288 B |
  ReadSingleMessage |  ManyArguments |     MsgPack |  1,022.3 ns |  19.59 ns |  24.053 ns |   978,145.4 | 0.0057 |     632 B |
 WriteSingleMessage |  ManyArguments |     MsgPack |  1,193.0 ns |  23.48 ns |  33.679 ns |   838,249.9 | 0.0095 |     776 B |
  ReadSingleMessage |    NoArguments |     MsgPack |    240.8 ns |  4.717 ns |   7.343 ns | 4,152,286.3 | 0.0041 |     352 B |
 WriteSingleMessage |    NoArguments |     MsgPack |    483.3 ns |  11.61 ns |  10.292 ns | 2,069,273.1 | 0.0076 |     744 B |
```


neuecc:
I'm messagepack-csharp author.
Especially the performance of Read is not as I expected.

This line may be `MessagePackSerializer.NonGeneric.Deserialize(type, ...)`
https://github.com/aspnet/SignalR/blob/a59d57d11697c39b257811f74acd6a31351e9e6a/src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs#L572
(this change improve performance. create dictionary vs create type, of course create type is faster).

Read twice reduce performance.
https://github.com/aspnet/SignalR/blob/a59d57d11697c39b257811f74acd6a31351e9e6a/src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs#L573
If requires readSize, use native formatter can read readSize.
`ContractlessStandardResolver.Instance.GetFormatter<T>().Deserialize()`.
But non-generic api can not get `T` formatter!
I'll add `NonGeneric.Deserialize(Type type, ArraySegment<byte> bytes, IFormatterResolver resolver, out int readSize)` overload.

string field was only used in exception but create every call.
https://github.com/aspnet/SignalR/blob/a59d57d11697c39b257811f74acd6a31351e9e6a/src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs#L207-L208

Thanks for check my lib and try to use.

anurse:
> Especially the performance of Read is not as I expected.

It's still better than what we had before! If there are more improvements, we'll probably look at making them in a separate PR.

@BrennanConroy can you check the case-sensitivity issue (can camelCase fields be deserialized to a .NET type with PascalCased properties)? That's the main functional issue with the current library. If it doesn't work, that's fine, we just need to look at contributing code to make it work.

> Thanks for check my lib and try to use.

We're glad you're excited @neuecc ! We may have a few things we need to fix or add to the library to get it to work for our scenario and we'd love to work with you to get those done.

BrennanConroy:
@neuecc is there a way of creating a custom resolver similar to 
```
MessagePack.Resolvers.CompositeResolver.RegisterAndSetAsDefault(
    // Resolve DateTime first
    MessagePack.Resolvers.NativeDateTimeResolver.Instance,
    MessagePack.Resolvers.ContractlessStandardResolver.Instance
);
```
Without it affecting the global state?

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -65,39 +66,42 @@ public bool TryParseMessages(ReadOnlyMemory<byte> input, IInvocationBinder binde
...
-                        return CreateCompletionMessage(unpacker, binder);
+                        return CreateCompletionMessage(input, ref startOffset, binder);
                     case HubProtocolConstants.CancelInvocationMessageType:
-                        return CreateCancelInvocationMessage(unpacker);
+                        return CreateCancelInvocationMessage(input, ref startOffset);
                     case HubProtocolConstants.PingMessageType:
                         return PingMessage.Instance;
                     case HubProtocolConstants.CloseMessageType:
-                        return CreateCloseMessage(unpacker);
+                        return CreateCloseMessage(input, ref startOffset);
```
JamesNK: Does `startOffset` return how far into the byte array we read with msgpack? If we know that then I think we can optimizations to `BinaryMessageParser.TryParseMessage`.  We should no longer need to search for the next VarInt because we know exactly where it will start: immediately after the current message ends.

Could leave this for a separate follow up issue.

JamesNK: @davidfowl

anurse: In theory, we don't need the length prefix at all. MsgPack is fully self-describing (like JSON). The main reason we have the length prefix and the record separator is to allow us to determine if we have enough data to process the whole message without having to dive into the parser (only to figure out we don't have enough data to complete the object).

With the length prefix/record separator, we can use simple logic (reading a var int or scanning for a single byte) to quickly determine if we have a full message.

BrennanConroy:
Still have the camelCase issue, filed https://github.com/neuecc/MessagePack-CSharp/issues/220

davidfowl:
Can you also run the broadcast benchmark?

JamesNK:
@anurse

> With the length prefix/record separator, we can use simple logic (reading a var int or scanning for a single byte) to quickly determine if we have a full message.

Why do we need to know if we have the full message? If Json.NET gets to the end of the text stream without the JSON object being completed then it will error. I assume msgpack will do the same if it runs out of bytes before it is done.

davidfowl:
We might also want to look at client compatibility, Android/iOS etc

BrennanConroy:
Before
```
       Method | Connections | Protocol |         Mean |      Error |     StdDev |      Op/s |  Gen 0 | Allocated |
------------- |------------ |--------- |-------------:|-----------:|-----------:|----------:|-------:|----------:|
 SendAsyncAll |           1 |  msgpack |     2.576 us |  0.0504 us |  0.0755 us | 388,212.4 | 0.0191 |   1.26 KB |
 SendAsyncAll |          10 |  msgpack |    17.481 us |  0.0971 us |  0.0908 us |  57,204.8 |      - |   1.83 KB |
 SendAsyncAll |        1000 |  msgpack | 1,445.012 us | 15.6230 us | 14.6138 us |     692.0 |      - |   63.7 KB |
```

After
```
       Method | Connections | Protocol |         Mean |     Error |    StdDev |      Op/s |  Gen 0 | Allocated |
------------- |------------ |--------- |-------------:|----------:|----------:|----------:|-------:|----------:|
 SendAsyncAll |           1 |  msgpack |     2.007 us | 0.0307 us | 0.0272 us | 498,238.0 | 0.0191 |   1.02 KB |
 SendAsyncAll |          10 |  msgpack |    16.262 us | 0.1100 us | 0.0859 us |  61,493.8 |      - |    1.6 KB |
 SendAsyncAll |        1000 |  msgpack | 1,360.875 us | 6.4879 us | 4.6912 us |     734.8 |      - |  63.48 KB |
```

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -24,7 +23,7 @@ public class MessagePackHubProtocol : IHubProtocol
         public static readonly string ProtocolName = "messagepack";
         public static readonly int ProtocolVersion = 1;
 
-        public SerializationContext SerializationContext { get; }
+        private static bool _setup;
```
davidfowl: Why not a static ctor?

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Microsoft.AspNetCore.SignalR.Protocols.MsgPack.csproj
```diff
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="MsgPack.Cli" Version="$(MsgPackCliPackageVersion)" />
+  <PackageReference Include="MessagePack" Version="$(MessagePackPackageVersion)" />
```
davidfowl: nit: spacing

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
BrennanConroy: Because I'm going to add options

neuecc:
＠BrennanConroy 
Yes, `CompositeResolver.RegisterAndSetAsDefault` should not use in library.
I'm recommending create library specific resolver and use it as default(and user can set custom resolver as option).

```csharp
public class SignalRDefaultResolver : IFormatterResolver
{
    public static readonly IFormatterResolver Instance = new SignalRDefaultResolver();

    static readonly IFormatterResolver[] resolvers = new[]
    {
        MessagePack.Resolvers.NativeDateTimeResolver.Instance,
        MessagePack.Resolvers.ContractlessStandardResolver.Instance
    };

    SignalRDefaultResolver()
    {

    }

    public IMessagePackFormatter<T> GetFormatter<T>()
    {
        return Cache<T>.formatter;
    }

    static class Cache<T>
    {
        public static readonly IMessagePackFormatter<T> formatter;

        static Cache()
        {
            foreach (var resolver in resolvers)
            {
                formatter = resolver.GetFormatter<T>();
                if (formatter != null) return;
            }
        }
    }
}
```

neuecc:
I saw GUID is string commit. https://github.com/aspnet/SignalR/pull/1698/commits/45c591c120f8921beac9c937f95059b49acb525c

If we can okay to use C# native Guid representation, we can add `UnsafeBinaryResolver`
that packs Guid(and decimal) as binary.
https://github.com/neuecc/MessagePack-CSharp/blob/9c7fac3be5f428b092ef08ea6b335577db2f5ef0/src/MessagePack/Formatters/UnsafeBinaryFormatters.cs
Of course, it is faster than "GUID is string".

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -269,11 +274,8 @@ public void WriteMessage(HubMessage message, Stream output)
             }
         }
 
-        private void WriteMessageCore(HubMessage message, Stream output)
+        private void WriteMessageCore(HubMessage message, Stream packer)
```
neuecc: MessagePack for C# is work on `byte[]` level so write to Stream(create temporary MemoryStream) is too bad.
My suggest of method signature are...

```csharp
public void WriteMessage(HubMessage message, Stream output)
{
    byte[] buffer = new byte[96]; // TODO:should rent from memory pool

    WriteMessageCore(message, ref buffer, out var length);
    BinaryMessageFormatter.WriteLengthPrefix(length, output);
    output.Write(buffer, 0, length);
}

private void WriteMessageCore(HubMessage message, ref byte[] packer, out int length)
{
    switch (message)
     ....
```

BrennanConroy: That's definitely something we'll want to look into in the future, but not for this PR

anurse:
> If we can okay to use C# native Guid representation, we can add UnsafeBinaryResolver

We are fine with GUIDs serializing as strings. We just need to make sure our tests expect that. I actually prefer that they are strings since it is more compatible with JavaScript.

BrennanConroy:
Current issue: Enum deserialization by default doesn't support string values, we can add `DynamicEnumAsStringResolver` to our default resolvers, but then Enum values aren't supported and strings become supported.

anurse:
> DynamicEnumAsStringResolver to our default resolvers, but then Enum values aren't supported and strings become supported.

Is the behavior bidirectional though? I think I'd be OK with enabling `DynamicEnumAsStringResolver` by default as long as it's consistent between serialization and deserialization.

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -373,9 +373,7 @@ describe("hubConnection", () => {
                     ByteArray: protocol.name === "json"
                         ? "aGVsbG8="
                         : new Uint8Array([0x68, 0x65, 0x6c, 0x6c, 0x6f]),
-                    GUID: protocol.name === "json"
-                        ? "00010203-0405-0607-0706-050403020100"
-                        : new Uint8Array([0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x07, 0x06, 0x05, 0x04, 0x03, 0x02, 0x01, 0x00]),
+                    GUID: "00010203-0405-0607-0706-050403020100",
```
anurse: Love that this is consistent between JSON and MessagePack now :)

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -38,7 +38,24 @@ public MessagePackHubProtocol()
 
         public MessagePackHubProtocol(IOptions<MessagePackHubProtocolOptions> options)
         {
-            SerializationContext = options.Value.SerializationContext;
+            var msgPackOptions = options.Value;
+            if (msgPackOptions.FormatterResolvers.Count != SignalRResolver.Resolvers.Count)
+            {
+                _resolver = new CombinedResolvers(msgPackOptions.FormatterResolvers);
+                return;
```
anurse: `return` in a constructor? I don't think I've ever seen that. I don't see why it wouldn't work, but I'm not a fan for some arbitrary reason I can't really articulate... Can we put this in a helper that creates the resolver instead?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/MessagePackHubProtocolTests.cs
```diff
@@ -73,202 +57,164 @@ public static IEnumerable<object[]> TestDataNames
             new ProtocolTestData(
                 name: "InvocationWithNoHeadersAndNoArgs",
                 message: new InvocationMessage(invocationId: "xyz", target: "method", argumentBindingException: null),
-                encoded: Array(HubProtocolConstants.InvocationMessageType, Map(), "xyz", "method", Array()),
```
anurse: RIP nice test failures :(. But I want this change more than I want that.

davidfowl:
Does this library run on the platforms we care about? Xamarin? UWP? .NET Framework (obviously it does here 😄 )

/cc @neuecc

anurse:
It's `netstandard2.0` so it should https://www.nuget.org/packages/MessagePack/

davidfowl:
> It's netstandard2.0 so it should https://www.nuget.org/packages/MessagePack/

Nah, does it use reflection emit on those platforms?

@neuecc have you tested your library on Xamarin and UWP applications?

anurse:
> does it use reflection emit on those platforms

Ah yeah, it does look like it uses Reflection.Emit

neuecc:
Yes, basically MessagePack-CSharp uses `System.Reflection.Emit` and does not provides raw reflection code.
The reason is performance, I decided must not provide slow performance code path.

I'm already using on Unity/IL2CPP(to iOS/Android) environment.
MessagePack-CSharp provides pre code-generation tool, it generates custom formatters and resolver.
I've used it on pre-compile time.
https://github.com/neuecc/MessagePack-CSharp/#pre-code-generationunityxamarin-supports

davidfowl:
> The reason is performance, I decided must not provide slow performance code path.

I get it but the end to end experience is a bit harder for SignalR customers as a result.

neuecc:
If I create `ContractlessReflectionObjectResolver` and replace it only for Xamarin, it may be what we desired.

davidfowl:
> If I create ContractlessReflectionObjectResolver and replace it only for Xamarin, it may be what we desired.

Right that sounds ideal, we can tell people that by default it's slower on platforms that don't support codegen (Xamarin, UWP, Unity) and then recommend that they use the non runtime code gen experience (and hopefully make it as friction free as possible with deep msbuild integration)

BrennanConroy:
https://github.com/neuecc/MessagePack-CSharp/issues/223 to track `ContractlessReflectionObjectResolver`

anurse:
Should we close this PR until the issue is resolved? It's not like keeping it open will stop it from becoming stale ;P

anurse:
@neuecc We're still very interested in switching, but we have to put this on hold for right now since we can't really drop support for Xamarin. We'll be looking at bringing this back before release though! It looks like your library will be a great fit here.

mikaelm12:
Thanks for clarifying @anurse ! I should have commented. We decided to close this during our triage meeting for the exact reason that Andrew mentioned.

neuecc:
Thanks for review and trying.
I've started to create ContractlessReflectionObjectResolver and it will release this weekend.
If you are interested yet, please try again at next time.

davidfowl:
@neuecc we're very interested. 

@BrennanConroy turns out we don't AOT on xamarin today already https://github.com/aspnet/SignalR/issues/1770

PR #1712, opened 2018-03-25
Upgrade the dependencies

davidfowl:


davidfowl:
Fixes the long path issue @JamesNK

PR #1714, opened 2018-03-25
Small changes

davidfowl:
- Don't allocate for empty arrays.
- Don't allocate the list of pre-serialized messages until writing

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMessage.cs
@@ -12,9 +12,8 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
         {
         }
 
-        // Initialize with capacity 2 for the 2 built in protocols
         private object _lock = new object();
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubMessage.cs
```diff
@@ -12,9 +12,8 @@ protected HubMessage()
         {
         }
 
-        // Initialize with capacity 2 for the 2 built in protocols
         private object _lock = new object();
-        private readonly List<SerializedMessage> _serializedMessages = new List<SerializedMessage>(2);
+        private List<SerializedMessage> _serializedMessages;
```
anurse: Does this save much given we always have at least one protocol enabled? Perhaps building a `SmallList<T>` that uses fields for the first two items would be worth it now?

davidfowl: The problem is reads not writes. Allocation up front means we incur this cost even if we never call write.

anurse: Gotcha

PR #1715, opened 2018-03-25
Fix for benchmarks

DamirAinullin:
Incorrect order of arguments.

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/BroadcastBenchmark.cs
@@ -1,14 +1,10 @@
 using System;
 using System.IO.Pipelines;
 using System.Threading;
-using System.Threading.Channels;
 using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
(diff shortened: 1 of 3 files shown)


davidfowl:
These are actually correct. The names of the properties are just confusing. They represent what you are writing to:

https://github.com/aspnet/SignalR/blob/9f0b52562c0bb8e7a17ac89c192338fa93ad9720/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs#L631-L635

anurse:
> The names of the properties are just confusing. They represent what you are writing to

I still go back and forth with these names myself :). I think these are the more sensible names, but I'm not positive :). The idea is that _from_ the **Application**, you write to/read from the **Transport**, so the pipe that the application gets is called **Transport** (and vice-versa). Definitely open to reconsidering these names though.

davidfowl:
im open to renaming the argument names and names of the duplex pipe creation but not the propert names on ConnectionContext.

DamirAinullin:
Yes, it is a bit confusing naming. May be it is better to use prefix "from" and "to", for example FromApplicationToTransport or just FromApplication? At least some comment will be good too I think. BTW thanks for such quick reaction for such small problem.

PR #1564, opened 2018-03-09
Remove support for binary over SSE and add transfer format to negotiation

anurse:
This removes support for Binary protocols over Server-Sent Events **and** removes the Data Encoder concept completely :tada:. 

If you try to handshake with `{ "protocol": "messagepack" }` over Server-Sent Events, the server will abort the connection. **Also** we now include the "Transfer Format" (Text or Binary) supported by each transport in the Negotiate payload, and the TS client (so far) honors that and skips over SSE when attempting to connect using the Message Pack protocol.

TODO Before Merging (Update notices will be posted as comments when they arrive so your notifications show up):
* [x] Tests for TS
* [x] Support for this in .NET Client, including Tests
* [x] Finish merging `ProtocolType` and the newly-renamed `TransferFormat` enum

I want to get this in front of some eyes ASAP (\*cough\* @davidfowl \*cough\*) so he can tell me how much he hates it sooner rather than later ;P.

Other stuff:
* Renamed `TransferMode` to `TransferFormat` (thanks @JamesNK, I like this name much better ;))
* Started changing transports to use `TransferFormat` instead of a separate `ProtocolType` enum.

PR #1718, opened 2018-03-25
IConnection refactoring

anurse:
Ok, here's the beast.

I suggest starting by just reviewing the new version of [`HubConnection`](https://github.com/aspnet/SignalR/blob/7cddcb5f7b829668588d9967a849722228773dab/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs) and [`HttpConnection`](https://github.com/aspnet/SignalR/blob/7cddcb5f7b829668588d9967a849722228773dab/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs) directly, rather than viewing a diff. They've changed pretty dramatically

There are some comments in-line to guide some... discussion questions for the class ;)

I'm going to be doing a bit of tidying up too.


anurse:
FYI: Currently untangling merge conflicts.

anurse:
Merge conflicts untangled.

src/Microsoft.AspNetCore.SignalR.Core/HubClients`T.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class HubClients<THub, T> : IHubClients<T> where THub : Hub
     {
-        private readonly HubLifetimeManager<THub> _lifetimeManager;
+        private HubLifetimeManager<THub> _lifetimeManager;
```
anurse: not sure where this came from, reverting.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -583,6 +576,252 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
...
+                    var result = await _connection.Transport.Input.ReadAsync();
+                    var buffer = result.Buffer;
+                    var consumed = buffer.Start;
+                    var examined = buffer.End;
+
+                    try
+                    {
+                        // read first message out of the incoming data
+                        if (TextMessageParser.TryParseMessage(ref buffer, out var payload))
+                        {
```
davidfowl: Check for a non empty buffer first

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -154,202 +110,179 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
-                            Log.TransportFailed(_logger, transportType, ex);
-                            // Try the next transport
-                            // Clear the negotiation response so we know to re-negotiate.
-                            negotiationResponse = null;
-                        }
+                        // Wait for the transport to complete its side of the pipes
+                        // REVIEW: We don't have to do this. But I think I do prefer that DisposeAsync only completes when everything is indeed complete
+                        // We could just unref the pipe and assume it'll shut down. All our transports will be tested to ensure they properly shut down though.
+                        await _transportPipe.Input.WaitForWriterToComplete();
+                        await _transportPipe.Output.WaitForReaderToComplete();
```
anurse: Open Question: Should we wait for the transport to shut down in `DisposeAsync`? I think so. Related question: Should HubConnection wait for the HttpConnection to dispose? Right now I have it doing so, but we could just unref it and then fire-and-forget Dispose. The difference is in how bugs with HttpConnection shutdown manifest: If we wait for DisposeAsync, they become hangs, if we don't, memory leaks :)

anurse:
Ok, I think it's ready to look at. I do plan to add some tests around how `HttpConnection` behaves with an uncooperative transport (one that doesn't properly shut down). Maybe we aren't worried about that case though?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -583,6 +576,252 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
...
+                while (true)
+                {
+                    var result = await _connection.Transport.Input.ReadAsync();
+                    var buffer = result.Buffer;
+                    var consumed = buffer.End; // TODO: Support partial messages
+                    var examined = buffer.End;
+
+                    try
+                    {
+                        if (buffer.IsEmpty && result.IsCompleted)
```
davidfowl: Copy the sever side loop style

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -583,6 +576,252 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
...
+                catch (Exception ex)
+                {
+                    Log.ErrorDuringClosedEvent(_logger, ex);
+                }
+            });
+        }
+
+        private ValueTask<FlushResult> WriteAsync(byte[] payload) =>
+            _connection.Transport.Output.WriteAsync(payload);
+
```
davidfowl: Did you introduce a lock around the write?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -583,6 +576,252 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
...
+                        if (timeoutTimer != null)
+                        {
+                            Log.ResettingKeepAliveTimer(_logger);
+                            timeoutTimer.Change(ServerTimeout, Timeout.InfiniteTimeSpan);
+                        }
+
+                        // We have data, process it
+                        var (close, exception) = await ProcessMessagesAsync(buffer);
+                        if (close)
```
davidfowl: I think we’ll want to reduce the amount of async state machines here eventually.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: The HttpConnection should wait on the transport to die. The HubConnection should wait for a fixed amount of time for graceful close then cancel its own read loop after.

client-ts/FunctionalTests/FunctionalTests.csproj
```diff
@@ -26,6 +25,11 @@
   </ItemGroup>
 
   <Target Name="ClientBuild" BeforeTargets="AfterBuild">
+      <!-- This will result in a double build of the JavaScript modules. Not sure the best way to deal with that right now. -->
+      <Exec Command="npm run build" WorkingDirectory="$(MSBuildThisFileDirectory)/.." />
```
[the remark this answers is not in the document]
anurse: Merge glitch

samples/ClientSample/RawSample.cs
```diff
@@ -35,36 +35,25 @@ internal static void Register(CommandLineApplication app)
...
 
-                while (!closeTcs.Task.IsCompleted)
-                {
-                    var line = await Task.Run(() => Console.ReadLine(), cts.Token);
-
-                    if (line == null)
-                    {
-                        break;
-                    }
+                _ = Task.Run(() => SendLoop(Console.In, connection.Transport.Output));
```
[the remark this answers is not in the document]
anurse: I think I put that in because I was seeing an issue. I'll remove it.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: It's always done inside the `_startLock`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Agreed, but I'd rather keep the logic in easier-to-understand methods right now and tidy that up afterwards.

samples/SocketsSample/wwwroot/sockets.html
```diff
@@ -38,7 +38,7 @@ <h1 id="transportName">Unknown Transport</h1>
                 event.preventDefault();
             });
 
-            connection.start().then(function () {
+            connection.start(signalR.TransferFormat.Text).then(function () {
```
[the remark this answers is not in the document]
anurse: a) We don't default to anything in the TS client, this sample was straight-up broken
b) I disagree on defautling to text. This is the low-level Connections API, designed to behave like Sockets do, and Sockets default to Binary.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
anurse: Yep, that's what it does currently: https://github.com/aspnet/SignalR/blob/7cddcb5f7b829668588d9967a849722228773dab/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L780

samples/SocketsSample/wwwroot/sockets.html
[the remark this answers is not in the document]
davidfowl: Yea no, the http connection should be text based by default. With this pattern sever sent events would never just work

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
```diff
@@ -13,7 +13,7 @@
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.AspNetCore.Connections;
 
-namespace Microsoft.AspNetCore.Sockets.Client
+namespace Microsoft.AspNetCore.Sockets.Client.Internal
```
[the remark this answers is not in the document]
davidfowl: pubternal is fine for now. Though I dunno why you did it as part of this PR

clients/ts/webdriver-tap-runner/package-lock.json
```diff
@@ -1099,6 +1099,14 @@
         }
       }
     },
+    "string_decoder": {
```
davidfowl: Is this relevant or are you going to revert this?

samples/ClientSample/RawSample.cs
```diff
@@ -78,5 +68,37 @@ internal static void Register(CommandLineApplication app)
...
+        private static async Task ReceiveLoop(TextWriter output, PipeReader input)
+        {
+            while (true)
+            {
+                var result = await input.ReadAsync();
+                var buffer = result.Buffer;
+
+                try
+                {
+                    if (result.IsCanceled || (buffer.IsEmpty && result.IsCompleted))
```
davidfowl: Copy the server side loop pattern:

if (!buffer.IsEmpty)
{
     // Do stuff with buffer
}
else if (result.IsCompleted)
{
    break;
}

davidfowl:
> Ok, I think it's ready to look at. I do plan to add some tests around how HttpConnection behaves with an uncooperative transport (one that doesn't properly shut down), and similarly for HubConnection's behavior with an uncooperative IConnection. Maybe we aren't worried about those cases though?

Not as a requirement before we commit this. This was the problem I hit last time and it halted progress. We have to get this first phase in when we own both sides (HttpConnection and HubConnection). Lets just make sure everything is good with that and then we can fix smaller things.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: Yep thats why I said eventually 😄

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: I hate that the start lock isn't just for starting? Can you rename it?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -154,202 +110,179 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+        private async Task StartAsyncCore(TransferFormat transferFormat)
         {
-            if (ChangeState(from: ConnectionState.Disconnected, to: ConnectionState.Connecting) != ConnectionState.Disconnected)
+            CheckDisposed();
+
+            if (_started)
             {
-                return Task.FromException(
-                    new InvalidOperationException($"Cannot start a connection that is not in the {nameof(ConnectionState.Disconnected)} state."));
+                return;
```
JamesNK: Log SkippingStart?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +109,283 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
+                {
+                    // We're already connected
+                    return;
+                }
+
+                CheckDisposed();
+
+                Log.Starting(_logger);
+
+                _connectionActive = new CancellationTokenSource();
```
davidfowl: It looks like this is used as a glorified bool. Where is it passed in?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -154,202 +110,179 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
 
-                    // We're going to search for the transfer format as a string because we don't want to parse
-                    // all the transfer formats in the negotiation response, and we want to allow transfer formats
-                    // we don't understand in the negotiate response.
-                    var transferFormatString = transferFormat.ToString();
-
-                    foreach (var transport in negotiationResponse.AvailableTransports)
+                    // Stop the transport, but we don't care if it throws.
+                    // The transport should also have completed the pipe with this exception.
+                    // REVIEW: We could make not throwing part of the contract here.
```
JamesNK: Unless something goes very wrong, throwing in dispose/close feels bad to me.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +109,283 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
+
+                // Check if we finished stopping (the check above proved that we were asked to stop)
+                // (We're the only ones who set _receiveTask, so this is thread-safe, because we're in our lock)
+                if (_receiveTask != null && !_receiveTask.IsCompleted)
```
davidfowl: Who is this for? People that call StopAsync then StartAsync without waiting on the StopAsync?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +109,283 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
 
-        public async Task<ChannelReader<object>> StreamAsChannelAsync(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default)
+        public async Task<ChannelReader<object>> StreamAsChannelAsync(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default) =>
+            await StreamAsChannelAsyncCore(methodName, returnType, args, cancellationToken).ForceAsync();
+
+        public async Task<object> InvokeAsync(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default) =>
+            await InvokeAsyncCore(methodName, returnType, args, cancellationToken).ForceAsync();
+
+        // REVIEW: We don't generally use cancellation tokens when writing to a pipe because the asynchrony is only the result of backpressure.
+        // However, this would be the only "invocation" method _without_ a cancellation token... which is odd.
```
davidfowl: We can still pass it to flush to cancel the flush if it *is* async

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +109,283 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
+                await HandshakeAsync();
+
+                // From here on, "ReceiveLoop" is in charge of the connection.
+                _receiveTask = ReceiveLoop();
+                Log.Started(_logger);
+            }
+            catch (Exception ex) when (_connection != null)
+            {
+                Log.ErrorStartingConnection(_logger, ex);
+                // We can't catch exceptions thrown by RecieveLoop because we don't await it, so we know it hasn't started yet.
```
davidfowl: Not really, if it throws before the first await..

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +109,283 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
+
+                // If we got partially started, we need to shut down the connection before we rethrow.
+                await _connection.DisposeAsync();
```
davidfowl: It might be cleaner to have a smaller try catch so we know exactly how far we got when we decide this was "ok" to do. I'm a big fan of letting everything unwind then calling dispose from a single place (event loopy).

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +109,283 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
+            finally
             {
-                Log.SendInvocationFailed(_logger, invocationMessage.InvocationId, ex);
-                throw;
+                _startLock.Release();
             }
         }
 
-        private async Task OnDataReceivedAsync(byte[] data)
+        private async Task<(bool close, Exception exception)> ProcessMessagesAsync(ReadOnlySequence<byte> buffer)
```
davidfowl: tuples...

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -583,6 +576,252 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
...
+            // Cancel any outstanding invocations
+            CancelOutstandingInvocations(closeException);
+
+            // Fire-and-forget the closed event
+            _ = Task.Run(() =>
+            {
+                try
+                {
+                    Log.InvokingClosedEventHandler(_logger);
+                    Closed?.Invoke(closeException);
```
davidfowl: This it the only place Closed gets called from now right?

src/Microsoft.AspNetCore.Sockets.Client.Http/PipeCompletionExtensions.cs
```diff
@@ -0,0 +1,41 @@
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    internal static class PipeCompletionExtensions
```
davidfowl: Why is this in product code if it's only used by tests?

src/Microsoft.AspNetCore.Sockets.Client.Http/PipeCompletionExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+                }
+                else
+                {
+                    ((TaskCompletionSource<object>)state).TrySetResult(null);
+                }
+            }, tcs);
+            return tcs.Task;
+        }
+
+        public static Task WaitForReaderToComplete(this PipeWriter writer)
```
davidfowl: Where is this used?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -847,11 +865,11 @@ public static IEnumerable<object[]> HubProtocolsAndTransportsAndHubPaths
         // This list excludes "special" hub paths like "default-nowebsockets" which exist for specific tests.
         public static string[] HubPaths = new[] { "/default", "/dynamic", "/hubT" };
 
-        public static IEnumerable<IHubProtocol> HubProtocols =>
-            new IHubProtocol[]
+        public static IDictionary<string, IHubProtocol> HubProtocols =>
```
davidfowl: nit: Just use concrete Dictionary

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -154,202 +110,179 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
         }
 
         public Task StartAsync() => StartAsync(TransferFormat.Binary);
```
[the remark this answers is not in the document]
davidfowl: Yea, can we kill the parameterless one? That could be an extension if we wanted to keep it.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: I still think this should wait on the previous call.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -154,202 +110,179 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
+                // We're going to search for the transfer format as a string because we don't want to parse
+                // all the transfer formats in the negotiation response, and we want to allow transfer formats
+                // we don't understand in the negotiate response.
+                var transferFormatString = transferFormat.ToString();
 
-                    Log.CompleteClosed(_logger);
-                    _logScope.ConnectionId = null;
+                foreach (var transport in negotiationResponse.AvailableTransports)
+                {
+                    // If we get disposed while starting, stop ASAP.
```
davidfowl: Why?

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -154,202 +110,179 @@ public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactor
...
-                    {
-                        Log.HttpConnectionClosed(_logger);
-                        return;
-                    }
+                    Log.DisposingHttpConnection(_logger);
 
-                    // This should only need to happen once
-                    connectUrl = CreateConnectUrl(Url, negotiationResponse.ConnectionId);
+                    // Complete our ends of the pipes.
+                    _transportPipe.Input.Complete(exception);
```
davidfowl: This will noop in the HubConnection case right? DisposeAsync should only ever be called once the HubConnection completes reading and writing right? Otherwise you'll see errors about trying to call GetMemory after completing the pipe.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
davidfowl: Ok I took another look at this and had flashbacks to when I tried.

- What should the behavior be if I create an HttpConnection, call StartAsync and then call DisposeAsync()? Currently it won't hang because you're forcefully completing the transport pipe no matter what. We need to make that part of the contract. DisposeAsync will complete the pipes so you cannot call it and then use the pipes.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -19,18 +19,17 @@ public class HubConnectionProtocolTests
         public async Task SendAsyncSendsANonBlockingInvocationMessage()
         {
             var connection = new TestConnection();
-            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+            var hubConnection = new HubConnection(() => connection, new JsonHubProtocol(), new LoggerFactory());
```
JamesNK: Create a helper method for creating HubConnection and reuse through tests. Helpers avoid updating dozens of tests each time the ctor changes.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
```diff
@@ -42,15 +41,19 @@ public class HubConnectionProtocolTests
...
         {
-            var connection = new TestConnection();
-            var hubConnection = new HubConnection(connection, new JsonHubProtocol(), new LoggerFactory());
+            var connection = new TestConnection(autoNegotiate: false);
+            var hubConnection = new HubConnection(() => connection, new JsonHubProtocol(), new LoggerFactory());
             try
             {
-                await hubConnection.StartAsync();
+                // We can't await StartAsync because it depends on the negotiate process!
+                var startTask = hubConnection.StartAsync().OrTimeout();
```
JamesNK: Is the standard to put the OrTimeout with the await or is that my imagination?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -0,0 +1,328 @@
...
+            // A helper that we wouldn't want to use in product code, but is fine for testing until IAsyncDisposable arrives :)
+            private static async Task AsyncUsing(HubConnection connection, Func<HubConnection, Task> action)
+            {
+                try
+                {
+                    await action(connection);
+                }
+                finally
+                {
+                    await connection.DisposeAsync();
```
JamesNK: OrTimeout?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -0,0 +1,328 @@
...
+
+                    // We should be stopped now
+                    var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => connection.SendAsync("Foo").OrTimeout());
+                    Assert.Equal($"The '{nameof(HubConnection.SendAsync)}' method cannot be called if the connection is not active", ex.Message);
+                });
+            }
+
+            [Theory]
+            [MemberData(nameof(MethodsNamesThatRequireActiveConnection))]
+            public async Task MethodsThatRequireActiveConnectionWaitForStopAndFailIfConnectionIsCurrentlyStopping(string name)
```
JamesNK: name -> methodName

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.AspNetCore.SignalR.Client.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
JamesNK: Hmmm? Since it is just test related code, why not make it public?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Oh yeah, it's a terrible name. Will rename.

samples/SocketsSample/wwwroot/sockets.html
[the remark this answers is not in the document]
anurse: We should have the Server-Sent Events transport detect non-Text data. The problem is that if we just give you SSE without you clearly indicating you want Text, then you can write random binary data and you'll just get weird protocol glitches and errors.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: It was a CTS before. Now that we lock more, it can probably be a `bool`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Correct

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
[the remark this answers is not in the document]
anurse: Do you not [want it](https://github.com/aspnet/SignalR/pull/1564#discussion_r174016719) anymore? Or do you want it as a default parameter or extension method? Just want clarification there.

I'm happy to remove the default entirely, but that wouldn't match with what you suggested in the TypeScript client. Just trying to get consistency here.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
anurse: Yep, we talked about that, I'll put it in. I did it in HubConnection already

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
anurse: Just to clarify, are you talking about making not throwing part of the `ITransport.StopAsync` contract? That's the exception we're suppressing here. It should be fine but I'll have to review the transport behavior.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: I'm going to need to clarify your position on `CancellationToken` and Pipelines; I'm getting conflicting messages ;)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: It's an `async` method, don't they still wrap exceptions above the first `await` in a faulted `Task`?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: internal method and `out` doesn't work because async.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Correct.

src/Microsoft.AspNetCore.Sockets.Client.Http/PipeCompletionExtensions.cs
anurse: `HttpConnection.DisposeAsync`

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Properties/AssemblyInfo.cs
anurse: It was late ;P

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: I don't want to flow it everywhere but cases like this we can pass it directly to the pipe API and let it do what it does.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: 👍

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
anurse: I had a version of this that worked without the lock originally. I removed that and didn't tidy these up.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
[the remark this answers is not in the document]
anurse: Spent a lot of time reviewing the transport code making sure it shutdown the way I expected it to. So I was there, and it felt easier to do it than filing an issue or opening a separate PR. Happy to revert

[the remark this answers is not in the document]
davidfowl: Nah, no revert now.

clients/ts/webdriver-tap-runner/package-lock.json
anurse: I believe it happens when the version of NPM is different maybe? I'm going to update to the latest released NPM and run `npm i` to see if that fixes it (my home desktop hasn't been updated in a while). If that doesn't fix it, I'll presume that this is the "new" ordering and we should check it in anyway.

anurse: Yep. Updated NPM and ran `./build /t:Restore` and everything went back to normal.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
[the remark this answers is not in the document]
davidfowl: Meh, keep it for now.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +109,283 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
-            // The stream invocation will be canceled by the CancelInvocationMessage, connection closing, or channel finishing.
-            using (cancellationToken.Register(token => ((CancellationTokenSource)token).Cancel(), invokeCts))
+        private async Task DisposeAsyncCore()
+        {
+            // Block a Start from happening until we've finished resolving.
+            await _startLock.WaitAsync();
+            try
             {
-                await InvokeStreamCore(methodName, irq, args);
+                if (_disposed)
```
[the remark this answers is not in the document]
anurse: Are you suggesting *adding* a check before the lock? We actually do that in `DisposeAsync` before we force ourselves onto the thread pool.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Yeah, this needed some tidying up.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -583,6 +576,252 @@ private bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
...
+            else
+            {
+                Log.TimedOutWaitingForReceiveLoop(_logger);
+                // We timed out waiting for the server/transport to shut down, cancel it, this will cause
+                // a TimeoutException indicating the server timeout elapsed to be signalled as
+                // the exception in the Closed event.
+                _connection.Transport.Input.CancelPendingRead();
+
+                Log.WaitingForCanceledReceiveLoop(_logger);
+                await _receiveTask;
```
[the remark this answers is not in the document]
anurse: No, I don't think we should catch it. Since we cancel it by cancelling the pipe, it will throw `TimeoutException` (see the `if(result.IsCanceled)` clause in `ReceiveLoop`) and we want that exception to manifest to the user since the server/transport did time out.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
anurse: Actually, it already does. Since `_started` is set at the end, then it will only be true if the connection is fully started. If the second call passes this check, it takes the lock which waits on the first call to finish. Once it's finished, `_started` is `true` and it no-ops.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.Log.cs
```diff
@@ -5,236 +5,124 @@
...
 
-            private static readonly Action<ILogger, Exception> _httpConnectionClosed =
-                LoggerMessage.Define(LogLevel.Debug, new EventId(2, "HttpConnectionClosed"), "Connection was closed from a different thread.");
+            private static readonly Action<ILogger, Exception> _skippingStart =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(2, "SkippingStart"), "Skipping start, connection is already started.");
 
-            private static readonly Action<ILogger, string, Uri, Exception> _startingTransport =
-                LoggerMessage.Define<string, Uri>(LogLevel.Debug, new EventId(3, "StartingTransport"), "Starting transport '{Transport}' with Url: {Url}.");
+            private static readonly Action<ILogger, Exception> _started =
+                LoggerMessage.Define(LogLevel.Information, new EventId(3, "Started"), "HttpConnection Started.");
```
BrennanConroy: lowercase started

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.Log.cs
```diff
@@ -5,236 +5,124 @@
...
 
-            private static readonly Action<ILogger, Exception> _processRemainingMessages =
-                LoggerMessage.Define(LogLevel.Debug, new EventId(4, "ProcessRemainingMessages"), "Ensuring all outstanding messages are processed.");
+            private static readonly Action<ILogger, Exception> _skippingDispose =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(5, "SkippingDispose"), "Skipping dispose, connection is already disposed.");
 
-            private static readonly Action<ILogger, Exception> _drainEvents =
-                LoggerMessage.Define(LogLevel.Debug, new EventId(5, "DrainEvents"), "Draining event queue.");
+            private static readonly Action<ILogger, Exception> _disposed =
+                LoggerMessage.Define(LogLevel.Information, new EventId(6, "Disposed"), "HttpConnection Disposed.");
```
BrennanConroy: lowercase disposed

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -92,60 +92,16 @@ public HttpConnection(Uri url, TransportType transportType, ILoggerFactory logge
...
-
-            return httpClient;
-        }
-
         public HttpConnection(Uri url, ITransportFactory transportFactory, ILoggerFactory loggerFactory, HttpOptions httpOptions)
         {
             Url = url ?? throw new ArgumentNullException(nameof(url));
-            _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
-            _logger = _loggerFactory.CreateLogger<HttpConnection>();
+            var loggerFactory1 = loggerFactory ?? NullLoggerFactory.Instance;
```
BrennanConroy: `loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;`

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
anurse: Yes, the goal here is to force the pipes to be completed if they weren't already. It's expected that the caller will have already completed the pipes before disposing.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionProtocolTests.cs
anurse: I don't know that we have a standard. I've mostly seen it with the async method itself, but I kinda agree it would be better with the `await`.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
anurse: So, I had `.OrTimeout` here before, but the idea here was that tests that use this aren't actually testing `DisposeAsync`. A timeout error in the finally block would suppress any exception from the test itself, so I was actually thinking of firing-and-forgetting this.

anurse: Added a comment to that effect. Happy to continue the discussion on whether we should `.OrTimeout` here though.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Properties/AssemblyInfo.cs
anurse: Oh, I remember why. It's because the `internal` type comes from a linked file that is in our product code. It has to be internal there. We could `#ifdef` the accessibility modifier I suppose but that gets gross.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs
```diff
@@ -125,6 +125,60 @@ private static class Log
...
+            private static readonly Action<ILogger, Exception> _waitingForPreviousStop =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(42, "WaitingForPreviousStop"), $"Waiting for previous {nameof(StopAsync)} to complete before restarting.");
+
+            private static readonly Action<ILogger, Exception> _errorStartingConnection =
+                LoggerMessage.Define(LogLevel.Error, new EventId(43, "ErrorStartingConnection"), "Error starting connection.");
+
+            private static readonly Action<ILogger, Exception> _started =
+                LoggerMessage.Define(LogLevel.Information, new EventId(44, "Started"), "HubConnection started.");
+
+            private static readonly Action<ILogger, string, Exception> _sendingCancelation =
```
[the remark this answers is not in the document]
anurse: YOUR LANGUAGE IS WEIRD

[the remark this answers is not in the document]
anurse: Seriously.

American English:
* Cancel
* Canceled
* Canceling
* Cancel**l**ation

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -421,302 +354,79 @@ private static Uri CreateConnectUrl(Uri url, string connectionId)
...
+                await transport.StartAsync(connectUrl, pair.Application, transferFormat, this);
             }
             catch (Exception ex)
             {
                 Log.ErrorStartingTransport(_logger, _transport, ex);
                 _transport = null;
                 throw;
             }
+
+            // We successfully started, set the transport properties (we don't want to set these until the transport is definitely running.
```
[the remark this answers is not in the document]
anurse: Brennan Conroy, Comment Compiler.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.Log.cs
```diff
@@ -38,6 +39,11 @@ private static class Log
             private static readonly Action<ILogger, Uri, Exception> _errorPolling =
                 LoggerMessage.Define<Uri>(LogLevel.Error, new EventId(9, "ErrorPolling"), "Error while polling '{PollUrl}'.");
 
+            // long? does properly format as "(null)" when null.
+            private static readonly Action<ILogger, int, long?, Exception> _pollResponseReceived =
+                LoggerMessage.Define<int, long?>(LogLevel.Error, new EventId(10, "PollResponseReceived"),
```
BrennanConroy: Not error

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs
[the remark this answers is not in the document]
JamesNK: Commonwealth English is true English 👑

clients/ts/FunctionalTests/FunctionalTests.csproj
```diff
@@ -5,6 +5,18 @@
...
+    <None Remove="selenium\run-ci-tests.ts" />
+    <None Remove="selenium\run-tests.ts" />
+    <None Remove="ts\Common.ts" />
+    <None Remove="ts\ConnectionTests.ts" />
+    <None Remove="ts\HubConnectionTests.ts" />
+    <None Remove="ts\index.ts" />
+    <None Remove="ts\Utils.ts" />
+    <None Remove="ts\WebDriverReporter.ts" />
+    <None Remove="ts\WebSocketTests.ts" />
+  </ItemGroup>
```
JamesNK: hmm?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,179 +108,288 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
+                    return;
+                }
+
+                CheckDisposed();
+
+                Log.Starting(_logger);
+
+                // Check if we finished stopping (the check above proved that we were asked to stop)
+                // (We're the only ones who set _receiveTask, so this is thread-safe, because we're in our lock)
+                if (_receiveTask != null && !_receiveTask.IsCompleted)
```
JamesNK: Copy `_recieveTask` to a local variable so there is no chance of an error because of the field getting set to null while using it here.

KPixel:
Quick one: `Microsoft.AspNetCore.Sockets.Http.csproj` no longer needs a reference to `Microsoft.AspNetCore.Sockets.Abstractions.csproj`. (Only .Client projects reference it now.)

clients/ts/FunctionalTests/FunctionalTests.csproj
anurse: It keeps coming back! I don't know what's causing it 😭

davidfowl:
> Quick one: Microsoft.AspNetCore.Sockets.Http.csproj no longer needs a reference to Microsoft.AspNetCore.Sockets.Abstractions.csproj. (Only .Client projects reference it now.)

It does because of the TransferMode enum.

KPixel:
> It does because of the TransferMode enum.

I assume you meant TransferFormat, and it is now in Connections.Abstractions.

[Sockets.Abstractions only contains IConnection](https://github.com/aspnet/SignalR/tree/anurse/1475-02-clean-up-httpconnection/src/Microsoft.AspNetCore.Sockets.Abstractions) which is only used by the Client projects.

anurse:
:up: :date: I've refactored the code to handle shut down better. I also had to remove the code that passed a `cancellationToken` in to `PipeWriter.WriteAsync` in `HubConnection.WriteAsync` because it seemed to be causing the later `PipeWriter.Complete` to not actually signal to the reader that it was complete. I think something else is going on there, but removing the token gets to passing tests, which is a good place to be while I investigate it so I committed it. Will be reviewing that code and checking in with @davidfowl about that in parallel.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -159,181 +104,281 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
+            AssertConnectionValid();
 
-            var invocationMessage = new InvocationMessage(null, target: methodName,
-                argumentBindingException: null, arguments: args);
+            var payload = _protocol.WriteToArray(hubMessage);
 
-            ThrowIfConnectionTerminated(invocationMessage.InvocationId);
+            Log.SendingMessage(_logger, hubMessage);
+            // REVIEW: If a token is passed in and is cancelled during FlushAsync it seems to break .Complete()...
+            await WriteAsync(payload, CancellationToken.None);
```
anurse: https://github.com/dotnet/corefx/issues/28492

davidfowl:
Why is the ci red?

anurse:
On it. I think there actually is a test that is broken. I may have spoken slightly too soon about the tests passing ;). Shouldn't be too tricky though.

src/Microsoft.AspNetCore.Sockets.Client.Http/ReadResultExtensions.cs
```diff
@@ -0,0 +1,13 @@
+namespace System.IO.Pipelines
+{
+    public static class ReadResultExtensions
+    {
+        public static void ThrowIfCanceled(this ReadResult readResult)
```
davidfowl: Where is this used?

anurse: It isn't anymore! Removed.

clients/ts/FunctionalTests/FunctionalTests.csproj
```diff
@@ -5,6 +5,18 @@
     <TypeScriptCompileBlocked>True</TypeScriptCompileBlocked>
   </PropertyGroup>
 
+  <ItemGroup>
```
davidfowl: Fix eeet

samples/SocketsSample/web.config
```diff
@@ -7,6 +7,8 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" startupTimeLimit="3600" requestTimeout="23:00:00" />
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" startupTimeLimit="3600" requestTimeout="23:00:00">
```
davidfowl: undo

davidfowl:
> I figured out why some of the lifecycle tests are flaky. It's because we now dispose the underlying connection outside the lock, meaning they can't guarantee the ordering of events any more (since the connection dispose was how we were synchronizing) without adding internal hooks. I'm going to refactor those tests this afternoon.

Should we skip them for now then?

davidfowl:
Yea the ConnectionLifecycle tests are flaky with this change.

JamesNK:
IMO skip new flakey tests and fix them in a follow up PR.

anurse:
> Should we skip them for now then?

No need, the fix is fairly simple. It just means it's harder for the test to guarantee that (for example) it's verifying that an `InvokeAsync` that happens while a `DisposeAsync` _is still processing_ waits for the dispose to complete and **then** fails. It's not a big deal as long as the observed behavior is correct (which is that the `InvokeAsync` throws an ODE)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -822,15 +815,23 @@ public Task InvokeAsync(object[] parameters)
...
         {
+            private volatile bool _stopping;
             private readonly HubConnection _hubConnection;
 
             private TaskCompletionSource<object> _stopTcs;
             private readonly object _lock = new object();
             private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>();
             private int _nextId;
-
+
```
davidfowl: nit: whitespace

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -922,36 +923,19 @@ public Task StopAsync(TimeSpan timeout)
                 Log.Stopping(_hubConnection._logger);
 
                 // Complete our write pipe, which should cause everything to shut down
-                Log.CompletingTransportPipe(_hubConnection._logger);
+                Log.TerminatingReceiveLoop(_hubConnection._logger);
                 try
                 {
-                    Connection.Transport.Output.Complete();
+                    Connection.Transport.Input.CancelPendingRead();
```
davidfowl: This can't throw.

davidfowl:
@anurse you are sleeping but I'm going to merge this. I ran it locally on OSX and it passes 😄

PR #1721, opened 2018-03-26
Don't create the span on netstandard

davidfowl:
- Directly pin the char[]


davidfowl:
You could say the same needs to be tested with any text reader implementation right? I think it’s not worth the effort unless there was a test suite we could use. So I’m not sold there’s value as it’s used in a single place. We just need to make sure the logic is correct and we need to makes sure we override what JsonTextReader uses.

JamesNK:
The tests don't need to be incredibly thorough - Json.NET uses that one method after all - I want to make sure that:
1. The method args are correctly used
2. Memory is completely read without accidentally skipping any bytes when it is so large that multiple Read calls are required to put everything into the buffer.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextReader.cs
```diff
@@ -11,10 +11,12 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
     internal class Utf8BufferTextReader : TextReader
     {
         private ReadOnlyMemory<byte> _utf8Buffer;
+        private Decoder _decoder;
 
         public Utf8BufferTextReader(ReadOnlyMemory<byte> utf8Buffer)
         {
             _utf8Buffer = utf8Buffer;
+            _decoder = Encoding.UTF8.GetDecoder();
```
anurse: Can this be cached statically?

davidfowl:
> LGTM. Could we just have a test or two that uses JsonTextReader? Then we are truly testing that the methods JsonTextReader uses are working.

That's what all of the JsonHubProtocol tests do.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextReader.cs
davidfowl: It's stateful and resttable so sorta but not really. Also the things did slow down a bit but it is more correct (after writing some unicode tests).

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
```diff
@@ -58,7 +58,8 @@ private static JsonTextWriter CreateJsonTextWriter(Stream output)
 
         private static JsonTextReader CreateJsonTextReader(ReadOnlyMemory<byte> payload)
         {
-            var textReader = new Utf8BufferTextReader(payload);
+            var textReader = new Utf8BufferTextReader();
```
JamesNK: Why doesn't this place use `Utf8BufferTextReader.Get(payload);`?

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextReader.cs
```diff
@@ -11,47 +11,64 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
...
-            _utf8Buffer = utf8Buffer;
+            _decoder = Encoding.UTF8.GetDecoder();
         }
 
-        public override int Read(char[] buffer, int index, int count)
+        public static Utf8BufferTextReader Get(ReadOnlyMemory<byte> utf8Buffer)
         {
-            if (_utf8Buffer.IsEmpty)
+            var reader = _cachedInstance;
+            if (reader == null)
```
JamesNK: When reader is not null I want a flag set on it to say that it is in use, and the flag is reset on `Return`. If `Get` is called and the cached instance is already in use then throw a friendly error message.

It probably won't happen but better to be safe the sorry.

davidfowl: @JamesNK I got you covered bro 😄 .

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
davidfowl: I didn't want to change the pattern. This creates the JsonTextReader, I want to put a try finally around the whole operation that does the return. I'll fix it up in another PR.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextReader.cs
```diff
@@ -11,47 +11,82 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
...
     {
         private ReadOnlyMemory<byte> _utf8Buffer;
+        private Decoder _decoder;
 
-        public Utf8BufferTextReader(ReadOnlyMemory<byte> utf8Buffer)
+        [ThreadStatic]
+        private static Utf8BufferTextReader _cachedInstance;
+
+#if DEBUG
+        private bool _inUse;
```
anurse: Isn't this going to block up if two threads are running parses? This should be thread static too, right?

JamesNK: It's a field on the class, and the class is on the thread static.

Issue #1391, opened 2018-02-01
Ensure we are tracing connection/request URLs on the client


anurse:
While investigating #1389 I noticed that we don't really trace the URLs that we are connecting to or the HTTP requests going out. So when an error occurs, we can't easily trace it back to the request that triggered it. We should ensure our transports trace out useful data about the outbound requests, and consider attaching a logging handler to the `HttpClient` we use

davidfowl:
We may need to scrub querystrings

anurse:
We should **definitely** scrub querystrings.

BrennanConroy:
?

anurse:
Just confirming that we should scrub querystrings, and moving this to the appropriate milestone :)

JamesNK:
What do we mean by scrub querystrings? URL only with no query string at all?

BrennanConroy:
`https://localhost:5000?id=1234&secret_token=super-secret` will become `https://localhost:5000?id=1234`

davidfowl:
That's a good question, are we stripping all query strings from logs as user data?

BrennanConroy:
I think everything except `id` should be stripped

anurse:
> That's a good question, are we stripping all query strings from logs as user data?

Yeah, I kinda dodged that because I'm not sure the best way to approach it. Generally, query strings should be loggable (as you're not supposed to put authentication in there). We're forced to put auth tokens in the query string in one place: WebSockets on the **JavaScript** Client. I think we can log query strings everywhere **except** in the WebSocket transport.

davidfowl:
It's a bit messy. People will be using these logs to look at what gets sent and filtering that out feels wrong.

JamesNK:
FYI the entire URL is already logged in the TS client

https://github.com/aspnet/SignalR/blob/d367bdc9aa7856be4c9b22c2fe97532f16d6641a/clients/ts/signalr/src/Transports.ts#L64

davidfowl:
I actually think this is fine.

PR #1723, opened 2018-03-26
Added logging of .NET client HTTP requests

JamesNK:
https://github.com/aspnet/SignalR/issues/1391

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -27,7 +27,9 @@ namespace Microsoft.AspNetCore.Sockets.Client
     public partial class HttpConnection : IConnection
     {
         private static readonly TimeSpan HttpClientTimeout = TimeSpan.FromSeconds(120);
+#if !NETCOREAPP2_1
         private static readonly Version Windows8Version = new Version(6, 2);
+#endif
 
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -100,8 +100,8 @@ namespace Microsoft.AspNetCore.Sockets.Client
 
         private HttpClient CreateHttpClient()
         {
-            HttpMessageHandler httpMessageHandler = null;
             var httpClientHandler = new HttpClientHandler();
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LoggingHttpMessageHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Microsoft.AspNetCore.Sockets.Client.Http.Internal
+{
+    public class LoggingHttpMessageHandler : DelegatingHandler
+    {
+        private readonly ILogger<LoggingHttpMessageHandler> _logger;
+
+        public LoggingHttpMessageHandler(HttpMessageHandler inner, ILoggerFactory loggerFactory) : base(inner)
+        {
+            _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<LoggingHttpMessageHandler>();
```
davidfowl: You don't need to be this defensive here. Just don't accept a null logger factory.

davidfowl:
So we separately log web sockets?

anurse:
Yeah, WebSockets don't use `HttpClient` (well technically the new implementation does but as an implementation detail) so we should also make sure we're logging the URL [there](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Sockets.Client.Http/WebSocketsTransport.cs#L327)

davidfowl:
@JamesNK need some websocket love

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -135,7 +139,10 @@ private HttpClient CreateHttpClient()
                 }
             }
 
-            var httpClient = httpMessageHandler == null ? new HttpClient() : new HttpClient(httpMessageHandler);
+            // Wrap message handler in a logging handler last to ensure it is always present
+            httpMessageHandler = new LoggingHttpMessageHandler(httpMessageHandler, _loggerFactory);
```
BrennanConroy: Might be worth checking if `_loggerFactory` is a `NullLoggerFactory`. Not necessary for this PR though

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LoggingHttpMessageHandler.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
```
BrennanConroy: copyright

PR #1728, opened 2018-03-26
Use more pooled Utf8BufferTextReaders

davidfowl:
- Use it in the HandshakeProtocol

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
@@ -56,33 +56,43 @@ namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
             return new JsonTextWriter(new StreamWriter(output, _utf8NoBom, 1024, leaveOpen: true));
         }
 
-        private static JsonTextReader CreateJsonTextReader(ReadOnlyMemory<byte> payload)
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
```diff
@@ -56,33 +56,43 @@ private static JsonTextWriter CreateJsonTextWriter(Stream output)
             return new JsonTextWriter(new StreamWriter(output, _utf8NoBom, 1024, leaveOpen: true));
         }
 
-        private static JsonTextReader CreateJsonTextReader(ReadOnlyMemory<byte> payload)
+        private static JsonTextReader CreateJsonTextReader(Utf8BufferTextReader textReader)
```
JamesNK: Move this to JsonUtils

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/JsonHubProtocol.cs
```diff
@@ -111,6 +111,8 @@ private HubMessage ParseMessage(TextReader textReader, IInvocationBinder binder)
                 using (var reader = new JsonTextReader(textReader))
```
JamesNK: Use new `JsonUtils.CreateJsonTextReader(textReader)` method to create JsonReader

Issue #1677, opened 2018-03-21
SignalR Client add the ability to unregister a handler without passing in the function handler


spudcud:
Right now the actual function handler has to be passed in to off in order to unregister it but keeping track of the handler is not desired. I want to unregister the event itself and don't care about what functions are registered to it. This used to be possible in the old SignalR version and would be nice to get added back in. I understand the desire to unregister a specific handler but unregistering all is useful too as we only ever have one per event.

`connection.off("myEvent", myHandler );`

Desired:
`connection.off("myEvent");`

davidfowl:
Seems reasonable. Care to send a pull request?

spudcud:
I'll test out some changes and then I'll look into figuring out how to submit the PR. Haven't done it yet so it will be a learning experience :)

davidfowl:
@spudcud We're here for moral support as well as technical support 😄.

spudcud:
Haha, thanks David!!

spudcud:
Morning. So, I unfortunately can't create a pull request. I have not been able to test the changes using dev latest. You can see the changes in the fork and here is the main change that I would recommend to allow for this feature and I added a test for it as well. I am not sure how to run the test or compile the typescript at this point. If you have any questions please reach out to me privately.

```javascript
    public off(methodName: string, method?: (...args: any[]) => void) {
        if (!methodName) {
            return;
        }

        methodName = methodName.toLowerCase();
        const handlers = this.methods[methodName];
        if (!handlers) {
            return;
        }

        if (method) {
            const removeIdx = handlers.indexOf(method);
            if (removeIdx !== -1) {
                handlers.splice(removeIdx, 1);
                if (handlers.length === 0) {
                    delete this.methods[methodName];
                }
            }
        }
        else {
            delete this.methods[methodName];
        }
    }
```

anurse:
Since you have a test already, feel free to just open a PR now and we can iterate on it there (the CI will automatically run the tests so you could luck out and everything could just work ;)). As I said above, feel free to reach out to me for assistance!

mikaelm12:
Done here: 3952eacfaa03af3519cb0d5666d2952bfa5f3de3

spudcud:
Thanks @mikaelm12 for getting it done. It will make my life a lot easier :)

mikaelm12:
@spudcud Glad to hear it!

PR #1729, opened 2018-03-26
Add ability to remove all handlers with just the method name

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1677
This allows you to remove all handlers for a method with just the method name. 
You can still pass in the method name and handler to off to just remove that specific call back.
I added tests for both scenarios.

### DIFF
--- a/clients/ts/signalr/spec/HubConnection.spec.ts
+++ b/clients/ts/signalr/spec/HubConnection.spec.ts
@@ -272,7 +272,6 @@ describe("HubConnection", () => {
         });
 
         it("all handlers can be unregistered with just the method name", async () => {
-            const warnings: string[] = [];
             const connection = new TestConnection();
             const hubConnection = new HubConnection(connection);
 
@@ -280,8 +279,9 @@ describe("HubConnection", () => {
 
             let count = 0;
             const handler = () => { count++; };
+            const secondHandler = () => { count++; };
             hubConnection.on("inc", handler);
-            hubConnection.on("inc", handler);
+            hubConnection.on("inc", secondHandler);
 
             connection.receive({
                 arguments: [],
@@ -305,7 +305,6 @@ describe("HubConnection", () => {
         });
 
         it("a single handler can be unregistered with the method name and handler", async () => {
-            const warnings: string[] = [];
             const connection = new TestConnection();
             const hubConnection = new HubConnection(connection);
 
@@ -313,8 +312,9 @@ describe("HubConnection", () => {
 
             let count = 0;
             const handler = () => { count++; };
+            const secondHandler = () => { count++; };
             hubConnection.on("inc", handler);
-            hubConnection.on("inc", handler);
+            hubConnection.on("inc", secondHandler);
 
             connection.receive({
                 arguments: [],
@@ -335,6 +335,28 @@ describe("HubConnection", () => {
             });
 
             expect(count).toBe(3);
+        });
+
+        it("can't register the same handler multiple times", async () => {
+            const connection = new TestConnection();
+            const hubConnection = new HubConnection(connection);
+
+            connection.receiveHandshakeResponse();
+
+            let count = 0;
+            const handler = () => { count++; };
+            hubConnection.on("inc", handler);
+            hubConnection.on("inc", handler);
+
+            connection.receive({
+                arguments: [],
+                invocationId: "0",
+                nonblocking: true,
+                target: "inc",
+                type: MessageType.Invocation,
+            });
+
+            expect(count).toBe(1);
         });
 
         it("callback invoked when servers invokes a method on the client", async () => {
--- a/clients/ts/signalr/spec/HubConnection.spec.ts
+++ b/clients/ts/signalr/spec/HubConnection.spec.ts
@@ -271,7 +271,7 @@ describe("HubConnection", () => {
             expect(warnings).toEqual(["No client method with the name 'message' found."]);
         });
 
-        it("all handlers can be unnregistered with just the method name", async () => {
+        it("all handlers can be unregistered with just the method name", async () => {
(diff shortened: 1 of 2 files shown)


clients/ts/signalr/spec/HubConnection.spec.ts
```diff
@@ -271,6 +271,72 @@ describe("HubConnection", () => {
...
+                invocationId: "0",
+                nonblocking: true,
+                target: "inc",
+                type: MessageType.Invocation,
+            });
+
+            expect(count).toBe(2);
+        });
+
+        it("a single handler can be unnregistered with the method name and handler", async () => {
```
BrennanConroy: unregistered

clients/ts/signalr/spec/HubConnection.spec.ts
```diff
@@ -271,6 +271,72 @@ describe("HubConnection", () => {
...
+
+            connection.receive({
+                arguments: [],
+                invocationId: "0",
+                nonblocking: true,
+                target: "inc",
+                type: MessageType.Invocation,
+            });
+
+            hubConnection.off("inc", handler);
```
BrennanConroy: This brings up an interesting question, if you call `off` with `handler` and have multiple `handler` registered, should it remove all instance of that callback?

mikaelm12: Hmm, I don't think so. Is that even a realistic scenario?

JamesNK: IMO we should follow `addEventListener`'s lead, which is ignore registering the same function multiple times.

https://stackoverflow.com/questions/26127336/javascript-event-addeventlistener-registering-multiple-times-for-same-function

mikaelm12: Okay. I think that makes sense but maybe should be a different issue. Thoughts?

JamesNK: It seems like it would be simple enough to do. Double check with Andrew or David that its behavior should change first.

anurse: I agree. It means scanning the list to see if any of the existing items `===` the new item when adding, but I think it's OK. Adding handlers is a startup task. Go ahead and put that in this PR.

clients/ts/signalr/spec/HubConnection.spec.ts
```diff
@@ -271,6 +271,72 @@ describe("HubConnection", () => {
             expect(warnings).toEqual(["No client method with the name 'message' found."]);
         });
 
+        it("all handlers can be unregistered with just the method name", async () => {
+            const warnings: string[] = [];
```
anurse: unused?

clients/ts/signalr/spec/HubConnection.spec.ts
```diff
@@ -271,6 +271,72 @@ describe("HubConnection", () => {
...
+                nonblocking: true,
+                target: "inc",
+                type: MessageType.Invocation,
+            });
+
+            expect(count).toBe(2);
+        });
+
+        it("a single handler can be unregistered with the method name and handler", async () => {
+            const warnings: string[] = [];
```
anurse: unused?

clients/ts/signalr/spec/HubConnection.spec.ts
mikaelm12: Copy pasta

clients/ts/signalr/spec/HubConnection.spec.ts
davidfowl: You guys are the masters of feature creep

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -316,11 +316,16 @@ export class HubConnection {
             this.methods[methodName] = [];
         }
 
-        this.methods[methodName].push(method);
+        // Preventing adding the same handler multiple times.
+        for (const method of this.methods[methodName]) {
+            if (method === newMethod) { return; }
```
BrennanConroy: nit: newline

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -316,11 +316,16 @@ export class HubConnection {
             this.methods[methodName] = [];
         }
 
-        this.methods[methodName].push(method);
+        // Preventing adding the same handler multiple times.
+        for (const method of this.methods[methodName]) {
```
anurse: You can use `indexOf`, it'll return `-1` if the item isn't in the list. So `if(this.methods.indexOf(newMethod) < 0) { return; }`

mikaelm12: I was unsure if indexOf used `===` but looks like it does 

>indexOf() compares searchElement to elements of the Array using strict equality (the same method used by the === or triple-equals operator).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

clients/ts/signalr/spec/HubConnection.spec.ts
mikaelm12: It always happens to me too 🙁

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -316,11 +316,14 @@ export class HubConnection {
             this.methods[methodName] = [];
         }
 
-        this.methods[methodName].push(method);
+        // Preventing adding the same handler multiple times.
+        if (this.methods[methodName].indexOf(newMethod) !== -1) { return; }
```
anurse: Put newlines in to match the others.

PR #1730, opened 2018-03-27
Add new SignalR scenarios

BrennanConroy:


benchmarks/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -10,23 +10,34 @@ namespace BenchmarkServer.Hubs
...
-                    await Clients.All.SendAsync("echo", DateTime.UtcNow);
+                    await Clients.All.SendAsync("send", DateTime.UtcNow);
+                    sent++;
                 }
             }
             catch (Exception e)
             {
                 Console.WriteLine(e);
             }
+            Console.WriteLine("Broadcast exited: Sent {0} messages", sent);
```
JamesNK: Since this is for performance, does `Console.WriteLine` have any impact? I guess it is out of the main loop.

benchmarks/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -10,23 +10,34 @@ namespace BenchmarkServer.Hubs
...
+            var sent = 0;
             try
             {
                 var t = new CancellationTokenSource();
                 t.CancelAfter(TimeSpan.FromSeconds(duration));
                 while (!t.IsCancellationRequested && !Context.ConnectionAborted.IsCancellationRequested)
                 {
-                    await Clients.All.SendAsync("echo", DateTime.UtcNow);
+                    await Clients.All.SendAsync("send", DateTime.UtcNow);
+                    sent++;
```
JamesNK: Do we worry about sending over int.MaxValue messages in stress tests? 🐴

BrennanConroy: Not doing stress tests 😛 
That variable is mostly a sanity check so I can make sure things are behaving as expected

benchmarks/BenchmarkServer/Hubs/EchoHub.cs
BrennanConroy: This is after the test has finished

benchmarks/BenchmarkServer/Hubs/EchoHub.cs
anurse: We will be doing stress tests at some point, but we can deal with that when we get there.

benchmarks/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -10,23 +10,34 @@ namespace BenchmarkServer.Hubs
...
 
-            Console.WriteLine("Echo exited");
+        public Task Echo(DateTime time)
+        {
+            return Clients.Client(Context.ConnectionId).SendAsync("send", time);
+        }
+
+        public Task EchoAll(DateTime time)
+        {
+            return Clients.All.SendAsync("send", time);
```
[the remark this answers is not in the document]
davidfowl: Meh I don't care enough to block this PR

PR #1735, opened 2018-03-27
Remove HubConnectionContext parameter from IHubProtocolResolver

### LINKED ISSUE #1724: Shall we allow DefaultHubProtocolResolver.GetProtocol accepts a parameter of HubConnnectionContext
I found [DefaultHubProtocolResolver:GetProtocol.GetProtocol](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubProtocolResolver.cs) refers to a unused parameter of HubConnnectionContext. It couples HubProtocolResolver and HubConnectionContext tightly. 

If the design here is to allow customized protocol for derived HubConnectionContext, it is better use another more concrete parameter instead of HubConnectionContext.

davidfowl:
Fixes #1724


anurse:
Wonder if we'll need `IConnectionContextAccessor` so people can get access to `ConnectionContext` in places where it isn't passed in 🙃

Issue #1726, opened 2018-03-26
Flaky tests: TransportThatFailsToStartOnceFallsBack


ryanbrandenburg:
[This test](http://aspnetci/viewLog.html?buildId=430868&tab=buildResultsDiv&buildTypeId=XPlat_Windows_Win7_Universe#testNameId4523811534513378872) started to fail recently:
```
System.InvalidOperationException : Unable to connect to the server with any of the available transports.
   at Microsoft.AspNetCore.Sockets.Client.HttpConnection.<StartAsyncInternal>d__46.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\modules\SignalR\src\Microsoft.AspNetCore.Sockets.Client.Http\HttpConnection.cs:line 290
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Sockets.Client.HttpConnection.<StartAsync>d__43.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\modules\SignalR\src\Microsoft.AspNetCore.Sockets.Client.Http\HttpConnection.cs:line 157
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.SignalR.Client.Tests.HttpConnectionTests.ConnectionLifecycle.<>c__DisplayClass5_1.<<TransportThatFailsToStartOnceFallsBack>b__1>d.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\modules\SignalR\test\Microsoft.AspNetCore.SignalR.Client.Tests\HttpConnectionTests.ConnectionLifecycle.cs:line 145
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1740, opened 2018-03-28
Check for Websockets to fix flaky tests

mikaelm12:
A couple of the fallback tests were flaky and upon further investigation I found that they were only failing on Win7 and Win2008 but they were failing consistently. 
We were getting 1 less start attempt than we expected which immediately tipped me off that it was a websockets availability issue.
The issue is that in certain scenarios we hit this code path https://github.com/aspnet/SignalR/blob/1fdd6511fadeafa0e77eac1c8f4958682af01560/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs#L246-L250

I added detection of WebSockets in the relevant tests which should take of of this.

Work tracked in issue: https://github.com/aspnet/SignalR/issues/1726
cc @ryanbrandenburg

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
@@ -147,8 +147,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
                         await connection.StartAsync(TransferFormat.Text);
                         if (!IsWebSocketsSupported())
                         {
-                            Assert.Equal(passThreshold - 1, startCounter);
-                            return;
+                            passThreshold -= 1;
                         }
                         Assert.Equal(passThreshold, startCounter);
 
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -142,7 +145,13 @@ Task OnTransportStart()
                     {
                         Assert.Equal(0, startCounter);
                         await connection.StartAsync(TransferFormat.Text);
+                        if (!IsWebSocketsSupported())
```
JamesNK: Is this a better or worse solution than having separate tests that are or are not conditional on web sockets? Thoughts @anurse @davidfowl ?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -153,6 +162,7 @@ Task OnTransportStart()
                 using (StartLog(out var loggerFactory))
                 {
                     var startCounter = 0;
+                    var availableTransports = 3;
```
anurse: Use `IsWebSocketsSupported` here to set the correct number rather than doing it down in the assert.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
anurse: I think it's fine this way. We don't want piles of skipped tests (they create some noise).

@mikaelm12 can you do this by doing `if(!IsWebSocketsSupported()) { passThreshold -= 1; }` and then just having a single Assert instead though?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs
```diff
@@ -168,11 +178,38 @@ Task OnTransportStart()
...
+                            {
+                                Assert.Equal(availableTransports - 1, startCounter);
+                                return;
+                            }
+                            Assert.Equal(availableTransports, startCounter);
                         });
                 }
             }
 
+            private static bool IsWebSocketsSupported()
```
anurse: Can we use the code from the condition attribute @JamesNK added? Put it in a helper method or something?

PR #1756, opened 2018-03-28
Always flush in WebSocketsTransport

BrennanConroy:


JamesNK:
Is this fixing an existing broken test? If not can you add a test for the scenario it is fixing.

davidfowl:
> Is this fixing an existing broken test? If not can you add a test for the scenario it is fixing.

No tests, it's a performance thing.

Issue #1614, opened 2018-03-16
Suppress detailed exception messages from the protocol messages by default


anurse:
Right now, we produce error messages in the protocol (`Completion`, `Close`, `HandshakeResponse`) containing exception messages. This need to be disabled by default as in SignalR for ASP.NET.

anurse:
Putting this in Preview 2 so we see it and look at it. It can probably move to the RC milestone, but it will need to happen by then at least.

anurse:
The big work item here is to find all the exceptions and check the flag to determine if detailed messages should be emitted.

PR #1763, opened 2018-03-29
Add detailed error option

BrennanConroy:
#1614

Opening now to solicit early feedback

Need to fill in the exception strings

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -40,6 +40,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
                 serviceScopeFactory,
                 new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>(NullLogger<DefaultHubLifetimeManager<TestHub>>.Instance)),
                 Options.Create(new HubOptions<TestHub>()),
+                Options.Create(new HubOptions()),
                 new Logger<DefaultHubDispatcher<TestHub>>(NullLoggerFactory.Instance));
 
             var pair = DuplexPipe.CreateConnectionPair(PipeOptions.Default, PipeOptions.Default);
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
             _dispatcher = new DefaultHubDispatcher<TestHub>(
                 serviceScopeFactory,
                 new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>(NullLogger<DefaultHubLifetimeManager<TestHub>>.Instance)),
-                new OptionsWrapper<HubOptions<TestHub>>(new HubOptions<TestHub>()),
+                Options.Create(new HubOptions<TestHub>()),
                 new Logger<DefaultHubDispatcher<TestHub>>(NullLoggerFactory.Instance));
 
             var pair = DuplexPipe.CreateConnectionPair(PipeOptions.Default, PipeOptions.Default);
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -17,6 +17,7 @@ using Microsoft.AspNetCore.Sockets;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
 {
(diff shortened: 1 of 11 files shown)


benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
```diff
@@ -38,6 +39,7 @@ public void GlobalSetup()
             _dispatcher = new DefaultHubDispatcher<TestHub>(
                 serviceScopeFactory,
                 new HubContext<TestHub>(new DefaultHubLifetimeManager<TestHub>(NullLogger<DefaultHubLifetimeManager<TestHub>>.Instance)),
+                new OptionsWrapper<HubOptions<TestHub>>(new HubOptions<TestHub>()),
```
davidfowl: Options.Create(...)

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -172,7 +175,16 @@ public override IReadOnlyList<Type> GetParameterTypes(string methodName)
                 if (hubMethodInvocationMessage.ArgumentBindingException != null)
                 {
                     Log.FailedInvokingHubMethod(_logger, hubMethodInvocationMessage.Target, hubMethodInvocationMessage.ArgumentBindingException);
-                    await SendInvocationError(hubMethodInvocationMessage, connection, $"Failed to invoke '{hubMethodInvocationMessage.Target}'. {hubMethodInvocationMessage.ArgumentBindingException.Message}");
+                    var errorMessage = string.Empty;
+                    if (!_hubOptions.EnableDetailedErrors)
+                    {
+                        errorMessage = "some error here";
```
anurse: I think the pattern of just cutting off the detailed message is sufficient. I do think both messages need a little work though so:

* DetailedErrors off: "Failed to invoke '{Target}' due to an error on the server."
* DetailedErrors on: "Failed to invoke '{Target}' due to an error on the server: {Details}"

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -225,6 +237,11 @@ public override IReadOnlyList<Type> GetParameterTypes(string methodName)
 
         private string BuildUnexpectedErrorMessage(string methodName, Exception exception)
         {
+            if (!_hubOptions.EnableDetailedErrors)
+            {
+                return "write an error";
```
anurse: "An unexpected error occurred invoking '{methodName}' on the server."

The method name isn't secret, the client already knew it.

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -256,6 +273,10 @@ private string BuildUnexpectedErrorMessage(string methodName, Exception exceptio
             }
             finally
             {
+                if (error != string.Empty && !_hubOptions.EnableDetailedErrors)
+                {
+                    error = "todo, put some generic error here";
```
anurse: "An error occurred on the server while streaming results"

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -326,7 +327,16 @@ public virtual void Abort()
             catch (Exception ex)
             {
                 Log.HandshakeFailed(_logger, ex);
-                await WriteHandshakeResponseAsync(new HandshakeResponseMessage($"An unexpected error occurred during connection handshake. {ex.GetType().Name}: {ex.Message}"));
+                var errorMessage = string.Empty;
+                if (enableDetailedErrors)
+                {
+                    errorMessage = $"An unexpected error occurred during connection handshake. {ex.GetType().Name}: {ex.Message}";
```
JamesNK: I'd like a helper method. Something like:

```
public string BuildErrorMessage(string message, Exception exception, bool includeExceptionDetails)
{
  if (!includeExceptionDetails) return message;

  return message += $" {exception.GetType().Name}: {exception.Message}";
}
```

We avoid duplicating the error messages throughout code and share logic around how exceptions are included with errors to the client.

anurse: I was kinda thinking something similar. If we ever get around to localization, it won't quite work that way because we can't reliably know how to concatenate these messages but that's a ways off :)

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -139,9 +139,14 @@ public class HubConnectionHandler<THub> : ConnectionHandler where THub : Hub
...
         private async Task SendCloseAsync(HubConnectionContext connection, Exception exception)
         {
-            CloseMessage closeMessage = exception == null
-                ? CloseMessage.Empty
-                : new CloseMessage($"Connection closed with an error. {exception.GetType().Name}: {exception.Message}");
+            var closeMessage = CloseMessage.Empty;
+
+            if (exception != null)
+            {
+                closeMessage = _enableDetailedErrors
```
anurse: Can't use the helper here?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -581,7 +581,7 @@ private Func<IConnection> GetHttpConnectionFactory(ILoggerFactory loggerFactory,
 
                     var channel = await connection.StreamAsChannelAsync<int>("Stream", "xyz");
                     var ex = await Assert.ThrowsAsync<HubException>(() => channel.ReadAllAsync().OrTimeout());
-                    Assert.StartsWith("Failed to invoke 'Stream'. Error binding arguments. Make sure that the types of the provided values match the types of the hub method being invoked.", ex.Message);
+                    Assert.StartsWith("Failed to invoke 'Stream' due to an error on the server.", ex.Message);
```
anurse: Can this be an `Assert.Equals` now?

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -1481,10 +1494,16 @@ public void HubsCannotHaveOverloadedMethods()
...
-        [Fact]
-        public async Task ReceiveCorrectErrorFromStreamThrowing()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task ReceiveCorrectErrorFromStreamThrowing(bool detailedErrors)
         {
-            var serviceProvider = HubConnectionHandlerTestUtils.CreateServiceProvider();
+            var serviceProvider = HubConnectionHandlerTestUtils.CreateServiceProvider(builder =>
+            builder.AddSignalR(options =>
```
anurse: Indent this one more?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
BrennanConroy: Just missed it

Issue #1629, opened 2018-03-16
Flaky tests: TransportThatFallsbackCreatesNewConnection et al


ryanbrandenburg:
[This](http://aspnetci/viewLog.html?buildId=423387&buildTypeId=XPlat_Windows_Win2008r2_Universe) test and others sometimes fails with a message like:
```
System.TimeoutException : Operation in TransportThatFallsbackCreatesNewConnection timed out at C:\BuildAgent\work\33bdfc1cae7b2a38\modules\SignalR\test\Microsoft.AspNetCore.SignalR.Tests\EndToEndTests.cs:69
   at System.Threading.Tasks.TaskExtensions.OrTimeout(Task task, TimeSpan timeout, String memberName, String filePath, Nullable`1 lineNumber)
   at Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.TransportThatFallsbackCreatesNewConnection() in C:\BuildAgent\work\33bdfc1cae7b2a38\modules\SignalR\test\Microsoft.AspNetCore.SignalR.Tests\EndToEndTests.cs:line 69
--- End of stack trace from previous location where exception was thrown ---
```

This particular subset of failures is being tracked in a larger list [here](http://aspnetci/httpAuth/app/rest/builds?locator=tag:timeout,sinceDate:201710010T000000-0001&fields=build(startDate,buildTypeId,status,branchName,webUrl,tags)).

ryanbrandenburg:
[This](http://aspnetci/project.html?tab=testDetails&testNameId=-1869633753029778556&projectId=XPlat&buildTypeId=XPlat_HighSierraOsxUniverse) fails more often than it passes on High Sierra.

BrennanConroy:
@mikaelm12

davidfowl:
I saw this locally as well.

mikaelm12:
This change should fix this: 81b0f60b5ae7a3efa2caab91a655d0099b4f2da2
Closing for now. We can reopen if the problem persists.

PR #1772, opened 2018-03-29
Check for Websockets in Fallback functional tests

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1629

This should fix the failures in this this functional test.

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
@@ -375,7 +375,7 @@ namespace Microsoft.AspNetCore.SignalR.Tests
 
             public FakeTransport()
             {
-                if (!IsWebSocketsSupported())
+                if (!TestHelpers.IsWebSocketsSupported())
                 {
                     availableTransports -= 1;
                 }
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -404,6 +414,23 @@ public Task StopAsync()
             }
         }
 
+        private static bool IsWebSocketsSupported()
```
JamesNK: Lets move this to `Microsoft.AspNetCore.SignalR.Tests.Utils` and reuse it everywhere (including that conditional test attribute)

mikaelm12: Yeah, probably about time to move this

JamesNK: In fact I can see a `TestHelpers.IsWebSocketsSupported()` method already exists in that project. Check with @anurse to see whether this logic is better than what is already in that method and possibly replace it.

mikaelm12: Yeah, I wen't ahead and replaced the logic. I can reverted it though. Thoughts? @davidfowl  @anurse

BrennanConroy: Remove

anurse: I think the new logic is good. This copy isn't needed though right?

PR #1773, opened 2018-03-29
Client logging fixes and improvements

JamesNK:


clients/ts/signalr/src/Transports.ts
```diff
@@ -313,17 +314,17 @@ export class LongPollingTransport implements ITransport {
...
-function getDataLength(data: any): number {
-    let length: number = null;
+function getDataDetail(data: any): string {
+    let length: string = null;
     if (data instanceof ArrayBuffer) {
-        length = data.byteLength;
-    } else if (data instanceof String) {
-        length = data.length;
+        length = `Binary data of length ${data.byteLength}`;
+    } else if (typeof data === "string") {
```
BrennanConroy: Whats wrong with `instanceof String`?

clients/ts/FunctionalTests/ts/TestLogger.ts
```diff
@@ -1,12 +1,17 @@
...
+import { ConsoleLogger, ILogger, LogLevel } from "@aspnet/signalr";
 
 export class TestLogger implements ILogger {
     public static instance: TestLogger = new TestLogger();
+    private static consoleLogger: ConsoleLogger = new ConsoleLogger(LogLevel.Trace);
 
     public messages: Array<[LogLevel, string]> = [];
 
     public log(logLevel: LogLevel, message: string): void {
+        // Capture log message so it can reported later
```
BrennanConroy: can **be**

clients/ts/signalr/src/Transports.ts
JamesNK: `instanceof String` will match `new String("Hello World")` 🤷‍♂️ 

https://stackoverflow.com/questions/16792051/how-to-instanceof-a-primitive-string-string-literal-in-javascript

JamesNK:
🆙 📅 

Added extra logging throughout TS client.

clients/ts/FunctionalTests/ts/TestLogger.ts
```diff
@@ -1,12 +1,17 @@
...
+    private static consoleLogger: ConsoleLogger = new ConsoleLogger(LogLevel.Trace);
 
     public messages: Array<[LogLevel, string]> = [];
 
     public log(logLevel: LogLevel, message: string): void {
+        // Capture log message so it can be reported later
         this.messages.push([logLevel, message]);
+
+        // Also write to browser console
+        TestLogger.consoleLogger.log(logLevel, message);
```
anurse: Nice, thanks for catching that too!

clients/ts/signalr/src/Transports.ts
anurse: Yeah, `instanceof String` **only** matches `new String(...)`, not literals. So, `typeof` is the idomatic way to do this here

anurse: image

anurse: JavaScript is weird yo.

PR #1774, opened 2018-03-29
Fix internal HubConnection state when handshake fails

BrennanConroy:


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -132,24 +132,25 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
                 // Start the connection
                 var connection = _connectionFactory();
                 await connection.StartAsync(_protocol.TransferFormat);
-                _connectionState = new ConnectionState(connection, this);
+                var connectionState = new ConnectionState(connection, this);
```
JamesNK: Maybe call this `startingConnectionState` to make it more obvious that it isn't ready to use yet?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -132,24 +132,25 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
...
                 {
                     Log.ErrorStartingConnection(_logger, ex);
 
                     // Can't have any invocations to cancel, we're in the lock.
-                    await _connectionState.Connection.DisposeAsync();
+                    await connectionState.Connection.DisposeAsync();
                     throw;
                 }
 
+                _connectionState = connectionState;
```
JamesNK: Add a comment why we wait until here before setting `_connectionState`.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -494,15 +495,15 @@ private void CheckDisposed()
             }
         }
 
-        private async Task HandshakeAsync()
+        private async Task HandshakeAsync(ConnectionState connectionState)
```
JamesNK: Also rename to startingConnectionState?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -132,24 +132,26 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
                 // Start the connection
                 var connection = _connectionFactory();
                 await connection.StartAsync(_protocol.TransferFormat);
-                _connectionState = new ConnectionState(connection, this);
+                var startingConnectionState = new ConnectionState(connection, this);
```
davidfowl: nit: initialConnectionState?

Issue #1485, opened 2018-02-22
Flaky test: EndToEndTests.ConnectionCanSendAndReceiveMessages


ryanbrandenburg:
This test has a [poor pass rate](http://aspnetci/project.html?tab=testDetails&testNameId=-192649964883647388&projectId=XPlat&buildTypeId=XPlat_Windows_Win2008r2_Universe) on Win2008 universe. It's unclear exactly what the problem is because the logging isn't informative (despite the fact that Log.Error takes an exception parameter, it doesn't use it by default). If instances like [this](http://aspnetci/viewLog.html?buildId=411473&tab=buildResultsDiv&buildTypeId=XPlat_Windows_Win2008r2_Universe#testNameId-192649964883647388) told us exactly what error occurred while sending we would know more.

CC @muratg for a flaky test.

An example failing test:
```
System.Net.WebException : The request was aborted: The request was canceled.
   at System.Net.ConnectStream.BeginRead(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
   at System.IO.Stream.<>c.<BeginEndReadAsync>b__43_0(Stream stream, ReadWriteParameters args, AsyncCallback callback, Object state)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncTrim[TInstance,TArgs](TInstance thisRef, TArgs args, Func`5 beginMethod, Func`3 endMethod)
   at System.IO.Stream.BeginEndReadAsync(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Stream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
   at System.IO.Stream.<CopyToAsyncInternal>d__27.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Net.Http.HttpContent.<LoadIntoBufferAsyncCore>d__49.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1778, opened 2018-03-30
Re-enable ConnectionCanSendAndReceiveMessages + logging

JamesNK:
https://github.com/aspnet/SignalR/issues/1485

Testing on CI

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -348,7 +348,7 @@ namespace Microsoft.AspNetCore.Sockets.Client.Http
             }
             catch (Exception ex)
             {
-                Log.ErrorStartingTransport(_logger, _transport, ex);
+                Log.ErrorStartingTransport(_logger, transport, ex);
                 _transport = null;
                 throw;
             }
@@ -356,6 +356,8 @@ namespace Microsoft.AspNetCore.Sockets.Client.Http
             // We successfully started, set the transport properties (we don't want to set these until the transport is definitely running).
             _transport = transport;
             _transportPipe = pair.Transport;
+
+            Log.TransportStarted(_logger, transport);
         }
 
         private HttpClient CreateHttpClient()
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
@@ -357,7 +357,7 @@ namespace Microsoft.AspNetCore.Sockets.Client.Http
             _transport = transport;
             _transportPipe = pair.Transport;
 
-            Log.TransportStarted(_logger, transport);
+            Log.TransportStarted(_logger, _transport);
         }
 
         private HttpClient CreateHttpClient()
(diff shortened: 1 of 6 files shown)


JamesNK:
Even in a loop of 1000 I can't replicate this locally, and a loop of 100 on the server. There is a corefx sockets bug that is sometimes coming up when running the test so many times but it isn't related to signalr.

I added some more logging while looking at this and made some minor fixes that I've left in.

src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs
```diff
@@ -348,14 +348,16 @@ private static Uri CreateConnectUrl(Uri url, string connectionId)
             }
             catch (Exception ex)
             {
-                Log.ErrorStartingTransport(_logger, _transport, ex);
+                Log.ErrorStartingTransport(_logger, transport, ex);
```
BrennanConroy: Nice catch

test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs
```diff
@@ -196,11 +196,11 @@ public EndToEndTests(ServerFixture<Startup> serverFixture, ITestOutputHelper out
             }
         }
 
-        [Theory(Skip = "https://github.com/aspnet/SignalR/issues/1485")]
+        [Theory]
         [MemberData(nameof(TransportTypesAndTransferFormats))]
         public async Task ConnectionCanSendAndReceiveMessages(TransportType transportType, TransferFormat requestedTransferFormat)
         {
-            using (StartLog(out var loggerFactory, testName: $"ConnectionCanSendAndReceiveMessages_{transportType.ToString()}"))
+            using (StartLog(out var loggerFactory, minLogLevel: LogLevel.Trace, testName: $"ConnectionCanSendAndReceiveMessages_{transportType.ToString()}_{requestedTransferFormat.ToString()}"))
```
BrennanConroy: :+1:

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
```diff
@@ -93,7 +93,17 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
 
             _application.Input.CancelPendingRead();
 
-            await Running;
+            try
+            {
+                await Running;
```
davidfowl: This should never throw. I would remove the try catch.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
```diff
@@ -200,7 +200,17 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
 
             _application.Input.CancelPendingRead();
 
-            await Running;
+            try
+            {
+                await Running;
```
davidfowl: This should never throw. I would remove the try catch.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/WebSocketsTransport.cs
```diff
@@ -348,10 +348,14 @@ private static Uri ResolveWebSocketsUrl(Uri url)
             {
                 await Running;
             }
-            catch
+            catch (Exception ex)
             {
+                Log.TransportStopped(_logger, ex);
                 // exceptions have been handled in the Running task continuation by closing the channel with the exception
```
davidfowl: We should make this consistent with the rest and make it never throw (tests would need to be updated)

PR #1779, opened 2018-03-30
Make shutdown logic for SSE and LongPolling more similar to WebSockets

davidfowl:
- Cancel reading from the application when initiating a transport stop
- Complete each side of the pipe in the place where the pipe is being consumed
- Errors from sending end up getting sent to the application
- The Running task never throws
- Removes ContinueWith
- Cancel the token from the sending side only
- Call CancelPendingFlush appropriately 
- Observe flush result

PS: Cancellation tokens still suck 😄

### DIFF
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
@@ -21,6 +21,7 @@ namespace Microsoft.AspNetCore.Sockets.Client.Internal
         private readonly HttpOptions _httpOptions;
         private readonly ILogger _logger;
         private IDuplexPipe _application;
+        // Volatile so that the poll loop sees the updated value set from a different thread
         private volatile Exception _error;
 
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
--- a/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
@@ -21,8 +21,7 @@ namespace Microsoft.AspNetCore.Sockets.Client.Internal
         private readonly HttpOptions _httpOptions;
         private readonly ILogger _logger;
         private IDuplexPipe _application;
-        private Task _sender;
-        private Task _poller;
+        private volatile Exception _error;
 
         private readonly CancellationTokenSource _transportCts = new CancellationTokenSource();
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
```diff
@@ -92,6 +114,7 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
                 return;
             }
 
+            using (response)
```
JamesNK: While the response is disposed if we reach here, an error thrown above at `response.EnsureSuccessStatusCode();` won't dispose the response. Connection could be left hanging because of `HttpCompletionOption.ResponseHeadersRead`.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
```diff
@@ -131,6 +157,13 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
                                     }
                                     break;
                             }
+
+                            // We canceled in the middle of applying back pressure
+                            // or if the consumer is done
+                            if (flushResult.IsCanceled || flushResult.IsCompleted)
+                            {
+                                break;
```
JamesNK: Add `Log.EventStreamEnded(_logger);` ?

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/LongPollingTransport.cs
```diff
@@ -52,40 +51,54 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
...
+            // Start sending and polling (ask for binary if the server supports it)
+            var receiving = Poll(url, _transportCts.Token);
+            var sending = SendUtils.SendMessages(url, _application, _httpClient, _httpOptions, _logger);
 
-            _transportCts.Cancel();
+            // Wait for send or receive to complete
+            var trigger = await Task.WhenAny(receiving, sending);
 
-            try
+            if (trigger == receiving)
```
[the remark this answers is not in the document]
davidfowl: We already log in each of the loops, this main loop is just doing coordination. If you were to look at the logs, the order would be obvious.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
davidfowl: good catch, I blame @mikaelm12. Fixing it.

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
davidfowl: No, the application stopped reading 😄 . I don't think we need a log for this.

davidfowl:
> In the interest of time, approved to merge. but would like @anurse to check it over later since he knows a lot more about the client than I 😸

He's going to *love* these changes 😄

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
```diff
@@ -116,12 +143,15 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
                         {
                             Log.ParsingSSE(_logger, input.Length);
                             var parseResult = _parser.ParseMessage(input, out consumed, out examined, out var buffer);
+                            FlushResult flushResult = default;
```
anurse: Not something I want to change now, but I had preferred `var flushResult = default(FlushResult)` in this situation, because it's more consistent with our "var-everywhere" style. Is this your preferred way to do this?

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
anurse: > I don't think we need a log for this.

Famous last words :trollface:

src/Microsoft.AspNetCore.Sockets.Client.Http/Internal/ServerSentEventsTransport.cs
davidfowl: I actually prefer the style you mention but you get more @pranavkm points when you use newer language features.

anurse: And something tracking Pranav Points

PR #1780, opened 2018-03-30
Dispose the http client in DisposeAsync

davidfowl:
- This layer of the stack is no longer reusable and this disposable was removed (I assume by mistake) in the big refactoring


anurse:
Yep, just a mistake. Good catch.

PR #1791, opened 2018-03-30
Use IBufferWriter in IHubProtocol

BrennanConroy:


src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextWriter.cs
```diff
@@ -77,12 +76,12 @@ public void SetWriter(IBufferWriter<byte> bufferWriter)
...
         public override void Write(char[] buffer, int index, int count)
         {
-            WriteInternal(buffer, index, count);
+            WriteInternal(buffer.AsSpan(index, count));
         }
 
         public override void Write(char[] buffer)
         {
-            WriteInternal(buffer, 0, buffer.Length);
+            WriteInternal(buffer.AsSpan());
```
davidfowl: You don't need AsSpan here

BrennanConroy:
For 1 million message writes on server
Before:
20.6 M allocations and 5.95 GB

After
18.54 M allocations and 2.94 GB

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextWriter.cs
```diff
@@ -120,6 +119,11 @@ public override void Write(char value)
             }
         }
 
+        public override void Write(string value)
+        {
+            WriteInternal(value.AsSpan());
```
JamesNK: ahhh, good add

BrennanConroy:
After
```
             Method |          Input | HubProtocol |      Mean |     Error |    StdDev |      Op/s |  Gen 0 |  Allocated|
------------------- |--------------- |------------ |----------:|----------:|----------:|----------:|-------:|----------:|
 WriteSingleMessage |   FewArguments |        Json |  3.052 us | 0.0597 us | 0.0998 us | 327,648.9 | 0.0305 |    2.99 KB|
 WriteSingleMessage |   FewArguments |     MsgPack |  2.006 us | 0.0356 us | 0.0500 us | 498,418.3 | 0.0267 |    2.74 KB|
 WriteSingleMessage | LargeArguments |        Json | 63.913 us | 1.2450 us | 1.5290 us |  15,646.3 | 0.4883 |   49.08 KB|
 WriteSingleMessage | LargeArguments |     MsgPack | 21.718 us | 0.3984 us | 0.7581 us |  46,045.4 | 0.9155 |   83.34 KB|
 WriteSingleMessage |  ManyArguments |        Json |  5.391 us | 0.1025 us | 0.1221 us | 185,487.1 | 0.0381 |    4.22 KB|
 WriteSingleMessage |  ManyArguments |     MsgPack |  3.044 us | 0.0632 us | 0.0649 us | 328,492.9 | 0.0343 |    3.01 KB|
 WriteSingleMessage |    NoArguments |        Json |  1.661 us | 0.0322 us | 0.0252 us | 602,123.9 | 0.0248 |    2.57 KB|
 WriteSingleMessage |    NoArguments |     MsgPack |  1.055 us | 0.0226 us | 0.0472 us | 947,488.9 | 0.0267 |     2.5 KB|
```

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextWriter.cs
```diff
@@ -77,12 +76,12 @@ public void SetWriter(IBufferWriter<byte> bufferWriter)
 
         public override void Write(char[] buffer, int index, int count)
         {
-            WriteInternal(buffer, index, count);
+            WriteInternal(buffer.AsSpan(index, count));
```
JamesNK: Is this going to cause some perf hit on environments where span isn't fully optimized? @davidfowl

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HubProtocolExtensions.cs
```diff
@@ -1,19 +1,15 @@
...
     public static class HubProtocolExtensions
     {
         public static byte[] WriteToArray(this IHubProtocol hubProtocol, HubMessage message)
         {
-            using (var ms = new LimitArrayPoolWriteStream())
-            {
-                hubProtocol.WriteMessage(message, ms);
-                return ms.ToArray();
-            }
+            var writer = new MemoryBufferWriter();
```
JamesNK: MemoryBufferWriter needs some perf thought put into, along with unit tests if it is going to be used everywhere. I wrote it pretty quickly to get unit tests passing

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextWriter.cs
anurse: We were talking about this just recently ;). We `#ifdef` and get a pointer from the span which mostly resolves any perf hit on slower platforms.

davidfowl: Creating the span is slightly slower but we do pin it to avoid some of the costs.

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/Utf8BufferTextWriter.cs
davidfowl: Showed up on the profile.

JamesNK:
More thought needs to go into the array buffer size and using pooling. Right now the minimum `MemoryBufferWriter` will allocate is 2KB which you can see show up in the msgpack with no arguments result, going from 0.4KB to 2.5KB.

davidfowl:
> More thought needs to go into the array buffer size and using pooling. Right now the minimum MemoryBufferWriter will allocate is 2KB which you can see show up in the msgpack with no arguments result, going from 0.4KB to 2.5KB.

Ya, we're fixing it now.

BrennanConroy:
```
             Method |          Input | HubProtocol |        Mean |       Error |      StdDev |        Op/s |  Gen 0 | Allocated |
------------------- |--------------- |------------ |------------:|------------:|------------:|------------:|-------:|----------:|
 WriteSingleMessage |   FewArguments |        Json |  2,922.8 ns |    71.32 ns |    66.72 ns |   342,132.5 | 0.0076 |     960 B |
 WriteSingleMessage |   FewArguments |     MsgPack |  1,595.6 ns |    30.26 ns |    33.63 ns |   626,729.0 | 0.0057 |     704 B |
 WriteSingleMessage | LargeArguments |        Json | 62,303.8 ns | 1,137.93 ns | 1,064.42 ns |    16,050.4 | 0.2441 |   25168 B |
 WriteSingleMessage | LargeArguments |     MsgPack | 19,543.8 ns |   382.93 ns |   639.79 ns |    51,167.0 | 0.5798 |   62320 B |
 WriteSingleMessage |  ManyArguments |        Json |  5,280.3 ns |   103.21 ns |   110.44 ns |   189,382.9 | 0.0153 |    2216 B |
 WriteSingleMessage |  ManyArguments |     MsgPack |  2,919.4 ns |    57.71 ns |    88.13 ns |   342,540.8 | 0.0076 |     976 B |
 WriteSingleMessage |    NoArguments |        Json |  1,477.0 ns |    31.92 ns |    48.74 ns |   677,038.5 | 0.0038 |     528 B |
 WriteSingleMessage |    NoArguments |     MsgPack |    850.8 ns |    15.41 ns |    14.41 ns | 1,175,297.9 | 0.0038 |     456 B |
```

src/Microsoft.AspNetCore.SignalR.Common/Internal/MemoryBufferWriter.cs
```diff
@@ -33,8 +33,7 @@ public Memory<byte> GetMemory(int sizeHint = 0)
 
             if (Segments.Count == 0 || Position == _segmentSize)
             {
-                // TODO: Rent memory from a pool
-                Segments.Add(new Memory<byte>(new byte[_segmentSize]));
+                Segments.Add(ArrayPool<byte>.Shared.Rent(_segmentSize));
```
JamesNK: What are the pros and cons of ArrayPool vs MemoryPool? https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Buffers/MemoryPool.cs

I've never used MemoryPool myself.

JamesNK:
I think some dotTracing to find out why performance hasn't improved despite lower memory usage would be useful. Especially for few/no arguments.

BrennanConroy:
Before, After for 1 million messages
capture

PR #1805, opened 2018-03-31
Added TCP client sample to HubSample

davidfowl:
- Ripped of Kestrel's SocketConnection to make a TcpConnection
IConnection implementation.
- Fixed issue with SignalR assuming there will always be a non-null user
on the ConnectionContext.

/cc @KPixel 

/cc @halter73 The connection object is usable with a single change (just removing the dispatcher). I think the unification is closer than we think 😄

### DIFF
--- a/samples/SignalRSamples/Program.cs
+++ b/samples/SignalRSamples/Program.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
+using System.Net;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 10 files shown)


samples/SignalRSamples/Program.cs
```diff
@@ -31,9 +32,8 @@ public static void Main(string[] args)
                     options.ListenLocalhost(5000);
 
                     // Hub bound to TCP end point
-                    options.ListenLocalhost(9001, builder =>
```
davidfowl: https://github.com/aspnet/KestrelHttpServer/pull/2447

PR #1806, opened 2018-03-31
Change how HttpConnection sets headers

davidfowl:
- Removed SendUtils.PrepareRequst and instead used DefaultRequestHeaders to apply HttpOptions to every outgoing HttpRequestMessage
- Modified how we check for the user agent request in tests


src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -396,6 +395,10 @@ private HttpClient CreateHttpClient()
                 }
             }
 
+            // Add the HttpOptions handler, this will apply any modifications to the http request
+            // (like applying headers, or the access token and adding the user agent).
+            httpMessageHandler = new HttpOptionsMessageHandler(httpMessageHandler, _httpOptions);
```
JamesNK: Only wrap the handler if `_httpOptions` isn't null? Then you can remove null options logic from the handler.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/HttpOptionsMessageHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+        private readonly HttpOptions _httpOptions;
+
+        public HttpOptionsMessageHandler(HttpMessageHandler inner, HttpOptions options) : base(inner)
+        {
+            _httpOptions = options;
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (_httpOptions?.Headers != null)
```
JamesNK: I'd like a comment explaining why these headers are set in a message handler instead of on the `HttpRequestMessage` prior to using `HttpClient`.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
davidfowl: Sounds good.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/HttpOptionsMessageHandler.cs
davidfowl: I don't understand. Are you asking for a comment that says we're using a message handler so we don't have to manually set common headers for every request?

Actually thinking about it now, I wonder if it makes sense to set DefaultHeaders on the http client itself

JamesNK: I asked for a comment because I was curious why we were changing from the `PrepareRequst` to using a handler. Is it just a refactor or does it fix an issue? If it fixes an issue then it would be good to say why we're doing something in such a way.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
JamesNK: Although you always want to set `request.Headers.UserAgent.Add(Constants.UserAgentHeader);` even if options are null.

I think you should rename the handler name if it isn't just for setting `HttpOptions`.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/HttpOptionsMessageHandler.cs
davidfowl: Oh, this is just a better way to execute code on *all* requests. We don't have to sprinkle code around the code base. Just like you did with the LoggingMessageHandler.

PR #1816, opened 2018-04-01
Clean up the SSE client side transport

davidfowl:
- Renamed StreamPipeConnection to PipeReaderFactory
- Flow the transport cancellation token to the CopyToAsync routine ~and use it in ReadAsync (as much as it pains me)~
- Use the PipeWriterStream from src/Common
- Other small cleanup and nits to make the style consistent with the other pipe reader loops


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/ServerSentEventsTransport.Log.cs
```diff
@@ -29,7 +29,7 @@ private static class Log
             private static readonly Action<ILogger, Exception> _transportStopping =
                 LoggerMessage.Define(LogLevel.Information, new EventId(6, "TransportStopping"), "Transport is stopping.");
 
-            private static readonly Action<ILogger, int, Exception> _messageToApp =
+            private static readonly Action<ILogger, int, Exception> _messageToApplication =
                 LoggerMessage.Define<int>(LogLevel.Debug, new EventId(7, "MessageToApp"), "Passing message to application. Payload size: {Count}.");
```
JamesNK: "MessageToApp" -> "MessageToApplication"

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/ServerSentEventsTransport.cs
```diff
@@ -116,57 +116,59 @@ public Task StartAsync(Uri url, IDuplexPipe application, TransferFormat transfer
...
+
+                                var parseResult = _parser.ParseMessage(buffer, out consumed, out examined, out var message);
+                                FlushResult flushResult = default;
+
+                                switch (parseResult)
+                                {
+                                    case ServerSentEventsMessageParser.ParseResult.Completed:
+                                        Log.MessageToApplication(_logger, message.Length);
+
+                                        flushResult = await _application.Output.WriteAsync(message);
```
JamesNK: So not using the cancellation token here?

davidfowl: No. We use CancelPendingFlush

davidfowl: I'd actually prefer to avoid the token in all cases but if the Stream ignores the cancellation token we don't want to hang (I dunno if that happens with all implementations of http client).

davidfowl: I'm going to make a slight tweak here to get rid of the CT usage in the pipe itself.

davidfowl: Ok did a bit more cleanup here.

davidfowl:
🆙 📅

src/Common/PipeWriterStream.cs
```diff
@@ -69,6 +69,11 @@ public override ValueTask WriteAsync(ReadOnlyMemory<byte> source, CancellationTo
 
         private ValueTask WriteCoreAsync(ReadOnlyMemory<byte> source, CancellationToken cancellationToken = default)
         {
+            if (cancellationToken.CanBeCanceled)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
```
JamesNK: Shouldn't this return a cancelled `ValueTask` instead of throw because there is no `async` keyword on the method?

davidfowl: Why?

davidfowl: @stephentoub are there any hard and fast rules here?

stephentoub: The guidance is for the cancellation exception to come via the returned task rather than propagating out synchronously.  That generally goes for any exception other than usage exceptions like ArgumentException.  It doesn't matter at all if you're just awaiting, e.g.
```C#
await WriteAsync(...);
```
as then the caller can't tell the difference, but in other situations it does matter, such as if you're using combinators, using ContinueWith or some other method that operates on the returned task, storing the task and using it for later, etc.

davidfowl: So basically don't use cancellationToken.ThrowIfCancellationRequested() unless the method is async.

PR #1819, opened 2018-04-01
Clean up the logging style for HttpConnectionManager

davidfowl:


src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.Log.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Microsoft.AspNetCore.Http.Connections
+{
+    public partial class HttpConnectionManager
+    {
+        private static class Log
+        {
+            // Category: ConnectionManager
+            private static readonly Action<ILogger, string, Exception> _createdNewConnection =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(1, nameof(CreatedNewConnection)), "New connection {TransportConnectionId} created.");
```
JamesNK: I have been wondering why `nameof` is not used for the event names. Why here and not other places? I would prefer one standard way of doing it, my preference being `nameof` unless there is a reason I don't know about.

davidfowl: File an issue

davidfowl: Seems like we're pretty inconsistent.

davidfowl: Actually, maybe these should be strings instead of being nameof, @anurse do these names have to be stable across releases? i.e. if the name of the method changes, we probably don't want the event name to change.

davidfowl: I'll undo the nameof here

Issue #1649, opened 2018-03-19
Non-serializable data


ryanbrandenburg:
[These](http://aspnetci/viewLog.html?buildId=423407&buildTypeId=Lite_UniverseTest&tab=buildLog&_focus=17506#_state=486) warnings:
```
Microsoft.AspNetCore.SignalR.Client.FunctionalTests: Non-serializable data ('System.Object[]') found for 'Microsoft.AspNetCore.SignalR.Client.FunctionalTests.HubConnectionTests.CanInvokeClientMethodFromServer'; falling back to single test case.
```
are appearing in builds of SignalR, they sound like they're preventing the full compliment of tests from running.

CC @jkotalik

jkotalik:
I don't think I'm the appropriate person for this task. cc @anurse @BrennanConroy @mikaelm12

anurse:
These issues don't cause any problem running the test. They simply stop Test Explorer from showing them as multiple tests.

anurse:
Re-opening to track making these data serializable so that we get a clean Test Explorer experience. It's suuuuuuper puntable :)

PR #1820, opened 2018-04-01
Fix non-serializable tests

JamesNK:
Also fix ignoring empty string on CloseMessage

https://github.com/aspnet/SignalR/issues/1649


JamesNK:
Outstanding are a couple of tests with the `WebSocketMessageType` enum parameter that get this error. Any idea why? Other parameterized tests have enum arguments and are fine.

https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs#L27-L111

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -1686,15 +1688,16 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
         {
             get
             {
+
```
davidfowl: super nit: extra space.

davidfowl:
> Outstanding are a couple of tests with the WebSocketMessageType enum parameter that get this error. Any idea why? Other parameterized tests have enum arguments and are fine.

Nope, thats odd.

JamesNK:
Creating an almost identical enum and using that instead fixed it 😕

test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
```diff
@@ -24,10 +24,18 @@ public WebSocketsTests(ITestOutputHelper output)
         {
         }
 
+        // For unknown reasons using the built-in WebSocketMessageType as a parameter
+        // causes "Non-serializable data" warnings in xunit
+        public enum TestWebSocketMessageType
+        {
+            Binary = 1,
+            Text = 0
```
anurse: Might be worth filing a bug, but I don't know exactly where it would go. I think maybe it's xunit here? `WebSocketMessageType` doesn't look any different from this enum really: https://source.dot.net/#System.Net.WebSockets/System/Net/WebSockets/WebSocketMessageType.cs,17aa513fb8c7e9fc

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -254,5 +269,25 @@ private static byte[] FormatMessageToArray(byte[] message)
             TextMessageFormatter.WriteRecordSeparator(output);
             return output.ToArray();
         }
+
+        public class ProtocolTestData1
```
anurse: ProtocolTestData**1**?

test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
anurse: Sounds like it's a problem on desktop framework when the enum is in the GAC. https://twitter.com/anurse/status/980825964757176320

bradwilson: https://github.com/xunit/xunit/blob/4d2679934462bc2bc459f340d8bcfce624ce1512/test/test.xunit.execution/Common/XunitSerializationInfoTests.cs#L200-L210

bradwilson: And not just enums, anything from the GAC (that we don't already special case): https://github.com/xunit/xunit/blob/4d2679934462bc2bc459f340d8bcfce624ce1512/test/test.xunit.execution/Common/XunitSerializationInfoTests.cs#L233-L245

test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
```diff
@@ -24,10 +24,18 @@ public WebSocketsTests(ITestOutputHelper output)
...
+        }
+
         [Theory]
-        [InlineData(WebSocketMessageType.Text)]
-        [InlineData(WebSocketMessageType.Binary)]
-        public async Task ReceivedFramesAreWrittenToChannel(WebSocketMessageType webSocketMessageType)
+        [InlineData(TestWebSocketMessageType.Text)]
+        [InlineData(TestWebSocketMessageType.Binary)]
```
bradwilson: As I suggested on Twitter, I believe we convert string data into enum values. I don't know if `nameof` supports enum values (to help preserve values across renames), but given your work-around, you've already given up on rename refactor support anyway...

test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
davidfowl: Another reason to hate the GAC

test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
anurse: `nameof` should work here actually, so we could try that.

bradwilson:
Did auto-conversion from string test data to Enum parameter not work correctly?

PR #1822, opened 2018-04-02
Detailed errors enabled for functional tests

BrennanConroy:
Review: `InvalidDataException` or `FormatException`?


anurse:
`FormatException`: The exception that is thrown when the format of an argument is invalid, or when a composite format string is not well formed.

`InvalidDataException`: The exception that is thrown when a data stream is in an invalid format.

`InvalidDataException` all the way. `FormatException` is for things like `string.Format` and `IFormattable`

PR #1824, opened 2018-04-02
Reaction to *Memory changes

pakrym:


test/Microsoft.AspNetCore.SignalR.Tests.Utils/PipeReaderExtensions.cs
```diff
@@ -28,8 +28,8 @@ public static class PipeReaderExtensions
                 }
                 finally
                 {
-                    // Consume nothing, just wait for everything
-                    pipeReader.AdvanceTo(result.Buffer.Start, result.Buffer.End);
+                    // Don't consume or advance
+                    pipeReader.AdvanceTo(result.Buffer.Start, result.Buffer.Start);
```
anurse: Is this correct? Why did the behavior change?

anurse: I suppose it doesn't really hurt the logic of this method though.

BrennanConroy: Before, the next read would wait until another message, and not grab the one that this function read.

pakrym: Because it might leave pipe in the state where subsequent ReadAsync would never complete(no more data is written) and I doubt it desired behavior of WaitToRead method.

PR #1825, opened 2018-04-02
Fix build

JamesNK:
Somehow this worked on Travis but failed on Appveyor/Windows 😕


davidfowl:
> How could that work on some and fail on others? This is plain boring managed code, no?

.NET Framework vs .NET Core. Travis runs .NET Core only.

Eilon:
So we think there's a bug in .NET Framework in this?

PR #1827, opened 2018-04-02
Couple logging additions

BrennanConroy:


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -553,7 +553,7 @@ private void CheckDisposed()
                     }
                 }
             }
-            catch (Exception ex)
+            catch (Exception ex) when (!(ex is HubException))
```
davidfowl: When would this ever happen?

BrennanConroy: When you received a handshake response with an `"error"` set

davidfowl: That seems like a product bug no?

davidfowl: Ah I see, this code looks like it needs to be reworked, it's throwing and logging then rethrowing...

davidfowl: Add a comment please

clients/ts/signalr/src/Utils.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000;
```
davidfowl: Isn't this just a helper file?

Issue #1662, opened 2018-03-20
The TS client doesn't support IPv6 address URLs in Edge


Tragetaschen:
Here's an annoying mess:

In Edge (Fall Creators Update), I cannot use IPv6 URLs (`http://[2003:…:3b]/`) in the browser when I want to connect to the server with a relative URL. In [the relevant code](https://github.com/aspnet/SignalR/blob/3a3cebb884ac4164c5f773a4ab652e5a19fbe290/client-ts/signalr/src/HttpConnection.ts#L238), `parser.host` is just `2003:…:3b` (without the brackets), so the resulting URL is invalid. `window.document.location.host` contains the brackets, though.

In Chrome, both `host` properties contain the brackets, so it works.
In Firefox for relative URLs, the `a`-element's `host` property is the empty string and the location's property contains the brackets, so things work.

Tragetaschen:
I've also created https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16476907/

anurse:
Given that this is an issue in Edge, I think our preference would certainly be to get it fixed there. If necessary though, we could probably work around it with custom URL parsing.

anurse:
We don't think we're going to be able to get to this in 2.1. It would mean building some URL parsing code and that's got a bunch of edge cases. We'd certainly consider a PR with lots of tests ;). There is a workaround in that you can resolve the absolute URL manually and pass it in. Hopefully Edge will have a fix for this.

mikaelm12:
Fixed in: e51852d0fc91c10b8eac641d137d10c9348589ec as pointed out in @Tragetaschen's comment here https://github.com/aspnet/SignalR/pull/1828#issuecomment-378514495. Good find!

Tragetaschen:
And for good measure: The issue has been marked fixed for Edge and should appear officially in the next (1809) release of Windows.

Issue #1436, opened 2018-02-12
signalr-client "normalizes" network-path references that start with "//".


686d7066:
We have a website and multiple endpoints that provide SignalR hubs.
e.g. the website is served from localhost:5000 and needs to be able to connect to a hub on localhost:5001.

Since the endpoint is not the same as the URL the website was served from we can't simply provide the HubName to the HubConnection. So we provide the URL as a network-path reference, because the website currently must support HTTP and HTTPS. 
So we pass '//localhost:5001/signalr/NotificationHub' to the HubConnection.
We are receiving the following message in the browsers console:

> Information: Normalizing '//localhost:5001/signalr/NotificationHub' to 'http://localhost:5001//localhost:5001/signalr/NotificationHub'

This happens with the signalr-client package with version 1.0.0-alpha2-final.

The client should not normalize URLs beginning with // by adding protocol AND host (again) and instead let the browser handle the network-path reference (to keep in line with https://tools.ietf.org/html/rfc3986#section-4.2) or alternatively only put the current protocol in front.

anurse:
Seems legit. Also fairly straightforward. If you're at all interested in contributing, we'd certainly take a PR that fixes this! We'll put it in our queue as well though.

pardahlman:
@anurse I'd be happy to try to help out here. The only thing I'm uncertain about is how I should go about testing this change... `resolveUrl`, `baseUrl` and `url` are all private. Any ideas?

anurse:
I think we'd do it through mocking our HttpClient interface, like we do with the other HubConnection tests in the TS client. They're a little rough, but you could probably copy one of the others. We just need to verify that the URL that goes to the HttpClient and then we can ignore remaining errors.

Wedvich:
As a dumb temporary workaround, we monkeypatched SignalR in our client app like this, which makes it work as expected for different HTTP/HTTPS endpoints:
```js
import * as signalR from '@aspnet/signalr';

const originalResolveUrl = signalR.HttpConnection.prototype.resolveUrl;
signalR.HttpConnection.prototype.resolveUrl = url =>
  originalResolveUrl(url.indexOf('//') === 0 ? window.document.location.protocol + url : url);
```
Looking forward to a _real_ fix though :smile:

anurse:
[...]

PR #1828, opened 2018-04-03
Don't normalize urls that start with //

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1436
Need to think about how to test this.

### DIFF
--- a/clients/ts/signalr/src/HttpConnection.ts
+++ b/clients/ts/signalr/src/HttpConnection.ts
@@ -264,7 +264,7 @@ export class HttpConnection implements IConnection {
 
     private resolveUrl(url: string): string {
         // startsWith is not supported in IE
-        if (url.lastIndexOf("https://", 0) === 0 || url.lastIndexOf("http://", 0) === 0 || url.lastIndexOf("//", 0) === 0) {
+        if (url.lastIndexOf("https://", 0) === 0 || url.lastIndexOf("http://", 0) === 0) {
             return url;
         }
 
@@ -272,20 +272,11 @@ export class HttpConnection implements IConnection {
             throw new Error(`Cannot resolve '${url}'.`);
         }
 
-        const parser = window.document.createElement("a");
-        parser.href = url;
+        const aTag = window.document.createElement("a");
+        aTag.href = url;
 
-        const baseUrl = (!parser.protocol || parser.protocol === ":")
-            ? `${window.document.location.protocol}//${(parser.host || window.document.location.host)}`
-            : `${parser.protocol}//${parser.host}`;
-
-        if (!url || url[0] !== "/") {
-            url = "/" + url;
-        }
-
-        const normalizedUrl = baseUrl + url;
-        this.logger.log(LogLevel.Information, `Normalizing '${url}' to '${normalizedUrl}'.`);
-        return normalizedUrl;
+        this.logger.log(LogLevel.Information, `Normalizing '${url}' to '${aTag.href}'.`);
+        return aTag.href;
     }
 
     private resolveNegotiateUrl(url: string): string {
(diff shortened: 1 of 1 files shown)


JamesNK:
You can test via passing in a test version of `IHttpClient`, and testing what URL `HttpConnection` gives to it.

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -264,7 +264,7 @@ export class HttpConnection implements IConnection {
 
     private resolveUrl(url: string): string {
         // startsWith is not supported in IE
-        if (url.lastIndexOf("https://", 0) === 0 || url.lastIndexOf("http://", 0) === 0) {
+        if (url.lastIndexOf("https://", 0) === 0 || url.lastIndexOf("http://", 0) === 0 || url.lastIndexOf("//", 0) === 0) {
```
anurse: This is going to fail on Node. It will return `//foo`, which isn't a valid URL. Node clients will get to line 271 and then throw, which makes sense, there's no "page" to resolve URLs relative to, so having Node clients throw when given a non-absolute URL (even one using `//`) is preferred.

Instead, let's refactor the part **below** the test for `window` further down, to use the `a` tag we're using, but more aggressively. I was doing some testing and I found that the following code pretty consistently normalizes URLs in Chrome (at least):

```javascript
const parser = document.createElement("a");
parser.href = url;
return parser.href; // Re-reading the property gives you the absolute URL, resolved by the browser.
```

As for testing, since we're testing browser-bound logic (URL resolution), the tests would need to be in our browser functional test suite. I'd still suggest using a test HttpClient though, since we don't really want to make real requests, just check that the URL resolution logic works.

anurse: As an added bonus, that will also resolve https://github.com/aspnet/SignalR/issues/1835 ; two birds, one stone ;)

mikaelm12:
Updated. Setting the href on the a tag handles some of the logic that we had. This is cleaner.

anurse:
Have you done some manual verification? Can you describe it here? We should look at tests, but as I said, there are difficulties. In the meantime I'd like a quick run down of how you tested.

Tragetaschen:
This kind of URL normalization would also fix #1662. `aTag.href` includes all the necessary brackets when set to a relative path in Edge.

image

anurse:
Nice find @Tragetaschen ! Double-whammy :)

davidfowl:
I think this is what old SignalR did https://github.com/SignalR/SignalR/blob/f5544f24c4a79bf2cca8850830a81bcfb67e1feb/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js#L396

anurse:
Yeah, but I can't see why. That logic means `new signalR.HubConnection('foo')` is always resolved as `http://[host]/foo`. That doesn't match with how URLs are resolved in any other place on a web page. Do we want to preserve the old behavior?

mikaelm12:
Discussed with @anurse a functional testing strategy. Will update the pr tonight

mikaelm12:
So I'm actually going to suggest that we move the functional testing of this out. I've done manual verification of the url normalization and getting the behavior we want 

Page relative normalization - "default" to "http://localhost:5000/foo/default"
normalize-url1

Absolute normalization - "/foo/default" to "http://localhost:5000/foo/default"
normalize-url2

Network path reference normalization - "//localhost:5000/foo/default" to ""http://localhost:5000/foo/default"
network-relativepath

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -272,20 +272,11 @@ export class HttpConnection implements IConnection {
             throw new Error(`Cannot resolve '${url}'.`);
         }
 
-        const parser = window.document.createElement("a");
-        parser.href = url;
+        const aTag = window.document.createElement("a");
+        aTag.href = url;
```
JamesNK: Because this can't easily be unit tested I think there should be a comment describing the different URL scenarios that this logic normalizes to keep future devs in the loop

Issue #1688, opened 2018-03-22
Rework HubConnectionBuilder


BrennanConroy:
~~Specifically we want to log unknown message types~~

Also, we can consider adding a service collection as the backend for the `HubConnectionBuilder` to have DI on the client.

Some context: https://github.com/aspnet/SignalR/pull/1666#discussion_r176460811

anurse:
If we do this right, we could end up with the same extension methods by having the new builder implement `ISignalRBuilder` (or something like that)

PR #1830, opened 2018-04-03
[WIP] Refactor HubConnectionBuilder

JamesNK:
https://github.com/aspnet/SignalR/issues/1688


src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -2,283 +2,42 @@
...
-                    WebSocketOptions = hubConnectionBuilder.GetWebSocketOptions(),
-                    Cookies = hubConnectionBuilder.GetCookies(),
-                    Proxy = hubConnectionBuilder.GetProxy(),
-                    UseDefaultCredentials = hubConnectionBuilder.GetUseDefaultCredentials(),
-                    ClientCertificates = hubConnectionBuilder.GetClientCertificates(),
-                    Credentials = hubConnectionBuilder.GetCredentials(),
-                };
+            Func<IConnection> createConnection = () => new HttpConnection(options.Url,
+                options.Transport ?? TransportType.All,
+                null, // TODO: Pass in logger factory
```
JamesNK: Thoughts on what to do here?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionBuilderExtensionsTests.cs
```diff
@@ -1,131 +1,131 @@
...
-{
-    public class HubConnectionBuilderExtensionsTests
-    {
-        [Fact]
-        public void WithProxyRegistersGivenProxy()
-        {
-            var connectionBuilder = new HubConnectionBuilder();
-            var proxy = Mock.Of<IWebProxy>();
-            connectionBuilder.WithProxy(proxy);
-            Assert.Same(proxy, connectionBuilder.GetProxy());
```
JamesNK: What level of testing is needed for getting and setting options on `WithHttpConnection`?

Because they're used to create `HttpConnection` there isn't a simple way to check they have been set. One idea is to make `_httpOptions` on `HttpConnection` internal.

src/Microsoft.AspNetCore.SignalR.Client.Core/IHubConnectionBuilder.cs
```diff
@@ -3,14 +3,17 @@
 
 using System;
 using Microsoft.AspNetCore.Connections;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
     public interface IHubConnectionBuilder
```
davidfowl: This should implement ISignalRBuilder (ISignalRBuilder should be moved into common)

samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
```diff
@@ -1,13 +1,18 @@
...
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
     public static class TcpHubConnectionBuilderExtensions
     {
         public static IHubConnectionBuilder WithEndPoint(this IHubConnectionBuilder builder, IPEndPoint endPoint)
         {
-            builder.ConfigureConnectionFactory(() => new TcpConnection(endPoint));
+            Func<IConnection> connectionFactory = () => new TcpConnection(endPoint);
+            builder.Services.Replace(new ServiceDescriptor(typeof(Func<IConnection>), connectionFactory));
```
davidfowl: We never use replace, we most use Add, last one wins.

JamesNK: So `Add` will add or replace a service, and `Replace` also will add or replace a service? Huh

Ok

src/Microsoft.AspNetCore.SignalR.Client.Core/IHubConnectionBuilder.cs
JamesNK: The problem with that is all of the `ISignalRBuilder` extension methods don't apply to the client. The only ones which kind of make sense are `AddJsonProtocol` and `AddMessagePackProtocol`, but they build an enumerable of protocols, while the client only makes sense with one.

Intellisense will be filled with extension methods that will confuse any devs that brought in non-client packages into their project.

davidfowl: Then I think we should consider doing ISignaRClientBuilder : ISignalRBuilder. The intent being that something will be shared (like the protocol) and should not require a developer to write 2 packages.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -2,283 +2,42 @@
...
+        public static IHubConnectionBuilder WithHttpConnection(this IHubConnectionBuilder hubConnectionBuilder, Action<HttpConnectionOptions> configureHttpConnection)
         {
-            if (url == null)
-            {
-                throw new ArgumentNullException(nameof(url));
-            }
-
-            return hubConnectionBuilder.WithUrl(new Uri(url));
-        }
+            HttpConnectionOptions options = new HttpConnectionOptions();
```
davidfowl: We should be using `IOptions<HttpConnectionOptions>` here.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
davidfowl: > Get the logger descriptor from the builder's ServiceCollection?

Definately not. We should defer it all. 

```
services.AddSingleton<Func<IConnection>>(serviceProvider =>
{
     return () => new HttpConnection(....)
});
```

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -752,10 +761,13 @@ private Func<IConnection> GetHttpConnectionFactory(ILoggerFactory loggerFactory,
...
                 var hubConnection = new HubConnectionBuilder()
-                    .WithUrl(_serverFixture.Url + "/default")
-                    .WithTransport(TransportType.WebSockets)
                     .WithLoggerFactory(loggerFactory)
-                    .WithWebSocketOptions(options => options.Cookies = cookieJar)
+                    .WithHttpConnection(options =>
+                    {
+                        options.Url = new Uri(_serverFixture.Url + "/default");
+                        options.Transport = TransportType.WebSockets;
+                        options.WebSocketOptions = o => o.Cookies = cookieJar;
```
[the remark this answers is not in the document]
anurse: [`ClientWebSocket.Options`](https://msdn.microsoft.com/en-us/library/system.net.websockets.clientwebsocket.options(v=vs.110).aspx) is read-only

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
```diff
@@ -6,68 +6,54 @@
...
-            return true;
-        }
+        public IServiceCollection Services { get; } = new ServiceCollection();
 
         public HubConnection Build()
         {
-            if (_connectionFactoryDelegate == null)
+            var provider = Services.BuildServiceProvider();
+
+            var connectionFactory = provider.GetService<Func<IConnection>>();
```
anurse: Should we make this an interface now to be clearer in DI? `IConnectionFactory`?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
```diff
@@ -6,68 +6,54 @@
...
 
-            IHubConnectionBuilder builder = this;
-
-            var loggerFactory = builder.GetLoggerFactory();
-            var hubProtocol = builder.GetHubProtocol();
+            var hubProtocol = provider.GetService<IHubProtocol>();
+            var loggerFactory = provider.GetService<ILoggerFactory>();
 
-            return new HubConnection(_connectionFactoryDelegate, hubProtocol ?? new JsonHubProtocol(), loggerFactory);
+            return new HubConnection(connectionFactory, hubProtocol ?? new JsonHubProtocol(), loggerFactory);
```
anurse: We can use [`ActivatorUtilities.CreateInstance`](https://github.com/aspnet/DependencyInjection/blob/dev/shared/Microsoft.Extensions.ActivatorUtilities.Sources/ActivatorUtilities.cs#L41) to just create `HubConnection` and resolve everything it needs out of DI.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -2,283 +2,42 @@
...
-        public static readonly string AccessTokenFactoryKey = "AccessTokenFactory";
-        public static readonly string WebSocketOptionsKey = "WebSocketOptions";
-        public static readonly string CookiesKey = "Cookies";
-        public static readonly string ProxyKey = "Proxy";
-        public static readonly string ClientCertificatesKey = "ClientCertificates";
-        public static readonly string CredentialsKey = "Credentials";
-        public static readonly string UseDefaultCredentialsKey = "UseDefaultCredentials";
-
-        public static IHubConnectionBuilder WithUrl(this IHubConnectionBuilder hubConnectionBuilder, string url)
+        public static IHubConnectionBuilder WithHttpConnection(this IHubConnectionBuilder hubConnectionBuilder, Action<HttpConnectionOptions> configureHttpConnection)
```
anurse: Make the `url` a first class argument. It's the only mandatory setting, so it should be "required" by the API

src/Microsoft.AspNetCore.SignalR.Client.Core/IHubConnectionBuilder.cs
davidfowl: Yep

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
anurse: If we had `IConnectionFactory` as an interface, we could just accept everything as constructor parameters to `HttpConnectionFactory`. Then we wouldn't need a delegate, we'd just register `HttpConnectionFactory` with `TryAdd` and `IOptions<HttpConnectionOptions>` with `Configure`.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionBuilderExtensionsTests.cs
anurse: I don't think it's necessary to test individual setters/getters like this. Most things should be testable by creating the container, building the HubConnection and then testing some setting or behavior. If we need to expose internal options objects for testing as a last resort, we can do that, but I think it'd be preferable to test behaviors where possible since that's what the customer actually cares about.

src/Microsoft.AspNetCore.SignalR.Client.Core/IHubConnectionBuilder.cs
JamesNK: Ok, but what is the benefit of sharing the `ISignalRBuilder` interface is nothing uses it directly.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
JamesNK: But it will be orders of magnitude slower because of reflection. Even using compiled expressions will be slower because they'll be interpreted on Android/iOS.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
JamesNK: Sounds good. That will also allow an overload of string and Uri.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionBuilderExtensionsTests.cs
JamesNK: All the HTTP settings would really only be testable by starting the connection and sending a HTTP request. It starts to become less a simple unit test and more an integration test. Maybe having an IConnectionFactory will make it easier...

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
davidfowl: Where would that interface live?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
davidfowl: Yea lets avoid that right now.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
JamesNK:
```
services.AddSingleton<Func<IConnection>>(serviceProvider =>
{
     return () => new HttpConnection(....)
});
```

Yeah I think that would work. The only thing is a func returning a func does feel a little inside baseball for anyone who wants to hook in their own IConnection.

An IConnectionFactory might simply things, but it would need access to the `ServiceCollection` when creating the IConnection instance to get the logger factory. How would that work? Set a logger on `HttpConnectionOptions` and get it that way?

anurse: > but it would need access to the ServiceCollection when creating the IConnection instance to get the logger factory

Why? It could accept `ILoggerFactory` in it's constructor and stash it away to use when constructing the connection. If necessary, it could also accept `IServerProvider` in it's constructor to get the container itself.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionBuilderExtensionsTests.cs
anurse: > It starts to become less a simple unit test and more an integration test.

Yeah, I agree there. For some of these settings we already have integration tests that use these methods and I don't see a big reason to duplicate that logic here.

src/Microsoft.AspNetCore.SignalR.Client.Core/IHubConnectionBuilder.cs
anurse: Yeah, it's a bit unclear to me what we would share. The current implementation uses `TryAddEnumerable` which means you could get in weird order-dependent situations if you try to add both.

davidfowl: The protocols will use them directly and we will remove the 2 msg pack dlls and just have a single one.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -28,28 +29,20 @@ public static IHubConnectionBuilder WithJsonProtocol(this IHubConnectionBuilder
...
+            if (loggerFactory == null)
+            {
+                throw new ArgumentNullException(nameof(loggerFactory));
+            }
+
+            hubConnectionBuilder.Services.AddSingleton(loggerFactory);
             return hubConnectionBuilder;
         }
 
         public static IHubConnectionBuilder WithLogger(this IHubConnectionBuilder hubConnectionBuilder, Action<ILoggerFactory> configureLogging)
```
davidfowl: I don't want you to fix this now but we have a design issue by newing up the LoggerFactory where it never gets disposed.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -2,283 +2,50 @@
...
+                Proxy = options.Proxy,
+                UseDefaultCredentials = options.UseDefaultCredentials,
+                ClientCertificates = options._clientCertificates,
+                Credentials = options.Credentials,
+            };
 
-            hubConnectionBuilder.AddSetting(WebSocketOptionsKey, configureWebSocketOptions);
+            Func<IConnection> createConnection = () => new HttpConnection(options.Url,
+                options.Transport ?? TransportType.All,
+                null, // TODO: Pass in logger factory
```
davidfowl: Why didn't you fix this?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -50,10 +50,12 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
             using (StartLog(out var loggerFactory, $"{nameof(CheckFixedMessage)}_{protocol.Name}_{transportType}_{path.TrimStart('/')}"))
             {
                 var connection = new HubConnectionBuilder()
-                    .WithUrl(_serverFixture.Url + path)
-                    .WithTransport(transportType)
                     .WithLoggerFactory(loggerFactory)
                     .WithHubProtocol(protocol)
+                    .WithHttpConnection(_serverFixture.Url + path, options =>
```
davidfowl: I would make the transport type a first class as well argument as well.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
JamesNK: Patience.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
JamesNK: Optional argument? Currently it default to All.

davidfowl: Yea

JamesNK: Got me thinking but is off topic - https://github.com/aspnet/SignalR/issues/1842

src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+        public Uri Url { get; set; }
+        public TransportType? Transport { get; set; }
+        public Func<HttpMessageHandler, HttpMessageHandler> MessageHandlerFactory { get; set; }
+        public bool? UseDefaultCredentials { get; set; }
+        public ICredentials Credentials { get; set; }
+        public IWebProxy Proxy { get; set; }
+        public Func<string> AccessTokenFactory { get; set; }
+        public Action<ClientWebSocketOptions> WebSocketOptions { get; set; }
+
+        public void AddClientCertificate(X509Certificate clientCertificate)
```
davidfowl: I'm not a fan of these methods. They make it harder to do general purpose collection copying (which can be useful for proxying things)

JamesNK:
🆙 📅

src/Microsoft.AspNetCore.SignalR.Client.MsgPack/Microsoft.AspNetCore.SignalR.Client.MsgPack.csproj
```diff
@@ -1,13 +0,0 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
davidfowl: 👏

src/Microsoft.AspNetCore.SignalR.Client.MsgPack/MessagePackProtocolHubConnectionBuilderExtensions.cs
```diff
@@ -1,19 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
davidfowl: 👏

src/Microsoft.AspNetCore.SignalR.Core/JsonProtocolDependencyInjectionExtensions.cs
```diff
@@ -16,7 +17,8 @@ public static class JsonProtocolDependencyInjectionExtensions
         /// </remarks>
         /// <param name="builder">The <see cref="ISignalRBuilder"/> representing the SignalR server to add JSON protocol support to.</param>
         /// <returns>The value of <paramref name="builder"/></returns>
-        public static ISignalRBuilder AddJsonProtocol(this ISignalRBuilder builder) => AddJsonProtocol(builder, _ => { });
+        public static TBuilder AddJsonProtocol<TBuilder>(this TBuilder builder) where TBuilder : ISignalRBuilder
```
davidfowl: Why is it generic? Why not just ISignalRBuilder

src/Microsoft.AspNetCore.SignalR.Core/JsonProtocolDependencyInjectionExtensions.cs
```diff
@@ -27,9 +29,10 @@ public static class JsonProtocolDependencyInjectionExtensions
         /// <param name="builder">The <see cref="ISignalRBuilder"/> representing the SignalR server to add JSON protocol support to.</param>
         /// <param name="configure">A delegate that can be used to configure the <see cref="JsonHubProtocolOptions"/></param>
         /// <returns>The value of <paramref name="builder"/></returns>
-        public static ISignalRBuilder AddJsonProtocol(this ISignalRBuilder builder, Action<JsonHubProtocolOptions> configure)
+        public static TBuilder AddJsonProtocol<TBuilder>(this TBuilder builder, Action<JsonHubProtocolOptions> configure) where TBuilder : ISignalRBuilder
         {
-            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IHubProtocol, JsonHubProtocol>());
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IHubProtocol, JsonHubProtocol>(
```
davidfowl: Why the change?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -23,11 +24,18 @@ public class ConnectionLifecycle
 
             public static IEnumerable<object[]> MethodsNamesThatRequireActiveConnection => MethodsThatRequireActiveConnection.Keys.Select(k => new object[] { k });
 
+            private HubConnection CreateHubConnection(Func<IConnection> connectionFactory)
+            {
+                IServiceCollection serviceCollection = new ServiceCollection();
+                var serviceProvider = serviceCollection.BuildServiceProvider();
+                return new HubConnection(connectionFactory, new JsonHubProtocol(), serviceProvider);
```
davidfowl: This looks messed up. Now I think we need to go further here and stick the HubConnection itself in the container so we can just resolve it.

src/Microsoft.AspNetCore.SignalR.MsgPack/MsgPackProtocolDependencyInjectionExtensions.cs
```diff
@@ -27,9 +29,10 @@ public static class MsgPackProtocolDependencyInjectionExtensions
         /// <param name="builder">The <see cref="ISignalRBuilder"/> representing the SignalR server to add MsgPack protocol support to.</param>
         /// <param name="configure">A delegate that can be used to configure the <see cref="MessagePackHubProtocolOptions"/></param>
         /// <returns>The value of <paramref name="builder"/></returns>
-        public static ISignalRBuilder AddMessagePackProtocol(this ISignalRBuilder builder, Action<MessagePackHubProtocolOptions> configure)
+        public static TBuilder AddMessagePackProtocol<TBuilder>(this TBuilder builder, Action<MessagePackHubProtocolOptions> configure) where TBuilder : ISignalRBuilder
         {
-            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IHubProtocol, MessagePackHubProtocol>());
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IHubProtocol, MessagePackHubProtocol>(
+                serviceProvider => new MessagePackHubProtocol(serviceProvider.GetService<IOptions<MessagePackHubProtocolOptions>>())));
```
davidfowl: Why the change?

src/Microsoft.AspNetCore.SignalR.Core/JsonProtocolDependencyInjectionExtensions.cs
JamesNK: Because the only extension methods that hang off ISignalRBuilder are AddJsonProtocol and AddMessagePackProcotol, and if ISignalRBuilder is returned then you can't chain anymore methods off it, e.g. WithUrl and Build below won't work.

```
var builder = new HubConnectionBuilder().AddHubProtocol().WithUrl(url).Build();
```

Need to return the passed in type using generics.

src/Microsoft.AspNetCore.SignalR.Core/JsonProtocolDependencyInjectionExtensions.cs
JamesNK: Because otherwise reflection and expression trees will be used to call the constructor, which we wanted to avoid on AOT.

src/Microsoft.AspNetCore.SignalR.Core/JsonProtocolDependencyInjectionExtensions.cs
davidfowl: I see, awful 😄

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
JamesNK: It doesn't matter here because it is a test, but do you want HubConnectionBuilder to use the container to create the HubConnection? Reflection, etc, will be involved.

src/Microsoft.AspNetCore.SignalR.Core/JsonProtocolDependencyInjectionExtensions.cs
davidfowl: Nahhh, lets keep it here. That happens once

samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
```diff
@@ -1,13 +1,18 @@
...
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
     public static class TcpHubConnectionBuilderExtensions
     {
         public static IHubConnectionBuilder WithEndPoint(this IHubConnectionBuilder builder, IPEndPoint endPoint)
         {
-            builder.ConfigureConnectionFactory(() => new TcpConnection(endPoint));
+            Func<IConnection> connectionFactory = () => new TcpConnection(endPoint);
+            builder.Services.AddSingleton(connectionFactory);
```
JamesNK: I think WithConnectionFactory should have its own extension method. Thoughts?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
```diff
@@ -2,72 +2,72 @@
...
         {
-            _settings[new KeyValuePair<string, Type>(name, typeof(T))] = value;
-        }
-
-        public bool TryGetSetting<T>(string name, out T value)
-        {
-            value = default;
-            if (!_settings.TryGetValue(new KeyValuePair<string, Type>(name, typeof(T)), out var setting))
+            Services = new ServiceCollection();
+            Services.AddSingleton<HubConnection>(serviceProvider =>
```
JamesNK: Right now this is a singleton. Should calling `Build()` on a HubConnectionBuilder multiple times return different instances?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
```diff
@@ -2,72 +2,72 @@
...
+                    throw new InvalidOperationException("Cannot create HubConnection instance. A connection was not configured.");
+                }
 
-            value = (T)setting;
-            return true;
+                var hubProtocol = serviceProvider.GetService<IHubProtocol>();
+                var loggerFactory = serviceProvider.GetService<ILoggerFactory>();
+
+                // Note that the root service provider that can be disposed is passed to HubConnection
+                return new HubConnection(connectionFactory, hubProtocol ?? new JsonHubProtocol(), _serviceProvider, loggerFactory);
```
JamesNK: HubConnection ctor is internal so I believe this is the only way to call it with ASP.NET Core's DI.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/HubConnectionSendBenchmark.cs
```diff
@@ -46,8 +49,15 @@ public void GlobalSetup()
             connection.Features.Set<IConnectionInherentKeepAliveFeature>(new TestConnectionInherentKeepAliveFeature());
             connection.Transport = _pipe;
 
+            Func<IConnection> connectionFactory = () => connection;
+
             var protocol = Protocol == "json" ? (IHubProtocol)new JsonHubProtocol() : new MessagePackHubProtocol();
-            _hubConnection = new HubConnection(() => connection, protocol, new NullLoggerFactory());
+
+            HubConnectionBuilder hubConnectionBuilder = new HubConnectionBuilder();
```
davidfowl: var

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -23,11 +24,18 @@ public class ConnectionLifecycle
 
             public static IEnumerable<object[]> MethodsNamesThatRequireActiveConnection => MethodsThatRequireActiveConnection.Keys.Select(k => new object[] { k });
 
+            private HubConnection CreateHubConnection(Func<IConnection> connectionFactory)
+            {
+                HubConnectionBuilder builder = new HubConnectionBuilder();
```
davidfowl: var

Issue #1733, opened 2018-03-27
Potential memory leak if pipes are not cleaned up on start failure


davidfowl:
https://github.com/aspnet/SignalR/blob/434147fe8e2952df6f44fc1adebe97a19d6eef09/src/Microsoft.AspNetCore.Sockets.Client.Http/HttpConnection.cs#L431-L446

Depending on how the transport failed, we need to make sure both ends of the pipe are completed when we decide to ignore a transport because starting throws.

anurse:
So all we need to do here is complete the pipes in `pair` within that `catch` block then.

PR #1836, opened 2018-04-03
Clean up pipe pair on transport start failure

mikaelm12:
Completing pipes when our transport fails to start.
Issue: https://github.com/aspnet/SignalR/issues/1733

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
@@ -345,6 +345,8 @@ namespace Microsoft.AspNetCore.Http.Connections.Client
             catch (Exception ex)
             {
                 Log.ErrorStartingTransport(_logger, transport, ex);
+
+                // Clean up pipes and null out transport when we fail to start.
                 pair.Transport.Input.Complete();
                 pair.Transport.Output.Complete();
                 pair.Application.Input.Complete();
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -356,6 +357,14 @@ private static Uri CreateConnectUrl(Uri url, string connectionId)
             Log.TransportStarted(_logger, _transport);
         }
 
+        private void completePipePair(DuplexPipe.DuplexPipePair duplexPipePair)
```
anurse: Pass the two pipes in separately. I don't like having `DuplexPipe.DuplexPipePair` here :). Just a style thing.

mikaelm12: Pulled the complete of the pipes out of the method entirely.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -345,6 +345,10 @@ private static Uri CreateConnectUrl(Uri url, string connectionId)
             catch (Exception ex)
             {
                 Log.ErrorStartingTransport(_logger, transport, ex);
+                pair.Transport.Input.Complete();
```
davidfowl: Add a comment

JamesNK:
Is this something that can be unit tested?

davidfowl:
Yes, making a fake transport that throws in StartAsync then observing that the application pipes are closed. The transport pipe is trickier because it's never exposed and it's nulled out.

PR #1792, opened 2018-03-30
Add ability to log message content

anurse:
Also improve how browser tests capture logs so you can review them easily from both the command prompt and the browser.

This is for post-preview2 (not worth destabilizing) so it is going in to `dev`.

PR #1838, opened 2018-04-03
Use xhr.onreadystatechange instead of xhr.onload

davidfowl:
- Added more logging

### DIFF
--- a/clients/ts/signalr/src/HttpClient.ts
+++ b/clients/ts/signalr/src/HttpClient.ts
@@ -92,7 +92,7 @@ export class DefaultHttpClient extends HttpClient {
                         request.abortSignal.onabort = null;
                     }
 
-                    // Some browsers report xhr.status == 0 when the 
+                    // Some browsers report xhr.status == 0 when the
                     // response has been cut off or there's been a TCP FIN.
                     // Treat it like a 200 with no response.
                     if (xhr.status === 0) {
--- a/clients/ts/signalr/src/HttpClient.ts
+++ b/clients/ts/signalr/src/HttpClient.ts
@@ -84,15 +84,24 @@ export class DefaultHttpClient extends HttpClient {
                 xhr.timeout = request.timeout;
             }
 
-            xhr.onload = () => {
-                if (request.abortSignal) {
-                    request.abortSignal.onabort = null;
-                }
+            xhr.onreadystatechange = () => {
+                this.logger.log(LogLevel.Trace, `onreadystatechange(${xhr.readyState}:${xhr.status}:${xhr.statusText})`);
 
-                if (xhr.status >= 200 && xhr.status < 300) {
-                    resolve(new HttpResponse(xhr.status, xhr.statusText, xhr.response || xhr.responseText));
-                } else {
-                    reject(new HttpError(xhr.statusText, xhr.status));
+                if (xhr.readyState === 4) {
+                    if (request.abortSignal) {
+                        request.abortSignal.onabort = null;
+                    }
+
+                    // Some browsers report xhr.status == 0 when the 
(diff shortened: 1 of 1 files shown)


davidfowl:
The build is fucked from @pakrym's changes. Not only do they fail because of package mismatches but ReceivedFramesAreWrittenToChannel test is hanging forever.

pakrym:
As I said, `dev` would get fixed after Publish passes and we can run "Update repos" in ~40m

davidfowl:
> As I said, dev would get fixed after Publish passes and we can run "Update repos" in ~40m

Okies

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -84,15 +84,24 @@ export class DefaultHttpClient extends HttpClient {
...
-                if (request.abortSignal) {
-                    request.abortSignal.onabort = null;
-                }
-
-                if (xhr.status >= 200 && xhr.status < 300) {
-                    resolve(new HttpResponse(xhr.status, xhr.statusText, xhr.response || xhr.responseText));
-                } else {
-                    reject(new HttpError(xhr.statusText, xhr.status));
+            xhr.onreadystatechange = () => {
+                this.logger.log(LogLevel.Trace, `onreadystatechange(${xhr.readyState}:${xhr.status}:${xhr.statusText})`);
```
JamesNK: I'd like the logging to be a bit friendlier, e.g. map ready state numbers to text - https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState

Something like "HTTP request state changed. New state '${readStateDescription}'. HTTP Status: ${status}: ${statusText}".

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -106,6 +115,7 @@ export class DefaultHttpClient extends HttpClient {
                 reject(new TimeoutError());
             };
 
+            this.logger.log(LogLevel.Trace, `xhr.send(${request.content || ""})`);
```
JamesNK: Isn't there a flag now for whether request content is displayed in logs in the TS client. Should reuse that logic from the transports. e.g. actual sent content vs content length.

clients/ts/signalr/src/HttpClient.ts
davidfowl: I don't want to map it to text, because we don't compare it to text. It needs to be the raw data. The only reason I left this one in here was for our own debugging. I'd be ok removing it.

clients/ts/signalr/src/HttpClient.ts
davidfowl: This flag https://github.com/aspnet/SignalR/pull/1792?

JamesNK: Yup

davidfowl: Ok I'm removing logging from this change. I don't want to block waiting for anything.

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -84,15 +84,22 @@ export class DefaultHttpClient extends HttpClient {
...
+                if (xhr.readyState === 4) {
+                    if (request.abortSignal) {
+                        request.abortSignal.onabort = null;
+                    }
+
+                    // Some browsers report xhr.status == 0 when the
+                    // response has been cut off or there's been a TCP FIN.
+                    // Treat it like a 200 with no response.
+                    if (xhr.status === 0) {
+                        resolve(new HttpResponse(200, null, null));
```
JamesNK: Should it have an "OK" (or something else) status text?

Issue #1782, opened 2018-03-30
HubConnection.StartAsync should timeout the handshake response


davidfowl:
We currently timeout the negotiate request on the server side but we don't timeout the handshake response on the client. This means if the server took long or never sent the handshake, the client would hang forever in StartAsync (which also means StopAsync would hang see https://github.com/aspnet/SignalR/issues/1781)

anurse:
Ooooooor, we allow a CancellationToken to be passed in \*ducks\*

anurse:
Just need to add a `CancellationToken` to `StartAsync`.

BrennanConroy:
`HttpOptions` should not have a `HandshakeTimeout` as that is a Hub only feature

Currently the only "option" we have for `HubConnection` on the C# client is a public `ServerTimeout` property. Should we add another public property for the timeout?

And we are adding a cancellation token to `HubConnection.StartAsync` as part of this?

anurse:
Good point. Sounds good. And yes, also add a cancellation token.

PR #1840, opened 2018-04-03
Add Handshake timeout to C# Client

BrennanConroy:
#1782

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -512,61 +512,58 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 throw new InvalidOperationException("The server disconnected before the handshake was completed");
             }
 
-            var handshakeCts = new CancellationTokenSource();
-            var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, handshakeCts.Token);
-            linkedCts.Token.Register(s =>
-            {
-                ((PipeReader)s).CancelPendingRead();
-            }, startingConnectionState.Connection.Transport.Input);
-            handshakeCts.CancelAfter(HandshakeTimeout);
-
             try
             {
-                while (true)
+                using (var handshakeCts = new CancellationTokenSource(HandshakeTimeout))
+                using (var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, handshakeCts.Token))
                 {
-                    var result = await startingConnectionState.Connection.Transport.Input.ReadAsync();
-                    var buffer = result.Buffer;
-                    var consumed = buffer.Start;
-                    var examined = buffer.End;
-
-                    try
+                    while (true)
                     {
-                        // Read first message out of the incoming data
-                        if (!buffer.IsEmpty)
+                        var result = await startingConnectionState.Connection.Transport.Input.ReadAsync(cts.Token);
+
+                        var buffer = result.Buffer;
+                        var consumed = buffer.Start;
+                        var examined = buffer.End;
+
+                        try
                         {
-                            if (HandshakeProtocol.TryParseResponseMessage(ref buffer, out var message))
+                            // Read first message out of the incoming data
+                            if (!buffer.IsEmpty)
                             {
-                                // Adjust consumed and examined to point to the end of the handshake
-                                // response, this handles the case where invocations are sent in the same payload
-                                // as the the negotiate response.
-                                consumed = buffer.Start;
-                                examined = consumed;
-
-                                if (message.Error != null)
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -22,6 +22,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
     public partial class HubConnection
     {
         public static readonly TimeSpan DefaultServerTimeout = TimeSpan.FromSeconds(30); // Server ping rate is 15 sec, this is 2 times that.
+        public static readonly TimeSpan DefaultHandshakeTimeout = TimeSpan.FromSeconds(5);
```
davidfowl: 15

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -510,47 +514,57 @@ private void CheckDisposed()
...
+                                    break;
+                                }
+                            }
+                            else if (result.IsCompleted)
+                            {
+                                // Not enough data, and we won't be getting any more data.
+                                throw new InvalidOperationException(
+                                    "The server disconnected before sending a handshake response");
+                            }
+                            else if (result.IsCanceled)
```
davidfowl: This will never trigger.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -510,47 +514,57 @@ private void CheckDisposed()
...
+                            {
+                                throw new OperationCanceledException(
```
davidfowl: This isn't necessarily true. Since you're linking 2 tokens, you'll want to catch OperationCancelledException and then check each of the tokens to see which one was cancelled.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -510,47 +514,57 @@ private void CheckDisposed()
...
-                    var buffer = result.Buffer;
-                    var consumed = buffer.Start;
-                    var examined = buffer.End;
-
-                    try
+                    while (true)
                     {
-                        // Read first message out of the incoming data
-                        if (!buffer.IsEmpty)
+                        var result = await startingConnectionState.Connection.Transport.Input.ReadAsync(cts.Token);
```
anurse: Do we want to do this? Or should we be using `CancelPendingRead`? @davidfowl

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Also, if we're going to throw an exception because of a timeout elapsing, it should be `TimeoutException`, even if a CancellationToken is how we are triggering it.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -135,7 +138,7 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
                 try
                 {
                     Log.HubProtocol(_logger, _protocol.Name, _protocol.Version);
-                    await HandshakeAsync(startingConnectionState);
+                    await HandshakeAsync(startingConnectionState, cancellationToken);
```
davidfowl: The fact that we can't pass the token to IConnection.StartAsync is an issue but not one that is relevant to this change.

davidfowl:
@BrennanConroy are you rebased on dev?

davidfowl:
Wait, it auto merges right? When did this run? Before or after my change?

BrennanConroy:
Looks like it has @anurse 's changes for detailed messages?

davidfowl:
Yea, this shit is still broken AF.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -339,6 +340,71 @@ IConnection ConnectionFactory()
...
+            {
+                var cts = new CancellationTokenSource();
+                var connection = new TestConnection(onStart: () =>
+                {
+                    cts.Cancel();
+                    return Task.CompletedTask;
+                }, autoHandshake: false);
+                var hubConnection = CreateHubConnection(connection);
+                // We want to make sure the cancellation is because of the token passed to StartAsync
+                hubConnection.HandshakeTimeout = TimeSpan.FromSeconds(20);
```
anurse: Set it to `Timeout.InfiniteTimespan` then. Then the handshake timeout will never fire.

PR #1841, opened 2018-04-03
fix #1727 by adding polyfills for IE

### LINKED ISSUE #1727: Fix Browser Tests for Edge
A lot of failing functional tests when using Edge. We need to investigate and resolve these

edge-browsertests

anurse:
While waiting for benchmarks to run I grabbed this one off the queue.

Turns out the failures in Edge were just the flakiness we're seeing everywhere. We pass everywhere now, after some small polyfills for IE (see below).

**Preview 2 Consideration**: We _may_ want to consider taking this for Preview 2. Without it, Binary support in IE is broken due to the use of `Uint8Array.indexOf` and `Uint8Array.slice`, neither of which IE supports. This would not need to fully re-spin everything since it only affects the JS client. Also, maybe we're OK with Binary in IE being broken in Preview 2. Just wanted to raise it.

image

### DIFF
--- a/clients/ts/signalr/src/Utils.ts
+++ b/clients/ts/signalr/src/Utils.ts
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000;
-
 export class Arg {
     public static isRequired(val: any, name: string): void {
         if (val === null || val === undefined) {
--- a/clients/ts/signalr/src/browser-index.ts
+++ b/clients/ts/signalr/src/browser-index.ts
@@ -5,4 +5,20 @@
 
 import "es6-promise/dist/es6-promise.auto.js";
 
+// Copy from Array.prototype into Uint8Array to polyfill on IE. It's OK because the implementations of indexOf and slice use properties
+// that exist on Uint8Array with the same name, and JavaScript is magic.
+// We make them 'writable' because the Buffer polyfill messes with it as well.
+if (!Uint8Array.prototype.indexOf) {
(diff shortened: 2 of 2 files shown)


clients/ts/signalr/src/Utils.ts
```diff
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000;
```
anurse: @BrennanConroy ಠ_ಠ

anurse: This is product code, not test code :)

clients/ts/signalr/src/browser-index.ts
```diff
@@ -5,4 +5,20 @@
 
 import "es6-promise/dist/es6-promise.auto.js";
 
+// Copy from Array.prototype into Uint8Array to polyfill on IE. It's OK because the implementations of indexOf and slice use properties
+// that exist on Uint8Array with the same name, and JavaScript is magic.
+// We make them 'writable' because the Buffer polyfill messes with it as well.
+if (!Uint8Array.prototype.indexOf) {
```
anurse: Suggested by Mozilla as the polyfill: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice#Polyfill

And it works, so 🤷‍♂️

clients/ts/signalr/src/Utils.ts
davidfowl: LOL

JamesNK: Should it be moved somewhere else or can it just be removed like this?

anurse: @BrennanConroy just got overzealous making this change to all the test code and spilled over in to product code.

Issue #1602, opened 2018-03-14
Sending AspNetCore.Identity.Application cookie to signalr 


KennethInfoEasy:
Hi, I'm trying to connect to an asp.net core 2 signalR hub from an .net core 2.0 console application. 
I'm trying to authenticate the application by sending a cookie with the signalR StartAsync method. The cookie is returned by using httpclient.

    var connection = new HubConnectionBuilder()
                .WithUrl("http://localhost:58215/testhub")
                .WithHeader("Set-Cookie", cookiestring)
                .WithConsoleLogger()
                .Build();

Test scenario 1: Without the Authorize attribute. I'm connected to the hub but the Context.User.Identity.Name is null. 
Test scenario 2. With the Authorize attribute i'm getting an exception from signalr. 

System.AggregateException: 'One or more errors occurred. (Invalid negotiation response received.)'
FormatException: Invalid negotiation response received.
JsonReaderException: Unexpected character encountered while parsing value: <. Path '', line 0, position 0

I get this exception also if I don't send the cookie  //   .WithHeader("Set-Cookie", cookiestring)

I'm unsure if the second scenario is related to the first.  Or this is two separate issues.  

Is there another way to authenticate the users logging in remotely (console or windows service)?

 I want to keep the default authentication for the rest of the web application, so changing everything to JWT is not an option.

BrennanConroy:
Check your server and client versions and make sure they are the same

KennethInfoEasy:
![signalrdependencies](https://user-images.githubusercontent.com/37364761/37435394-707e47e8-27e3-11e8-8fe0-1af739335514.png)
I belive i'm good there. I checked that before submitting. @BrennanConroy Thanks for the quick reply though 👍

BrennanConroy:
I'm not sure setting cookies via `.WithHeader("Set-Cookie", ...)` is a good idea. You could try using `.WithMessageHandler` and providing your own wrapper that adds cookies to the httprequest. In the latest dev bits we have `.WithCookie` FYI

BrennanConroy:
Also, I was just made aware that on the request it's "Cookie" instead of "Set-Cookie"

KennethInfoEasy:
@BrennanConroy  Nice news about the .WithCookie.. Will that be release to nuget any time soon ? 
Did I mix the cookies up? :) I thought you put Set-Cookie when sending back to server. I'll try :) 

What about the exception? Since I get then exception when not setting the withheader option.

[...]

PR #1848, opened 2018-04-04
Set X-Requested-With on all requests

BrennanConroy:
#1602

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
@@ -428,6 +428,10 @@ namespace Microsoft.AspNetCore.Http.Connections.Client
                 }
             }
 
+            httpClient.DefaultRequestHeaders.Remove("X-Requested-With");
+            // Tell auth middleware to 401 instead of redirecting
+            httpClient.DefaultRequestHeaders.Add("X-Requested-With", "XMLHttpRequest");
+
             return httpClient;
         }
 
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO.Pipelines;
 using System.Linq;
 using System.Net.Http;
@@ -91,6 +92,21 @@ namespace Microsoft.AspNetCore.Http.Connections.Client
             {
                 _httpClient = CreateHttpClient();
             }
+
+            // Set the X-Requested-With header on httpOptions for WebSockets
+            if (httpOptions == null)
+            {
+                httpOptions = new HttpOptions();
+            }
+            if (httpOptions.Headers == null)
+            {
+                httpOptions.Headers = new Dictionary<string, string>();
+            }
+            else
+            {
+                httpOptions.Headers.Remove("X-Requested-With");
+            }
+            httpOptions.Headers.Add("X-Requested-With", "XMLHttpRequest");
 
             _transportFactory = new DefaultTransportFactory(transportType, _loggerFactory, _httpClient, httpOptions);
             _logScope = new ConnectionLogScope();
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.IO.Pipelines;
 using System.Linq;
 using System.Net.Http;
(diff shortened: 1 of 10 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Transport.cs
```diff
@@ -66,6 +66,46 @@ public class Transport
                 Assert.True(requestsExecuted);
             }
 
+            [Theory]
+            [InlineData(HttpTransportType.LongPolling)]
+            [InlineData(HttpTransportType.ServerSentEvents)]
```
anurse: Can you check how WebSockets behaves (as we discussed in person)? I think it may need this header too to avoid the 302.

BrennanConroy: 👍 
We're going to also set the header on httpOptions so that WebSockets gets the header set too

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -92,6 +93,21 @@ public HttpConnection(Uri url, HttpTransportType transportType, ILoggerFactory l
...
+            {
+                httpOptions = new HttpOptions();
+            }
+            if (httpOptions.Headers == null)
+            {
+                httpOptions.Headers = new Dictionary<string, string>();
+            }
+            else
+            {
+                httpOptions.Headers.Remove("X-Requested-With");
```
JamesNK: const X-Requested-With

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -92,6 +93,21 @@ public HttpConnection(Uri url, HttpTransportType transportType, ILoggerFactory l
...
+            }
+            httpOptions.Headers.Add("X-Requested-With", "XMLHttpRequest");
```
JamesNK: const XMLHttpRequest

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -428,6 +444,10 @@ private HttpClient CreateHttpClient()
                 }
             }
 
+            httpClient.DefaultRequestHeaders.Remove("X-Requested-With");
+            // Tell auth middleware to 401 instead of redirecting
+            httpClient.DefaultRequestHeaders.Add("X-Requested-With", "XMLHttpRequest");
```
JamesNK: Use consts

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -92,6 +93,21 @@ public HttpConnection(Uri url, HttpTransportType transportType, ILoggerFactory l
                 _httpClient = CreateHttpClient();
             }
 
+            // Set the X-Requested-With header on httpOptions for WebSockets
```
JamesNK: Hmm, this feels kind of messy. What about setting this header manually inside the `WebSocketsTransport`. It keeps the transport specific logic closer to the implementation.

BrennanConroy: Yeah, we were talking about that, not sure if we landed anywhere

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
```diff
@@ -82,6 +85,8 @@ public WebSocketsTransport(HttpOptions httpOptions, ILoggerFactory loggerFactory
                 _closeTimeout = httpOptions.CloseTimeout;
             }
 
+            _webSocket.Options.SetRequestHeader(XRequestedWithName, XRequestedWithValue);
```
JamesNK: Move to the top so it is possible to change the header value in `WebSocketsOptions`.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
```diff
@@ -23,6 +23,9 @@ public partial class WebSocketsTransport : ITransport
         private readonly TimeSpan _closeTimeout;
         private volatile bool _aborted;
 
+        private static readonly string XRequestedWithName = "X-Requested-With";
+        private static readonly string XRequestedWithValue = "XMLHttpRequest";
```
JamesNK: Hmm, if these aren't shared then making them consts is kind of pointless. Is there somewhere public internal to place them? If not then just have the strings inline.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
anurse: Just do it on the ClientWebSocketOptions and HttpClient directly

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Transport.cs
JamesNK: I wrote an endpoint that echos back a HTTP header for testing the user agent. Checkout WebSocketsTransportSendsUserAgent. You should be able to test web sockets passing the header with the same logic.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
BrennanConroy: I think we want to override this header

PR #1849, opened 2018-04-04
fix #1815 by disposing linked cts

### LINKED ISSUE #1815: We don't dispose the linked CancellationTokenSource on Streaming invocations
Seems like we may have a leak here:

https://github.com/aspnet/SignalR/blob/ac5dfcf9092d4919bda392e2782c0972f4389047/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs#L378-L383

We create a linked token source but then we don't keep track of it. We should dispose it properly (though I need to think about if it's really a slow leak), maybe if both tokens always fire it's fine.

anurse:
Snapshots taken immediately before a streaming invocation, then immediate after it completes. Compared memory traffic between the snapshots.

Before:

before

After:

after

Fixes #1815

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
@@ -193,7 +193,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
                     if (isStreamedInvocation)
                     {
-                        if (!TryGetStreamingEnumerator(connection, hubMethodInvocationMessage.InvocationId, descriptor, result, out var enumerator, out var cts))
+                        if (!TryGetStreamingEnumerator(connection, hubMethodInvocationMessage.InvocationId, descriptor, result, out var enumerator, out var streamCts))
                         {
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -347,39 +350,43 @@ private void InitializeHub(THub hub, HubConnectionContext connection)
             {
                 Log.NonStreamingMethodCalledWithStream(_logger, hubMethodInvocationMessage);
                 await connection.WriteAsync(CompletionMessage.WithError(hubMethodInvocationMessage.InvocationId,
-                    $"The client attempted to invoke the non-streaming '{hubMethodInvocationMessage.Target}' method in a streaming fashion."));
+                    $"The client attempted to invoke the non-streaming '{hubMethodInvocationMessage.Target}' with a streaming invocation."));
```
anurse: Decided to tidy this message up while I was here.

anurse:
Also, turns out we were leaking the subscription when streaming an IObservable, which meant a test started failing because we were no longer listening to the ConnectionAborted token (since we dispose the CTS when we finish enumerating). I've fixed that logic.

Issue #1856, opened 2018-04-05
Move JsonHubProtocol to its own package


JamesNK:
`JsonHubProtocol` is currently in `Microsoft.AspNetCore.SignalR.Common`.

`AddJsonProtocol` is currently in `Microsoft.AspNetCore.SignalR.Core`. (note that these helper methods are now also intented to be used by the client)

Consider moving both of these to a `Microsoft.AspNetCore.SignalR.Protocols.Json` package. At the very least move `AddJsonProtocol` to common so clients don't need to pull in core into their app.

PR #1861, opened 2018-04-05
Added Microsoft.AspNetCore.SignalR.Protocols.Json

davidfowl:
- Fixed package descriptions
- Call AddJsonProtocol from both the SignalR.Core (client and server)

Fixes https://github.com/aspnet/SignalR/issues/1856


JamesNK:
Should protocol tests be moved from Microsoft.AspNetCore.SignalR.Common.Tests into their own assemblies. I don't really care.

davidfowl:
> Should protocol tests be moved from Microsoft.AspNetCore.SignalR.Common.Tests into their own assemblies. I don't really care.

Yea, we can file a task to clean up the test structuring, it seems like we got lazy at some point and started throwing shit in the same assemblies.

halter73:
I'll fix the actual benchmarks repo. That comment was so nitty, I'm now embarrassed for having left it 😄

PR #1868, opened 2018-04-05
Allow benchmarks to specify runtime version

BrennanConroy:
@sebastienros @anurse

### DIFF
--- a/benchmarks/BenchmarkServer/BenchmarkServer.csproj
+++ b/benchmarks/BenchmarkServer/BenchmarkServer.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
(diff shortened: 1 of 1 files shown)


davidfowl:
https://github.com/aspnet/SignalR/blob/dev/Directory.Build.targets#L3

/cc @natemcmaster

natemcmaster:
Directory.Build.targets comes last, so you will probably still end up with MicrosoftNETCoreApp21PackageVersion, not BenchmarksRuntimeFrameworkVersion. Any reason you can't just set MicrosoftNETCoreApp21PackageVersion?

sebastienros:
That's another solution, historically there has always been these `Benchmarks[XXX]` variables, I could just use this one and the 2.0 counterpart instead, that would just work OOB with KoreBuild repos.

PR #1870, opened 2018-04-05
Comment why we set X-Requested-With

BrennanConroy:


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
```diff
@@ -82,6 +82,7 @@ public WebSocketsTransport(HttpOptions httpOptions, ILoggerFactory loggerFactory
                 _closeTimeout = httpOptions.CloseTimeout;
             }
 
+            // Set this header so the server auth middleware will set an Unauthorized instead of Redirect status code
```
davidfowl: Point to the code in security that does this

Issue #1032, opened 2017-10-18
Exceptions from the ChannelObserver's OnNext method


Tragetaschen:
In alpha2, the `ChannelObserver.OnNext` method can throw an `OperationCanceledException` and this currently hits my application in places I didn't expect. Former SignalR didn't let these connection specific exceptions enter my application. The [guidelines](https://docs.microsoft.com/en-us/dotnet/standard/events/observer-design-pattern-best-practices#the-observer----implementing-the-onerror-method) for implementing `OnNext` suggest to not throw in the interface implementations.

The code for `OnNext` has changed since to not throw OCE, but there's still an IOE that some day might pop up (when TryWrite might start to fail).

Are SignalR specific exceptions in a stream something I have to deal with on the stream level?

davidfowl:
@Tragetaschen do you want to send a PR for this one?

Tragetaschen:
[There](https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs#L101)'s quite some explanation about how channels work and could work that lower my confidence in any change I could make—especially since I haven't looked into channels yet.

Given my time budget right now and to my dismay: Pass :-/

anurse:
We should prevent exceptions from escaping [`ChannelObserver.OnNext`](https://github.com/aspnet/SignalR/blob/d23f2beafc64336918e240ff670c8c71a39ed013/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs#L50).

mikaelm12:
Hmm, yeah. We should be letting that exception propagate into the Observable. Should we just no-op then if the channel is already closed?

mikaelm12:
Done: 4ddf8664c0a8da3b14f147c46b971b73a39538e9

Issue #1864, opened 2018-04-05
Review unbounded channel usage in observable stream return scenario


davidfowl:
Today we adapt the `IObservable<T>` return type into an unbounded channel internally to write messages to the client. This means the Observer will never experience back pressure if the client backs up, memory will just grow forever. We need to consider either using a bounded channel here, or changing the approach to properly flow back pressure from the client to the observer.

https://github.com/aspnet/SignalR/blob/61c3d0c4030033013d4ad155d0f284aba384070a/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs#L17

https://github.com/aspnet/SignalR/blob/61c3d0c4030033013d4ad155d0f284aba384070a/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs#L57-L63

anurse:
The thing is, what would we do if the buffer is full? Synchronously block?

anurse:
My thought is that if you want back-pressure to prevent you from gobbling up memory, use a Channel and you can configure your buffering.

davidfowl:
> The thing is, what would we do if the buffer is full? Synchronously block?

Yes. Block hard.

anurse:
I mean, we can do that, but it's a big cannon to point at people's feet. So are unbuffered channels. Backpressure seems to be a pretty [hot](https://github.com/ReactiveX/RxJava/wiki/Backpressure) issue in [RxLand](https://github.com/Reactive-Extensions/Rx.NET/issues/19).

My concern is that `IObservable` is going to seem like the "easy" way to do streaming (since `Channel<T>` is so new) and people are going to flock to that rather than the more appropriate Channel API.

I think we need some people with more Rxperience to chime in here.

davidfowl:
We're going to remove Rx.

PR #1872, opened 2018-04-05
Don't throw from Observer's OnNext

mikaelm12:
We shouldn't be throwing from the Observer. If `OnNext` is called after the underlying channel is completed then it should just no-op.
Issue : https://github.com/aspnet/SignalR/issues/1032

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
@@ -62,8 +62,8 @@ namespace Microsoft.AspNetCore.SignalR.Internal
                     // Wait for a spot
                     if (!_output.WaitToWriteAsync(_cancellationToken).Result)
                     {
-                        // Channel was closed.
-                        throw new InvalidOperationException("Output channel was closed");
+                        // Channel was closed so we just no-op. The observer shouldn't throw.
+                        return;
                     }
                 }
             }
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
@@ -16,9 +16,9 @@ namespace Microsoft.AspNetCore.SignalR.Internal
             // TODO: Allow bounding and optimizations?
             var channel = Channel.CreateUnbounded<object>();
             var observer = new ObserverState();
-            var channelObserver = new ChannelObserver<T>(channel.Writer, cancellationToken);
+            var channelObserver = new ChannelObserver<T>(channel.Writer);
             observer.Subscription = observable.Subscribe(channelObserver);
-            observer.TokenRegistration = cancellationToken.Register(obs => { ((ChannelObserver<T>)obs).OnCompleted(); }, channelObserver);
+            observer.TokenRegistration = cancellationToken.Register(obs => ((ChannelObserver<T>)obs).OnCompleted(), channelObserver);
 
             // Make sure the subscription and token registration is disposed when enumeration is completed.
             return new AsyncEnumerator<object>(channel.Reader, cancellationToken, observer);
@@ -40,12 +40,10 @@ namespace Microsoft.AspNetCore.SignalR.Internal
         private class ChannelObserver<T> : IObserver<T>
         {
             private readonly ChannelWriter<object> _output;
-            private readonly CancellationToken _cancellationToken;
 
-            public ChannelObserver(ChannelWriter<object> output, CancellationToken cancellationToken)
+            public ChannelObserver(ChannelWriter<object> output)
             {
                 _output = output;
-                _cancellationToken = cancellationToken;
             }
 
             public void OnCompleted()
@@ -60,12 +58,6 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
             public void OnNext(T value)
             {
-                if (_cancellationToken.IsCancellationRequested)
-                {
-                    // Noop, someone else is handling the cancellation
-                    return;
-                }
-
                 // This will block the thread emitting the object if the channel is bounded and full
                 // I think this is OK, since we want to push the backpressure up. However, we may need
                 // to find a way to force the entire subscription off to a dedicated thread in order to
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -148,6 +148,66 @@ public class HubConnectionHandlerTests
...
+                var subscribeTask = Subscribe();
+
+                await waitForSubscribe.Task.OrTimeout();
+
+                Assert.Single(observable.Observers);
+
+                observable.OnNext(1);
+
+                // Calling OnComplete to complete the observer. Further calls to OnNext should no-op
+                observable.Observers[0].OnCompleted();
```
davidfowl: I think we want to kill the client connection, then trigger OnNext

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -148,6 +148,66 @@ public class HubConnectionHandlerTests
             }
         }
 
+        [Fact]
+        public async Task OberserverDoestThrowWhenOnNextIsCalledAfterChannelIsCompleted()
```
BrennanConroy: Doest

src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
```diff
@@ -62,8 +62,8 @@ public void OnNext(T value)
                     // Wait for a spot
                     if (!_output.WaitToWriteAsync(_cancellationToken).Result)
```
JamesNK: Off topic: `.Result`? I'm guessing because `OnNext` is not async. Should there be a comment explaining why and consequences?

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
mikaelm12: If we kill the client connection here then we can't verify the other call to OnNext succeeded since we wait on it later on. 
I don't think this is the prettiest but it covers the exact case that we want

davidfowl: I don’t think it’s testing the right thing. In this scenario the client disconnected and the producer is happily calling OnNext

src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
mikaelm12: Hm, that's what it seems like to me. Doesn't seem necessary but I'll add it if people want it

mikaelm12:
Updated

mikaelm12:
Rip, fixing this

src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
anurse: Yeah. Also because we use unbounded channels, this will never go async.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -148,6 +148,58 @@ public class HubConnectionHandlerTests
...
+                waitForDispose.TrySetResult(null);
+            };
+
+            using (var client = new TestClient())
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler);
+
+                async Task Subscribe()
+                {
+                    var results = await client.StreamAsync(nameof(ObservableHub.Subscribe));
```
davidfowl: nit: delete results

src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
anurse: If the token is canceled, this will throw `OperationCanceledException`, we wouldn't want that leaking out either, right? Should we just wrap the thing in a `try`/`catch`?

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -148,6 +148,58 @@ public class HubConnectionHandlerTests
...
+                }
+
+                var subscribeTask = Subscribe();
```
anurse: This can just be `var subscribeTask = client.StreamAsync(nameof(ObservableHub.Subscribe))`. We can still `await` it below and just discard the result without needing the local function.

src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
davidfowl: I spoke to @mikaelm12 , he's fixing it.

anurse: Though depending on #1864 , it may start mattering again later. We can deal with that when we resolve that issue though.

JamesNK:
Is there a test for cancelling in OnNext? 😺

src/Microsoft.AspNetCore.SignalR.Core/Internal/AsyncEnumeratorAdapters.cs
```diff
@@ -15,11 +15,26 @@ public static IAsyncEnumerator<object> FromObservable<T>(IObservable<T> observab
         {
             // TODO: Allow bounding and optimizations?
             var channel = Channel.CreateUnbounded<object>();
+            var observer = new ObserverState();
+            var channelObserver = new ChannelObserver<T>(channel.Writer, cancellationToken);
```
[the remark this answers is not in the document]
mikaelm12: Nope. This can be removed

mikaelm12:
> Is there a test for cancelling in OnNext

The Observer no longer has access to the cancellation token.

PR #1873, opened 2018-04-05
HttpTransportType refactor

JamesNK:
- Rename file to HttpTransportType.cs
- Add HttpTransportType.None
- Move All to static readonly field

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionOptions.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionOptions.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http.Connections.Internal;
 
 namespace Microsoft.AspNetCore.Http.Connections
 {
@@ -10,7 +11,7 @@ namespace Microsoft.AspNetCore.Http.Connections
     {
         public IList<IAuthorizeData> AuthorizationData { get; } = new List<IAuthorizeData>();
 
-        public HttpTransportType Transports { get; set; } = HttpTransportType.All;
+        public HttpTransportType Transports { get; set; } = HttpTransports.All;
 
         public WebSocketOptions WebSockets { get; } = new WebSocketOptions();
 
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
@@ -39,8 +39,8 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
         }
 
         [Theory]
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNetCore.Http.Connections/HttpConnectionOptions.cs
```diff
@@ -3,14 +3,15 @@
...
+using Microsoft.AspNetCore.Http.Connections.Internal;
 
 namespace Microsoft.AspNetCore.Http.Connections
 {
     public class HttpConnectionOptions
     {
         public IList<IAuthorizeData> AuthorizationData { get; } = new List<IAuthorizeData>();
 
-        public HttpTransportType Transports { get; set; } = HttpTransportType.All;
+        public HttpTransportType Transports { get; set; } = HttpTransports.All;
```
davidfowl: Pubternal types in public namespace? Naw.

JamesNK: This syntax is compiled down so that the value is set inside the `HttpConnectionOptions` ctor I thought?

JamesNK: It is but I updated class to use ctor to set defaults to make it more obvious.

src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionOptions.cs
```diff
@@ -17,8 +18,13 @@ public class HttpConnectionOptions
...
         internal CookieContainer _cookies;
 
+        public HttpConnectionOptions()
+        {
+            Transport = HttpTransports.All;
+        }
+
         public Uri Url { get; set; }
-        public HttpTransportType? Transport { get; set; }
+        public HttpTransportType Transport { get; set; }
```
JamesNK: Should this be pluralized?

anurse: Yeah, we should be consistent with the server. I like `Transports` here.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -39,8 +39,8 @@ public void ConnectionReturnsUrlUsedToStartTheConnection()
         }
 
         [Theory]
-        [InlineData((HttpTransportType)0)]
-        [InlineData(HttpTransportType.All + 1)]
+        [InlineData(HttpTransportType.None)]
+        [InlineData((HttpTransportType)int.MaxValue)]
```
anurse: I don't think this should be invalid. I know it was before, but I don't think it makes sense. In general, having all bits set in a flags enum isn't a problem. Neither do we need to check for negative numbers (in `DefaultTransportFactory` constructor) since this is a bit flag, not a number.

Issue #1642, opened 2018-03-19
Block overlapping incoming requests to the POST endpoint for the same connection


davidfowl:
We need to lock (semaphore slim) while writing to the application pipe here https://github.com/aspnet/SignalR/blob/45f4beccb9f35eb771f634bbd632b011fce005d3/src/Microsoft.AspNetCore.Sockets.Http/HttpConnectionDispatcher.cs#L455-L459. Today we're just getting lucky that clients don't make this explode.

PR #1876, opened 2018-04-05
Lock when sending data to connection

JamesNK:
https://github.com/aspnet/SignalR/issues/1642

I've confirmed that the test failed before the lock.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
@@ -450,7 +450,17 @@ namespace Microsoft.AspNetCore.Http.Connections
             }
 
             var pipeWriterStream = new PipeWriterStream(connection.Application.Output);
-            await context.Request.Body.CopyToAsync(pipeWriterStream);
+
+            await connection.Lock.WaitAsync();
+
+            try
+            {
+                await context.Request.Body.CopyToAsync(pipeWriterStream);
+            }
+            finally
+            {
+                connection.Lock.Release();
+            }
 
             Log.ReceivedBytes(_logger, pipeWriterStream.Length);
         }
--- a/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
@@ -451,6 +451,8 @@ namespace Microsoft.AspNetCore.Http.Connections
 
             var pipeWriterStream = new PipeWriterStream(connection.Application.Output);
 
+            // REVIEW: Consider spliting the connection lock into a read lock and a write lock
+            // Need to think about HttpConnection.DisposeAsync and whether one or both locks would be needed
             await connection.Lock.WaitAsync();
 
             try
--- a/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
@@ -452,7 +452,7 @@ namespace Microsoft.AspNetCore.Http.Connections
             var pipeWriterStream = new PipeWriterStream(connection.Application.Output);
 
             // REVIEW: Consider spliting the connection lock into a read lock and a write lock
-            // Need to think about HttpConnection.DisposeAsync and whether one or both locks would be needed
+            // Need to think about HttpConnectionContext.DisposeAsync and whether one or both locks would be needed
             await connection.Lock.WaitAsync();
 
             try
--- a/test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
+++ b/test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
@@ -338,6 +338,59 @@ namespace Microsoft.AspNetCore.Http.Connections.Tests
             }
         }
 
+        [Theory]
+        [InlineData(HttpTransportType.LongPolling)]
+        [InlineData(HttpTransportType.ServerSentEvents)]
+        public async Task PostSendsToConnectionInParallel(HttpTransportType transportType)
+        {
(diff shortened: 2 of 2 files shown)


test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -338,6 +338,59 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
...
+            using (StartLog(out var loggerFactory, LogLevel.Debug))
+            {
+                var manager = CreateConnectionManager(loggerFactory);
+                var dispatcher = new HttpConnectionDispatcher(manager, loggerFactory);
+                var connection = manager.CreateConnection();
+                connection.Items[ConnectionMetadataNames.Transport] = transportType;
+
+                List<Task> sendTasks = new List<Task>();
+                // This tests thread safety of sending multiple pieces of data to a connection at once
+                for (int i = 0; i < 1000; i++)
```
BrennanConroy: We try not to test race conditions this way.

If you can't find a way to block inside the lock via a mock or something then I don't think we should have a test for this

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
```diff
@@ -450,7 +450,17 @@ private static bool ServerHasWebSockets(IFeatureCollection features)
             }
 
             var pipeWriterStream = new PipeWriterStream(connection.Application.Output);
-            await context.Request.Body.CopyToAsync(pipeWriterStream);
+
+            await connection.Lock.WaitAsync();
```
davidfowl: I'm wondering if we should expose a writer lock so this doesn't overlap with the locking for polls.

davidfowl: Yes, lets do that, lets have a SendLock and ReceiveLock

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
JamesNK: Ok, I'm not a big fan of brute forcing concurrency issues either.

I'll see if I can come up with something more elegant.

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
JamesNK: What about actions like start and stop? Would they lock both SendLock and ReceiveLock?

davidfowl: Stop and Start?

davidfowl: Meh locking is fine it keeps dispose simple

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
```diff
@@ -450,7 +450,19 @@ private static bool ServerHasWebSockets(IFeatureCollection features)
             }
 
             var pipeWriterStream = new PipeWriterStream(connection.Application.Output);
-            await context.Request.Body.CopyToAsync(pipeWriterStream);
+
+            // REVIEW: Consider spliting the connection lock into a read lock and a write lock
+            // Need to think about HttpConnection.DisposeAsync and whether one or both locks would be needed
```
anurse: You mean `HttpConnectionContext.DisposeAsync` right? I'm not sure that will do anything as the code is currently structured, so we'll have to think about it a bit.

If the context is disposed, the `PipeWriter` we're writing to [will be completed](https://github.com/aspnet/SignalR/blob/b0c4e9d0f71766340a6285ab0d5c17e01542f6be/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionContext.cs#L154) and the `CopyToAsync` will start throwing exceptions. Even if we lock, that will still happen. We'd need to check some kind of "is disposed" state to prevent that. We'll need to check the `HttpConnectionContext.Status` property, which requires that current lock be taken.

Maybe it's this method that will need to take both locks? That seems unpleasant though.

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -338,6 +338,93 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
...
+            using (StartLog(out var loggerFactory, LogLevel.Debug))
+            {
+                var manager = CreateConnectionManager(loggerFactory);
+                var dispatcher = new HttpConnectionDispatcher(manager, loggerFactory);
+                var connection = manager.CreateConnection();
+                connection.Items[ConnectionMetadataNames.Transport] = transportType;
+
+                // Allow a maximum of one caller to use code at one time
+                var callerTracker = new SemaphoreSlim(1, 1);
+                var cts = new TaskCompletionSource<bool>();
```
anurse: mega-nit: `tcs`

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -338,6 +338,93 @@ public HttpConnectionDispatcherTests(ITestOutputHelper output) : base(output)
...
+                _callerTracker = callerTracker;
+                _waitTask = waitTask;
+            }
+
+            public override async Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
+            {
+                // Will return false if all available locks from semaphore are taken
+                if (!_callerTracker.Wait(0))
+                {
+                    throw new Exception("Too many callers.");
```
anurse: Devious. I love it.

JamesNK: I'm not sure whether it is clever or dumb 🤷‍♂️

PR #1878, opened 2018-04-05
0 byte read in WebSockets

BrennanConroy:
#1582 

Don't have any numbers yet

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
@@ -179,27 +179,22 @@ namespace Microsoft.AspNetCore.Http.Connections.Client.Internal
         {
             try
             {
-                var endOfMessage = true;
                 while (true)
                 {
 #if NETCOREAPP2_1
-                    // If there was a read that had a 'false' EndOfMessage then we should skip the 0 byte read since there is an in-progress frame
-                    if (endOfMessage)
+                    // Do a 0 byte read so that idle connections don't allocate a buffer when waiting for a read
+                    var result = await socket.ReceiveAsync(Memory<byte>.Empty, CancellationToken.None);
+
+                    if (result.MessageType == WebSocketMessageType.Close)
                     {
-                        // Do a 0 byte read so that idle connections don't allocate a buffer when waiting for a read
-                        var result = await socket.ReceiveAsync(Memory<byte>.Empty, CancellationToken.None);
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
```diff
@@ -181,8 +181,23 @@ public WebSocketsTransport(HttpOptions httpOptions, ILoggerFactory loggerFactory
...
+                        {
+                            throw new InvalidOperationException($"Websocket closed with error: {_webSocket.CloseStatus}.");
+                        }
 
+                        return;
+                    }
+#endif
+                    var memory = _application.Output.GetMemory();
 #if NETCOREAPP2_1
                     var receiveResult = await socket.ReceiveAsync(memory, CancellationToken.None);
```
anurse: Put a comment here that we're guaranteed not to get a close status because the zero-byte read didn't return a close frame.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/WebSocketsTransport.cs
```diff
@@ -140,6 +140,15 @@ public WebSocketsTransport(WebSocketOptions options, IDuplexPipe application, Ht
...
+#if NETCOREAPP2_1
+                    // Do a 0 byte read so that idle connections don't allocate a buffer when waiting for a read
+                    var result = await socket.ReceiveAsync(Memory<byte>.Empty, CancellationToken.None);
+
+                    if (result.MessageType == WebSocketMessageType.Close)
+                    {
+                        return;
+                    }
+#endif
                     var memory = _application.Output.GetMemory();
```
anurse: Ditto comment

test/Microsoft.AspNetCore.Http.Connections.Tests/TestWebSocketConnectionFeature.cs
```diff
@@ -106,23 +107,45 @@ public override void Dispose()
...
-                    await _input.WaitToReadAsync();
-
-                    if (_input.TryRead(out var message))
+                    if (_internalBuffer.Buffer == null || _internalBuffer.Buffer.Length == 0)
                     {
-                        if (message.MessageType == WebSocketMessageType.Close)
+                        await _input.WaitToReadAsync();
+
+                        WebSocketMessage message;
+                        if (_input.TryRead(out message))
```
anurse: y u no `out var`?

PR #1879, opened 2018-04-06
MessagePack-CSharp returns

BrennanConroy:
#1690


davidfowl:
Did we do all the things to universe for this package already?

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/MessagePackHubProtocolTests.cs
```diff
@@ -299,19 +264,15 @@ public void ParseMessages(string testDataName)
...
             var expectedMessage = new InvocationMessage(invocationId: "xyz", target: "method", argumentBindingException: null);
-            var encodedObj = Array(HubProtocolConstants.InvocationMessageType, Map(), "xyz", "method", Array(), "ex");
-            var binary = "lgGAo3h5eqZtZXRob2SQomV4";
 
             // Verify that the input binary string decodes to the expected MsgPack primitives
-            var bytes = Convert.FromBase64String(binary);
-            var obj = Unpack(bytes);
-            Assert.Equal(encodedObj, obj);
+            //Array(HubProtocolConstants.InvocationMessageType, Map(), "xyz", "method", Array(), "ex");
+            var bytes = new byte[] { 0x96, 0x01, 0x80, 0xa3, 0x78, 0x79, 0x7a, 0xa6, 0x6d, 0x65, 0x74, 0x68, 0x6f, 0x64, 0x90, 0xa2, 0x65, 0x78 };
```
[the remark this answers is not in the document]
JamesNK: For non-control characters.
For control characters create some consts and use those for readibility.

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -379,9 +379,7 @@ describe("hubConnection", () => {
                     ByteArray: protocol.name === "json"
                         ? "aGVsbG8="
                         : new Uint8Array([0x68, 0x65, 0x6c, 0x6c, 0x6f]),
-                    GUID: protocol.name === "json"
-                        ? "00010203-0405-0607-0706-050403020100"
-                        : new Uint8Array([0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x07, 0x06, 0x05, 0x04, 0x03, 0x02, 0x01, 0x00]),
```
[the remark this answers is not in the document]
JamesNK: Dates are a SOB when serializing, especially bad in .NET because `DateTime` pretends to know about time zones.

I think it would be useful to have a range of tests for serializing and deserializing DateTime with different DateTime.Kind values, and DateTimeOffset.

It's complex and fiddly and a pain in the butt, but unfortunatly important.

test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/MessagePackHubProtocolTests.cs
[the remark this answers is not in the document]
anurse: Yeah, unfortunately most of the control characters are ORs of a token and a length, for example an Array is `0x90 | length` (for lengths less than `0xF`) so they aren't exactly a good fit for constants. Helper functions could work though.

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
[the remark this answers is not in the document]
JamesNK: (and if you mess it up in the beginning you can't easily change how time zones are handled without breaking everyone in subtle and hard to diagnose ways)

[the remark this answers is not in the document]
BrennanConroy: >DateTime is serialized to MessagePack Timestamp format, it serialize/deserialize UTC and loses Kind info.

I used the native datetime formatter to solve this issue, but that brings other issues. I'm not sure how we can do datetime then...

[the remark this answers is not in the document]
anurse: Oh yeah, I do remember that. We could serialize to strings I suppose

BrennanConroy:
Changed some of the datetime behavior, didn't change a lot of the tests cause it's tedious, and we haven't 100% decided what we want to do

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -544,14 +570,118 @@ private static object DeserializeObject(Unpacker unpacker, Type type, string fie
...
+            public static readonly CustomDateTimeFormatter Instance = new CustomDateTimeFormatter();
+
+            CustomDateTimeFormatter()
+            {
+            }
+
+            public int Serialize(ref byte[] bytes, int offset, DateTime value, IFormatterResolver formatterResolver)
+            {
+                value = DateTime.SpecifyKind(value, DateTimeKind.Utc);
+                return MessagePackBinary.WriteDateTime(ref bytes, offset, value);
```
anurse: You specify the kind as `Utc` because otherwise `MessagePackBinary` will try to convert it, right? So this has the effect of writing the literal ticks value, ignoring the `Kind`?

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
```diff
@@ -544,14 +570,118 @@ private static object DeserializeObject(Unpacker unpacker, Type type, string fie
...
+            }
+
+            public DateTime Deserialize(byte[] bytes, int offset, IFormatterResolver formatterResolver, out int readSize)
+            {
+                return MessagePackBinary.ReadDateTime(bytes, offset, out readSize);
```
anurse: What's the `Kind` value for this? `Unspecified`? That's what I'm expecting.

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
BrennanConroy: It think it writes the time as is, ignoring any `Kind` value.

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
BrennanConroy: I think so, didn't check in the debugger

BrennanConroy: It reads as UTC

anurse: We should use SpecifyKind to force the kind to Unspecified then. We don't know the kind, so we should model that properly.;

src/Microsoft.AspNetCore.SignalR.Protocols.MsgPack/Internal/Protocol/MessagePackHubProtocol.cs
anurse: Yeah, that's my interpretation too.

BrennanConroy:
before
```
             Method |          Input | HubProtocol |        Mean |     Error |      StdDev |        Op/s |  Gen 0 | Allocated |
------------------- |--------------- |------------ |------------:|----------:|------------:|------------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |     MsgPack |  1,635.8 ns |  32.37 ns |    42.09 ns |   611,325.7 | 0.0057 |     696 B |
 WriteSingleMessage |   FewArguments |     MsgPack |  1,301.6 ns |  26.32 ns |    51.33 ns |   768,305.2 | 0.0038 |     472 B |
  ReadSingleMessage | LargeArguments |     MsgPack | 14,914.1 ns | 296.18 ns |   815.77 ns |    67,050.5 | 0.4578 |   41552 B |
 WriteSingleMessage | LargeArguments |     MsgPack | 25,052.0 ns | 500.19 ns | 1,217.54 ns |    39,916.9 | 0.5798 |   61848 B |
  ReadSingleMessage |  ManyArguments |     MsgPack |  3,192.3 ns |  61.46 ns |    68.31 ns |   313,251.5 | 0.0076 |    1160 B |
 WriteSingleMessage |  ManyArguments |     MsgPack |  2,489.9 ns |  49.07 ns |    70.37 ns |   401,628.6 | 0.0038 |     744 B |
  ReadSingleMessage |    NoArguments |     MsgPack |    618.8 ns |  12.32 ns |    23.43 ns | 1,616,124.4 | 0.0029 |     304 B |
 WriteSingleMessage |    NoArguments |     MsgPack |    508.2 ns |  10.80 ns |    19.48 ns | 1,967,558.0 | 0.0019 |     224 B |
```

after
```
  ReadSingleMessage |   FewArguments |     MsgPack |    700.2 ns |  14.137 ns |  27.904 ns | 1,428,147.7 | 0.0019 |     224 B |
 WriteSingleMessage |   FewArguments |     MsgPack |    827.1 ns |  13.593 ns |  12.050 ns | 1,209,106.2 |      - |      48 B |
  ReadSingleMessage | LargeArguments |     MsgPack | 14,247.5 ns | 283.278 ns | 552.511 ns |    70,187.5 | 0.2441 |   30912 B |
 WriteSingleMessage | LargeArguments |     MsgPack | 14,890.8 ns | 293.065 ns | 337.495 ns |    67,155.7 | 0.1678 |   20528 B |
  ReadSingleMessage |  ManyArguments |     MsgPack |  1,294.3 ns |  25.044 ns |  24.597 ns |   772,628.6 | 0.0038 |     400 B |
 WriteSingleMessage |  ManyArguments |     MsgPack |  1,615.4 ns |  38.943 ns |  43.285 ns |   619,041.2 |      - |      72 B |
  ReadSingleMessage |    NoArguments |     MsgPack |    360.6 ns |   7.037 ns |   8.643 ns | 2,773,000.1 | 0.0010 |     120 B |
 WriteSingleMessage |    NoArguments |     MsgPack |    538.5 ns |  10.987 ns |  20.904 ns | 1,856,853.8 |      - |      40 B |
```

Small Op/s regression for the NoArgs Write, but that's something we can look into later

neuecc:
Thank you for merged!

> Small Op/s regression for the NoArgs Write, but that's something we can look into later

I think this is caused by `MemoryBufferWriter : Stream` and `private void WriteMessageCore(HubMessage message, Stream packer)` is not good choice.
`MessagePackBinary.WriteXxx(Stream)` write to internal memory pool and copy to stream.

core api can change to
```csharp
public void WriteMessage(HubMessage message, IBufferWriter<byte> output)
{
    // concept...
    var writerOriginal = ArrayPool<byte>.Shared.Rent(4096);
    var writer = writerOriginal;
    try
    {
        var len = WriteMessageCore(message, ref writer);

        BinaryMessageFormatter.WriteLengthPrefix(writer.Length, output);
        output.Write(new ReadOnlySpan<byte>(writer, 0, len));
    }
    finally
    {
        ArrayPool<byte>.Shared.Return(writerOriginal);
    }
}
```

benaadams:
> Small Op/s regression for the NoArgs Write, but that's something we can look into later

Regained by https://github.com/aspnet/SignalR/pull/1915

davidfowl:
@neuecc we'd have to handle growing the buffer as well.

Issue #1857, opened 2018-04-05
Does bearer token expiration still work?


moozzyk:
The idea behind having accessTokenFactory was to call the factory before sending each HTTP request to obtain the current token. This way, if the user knows the current token is about to expire, they can get a new token from the server and return it from the factory to prevent from stopping the connection with an error due to an expired token. 
Today I found this:
https://github.com/aspnet/SignalR/blob/32b4d5cc6ce8a938d83d473a7e0030184b008e4a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs#L427
which, if I read the code correctly, means that the access token value will be set once when starting the connection and SignalR will not ask the user for the token again.

davidfowl:
Ah yes, I made this change. We'll take a look at this. We need to change this to be part of a message handler instead. 

Good catch! (we need a test 😉)

PR #1880, opened 2018-04-06
Fix AccessTokenFactory not being called with each request

JamesNK:
https://github.com/aspnet/SignalR/issues/1857

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/AccessTokenHttpMessageHandler.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/AccessTokenHttpMessageHandler.cs
@@ -1,0 +1,28 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/AccessTokenHttpMessageHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+        private readonly Func<string> _accessTokenFactory;
+
+        public AccessTokenHttpMessageHandler(HttpMessageHandler inner, Func<string> accessTokenFactory) : base(inner)
+        {
+            _accessTokenFactory = accessTokenFactory ?? throw new ArgumentNullException(nameof(accessTokenFactory));
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", _accessTokenFactory());
```
JamesNK: Is there a reason `Add("Authorization", $"Bearer {_httpOptions.AccessTokenFactory()}");` was being used over `Authorization = new AuthenticationHeaderValue("Bearer", _accessTokenFactory());`?

anurse: Not that I'm aware of.

Issue #1866, opened 2018-04-05
Change NegotiateProtocol to use an IBufferWriter<byte> instead of Stream


davidfowl:
https://github.com/aspnet/SignalR/blob/61c3d0c4030033013d4ad155d0f284aba384070a/src/Microsoft.AspNetCore.Http.Connections.Common/Internal/NegotiateProtocol.cs#L22

Which will lead to changing https://github.com/aspnet/SignalR/blob/61c3d0c4030033013d4ad155d0f284aba384070a/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs#L420.

In the name of performance

PR #1881, opened 2018-04-06
Serialize NegotiateResponse with IBufferWriter

JamesNK:
https://github.com/aspnet/SignalR/issues/1866

TODO: Remove old implementation


JamesNK:
```
                           Method |     Mean |     Error |    StdDev |      Op/s |  Gen 0 | Allocated |
--------------------------------- |---------:|----------:|----------:|----------:|-------:|----------:|
       WriteResponse_MemoryStream | 1.715 us | 0.0621 us | 0.0716 us | 582,941.5 | 0.0687 |    6176 B |
 WriteResponse_MemoryBufferWriter | 1.613 us | 0.0169 us | 0.0195 us | 620,031.4 | 0.0038 |     488 B |
```

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/NegotiateProtocolBenchmark.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
```
mikaelm12: nit: Header here

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
```diff
@@ -419,7 +420,16 @@ private static byte[] GetNegotiatePayload(string connectionId, HttpContext conte
             var ms = new MemoryStream();
             NegotiateProtocol.WriteResponse(response, ms);
 
-            return ms.ToArray();
+            var writer = MemoryBufferWriter.Get();
+            try
+            {
+                NegotiateProtocol.WriteResponse(response, writer);
+                return writer.ToArray();
```
JamesNK: I think this `ToArray` is unnessessary. Could pass `MemoryBufferWriter` into this method, serialize response, and then get a `ReadOnlySequence<byte>` from writer and pass that to `context.Response.Body.WriteAsync`.

Just need to add support to `MemoryBufferWriter` for getting a sequence from it.

JamesNK: Will look into this tomorrow

src/Microsoft.AspNetCore.Http.Connections.Common/Microsoft.AspNetCore.Http.Connections.Common.csproj
```diff
@@ -14,4 +14,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonPackageVersion)" />
     <PackageReference Include="System.Buffers" Version="$(SystemBuffersPackageVersion)" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.SignalR.Common\Microsoft.AspNetCore.SignalR.Common.csproj" />
```
davidfowl: Nah, this layering is wrong. We need to move the Memory stuff to Common

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
davidfowl: I have some concerns about this and the thread static caching. I don't think it'll play well together.

davidfowl: Another option would be to not use the thread static cache here but expose the `ReadOnlySequence<byte>` and write it directly response.

JamesNK:
```
                           Method |     Mean |     Error |    StdDev |      Op/s |  Gen 0 | Allocated |
--------------------------------- |---------:|----------:|----------:|----------:|-------:|----------:|
       WriteResponse_MemoryStream | 2.870 us | 0.1807 us | 0.1690 us | 348,425.5 | 0.6866 |    6176 B |
 WriteResponse_MemoryBufferWriter | 1.943 us | 0.0162 us | 0.0151 us | 514,705.6 | 0.0381 |     368 B |
```

image

src/Microsoft.AspNetCore.Http.Connections.Common/Microsoft.AspNetCore.Http.Connections.Common.csproj
davidfowl: @JamesNK are you going to move it?

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
```diff
@@ -385,17 +387,27 @@ private Task ProcessNegotiate(HttpContext context, HttpConnectionOptions options
...
+                WriteNegotiatePayload(writer, connection.ConnectionId, context, options);
 
-            // Write it out to the response with the right content length
-            context.Response.ContentLength = negotiateResponseBuffer.Length;
-            return context.Response.Body.WriteAsync(negotiateResponseBuffer, 0, negotiateResponseBuffer.Length);
+                Log.NegotiationRequest(_logger);
+
+                // Write it out to the response with the right content length
+                context.Response.ContentLength = writer.Length;
+                return writer.CopyToAsync(context.Response.Body);
```
davidfowl: You need to await here of the MemoryBufferWriter is going to be returned before the operation is complete *if* it goes async.

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
```diff
@@ -207,14 +207,15 @@ public void CloseConnections()
                 // Stop firing the timer
                 _timer?.Dispose();
 
-                var tasks = new List<Task>();
```
davidfowl: Undo, this looks like it snuck in from another change?

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
JamesNK: Good catch

src/Microsoft.AspNetCore.SignalR.Common/Internal/Protocol/HandshakeProtocol.cs
```diff
@@ -16,6 +17,22 @@ public static class HandshakeProtocol
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
+        public static readonly ReadOnlyMemory<byte> SuccessHandshakeData;
```
davidfowl: Public fields...

PR #1884, opened 2018-04-06
Remove extra class

davidfowl:
- Merged 2 DI classes in SignalR


src/Microsoft.AspNetCore.SignalR/SignalRDependencyInjectionExtensions.cs
```diff
@@ -9,6 +9,13 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public static class SignalRDependencyInjectionExtensions
     {
+        public static ISignalRServerBuilder AddHubOptions<THub>(this ISignalRServerBuilder signalrBuilder, Action<HubOptions<THub>> options) where THub : Hub
+        {
+            signalrBuilder.Services.AddSingleton<IConfigureOptions<HubOptions<THub>>, HubOptionsSetup<THub>>();
```
JamesNK: Hmm, I didn't do this with options in the `HubConnectionBuilderExtensions`.

PR #1885, opened 2018-04-06
Allocate pipe writer stream per connection

davidfowl:
- Don't create the PipeWriterStream per operation, make it per Connection
- Reduce the buffer size for CopyToAsync operations to 4K where possible instead of 81K (the default)

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/PipeReaderFactory.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/PipeReaderFactory.cs
@@ -30,9 +30,7 @@ namespace System.IO.Pipelines
             {
                 try
                 {
-                    // REVIEW: Should we use the default buffer size here?
-                    // 81920 is the default bufferSize, there is no stream.CopyToAsync overload that takes only a cancellationToken
-                    await stream.CopyToAsync(new PipeWriterStream(pipe.Writer), bufferSize: 81920, cancellationToken);
+                    await stream.CopyToAsync(new PipeWriterStream(pipe.Writer), bufferSize: 4096, cancellationToken);
                 }
                 catch (OperationCanceledException)
                 {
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
```diff
@@ -449,22 +449,25 @@ private static bool ServerHasWebSockets(IFeatureCollection features)
...
 
             try
             {
-                await context.Request.Body.CopyToAsync(pipeWriterStream);
+                await context.Request.Body.CopyToAsync(connection.ApplicationStream, bufferSize);
+
+                Log.ReceivedBytes(_logger, connection.ApplicationStream.Length);
+
+                // Clear the amount of read bytes so logging is accurate
+                connection.ApplicationStream.Reset();
```
JamesNK: Unit test

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/PipeReaderFactory.cs
```diff
@@ -30,9 +30,7 @@ public static PipeReader CreateFromStream(PipeOptions options, Stream stream, Ca
             {
                 try
                 {
-                    // REVIEW: Should we use the default buffer size here?
-                    // 81920 is the default bufferSize, there is no stream.CopyToAsync overload that takes only a cancellationToken
-                    await stream.CopyToAsync(new PipeWriterStream(pipe.Writer), bufferSize: 81920, cancellationToken);
+                    await stream.CopyToAsync(new PipeWriterStream(pipe.Writer), bufferSize: 4096, cancellationToken);
```
JamesNK: What about we write a custom method for copying from a stream directly to a pipe writer? Can avoid allocating the `PipeWriterStream` and maybe allocating the buffer. Can be done in the future, but add a comment.

JamesNK: Or a new issue

davidfowl: Not a pipe writer a pipe reader. See https://github.com/dotnet/corefx/issues/27246
. Pretty out of scope right now

PR #1888, opened 2018-04-07
Crank 2: Let's Get Crankier

anurse:
[Crank](https://github.com/SignalR/DCrank) returns! The bulk of this is a port of that code to .NET Core and ASP.NET Core, but see below for some more info.

This PR adds a new .NET Core Console app called "Crankier" to the benchmarks folder. There are three "commands" in this app:

* `local` - Runs a self-contained Agent
* `agent` - Runs an Agent that connects to a controller (not yet implemented)
* `worker` - Runs a single Worker node (launched automatically by the Agent)

The Agent spins up a series of Worker processes, communicating via stdio. Each Worker start a number of connections and send status updates to the Agent. Once all connections across all Workers have been established, the Agent instructs the Workers to send a payload to the server at a regular interval, for a specified duration. Finally, when the duration elapses, the Workers stop all their connections.

It's configured to work properly against the BenchmarkServer EchoHub but any Hub with the necessary method will work.

Other notable changes:
* It connects at the Hub level, because a) that was easier and b) it means we get the server sending us heartbeats (driving traffic that way)
* It doesn't use our coding style. I'll try to do a formatting pass this weekend.

### DIFF
--- a/benchmarks/BenchmarkServer/Hubs/EchoHub.cs
+++ b/benchmarks/BenchmarkServer/Hubs/EchoHub.cs
@@ -41,10 +41,9 @@ namespace BenchmarkServer.Hubs
             return Clients.All.SendAsync("send", time);
         }
 
-        public Task SendPayload(string payload)
+        public void SendPayload(string payload)
         {
             // Dump the payload, we don't care
-            return Task.CompletedTask;
         }
     }
 }
--- a/benchmarks/BenchmarkServer/Hubs/EchoHub.cs
+++ b/benchmarks/BenchmarkServer/Hubs/EchoHub.cs
@@ -1,7 +1,8 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
@@ -38,6 +39,12 @@ namespace BenchmarkServer.Hubs
         public Task EchoAll(DateTime time)
         {
             return Clients.All.SendAsync("send", time);
+        }
+
+        public Task SendPayload(string payload)
+        {
+            // Dump the payload, we don't care
+            return Task.CompletedTask;
         }
     }
 }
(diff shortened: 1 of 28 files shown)


benchmarks/Crankier/WorkerReceiver.cs
```diff
@@ -0,0 +1,76 @@
...
+                    }
+                }
+            });
+        }
+
+        public void Stop()
+        {
+            _receiveMessageCts.Cancel();
+        }
+
```
JamesNK: nit

benchmarks/Crankier/Worker.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        private async Task SendStatusUpdate(CancellationToken cancellationToken)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                int connectedCount = 0;
+                int connectingCount = 0;
+                int disconnectedCount = 0;
+                int reconnectingCount = 0;
+                int faultedCount = 0;
```
JamesNK: vaarrrr

benchmarks/Crankier/Worker.cs
```diff
@@ -0,0 +1,170 @@
...
+            await _agent.Pong(_processId, value);
+            Log("Worker sent pong command with value {0}.", value);
+        }
+
+        public async Task Connect(string targetAddress, HttpTransportType transportType, int numberOfConnections)
+        {
+            Log("Worker received connect command with target address {0} and number of connections {1}", targetAddress, numberOfConnections);
+
+            _targetConnectionCount += numberOfConnections;
+            for (int count = 0; count < numberOfConnections; count++)
```
JamesNK: var

JamesNK:
And copyright header everywhere.

benchmarks/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -39,5 +40,11 @@ public Task EchoAll(DateTime time)
         {
             return Clients.All.SendAsync("send", time);
         }
+
+        public Task SendPayload(string payload)
```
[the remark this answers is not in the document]
anurse: Yeah, all part of tidying up the old code :). Will do that, and `var` all the things.

anurse:
:up: :date: Got copyright headers added and style mostly in sync.

PR #1890, opened 2018-04-07
Removing native support for IObservable<T>

### LINKED ISSUE #1864: Review unbounded channel usage in observable stream return scenario

davidfowl:
- There are too many issues and questions with respect to back pressure and the buffering policy we should inact when the client being streamed to can't support the data being pushed via OnNext.
As a result, we're dropping support for IObservable but keeping ChannelReader and we'll eventually support IAsyncEnumerable when that makes it into the BCL.

The choices were:
1. Blocking thread pool threads randomly
2. Memory leaks with unbounded channels
3. Remove Rx

This is option 3

Fixes #1864


davidfowl:
Hmm, I don't see this failure locally 😞

JanEggers:
I think there needs to be some other mechanic, rather than removing a good api.

I would like that clients get disconnected that cant keep up with see server. And generate a warning log so devs may react.
removing api or slowing down all other clients with backpreasure mechanics are no benificial options i think.

davidfowl:
We're removing it. The good news is that it just means there needs to be a helper library that can convert one `IObservable<T>` to `ChannelReader<T>` (and `IAsyncEnumerable<T>` when we get it). We're just not going to be the ones to own that library or the policy for how back pressure should work (dropping items vs blocking etc).

clients/ts/FunctionalTests/ts/Utils.ts
```diff
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000;
```
anurse: Removed because....?

davidfowl: We don't need it anymore. The hangs were caused by that bug I fixed not the default 5 second timeout.

davidfowl:
Ok I added an example of adapting the Observable to a ChannelReader https://github.com/aspnet/SignalR/pull/1890/files#diff-9204e37923b3d207fe00b933a96f1404R14. This can be turned into a helper function. Note the backpressure policy (there is none).

test/Microsoft.AspNetCore.SignalR.Tests/HubEndpointTestUtils/Hubs.cs
```diff
@@ -558,9 +465,11 @@ public ChannelReader<string> BlockingStream()
...
-        public IObservable<int> ThrowStream()
+        public ChannelReader<int> ThrowStream()
         {
-            return Observable.Throw<int>(new Exception("Exception from observable"));
+            var channel = Channel.CreateUnbounded<int>();
+            channel.Writer.TryComplete(new Exception("Exception from channel"));
+            return channel.Reader;
         }
 
         private class CountingObservable : IObservable<string>
```
JamesNK: Still used? It's worth doing a search for usings of namespace or references to the package in project files that are still floating around.

samples/SignalRSamples/ObservableExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public static class ObservableExtensions
+    {
+        public static ChannelReader<T> AsChannelReader<T>(this IObservable<T> observable)
+        {
+            // This sample shows adapting an observable to a ChannelReader without 
+            // back pressure, if the connection is slower than the producer, memory will
+            // start to increase.
+
+            // If the channel is unbounded, TryWrite will return false and effectively
```
anurse: If the channel is **bounded**

anurse: Weird.

PR #1893, opened 2018-04-07
Clean up client ITransport interface

davidfowl:
- Transports own the pipe creation. The fact they are 2 sides is an implementation detail.
- The client transports now internally create the duplex pipe instead of the HttpConnection. It's just a fancy transport selector.
- Removed application and IConnection from ITransport.StartAsync, transports are now completely independent.
- Implemented IDuplexPipe on ITransport

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
@@ -9,7 +9,9 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Connections.Features;
 using Microsoft.AspNetCore.Http.Connections.Client.Internal;
+using Microsoft.AspNetCore.Http.Connections.Features;
 using Microsoft.AspNetCore.Http.Connections.Internal;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.Extensions.Logging;
@@ -30,8 +32,6 @@ namespace Microsoft.AspNetCore.Http.Connections.Client
         private bool _started;
         private bool _disposed;
 
-        private IDuplexPipe _transportPipe;
-
         private readonly HttpClient _httpClient;
         private readonly HttpOptions _httpOptions;
         private ITransport _transport;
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/ServerSentEventsTransport.cs
```diff
@@ -72,6 +74,13 @@ public ServerSentEventsTransport(HttpClient httpClient, ILoggerFactory loggerFac
                 throw;
             }
 
+            // Create the pipe pair (Application's writer is connected to Transport's reader, and vice versa)
+            var options = ClientPipeOptions.DefaultOptions;
+            var pair = DuplexPipe.CreateConnectionPair(options, options);
+
+            _transport = pair.Application;
+            _application = pair.Transport;
```
anurse: I know the names of the ones on DuplexPipe don't really matter, but can you make them line up to avoid confusion?

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
```diff
@@ -117,6 +117,13 @@ public WebSocketsTransport(HttpOptions httpOptions, ILoggerFactory loggerFactory
 
             await _webSocket.ConnectAsync(resolvedUrl, CancellationToken.None);
 
+            // Create the pipe pair (Application's writer is connected to Transport's reader, and vice versa)
+            var options = ClientPipeOptions.DefaultOptions;
+            var pair = DuplexPipe.CreateConnectionPair(options, options);
+
+            _transport = pair.Application;
+            _application = pair.Transport;
```
anurse: Ditto. Line the names up.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -331,34 +327,30 @@ private static Uri CreateConnectUrl(Uri url, string connectionId)
...
-                pair.Application.Input.Complete();
-                pair.Application.Output.Complete();
                 _transport = null;
                 throw;
             }
 
+            if (transportType == HttpTransportType.LongPolling)
+            {
+                // Disable keep alives for long polling
+                Features.Set<IConnectionInherentKeepAliveFeature>(new ConnectionInherentKeepAliveFeature(_httpClient.Timeout));
```
JamesNK: You had to move this here because a transport is no longer passed the connection, but it feels like the implementation detail of long polling is leaking into the generic HttpConnection class. Is there a better way?

davidfowl: This mirrors the server side and I think it's fine to have it here. The transport doesn't care about features, the HttpConnection layer is coupled to knowledge of specific transports and is where the concept of features is introduced. This is IMO, the right place.

PR #1895, opened 2018-04-07
The grand unification of ConnectionContext

### LINKED ISSUE #1892: Rationalize IConnection and ConnectionContext
This is the last remaining piece of work to do from https://github.com/aspnet/SignalR/issues/1453.

davidfowl:
This change rationalizes the 2 very similar abstractions that exist in Connections.Abstractions, IConnection and ConnectionContext. It also introduces an IConnectionFactory to SignalR that is used to create a new ConnectionContext for a HubConnection.

- HubConnection just completes both ends of the transport pipe instead of calling DisposeAsync.
- Implemented ConnectionContext on HttpConnection and added HttpConnectionFactory
-  Updated tests

PS: I had to skip a test (for discussion in this PR) because StopAsync is now fire and forget, we can't properly test a half completed stop anymore without adding some test hooks to HubConnection (just one test fails), not worth it IMO.

Fixes #1892


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionBuilderTests.cs
```diff
@@ -19,7 +20,7 @@ public class HubConnectionBuilderTests
         public void HubConnectionBuiderThrowsIfConnectionFactoryNotConfigured()
         {
             var ex = Assert.Throws<InvalidOperationException>(() => new HubConnectionBuilder().Build());
-            Assert.Equal("Cannot create HubConnection instance. A connection was not configured.", ex.Message);
+            Assert.Equal("Cannot create HubConnection instance. An IConnectionFactory was not configured.", ex.Message);
```
anurse: nit: `nameof` (in both places actually)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -91,16 +98,16 @@ IConnection ConnectionFactory()
                 });
             }
 
-            [Fact]
+            [Fact(Skip = "This test would require hooks into the HubConnection to function again")]
```
anurse: Because we don't `DisposeAsync` anymore, right?

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -17,7 +18,7 @@
 
 namespace Microsoft.AspNetCore.SignalR.Client.Tests
 {
-    internal class TestConnection : IConnection
+    internal class TestConnection : ConnectionContext, IConnection
```
anurse: Why is `IConnection` still around?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
davidfowl: Correct.

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
davidfowl: We pass it to the ITransport. I plan to get rid of it when I merge this https://github.com/aspnet/SignalR/pull/1893

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionBuilderTests.cs
davidfowl: I think the test should fail if the message changes.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
davidfowl: > Re-read your PR description. Should we just remove the test?

Yes, I just wanted to discuss it. I'll remove it.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -55,9 +56,15 @@ public IDuplexPipe Transport
                 }
                 return _transportPipe;
             }
+            set
+            {
+                throw new NotSupportedException("The transport pipe isn't settable");
```
JamesNK: nit: end with full stop (period for @mikaelm12 😸)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -128,8 +128,7 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
                 Log.Starting(_logger);
 
                 // Start the connection
-                var connection = _connectionFactory();
-                await connection.StartAsync(_protocol.TransferFormat);
+                var connection = await _connectionFactory.ConnectAsync(_protocol.TransferFormat);
```
JamesNK: Why isn't connection factory just creating the connection and StartAsync being called afterwards here? I'd like to understand more because my gut reaction is I'm not a fan of the factory creating and starting.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -144,7 +143,8 @@ public IDisposable On(string methodName, Type[] parameterTypes, Func<object[], o
                     Log.ErrorStartingConnection(_logger, ex);
 
                     // Can't have any invocations to cancel, we're in the lock.
-                    await startingConnectionState.Connection.DisposeAsync();
+                    startingConnectionState.Connection.Transport.Output.Complete();
+                    startingConnectionState.Connection.Transport.Input.Complete();
```
JamesNK: These can never be null?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
```diff
@@ -40,10 +41,10 @@ public HubConnection Build()
             // The service provider is disposed by the HubConnection
             var serviceProvider = Services.BuildServiceProvider();
 
-            var connectionFactory = serviceProvider.GetService<Func<IConnection>>();
+            var connectionFactory = serviceProvider.GetService<IConnectionFactory>();
             if (connectionFactory == null)
             {
-                throw new InvalidOperationException("Cannot create HubConnection instance. A connection was not configured.");
+                throw new InvalidOperationException($"Cannot create HubConnection instance. An {nameof(IConnectionFactory)} was not configured.");
```
JamesNK: If we're using nameof, also nameof(HubConnection)?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilderExtensions.cs
```diff
@@ -12,9 +13,9 @@ namespace Microsoft.AspNetCore.SignalR.Client
 {
     public static class HubConnectionBuilderExtensions
     {
-        public static IHubConnectionBuilder WithConnectionFactory(this IHubConnectionBuilder hubConnectionBuilder, Func<IConnection> connectionFactory)
+        public static IHubConnectionBuilder WithConnectionFactory(this IHubConnectionBuilder hubConnectionBuilder, Func<TransferFormat, Task<ConnectionContext>> connectionFactory)
```
JamesNK: Check this argument is not null. Will throw a NRE in the `DelegateConnectionFactory` if null.

src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public class HttpConnectionFactory : IConnectionFactory
+    {
+        private readonly HttpConnectionOptions _options;
+        private readonly ILoggerFactory _loggerFactory;
+
+        public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFactory loggerFactory)
+        {
+            _options = options.Value;
```
JamesNK: Check arguments for null.

src/Microsoft.AspNetCore.SignalR.Client/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -79,32 +73,7 @@ private static IHubConnectionBuilder WithUrlCore(this IHubConnectionBuilder hubC
...
-                Func<IConnection> createConnection = () => new HttpConnection(
-                    value.Url,
-                    value.Transports,
-                    services.GetService<ILoggerFactory>(),
-                    httpOptions);
-
-                return createConnection;
-            });
-
+            hubConnectionBuilder.Services.AddSingleton<IConnectionFactory, HttpConnectionFactory>();
```
JamesNK: 👍

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: Because we're deleting IConnection and ConnectionContext isn't getting StartAsync and StopAsync. IConnectionFactory (we can rename it if that makes you feel better) needs to both create and Start the ConnectionContext.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: No

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionBuilder.cs
davidfowl: Sounds good

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
JamesNK: Ok. I haven't got a better name than IConnectionFactory.

What do you think about ConnectAsync having IHubProtocol as the argument instead of just TransferFormat?

src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionFactory.cs
davidfowl: Just FYI, in practice these can't be null because this is activated by the DI container. It is a public class though so I'll do it anyways although I *despise* the gratuitous layers of argument checking in .NET.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
davidfowl: Too much coupling? What else would it do with the IHubProtocol?

davidfowl: I think this is the right contract.

JamesNK: Yeah. Plus IHubProtocol is pub internal. I just worry about what happens if ConnectAsync needs more information in the future but the interface limitation is TransferFormat as the only argument.

Maybe some kind of ConnectContext class that has TransferFormat as a property. Or am I worrying too much?

davidfowl: First, I understand the concern now. But lets try to think of something realistic we'd want to add here first. It would need to be something that couldn't be communicated via the construction of the factory itself.  

A context doesn't seem unreasonable.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
JamesNK: Move this cleanup stuff to its own method. Reuse 664-665.

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
davidfowl: Am removing it now

davidfowl:
> Plus finish removing IConnection https://github.com/aspnet/SignalR/pull/1895/files#r179925453

Done!

PR #1900, opened 2018-04-08
Use in for ReadOnlySequence

davidfowl:
It's a big enough struct that there *could* be a benefit (I didn't run any perf tests though...)

### DIFF
--- a/samples/ClientSample/Tcp/SocketSender.cs
+++ b/samples/ClientSample/Tcp/SocketSender.cs
@@ -25,7 +25,7 @@ namespace ClientSample
             _eventArgs.Completed += (_, e) => ((SocketAwaitable)e.UserToken).Complete(e.BytesTransferred, e.SocketError);
         }
 
-        public SocketAwaitable SendAsync(ReadOnlySequence<byte> buffers)
+        public SocketAwaitable SendAsync(in ReadOnlySequence<byte> buffers)
         {
             if (buffers.IsSingleSegment)
             {
@@ -74,7 +74,7 @@ namespace ClientSample
             return _awaitable;
         }
 
-        private List<ArraySegment<byte>> GetBufferList(ReadOnlySequence<byte> buffer)
+        private List<ArraySegment<byte>> GetBufferList(in ReadOnlySequence<byte> buffer)
         {
             Debug.Assert(!buffer.IsEmpty);
             Debug.Assert(!buffer.IsSingleSegment);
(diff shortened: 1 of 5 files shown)


lahma:
I've tested `in` to benefit even [something as small as nullable bool](https://gist.github.com/lahma/3649ef6c800c4688806f6dfeb61bedc0).  Do you have generic guidance like for struct vs class where it's based on sum of field sizes?

davidfowl:
> I've tested in to benefit even something as small as nullable bool. Do you have generic guidance like for struct vs class where it's based on sum of field sizes?

Wow those are huge gains. Yes, we had a discussion a while back on a corefx thread that I can't find now when to use ref. I believe it was if the size of the struct was > 16 bytes? @AndyAyersMS @pakrym @benaadams

pakrym:
Yep, 2x pointer size was the conclusion. ROS is 24 bytes though.

benaadams:
> Yep, 2x pointer size was the conclusion

Structs up to 2x pointer size *could* be passed by reg (on Unix x64, perhaps) https://github.com/dotnet/corefx/issues/27238#issuecomment-366546338 and in future the Jit *may* reduce the copies that happen when pass by value (but currently doesn't [Eliminate unecessary copies when passing structs](https://github.com/dotnet/coreclr/issues/16619))

> I've tested in to benefit even something as small as nullable bool. Do you have generic guidance like for struct vs class where it's based on sum of field sizes?

Needs to be a `readonly struct` type (or have no properties/methods, only fields); otherwise you are just pushing the copy into the method (and can cause multiple copies).

**In a public api** to future proof (changing from `in` to byval would be a [breaking change](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md)) 

If the struct has references; 2x pointer size is probably a good limit. Pointers always need to be decomposed to a register (GC reporting) and 2 registers is the current max byval being looked at.

There is also a vector calling convention (which isn't currently used); so for stucts without any references at some point in the future larger sizes could also be passed by register. The [size candidates](https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/first-class-structs.md#normalizing-struct-types) in bytes are 1, 2, 4, 8, 12, 16, 32 (if cpu supported). So if you had an 18 byte struct, that wouldn't be a candidate for register passing.

**In a private api**, up to you, but be prepared to change it when the Jit becomes more capable; though larger than a single pointer would currently benefit. 

>  ROS is 24 bytes though.

On x64 its at least 28 bytes due to alignment of the pointers
```
[ReadOnlySequence                        ]
[SequencePosition][pad ][SequencePosition]
[[ptr     ][int ]][pad ][[ptr     ][int ]]
```

lahma:
@benaadams thank you for taking the time and explaining in such detail, something I can come back to for reference.

lahma:
So now someone needs to write these rules as Roslyn/R# analyzer. These are not the easiest to keep in mind when so many moving parts. Thanks again :)

AndyAyersMS:
I think Ben's advice is reasonable. Generally, pass small structs by value and larger ones by reference.

Also the jit's codegen is still quirky in many cases and can make inferring general principles from measurements challenging. Opened dotnet/coreclr#17473 with some notes on the jit codegen for your example.

PR #1906, opened 2018-04-08
Use expression body properties and implicit arrays where possible

JamesNK:


src/Microsoft.AspNetCore.SignalR.Core/DynamicHub.cs
```diff
@@ -9,7 +9,7 @@ public class DynamicHub : Hub
 
         public new DynamicHubClients Clients
         {
-            get { return _clients ?? new DynamicHubClients(base.Clients); }
+            get => _clients ?? new DynamicHubClients(base.Clients);
```
JamesNK: Should this return a new `DynamicHubClients` every time?

src/Common/ForceAsyncAwaiter.cs
```diff
@@ -34,7 +34,8 @@ internal struct ForceAsyncAwaiter : ICriticalNotifyCompletion
 
         public ForceAsyncAwaiter GetAwaiter() { return this; }
 
-        public bool IsCompleted { get { return false; } } // the purpose of this type is to always force a continuation
+        public bool IsCompleted // the purpose of this type is to always force a continuation
```
davidfowl: Fix this formatting so the comment is above the property and the => false is on the same line

src/Microsoft.AspNetCore.SignalR.Core/DynamicHub.cs
davidfowl: No I think it should look like this https://github.com/aspnet/SignalR/blob/5b6e81387aa2a301149d9c827273379ee617d2cd/src/Microsoft.AspNetCore.SignalR.Core/Hub%60T.cs#L10-L20

JamesNK: Ok, thought so.

PR #1911, opened 2018-04-09
Support async access token factory

veloek:
Access tokens may be stored in such a way that an async method is
necessary to retrieve them.


davidfowl:
@veloek this looks like a good change in general but it looks like you broke something and you also need to make the change for the C# client.

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -87,7 +87,7 @@ export class HttpConnection implements IConnection {
                 // No fallback or negotiate in this case.
                 await this.transport.connect(this.url, transferFormat, this);
             } else {
-                const token = this.options.accessTokenFactory();
+                const token = await this.options.accessTokenFactory();
```
JamesNK: What happens if you await a string?

davidfowl: Nothing good I'd imagine.

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -13,7 +13,7 @@ export interface IHttpConnectionOptions {
     httpClient?: HttpClient;
     transport?: TransportType | ITransport;
     logger?: ILogger | LogLevel;
-    accessTokenFactory?: () => string;
+    accessTokenFactory?: () => string|Promise<string>;
```
davidfowl: Can't this just be `Promise<string>`?

clients/ts/signalr/src/HttpConnection.ts
veloek: I tested returning both promise and string and it worked just the same.

clients/ts/signalr/src/HttpConnection.ts
veloek: That would be a breaking change though, right? Keeping it optional seems better to me.

davidfowl: There are no breaking changes in the until this library ships. We make breaking changes between previews all the time (in fact we already have a bunch).

clients/ts/signalr/src/HttpConnection.ts
veloek: From the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await): "If the value of the expression following the await operator is not a Promise, it's converted to a resolved Promise."

davidfowl:
Run build.cmd or build.sh

clients/ts/signalr/src/HttpConnection.ts
anurse: JavaScript is weird. But hey, if it works :)

clients/ts/signalr/src/HttpConnection.ts
anurse: I think it should stay `string | Promise<string>`. We use the `|` syntax all over the place for convenience features like this, and TypeScript lets you await a literal value. It should be spaced out though, like the others.

veloek:
Thanks guys, got everything building and the tests are green now.
A bit unsure about the C# client though. Should I change the `AccessTokenFactory` to be of type `Func<Task<string>>` or should it somehow be optional to have an async factory, like in the TypeScript client?

anurse:
I'm OK with making that one just `Task<string>`. C# doesn't have the same flexibility here ;).

JamesNK:
> I'm OK with making that one just Task<string>. C# doesn't have the same flexibility here ;).

I think it needs to `Func<Task<string>>`, otherwise it can't change once set.

veloek:
CI build is getting a timeout exception, but I can't reproduce it locally.

davidfowl:
Don't worry about that particular exception, I filed this https://github.com/aspnet/SignalR/issues/1927

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -708,7 +708,7 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
             {
                 var httpResponse = await new HttpClient().GetAsync(_serverFixture.Url + "/generateJwtToken");
                 httpResponse.EnsureSuccessStatusCode();
-                var token = await httpResponse.Content.ReadAsStringAsync();
+                var token = httpResponse.Content.ReadAsStringAsync();
```
davidfowl: What happened here?

veloek: Well, the factory is returning the task it self, so it's not necessary to await it here.

davidfowl: I see, can we change this test to move that logic all into the AccessTokenFactory. That'll be new coverage since everything else we have is pretty static.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -706,15 +706,18 @@ public HubConnectionTests(ServerFixture<Startup> serverFixture, ITestOutputHelpe
         {
             using (StartLog(out var loggerFactory, $"{nameof(ClientCanUseJwtBearerTokenForAuthentication)}_{transportType}"))
             {
-                var httpResponse = await new HttpClient().GetAsync(_serverFixture.Url + "/generateJwtToken");
-                httpResponse.EnsureSuccessStatusCode();
-                var token = await httpResponse.Content.ReadAsStringAsync();
+                async Task<string> accessTokenFactory()
```
davidfowl: tiny nit: a in accessTokenFactory should be upper case.

davidfowl:
One tiny nit

clients/ts/signalr/src/Transports.ts
```diff
@@ -52,9 +52,9 @@ export class WebSocketTransport implements ITransport {
 
         this.logger.log(LogLevel.Trace, "(WebSockets transport) Connecting");
 
+        const token = await this.accessTokenFactory();
         return new Promise<void>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
-            const token = this.accessTokenFactory();
```
davidfowl: I just realized we had a bug here where if this threw an exception we'd never call reject on the promise.

davidfowl:
Solid PR @veloek !

veloek:
Thanks! Always exiting to contribute to a new project and this was a very pleasant experience :)

PR #1913, opened 2018-04-09
Remove unused JsonSerializer

davidfowl:


anurse:
Nice catch, my bad :)

Issue #1865, opened 2018-04-05
Optimize ServerSentEventsMessageFormatter


davidfowl:
Now that we have some benchmarks for SSE (formatting and parsing) we can do some optimizations:

These are some ideas

- Use the MemoryPoolBufferWriter instead of allocating a memory stream every time and change from using Stream to using that.
- Don't convert the entire payload into an array if if's multi-segmented add support for handling `ReadOnlySequence<byte>` directly
- Use in for `ReadOnlySequence<byte>` argument.

This idea needs to be tested but it might not work because we don't want to write partial responses to the javascript client:

- Don't allocate the memory stream unless the payload contains \n. If it doesn't we can just write the buffers directly
 to the stream without modification. While it does mean that there will be smaller writes, should be fine for the most part
 since we're using reasonably sized buffers.

PR #1916, opened 2018-04-09
SSE formatting refactor

JamesNK:
https://github.com/aspnet/SignalR/issues/1865


src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
```diff
@@ -27,37 +27,15 @@ public static class ServerSentEventsMessageFormatter
...
-            {
-                WriteMessage(payload.First, ms);
-            }
-            else
-            {
-                WriteMessage(payload.ToArray(), ms);
-            }
+            WriteMessage(payload, ms);
 
             ms.Position = 0;
```
JamesNK: Why is a MemoryStream written to instead of writing directly to the response output stream? Is it because writing to the response output stream should use `WriteAsync` and didn't want the overhead of async while parsing?

davidfowl: We can’t write partial messages to the JavaScript client. It can handle batches but not partial messages

davidfowl: We could use our new more efficient memory stream. Orrr we could port pipelines to JavaScript 😬

davidfowl: Yes also writing directly from the parsing loop would require it to be async.

davidfowl: I actually think we should do a quick check for a new line. If there’s none then skip this whole thing

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
```diff
@@ -69,56 +47,130 @@ private static void WritePayload(ReadOnlyMemory<byte> payload, Stream output)
...
-                    keepWriting = false;
-                }
-                if (sliceEnd > 0 && span[sliceEnd - 1] == '\r')
-                {
-                    sliceEnd--;
-                }
 
-                var slice = payload.Slice(0, sliceEnd);
+            if (sequence.Length > 0)
+            {
```
davidfowl: I think we need to use the BufferReader for this

davidfowl: It’s in Common

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
JamesNK: Even if there was no new line, you would still need to write "data: " and "\r\n" with the content in one call.

davidfowl: No that’s not a problem. It’s partial frames within the SSE frames. So you can write data first then the payload then \r\n. The browser (EventSource api) buffers until you get an entire frame.

davidfowl: This is a problem for example:


data: { “target”: “Metho\r\n
data: d”, “args”: []\r\n
data: }\r\n

JamesNK: Will that ever be a problem? JSON doesn't have raw \r\n in its content. It is always escaped when written inside a string.

As long as you aren't send pretty printed JSON (e.g. indented with new lines) then you'll never have that issue.

davidfowl: Exactly why old SignalR worked because it was only json and we never had this logic. That’s why I want the fast path. The reason this logic exists here is because this transport layer doesn’t know we’re writing JSON.

anurse: Still, I don't really like making this assumption in the transport if we can avoid it.

davidfowl: Nah, we shouldn't I think we should do a PositionOf and if that returns null. Skip everything.

JamesNK:
Unable to use BufferReader.cs because access to Memory<byte> is required.

🆙 📅

JamesNK:
After
``` 
             Method |          Input |       Mean |      Error |      StdDev |        Op/s |  Gen 0 | Allocated |
------------------- |--------------- |-----------:|-----------:|------------:|------------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |   404.9 ns |  4.1147 ns |   3.8489 ns | 2,469,477.2 | 0.0086 |      80 B |
 WriteSingleMessage |   FewArguments |   141.7 ns |  1.4626 ns |   1.3681 ns | 7,057,242.1 |      - |       0 B |
  ReadSingleMessage | LargeArguments | 4,408.6 ns | 86.8975 ns | 188.9079 ns |   226,826.9 | 1.8311 |   20560 B |
 WriteSingleMessage | LargeArguments |   697.3 ns |  3.0253 ns |   2.5263 ns | 1,434,152.2 |      - |       0 B |
  ReadSingleMessage |  ManyArguments |   399.7 ns |  1.1366 ns |   0.9491 ns | 2,501,599.6 | 0.0138 |     128 B |
 WriteSingleMessage |  ManyArguments |   150.2 ns |  1.6419 ns |   1.3711 ns | 6,658,031.5 |      - |       0 B |
  ReadSingleMessage |    NoArguments |   400.3 ns |  1.9463 ns |   1.6253 ns | 2,498,085.9 | 0.0076 |      72 B |
 WriteSingleMessage |    NoArguments |   139.4 ns |  0.6454 ns |   0.5721 ns | 7,175,938.3 |      - |       0 B |
```

Before
```
             Method |          Input |        Mean |      Error |      StdDev |      Median |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |------------:|-----------:|------------:|------------:|------------:|-------:|-------:|----------:|
  ReadSingleMessage |   FewArguments |    394.2 ns |   3.951 ns |   3.2995 ns |    393.3 ns | 2,536,667.5 | 0.0086 |      - |      80 B |
 WriteSingleMessage |   FewArguments |    239.3 ns |   5.044 ns |   5.1801 ns |    239.1 ns | 4,179,242.9 | 0.0391 |      - |     352 B |
  ReadSingleMessage | LargeArguments |  4,577.5 ns | 100.078 ns | 133.6007 ns |  4,564.4 ns |   218,458.5 | 1.8311 |      - |   20560 B |
 WriteSingleMessage | LargeArguments | 11,282.8 ns | 310.353 ns | 828.3943 ns | 10,967.3 ns |    88,630.4 | 4.1199 | 0.0763 |   62008 B |
  ReadSingleMessage |  ManyArguments |    414.4 ns |   1.056 ns |   0.9365 ns |    414.4 ns | 2,412,840.6 | 0.0138 |      - |     128 B |
 WriteSingleMessage |  ManyArguments |    241.5 ns |   4.774 ns |   4.2320 ns |    243.1 ns | 4,141,111.0 | 0.0391 |      - |     352 B |
  ReadSingleMessage |    NoArguments |    403.5 ns |   3.907 ns |   3.2623 ns |    402.8 ns | 2,478,360.1 | 0.0076 |      - |      72 B |
 WriteSingleMessage |    NoArguments |    239.4 ns |   5.578 ns |   7.6351 ns |    237.5 ns | 4,177,410.4 | 0.0391 |      - |     352 B |
```

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
```diff
@@ -69,56 +59,131 @@ private static void WritePayload(ReadOnlyMemory<byte> payload, Stream output)
             // end up with an empty payload buffer, BUT we need to write it as an empty 'data:' line, so we need
             // to use a condition that ensure the only time we stop writing is when we write the slice after the final
             // newline.
-            var keepWriting = true;
```
davidfowl: Can we uber simplify this code and since it's now the slow path, just do a very simple PositionOf and Write?

JamesNK:
Before simplifying slow path (I added msgpack with a line break byte in the content to trigger the slow path)

```
             Method |          Input | Protocol |        Mean |       Error |      StdDev |      Median |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |--------- |------------:|------------:|------------:|------------:|------------:|-------:|-------:|----------:|
 WriteSingleMessage |   FewArguments |  msgpack |    361.0 ns |   7.2450 ns |  17.6354 ns |    355.9 ns | 2,770,143.0 | 0.0391 |      - |     352 B |
 WriteSingleMessage | LargeArguments |  msgpack | 11,413.7 ns | 330.1178 ns | 903.6921 ns | 11,174.7 ns |    87,613.7 | 4.1199 | 0.0458 |   61952 B |
 WriteSingleMessage |  ManyArguments |  msgpack |    367.0 ns |   7.6908 ns |  18.4266 ns |    359.3 ns | 2,724,762.8 | 0.0391 |      - |     352 B |
 WriteSingleMessage |    NoArguments |  msgpack |    353.1 ns |   7.3014 ns |  13.1660 ns |    349.8 ns | 2,832,150.1 | 0.0386 |      - |     352 B |
```

After simplifying slow path. It is about 50% slower
```
             Method |          Input | Protocol |        Mean |       Error |        StdDev |      Median |        Op/s |  Gen 0 |  Gen 1 | Allocated |
------------------- |--------------- |--------- |------------:|------------:|--------------:|------------:|------------:|-------:|-------:|----------:|
 WriteSingleMessage |   FewArguments |  msgpack |    757.3 ns |  12.7105 ns |    11.8894 ns |    757.7 ns | 1,320,521.8 | 0.0391 |      - |     352 B |
 WriteSingleMessage | LargeArguments |  msgpack | 15,035.6 ns | 296.0861 ns |   570.4574 ns | 14,765.3 ns |    66,508.9 | 4.1199 | 0.0305 |   61952 B |
 WriteSingleMessage |  ManyArguments |  msgpack |    737.3 ns |  16.3253 ns |    14.4720 ns |    735.4 ns | 1,356,341.9 | 0.0391 |      - |     352 B |
 WriteSingleMessage |    NoArguments |  msgpack |    714.9 ns |  25.0704 ns |    23.4509 ns |    703.4 ns | 1,398,884.5 | 0.0391 |      - |     352 B |
```

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
```diff
@@ -12,113 +12,101 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
...
-            //   data: \r\n
-            //   \r\n
-            // Since we slice past the newline when we find it, after writing "Hello" in the previous example, we'll
-            // end up with an empty payload buffer, BUT we need to write it as an empty 'data:' line, so we need
-            // to use a condition that ensure the only time we stop writing is when we write the slice after the final
-            // newline.
+            throw new InvalidOperationException("Could not get last segment from sequence.");
+        }
+
+        private static async Task WriteMessageToMemory(Stream output, ReadOnlySequence<byte> payload)
```
davidfowl: Make this synchronous and make the type MemoryStream.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
```diff
@@ -12,113 +12,101 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
...
-                    sliceEnd = payload.Length;
-                    nextSliceStart = sliceEnd + 1;
+                var sliceEnd = payload.PositionOf(LineFeed) ?? payload.End;
 
-                    // This is the last span
-                    keepWriting = false;
-                }
-                if (sliceEnd > 0 && span[sliceEnd - 1] == '\r')
+                ReadOnlySequence<byte> lineSegment;
+                if (sliceEnd == payload.End)
```
davidfowl: I don't think this comparison is safe. SequencePosition(s) can't be compared.

/cc @pakrym

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
```diff
@@ -12,113 +12,101 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
...
+                    lineSegment = payload;
+                    payload = ReadOnlySequence<byte>.Empty;
+                    keepWriting = false;
                 }
                 else
                 {
-                    payload = payload.Slice(nextSliceStart);
+                    lineSegment = payload.Slice(payload.Start, sliceEnd);
+
+                    if (sliceEnd != payload.Start)
```
davidfowl: Same here.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
```diff
@@ -12,113 +12,101 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
...
+                    {
+                        // Check if the line ended in \r\n. If it did then trim the \r
+                        var memory = GetLastSegment(in lineSegment, out var offset);
```
davidfowl: Why are you specifying in here? It's implicit and doesn't need to be specified at the call site.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionStoreTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        public void ConnectionIdUsesOrdinalComparison()
+        {
+            CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
+
+            var s1 = "Strasse";
+            var s2 = "Straße";
+
+            bool sdf = s1.Equals(s2, StringComparison.Ordinal);           //false
+
+            bool sdfdf = s1.Equals(s2, StringComparison.InvariantCulture);  //true
```
anurse: Doing a little culture spelunking?

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/ServerSentEventsBenchmark.cs
```diff
@@ -12,36 +13,51 @@ public class ServerSentEventsBenchmark
     {
         private ServerSentEventsMessageParser _parser;
         private byte[] _sseFormattedData;
-        private byte[] _rawData;
+        private ReadOnlySequence<byte> _rawData;
 
         [Params(Message.NoArguments, Message.FewArguments, Message.ManyArguments, Message.LargeArguments)]
         public Message Input { get; set; }
 
+        [Params("json", "msgpack")]
```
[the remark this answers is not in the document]
JamesNK: msgpack was to test new lines bytes. I'll change it to be formatted JSON to make more sense.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
JamesNK: Ah, VS squiggles it without the in but it still compiles. I'll remove.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
pakrym: Yeah, we need to remove those operators or at lease comments, it's not guaranteed that comparison would be correct.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
pakrym: https://github.com/dotnet/corefx/issues/29001

JamesNK: Can you comment on how to achieve this logic without comparing SequencePositions? https://github.com/aspnet/SignalR/blob/1ca2e5f2638f498bb4b57a9d2c69084ecf99bc8f/src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs#L71-L111

pakrym: You can replace first `sliceEnd == payload.End` with `sliceEnd.HasValue`.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
anurse: You can check the length of `lineSegment` here rather than comparing the `SequencePosition`

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs
anurse: Here you can just check if `sliceEnd` is `null` rather than using `?? payload.End`

JamesNK:
https://github.com/aspnet/SignalR/blob/bf60197a2f06fb22c8b0eeb0282cb83cfc9977df/src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs#L45-L111

@pakrym Better? ReadOnlySequence is challenging to work with without documentation/samples.

pakrym:
Yep, it's better. But I think entire thing could be written to just got through ReadOnlyMemory instances and do IndexOf on them.

JamesNK:
> Yep, it's better. But I think entire thing could be written to just got through ReadOnlyMemory instances and do IndexOf on them.

Like this? https://github.com/aspnet/SignalR/blob/c2636f4eb0912a118a2de20c5c9900407a6211d4/src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/ServerSentEventsMessageFormatter.cs

anurse:
That loop seems a little twistier than it needs to be. I think it could just be two simple nested loops, one over Segments using `TryGet` to get a segment `memory` and an inner one while `memory.Length > 0` (to write all the data from that segment). The logic should be able to live inside the inner loop, as long as the various state flags (trailing CR, at a new line, etc.) are tracked outside the outer loop. Looking at your most recent suggestion, it looks like you've got all the state tracking, but I think it could be a little simpler.

I'm also OK leaving it as you had [here](https://github.com/aspnet/SignalR/pull/1916#issuecomment-380277022) though. I think simpler is better unless the performance is significantly different.

Life with RoS<T> right now is unfortunately complicated because it's changing frequently and isn't documented yet :(. I'd rather we get this simple and right than try to hyper-optimize, unless we know it's a major gain.

JamesNK:
Simpler one is about 50% slower https://github.com/aspnet/SignalR/pull/1916#issuecomment-380002836

I'm kind of over looking at it tbh.

JamesNK:
I'm going to merge what we have. Improvements can be made post 2.1.

PR #1921, opened 2018-04-09
Split HandshakeProtocol into another class

davidfowl:
Moved the handshake protocol handling out of HubConnection

### DIFF
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -178,6 +178,11 @@ export class HubConnection {
     }
 
     public async start(): Promise<void> {
+        const handshakeRequest: HandshakeRequestMessage = {
+            protocol: this.protocol.name,
+            version: this.protocol.version,
+        };
+
         this.logger.log(LogLevel.Trace, "Starting HubConnection.");
 
         this.receivedHandshakeResponse = false;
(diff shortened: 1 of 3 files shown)


clients/ts/signalr/src/HubConnection.ts
```diff
@@ -106,39 +109,12 @@ export class HubConnection {
...
-                const responseLength = separatorIndex + 1;
-                messageData = textData.substring(0, responseLength);
-                remainingData = (textData.length > responseLength) ? textData.substring(responseLength) : null;
-            }
-
-            // At this point we should have just the single handshake message
-            const messages = TextMessageFormat.parse(messageData);
-            responseMessage = JSON.parse(messages[0]);
+        try {
+            const result = this.handshakeProtocol.processHandshakeResponse(data);
```
anurse: Use destructuring syntax:

```
[remainingData, responseMessage] = this.handshakeProtocol.processHandshakeResponse(data);
```

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -41,6 +43,7 @@ export class HubConnection {
         this.timeoutInMilliseconds = options.timeoutInMilliseconds || DEFAULT_TIMEOUT_IN_MS;
 
         this.protocol = options.protocol || new JsonHubProtocol();
+        this.handshakeProtocol = new HandshakeProtocol();
```
JamesNK: IMO just have `HandshakeProtocol` expose static methods.

clients/ts/signalr/src/HubConnection.ts
davidfowl: Ohhhhh

Issue #1863, opened 2018-04-05
Connect to redis asynchronously


davidfowl:
Today we block a thread while trying to connect to redis, this would only be beneficial if we did it at startup and it caused startup to fail (or maybe that's a bad thing). In reality, we may end up blocking a thread pool thread sometime layer down the road which would lead to starvation. We should change this to be asynchronous.

See https://github.com/aspnet/SignalR/blob/61c3d0c4030033013d4ad155d0f284aba384070a/src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs#L21

mikaelm12:
Done here: e2169ceda636797ebe76ba1a26abd0ae59a0802e

PR #1922, opened 2018-04-09
Connect to Redis async

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1863

### DIFF
--- a/samples/ChatSample/RedisUserTracker.cs
+++ b/samples/ChatSample/RedisUserTracker.cs
@@ -28,14 +28,14 @@ namespace ChatSample
         private const int ScanInterval = 5; //seconds
         private const int ServerInactivityTimeout = 30; // seconds
 
-        private readonly IConnectionMultiplexer _redisConnection;
-        private readonly IDatabase _redisDatabase;
-        private readonly ISubscriber _redisSubscriber;
+        private IConnectionMultiplexer _redisConnection;
+        private IDatabase _redisDatabase;
+        private ISubscriber _redisSubscriber;
 
         private const string UserAddedChannelName = "UserAdded";
         private const string UserRemovedChannelName = "UserRemoved";
-        private readonly RedisChannel _userAddedChannel;
-        private readonly RedisChannel _userRemovedChannel;
+        private RedisChannel _userAddedChannel;
+        private RedisChannel _userRemovedChannel;
 
         private readonly ILogger _logger;
 
@@ -44,8 +44,9 @@ namespace ChatSample
         private HashSet<UserDetails> _users;
         private readonly object _lockObj = new object();
         private readonly SemaphoreSlim _userSyncSempaphore = new SemaphoreSlim(initialCount: 1);
+        private readonly RedisOptions _options;
 
-        private readonly Timer _timer;
+        private Timer _timer;
 
         public event Action<UserDetails[]> UsersJoined;
         public event Action<UserDetails[]> UsersLeft;
@@ -57,7 +58,18 @@ namespace ChatSample
             _users = new HashSet<UserDetails>(_userEqualityComparer);
 
             _logger = loggerFactory.CreateLogger<RedisUserTracker<THub>>();
-            (_redisConnection, _redisDatabase) = StartRedisConnection(options.Value);
+            _options = options.Value;
+        }
+
+        private async Task EstablishRedisConnection()
+        {
+            // TODO: handle connection failures
+            _redisConnection = await ConnectToRedis(_options, _logger);
+            _redisDatabase = _redisConnection.GetDatabase(_options.Options.DefaultDatabase.GetValueOrDefault());
+
+            // Register connection
+            _redisDatabase.SetAdd(ServerIndexRedisKey, ServerId);
+            _redisDatabase.StringSet(LastSeenRedisKey, DateTimeOffset.UtcNow.Ticks);
 
             _timer = new Timer(Scan, this, TimeSpan.FromMilliseconds(0), TimeSpan.FromSeconds(ScanInterval));
 
@@ -88,30 +100,17 @@ namespace ChatSample
             });
         }
 
-        private (IConnectionMultiplexer, IDatabase) StartRedisConnection(RedisOptions options)
-        {
-            // TODO: handle connection failures
-            var redisConnection = ConnectToRedis(options, _logger);
-            var redisDatabase = redisConnection.GetDatabase(options.Options.DefaultDatabase.GetValueOrDefault());
-
-            // Register connection
-            redisDatabase.SetAdd(ServerIndexRedisKey, ServerId);
-            redisDatabase.StringSet(LastSeenRedisKey, DateTimeOffset.UtcNow.Ticks);
-
-            return (redisConnection, redisDatabase);
-        }
-
-        private static IConnectionMultiplexer ConnectToRedis(RedisOptions options, ILogger logger)
+        private static async Task<IConnectionMultiplexer> ConnectToRedis(RedisOptions options, ILogger logger)
         {
             var loggerTextWriter = new LoggerTextWriter(logger);
             if (options.Factory != null)
             {
-                return options.Factory(loggerTextWriter);
+                return await options.Factory(loggerTextWriter);
             }
 
             if (options.Options.EndPoints.Any())
             {
-                return ConnectionMultiplexer.Connect(options.Options, loggerTextWriter);
+                return await ConnectionMultiplexer.ConnectAsync(options.Options, loggerTextWriter);
             }
 
             var configurationOptions = new ConfigurationOptions();
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -53,9 +53,9 @@ namespace Microsoft.AspNetCore.SignalR.Redis
         {
             if (_redisServerConnection == null)
             {
+                await _connectionLock.WaitAsync();
                 try
                 {
-                    await _connectionLock.WaitAsync();
                     if (_redisServerConnection == null)
                     {
                         var writer = new LoggerTextWriter(_logger);
--- a/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
@@ -49,65 +49,6 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             RedisLog.ConnectingToEndpoints(_logger, options.Value.Options.EndPoints, _serverName);
         }
 
-        private async Task EnsureRedisServerConnection()
-        {
-            if (_redisServerConnection == null)
-            {
-                await _connectionLock.WaitAsync();
-                try
-                {
-                    if (_redisServerConnection == null)
-                    {
-                        var writer = new LoggerTextWriter(_logger);
-                        _redisServerConnection = await _options.ConnectAsync(writer);
-                        _bus = _redisServerConnection.GetSubscriber();
-                        _redisServerConnection.ConnectionRestored += (_, e) =>
-                        {
-                            // We use the subscription connection type
-                            // Ignore messages from the interactive connection (avoids duplicates)
-                            if (e.ConnectionType == ConnectionType.Interactive)
-                            {
-                                return;
-                            }
-
-                            RedisLog.ConnectionRestored(_logger);
-                        };
-
-                        _redisServerConnection.ConnectionFailed += (_, e) =>
-                        {
-                            // We use the subscription connection type
-                            // Ignore messages from the interactive connection (avoids duplicates)
-                            if (e.ConnectionType == ConnectionType.Interactive)
-                            {
-                                return;
-                            }
-
-                            RedisLog.ConnectionFailed(_logger, e.Exception);
-                        };
-
-                        if (_redisServerConnection.IsConnected)
-                        {
-                            RedisLog.Connected(_logger);
-                        }
-                        else
-                        {
-                            RedisLog.NotConnected(_logger);
-                        }
-
-                        SubscribeToAll();
-                        SubscribeToGroupManagementChannel();
-                        SubscribeToAckChannel();
-                    }
-                }
-                finally
-                {
-                    _connectionLock.Release();
-                }
-
-
-            }
-        }
-
         public override async Task OnConnectedAsync(HubConnectionContext connection)
         {
             await EnsureRedisServerConnection();
@@ -165,10 +106,10 @@ namespace Microsoft.AspNetCore.SignalR.Redis
             return Task.WhenAll(tasks);
         }
 
-        public override async Task SendAllAsync(string methodName, object[] args)
+        public override Task SendAllAsync(string methodName, object[] args)
         {
             var message = _protocol.WriteInvocation(methodName, args);
-            await PublishAsync(_channels.All, message);
+            return PublishAsync(_channels.All, message);
         }
 
         public override Task SendAllExceptAsync(string methodName, object[] args, IReadOnlyList<string> excludedIds)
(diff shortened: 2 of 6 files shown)


samples/ChatSample/RedisUserTracker.cs
```diff
@@ -106,7 +106,7 @@ private static IConnectionMultiplexer ConnectToRedis(RedisOptions options, ILogg
             var loggerTextWriter = new LoggerTextWriter(logger);
             if (options.Factory != null)
             {
-                return options.Factory(loggerTextWriter);
+                return options.Factory(loggerTextWriter).Result;
```
mikaelm12: 😅

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -151,10 +157,11 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
             return Task.WhenAll(tasks);
         }
 
-        public override Task SendAllAsync(string methodName, object[] args)
+        public override async Task SendAllAsync(string methodName, object[] args)
         {
+            await EnsureRedisServerConnection();
```
davidfowl: This should be in publish and subscribe, I don't think it needs to be anywhere else. Those are the 2 redis operations we do today.

samples/ChatSample/RedisUserTracker.cs
anurse: Gotta keep pushing `async` up the stack :)

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -52,51 +52,57 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+                _bus = _redisServerConnection.GetSubscriber();
+                _redisServerConnection.ConnectionRestored += (_, e) =>
+                {
                 // We use the subscription connection type
                 // Ignore messages from the interactive connection (avoids duplicates)
                 if (e.ConnectionType == ConnectionType.Interactive)
-                {
-                    return;
-                }
+                    {
```
anurse: spacing?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -52,51 +52,57 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
             RedisLog.ConnectingToEndpoints(_logger, options.Value.Options.EndPoints, _serverName);
-            _redisServerConnection = _options.Connect(writer);
+        }
 
-            _redisServerConnection.ConnectionRestored += (_, e) =>
+        private async Task EnsureRedisServerConnection()
+        {
+            if (_redisServerConnection == null)
             {
+                var writer = new LoggerTextWriter(_logger);
```
anurse: Need a lock. The lifetime manager can be called from multiple concurrent tasks.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: Yeah, got a little over zealous in adding this

src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs
```diff
@@ -16,10 +17,10 @@ public class RedisOptions
             AbortOnConnectFail = false
         };
 
-        public Func<TextWriter, IConnectionMultiplexer> Factory { get; set; }
+        public Func<TextWriter, Task<IConnectionMultiplexer>> Factory { get; set; }
 
         // TODO: Async
```
anurse: Remove the comment ;)

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
JamesNK: (semaphore lock since it is async)

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -412,7 +419,7 @@ private Task PublishAsync(string channel, byte[] payload)
 
         public void Dispose()
         {
-            _bus.UnsubscribeAll();
+            _bus?.UnsubscribeAll();
             _redisServerConnection.Dispose();
```
JamesNK: ?.Dispose here too

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -552,7 +553,7 @@ public class TestObject
 
         private RedisHubLifetimeManager<MyHub> CreateLifetimeManager(TestRedisServer server, MessagePackHubProtocolOptions messagePackOptions = null, JsonHubProtocolOptions jsonOptions = null)
         {
-            var options = new RedisOptions() { Factory = t => new TestConnectionMultiplexer(server) };
+            var options = new RedisOptions() { Factory = async (t) => await Task.FromResult(new TestConnectionMultiplexer(server)) };
```
JamesNK: Can be: `Factory = (t) => Task.FromResult<IConnectionMultiplexer>(new TestConnectionMultiplexer(server))`

src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs
```diff
@@ -16,10 +17,10 @@ public class RedisOptions
...
-        internal IConnectionMultiplexer Connect(TextWriter log)
+        internal async Task<IConnectionMultiplexer> Connect(TextWriter log)
         {
             if (Factory == null)
```
JamesNK: Assign factory to a local variable inside this method so there is no way the field can be set to null out from under you inbetween this test and using the factory.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -52,51 +53,59 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
 
-            var writer = new LoggerTextWriter(logger);
             RedisLog.ConnectingToEndpoints(_logger, options.Value.Options.EndPoints, _serverName);
-            _redisServerConnection = _options.Connect(writer);
+        }
 
-            _redisServerConnection.ConnectionRestored += (_, e) =>
+        private async Task EnsureRedisServerConnection()
+        {
+            _redisConnectionSemaphor.Wait();
```
JamesNK: Rename to _connectionLock.
WaitAsync for non-blocking.
Double check locking (i.e. if _redisServerConnection is not null then exit method immediately)
Place logic to start connection inside try/finally to ensure _lock.Release is always called

src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs
```diff
@@ -16,12 +17,12 @@ public class RedisOptions
             AbortOnConnectFail = false
         };
 
-        public Func<TextWriter, IConnectionMultiplexer> Factory { get; set; }
+        public Func<TextWriter, Task<IConnectionMultiplexer>> Factory { get; set; }
 
-        // TODO: Async
-        internal IConnectionMultiplexer Connect(TextWriter log)
+        internal async Task<IConnectionMultiplexer> Connect(TextWriter log)
```
anurse: nit: rename `ConnectAsync`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -52,51 +53,68 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+            if (_redisServerConnection == null)
             {
-                // We use the subscription connection type
-                // Ignore messages from the interactive connection (avoids duplicates)
-                if (e.ConnectionType == ConnectionType.Interactive)
+                try
                 {
-                    return;
-                }
+                    await _connectionLock.WaitAsync();
```
BrennanConroy: this goes outside the try

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -52,51 +53,68 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+                    if (_redisServerConnection == null)
+                    {
+                        var writer = new LoggerTextWriter(_logger);
+                        _redisServerConnection = await _options.Connect(writer);
+                        _bus = _redisServerConnection.GetSubscriber();
+                        _redisServerConnection.ConnectionRestored += (_, e) =>
+                        {
+                        // We use the subscription connection type
+                        // Ignore messages from the interactive connection (avoids duplicates)
+                        if (e.ConnectionType == ConnectionType.Interactive)
```
BrennanConroy: formatting

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -52,51 +53,68 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
...
+                        if (_redisServerConnection.IsConnected)
+                        {
+                            RedisLog.Connected(_logger);
+                        }
+                        else
+                        {
+                            RedisLog.NotConnected(_logger);
+                        }
+
+                        SubscribeToAll();
```
BrennanConroy: ??? We only need to subscribe once, the library handles resubscribing on connection restores

BrennanConroy: Also, you aren't guaranteed to set this callback before you connect to redis so you would miss setting up all the subscriptions

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -151,10 +169,10 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
             return Task.WhenAll(tasks);
         }
 
-        public override Task SendAllAsync(string methodName, object[] args)
+        public override async Task SendAllAsync(string methodName, object[] args)
         {
             var message = _protocol.WriteInvocation(methodName, args);
-            return PublishAsync(_channels.All, message);
+            await PublishAsync(_channels.All, message);
```
BrennanConroy: revert these

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -314,10 +332,11 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
             return Task.CompletedTask;
         }
 
-        private Task PublishAsync(string channel, byte[] payload)
+        private async Task PublishAsync(string channel, byte[] payload)
         {
+            await EnsureRedisServerConnection();
```
BrennanConroy: This shouldn't be needed right?

src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs
```diff
@@ -16,12 +17,12 @@ public class RedisOptions
...
         {
-            if (Factory == null)
+            var localFactory = Factory;
```
BrennanConroy: Why?

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -309,6 +309,7 @@ public class RedisHubLifetimeManagerTests
         [Fact]
         public async Task RemoveGroupFromConnectionOnDifferentServerNotInGroupDoesNothing()
         {
+            while (System.Diagnostics.Debugger.IsAttached) { }
```
BrennanConroy: Fix your VS 😛

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: We should be consistent, but we already have inconsistency :(.

https://github.com/aspnet/SignalR/blob/8a3516284e0019dcaaffbe695c88907494382868/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L116-L118

https://github.com/aspnet/SignalR/blob/8a3516284e0019dcaaffbe695c88907494382868/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs#L174-L177

src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs
mikaelm12: > Assign factory to a local variable inside this method so there is no way the field can be set to null out from under you inbetween this test and using the factory.

Suggestion by @JamesNK

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: > We only need to subscribe once, the library handles resubscribing on connection restores

Okay. This is how it was implemented before. I'll fix it.

anurse: Make sure to verify that actually works. It's possible the API won't allow you to subscribe until after connecting

anurse: @BrennanConroy I think you're confusing this because of indentation. This logic only happens on initial startup.

@mikaelm12 Fix the indentation :). If this is the same as the initial code, just leave it in this PR.

mikaelm12: Turns out formatting is important 😄

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
mikaelm12: This is needed since there are scenarios where OnConnected isn't called

src/Microsoft.AspNetCore.SignalR.Redis/RedisOptions.cs
JamesNK: Add a comment:
> Factory is publically settable. Assign to local variable before null check for thread safety.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
JamesNK: It does actually matter - https://stackoverflow.com/questions/24139084/semaphoreslim-waitasync-before-after-try-block

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -45,51 +46,71 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
             _channels = new RedisChannels(typeof(THub).FullName);
             _protocol = new RedisProtocol(hubProtocolResolver.AllProtocols);
 
-            var writer = new LoggerTextWriter(logger);
             RedisLog.ConnectingToEndpoints(_logger, options.Value.Options.EndPoints, _serverName);
-            _redisServerConnection = _options.Connect(writer);
+        }
 
-            _redisServerConnection.ConnectionRestored += (_, e) =>
+        private async Task EnsureRedisServerConnection()
```
anurse: nit: move this below `public` methods

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: `IHubContext` could be used before any connections actually exist on this server. @mikaelm12 's right here, we aren't guaranteed that the connection is up when we get here.

PR #1925, opened 2018-04-09
fix #1281 by adding support for DELETE endpoint on server

### LINKED ISSUE #1281: Consider Adding Graceful Termination of LongPolling Transport
Depending on the timeout for the termination of the LongPolling transport causes a delay in the OnDisconnected method being called. For example I was seeing this in the SocketsSample. When refreshing a client and connecting again you eventually get the disconnect message from your previous session which is odd.

anurse:
Add support for a `DELETE` endpoint to stop the long polling connection gracefully (rather than just waiting for it to time out).

TODO:
* [x] C# Client support including Functional Testing
* [x] TS Client support
* [x] Functional testing of TS


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
```diff
@@ -185,6 +185,24 @@ public Task StartAsync(Uri url, TransferFormat transferFormat)
...
+                _ = SendDeleteRequest(pollUrl);
+            }
+        }
+
+        private async Task SendDeleteRequest(Uri pollUrl)
+        {
+            try
+            {
+                Log.SendingDeleteRequest(_logger, pollUrl);
+                var resp = await _httpClient.DeleteAsync(pollUrl);
```
JamesNK: response. Abbreviations trigger me.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.Log.cs
```diff
@@ -99,6 +108,30 @@ public static void PollResponseReceived(ILogger logger, HttpResponseMessage resp
                         response.Content.Headers.ContentLength ?? -1, null);
                 }
             }
+
+            public static void SendingDeleteRequest(ILogger logger, Uri pollUrl)
+            {
+                if (logger.IsEnabled(LogLevel.Debug))
+                {
+                    _sendingDeleteRequest(logger, pollUrl.ToString(), null);
```
JamesNK: Does the logging not ToString Uri's automatically? I'm sure I have seen logging that passes the Uri directly.

davidfowl: Yep, we should be doing that.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
```diff
@@ -185,6 +185,24 @@ public Task StartAsync(Uri url, TransferFormat transferFormat)
                 _application.Output.Complete(_error);
 
                 Log.ReceiveStopped(_logger);
+
+                // Send the DELETE request to clean-up the connection on the server, but don't wait for it
+                _ = SendDeleteRequest(pollUrl);
```
davidfowl: Why not await this?

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1744,12 +1839,18 @@ private static DefaultHttpContext MakeRequest(string path, ConnectionContext con
...
-        private static void SetTransport(HttpContext context, HttpTransportType transportType)
+        private static void SetTransport(HttpContext context, HttpTransportType transportType, CancellationToken terminateTransport = default)
         {
+            context.RequestAborted = terminateTransport;
             switch (transportType)
             {
                 case HttpTransportType.WebSockets:
-                    context.Features.Set<IHttpWebSocketFeature>(new TestWebSocketConnectionFeature());
+                    var feature = new TestWebSocketConnectionFeature();
+                    if (terminateTransport.CanBeCanceled)
```
davidfowl: Eerr, why?

davidfowl:
> I'm curious why not use CloseMessage?

Decided that this should be at the transport layer since it's impossible to terminate the connection at that level.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
anurse: I'm on board with you. For some reason I keep doing it with `resp` though...

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
anurse: I was probably going overboard trying to shut down the mock connection. I can yank it.

moozzyk:
`abort` comes back... Can you update the spec?

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.cs
```diff
@@ -487,6 +492,36 @@ private static bool ServerHasWebSockets(IFeatureCollection features)
...
+                return;
+            }
+
+            Log.TerminatingConection(_logger);
+
+            // Complete the receiving end of the pipe
+            connection.Application.Output.Complete();
+
+            // Dispose the connection gracefully, but don't wait for it
+            _ = connection.DisposeAsync(closeGracefully: true);
```
davidfowl: I think this needs to be call `_ =_manager.DisposeAndRemoveAsync(connection, closeGracefully: false);`, otherwise nobody will remove the disposed connection from the list.

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1616,6 +1616,93 @@ public TrackingMemoryStream(SemaphoreSlim callerTracker, Task waitTask)
...
+
+                await pollTask;
+
+                // Verify that everything shuts down
+                await connection.ApplicationTask.OrTimeout();
+                await connection.TransportTask.OrTimeout();
+
+                // Verify the response from the DELETE request
+                Assert.Equal(StatusCodes.Status202Accepted, deleteContext.Response.StatusCode);
+                Assert.Equal("text/plain", deleteContext.Response.ContentType);
```
davidfowl: Assert that the connection manager no longer tracks the connection.

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1616,6 +1616,93 @@ public TrackingMemoryStream(SemaphoreSlim callerTracker, Task waitTask)
...
+                var context = MakeRequest("/foo", connection);
+
+                var services = new ServiceCollection();
+                services.AddSingleton<TestConnectionHandler>();
+                var builder = new ConnectionBuilder(services.BuildServiceProvider());
+                builder.UseConnectionHandler<TestConnectionHandler>();
+                var app = builder.Build();
+                var options = new HttpConnectionOptions();
+
+                var pollTask = dispatcher.ExecuteAsync(context, options, app).OrTimeout();
```
davidfowl: Do you want the timeout here instead of on the awaiting of the pollTask itself?

anurse: We usually do it this way, but I do think it fits better on the `await`. I can put it there.

davidfowl: Yah

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.Log.cs
```diff
@@ -40,6 +40,12 @@ private static class Log
             private static readonly Action<ILogger, Exception> _negotiationRequest =
                 LoggerMessage.Define(LogLevel.Debug, new EventId(10, "NegotiationRequest"), "Sending negotiation response.");
 
+            private static readonly Action<ILogger, HttpTransportType, Exception> _receivedDeleteRequestForUnsupportedTransport =
+                LoggerMessage.Define<HttpTransportType>(LogLevel.Trace, new EventId(11, "ReceivedDeleteRequestForUnsupportedTransport"), "Recieved DELETE request for unsupported transport: {TransportType}");
```
BrennanConroy: i before e except after **whenever you feel like it**

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/LongPollingTransport.cs
```diff
@@ -1,13 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
```
BrennanConroy: sort

anurse:
:up: :date: C# client and functional tests, and TypeScript client (no functional tests). Latest PR feedback not yet addressed.

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.Log.cs
anurse: Ugh, that one is literally the bane of my existence.

anurse:
Something we need to talk about: Should this wait to allow the client to drain messages or is the client expected to never come back after the `DELETE`?

anurse:
:up: :date: added support for TypeScript and functional tests. There was a bug in the way transport shutdown occurred on the TypeScript client where the connection didn't await `ITransport.stop` during `HttpConnection.stop`. Also, `HttpConnection.stop` called `HttpConnection.onclose` itself, meaning it could be double-called if `ITransport.onclose` fired. I changed that logic so that `HttpConnection.stop` just stops the transport and returns, and `HttpConnection.onclose` is fired when `ITransport.onclose` fires.

As part of that, I split the transports into three files. I did that in a single commit so if you want to see the TypeScript client changes without the wall of red and green text, look at this commit: [link to come]

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -5,9 +5,12 @@ import { ConnectionClosed, DataReceived } from "./Common";
 import { DefaultHttpClient, HttpClient } from "./HttpClient";
 import { IConnection } from "./IConnection";
 import { ILogger, LogLevel } from "./ILogger";
+import { ITransport, TransferFormat, TransportType } from "./ITransport";
```
davidfowl: super nice

clients/ts/signalr/src/ITransport.ts
```diff
@@ -0,0 +1,24 @@
...
+    LongPolling,
+}
+
+export enum TransferFormat {
+    Text = 1,
+    Binary,
+}
+
+export interface ITransport {
+    connect(url: string, transferFormat: TransferFormat, connection: IConnection): Promise<void>;
```
davidfowl: I removed the last argument

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -0,0 +1,166 @@
...
+            abortSignal: this.pollAbort.signal,
+            headers: {},
+            timeout: 90000,
+        };
+
+        if (transferFormat === TransferFormat.Binary) {
+            pollOptions.responseType = "arraybuffer";
+        }
+
+        const token = this.accessTokenFactory();
```
davidfowl: where's the await? Did you make these changes before all of the other changes?

anurse: probably, I'll reconcile it

davidfowl: Yes you did you basically don't have changes from the TS changes that were made. The accessToken factory and my change

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -0,0 +1,109 @@
...
+                                this.onclose(error);
+                            }
+                            return;
+                        }
+                    }
+                };
+
+                eventSource.onerror = (e: any) => {
+                    const error = new Error(e.message || "Error occurred");
+                    reject(error);
```
BrennanConroy: Does order matter here? Reject at the end makes more sense to me

clients/ts/signalr/src/Utils.ts
```diff
@@ -15,3 +18,52 @@ export class Arg {
...
+    // Trim of trailing space.
+    return str.substr(0, str.length - 1);
+}
+
+export async function sendMessage(logger: ILogger, transportName: string, httpClient: HttpClient, url: string, accessTokenFactory: () => string, content: string | ArrayBuffer, logMessageContent: boolean): Promise<void> {
+    let headers;
+    const token = accessTokenFactory();
+    if (token) {
+        headers = {
+            ["Authorization"]: `Bearer ${accessTokenFactory()}`,
```
BrennanConroy: Use `token`

anurse:
Rebasing on dev to get the missing changes. Not looking forward to that merge :)

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -238,22 +242,25 @@ export class HttpConnection implements IConnection {
...
             await this.startPromise;
         } catch (e) {
             // this exception is returned to the user as a rejected Promise from the start method
         }
-        this.stopConnection(/*raiseClosed*/ previousState === ConnectionState.Connected, error);
-    }
 
-    private stopConnection(raiseClosed: boolean, error?: Error) {
+        // The transport's onclose will trigger stopConnection which will run our onclose event.
         if (this.transport) {
```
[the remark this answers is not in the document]
anurse: If the transport is null, we haven't started yet. In fact, I should clear out the transport in `stopConnection` so a second call to stop no-ops.

clients/ts/signalr/src/ServerSentEventsTransport.ts
anurse: Actually, we should do one or the other. If `onopen` has fired, we resolve the promise (to say we're connected), and we shouldn't then reject it later. If `onopen` hasn't fired, this is an error during the initial connection and we should reject the promise. I'll tidy this up as well.

clients/ts/signalr/src/Utils.ts
anurse: derp

clients/ts/signalr/src/Utils.ts
```diff
@@ -15,3 +18,52 @@ export class Arg {
         }
     }
 }
+
+export function getDataDetail(data: any, includeContent: boolean): string {
+    let length: string = null;
+    if (data instanceof ArrayBuffer) {
+        length = `Binary data of length ${data.byteLength}.`;
+        if (includeContent) {
+            length += ` Content: '${formatArrayBuffer(data)}'.`;
```
BrennanConroy: Could we remove the '.' from here, logs are now showing up with '..' at the end because the actual log location also puts a '.'

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.Log.cs
BrennanConroy: Still need to fix this, also, full stop

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionDispatcher.Log.cs
```diff
@@ -40,6 +40,12 @@ private static class Log
...
+
+            private static readonly Action<ILogger, Exception> _terminatingConnection =
+                LoggerMessage.Define(LogLevel.Trace, new EventId(12, "TerminatingConection"), "Terminating Long Polling connection due to a DELETE request");
```
BrennanConroy: full-stop

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1616,6 +1616,141 @@ public TrackingMemoryStream(SemaphoreSlim callerTracker, Task waitTask)
...
+
+                // Issue the delete request and make sure the poll completes
+                var deleteContext = new DefaultHttpContext();
+                deleteContext.Request.Path = "/foo";
+                deleteContext.Request.QueryString = new QueryString($"?id={connection.ConnectionId}");
+                deleteContext.Request.Method = "DELETE";
+
```
BrennanConroy: Assert.False(pollTask.IsCompleted)

clients/ts/FunctionalTests/ts/Common.ts
```diff
@@ -1,31 +1,31 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-import { IHubProtocol, JsonHubProtocol, TransportType } from "@aspnet/signalr";
+import { IHubProtocol, JsonHubProtocol, HttpTransportType } from "@aspnet/signalr";
```
JamesNK: 👍

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -559,7 +559,34 @@ describe("hubConnection", () => {
                 }
             });
 
-            if (transportType !== TransportType.LongPolling) {
+            it("can connect to hub with authorization using async token factory", async (done) => {
+                const message = "你好，世界！";
+
+                try {
+                    const hubConnection = new HubConnection("/authorizedhub", {
+                        accessTokenFactory: () => getJwtToken("http://" + document.location.host + "/generateJwtToken"),
```
JamesNK: Is this from the JS async token PR? Curious why it is show up here. Did you move it?

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -5,13 +5,16 @@ import { ConnectionClosed, DataReceived } from "./Common";
...
-import { ITransport, LongPollingTransport, ServerSentEventsTransport, TransferFormat, TransportType, WebSocketTransport } from "./Transports";
+import { LongPollingTransport } from "./LongPollingTransport";
+import { ServerSentEventsTransport } from "./ServerSentEventsTransport";
 import { Arg } from "./Utils";
+import { WebSocketTransport } from "./WebSocketTransport";
 
 export interface IHttpConnectionOptions {
     httpClient?: HttpClient;
-    transport?: TransportType | ITransport;
+    transport?: HttpTransportType | ITransport;
```
JamesNK: Does TS support flags? Is this a single transport type or multiple? I'm just wondering if this should be puralized like in C# client?

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    private url: string;
+    private pollXhr: XMLHttpRequest;
+    private pollAbort: AbortController;
+    private shutdownTimeout: number;
+    private running: boolean;
+
+    constructor(httpClient: HttpClient, accessTokenFactory: () => string | Promise<string>, logger: ILogger, logMessageContent: boolean) {
+        this.httpClient = httpClient;
+        this.accessTokenFactory = accessTokenFactory || (() => null);
```
JamesNK: nit: Defaulting to a null string would be a bit more efficient.

clients/ts/signalr/src/HttpConnection.ts
anurse: This wasn't configured as a flags for some reason. I'm not sure if @moozzyk tried it and it didn't work or just didn't do it. I'll file an issue and work it out today.

anurse: https://github.com/aspnet/SignalR/issues/1961

clients/ts/signalr/src/LongPollingTransport.ts
anurse: You mean default to `null` rather than a `null`-returning function? We'd have to change the signature and call site.

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
anurse: Because I split `Transports.ts` into four files (one for each transport and one for the interface) it got wiped out in a bad rebase.

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
```diff
@@ -123,6 +120,9 @@ public Task StartAsync(Uri url, TransferFormat transferFormat)
                 throw;
             }
 
+            _transport.Output.Complete();
```
davidfowl: Why does this matter?

clients/ts/signalr/src/LongPollingTransport.ts
JamesNK: Ohhhh, I thought it was a string or a func, but it is always a func. nvm

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
davidfowl: Completing the pipes was causing the poll loop to throw, how?

anurse: > how?

It wasn't throwing, I mistook that. It was simply completing "gracefully" as though the server terminated the poll.

anurse:
:up: :date: feedback resolved, holding on Approval and green builds

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import { AbortController } from "./AbortController";
+import { DataReceived, TransportClosed } from "./Common";
+import { HttpError, TimeoutError } from "./Errors";
+import { HttpClient, HttpRequest } from "./HttpClient";
+import { IConnection } from "./IConnection";
```
davidfowl: You can remove these now

clients/ts/signalr/src/ITransport.ts
```diff
@@ -0,0 +1,24 @@
+import { DataReceived, TransportClosed } from "./Common";
+import { IConnection } from "./IConnection";
```
davidfowl: nit: Remove this is unused

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -0,0 +1,174 @@
...
+                    // tslint:disable-next-line:no-string-literal
+                    pollOptions.headers["Authorization"] = `Bearer ${token}`;
+                }
+
+                try {
+                    const pollUrl = `${url}&_=${Date.now()}`;
+                    this.logger.log(LogLevel.Trace, `(LongPolling transport) polling: ${pollUrl}`);
+                    const response = await this.httpClient.get(pollUrl, pollOptions);
+
+                    if (!this.running) {
```
davidfowl: Why do we need this

anurse: As I explained elsewhere, this is just diagnostic. I can yank it.

clients/ts/signalr/src/HttpConnection.ts
moozzyk: There was no fallback originally so flags did not make much sense.

Issue #1905, opened 2018-04-08
SignalR and culture


JamesNK:
SignalR server and client should run consistently regardless of the computer's configured culture.

Ensure `CultureInfo.InvariantCulture` is used when converting between strings and values, and when comparing strings.

http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html

pranavkm:
For Mvc's functional tests, we use a middleware that sets up the culture to a non-US one (https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNetCore.Mvc.FunctionalTests/Infrastructure/CultureReplacerMiddleware.cs) - think we pick `en-GB`. You could consider something along these lines to start with.

jaredpar:
> Roslyn started running their PR tests on a Spanish OS 

That's correct. We are essentially running our debug unit test 32 bit run on Spanish OS. 

> I'm pretty sure they use the .NET CI infrastructure we're transitioning to

Yes if you're on the .NET CI infrastructure you can just use the image name Windows.10.Amd64.ClientRS3.ES.Open. Here is the PR where we merged the work back into our production branches.

https://github.com/dotnet/roslyn/pull/25874

PR #1928, opened 2018-04-10
Use invariant culture and ordinal comparisons

JamesNK:
- Ordinal comparison on group names, connection IDs, invocation IDs and headers
- A couple of other minor culture fixes


JamesNK:
https://github.com/aspnet/SignalR/issues/1905

JamesNK:
In the future might want to consider some code analysis for culture comparisons.

BrennanConroy:
Do we have a testing strategy for different cultures?

JamesNK:
Simplest way to do it is run the tests again with a different culture.

PR #1933, opened 2018-04-10
Change the HubSample to reconnect to the server

davidfowl:
- Modified the sample to test our reconnect patterns
- Add IConnectionInherentKeepAliveFeature to the tcp connection impl since Kestrel doesn't implement the heartbeat (yet).

### DIFF
--- a/samples/ClientSample/HubSample.cs
+++ b/samples/ClientSample/HubSample.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Linq;
 using System.Net;
 using System.Threading;
@@ -34,7 +35,6 @@ namespace ClientSample
                 .WithLogging(logging =>
                 {
                     logging.AddConsole();
-                    logging.SetMinimumLevel(LogLevel.Trace);
                 });
 
             if (uri.Scheme == "net.tcp")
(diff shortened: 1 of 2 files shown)


samples/ClientSample/Tcp/TcpConnection.cs
```diff
@@ -4,13 +4,15 @@
...
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Connections.Features;
 using Microsoft.AspNetCore.Http.Features;
 
 namespace ClientSample
 {
-    public class TcpConnection : ConnectionContext
+    public class TcpConnection : ConnectionContext, IConnectionInherentKeepAliveFeature
```
anurse: Wat? TCP doesn't have inherent keep-alive...

samples/ClientSample/HubSample.cs
```diff
@@ -48,71 +48,102 @@ internal static void Register(CommandLineApplication app)
...
-                    var line = await (Task<string>)completedTask;
+                        if (line == null || closedTokenSource.Token.IsCancellationRequested)
+                        {
+                            break;
+                        }
 
-                    if (line == null)
+                        await connection.InvokeAsync<object>("Send", line);
+                    }
+                    catch (IOException)
```
anurse: I don't like that we have to catch various kinds of exceptions to detect restartable conditions... Not sure the best way to approach that though.

samples/ClientSample/Tcp/TcpConnection.cs
davidfowl: > Add IConnectionInherentKeepAliveFeature to the tcp connection impl since Kestrel doesn't implement the heartbeat (yet).

Read my description brah

anurse: Put it in a comment then brah

PR #1935, opened 2018-04-10
Clean up timer code with custom awaitable

davidfowl:
- Gets rid of locks by using an event loop

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
@@ -28,11 +28,9 @@ namespace Microsoft.AspNetCore.Http.Connections
 
         private readonly ConcurrentDictionary<string, (HttpConnectionContext Connection, ValueStopwatch Timer)> _connections =
             new ConcurrentDictionary<string, (HttpConnectionContext Connection, ValueStopwatch Timer)>(StringComparer.Ordinal);
-        private Timer _timer;
+        private TimerAwaitable _timer;
         private readonly ILogger<HttpConnectionManager> _logger;
         private readonly ILogger<HttpConnectionContext> _connectionLogger;
-        private readonly object _executionLock = new object();
-        private bool _disposed;
 
         public HttpConnectionManager(ILoggerFactory loggerFactory, IApplicationLifetime appLifetime)
         {
@@ -40,22 +38,14 @@ namespace Microsoft.AspNetCore.Http.Connections
             _connectionLogger = loggerFactory.CreateLogger<HttpConnectionContext>();
             appLifetime.ApplicationStarted.Register(() => Start());
             appLifetime.ApplicationStopping.Register(() => CloseConnections());
+            _timer = new TimerAwaitable(_heartbeatTickRate, _heartbeatTickRate);
         }
 
         public void Start()
         {
-            lock (_executionLock)
-            {
-                if (_disposed)
-                {
-                    return;
-                }
+            _timer.Start();
 
-                if (_timer == null)
-                {
-                    _timer = new Timer(Scan, this, _heartbeatTickRate, _heartbeatTickRate);
-                }
-            }
+            _ = ExecuteTimerLoop();
         }
 
         public bool TryGetConnection(string id, out HttpConnectionContext connection)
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
```diff
@@ -28,34 +28,24 @@ public partial class HttpConnectionManager
...
-        private readonly object _executionLock = new object();
-        private bool _disposed;
 
         public HttpConnectionManager(ILoggerFactory loggerFactory, IApplicationLifetime appLifetime)
         {
             _logger = loggerFactory.CreateLogger<HttpConnectionManager>();
             _connectionLogger = loggerFactory.CreateLogger<HttpConnectionContext>();
             appLifetime.ApplicationStarted.Register(() => Start());
             appLifetime.ApplicationStopping.Register(() => CloseConnections());
+            _timer = new TimerAwaitable(_heartbeatTickRate, _heartbeatTickRate);
```
anurse: Since we're going to be `await`ing this, maybe give it a name that makes sense for that like `_nextHeartbeat`. So the code is `await _nextHeartbeat`

src/Microsoft.AspNetCore.Http.Connections/Internal/TimerAwaitable.cs
```diff
@@ -0,0 +1,99 @@
...
+        public void UnsafeOnCompleted(Action continuation)
+        {
+            OnCompleted(continuation);
+        }
+
+        public void Stop()
+        {
+            _running = false;
+
+            Tick();
```
anurse: You're going to fire the timer when someone calls `Stop`?

davidfowl: You have to yield the await.

anurse: Should it return a bool then so you can skip your action? Then the loop is `while(await _nextHeartbeat) { ... }`

I get that you're obsessed with event loops now but I'm not 100% convinced that this is "better" yet...

davidfowl: Seems fine, it's 1000% better objectively so but to be honest, I don't really care if we merge this change right now. It's late. I can do it in another release. But:
- It removes the locking
- It's pull so there's no overlapping ticks possible
- It fixes the races with disposal and changing the timer interval
- The code that runs in the tick callback can be async without being async void

anurse: Ok, I'm convinced, let's do it

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
```diff
@@ -114,118 +104,87 @@ private static string MakeNewConnectionId()
...
-        }
-
-        public void Scan()
-        {
-            // If we couldn't get the lock it means one of 2 things are true:
-            // - We're about to dispose so we don't care to run the scan callback anyways.
-            // - The previous Scan took long enough that the next scan tried to run in parallel
-            // In either case just do nothing and end the timer callback as soon as possible
-            if (!Monitor.TryEnter(_executionLock))
+            using (_timer)
```
JamesNK: I'm not sure what this using statement is doing. Would a comment explaining what is going on be useful?

src/Microsoft.AspNetCore.Http.Connections/Internal/TimerAwaitable.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public void UnsafeOnCompleted(Action continuation)
+        {
+            OnCompleted(continuation);
+        }
+
+        public void Stop()
+        {
+            _running = false;
```
JamesNK: Check if disposed and throw (or return)

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
```diff
@@ -114,118 +104,87 @@ private static string MakeNewConnectionId()
...
-                        // Capture the connection state
-                        status = connection.Status;
-
-                        lastSeenUtc = connection.LastSeenUtc;
-                    }
-                    finally
-                    {
-                        connection.Lock.Release();
-                    }
+                    await connection.Lock.WaitAsync();
```
JamesNK: Move above try

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
anurse: Part of the goal here is to change the timer callback into a loop, so in this case we have a concrete, lexical, lifetime for the timer which means we can use `using`

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
mikaelm12: > It does actually matter - https://stackoverflow.com/questions/24139084/semaphoreslim-waitasync-before-after-try-bloc

davidfowl:
🆙 📅

src/Microsoft.AspNetCore.Http.Connections/HttpConnectionManager.cs
```diff
@@ -28,34 +28,25 @@ public partial class HttpConnectionManager
 
         private readonly ConcurrentDictionary<string, (HttpConnectionContext Connection, ValueStopwatch Timer)> _connections =
             new ConcurrentDictionary<string, (HttpConnectionContext Connection, ValueStopwatch Timer)>(StringComparer.Ordinal);
-        private Timer _timer;
+        private TimerAwaitable _nextHeartbeat;
```
JamesNK: readonly

src/Microsoft.AspNetCore.Http.Connections/Internal/TimerAwaitable.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        private TimeSpan _period;
+
+        private TimeSpan _dueTime;
+        private bool _disposed;
+        private bool _running = true;
+
+        public TimerAwaitable(TimeSpan dueTime, TimeSpan period)
+        {
+            _dueTime = dueTime;
```
JamesNK: readonly

src/Microsoft.AspNetCore.Http.Connections/Internal/TimerAwaitable.cs
```diff
@@ -0,0 +1,108 @@
...
+            _period = period;
```
JamesNK: readonly

src/Microsoft.AspNetCore.Http.Connections/Internal/TimerAwaitable.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        public void Start()
+        {
+            if (_timer == null)
+            {
+                lock (this)
```
txdv: Isn't it a good practice to have an explicit lock object?

txdv:
The class TimerAwaitable is dope, would it make sense to provide it in corefx?

src/Microsoft.AspNetCore.Http.Connections/Internal/TimerAwaitable.cs
```diff
@@ -14,11 +14,12 @@ public class TimerAwaitable : IDisposable, ICriticalNotifyCompletion
...
         private static readonly Action _callbackCompleted = () => { };
 
-        private TimeSpan _period;
+        private readonly TimeSpan _period;
 
-        private TimeSpan _dueTime;
+        private readonly TimeSpan _dueTime;
         private bool _disposed;
         private bool _running = true;
+        private object _lockObj = new object();
```
JamesNK: readonly!

PR #1938, opened 2018-04-10
Rename HttpTransportType

mikaelm12:
API review task
HttpTransportType -> HttpTransportTypes


JamesNK:
Ehh. Pluralization make sense on the external APIs because a dev is specifying a number of transport types, but internally SignalR tends to use it without flags to select a single transport 🤷‍♂️

mikaelm12:
> but internally SignalR tends to use it without flags to select a single transport

Yeah, I agree. That does feel weird.

davidfowl:
I vote no. I didn’t like it in the meeting either

mikaelm12:
Alright. Closing this then

PR #1939, opened 2018-04-10
Rename Callback -> Configure in UseConnections

mikaelm12:
API review task
Renaming callback to configure

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/ConnectionsAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/ConnectionsAppBuilderExtensions.cs
@@ -12,6 +12,11 @@ namespace Microsoft.AspNetCore.Builder
     {
         public static IApplicationBuilder UseConnections(this IApplicationBuilder app, Action<ConnectionsRouteBuilder> configure)
         {
+            if (configure == null)
+            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Http.Connections/ConnectionsAppBuilderExtensions.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.AspNetCore.Builder
 {
     public static class ConnectionsAppBuilderExtensions
     {
-        public static IApplicationBuilder UseConnections(this IApplicationBuilder app, Action<ConnectionsRouteBuilder> callback)
+        public static IApplicationBuilder UseConnections(this IApplicationBuilder app, Action<ConnectionsRouteBuilder> configure)
```
JamesNK: Need to check this is not null

anurse: `callback = callback ?? throw new ArgumentNullException(nameof(callback));`

@davidfowl 'll love it ;P

JamesNK: nooo

Issue #1896, opened 2018-04-07
Unify HttpConnectionOptions and HttpOptions


davidfowl:
I'm not sure why we have 2 very similar types on the client side by regardless, we need to unify them. We might also want to consider adding Client to the name HttpClientOptions since it'll conflict with the server if you use them in the same app.

/cc @JamesNK do you know why we have both?

JamesNK:
I'll remove HttpConnectionOptions, rename HttpOptions to HttpClientOptions, and have the factory use it.

PR #1942, opened 2018-04-11
Refactor HttpOptions in client

JamesNK:
- Rename HttpOptions to HttpConfigurationOptions
- Remove HttpConfigurationOptions from SignalR client project
- Updated client HttpConnection ctors


JamesNK:
https://github.com/aspnet/SignalR/issues/1896

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnectionOptions.cs
```diff
@@ -8,23 +8,54 @@
 using System.Net.WebSockets;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Connections.Internal;
 
 namespace Microsoft.AspNetCore.Http.Connections.Client
 {
-    public class HttpOptions
+    public class HttpConnectionOptions
```
JamesNK: There was talk about renaming this. Having a `HttpConnectionOptions` to configure `HttpConnection` makes sense to me.

A `HttpConnectionClientOptions` to configure `HttpConnection` doesn't make as much sense. I think if we rename one we need to rename the other.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -72,48 +72,51 @@ public HttpConnection(Uri url, HttpTransportType transports)
...
+            return new HttpConnectionOptions { Url = url, Transports = transports };
         }
 
-        public HttpConnection(Uri url, HttpTransportType transports, ILoggerFactory loggerFactory, HttpOptions httpOptions)
+        public HttpConnection(HttpConnectionOptions httpConnectionOptions, ILoggerFactory loggerFactory)
         {
-            Url = url ?? throw new ArgumentNullException(nameof(url));
+            if (httpConnectionOptions.Url == null)
+            {
+                throw new InvalidOperationException("Cannot create a HttpConnection without a URL.");
```
anurse: I think this should be an `ArgumentException`. It's an invalid argument.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -72,48 +72,51 @@ public HttpConnection(Uri url, HttpTransportType transports)
...
+            }
+
+            Url = httpConnectionOptions.Url;
```
anurse: The `Url` property could just forward to `_httpConnectionOptions`. :bike: 🏠

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -72,48 +72,51 @@ public HttpConnection(Uri url, HttpTransportType transports)
...
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
 
             _logger = _loggerFactory.CreateLogger<HttpConnection>();
-            _httpOptions = httpOptions;
+            _httpConnectionOptions = httpConnectionOptions;
 
-            _requestedTransports = transports;
+            _requestedTransports = httpConnectionOptions.Transports;
```
anurse: Probably don't need this field either

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnectionOptions.cs
anurse: Yeah, it does seem weird to rename this type. We'll have to thing about it a bit. There's no user-facing "HttpConnection" type on the server, so it makes sense to rename that one to something else.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnectionOptions.cs
```diff
@@ -8,23 +8,54 @@
...
+        }
+
         /// <summary>
         /// Gets or sets a delegate for wrapping or replacing the <see cref="HttpMessageHandlerFactory"/>
         /// that will make HTTP requests the server.
         /// </summary>
         public Func<HttpMessageHandler, HttpMessageHandler> HttpMessageHandlerFactory { get; set; }
 
-        public IReadOnlyCollection<KeyValuePair<string, string>> Headers { get; set; }
+        public IDictionary<string, string> Headers
```
anurse: Auto-properties with initializers?

JamesNK: It's not so as to throw if the collection is set to null.

anurse: I don't think we usually do things that way with these. We generally throw at the consumer if the value is null.

anurse: TBH though, I might prefer this. It's unambiguously wrong to set these to `null`, since we preinitialize them for you.

davidfowl:
The red is @mikaelm12 's redis flakiness 😢

Issue #1862, opened 2018-04-05
HubConnection: Stop copying invocation handlers on every message dispatch


davidfowl:
Today we're copying the list of callbacks (which is usually one) every time we dispatch a hub message. We should avoid doing that.

https://github.com/aspnet/SignalR/blob/80f87e7730e04e6979ee934126be7d608fc43724/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L440-L447

See an approach to fixing this by @FlorianGrimm  (who originally had some git issues).

https://github.com/aspnet/SignalR/pull/1598

PR #1945, opened 2018-04-11
Dont always copy handlers

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1862
Previously we were always making copies of the the invocation handler list in preparation for dispatching an invocation. Now we only make a new copy if there have been updates to the set of handlers.

https://github.com/aspnet/SignalR/blob/80f87e7730e04e6979ee934126be7d608fc43724/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L440-L447

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -791,10 +791,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             public void Dispose()
             {
-                lock (_handlerList)
-                {
-                    _handlerList.Remove(_handler);
-                }
+                 _handlerList.Remove(_handler);
             }
         }
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -34,7 +34,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private readonly IHubProtocol _protocol;
         private readonly IServiceProvider _serviceProvider;
         private readonly IConnectionFactory _connectionFactory;
-        private readonly ConcurrentDictionary<string, List<InvocationHandler>> _handlers = new ConcurrentDictionary<string, List<InvocationHandler>>(StringComparer.Ordinal);
+        private readonly ConcurrentDictionary<string, InvocationHandlerList> _handlers = new ConcurrentDictionary<string, InvocationHandlerList>(StringComparer.Ordinal);
         private bool _disposed;
 
         // Transient state to a connection
@@ -87,7 +87,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             // It's OK to be disposed while registering a callback, we'll just never call the callback anyway (as with all the callbacks registered before disposal).
             var invocationHandler = new InvocationHandler(parameterTypes, handler, state);
-            var invocationList = _handlers.AddOrUpdate(methodName, _ => new List<InvocationHandler> { invocationHandler },
+            var invocationList = _handlers.AddOrUpdate(methodName, _ => new InvocationHandlerList(invocationHandler) ,
                 (_, invocations) =>
                 {
                     lock (invocations)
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -788,19 +781,60 @@ private Task WaitConnectionLockAsync([CallerMemberName] string memberName = null
...
             {
-                lock (_handlerList)
+                 _handlerList.Remove(_handler);
+            }
+        }
+
+        private class InvocationHandlerList
+        {
+            private readonly List<InvocationHandler> _invocationHandlers;
+            private InvocationHandler[] _CopiedHandlers;
```
davidfowl: why is this uppercase?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -788,19 +781,60 @@ private Task WaitConnectionLockAsync([CallerMemberName] string memberName = null
...
+                _invocationHandlers = new List<InvocationHandler>() { handler };
+            }
+
+            internal InvocationHandler[] GetCopiedHandlers()
+            {
+                if (_CopiedHandlers == null)
+                {
+                    lock (_invocationHandlers)
+                    {
+                        _CopiedHandlers = _invocationHandlers.ToArray();
```
anurse: 1. Capture `_CopiedHandlers` into a local before testing against `null`, to avoid another thread messing with it.
2. Check `_CopiedHandlers` again when inside the lock to avoid another thread setting it after your null check.

So, something like this:

```csharp
// First, just grab the current value without the lock
var handlers = _CopiedHandlers;

// If handlers is non-null, we can just return it.
if(handlers == null) 
{
    lock(_invocationHandlers)
    {
        // Did another thread already set it?
        if(_CopiedHandlers == null)
        {
            // No, so initialize it
            _CopiedHandlers = _invocationHandlers.ToArray();
        }
        // Either way, copy the (now initialized) value locally while inside the lock
        handlers = _CopiedHandlers;
    }
}
return handlers;
```

That way, the fast path, when handlers are being retrieved and nobody is messing with them, will occur outside the lock, as desired, but when we do need to take the lock, we do everything inside the lock before releasing it.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -788,19 +781,60 @@ private Task WaitConnectionLockAsync([CallerMemberName] string memberName = null
...
+            public Subscription(InvocationHandler handler, InvocationHandlerList handlerList)
             {
                 _handler = handler;
                 _handlerList = handlerList;
             }
 
             public void Dispose()
             {
-                lock (_handlerList)
+                 _handlerList.Remove(_handler);
```
BrennanConroy: nit: 5 spaces

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -788,19 +781,60 @@ private Task WaitConnectionLockAsync([CallerMemberName] string memberName = null
...
+            }
+
+            internal InvocationHandlerList Add(InvocationHandler handler)
+            {
+                lock (_invocationHandlers)
+                {
+                    _invocationHandlers.Add(handler);
+                    _CopiedHandlers = null;
+                }
+                return this;
```
BrennanConroy: why is this returning?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -788,19 +781,65 @@ private Task WaitConnectionLockAsync([CallerMemberName] string memberName = null
...
             {
-                lock (_handlerList)
+                _handlerList.Remove(_handler);
+            }
+        }
+
+        private class InvocationHandlerList
+        {
+            private readonly List<InvocationHandler> _invocationHandlers;
+            private InvocationHandler[] _copiedHandlers;
```
JamesNK: Add a comment about what this is. Something like:
>A cached copy of the handlers that doesn't change for thread safety. Adding or removing a handler sets this to null

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -788,19 +781,67 @@ private Task WaitConnectionLockAsync([CallerMemberName] string memberName = null
...
+            // A lazy cached copy of the handlers that doesn't change for thread safety. 
+            // Adding or removing a handler sets this to null.
+            private InvocationHandler[] _copiedHandlers;
+
+            internal InvocationHandlerList(InvocationHandler handler)
+            {
+                _invocationHandlers = new List<InvocationHandler>() { handler };
+            }
+
+            internal InvocationHandler[] GetCopiedHandlers()
```
anurse: nit: Just name it `GetHandlers`. I don't like the name `CopiedHandlers` and want to keep it as private as possible ;P

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -788,19 +781,67 @@ private Task WaitConnectionLockAsync([CallerMemberName] string memberName = null
...
+            {
+                var handlers = _copiedHandlers;
+                if (handlers == null)
+                {
+                    lock (_invocationHandlers)
+                    {
+                        // Check if the handlers are set, if not we'll copy them over.
+                        if (_copiedHandlers == null)
+                        {
+                            _copiedHandlers = _invocationHandlers.ToArray();
```
JamesNK: Ahh, I just thought of something. If _invocationHandlers is empty there set _copiedHandlers to Array.Empty. (Does ToArray do that internally? idk)

Issue #1944, opened 2018-04-11
Flaky test: RedisEndToEndTests.HubConnectionCanSendAndReceiveGroupMessages


JamesNK:
https://travis-ci.org/aspnet/SignalR/jobs/364910777#L1422

```
[xUnit.net 00:00:23.0712597]     Microsoft.AspNetCore.SignalR.Redis.Tests.RedisEndToEndTests.HubConnectionCanSendAndReceiveGroupMessages(transportType: WebSockets, protocolName: "messagepack") [FAIL]
Failed   Microsoft.AspNetCore.SignalR.Redis.Tests.RedisEndToEndTests.HubConnectionCanSendAndReceiveGroupMessages(transportType: WebSockets, protocolName: "messagepack")
Error Message:
 System.TimeoutException : Operation in HubConnectionCanSendAndReceiveGroupMessages timed out at /_/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs:86
Stack Trace:
   at System.Threading.Tasks.TaskExtensions.<OrTimeout>d__6`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.SignalR.Redis.Tests.RedisEndToEndTests.<HubConnectionCanSendAndReceiveGroupMessages>d__3.MoveNext() in /_/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs:line 86
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1959, opened 2018-04-11
increase jasmine timeout and add timestamps to logs

anurse:
cc @ryanbrandenburg 

Hoping to merge ASAP.

### DIFF
--- a/clients/ts/FunctionalTests/ts/ConnectionTests.ts
+++ b/clients/ts/FunctionalTests/ts/ConnectionTests.ts
@@ -10,6 +10,9 @@ const commonOptions: IHttpConnectionOptions = {
     logger: TestLogger.instance,
 };
 
+// On slower CI machines, these tests sometimes take longer than 5s
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 10 * 1000;
+
 describe("connection", () => {
     it("can connect to the server without specifying transport explicitly", (done) => {
         const message = "Hello World!";
@@ -88,7 +91,7 @@ describe("connection", () => {
                 connection.onclose = (error) => {
                     // Search the logs for the message content
                     expect(TestLogger.instance.currentLog.messages.length).toBeGreaterThan(0);
-                    for (const [_, logMessage] of TestLogger.instance.currentLog.messages) {
+                    for (const [_, __, logMessage] of TestLogger.instance.currentLog.messages) {
                         expect(logMessage).not.toContain(message);
                     }
                     done();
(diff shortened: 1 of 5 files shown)


clients/ts/FunctionalTests/ts/TestLogger.ts
```diff
@@ -34,7 +34,7 @@ export class TestLogger implements ILogger {
     public currentLog: TestLog = new TestLog();
 
     public log(logLevel: LogLevel, message: string): void {
-        this.currentLog.addMessage(logLevel, message);
+        this.currentLog.addMessage(new Date(), logLevel, message);
```
JamesNK: Is this local time or UTC? I'd prefer UTC

clients/ts/FunctionalTests/ts/ConnectionTests.ts
```diff
@@ -10,6 +10,9 @@ const commonOptions: IHttpConnectionOptions = {
     logger: TestLogger.instance,
 };
 
+// On slower CI machines, these tests sometimes take longer than 5s
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 10 * 1000;
```
JamesNK: If a longer timeout is required then it should be automatic for all tests. Is there a common test file included in every test to include this?

clients/ts/FunctionalTests/ts/TestLogger.ts
anurse: Sure, I can do that. We only really care about the elapsed time.

clients/ts/FunctionalTests/ts/ConnectionTests.ts
anurse: Because of the way rollup combines modules, I wasn't positive that I could get it in one place so I felt like the easiest quick fix was to set it everywhere. After this PR I'll do some work to find the right place to put this.

anurse:
> Looks like a bunch of tests failed, are they related?

https://github.com/aspnet/SignalR/issues/1944

clients/ts/FunctionalTests/ts/TestLogger.ts
anurse: `new Date` is UTC, turns out :).

anurse:
```
dotnet.exe : The active test run was aborted. Reason: Additional text encountered after finished reading JSON content: ,. Path '', line 1, position 0.
```

doesn't seem related...

BrennanConroy:
Will supposedly be fixed by https://github.com/aspnet/BuildTools/pull/627

PR #1965, opened 2018-04-11
Move Benchmark Apps to new folder and use .All package when Benchmarks runs our scenarios

BrennanConroy:
cc @sebastienros


BrennanConroy:
Isn't that what I did https://github.com/aspnet/SignalR/pull/1965/files#diff-2a0a6e8c81f03ba6e58438da722c8936R7

sebastienros:
Yes, then just the scenario.sh then after

PR #1968, opened 2018-04-12
fix #1914 by switching RedisProtocol to MsgPack

### LINKED ISSUE #1914: Consider using msgpack for Redis protocol
Right now the protocol is completely internal and uses the BinaryWriter APIs in .NET (which don't interoperate with anything else). If we want to open this up for allowing others to publish to clients via redis we should use something more standard (and also document and version the protocol).

anurse:
Fixes #1914.

To make this work, I had to move the code to serialize a `SerializeHubMessage` into the Redis assembly (otherwise the core would have to reference MessagePack-CSharp). I think that's actually cleaner anyway, as it's the only place that needs to serialize/deserialize these.

That also means `SerializeHubMessage` gained a constructor that takes an `IEnumerable<SerializedMessage>` and `SerializedMessage` went from private-nested to pubternal. Both reasonable changes I think, since other lifetime managers will need to construct `SerializedHubMessage` from serialized messages.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
@@ -23,7 +23,8 @@ namespace Microsoft.AspNetCore.SignalR.Internal
         {
             for (var i = 0; i < messages.Count; i++)
             {
-                SetCache(messages[i].ProtocolName, messages[i].Serialized);
+                var message = messages[i];
+                SetCache(message.ProtocolName, message.Serialized);
             }
         }
 
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
@@ -19,11 +19,11 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
         public HubMessage Message { get; }
 
-        public SerializedHubMessage(IEnumerable<SerializedMessage> messages)
+        public SerializedHubMessage(IReadOnlyList<SerializedMessage> messages)
         {
-            foreach (var message in messages)
+            for (var i = 0; i < messages.Count; i++)
             {
-                SetCache(message.ProtocolName, message.Serialized);
+                SetCache(messages[i].ProtocolName, messages[i].Serialized);
             }
         }
 
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
 using Microsoft.AspNetCore.SignalR.Internal.Protocol;
 
 namespace Microsoft.AspNetCore.SignalR.Internal
@@ -20,8 +19,12 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
         public HubMessage Message { get; }
 
-        private SerializedHubMessage()
+        public SerializedHubMessage(IEnumerable<SerializedMessage> messages)
         {
+            foreach (var message in messages)
+            {
+                SetCache(message.ProtocolName, message.Serialized);
+            }
         }
 
         public SerializedHubMessage(HubMessage message)
@@ -44,46 +47,6 @@ namespace Microsoft.AspNetCore.SignalR.Internal
             }
 
             return serialized;
-        }
-
-        public static void WriteAllSerializedVersions(BinaryWriter writer, HubMessage message, IReadOnlyList<IHubProtocol> protocols)
-        {
-            // The serialization format is based on BinaryWriter
-            // * 1 byte number of protocols
-            // * For each protocol:
-            //   * Length-prefixed string using 7-bit variable length encoding (length depends on BinaryWriter's encoding)
-            //   * 4 byte length of the buffer
-            //   * N byte buffer
-
-            if (protocols.Count > byte.MaxValue)
-            {
-                throw new InvalidOperationException($"Can't serialize cache containing more than {byte.MaxValue} entries");
-            }
-
-            writer.Write((byte)protocols.Count);
-            foreach (var protocol in protocols)
-            {
-                writer.Write(protocol.Name);
-
-                var buffer = protocol.GetMessageBytes(message);
-                writer.Write(buffer.Length);
-                writer.Write(buffer);
-            }
-        }
-
-        public static SerializedHubMessage ReadAllSerializedVersions(BinaryReader reader)
-        {
-            var cache = new SerializedHubMessage();
-            var count = reader.ReadByte();
-            for (var i = 0; i < count; i++)
-            {
-                var protocol = reader.ReadString();
-                var length = reader.ReadInt32();
-                var serialized = reader.ReadBytes(length);
-                cache.SetCache(protocol, serialized);
-            }
-
-            return cache;
         }
 
         private void SetCache(string protocolName, byte[] serialized)
@@ -144,18 +107,6 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
             result = default;
             return false;
-        }
-
-        private readonly struct SerializedMessage
-        {
-            public string ProtocolName { get; }
-            public byte[] Serialized { get; }
-
-            public SerializedMessage(string protocolName, byte[] serialized)
-            {
-                ProtocolName = protocolName;
-                Serialized = serialized;
-            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
@@ -9,7 +9,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
     {
         public static int ReadArrayHeader(ref ReadOnlyMemory<byte> data)
         {
-            var arr = GetArray(in data);
+            var arr = GetArray(data);
             var val = MessagePackBinary.ReadArrayHeader(arr.Array, arr.Offset, out var readSize);
             data = data.Slice(readSize);
             return val;
@@ -17,7 +17,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
 
         public static int ReadMapHeader(ref ReadOnlyMemory<byte> data)
         {
-            var arr = GetArray(in data);
+            var arr = GetArray(data);
             var val = MessagePackBinary.ReadMapHeader(arr.Array, arr.Offset, out var readSize);
             data = data.Slice(readSize);
             return val;
@@ -25,7 +25,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
 
         public static string ReadString(ref ReadOnlyMemory<byte> data)
         {
-            var arr = GetArray(in data);
+            var arr = GetArray(data);
             var val = MessagePackBinary.ReadString(arr.Array, arr.Offset, out var readSize);
             data = data.Slice(readSize);
             return val;
@@ -33,7 +33,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
 
         public static byte[] ReadBytes(ref ReadOnlyMemory<byte> data)
         {
-            var arr = GetArray(in data);
+            var arr = GetArray(data);
             var val = MessagePackBinary.ReadBytes(arr.Array, arr.Offset, out var readSize);
             data = data.Slice(readSize);
             return val;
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisProtocol.cs
```diff
@@ -35,136 +39,181 @@ public RedisProtocol(IReadOnlyList<IHubProtocol> protocols)
...
             // See WriteAck for format
-            using (var stream = new MemoryStream(data))
-            using (var reader = new BinaryReaderWithVarInt(stream, _utf8NoBom))
+
+            var length = MsgPackUtil.ReadArrayHeader(ref data);
+
+            if (length < 1)
             {
-                return reader.ReadVarInt();
+                throw new InvalidDataException("Insufficient items in Ack array.");
```
JamesNK: How about adding a helper method to MsgPackUtil for this check: `MsgPackUtil.ValidateItems(1, "Ack");`

Should be re-usable here and in the hub protocol.

anurse:
Before:

```
                         Method |        Mean |     Error |    StdDev |         Op/s |  Gen 0 | Allocated |
------------------------------- |------------:|----------:|----------:|-------------:|-------:|----------:|
                       WriteAck |   127.04 ns |  2.547 ns |  4.040 ns |  7,871,521.6 | 0.0060 |     544 B |
              WriteGroupCommand |   336.23 ns |  7.011 ns |  5.854 ns |  2,974,148.5 | 0.0091 |     848 B |
    WriteInvocationNoExclusions |   625.29 ns | 12.193 ns | 11.406 ns |  1,599,266.4 | 0.0114 |    1008 B |
 WriteInvocationSmallExclusions |   784.52 ns | 10.489 ns |  9.298 ns |  1,274,662.8 | 0.0124 |    1112 B |
 WriteInvocationLargeExclusions | 2,707.48 ns | 50.621 ns | 70.963 ns |    369,346.7 | 0.0343 |    3288 B |
                        ReadAck |    82.13 ns |  2.152 ns |  2.013 ns | 12,175,339.3 | 0.0026 |     240 B |
               ReadGroupCommand |   316.08 ns |  6.310 ns |  9.050 ns |  3,163,801.5 | 0.0091 |     808 B |
     ReadInvocationNoExclusions |   327.39 ns |  5.570 ns |  4.937 ns |  3,054,439.0 | 0.0095 |     904 B |
  ReadInvocationSmallExclusions |   468.53 ns | 10.950 ns | 13.035 ns |  2,134,320.7 | 0.0134 |    1136 B |
  ReadInvocationLargeExclusions | 1,903.16 ns | 26.269 ns | 24.572 ns |    525,442.2 | 0.0305 |    3008 B |
```

After:

```
                         Method |        Mean |      Error |    StdDev |         Op/s |  Gen 0 | Allocated |
------------------------------- |------------:|-----------:|----------:|-------------:|-------:|----------:|
                       WriteAck |   165.38 ns |  3.5808 ns |  3.677 ns |  6,046,859.2 | 0.0002 |      32 B |
              WriteGroupCommand |   304.91 ns |  6.1242 ns | 16.661 ns |  3,279,661.3 | 0.0005 |      56 B |
    WriteInvocationNoExclusions |   803.76 ns | 14.0594 ns | 13.151 ns |  1,244,155.1 | 0.0029 |     280 B |
 WriteInvocationSmallExclusions | 1,132.78 ns | 22.3431 ns | 22.945 ns |    882,781.6 | 0.0038 |     384 B |
 WriteInvocationLargeExclusions | 2,363.62 ns | 33.1100 ns | 30.971 ns |    423,080.1 | 0.0114 |     992 B |
                        ReadAck |    42.44 ns |  0.9071 ns |  2.156 ns | 23,564,846.3 |      - |       0 B |
               ReadGroupCommand |   234.88 ns |  2.0854 ns |  1.849 ns |  4,257,524.1 | 0.0014 |     128 B |
     ReadInvocationNoExclusions |   393.87 ns |  4.0405 ns |  3.582 ns |  2,538,908.8 | 0.0033 |     312 B |
  ReadInvocationSmallExclusions |   587.95 ns | 11.5178 ns | 14.976 ns |  1,700,815.8 | 0.0067 |     544 B |
  ReadInvocationLargeExclusions | 1,956.98 ns | 38.8553 ns | 46.254 ns |    510,992.7 | 0.0229 |    2416 B |
```

I'm not terribly surprised by some of the variation. The MessagePack encoding is slightly more complicated than BinaryWriter, but MessagePack is more broadly compatible. Much fewer allocations though since we're not using `MemoryStream` and `BinaryWriter`

anurse:
Also there's some room for improvement if we use `MessagePackBinary` more directly rather than giving it a Stream.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisProtocol.cs
anurse: I think we should keep it to this type (and have a similar one in the hub protocol) rather than sharing it. We don't really want to share a public or internal type between the two assemblies and they have somewhat different usage.

src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
```diff
@@ -20,8 +19,12 @@ public class SerializedHubMessage
 
         public HubMessage Message { get; }
 
-        private SerializedHubMessage()
+        public SerializedHubMessage(IEnumerable<SerializedMessage> messages)
```
davidfowl: Ew IEnumerable? Who so allocatey, this is a per message allocation.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var arr = GetArray(in data);
+            var val = MessagePackBinary.ReadString(arr.Array, arr.Offset, out var readSize);
+            data = data.Slice(readSize);
+            return val;
+        }
+
+        public static byte[] ReadBytes(ref ReadOnlyMemory<byte> data)
+        {
+            var arr = GetArray(in data);
```
davidfowl: is this when you have to pass in explicitly?

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public static int ReadInt32(ref ReadOnlyMemory<byte> data)
+        {
+            var arr = GetArray(in data);
+            var val = MessagePackBinary.ReadInt32(arr.Array, arr.Offset, out var readSize);
+            data = data.Slice(readSize);
+            return val;
+        }
+
+        public static byte ReadByte(ref ReadOnlyMemory<byte> data)
```
davidfowl: This cant be efficient

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -383,7 +383,7 @@ private void SubscribeToAll()
                 {
                     RedisLog.ReceivedFromChannel(_logger, _channels.All);
 
-                    var invocation = _protocol.ReadInvocation(data);
+                    var invocation = _protocol.ReadInvocation((byte[])data);
```
davidfowl: I see, this is because of RedisValue

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
anurse: Apparently. I was getting a compiler error without it. I'll try removing it and verify it.

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
anurse: Yeah, it has conversions to `byte[]` and `string`, not `ReadOnlyMemory<byte>`... yet... ;)

src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
anurse: I could make it require an array or a concrete list. This is a pubternal API though and something that lifetime managers will have to call.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
anurse: I could bypass MessagePack-CSharp completely. It's always going to be a single byte. This is about reading a single-byte integer btw, not about reading large quantities of data.

src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
anurse: I guess we could try-cast for optimization purposes?

davidfowl: Yes, we can't allocate per message like this.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var arr = GetArray(in data);
+            var val = MessagePackBinary.ReadByte(arr.Array, arr.Offset, out var readSize);
+            data = data.Slice(readSize);
+            return val;
+        }
+
+        private static ArraySegment<byte> GetArray(in ReadOnlyMemory<byte> data)
```
davidfowl: Don't use in for ReadOnlyMemory

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisProtocol.cs
```diff
@@ -35,136 +39,172 @@ public RedisProtocol(IReadOnlyList<IHubProtocol> protocols)
...
             // See WriteAck for format
-            using (var stream = new MemoryStream(data))
-            using (var reader = new BinaryReaderWithVarInt(stream, _utf8NoBom))
+            ValidateArraySize(ref data, 1, "Ack");
+            return MsgPackUtil.ReadInt32(ref data);
+        }
+
+        private void WriteSerializedHubMessage(Stream stream, SerializedHubMessage message)
+        {
+            // Written as a MessagePack 'map' where the keys are the name of the protocol (as a MessagePack 'str')
```
davidfowl: Why not an array of pairs?

src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
davidfowl: Nah IList and do  a for loop

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisProtocol.cs
anurse: Identical size and this is more "accurate"

anurse: A map is literally encoded as an array of pairs but with a different header value

src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
anurse: Derp. Don't know why I missed that one. `IReadOnlyList` ok? (assuming a for-loop).

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
anurse: Can you clarify what you mean? The whole pattern or this method in particular? We can require an array, since we know it's always an array, it just means more offset tracking. I can try and benchmark it if you want.

src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedHubMessage.cs
davidfowl: Does that work for array?

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
davidfowl: We can do that after I guess. Yes, I'm guessing the ref adds more overhead because of the slicing. Doesn't need to be in this PR.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisProtocol.cs
davidfowl: I see

src/Microsoft.AspNetCore.SignalR.Redis/Internal/MsgPackUtil.cs
anurse: Sure. Ideally, we'd get MessagePack-CSharp to support `ReadOnlySpan` natively ;)

PR #1970, opened 2018-04-12
fix #1951 by transpiling to ES5

### LINKED ISSUE #1951: preview2 cannot be minified by create-react-app
When trying to run `npm run build` using `create-react-app`? I got this error:
```
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file:

        ./node_modules/@aspnet/signalr/dist/esm/Loggers.js:4

Read more here: http://bit.ly/2tRViJ9
```

With preview1 everything works fine.

anurse:
I also added a folder for "Consumption Tests". This is a place where we can add some automation around consuming SignalR from various places.

The `react-app` folder has a test script that just installs `@aspnet/signalr` and `@aspnet/signalr-protocol-msgpack` and then runs `npm run build` (which is what fails in #1951). Before the changes I made to the `signalr` and `signalr-protocol-msgpack` `package.json` files, this test failed, and after, it passes.

The consumption tests are wired in to the `./build` script, but not in the `npm test` in `client\ts` (since most people won't need to run this)

I'm not 100% sold on the consumption test yet. Interested in others' opinion. If we decide not to do that, I can just trim this back to the actual change, to transpile our ESM version to ES5.


anurse:
We decided to remove the test, it's a little awkward and doesn't quite feel right. I was able to verify that it should work in a `create-react-app` app now though.

PR #1975, opened 2018-04-12
Remove Binary over text sample

davidfowl:
We don't support it anymore


moozzyk:
Such a cool features (no one would use).

davidfowl:
😉

PR #1981, opened 2018-04-12
polyfill UInt8Array.forEach for IE

### LINKED ISSUE #1979: Verify Browser Tests in IE
They're broken, I think because of UInt8Array.forEach not existing. I can easily polyfill. Filing this to track.

anurse:
Fixes #1979 

image

### DIFF
--- a/clients/ts/signalr/src/browser-index.ts
+++ b/clients/ts/signalr/src/browser-index.ts
@@ -20,5 +20,11 @@ if (!Uint8Array.prototype.slice) {
         writable: true,
     });
 }
+if (!Uint8Array.prototype.forEach) {
+    Object.defineProperty(Uint8Array.prototype, "forEach", {
+        value: Array.prototype.forEach,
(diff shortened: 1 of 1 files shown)


mikaelm12:
I'm actually seeing a number of falky failures in the browser tests in IE. I get between 2 and 7 failures every run from this branch.

anurse:
I've noticed some odd behavior with IE throwing "SecurityError" or blocking HTTP requests sometimes. I'm not yet sure if it's something we're doing wrong or something IE is doing wrong...

anurse:
We can open a new issue to track those though.

moozzyk:
AFAIK "Security error" on IE just means you have reached connection limit (e.g.: https://github.com/SignalR/SignalR/pull/2112)

anurse:
Hrm, the tests shouldn't be running in parallel though so I'm not sure how we're running out.

Issue #1976, opened 2018-04-12
API Review work master list


anurse:
* Http.Connections
  * [ ] Remove string overloads of MapConnectionHandler, MapConnections on ConnectionsRouteBuilder
  * [ ] Pubternalize HttpConnectionContext
  * [ ] Unnest HttpConnectionContext.ConnectionStatus (and Pubternalize)
  * [ ] Pubternalize HttpConnectionDispatcher
  * [ ] Pubternalize HttpConnectionManager
  * [ ] Rename `HttpConnectionOptions` (ideas: `HttpConnectionDispatcherOptions`, `HttpConnectionServerOptions`)
  * [ ] Funkifize `WebSocketOptions.SubProtocol` (signature `Func<IReadOnlyList<string>, string>` that receives the subprotocols that the client requested)
* Http.Connections.Client
  * [ ] Internalize `Constants`
  * [ ] Internalize `DefaultTransportFactory`
  * [ ] Reduce constructors on `HttpConnection` to:
    1. `(Uri)`
    2. `(Uri, HttpTransportType)`
    3. `(HttpOptions, ILoggerFactory)` // Throw ArgumentException if HttpOptions.Url is null
    4. `internal (HttpOptions, ILoggerFactory, ITransportFactory)`
  * [ ] Internalize `HttpConnection.Url` (or remove if not used)
  * [ ] Check in on current state of IAsyncDisposable, ensure `DisposeAsync` aligns as much as possible
  * [ ] Default to `TransferFormat.Binary` in `StartAsync(TransferFormat)` via overload
  * [ ] Rename `HttpConnectionOptions.AccessTokenFactory` to `AccessTokenProvider`
  * [ ] Internalize `ITransport`
  * [ ] Internalize `ITransportFactory`
* Http.Connections.Client.Internal
  * [ ] Internalize everything **except** the concrete transport classes
  * [ ] Internalize `Running` task on transport classes
* Http.Connections.Features
  * [ ] Change `IConnectionInherentKeepAliveFeature` to a `bool` rather than a `TimeSpan`, implement directly on `HttpConnection` and remove implementation class
* Http.Connections.Internal
  * [ ] Internalize: 
    * [ ] `AuthorizeHelper`
    * [ ] `AvailableTransport`
    * [ ] `AwaitableThreadPool`
    * [ ] `ConnectionLogScope`
    * [ ] `ServerSentEventsMessageFormatter`
* Http.Connections.Internal.Transports
  * Nothing
[...]

PR #1982, opened 2018-04-12
Make Proxies Internal

mikaelm12:
Part of https://github.com/aspnet/SignalR/issues/1976


src/Microsoft.AspNetCore.SignalR.Core/Proxies.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public class UserProxy<THub> : IClientProxy where THub : Hub
+    internal class UserProxy<THub> : IClientProxy where THub : Hub
```
JamesNK: Should these be moved to an internal namespace?

mikaelm12: Yeah, you're right

PR #1986, opened 2018-04-12
Update usage of TestSink

JunTaoLuo:
Preparing for aspnet/Logging#804

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Security.Cryptography.X509Certificates;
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -118,9 +119,9 @@ public void CannotCreateConnectionWithNullUrlOnOptions()
                 // ignore connection error
             }
 
-            Assert.Equal(2, testSink.Writes.Count);
-            Assert.Equal("SendingHttpRequest", testSink.Writes[0].EventId.Name);
-            Assert.Equal("UnsuccessfulHttpResponse", testSink.Writes[1].EventId.Name);
+            Assert.Equal(2, testSink.Writes.Count());
+            Assert.Equal("SendingHttpRequest", testSink.Writes.First().EventId.Name);
```
JunTaoLuo: Is order important here?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
```diff
@@ -118,9 +119,9 @@ public void CannotCreateConnectionWithNullUrlOnOptions()
...
+            Assert.Equal("UnsuccessfulHttpResponse", testSink.Writes.Skip(1).First().EventId.Name);
```
pakrym: This is pretty bad, use `Assert.Collection`.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.cs
davidfowl: Can you convert to a list first up top then use that list?

JamesNK: How about `var writes = testSink.Writes.ToList();` and then use the list rather than using linq everywhere?

JunTaoLuo:
:up::date:

PR #1992, opened 2018-04-13
API review changes

davidfowl:
- Changed GetMessageBytes to return ReadOnlyMemory
- Make HandshakeProtocol.SuccessHandshakeData a readonly field

## Before

```
             Method |       Input | HubProtocol |       Mean |     Error |    StdDev |        Op/s |  Gen 0 | Allocated |
------------------- |------------ |------------ |-----------:|----------:|----------:|------------:|-------:|----------:|
 WriteSingleMessage | NoArguments |        Json | 1,074.8 ns | 11.669 ns | 10.915 ns |   930,399.5 | 0.0114 |     392 B |
 WriteSingleMessage | NoArguments |     MsgPack |   344.3 ns |  7.487 ns |  6.637 ns | 2,904,147.3 | 0.0010 |      40 B |
```

## After

```
              Method |       Input | HubProtocol |       Mean |     Error |    StdDev |        Op/s |  Gen 0 | Allocated |
------------------- |------------ |------------ |-----------:|----------:|----------:|------------:|-------:|----------:|
 WriteSingleMessage | NoArguments |        Json | 1,112.9 ns | 12.325 ns | 11.529 ns |   898,513.8 | 0.0114 |     392 B |
 WriteSingleMessage | NoArguments |     MsgPack |   366.7 ns |  7.010 ns |  7.791 ns | 2,726,694.7 | 0.0010 |      40 B |
```

There's a performance hit here, but it might still be worth it?

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedMessage.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 
 namespace Microsoft.AspNetCore.SignalR.Internal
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.SignalR.Core/Internal/SerializedMessage.cs
```diff
@@ -1,11 +1,13 @@
+using System;
```
mikaelm12: nit: Header

BrennanConroy:
Why did it get slower?

davidfowl:
We're creating a `ROMemory<byte>` over the array every time, that's not free.

PR #1994, opened 2018-04-13
Make the protocol and messages public

davidfowl:
- More API review feedback


src/Microsoft.AspNetCore.SignalR.Common/CloseMessage.cs
```diff
@@ -4,7 +4,7 @@
 using System;
 using System.Collections.Generic;
 
-namespace Microsoft.AspNetCore.SignalR.Internal.Protocol
+namespace Microsoft.AspNetCore.SignalR
```
JamesNK: Hmmm, I'm not sure about putting them in the main namespace. How often will devs interact with these classes?

davidfowl: You think a .Protocol namespace?

JamesNK: Yeah

davidfowl: Will do.

PR #1999, opened 2018-04-13
Delegate disposable to the IConnectionFactory

### LINKED ISSUE #1996: HttpClient is never disposed when using HubConnection with HttpConnection
We no longer call DisposeAsync and that's where the disposal of the HttpConnection happens

davidfowl:
- Added DisposeAsync to the IConnectionFactory. It's responsible for disposing the connection after the pipe has closed.
- Added dispose callback to WithConnectionFactory

Fixes #1996

### DIFF
--- a/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
+++ b/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
@@ -31,7 +31,10 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         public static IHubConnectionBuilder WithEndPoint(this IHubConnectionBuilder builder, EndPoint endPoint)
         {
-            builder.WithConnectionFactory(format => new TcpConnection(endPoint).StartAsync());
+            builder.WithConnectionFactory(format => new TcpConnection(endPoint).StartAsync(), connection =>
+            {
+                return ((TcpConnection)connection).DisposeAsync();
+            });
 
             return builder;
         }
--- a/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
+++ b/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
@@ -31,10 +31,10 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         public static IHubConnectionBuilder WithEndPoint(this IHubConnectionBuilder builder, EndPoint endPoint)
         {
-            builder.WithConnectionFactory(format => new TcpConnection(endPoint).StartAsync(), connection =>
-            {
-                return ((TcpConnection)connection).DisposeAsync();
-            });
+            builder.WithConnectionFactory(
+                format => new TcpConnection(endPoint).StartAsync(),
+                connection => ((TcpConnection)connection).DisposeAsync()
+            );
 
             return builder;
         }
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
@@ -15,8 +15,6 @@ namespace Microsoft.AspNetCore.Http.Connections.Client.Internal
 {
     public partial class LongPollingTransport : ITransport
     {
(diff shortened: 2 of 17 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/IConnectionFactory.cs
```diff
@@ -9,5 +9,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
     public interface IConnectionFactory
```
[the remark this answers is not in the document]
davidfowl: lol yes! https://github.com/aspnet/Mvc/blob/e34b4e8335468299a75eb0b9e2b0f2296e81dc7a/src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerFactory.cs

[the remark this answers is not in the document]
JamesNK: Well played. You win this round.

davidfowl:
```
Microsoft.AspNetCore.Http.Connections.Client.Internal.LongPollingTransport[9]
      Error while polling 'http://127.0.0.1:38171/hubT?id=c7o8lfOOGsAW-jWMwVi74g'.
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'System.Net.Http.HttpClient'.
   at System.Net.Http.HttpClient.CheckDisposed()
   at System.Net.Http.HttpClient.SendAsync(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Http.Connections.Client.Internal.LongPollingTransport.Poll(Uri pollUrl, CancellationToken cancellationToken) in /_/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs:line 146
[32minfo: Microsoft.AspNetCore.Http.Connections.Client.Internal.LongPollingTransport[1]
```

davidfowl:
> FYI WithConnectionFactory is going bye-byte as part of the code review.

Yep, but I'm not fixing that, at least not here 😄

samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
```diff
@@ -31,7 +31,10 @@ public static IHubConnectionBuilder WithEndPoint(this IHubConnectionBuilder buil
 
         public static IHubConnectionBuilder WithEndPoint(this IHubConnectionBuilder builder, EndPoint endPoint)
         {
-            builder.WithConnectionFactory(format => new TcpConnection(endPoint).StartAsync());
+            builder.WithConnectionFactory(format => new TcpConnection(endPoint).StartAsync(), connection =>
```
JamesNK: I don't like this formatting

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
```diff
@@ -95,12 +92,12 @@ public Task StartAsync(Uri url, TransferFormat transferFormat)
                 // This will also cause the poll to return.
                 await SendDeleteRequest(url);
 
-                // This timeout is only to ensure the poll is cleaned up despite a misbehaving server.
-                // It doesn't need to be configurable.
-                _transportCts.CancelAfter(ShutdownTimeout);
+                _transportCts.Cancel();
```
davidfowl: @anurse I'm basically reverting that change to wait for graceful shutdown. For 2 reasons:

- The only thing that matters is that we send the request to make sure the server does aggressively closes the connection.
- We need to wait for the other side of the loop to end because when this returns the HttpConnection disposes the http client.
- The tests need to be updated because none of them handle the delete properly meaning things end up hanging for 5 seconds extending the test time.

anurse: I'm OK with that. My only concern was to avoid logging "failures" when the poll is cancelled due to graceful shutdown. I think we don't log that as an error though so it's fine.

PR #2004, opened 2018-04-13
Internalize more classes

BrennanConroy:
Part of #1976


samples/ChatSample/PresenceHubLifetimeManager.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.AspNetCore.SignalR.Redis;
+using Microsoft.AspNetCore.SignalR.Internal;
```
mikaelm12: Sort

PR #2015, opened 2018-04-13
API Review Dependency Injection

BrennanConroy:
#1976

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/SignalRConnectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/SignalRConnectionBuilderExtensions.cs
@@ -1,7 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Connections;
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.SignalR.Core/SignalRConnectionBuilderExtensions.cs
```diff
@@ -1,14 +1,23 @@
...
 namespace Microsoft.AspNetCore.SignalR
 {
     public static class SignalRConnectionBuilderExtensions
     {
         public static IConnectionBuilder UseHub<THub>(this IConnectionBuilder connectionBuilder) where THub : Hub
         {
+            var marker = connectionBuilder.ApplicationServices.GetService(typeof(SignalRMarkerService));
+            if (marker == null)
+            {
+                throw new InvalidOperationException("Unable to find the SignalR service. Please add it by " +
```
anurse: Copy from MVC: 

https://github.com/aspnet/Mvc/blob/ec31ff0c28961be64ad0b8228eba3a084086f60b/src/Microsoft.AspNetCore.Mvc.Core/Resources.resx#L210-L212

"Unable to find the required services. Please add all the required services by calling 'IServiceCollection.AddSignalR' inside the call to 'ConfigureServices(...)' in the application startup code."

PR #1972, opened 2018-04-12
Add VerifyNoErrorsScope

JamesNK:
re: https://github.com/aspnet/SignalR/issues/1883#issuecomment-380676048

Errors can be thrown in background threads, logged, and not rethrown. This "scope" will throw if any errors were logged.

Problems with what is currently in the PR:
1. ~~Doesn't compose with other loggers. Probably can be fixed easily~~
2. ~~Needs to be manually added to a test~~ Hooked into `LoggedTest` and replaced with `VerifiableLoggedTest`

PR #2020, opened 2018-04-14
Send delete request after poll ends

davidfowl:
Should fix the 404s in https://github.com/aspnet/SignalR/pull/1972#issuecomment-381314296. The idea is to unwind the client side loop before sending the delete request.


anurse:
Yeah, this is OK I guess. The original intent was that the `DELETE` request would cause the server to complete the poll. It doesn't really matter though, so I'm fine with this.

davidfowl:
Yea the main thing is that the server OnDisconnectedAsync get fired aggressively

PR #2022, opened 2018-04-14
 API review Microsoft.AspNetCore.SignalR.Internal

JamesNK:
- Make DefaultHubDispatcher<THub> methods virtual **(methods are overrides so already virtual)**
- Make DefaultHubProtocolResovler methods virtual
- Move IHubProtocolResolver out of internal namespace
- Change IHubProtocolResolver.GetProtocol to use IReadOnlyList<string>


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
```
davidfowl: LINQ....

JamesNK: Would you prefer `SupportedProtocols` on `HubOptions` was also a readonly list?

davidfowl: It needs to be mutable for users to tweak

PR #2027, opened 2018-04-14
Lazy initialize HttpConnectionContext.Items

JamesNK:
From API review

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
@@ -26,11 +26,13 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
                                          IHttpContextFeature,
                                          IHttpTransportFeature
     {
+        private readonly object _itemsLock = new object();
         private readonly object _heartbeatLock = new object();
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
```diff
@@ -95,7 +97,24 @@ public HttpConnectionContext(string id, IDuplexPipe transport, IDuplexPipe appli
...
+                    {
+                        if (_items == null)
+                        {
+                            _items = new ConnectionItems(new ConcurrentDictionary<object, object>());
+                        }
+                    }
+                }
+                return _items;
+            }
+            set => _items = value ?? throw new ArgumentNullException(nameof(value));
```
JamesNK: Should `Items` be settable? If it threw an error then the double check lock could probably be replaced with a `Lazy<ConnectionItems>`.

davidfowl: No lazy plz. This is fine.

JamesNK: le approve?

Issue #2031, opened 2018-04-15
flaky test: Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.CanStartAndStopConnectionUsingGivenTransport


davidfowl:
```
dotnet.exe : [xUnit.net 00:00:08.1859729]     Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.CanStartAndStopConnectionUsingGivenTransport(transportType: LongPolling) [FAIL]
At C:\Users\appveyor\.dotnet\buildtools\korebuild\2.1.0-preview3-17002\scripts\common.psm1:7 char:5
+     & $_cmd @args
+     ~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: ([xUnit.net 00:0...Polling) [FAIL]:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Failed   Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.CanStartAndStopConnectionUsingGivenTransport(transportType: LongPolling)
Error Message:
 System.Exception : 1 error(s) logged.
Microsoft.AspNetCore.Http.Connections.Client.Internal.LongPollingTransport - ErrorSendingDeleteRequest - Error sending DELETE request to 'http://127.0.0.1:3224/echo?id=IES88WFl2FJN1VfJPzCxJQ'.
===================
System.Net.Http.HttpRequestException: Response status code does not indicate success: 400 (Bad Request).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at Microsoft.AspNetCore.Http.Connections.Client.Internal.LongPollingTransport.<SendDeleteRequest>d__20.MoveNext() in C:\projects\signalr\src\Microsoft.AspNetCore.Http.Connections.Client\Internal\LongPollingTransport.cs:line 211
===================
Stack Trace:
   at Microsoft.AspNetCore.SignalR.Tests.VerifyNoErrorsScope.Dispose() in C:\projects\signalr\test\Microsoft.AspNetCore.SignalR.Tests.Utils\VerifyNoErrorsScope.cs:line 63
   at Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.<CanStartAndStopConnectionUsingGivenTransport>d__4.MoveNext() in C:\projects\signalr\test\Microsoft.AspNetCore.SignalR.Tests\EndToEndTests.cs:line 85
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #2032, opened 2018-04-15
Flush first long poll immediately

davidfowl:
There are other approaches, I wanted to see how big of a change this one would be.

#2031

PS: Need to fix javascript tests


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
```diff
@@ -49,6 +49,14 @@ public Task StartAsync(Uri url, TransferFormat transferFormat)
 
             Log.StartTransport(_logger, transferFormat);
 
+
+            // Do the initial poll, this will return immediately without data
```
JamesNK: What happens if in the short time between the connection starting and this first "poll" a message is written?

test/Microsoft.AspNetCore.SignalR.Client.Tests/ResponseUtils.cs
```diff
@@ -49,6 +51,11 @@ public static bool IsLongPollDeleteRequest(HttpRequestMessage request)
                    (request.RequestUri.PathAndQuery.Contains("?id=") || request.RequestUri.PathAndQuery.Contains("&id="));
         }
 
+        public static bool IsSSERequest(HttpRequestMessage request)
```
JamesNK: Nit: IsSseRequest

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs
davidfowl: What do you mean? The pipe will have the data and it’ll work like normal.

JamesNK: Doesn't the poll response need to be written to the application like `Poll` does?

JamesNK: Ok. I think the terminology needs to be cleaned up a little because the first request when using long polling isn't really a poll because it will never return data, it's more like it is used to initialize polling.

Splitting hairs but could you update the comment to say something like

```
// Make initial long polling request
// Server side implementation of long polling will always return the first request immediately and
// without data so response content can be ignored
```

davidfowl: Sure.

moozzyk: https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Core/PersistentConnection.cs#L32

davidfowl: Throw back 😄

JamesNK:
🆙 📅

PR #2033, opened 2018-04-15
API review changes for Microsoft.AspNetCore.SignalR

JamesNK:
- Pubternalize DefaultHubActivator and make methods virtual
- Remove HubConnectionContext.Input **(used by dispatcher so made internal instead)**
- Internalize DefaultHubCallerContext **(used by ChatSample.PresenceHubLifetimeManager so made pub internal)**
- Internalize HubConnectionStore **(used by RedisHubLifetimeManager so made pub internal)**
- Change order arguments in HubLifetimeManager.AddGroupAsync/RemoveGroupAsync
- Change order arguments in IGroupManager.AddGroupAsync/RemoveGroupAsync
- Rename: IClientProxyExtensions -> ClientProxyExtensions
- Remove HubRouteBuilder.MapHub(string)
- Rename excludedIds to excludedConnectionIds in IHubClients.NExcept methods.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/IGroupManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/IGroupManager.cs
@@ -7,7 +7,7 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public interface IGroupManager
     {
-        Task AddAsync(string connectionId, string groupName);
-        Task RemoveAsync(string connectionId, string groupName);
+        Task AddAsync(string groupName, string connectionId);
+        Task RemoveAsync(string groupName, string connectionId);
     }
 }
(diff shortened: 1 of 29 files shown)


src/Microsoft.AspNetCore.SignalR.Core/IGroupManager.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public interface IGroupManager
     {
-        Task AddAsync(string connectionId, string groupName);
-        Task RemoveAsync(string connectionId, string groupName);
+        Task AddAsync(string groupName, string connectionId);
```
davidfowl: I gotta say this change I’m not sure about m. It’s going to break all of the apps and I’m not even sure it’s right... maybe we should instead rename it AddToGroup

davidfowl:
Can you change the order of parameters back? You’re now adding the ConnectionId to a particular group

JamesNK:
That would result in...

DefaultHubLifetimeManager.cs:
`public override Task AddGroupAsync(string groupName, string connectionId)`

IGroupManager.cs:
`Task AddToGroupAsync(string connectionId, string groupName);`

davidfowl:
Once ci passes

PR #2043, opened 2018-04-16
API review changes

davidfowl:
- Make SerializedHubMessage and SerializedMessage
public
- Made HubConnectionStore public
- Move SendUtils and Utils to the internal folder
- Make Microsoft.AspNetCore.Http.Connections.Common/Internal public.
- Move InvocationRequest to internal folder

Addresses https://github.com/aspnet/SignalR/issues/1976#issuecomment-381518117


src/Microsoft.AspNetCore.Http.Connections.Common/AvailableTransport.cs
```diff
@@ -3,7 +3,7 @@
 
 using System.Collections.Generic;
 
-namespace Microsoft.AspNetCore.Http.Connections.Internal
+namespace Microsoft.AspNetCore.Http.Connections
```
anurse: You're going to have to `IList<string>` the `TransferFormats` property if this is going to be public...

BrennanConroy: Same in NegotiationResponse

Issue #2008, opened 2018-04-13
WinForms net461/471 with SignalR.Client throws on Connect


HaoK:
Using 2.1-preview2 fina lbits:
 
Setup SignalR server with /hub/chat via: https://blogs.msdn.microsoft.com/webdev/2018/02/27/asp-net-core-2-1-0-preview1-getting-started-with-signalr/
This is working via chrome at https://localhost:44343/hubs/chat

Setup WinForms client:
```
•	New WinForms project (net461) + Add preview2 SignalR.Client
•	  <package id="Microsoft.AspNetCore.Connections.Abstractions" version="2.1.0-preview2-final" targetFramework="net461" />
•	  <package id="Microsoft.AspNetCore.Http.Connections.Client" version="1.0.0-preview2-final" targetFramework="net461" />
•	  <package id="Microsoft.AspNetCore.Http.Connections.Common" version="1.0.0-preview2-final" targetFramework="net461" />
•	  <package id="Microsoft.AspNetCore.Http.Features" version="2.1.0-preview2-final" targetFramework="net461" />
•	  <package id="Microsoft.AspNetCore.SignalR.Client" version="1.0.0-preview2-final" targetFramework="net461" />
•	  <package id="Microsoft.AspNetCore.SignalR.Client.Core" version="1.0.0-preview2-final" targetFramework="net461" />
•	  <package id="Microsoft.AspNetCore.SignalR.Common" version="1.0.0-preview2-final" targetFramework="net461" />
•	  <package id="Microsoft.Extensions.Configuration" version="2.1.0-preview2-final" targetFramework="net461" />
```

Trying to connect via:

``` 
        const string ServerURI = "https://localhost:44343/hubs/chat";
            Console.WriteLine("Connecting to {0}", ServerURI);
 
            Connection = new HubConnectionBuilder().WithUrl(ServerURI).Build();
            Connection.Closed += Connection_Closed;
            Connection.On<string, string>("AddMessage", (name, message) =>
                this.Invoke((Action)(() =>
                    RichTextBoxConsole.AppendText(String.Format("{0}: {1}" + Environment.NewLine, name, message))
                ))
            );
            try
            {
                await Connection.StartAsync(); => Blows up with: 
 ```

```
[...]

PR #2046, opened 2018-04-16
add WinForms sample

anurse:
Associated with #2008 

Adds a Windows Forms Client. I don't believe this will fail to build on macOS/Linux, but if it does the CI will tell us and we can adjust. The client doesn't work right now because of #2008 but it builds

image


samples/WindowsFormsSample/ChatForm.cs
```diff
@@ -0,0 +1,132 @@
+using System;
```
JamesNK: Copyright header

samples/WindowsFormsSample/Program.cs
```diff
@@ -0,0 +1,22 @@
+using System;
```
JamesNK: Copyright header.

davidfowl:
When travis works I'll buy it

anurse:
FYI this won't be merged until the branching is finished, and it'll be going into dev.

anurse:
```
Could not run the "GenerateResource" task because MSBuild could not create or connect to a task host with runtime "CLR4" and architecture "x86".
```

☹️ 

I'll look at disabling this on non-Windows.

anurse:
:up: :date: nits picked, and I think I found a sneaky way to make the build no-op on non-Windows.

BrennanConroy:
I don't think VSCode really uses the .sln

davidfowl:
It does, actutslly the more I think about it let’s move this to the samples repo.

anurse:
It doesn't load in VS code. The rest of the projects work fine but you do get a warning. Closing this.

Issue #1963, opened 2018-04-11
API review TS client


JamesNK:
Should review what the TS client makes public.

anurse:
Turns out many of the things we thought were exported were not and I was just not using the tool right (or the tool doesn't support what I was doing.

From Chrome Dev Tools looking at the actual `signalR` namespace:

![image](https://user-images.githubusercontent.com/7574/38838831-e5e1fddc-418c-11e8-8987-690b15cee6e0.png)

anurse:
This is all done now.

PR #2050, opened 2018-04-17
Simple TypeScript API Review Feedback

anurse:
* Review Function Types exported by Common and remove if we don't need them anymore - **I decided to remove them entirely, they aren't really necessary and create a false sense of equivalence between different events**
* `IConnection.send` should be `string | ArrayBuffer` - **I also changed the `onreceive` event**
* Add `IConnection.start` overload
* Add `HubConnection.invoke<T>`
* Add `HubConnection.stream` that returns `Observable<any>`
* Just add `LogLevel.Debug` and do a quick scan to change some `Trace`s to `Debug`s - **I changed most of the connection-level events from Trace to Debug. Message-level events stayed at Trace**
* Don't export `ConsoleLogger` (use `LogLevel.[level]` instead) or `NullLogger` - **I left `NullLogger` but made it's constructor private and used a static `instance` field**
* Remove `LoggerFactory` and just inline logger creation

* Rename `Observable<T>` and `Observer<T>` to `Stream<T>` and `StreamSubscriber<T>` and make them classes so people can attach things to the prototype (including a conversion to Rx Observable) - **Not so sure about making these classes, it makes some of the code awkward. I did it and you can kinda see the awkwardness in the tests**

**Note** Not merging until we `release/2.1` re-opens. Will rebase when that happens.

Still some other API review items to come, details in #1963


clients/ts/FunctionalTests/rollup.config.js
```diff
@@ -8,7 +8,7 @@ import commonjs from 'rollup-plugin-commonjs'
 import resolve from 'rollup-plugin-node-resolve'
 
 export default {
-    input: path.join(__dirname, "obj", "js", "index.js"),
+    input: path.join(__dirname, "obj", "js", "FunctionalTests", "ts", "index.js"),
```
anurse: This is needed because now that the tests are importing from the `Utils.ts` file **inside** `@aspnet/signalr`, TypeScript changed the output folder structure.

clients/ts/FunctionalTests/ts/TestLogger.ts
```diff
@@ -1,4 +1,7 @@
-import { ConsoleLogger, ILogger, LogLevel } from "@aspnet/signalr";
+import { ILogger, LogLevel } from "@aspnet/signalr";
+
+// Pull some useful internal utilities from the signalr library.
```
anurse: This is a sneaky thing tests can do. Since JavaScript imports work on the file level, we can actually just grab internals straight out of the file in which they are defined.

JamesNK: Put that as a comment in the ts file

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -198,13 +85,13 @@ export class HubConnection {
...
     public stop(): Promise<void> {
-        this.logger.log(LogLevel.Trace, "Stopping HubConnection.");
+        this.logger.log(LogLevel.Debug, "Stopping HubConnection.");
 
         this.cleanupTimeout();
         return this.connection.stop();
     }
 
-    public stream<T>(methodName: string, ...args: any[]): Observable<T> {
+    public stream<T = any>(methodName: string, ...args: any[]): StreamResult<T> {
```
JamesNK: default generic type?

JamesNK:
As part of this double check and fix any of the JS samples that have broken. They definitely use the console logger.

clients/ts/signalr/src/HubConnection.ts
anurse: Yeah, it's a thing apparently.

JamesNK:
btw `signalR.TransportType` needs to be renamed to `signalR.HttpTransportType`. I needed to fix that to get some of the sample apps working

anurse:
:up: :date: samples updated

clients/ts/signalr/src/ITransport.ts
```diff
@@ -1,4 +1,3 @@
-import { DataReceived, TransportClosed } from "./Common";
```
JamesNK: Copyright header

anurse: TypeScript auto-import puts things above any leading comments :(. Fixed

anurse:
Any concerns @davidfowl ?

clients/ts/signalr/src/HubConnection.ts
davidfowl: Should this be a `Promise<StreamResult<T>>`?

anurse:
Talked to @davidfowl . We're going to go back to an interface for StreamResult and friends

clients/ts/signalr/src/Utils.ts
```diff
@@ -67,3 +69,104 @@ export async function sendMessage(logger: ILogger, transportName: string, httpCl
...
+                    console.error(`${LogLevel[logLevel]}: ${message}`);
+                    break;
+                case LogLevel.Warning:
+                    console.warn(`${LogLevel[logLevel]}: ${message}`);
+                    break;
+                case LogLevel.Information:
+                    console.info(`${LogLevel[logLevel]}: ${message}`);
+                    break;
+                default:
+                    console.log(`${LogLevel[logLevel]}: ${message}`);
```
BrennanConroy: There is a console.debug it looks like https://console.spec.whatwg.org/#debug

clients/ts/signalr/src/Utils.ts
```diff
@@ -67,3 +69,104 @@ export async function sendMessage(logger: ILogger, transportName: string, httpCl
...
+    private readonly minimumLogLevel: LogLevel;
+
+    constructor(minimumLogLevel: LogLevel) {
+        this.minimumLogLevel = minimumLogLevel;
+    }
+
+    public log(logLevel: LogLevel, message: string): void {
+        if (logLevel >= this.minimumLogLevel) {
+            switch (logLevel) {
+                case LogLevel.Error:
```
BrennanConroy: Should critical also be error?

PR #2058, opened 2018-04-17
flow cancellationtoken through HttpConnection.StartAsync

anurse:
Related to API reviewing

### DIFF
--- a/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
+++ b/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Net;
 using System.Threading;
@@ -5,7 +8,6 @@ using System.Threading.Tasks;
 using ClientSample;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection.Extensions;
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
--- a/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
+++ b/samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Net;
+using System.Threading;
 using System.Threading.Tasks;
 using ClientSample;
 using Microsoft.AspNetCore.Connections;
@@ -46,7 +47,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 _endPoint = endPoint;
             }
 
-            public Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat)
+            public Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, CancellationToken cancellationToken)
             {
                 return new TcpConnection(_endPoint).StartAsync();
             }
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
@@ -213,11 +213,11 @@ namespace Microsoft.AspNetCore.Http.Connections.Client
             if (_httpConnectionOptions.Transports == HttpTransportType.WebSockets)
             {
                 Log.StartingTransport(_logger, _httpConnectionOptions.Transports, _httpConnectionOptions.Url);
-                await StartTransport(_httpConnectionOptions.Url, _httpConnectionOptions.Transports, transferFormat);
+                await StartTransport(_httpConnectionOptions.Url, _httpConnectionOptions.Transports, transferFormat, cancellationToken);
             }
             else
             {
-                var negotiationResponse = await GetNegotiationResponse();
+                var negotiationResponse = await GetNegotiationResponse(cancellationToken);
 
                 // This should only need to happen once
                 var connectUrl = CreateConnectUrl(_httpConnectionOptions.Url, negotiationResponse.ConnectionId);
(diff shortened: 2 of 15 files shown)


samples/ClientSample/Tcp/TcpHubConnectionBuilderExtensions.cs
```diff
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
```
mikaelm12: nit:Header

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -229,6 +233,8 @@ internal HttpConnection(HttpConnectionOptions httpConnectionOptions, ILoggerFact
 
                 foreach (var transport in negotiationResponse.AvailableTransports)
                 {
+                    cancellationToken.ThrowIfCancellationRequested();
```
davidfowl: This is probably another one of my biggest beef with CT. You just sprinkle calls to this in arbitrary places in the code. You could also check cancellationToken.IsCancellationRequested to break out of the loop early...

src/Microsoft.AspNetCore.SignalR.Client.Core/IConnectionFactory.cs
```diff
@@ -1,15 +1,16 @@
...
 using Microsoft.AspNetCore.Connections;
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
     public interface IConnectionFactory
     {
-        Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat);
+        Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, CancellationToken cancellationToken);
 
         Task DisposeAsync(ConnectionContext connection);
```
davidfowl: What about DisposeAsync?

anurse: > DisposeAsync accepting a CancellationToken: while in theory it makes sense that anything async can be canceled, disposal is about cleanup, closing things out, free'ing resources, etc., which is generally not something that should be canceled; cleanup is still important for work that's canceled. The same CancellationToken that caused the actual work to be canceled would typically be the same token passed to DisposeAsync, making DisposeAsync worthless because cancellation of the work would cause DisposeAsync to be a nop. If someone wants to avoid being blocked waiting for disposal, they can avoid waiting on the resulting Task, or wait on it only for some period of time.

https://github.com/dotnet/csharplang/blob/195efa07806284d7b57550e7447dc8bd39c156bf/proposals/async-streams.md#iasyncdisposable

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -115,17 +115,21 @@ internal HttpConnection(HttpConnectionOptions httpConnectionOptions, ILoggerFact
             _transportFactory = transportFactory;
         }
 
-        public async Task StartAsync()
+        public Task StartAsync() => StartAsync(TransferFormat.Binary, CancellationToken.None);
+
+        public Task StartAsync(CancellationToken cancellationToken) => StartAsync(TransferFormat.Binary, cancellationToken);
+
+        public async Task StartAsync(TransferFormat transferFormat)
```
mikaelm12: Can we get rid of this overload by by changing the next one to use `CancellationToken cancellationToken = default`

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
anurse: The question is how does the user want to react to cancellation. I don't like that it uses exceptions but that seems to be the common pattern. If we just break the loop, you'd get the error that "no transport could be connected" rather than an `OperationCancelledException`

src/Microsoft.AspNetCore.SignalR.Client.Core/IConnectionFactory.cs
davidfowl: Should we remove it from StopAsync?

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
anurse: Can I get clarity on our policy with overloads vs default arguments @davidfowl ? Previously I'd heard that default arguments are "bad" because they are a language-specific feature and it depends on the caller's compiler filling in the default. I'm happy to use them if we don't care about that. It makes a ton of things easier.

src/Microsoft.AspNetCore.SignalR.Client.Core/IConnectionFactory.cs
anurse: Stop isn't quite the same as Dispose though. Dispose is about releasing resources. Stop is about graceful shutdown of the connection. I think cancelling stop would make sense, but it should put the connection in the "disgracefully terminated" state, rather than restoring it back to running. It can cancel the DELETE request, and other shutdown processes but it should continue to shut down all the local resources associated with the connection (by Disposing the HttpConnection, which isn't cancelable as per above).

Today, we don't do anything with the CT in StopAsync though, so it's somewhat moot whether we keep it or not. If we remove it, adding it back via default parameter is breaking but via overload wouldn't be breaking.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
anurse: I see we've got https://github.com/aspnet/Home/wiki/Engineering-guidelines#async-method-patterns

davidfowl:
I think we need to reduce the scope of this PR by exposing the token on top level APIs but don't bother flowing it to the other APIs like ITransport.

JamesNK:
Starting the transport is part of starting the connection. Not passing it the token would hamstring the effectiveness of cancelling.

davidfowl:
I dislike flowing cancellation throughout the stack and using it in various places untested and throwing from arbitrary points in the code where we don’t know what state we’re in.

anurse:
We're going to go super simple and just put CTs at the public APIs and ignore them for now.

PR #2063, opened 2018-04-17
HubConnectionBuilder.ts

anurse:
HubConnectionBuilder for TypeScript.

Unit Tests all updated and passing. Functional Tests not yet updated.

TODO:
* [x] Update Functional Tests
* [x] Unit tests for HubConnectionBuilder
* [x] Update Samples (guh)


clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export class HubConnectionBuilder {
+    private protocol: any;
+    private httpConnectionOptions: IHttpConnectionOptions;
+    private url: string;
+    private logger: ILogger;
+
+    public configureLogging(logLevel: LogLevel): HubConnectionBuilder;
+    public configureLogging(logger: ILogger): HubConnectionBuilder;
+    public configureLogging(logLevelOrLogger: LogLevel | ILogger): HubConnectionBuilder {
```
JamesNK: Are these parameter names idiomatic TS?

clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -0,0 +1,72 @@
...
+        return this;
+    }
+
+    public build(): HubConnection {
+        // If httpConnectionOptions has a logger, use it. Otherwise, override it with the one
+        // provided to configureLogger
+        const httpConnectionOptions = this.httpConnectionOptions || {};
+
+        // If it's 'null', the user **explicitly** asked for null, don't mess with it.
+        if (httpConnectionOptions.logger === undefined) {
```
JamesNK: defs want a unit test to check this

clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -0,0 +1,72 @@
...
+            // If our logger is undefined or null, that's OK, the HttpConnection constructor will handle it.
+            httpConnectionOptions.logger = this.logger;
+        }
+
+        // Now create the connection
+        if (!this.url) {
+            throw new Error("The 'HubConnectionBuilder.withUrl' method must be called before building the connection");
```
JamesNK: full stop

clients/ts/signalr/src/HubConnectionBuilder.ts
anurse: I'll do a little looking to check but one thing to note is that when you do overloads like this in TypeScript, the actual parameter names (the signature that actually has a body) don't actually appear to the user in IntelliSense:

image
image

anurse:
:up: :date: no longer WIP

clients/ts/signalr-protocol-msgpack/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aspnet/signalr-protocol-msgpack",
-  "version": "1.0.0-rc1-t000",
+  "version": "1.0.0-preview3-t000",
```
JamesNK: Is this right or a change leaking from dev branch?

clients/ts/signalr/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aspnet/signalr",
-  "version": "1.0.0-rc1-t000",
+  "version": "1.0.0-preview3-t000",
```
JamesNK: Ditto

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -16,12 +16,14 @@ export interface HttpRequest {
...
+    constructor(statusCode: number);
+    constructor(statusCode: number, statusText: string);
     constructor(statusCode: number, statusText: string, content: string);
     constructor(statusCode: number, statusText: string, content: ArrayBuffer);
     constructor(
         public readonly statusCode: number,
-        public readonly statusText: string,
-        public readonly content: string | ArrayBuffer) {
+        public readonly statusText?: string,
+        public readonly content?: string | ArrayBuffer) {
```
JamesNK: What is this change for? I'm not 100% up to speed on the intricacies of TS overloading.

clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -0,0 +1,66 @@
...
+import { NullLogger } from "./Loggers";
+import { ConsoleLogger } from "./Utils";
+
+export class HubConnectionBuilder {
+    private protocol: any;
+    private httpConnectionOptions: IHttpConnectionOptions;
+    private url: string;
+    private logger: ILogger;
+
+    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder {
```
JamesNK: What happens if null is passed in? Needs to be handled or validated to give a nice error msg.

clients/ts/signalr/src/HttpClient.ts
anurse: This is just how you make overloads, you add new signatures and you make the actual function definition accept optional parameters as needed so that all the signature combinations can be passed in properly, since JavaScript doesn't **actually** have overloads.

See "Overloads" in https://www.typescriptlang.org/docs/handbook/functions.html

clients/ts/signalr/package-lock.json
anurse: Derp

clients/ts/FunctionalTests/package.json
```diff
@@ -22,8 +22,8 @@
     "build:rollup": "node ../node_modules/rollup/bin/rollup -c",
     "pretest": "npm run build",
     "test": "dotnet build && npm run test-only",
-    "test-only": "ts-node --project ./selenium/tsconfig.json ./selenium/run-tests.ts",
-    "ci-test": "ts-node --project ./selenium/tsconfig.json ./selenium/run-ci-tests.ts"
+    "test-only": "ts-node --project ./selenium/tsconfig-selenium.json ./selenium/run-tests.ts",
+    "ci-test": "ts-node --project ./selenium/tsconfig-selenium.json ./selenium/run-ci-tests.ts"
```
anurse: Renaming this file fixes some problems with VS Code detecting the right tsconfig.json. It makes IntelliSense work better in the FunctionalTests

anurse:
:up: :date: PR feedback

clients/ts/signalr-protocol-msgpack/package-lock.json
JamesNK: Still need to change back?

clients/ts/signalr/package-lock.json
JamesNK: Ditto

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -31,26 +26,21 @@ export class HubConnection {
...
     private receivedHandshakeResponse: boolean;
 
-    constructor(url: string, options?: IHubConnectionOptions);
-    constructor(connection: IConnection, options?: IHubConnectionOptions);
-    constructor(urlOrConnection: string | IConnection, options: IHubConnectionOptions = {}) {
-        options = options || {};
+    public serverTimeoutInMilliseconds: number;
 
-        this.timeoutInMilliseconds = options.timeoutInMilliseconds || DEFAULT_TIMEOUT_IN_MS;
+    constructor(connection: IConnection, logger: ILogger, protocol: IHubProtocol) {
```
davidfowl: So clean 😄  I'm just sad it isn't a connection factory 😢

anurse: Might still be able to make that happen ;)

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -31,26 +26,21 @@ export class HubConnection {
...
+        Arg.isRequired(connection, "connection");
+        Arg.isRequired(logger, "logger");
+        Arg.isRequired(protocol, "protocol");
 
-        this.protocol = options.protocol || new JsonHubProtocol();
-        this.handshakeProtocol = new HandshakeProtocol();
+        this.serverTimeoutInMilliseconds = DEFAULT_TIMEOUT_IN_MS;
```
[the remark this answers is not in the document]
anurse: I think it's better to make the property name super clear. On the server it takes a `TimeSpan` so the units are clear. I don't like numeric timestamp properties that don't make it clear what the units are, but I'm not super passionate about that :)

clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -0,0 +1,88 @@
...
+import { HubConnection, JsonHubProtocol } from "./HubConnection";
+import { IHubProtocol } from "./IHubProtocol";
+import { ILogger, LogLevel } from "./ILogger";
+import { HttpTransportType } from "./ITransport";
+import { NullLogger } from "./Loggers";
+import { Arg, ConsoleLogger } from "./Utils";
+
+export class HubConnectionBuilder {
+    /** @internal */
+    public protocol: any;
```
anurse: We need to talk a bit about this. It's really the only way to test the `transportType` overload of `withUrl`.

In our tsconfig, we set `stripInternal: true`, which tells TypeScript to **strip** items marked with the JSDoc `@internal` annotation **from our .d.ts files**. This means:

* The properties are still "public" in JavaScript, meaning they can be read/written from JavaScript
* TypeScript code that references our compiled output (which are `.js` files with `.d.ts` buddies) cannot access this property (unless they cast to `any`, which "turns off" type checking)
* TypeScript code that references the `.ts` file directly (our tests) can see the property just fine.

This is basically The Only Way to do something like "Internals Visible To" in TypeScript.

clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export class HubConnectionBuilder {
+    /** @internal */
+    public protocol: IHubProtocol;
+    /** @internal */
+    public httpConnectionOptions: IHttpConnectionOptions;
+    /** @internal */
+    public url: string;
+    /** @internal */
+    public logger: ILogger;
```
anurse: We need to talk a bit about this. It's really the only way to test the `transportType` overload of `withUrl`.

In our tsconfig, we set `stripInternal: true`, which tells TypeScript to **strip** items marked with the JSDoc `@internal` annotation **from our .d.ts files**. This means:

* The properties are still "public" in JavaScript, meaning they can be read/written from JavaScript
* TypeScript code that references our compiled output (which are `.js` files with `.d.ts` buddies) cannot access this property (unless they cast to `any`, which "turns off" type checking)
* TypeScript code that references the `.ts` file directly (our tests) can see the property just fine.

This is basically The Only Way to do something like "Internals Visible To" in TypeScript.

clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -0,0 +1,88 @@
...
+            this.logger = new ConsoleLogger(logging);
+        }
+
+        return this;
+    }
+
+    public withUrl(url: string): HubConnectionBuilder;
+    public withUrl(url: string, options: IHttpConnectionOptions): HubConnectionBuilder;
+    public withUrl(url: string, transportType: HttpTransportType): HubConnectionBuilder;
+    public withUrl(url: string, transportTypeOrOptions?: IHttpConnectionOptions | HttpTransportType): HubConnectionBuilder {
```
anurse: I can't find evidence pro or con this as a TypeScript idiom, but I like it. It's clearer for us when we're working on the code. The name `transportTypeOrOptions` does not appear in IntelliSense at all.

anurse: https://blog.mariusschulz.com/2016/08/18/function-overloads-in-typescript provides some support, but is just a rando blog I found ;)

clients/ts/signalr/src/HubConnectionBuilder.ts
davidfowl: This seems fine. I am OK with this.

samples/ChatSample/Views/Home/Index.cshtml
```diff
@@ -17,7 +17,10 @@
 <script src="lib/signalr-client/signalr.js"></script>
 <script>
 let transportType = signalR.HttpTransportType[getParameterByName('transport')] || signalR.HttpTransportType.WebSockets;
-let connection = new signalR.HubConnection('/chat', { transport: transportType, logger: signalR.LogLevel.Information });
+let connection = new signalR.HubConnectionBuilder()
+    .configureLogging(signalR.LogLevel.Information)
+    .withUrl("/chat", { transport: transportType })
```
davidfowl: Update the sample to use the new overload

samples/SignalRSamples/wwwroot/streaming.html
```diff
@@ -53,7 +53,10 @@ <h2>Results</h2>
             });
 
             click('connectButton', function () {
-                connection = new signalR.HubConnection('/streaming', { transport: transportType, logger: signalR.LogLevel.Trace });
+                connection = new signalR.HubConnectionBuilder()
+                    .configureLogging(signalR.LogLevel.Trace)
+                    .withUrl("/streaming", { transport: transportType })
```
davidfowl: Update the sample to use the new overload

clients/ts/signalr/src/HubConnectionBuilder.ts
anurse: It doesn't really matter and this is a little cleaner. A `private` member of a TypeScript class is always "public" to JavaScript. This way means that we get IntelliSense in the places we want it. Using `any` has the same "downsides" as this technique (the member is visible to JS) but none of the benefits.

PR #2064, opened 2018-04-17
Fix #1989 by adding a 'binding failure' pseudo-message

### LINKED ISSUE #1989: [API Review] Make "synthetic" HubMessage for Invocations with binding errors
Rather than an "ArgumentBindingException" property, let's make our IHubProtocol implementations return a new `HubMessage`-derived type `BindingErrorMessage` (name to be :bike: :house:ed) that identifies the Invocation ID (so a response can be sent) and the error.

anurse:
Feel free to open the naming :bike: :house:...


test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisProtocolTests.cs
```diff
@@ -84,22 +84,24 @@ public void WriteGroupCommand(string testName)
         }
 
         // The actual invocation message doesn't matter
-        private static InvocationMessage _testMessage = new InvocationMessage("target", null, Array.Empty<object>());
-        private static Dictionary<string, ProtocolTestData<RedisInvocation>> _invocationTestData = new[]
+        private static InvocationMessage _testMessage = new InvocationMessage("target", Array.Empty<object>());
+
+        // We use a func so we are guaranteed to get a new SerializedHubMessage for each test
```
anurse: I noticed some subtle issues here when re-running tests. The SerializedHubMessage was being reused between tests. I'll do some investigation, but this Func change fixes the issue.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HubMethodInvocationMessage.cs
```diff
@@ -3,60 +3,37 @@
...
 
-            if ((arguments == null && argumentBindingException == null) || (arguments?.Length > 0 && argumentBindingException != null))
-            {
-                throw new ArgumentException($"'{nameof(argumentBindingException)}' and '{nameof(arguments)}' are mutually exclusive");
-            }
-
             Target = target;
-            _arguments = arguments;
-            _argumentBindingException = argumentBindingException;
+            Arguments = arguments;
```
JamesNK: Check arguments is not null

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisProtocolTests.cs
JamesNK: I don't feel strongly about it but instead of making that one piece a func I'd rather the invocation test data was completed recreated each time, i.e. a new copy is returned from a method when required rather than stored as a field.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HubMethodInvocationMessage.cs
anurse: We allowed `null` before, and the protocols set it to null when there are no arguments. I can look at changing that though.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisProtocolTests.cs
anurse: I'm OK with that approach, it's a little cleaner.

PR #2079, opened 2018-04-18
fix #2078 by adding locking

### LINKED ISSUE #2078: SerializedHubMessage doesn't lock properly
https://github.com/aspnet/SignalR/blob/b3b5a13f2c0ef1707278bdeaabfb0f977f19a49e/src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs

If multiple threads are accessing the `SerializedHubMessage`, a race can occur between `TryGetCached` and `SetCached` which can cause duplicate serializations to be registered, and "collection modified during enumeration" errors (cc @BrennanConroy).

We had locking here before but it got removed in refactoring because it was presumed that outer locking would be sufficient, but the outer locking is per connection, and multiple connections can be writing the same message at the same time (in broadcast scenarios).

The fix is simple: Add a lock to `SerializedHubMessage` and take it when accessing the list of serializations.

anurse:
@BrennanConroy saw some concurrency issues in the benchmarks and we found that this was the cause.

Fixes #2078

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.SignalR
         public ReadOnlyMemory<byte> GetSerializedMessage(IHubProtocol protocol)
         {
             // Double-check locking!
-            if (!TryGetCached(protocol.Name, out var serialized))
+            if (!TryGetCachedFast(protocol.Name, out var serialized))
             {
                 lock (_lock)
                 {
--- a/src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using Microsoft.AspNetCore.SignalR.Protocol;
 
 namespace Microsoft.AspNetCore.SignalR
@@ -17,6 +18,7 @@ namespace Microsoft.AspNetCore.SignalR
         private SerializedMessage _cachedItem2;
         private IList<SerializedMessage> _cachedItems;
         private readonly object _lock = new object();
+        private int _count = 0;
 
         public HubMessage Message { get; }
 
@@ -58,23 +60,24 @@ namespace Microsoft.AspNetCore.SignalR
             return serialized;
         }
 
+        // Used for unit testing.
         internal IEnumerable<SerializedMessage> GetAllSerializations()
         {
-            if (_cachedItem1.ProtocolName == null)
+            if (_count < 1)
             {
                 yield break;
             }
 
             yield return _cachedItem1;
 
-            if (_cachedItem2.ProtocolName == null)
+            if (_count < 2)
             {
                 yield break;
             }
 
             yield return _cachedItem2;
 
-            if (_cachedItems == null)
+            if (_count < 3)
             {
                 yield break;
             }
@@ -90,17 +93,23 @@ namespace Microsoft.AspNetCore.SignalR
             if (_cachedItem1.ProtocolName == null)
             {
                 _cachedItem1 = new SerializedMessage(protocolName, serialized);
+                Interlocked.Increment(ref _count);
             }
             else if (_cachedItem2.ProtocolName == null)
             {
                 _cachedItem2 = new SerializedMessage(protocolName, serialized);
+                Interlocked.Increment(ref _count);
             }
             else
             {
                 if (_cachedItems == null)
                 {
                     _cachedItems = new List<SerializedMessage>();
+                    Interlocked.Increment(ref _count);
                 }
+
+                // No need to continue updating _count. It's just used to track the fields. The list
+                // always has to be accessed under a lock (except in the constructor) so we don't need the counter.
 
                 foreach (var item in _cachedItems)
                 {
@@ -115,21 +124,32 @@ namespace Microsoft.AspNetCore.SignalR
             }
         }
 
-        private bool TryGetCached(string protocolName, out ReadOnlyMemory<byte> result)
+        private bool TryGetCachedFast(string protocolName, out ReadOnlyMemory<byte> result)
         {
-            if (string.Equals(_cachedItem1.ProtocolName, protocolName, StringComparison.Ordinal))
+            if (_count > 0 && string.Equals(_cachedItem1.ProtocolName, protocolName, StringComparison.Ordinal))
             {
                 result = _cachedItem1.Serialized;
                 return true;
             }
 
-            if (string.Equals(_cachedItem2.ProtocolName, protocolName, StringComparison.Ordinal))
+            if (_count > 1 && string.Equals(_cachedItem2.ProtocolName, protocolName, StringComparison.Ordinal))
             {
                 result = _cachedItem2.Serialized;
                 return true;
             }
 
-            if (_cachedItems != null)
+            result = ReadOnlyMemory<byte>.Empty;
+            return false;
+        }
+
+        private bool TryGetCached(string protocolName, out ReadOnlyMemory<byte> result)
+        {
+            if (TryGetCachedFast(protocolName, out result))
+            {
+                return true;
+            }
+
+            if (_count > 2)
             {
                 foreach (var serializedMessage in _cachedItems)
                 {
--- a/src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
@@ -90,6 +90,11 @@ namespace Microsoft.AspNetCore.SignalR
 
         private void SetCache(string protocolName, ReadOnlyMemory<byte> serialized)
         {
+            // We set the fields before moving on to the list, if we need it to hold more than 2 items.
+            // In order to prevent "shearing" (where some of the fields of the struct are set by one thread,
+            // while another thread is reading the struct), we atomically increment a counter to ensure
+            // consumers know which structs have been fully assigned.
+            
             if (_cachedItem1.ProtocolName == null)
             {
                 _cachedItem1 = new SerializedMessage(protocolName, serialized);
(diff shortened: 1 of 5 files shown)


davidfowl:
Hah! Nice one

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -16,6 +16,7 @@ public class SerializedHubMessage
         private SerializedMessage _cachedItem1;
         private SerializedMessage _cachedItem2;
         private IList<SerializedMessage> _cachedItems;
+        private object _lock = new object();
```
JamesNK: image

😸

JamesNK:
I wonder if this can be optimized. Right now we take the lock every time this is called. Accessing the first two fields should be thread safe and could happen outside of the lock.

JamesNK:
Also is there anyway we can unit test this to prevent future regression? 😄

anurse:
:up: :date: I found a way to write a couple tests.

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -35,21 +36,55 @@ public SerializedHubMessage(HubMessage message)
 
         public ReadOnlyMemory<byte> GetSerializedMessage(IHubProtocol protocol)
         {
+            // Double-check locking!
             if (!TryGetCached(protocol.Name, out var serialized))
```
JamesNK: This will check the list which isn't thread safe. Should only check:

```
             if (string.Equals(_cachedItem1.ProtocolName, protocolName, StringComparison.Ordinal))
             {
                 result = _cachedItem1.Serialized;
                 return true;
             }
 
             if (string.Equals(_cachedItem2.ProtocolName, protocolName, StringComparison.Ordinal))
             {
                 result = _cachedItem2.Serialized;
                 return true;
             }
```

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -35,21 +36,55 @@ public SerializedHubMessage(HubMessage message)
...
+                        serialized = protocol.GetMessageBytes(Message);
+                        SetCache(protocol.Name, serialized);
+                    }
+                }
             }
 
             return serialized;
         }
 
+        internal IEnumerable<SerializedMessage> GetAllSerializations()
```
JamesNK: Add a comment that this is for unit tests so @davidfowl doesn't freak about the yield.

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
anurse: Yep, good catch

anurse: Shoot, I think this could shear... The struct is two fields so it's not guaranteed to set atomically.

anurse: To clarify (I was headed out the door when I found this): When setting this, since the `_cachedItem` fields are multi-field structs, a thread could set the Protocol Name but **not** yet have set the Serialized message when the reader comes along to check for a cached copy. This would result in returning an empty `ReadOnlyMemory` as the serialized version for a message (which would be A Bad Thing).

JamesNK: gotcha

anurse: We can probably keep the optimization if we use an atomic counter to track the number of items. I'll look at that.

anurse: Yeah, we just need an atomic way to know that a field has been fully assigned. An `Interlocked.Increment`ed counter should achieve that

anurse:
Ok, updated with the atomic counter to keep the fast path thread-safe. I think it's worth it because the bulk of the accesses are from concurrent `WriteAsync` calls that are accessing already-cached data. So after the first serialization makes it into the cache, we'll want to avoid the lock for all further retrieval of that cached item

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -80,21 +124,32 @@ private void SetCache(string protocolName, ReadOnlyMemory<byte> serialized)
             }
         }
 
-        private bool TryGetCached(string protocolName, out ReadOnlyMemory<byte> result)
+        private bool TryGetCachedFast(string protocolName, out ReadOnlyMemory<byte> result)
         {
-            if (string.Equals(_cachedItem1.ProtocolName, protocolName, StringComparison.Ordinal))
+            if (_count > 0 && string.Equals(_cachedItem1.ProtocolName, protocolName, StringComparison.Ordinal))
```
JamesNK: Add a comment that the count field is used to avoid problems with the struct sheering

anurse: 👍 

Though that made me thing of another thing: Do we need to make the fields `volatile` to ensure read/write barriers are used? Could out-of-order execution hit us here? I don't remember the specifics of that :(. @davidfowl ?

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -35,38 +38,84 @@ public SerializedHubMessage(HubMessage message)
...
         {
+            // We set the fields before moving on to the list, if we need it to hold more than 2 items.
+            // In order to prevent "shearing" (where some of the fields of the struct are set by one thread,
+            // while another thread is reading the struct), we atomically increment a counter to ensure
+            // consumers know which structs have been fully assigned.
+            
             if (_cachedItem1.ProtocolName == null)
             {
                 _cachedItem1 = new SerializedMessage(protocolName, serialized);
+                Interlocked.Increment(ref _count);
```
JamesNK: Is Interlocked.Increment required? SetCache is running under a lock so only one thread will access it at a time.

anurse: The read isn't under the lock though. But I guess that's OK too now that I think about it. So yeah, Interlocked not required :)

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -35,38 +37,84 @@ public SerializedHubMessage(HubMessage message)
...
+                        SetCache(protocol.Name, serialized);
+                    }
+                }
             }
 
             return serialized;
         }
 
+        // Used for unit testing.
+        internal IEnumerable<SerializedMessage> GetAllSerializations()
```
davidfowl: At first I was like wtf then I saw the comment 😄

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -35,38 +37,84 @@ public SerializedHubMessage(HubMessage message)
...
+            foreach (var item in _cachedItems)
+            {
+                yield return item;
+            }
+        }
+
         private void SetCache(string protocolName, ReadOnlyMemory<byte> serialized)
         {
+            // We set the fields before moving on to the list, if we need it to hold more than 2 items.
+            // In order to prevent "shearing" (where some of the fields of the struct are set by one thread,
```
davidfowl: "tearing"

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
JamesNK: i got you fam https://github.com/aspnet/SignalR/pull/2079#discussion_r182593089

test/Microsoft.AspNetCore.SignalR.Tests/SerializedHubMessageTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            var invocation = new InvocationMessage("Foo", new object[0]);
+            var message = new SerializedHubMessage(invocation);
+            var protocol = new DummyHubProtocol("p1");
+
+            // This should cache it
+            _ = message.GetSerializedMessage(protocol);
+
+            // Get it again
+            var serialized = message.GetSerializedMessage(protocol);
+
```
davidfowl: nit:whitespace...

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
anurse: I've definitely heard both terms but meh, I don't care

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
anurse: Sounds like @stephentoub confirmed my fears, that `_count` could be observed to be updated from the other thread before `_cachedItem` is observed to be updated. We're going to remove the "fast" path for now (we had locking here earlier, before my refactor that broke it). If we can get a correct lock-free path for reading the cache we can optimize later.

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
anurse: Also, we really should load up the list in the lock (even if this is only used in unit tests, let's do it right). Changing it to load up a list in the lock and then return that.

anurse:
:up: :date: took away the nice things

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -16,15 +16,19 @@ public class SerializedHubMessage
...
         private SerializedMessage _cachedItem2;
         private IList<SerializedMessage> _cachedItems;
+        private readonly object _lock = new object();
 
         public HubMessage Message { get; }
 
         public SerializedHubMessage(IReadOnlyList<SerializedMessage> messages)
         {
-            for (var i = 0; i < messages.Count; i++)
+            lock (_lock)
```
davidfowl: nit: Remove this lock

src/Microsoft.AspNetCore.SignalR.Core/SerializedHubMessage.cs
```diff
@@ -67,6 +104,9 @@ private void SetCache(string protocolName, ReadOnlyMemory<byte> serialized)
                     _cachedItems = new List<SerializedMessage>();
                 }
 
+                // No need to continue updating _count. It's just used to track the fields. The list
```
BrennanConroy: remove comment

Issue #2080, opened 2018-04-18
LongPolling connections can deadlock if pipe backpressures


BrennanConroy:
If enough messages are written to the connection then the pipe can backpressure
https://github.com/aspnet/SignalR/blob/b3b5a13f2c0ef1707278bdeaabfb0f977f19a49e/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs#L487

And it is sitting inside the lock
https://github.com/aspnet/SignalR/blob/b3b5a13f2c0ef1707278bdeaabfb0f977f19a49e/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs#L304

And when the LongPolling transport reads from the pipe and tries to re-poll it will try to acquire the lock which results in a deadlock
https://github.com/aspnet/SignalR/blob/b3b5a13f2c0ef1707278bdeaabfb0f977f19a49e/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs#L473

PR #2081, opened 2018-04-18
Split lock into WriteLock and StateLock to prevent deadlocks and unneeded lock contention

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2080

Tested this change with the load test runs and it produces the expected throughput instead of hanging.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
@@ -80,7 +80,8 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
 
         public HttpTransportType TransportType { get; set; }
 
-        public SemaphoreSlim Lock { get; } = new SemaphoreSlim(1, 1);
+        public SemaphoreSlim WriteLock { get; } = new SemaphoreSlim(1, 1);
+        public SemaphoreSlim StateLock { get; } = new SemaphoreSlim(1, 1);
 
         // Used for testing only
         internal Task DisposeAndRemoveTask { get; set; }
@@ -181,7 +182,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
 
             try
             {
-                await Lock.WaitAsync();
+                await StateLock.WaitAsync();
 
                 if (Status == HttpConnectionStatus.Disposed)
                 {
@@ -201,7 +202,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
             }
             finally
             {
-                Lock.Release();
+                StateLock.Release();
             }
 
             await disposeTask;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
```diff
@@ -179,10 +180,9 @@ public void TickHeartbeat()
         {
             var disposeTask = Task.CompletedTask;
 
+            await StateLock.WaitAsync();
```
JamesNK: Since we're only waiting on the statelock, does this mean that dispose can start while writes are in progress?

BrennanConroy:
Wrote a test!

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1797,6 +1797,60 @@ bool ExpectedErrors(WriteContext writeContext)
...
+                    context.Request.Method = "POST";
+                    var values = new Dictionary<string, StringValues>();
+                    values["id"] = connection.ConnectionId;
+                    var qs = new QueryCollection(values);
+                    context.Request.Query = qs;
+                    var buffer = Encoding.UTF8.GetBytes("Hello, world");
+                    requestBody.Write(buffer, 0, buffer.Length);
+                    requestBody.Seek(0, SeekOrigin.Begin);
+
+                    var t = loggerFactory.CreateLogger("testing");
```
davidfowl: Remove this......

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1797,6 +1797,60 @@ bool ExpectedErrors(WriteContext writeContext)
...
+
+                    // Write some data to the pipe to fill it up and make the next write wait
+                    await connection.ApplicationStream.WriteAsync(buffer, 0, buffer.Length).OrTimeout();
+
+                    // This will block until the pipe is unblocked
+                    var sendTask = dispatcher.ExecuteAsync(context, options, app).OrTimeout();
+                    Assert.False(sendTask.IsCompleted);
+
+                    var poll = MakeRequest("/foo", connection);
```
davidfowl: pollContext

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
davidfowl: We but we don't care about that race. It does mean the client will 500, which we could consider catching...

JamesNK: Ok, Could it be prevented full stop by using `await Task.WhenAll(StateLock.WaitAsync(), WriteLock.WaitAsync());` and then releasing both? Or do you think it is not worth the complexity?

BrennanConroy: "Nah" - @davidfowl 2018

PR #2085, opened 2018-04-18
Add connection id to sink log for unit tests

JamesNK:


test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -204,16 +205,18 @@ private class LogSinkLogger : ILogger
         {
             private readonly string _categoryName;
             private readonly LogSinkProvider _logSinkProvider;
+            private readonly LoggerExternalScopeProvider _scopeProvider;
```
anurse: @pakrym remind me how the external scope stuff works again? If we implement `ISupportsExternalScopes` don't we get some of this for free?

JamesNK: Basically I want to capture some information from the log scope. But the way I have done it means it only works when the information has been explicitly added to the ILogger that is currently writing the data. If an ILogger is created later with ILoggerFactory and it doesn't have the scope then the information is not available.

It has been hinted that there is some AsyncLocal magic to get around this issue?

pakrym: > You can then just use .ForEachScope on that in your Log method to get the scope data. And I think you don't have to implement BeginScope at all when you do that (i.e. it can just no-op and return null)

You are completely right.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -223,9 +225,43 @@ public bool IsEnabled(LogLevel logLevel)
...
+                            connectionId = id;
+                        }
+                    }
+                }, null);
+                return connectionId;
             }
         }
-    }
 
+        public void SetScopeProvider(IExternalScopeProvider scopeProvider)
```
pakrym: remove

pakrym: My bad, file layout confused me.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs
```diff
@@ -223,9 +225,43 @@ public bool IsEnabled(LogLevel logLevel)
 
             public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
             {
-                _logSinkProvider.Log(_categoryName, logLevel, eventId, state, exception, formatter);
+                var connectionId = GetConnectionId();
+
+                Func<TState, Exception, string> newFormatter = (s, e) =>
+                {
+                    StringBuilder sb = new StringBuilder();
+                    if (connectionId != null)
```
anurse: Good thing this is test code or @davidfowl would be exploding about the capture ;)

Issue #2084, opened 2018-04-18
A client can easily overwhelm the servers memory by constantly sending to a broadcast hub method


BrennanConroy:
We currently fire-and-forget hub invocations
https://github.com/aspnet/SignalR/blob/9a0c190f487b9f48eaaa83fb4f3c1675ee21bdc0/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs#L187
This can cause high lock contention and quickly grow the servers memory when many invocations are trying to write to the same connection. A quick test where we await the invocations shows stable low memory without affecting RPS in the load test that found this.

**Server Hub**

```c#
public class BroadcastHub : Hub
{
    public Task Broadcast(string message)
    {
        return Clients.All.SendAsync(message);
    }
}
```

**Client**

```c#
while (true)
{
    hubConnection.SendAsync("Broadcast", "flood server");
}
```

muratg:
Approved.

PR #2086, opened 2018-04-18
Run hub invocations serially

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2084

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
@@ -184,7 +184,7 @@ namespace Microsoft.AspNetCore.SignalR
                             {
                                 // Don't wait on the result of execution, continue processing other
                                 // incoming messages on this connection.
-                                _ = _dispatcher.DispatchMessageAsync(connection, message);
+                                await _dispatcher.DispatchMessageAsync(connection, message);
                             }
                         }
                         else if (result.IsCompleted)
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
@@ -164,9 +164,6 @@ namespace Microsoft.AspNetCore.SignalR
 
         private async Task DispatchMessagesAsync(HubConnectionContext connection)
         {
-            // Since we dispatch multiple hub invocations in parallel, we need a way to communicate failure back to the main processing loop.
-            // This is done by aborting the connection.
-
             try
             {
                 var input = connection.Input;
@@ -182,8 +179,6 @@ namespace Microsoft.AspNetCore.SignalR
                         {
                             while (protocol.TryParseMessage(ref buffer, _dispatcher, out var message))
                             {
-                                // Don't wait on the result of execution, continue processing other
-                                // incoming messages on this connection.
                                 await _dispatcher.DispatchMessageAsync(connection, message);
                             }
                         }
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
@@ -179,6 +179,8 @@ namespace Microsoft.AspNetCore.SignalR
                         {
                             while (protocol.TryParseMessage(ref buffer, _dispatcher, out var message))
                             {
+                                // Messages are dispatched sequentially and will block other messages from being processed until they complete.
+                                // Streaming methods will run sequentially until they start streaming, then they will fire-and-forget allowing other messages to run.
                                 await _dispatcher.DispatchMessageAsync(connection, message);
                             }
                         }
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
@@ -88,8 +88,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
                 case StreamInvocationMessage streamInvocationMessage:
                     Log.ReceivedStreamHubInvocation(_logger, streamInvocationMessage);
-                    // Fire-and-forget stream invocations, otherwise they would block other hub invocations from being able to run
-                    _ = ProcessInvocation(connection, streamInvocationMessage, isStreamedInvocation: true);
+                    await ProcessInvocation(connection, streamInvocationMessage, isStreamedInvocation: true);
                     break;
 
                 case CancelInvocationMessage cancelInvocationMessage:
@@ -171,7 +170,11 @@ namespace Microsoft.AspNetCore.SignalR.Internal
         {
             var methodExecutor = descriptor.MethodExecutor;
 
-            using (var scope = _serviceScopeFactory.CreateScope())
+            var disposeScope = true;
+            var scope = _serviceScopeFactory.CreateScope();
+            IHubActivator<THub> hubActivator = null;
+            THub hub = null;
+            try
             {
                 if (!await IsHubMethodAuthorized(scope.ServiceProvider, connection.User, descriptor.Policies))
                 {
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -184,7 +184,7 @@ public class HubConnectionHandler<THub> : ConnectionHandler where THub : Hub
                             {
                                 // Don't wait on the result of execution, continue processing other
                                 // incoming messages on this connection.
```
muratg: Not valid, remove.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -184,7 +184,7 @@ public class HubConnectionHandler<THub> : ConnectionHandler where THub : Hub
...
-                                _ = _dispatcher.DispatchMessageAsync(connection, message);
+                                await _dispatcher.DispatchMessageAsync(connection, message);
```
JamesNK: Does this mean only one message can be invoked at a time for a connection? What happens if a hub method takes 5 seconds to complete because it is calling a slow DB? Do all other messages wait for it to finish?

JamesNK: Scenario: I am connected to jabbr Core. I send David a private message containing a link to github. On the server jabbr parses the message and sees the github link, makes a request to github to include more details. At the same time I am sending messages in another chat room. Will those other messages only show up once the private message has finished being processed?

anurse: It does, but I believe this is also how ASP.NET SignalR works. This helps propagate back-pressure on the server back up through the client. If the server is backed up waiting to send data (in response to received messages), it slows down processing received messages, which eventually slows down the client, etc...

If you want to invoke a long-running thing in the background, you can fire and forget a task in the body of your Hub method and use `IHubContext<T>` to send a response to the Client via a Client invocation when the long-running task is complete.

JamesNK: Ok, as long as everyone who knows more about SignalR understands and expects that behavior. It doesn't seem like something can can easily be undone. If V1 runs messages serially then V2 won't be able to change to run in parallel without breaking some hearts.

I think it should be added as a comment here, and also spelled out in documentation. Devs are use to firing 10 requests at MVC and having them run in parallel.

JamesNK: Although I guess if there is load balancing and multiple instances of the SignalR server then messages could still end up running in parallel so people shouldn't rely on that logic.

davidfowl: Yes this is how old SignalR worked when you used websockets. Regardless this is a better way to handle dispatching. Running multiple calls to the hub in parallel means there's no self throttling.

JamesNK: ok!

JamesNK:
I'd like a couple of unit tests
1. Sequentially dispatching invoke messages
2. Multiple streams returning results and not blocking other messages being dispatched

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -1982,6 +1982,77 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+        {
+            var tcsService = new TcsService();
+            var serviceProvider = HubConnectionHandlerTestUtils.CreateServiceProvider(builder =>
+            {
+                builder.AddSingleton(tcsService);
+            });
+            var connectionHandler = serviceProvider.GetService<HubConnectionHandler<BlockingHub>>();
+
+            // Because we use PipeScheduler.Inline the hub invocations will run inline until they wait, which happens inside the BlockingMethod call
+            using (var client = new TestClient(new JsonHubProtocol()))
```
davidfowl: Isn't JsonHubProtocol the default?

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTestUtils/Hubs.cs
```diff
@@ -491,6 +493,34 @@ public class SimpleTypedHub : Hub<ITypedHubClient>
         }
     }
 
+    public class BlockingHub : Hub
+    {
+        private TcsService _tcs;
```
davidfowl: _tcsService not tcs

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTestUtils/Hubs.cs
```diff
@@ -491,6 +493,34 @@ public class SimpleTypedHub : Hub<ITypedHubClient>
...
+
+        public BlockingHub(TcsService tcs)
+        {
+            _tcs = tcs;
+        }
+
+        public async Task<int> BlockingMethod()
```
davidfowl: This isn't blocking...

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -1982,6 +1982,77 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler).OrTimeout();
+
+                // Long running hub invocation to test that other invocations will not run until it is completed
+                var longRunningTask = client.SendInvocationAsync(nameof(LongRunningHub.LongRunningMethod), nonBlocking: false);
+                // Wait for the long running method to start
+                await tcsService.StartedMethod.Task.OrTimeout();
+
+                // Invoke another hub method which will wait for the first method to finish
+                var nonBlockingTask = client.SendInvocationAsync(nameof(LongRunningHub.SimpleMethod), nonBlocking: false);
```
davidfowl: `Assert.False(nonBlockingTask.IsCompleted)`

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -88,7 +88,8 @@ public partial class DefaultHubDispatcher<THub> : HubDispatcher<THub> where THub
 
                 case StreamInvocationMessage streamInvocationMessage:
                     Log.ReceivedStreamHubInvocation(_logger, streamInvocationMessage);
-                    await ProcessInvocation(connection, streamInvocationMessage, isStreamedInvocation: true);
+                    // Fire-and-forget stream invocations, otherwise they would block other hub invocations from being able to run
+                    _ = ProcessInvocation(connection, streamInvocationMessage, isStreamedInvocation: true);
```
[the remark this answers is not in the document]
anurse: I mostly agree with @JamesNK here. Though to be fair, this isn't a full defense against bad code, there's still nothing stopping a poorly-written Hub method from `Task.Run`ing stuff. Still, if we `await` invocation of the method and then only fire-and-forget the logic to drain the stream, it would fit user expectations more I think.

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -227,51 +233,60 @@ public override IReadOnlyList<Type> GetParameterTypes(string methodName)
                     await SendInvocationError(hubMethodInvocationMessage.InvocationId, connection,
                         ErrorMessageHelper.BuildErrorMessage($"An unexpected error occurred invoking '{hubMethodInvocationMessage.Target}' on the server.", ex, _enableDetailedErrors));
                 }
-                finally
+            }
+            finally
+            {
+                if (disposeScope)
```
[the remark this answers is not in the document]
BrennanConroy: Haha

Issue #2069, opened 2018-04-18
flaky test: Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport


davidfowl:
```
[xUnit.net 00:00:14.4555490]     Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport(message: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"...) [FAIL]

  Failed   Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport(message: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"...)

[39;49m[31mRUNDOTNET : error Message:  [/Users/travis/.dotnet/buildtools/korebuild/2.1.0-preview3-17018/KoreBuild.proj]

[39;49m   System.Exception : 1 error(s) logged.

  Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionManager - FailedDispose - Failed disposing connection yTeG2nC1VbY0hQOwC6ydHA.

  ===================

  System.Net.WebSockets.WebSocketException (0x80004005): The WebSocket is in an invalid state ('CloseReceived') for this operation. Valid states are: 'Open, CloseSent'

     at System.Net.WebSockets.WebSocketValidate.ThrowIfInvalidState(WebSocketState currentState, Boolean isDisposed, WebSocketState[] validStates)

     at System.Net.WebSockets.ManagedWebSocket.ReceiveAsync(ArraySegment`1 buffer, CancellationToken cancellationToken)

  --- End of stack trace from previous location where exception was thrown ---

     at System.Net.WebSockets.WebSocket.ReceiveAsync(Memory`1 buffer, CancellationToken cancellationToken)

     at Microsoft.AspNetCore.Http.Connections.Internal.Transports.WebSocketsTransport.StartReceiving(WebSocket socket) in /_/src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/WebSocketsTransport.cs:line 147

     at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()

[...]

PR #2089, opened 2018-04-19
Global suppress low impact logging verification in tests

JamesNK:
Suppress low impact test failures from #2034 and #2069


JamesNK:
Change is to tests only so ask not required.

PR #2092, opened 2018-04-19
Reduce redundant dependencies (#2091)

davidfowl:


src/Microsoft.AspNetCore.Http.Connections.Client/Microsoft.AspNetCore.Http.Connections.Client.csproj
```diff
@@ -19,8 +19,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
-    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonPackageVersion)" />
```
JamesNK: Boo

anurse: Guess that means JSON.NET is redundant now :)

PR #2096, opened 2018-04-19
fix websocket sample

anurse:


samples/WebSocketSample/Program.cs
```diff
@@ -11,15 +11,24 @@ namespace WebSocketSample
 {
     public class Program
     {
-        public static void Main(string[] args)
+        public static int Main(string[] args)
```
BrennanConroy: async main 😝

anurse: I'll async you!

PR #2099, opened 2018-04-19
fix #2097 by adding GetHttpContext to HubConnectionContext

### LINKED ISSUE #2097: Add GetHttpContext extension method to HubConnectionContext
We have it on ConnectionContext and HubCallerContext, but not HubConnectionContext :).

As a workaround, you can use the following:

```csharp
var httpContext = context.Features.Get<IHttpContextFeature>()?.HttpContext;
```

anurse:
Also added some tests for `IUserIdProvider` that were missing, see comments inline.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR/GetHttpContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR/GetHttpContextExtensions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections.Features;
 
@@ -10,11 +11,19 @@ namespace Microsoft.AspNetCore.SignalR
     {
         public static HttpContext GetHttpContext(this HubCallerContext connection)
         {
+            if (connection == null)
+            {
+                throw new ArgumentNullException(nameof(connection));
+            }
             return connection.Features.Get<IHttpContextFeature>()?.HttpContext;
         }
 
         public static HttpContext GetHttpContext(this HubConnectionContext connection)
         {
+            if (connection == null)
+            {
+                throw new ArgumentNullException(nameof(connection));
+            }
             return connection.Features.Get<IHttpContextFeature>()?.HttpContext;
         }
     }
--- a/src/Microsoft.AspNetCore.SignalR/GetHttpContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.SignalR/GetHttpContextExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Http;
@@ -6,9 +6,14 @@ using Microsoft.AspNetCore.Http.Connections.Features;
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public static class HubCallerContextExtensions
+    public static class GetHttpContextExtensions
     {
         public static HttpContext GetHttpContext(this HubCallerContext connection)
+        {
+            return connection.Features.Get<IHttpContextFeature>()?.HttpContext;
+        }
+
+        public static HttpContext GetHttpContext(this HubConnectionContext connection)
         {
             return connection.Features.Get<IHttpContextFeature>()?.HttpContext;
         }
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.SignalR/GetHttpContextExtensions.cs
```diff
@@ -1,16 +1,21 @@
...
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections.Features;
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public static class HubCallerContextExtensions
+    public static class GetHttpContextExtensions
```
anurse: It felt unnecessary to create a whole new class just to duplicate this extension method. I'm OK splitting this out into two again if people really don't like this.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -1705,6 +1705,57 @@ public static IEnumerable<object[]> StreamingMethodAndHubProtocols
             }
         }
 
+        [Fact]
+        public async Task ConnectionUserIdIsAssignedByUserIdProvider()
```
anurse: `IUserIdProvider` never had a real test here...

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -864,15 +864,14 @@ bool ExpectedErrors(WriteContext writeContext)
             }
         }
 
-        [Theory]
-        [MemberData(nameof(TransportTypes))]
-        public async Task CheckHttpConnectionFeatures(HttpTransportType transportType)
+        [Fact]
+        public async Task CheckHttpConnectionFeatures()
```
anurse: I didn't want the test I was adding (below) to vary by transport because it seemed unnecessary. This was another test that was very similar (not testing anything transport-specific) so I just removed the parameter from it as well. If people things we should revert this, I should make my test vary by transport too.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HeaderUserIdProvider.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.AspNetCore.SignalR.Client.FunctionalTests
+{
+    internal class HeaderUserIdProvider : IUserIdProvider
+    {
+        public static readonly string HeaderName = "Super-Secure-UserName";
+
+        public string GetUserId(HubConnectionContext connection)
+        {
+            // Super-secure user id provider :)
```
davidfowl: Don't let Barry see this 😄

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Startup.cs
```diff
@@ -25,6 +25,7 @@ public void ConfigureServices(IServiceCollection services)
                 options.EnableDetailedErrors = true;
             })
                 .AddMessagePackProtocol();
+            services.AddSingleton<IUserIdProvider, HeaderUserIdProvider>();
```
davidfowl: If we ever want to test something else, we're fucked 😄

anurse: Not really...

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HeaderUserIdProvider.cs
anurse: I'll rename the header and comments just in case someone doesn't understand my sarcasm ;)

anurse:
This is all part of API reviews of SignalR, which was pre-approved by Shiproom

src/Microsoft.AspNetCore.SignalR/GetHttpContextExtensions.cs
```diff
@@ -1,16 +1,21 @@
...
     {
         public static HttpContext GetHttpContext(this HubCallerContext connection)
         {
             return connection.Features.Get<IHttpContextFeature>()?.HttpContext;
         }
+
+        public static HttpContext GetHttpContext(this HubConnectionContext connection)
+        {
+            return connection.Features.Get<IHttpContextFeature>()?.HttpContext;
```
JamesNK: Check `connection` argument is not null? +other overload

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
JamesNK: 👍

src/Microsoft.AspNetCore.SignalR/GetHttpContextExtensions.cs
anurse: Sure, but it should `ArgumentNullException` there. I'll add that.

davidfowl:
@anurse isn't this for the release branch.

anurse:
Yep, derp.

PR #2101, opened 2018-04-19
CancellationToken EVERYWHERE

analogrelay:
* Add CancellationToken to all the public APIs that might be cancelable in the future.
* Also cleaned up some extra usings as I was scanning through files
* Removed unnecessary parameter from `HubConnection+ConnectionState.StopAsync`

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -6,6 +6,7 @@ using System.Buffers;
 using System.IO;
 using System.IO.Pipelines;
 using System.Reactive.Linq;
+using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
@@ -82,7 +83,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
             {
             }
 
-            public override ValueTask WriteAsync(HubMessage message)
+            public override ValueTask WriteAsync(HubMessage message, CancellationToken cancellationToken)
             {
                 if (message is CompletionMessage completionMessage)
                 {


src/Microsoft.AspNetCore.SignalR.Core/ClientProxyExtensions.cs
```diff
@@ -13,10 +12,11 @@ public static class ClientProxyExtensions
         /// </summary>
         /// <param name="clientProxy">The <see cref="IClientProxy"/></param>
         /// <param name="method">name of the method to invoke</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
         /// <returns>A task that represents when the data has been sent to the client.</returns>
-        public static Task SendAsync(this IClientProxy clientProxy, string method)
+        public static Task SendAsync(this IClientProxy clientProxy, string method, CancellationToken cancellationToken = default)
```
anurse: This one may be debatable. The `SendAsync` on the client can be canceled. I think these APIs should mirror as much as possible (particularly useful when Client Hubs become a thing)

src/Microsoft.AspNetCore.SignalR.Core/HubLifetimeManager.cs
```diff
@@ -2,37 +2,40 @@
...
+        public abstract Task SendUserAsync(string userId, string methodName, object[] args, CancellationToken cancellationToken = default);
 
-        public abstract Task SendUsersAsync(IReadOnlyList<string> userIds, string methodName, object[] args);
+        public abstract Task SendUsersAsync(IReadOnlyList<string> userIds, string methodName, object[] args, CancellationToken cancellationToken = default);
 
-        public abstract Task AddToGroupAsync(string connectionId, string groupName);
+        public abstract Task AddToGroupAsync(string connectionId, string groupName, CancellationToken cancellationToken = default);
 
-        public abstract Task RemoveFromGroupAsync(string connectionId, string groupName);
+        public abstract Task RemoveFromGroupAsync(string connectionId, string groupName, CancellationToken cancellationToken = default);
```
anurse: Since these are all called by methods that just got CTs (`IClientProxy.Send` and `IGroupManager.AddToGroupAsync/RemoveFromGroupAsync`) the lifetime manager should get to observe cancellation if it wants to.

anurse:
Part of pre-approved API reviews

src/Microsoft.AspNetCore.SignalR.Client.Core/IConnectionFactory.cs
```diff
@@ -1,15 +1,18 @@
...
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
     public interface IConnectionFactory
     {
-        Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat);
+        Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, CancellationToken cancellationToken = default);
```
JamesNK: I'm not sure if it should have a default on interfaces. Should look and see what async APIs with interfaces do in corefx.

src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionFactory.cs
```diff
@@ -31,10 +32,10 @@ public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFac
             _loggerFactory = loggerFactory;
         }
 
-        public async Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat)
+        public async Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, CancellationToken cancellationToken)
```
JamesNK: No default? Add a comment explaining why.

src/Microsoft.AspNetCore.SignalR.Core/IClientProxy.cs
```diff
@@ -17,7 +18,8 @@ public interface IClientProxy
         /// </summary>
         /// <param name="method">Name of the method to invoke.</param>
         /// <param name="args">A collection of arguments to pass to the client.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
         /// <returns>A task that represents when the data has been sent to the client.</returns>
-        Task SendCoreAsync(string method, object[] args);
+        Task SendCoreAsync(string method, object[] args, CancellationToken cancellationToken = default);
```
JamesNK: See earlier comment about default's on interfaces

src/Microsoft.AspNetCore.SignalR.Core/IGroupManager.cs
```diff
@@ -1,13 +1,14 @@
...
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.SignalR
 {
     public interface IGroupManager
     {
-        Task AddToGroupAsync(string connectionId, string groupName);
-        Task RemoveFromGroupAsync(string connectionId, string groupName);
+        Task AddToGroupAsync(string connectionId, string groupName, CancellationToken cancellationToken = default);
+        Task RemoveFromGroupAsync(string connectionId, string groupName, CancellationToken cancellationToken = default);
```
JamesNK: +1

src/Microsoft.AspNetCore.SignalR.Client.Core/IConnectionFactory.cs
anurse: I think it's fine, `Foo(int bar = 4)` and `Foo(int bar)` are identical signatures at the IL level. The default value is basically just a hint to compilers that support it. We want this API to take a CancellationToken, so we have it on our API. The default value is just a useful thing to have.

The issues arise when you treat optional parameters like overloads. Adding an optional parameter to an existing API is a breaking change, for example.

I'll look to see if I can find prior art, and I'm OK removing the default (which is the alternative) if people are concerned.

src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionFactory.cs
anurse: just missed it

src/Microsoft.AspNetCore.SignalR.Client.Core/IConnectionFactory.cs
davidfowl: It's fine and part of our guidelines

davidfowl:
Ugh

davidfowl:
ugh, I guess... 😄

anurse:
Is that Approval or not? I want to get this in ASAP today.

src/Microsoft.AspNetCore.SignalR.Core/Internal/TypedClientBuilder.cs
```diff
@@ -18,6 +19,8 @@ internal static class TypedClientBuilder<T>
         // There is one static instance of _builder per T
         private static readonly Lazy<Func<IClientProxy, T>> _builder = new Lazy<Func<IClientProxy, T>>(() => GenerateClientBuilder());
 
+        private static readonly PropertyInfo CancellationTokenNoneProperty = typeof(CancellationToken).GetProperty("None", BindingFlags.Public | BindingFlags.Static);
```
davidfowl: Did you test this change?

davidfowl:
> Is that Approval or not? I want to get this in ASAP today. (I'll fix the conflict shortly)

I'd like to see a test for the ref emit case that works with and without a token. That's the only scary part of this change.

anurse:
What do you mean? All the ref emit tests hit this code path. I **have not** added support for the client proxy type including a `CancellationToken` at this time, so there's no new scenario here, it's just that the method the generated proxy calls now has an additional parameter that needs to be set.

davidfowl:
I see, the method bodies changed, but you don't support flowing it through.

davidfowl:
That shouldn't cause any issues supporting it later right?

anurse:
I could have done that in this change, but wanted to keep it as limited as possible. We'd have to figure out exactly how it should flow too (is it just if there's a CT argument? does it have to be the last argument? what if there are more than one?). It's not hard to implement, I just didn't think we should implement it now. I only really remembered to update this code because all the tests for it failed ;)

anurse:
> That shouldn't cause any issues supporting it later right?

I don't believe so. If we wanted to support it now, we'd also want to make it optional. So we'd need to look at the method we're implementing, check for a CT and if it has one, we'd decide it should be used for this purpose, rather than treating it as an argument to be sent to the client. That logic doesn't really change from 1.0 to 1.1. A CT doesn't make sense as a normal argument, so making the decision in 1.1 to treat CTs in the argument list as tokens for cancelling the send itself seems safe.

davidfowl:
So today its just broken because we don't even detect a CT in the signature. Cool.

anurse:
Conflicts resolved (though git didn't have any conflicts... I just rebased and pushed and GitHub was happy). Ready to merge with Approval

davidfowl:
@anurse there's another conflict for ya 😏

anurse:
Simple one though, you deleted a file I changed.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2181,6 +2181,15 @@ public void Release(THub hub)
             }
         }
 
+        public class AnotherTypedHub : Hub<IAnotherTypedHubClient>
+        {
+        }
+
+        public interface IAnotherTypedHubClient
+        {
+            void Foo();
```
davidfowl: This will blow up after the other change 😄

src/Microsoft.AspNetCore.SignalR.Core/Internal/Proxies.cs
```diff
@@ -17,7 +18,7 @@ public UserProxy(HubLifetimeManager<THub> lifetimeManager, string userId)
             _userId = userId;
         }
 
-        public Task SendCoreAsync(string method, object[] args)
+        public Task SendCoreAsync(string method, object[] args, CancellationToken cancellationToken = default)
         {
             return _lifetimeManager.SendUserAsync(_userId, method, args);
```
BrennanConroy: These methods all now take the `cancellationToken`

src/Microsoft.AspNetCore.SignalR.Core/ClientProxyExtensions.cs
```diff
@@ -1,6 +1,5 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+using System;
```
BrennanConroy: copyright

src/Microsoft.AspNetCore.SignalR.Core/Internal/GroupManager.cs
```diff
@@ -0,0 +1,28 @@
...
+        private readonly HubLifetimeManager<THub> _lifetimeManager;
+
+        public GroupManager(HubLifetimeManager<THub> lifetimeManager)
+        {
+            _lifetimeManager = lifetimeManager;
+        }
+
+        public Task AddToGroupAsync(string connectionId, string groupName, CancellationToken cancellationToken = default)
+        {
+            return _lifetimeManager.AddToGroupAsync(connectionId, groupName);
```
BrennanConroy: pass token in

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
anurse: Yep, hoping to get that one in and then rebase this on top.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -64,6 +64,7 @@ public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAl
         // Currently used only for streaming methods
         internal ConcurrentDictionary<string, CancellationTokenSource> ActiveRequestCancellationSources { get; } = new ConcurrentDictionary<string, CancellationTokenSource>(StringComparer.Ordinal);
 
+        // REVIEW: Cancellable? The caller has no way to cancel without it (they can't CancelPendingFlush).
         public virtual ValueTask WriteAsync(HubMessage message)
```
JamesNK: Was a decision made one way or the other on this method?

anurse: Ah no, good point. Paging @davidfowl.

I lean towards adding a token because normally you'd be able to cancel the backpressure wait on 
a pipe using `CancelPendingFlush`, but the person who calls this method doesn't have access to the pipe. They could abort the connection I suppose.

anurse: Following our plan of doing the minimal work for RC, I'm going to put these on the API so they're there in RC. We don't have to do anything with them until we've decided exactly how we want them to work.

PR #2105, opened 2018-04-20
privatize IConnection and HttpConnection

anurse:
Makes HttpConnection private. We're not able to get the API into the shape we want and it's only necessary for accessing ConnectionHandlers directly, so we'll make it private until we can tidy up the API.


clients/ts/FunctionalTests/ts/ConnectionTests.ts
```diff
@@ -1,148 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-import { HttpConnection, HttpTransportType, IHttpConnectionOptions, LogLevel, TransferFormat } from "@aspnet/signalr";
```
davidfowl: We shouldn't be deleting any tests.

anurse:
For 2.1, I'll just import `HttpConnection.ts` into this test directly by path. It invalidates the test as far as testing any user scenario but it should keep the other tests clean. After 2.1 we'll have unit tests for this type that are more appropriate.

I don't think it's relevant to do functional testing with a type we don't support. I think it's valuable to have tests for this component, as with any of our tests. Now that `HttpConnection` is an implementation detail, it's weird to test it in this way, but I realize we don't really have any other tests.

Please make sure you've reviewed there rest of this so that I can merge this shortly after making that change, I have some mechanical TypeScript clean-up (TSLinting, cleaning up imports, etc.) I want to do before the Ask Mode deadline.

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -12,8 +12,6 @@ import { IStreamResult } from "./Stream";
 import { TextMessageFormat } from "./TextMessageFormat";
 import { Arg, createLogger, Subject } from "./Utils";
 
-export { JsonHubProtocol };
```
davidfowl: Why?

davidfowl:
> I don't think it's relevant to do functional testing with a type we don't support. I think it's valuable to have tests for this component, as with any of our tests. Now that HttpConnection is an implementation detail, it's weird to test it in this way, but I realize we don't really have any other tests.

It is because we're going to support this type, we just ran out of time so lets keep them. The direct file import is good.

Everything else looks pretty benign. I do have a question about removing the export the JsonHubProtocol from the HubConnection.ts. What are the implications there?

anurse:
I just moved it. It was weird that it was imported and then re-exported from `HubConnection`, so I removed it from there and export it directly out of `index.`

Issue #2107, opened 2018-04-20
Make Hub and Hub<T> abstract


JamesNK:
There is no reason for these types to be directly created.

davidfowl:
This is kinda now or never...

PR #2108, opened 2018-04-20
Change Hub and friends to abstract

JamesNK:
https://github.com/aspnet/SignalR/issues/2107

Does this fall under API review banner? It needs to be done before release or never.


davidfowl:
Yes it does

PR #2113, opened 2018-04-20
Removing the chat sample

davidfowl:
- People are copying the presence prototype as if it's recommended instead of treating it like a prototype. Removing this because we want to do something more integrated in the future.


moozzyk:
So much sweat put into this lua and it is now 🌬 

😉

jedjohan:
Oh no. Used it as a recommendation. And now it´s all gone. Any hints on what "more integrated" means ? Need to keep inmemory list of connections and that sample was great as a base 😀

davidfowl:
> Oh no. Used it as a recommendation. And now it´s all gone. Any hints on what "more integrated" means ? Need to keep inmemory list of connections and that sample was great as a base 😀

Feel free to copy it .

PR #2117, opened 2018-04-20
fix #2116 by requiring task return values from typed client

### LINKED ISSUE #2116: Ensure that interfaces used in Hub<T> always return Task for methods
Today we support both void returning methods and Task returning methods, we should change this to always require a task returning method. Today we fire and forget if the method is void returning. This leads to unobserved Task exceptions in your code if for example you were using redis and it went down.

anurse:
fixes #2116 

We now throw a useful error if you have a `void`-returning method in your TypedHubClient proxy.


test/Microsoft.AspNetCore.SignalR.Tests/Internal/TypedClientBuilderTests.cs
```diff
@@ -0,0 +1,220 @@
+using System;
```
BrennanConroy: copy

anurse:
image

Issue #2135, opened 2018-04-23
C# Client: Don't hop off the event loop for message dispatching


BrennanConroy:
https://github.com/aspnet/SignalR/pull/1952

Hopping off the event loop for every message is very costly. We should stay on the event loop and document that users should be careful of blocking in their callbacks.

PR #2136, opened 2018-04-23
Remove unnecessary hop

BrennanConroy:
#2135


JamesNK:
Is it unit testable that we're already on the thread pool? 😸

BrennanConroy:
I wrote something for a test, not 100% sure it is accurate

PR #2137, opened 2018-04-23
fix #2134 by disposing httpconnection if start fails

### LINKED ISSUE #2134: HttpClient isn't disposed if transport fails to start when using HttpConnection and HubConnection
We don't dispose the HttpClient if the transport fails to Start. This could cause resource leaks when using HubConnection in a retry loop since it can't call DisposeAsync because it doesn't even have a reference to the ConnectionContext if ConnectAsync fails to start.

anurse:
I started the test, need to actually run it and verify it :).

The fix is good though and I wanted to get the PR up.

Fixes #2134

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionFactory.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionFactory.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             {
                 throw new ArgumentNullException(nameof(loggerFactory));
             }
-            
+
             _httpConnectionOptions = options.Value;
             _loggerFactory = loggerFactory;
         }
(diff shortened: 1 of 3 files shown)


davidfowl:
CI is failing for your test

anurse:
Yeah, I think I know what's up.

anurse:
I hadn't really run the test, I just wanted to get the PR up for shiproom ;P

anurse:
```
11:21:47   The active test run was aborted. Reason: Unexpected character encountered while parsing value: i. Path '', line 0, position 0.
11:21:48   
11:21:48   Total tests: Unknown. Passed: 1174. Failed: 0. Skipped: 11.
11:21:48   Test Run Aborted.
11:21:48   Test execution time: 35.2068 Seconds
```

the Jenkins build is failing with https://github.com/Microsoft/vstest/issues/1542 too now.

anurse:
@dotnet-bot test this please

anurse:
The plan is to merge this to `dev` and cherry pick it back in to `release/2.1` when the branch opens for RTM

src/Microsoft.AspNetCore.SignalR.Client/HttpConnectionFactory.cs
```diff
@@ -27,16 +27,25 @@ public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFac
...
-            return connection;
+            try
+            {
+                await connection.StartAsync(transferFormat, cancellationToken);
+                return connection;
+            }
+            catch
+            {
+                // Make sure the connection is disposed, in case it allocated any resources before failing.
+                await connection.DisposeAsync();
```
JamesNK: Should this be wrapped in its own try catch? If an error is thrown out of DisposeAsync it will hide the error that originally came from StartAsync

davidfowl: This never throws 😆  (famous last words)

anurse: A `Dispose` that throws an exception (other than an execution-engine exception like `StackOverflowException`) is an incorrectly-written `Dispose`.

PR #2143, opened 2018-04-25
EchoAll scenario doesn't work well with warmup

BrennanConroy:
Warmup runs will fill up the connection pipes on the server and when the connections are closed the server will still be processing messages because that's what we do. Then the main run will start while the warmup messages are still running, ruining our benchmark results


BrennanConroy:
https://github.com/aspnet/SignalR/issues/2144

Issue #2131, opened 2018-04-22
Minor API related things found while documenting


JamesNK:
Some minor API things I have noticed while documenting. I don't know if these meet the change bar.

IHubClientsExtensions -> should be named HubClientsExtensions. (since it itself is not an interface)

DynamicHubClients.Users(IReadOnlyList<string> users) -> should be userIds for consistency

JamesNK:
Should `SerializedHubMessage` and `SerializedMessage` located in Core be under the `Microsoft.AspNetCore.SignalR.Protocol` namespace?

JamesNK:
Was there a rule about public types exposing pub-internal types? `ConnectionsRouteBuilder` takes the pub-internal `HttpConnectionDispatcher` in its public ctor

JamesNK:
`RedisDependencyInjectionExtensions.AddRedis` overloads, should rename `builder` to `signalrBuilder` to be consistent with other extension methods.

JamesNK:
`SignalRDependencyInjectionExtensions.AddHubOptions` should rename `options` parameter to `configure` to be consistent with other configuration callbacks.

anurse:
> Not ideal but the ctor is for DI so no change required.

How is it for DI? It's constructed directly in the [one place it's used](https://github.com/aspnet/SignalR/search?utf8=%E2%9C%93&q=ConnectionsRouteBuilder&type=), which is also inside the same assembly. Could the constructor just be internal?

JamesNK:
You're right. I thought `HttpConnectionDispatcher` was passed in via DI but it is resolved manually and then the ctor is used. Will change it to internal.

PR #2147, opened 2018-04-26
Minor API cleanup

JamesNK:
https://github.com/aspnet/SignalR/issues/2131

Waiting on approval for merge into 2.1 RTM.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/ConnectionsRouteBuilder.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/ConnectionsRouteBuilder.cs
@@ -18,12 +18,7 @@ namespace Microsoft.AspNetCore.Http.Connections
         private readonly HttpConnectionDispatcher _dispatcher;
         private readonly RouteBuilder _routes;
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="ConnectionsRouteBuilder"/> class.
-        /// </summary>
-        /// <param name="routes">The underlying <see cref="RouteBuilder"/>.</param>
-        /// <param name="dispatcher">The dispatcher.</param>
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.SignalR.Core/Protocol/SerializedHubMessage.cs
```diff
@@ -3,9 +3,8 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.AspNetCore.SignalR.Protocol;
 
-namespace Microsoft.AspNetCore.SignalR
+namespace Microsoft.AspNetCore.SignalR.Protocol
```
davidfowl: Why is this in protocol?

src/Microsoft.AspNetCore.SignalR.Core/Protocol/SerializedMessage.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.AspNetCore.SignalR
+namespace Microsoft.AspNetCore.SignalR.Protocol
```
davidfowl: Why is this in protocol?

JamesNK: All the other message classes are in `Microsoft.AspNetCore.SignalR.Protocol`. It feels weird that these "internalish only" infrastructure classes for messages are in the root SignalR namespace.

davidfowl: This isn't a message type. It's sits above all protocols and that's why it's in the root. There's a public WriteAsync on HubConnectionContext that takes this in.

Issue #2148, opened 2018-04-26
We never set HttpConnection.ConnectionId


davidfowl:
Makes it really hard to do logging that shows the connection id (Impossible 😏).

anurse:
https://github.com/aspnet/SignalR/commit/41c8dcf44972363289add219ae7b2fa7575c38e9

PR #2154, opened 2018-04-27
Fix not setting HttpConnection.ConnectionId

JamesNK:
https://github.com/aspnet/SignalR/issues/2148

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
@@ -82,6 +82,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Client
             set => throw new InvalidOperationException("The ConnectionId is set internally and should not be set by user code.");
         }
 
+        /// <inheritdoc />
         public override IDictionary<object, object> Items { get; set; } = new ConnectionItems();
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -62,7 +62,14 @@ public override IDuplexPipe Transport
         }
 
         public override IFeatureCollection Features { get; } = new FeatureCollection();
-        public override string ConnectionId { get; set; }
+
+        // ConnectionId will be null if negotate was skipped
+        public override string ConnectionId
+        {
+            get => _connectionId;
+            set => throw new InvalidOperationException("The ConnectionId is set internally when the HttpConnection starts by negotate.");
```
anurse: "The ConnectionId is set internally and should not be set by user code" (no need to go into detail about how it's set :))

anurse: This being our last chance to do so, should we consider making `ConnectionContext.ConnectionId` read-only? We generally make these properties read-write for testability (following in `HttpContext`s footsteps), but this is a bit of a special case.

JamesNK: It's get/set on `ConnectionContext`.

JamesNK: > (no need to go into detail about how it's set :))

Devs might wonder about when it is set. Also as far as I can tell it is only set when there is a negotiate. If skipNegotiate is true then it will be null.

davidfowl: No it shouldn’t be readonly. I’m not sure about throwing either

anurse: > If skipNegotiate is true then it will be null.

That's correct but it's "by-design" in that no connection ID is needed when using a direct WebSocket connection (since it's handled by a single request)

davidfowl: Yep!

PR #2162, opened 2018-04-29
Simplify error handling in HubConnectionHandler

davidfowl:
- Since we execute hub methods inline, there's no need to abort the connection on unexpected exceptions.
- Don't pass the cancellation token to ReadAsync, instead just use CancelPendingRead.
- Don't treat OperationCancelledException errors as errors.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
@@ -100,7 +100,12 @@ namespace Microsoft.AspNetCore.SignalR
             {
                 await DispatchMessagesAsync(connection);
             }
-            catch (Exception ex) when (!(ex is OperationCanceledException))
+            catch (OperationCanceledException)
+            {
+                // Don't treat OperationCanceledException as an error, it's basically a "control flow"
+                // exception to stop things from running
+            }
+            catch (Exception ex)
             {
                 Log.ErrorProcessingRequest(_logger, ex);
 
@@ -177,13 +182,11 @@ namespace Microsoft.AspNetCore.SignalR
                     {
                         break;
                     }
-                    
+
                     if (!buffer.IsEmpty)
                     {
                         while (protocol.TryParseMessage(ref buffer, _dispatcher, out var message))
                         {
-                            // Messages are dispatched sequentially and will block other messages from being processed until they complete.
-                            // Streaming methods will run sequentially until they start streaming, then they will fire-and-forget allowing other messages to run.
                             await _dispatcher.DispatchMessageAsync(connection, message);
                         }
                     }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -100,7 +100,7 @@ public class HubConnectionHandler<THub> : ConnectionHandler where THub : Hub
             {
                 await DispatchMessagesAsync(connection);
             }
-            catch (Exception ex)
+            catch (Exception ex) when (!(ex is OperationCanceledException))
```
JamesNK: Everywhere else we explicitly catch `OperationCanceledException` and don't do anything with it. Should stick with that for consistency (also it gives a nice space for a comment explaining why canceled exception is gnored)

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -142,26 +142,18 @@ public override IReadOnlyList<Type> GetParameterTypes(string methodName)
         private async Task ProcessInvocation(HubConnectionContext connection,
             HubMethodInvocationMessage hubMethodInvocationMessage, bool isStreamedInvocation)
         {
-            try
+            // If an unexpected exception occurs then we want to kill the entire connection
+            // by ending the processing loop
```
JamesNK: Is this comment still true?

JamesNK:
Unit tests 😱

davidfowl:
Unit tests for what?

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
davidfowl: Yes we’re just changing how his happens.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
davidfowl: Sure

JamesNK:
I think this is important behavior and should be unit tested:
* Since we execute hub methods inline, there's no need to abort the connection on unexpected exceptions.

davidfowl:
I think you misunderstood. The behavior didn’t change. It’s just simplified. Before we had to abort this way because it was fire and forget. Now we can remove the complex control flow and any exception in DispatchMessageAsync will throw inline and stop the loop.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -164,44 +164,41 @@ public class HubConnectionHandler<THub> : ConnectionHandler where THub : Hub
...
-                        else if (result.IsCompleted)
+                        break;
+                    }
+                    
+                    if (!buffer.IsEmpty)
+                    {
+                        while (protocol.TryParseMessage(ref buffer, _dispatcher, out var message))
                         {
-                            break;
+                            // Messages are dispatched sequentially and will block other messages from being processed until they complete.
```
davidfowl: This comment should really be moved into the DefaultHubDispatcher

davidfowl:

PR #2166, opened 2018-04-30
Don't throw from AbortAsync

davidfowl:
- Log from inside of HubConnectionContext if the user callback failed.
- Use ThreadPool.QUWI instead of Task.Factory.StartNew.
- Remove try catch from HubConnectionHandler

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -22,7 +22,7 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class HubConnectionContext
     {
-        private static readonly Action<object> _abortedCallback = AbortConnection;
+        private static readonly WaitCallback _abortedCallback = AbortConnection;
 
         private readonly ConnectionContext _connectionContext;
         private readonly ILogger _logger;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -271,7 +271,7 @@ public virtual void Abort()
             Input.CancelPendingRead();
 
             // We fire and forget since this can trigger user code to run
-            Task.Factory.StartNew(_abortedCallback, this);
+            ThreadPool.QueueUserWorkItem(_abortedCallback, this);
```
anurse: For my future reference, what's the reason for doing this over `Task.Factory.StartNew`?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -495,7 +498,12 @@ public static void ProtocolVersionFailed(ILogger logger, string protocolName, in
             {
                 _protocolVersionFailed(logger, protocolName, version, null);
             }
-        }
 
+            public static void AbortFailed(ILogger logger, Exception exception)
+            {
+                _abortFailed(logger, exception);
+            }
+
```
anurse: nit: extra line

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
davidfowl: 2 reasons:
- Use the global queue instead of the local queue (which means it's more "fair" for other scheduled work items). This ends up being faster most of the time.
- Don't need to allocate a task because this is fire and forget

PR #2167, opened 2018-04-30
Remove unnecessary state machines in DefautHubDispatcher

davidfowl:
- Remove state machines for async tail calls
- Added fast path for auth check when there are no policies

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
@@ -73,7 +73,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
             }
         }
 
-        public override async Task DispatchMessageAsync(HubConnectionContext connection, HubMessage hubMessage)
+        public override Task DispatchMessageAsync(HubConnectionContext connection, HubMessage hubMessage)
         {
             // Messages are dispatched sequentially and will stop other messages from being processed until they complete.
             // Streaming methods will run sequentially until they start streaming, then they will fire-and-forget allowing other messages to run.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Core/Internal/TaskCache.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    internal static class TaskCahe
```
anurse: Le Task Cahé

davidfowl: Fixed!

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -118,6 +115,8 @@ public partial class DefaultHubDispatcher<THub> : HubDispatcher<THub> where THub
                     Log.UnsupportedMessageReceived(_logger, hubMessage.GetType().FullName);
                     throw new NotSupportedException($"Received unsupported message: {hubMessage}");
             }
+
+            return Task.CompletedTask;
```
JamesNK: nit: What about making PingMessage and CancelInvocationMessage return completed task rather than breaking. Means this return can be removed

davidfowl:

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -323,14 +322,19 @@ private void InitializeHub(THub hub, HubConnectionContext connection)
             hub.Groups = _hubContext.Groups;
         }
 
-        private async Task<bool> IsHubMethodAuthorized(IServiceProvider provider, ClaimsPrincipal principal, IList<IAuthorizeData> policies)
+        private Task<bool> IsHubMethodAuthorized(IServiceProvider provider, ClaimsPrincipal principal, IList<IAuthorizeData> policies)
```
anurse: nit: This (and the Slow variant) should probably have an `Async` suffix for consistency... though they didn't have that before this PR so 🤷‍♂️

Issue #2048, opened 2018-04-16
TypeScript clean-up and testing improvements


anurse:
* Overhaul TypeScript Functional tests to use promises
* Mock out things like WebSocket/EventSource so they can be tested

etc.

anurse:
Things to do here:

* Add `WebSocket` and `EventSource` constructors to `IHttpConnectionOptions` as Internal (`public` but with `/** @internal */` comment)
* If those aren't set, **and** `window` is defined (because we're in a browser) initialize these to `window.WebSocket`/`window.EventSource`
* Flow those through to the transports
* Write a proof-of-concept unit test using mock `WebSocket` and `EventSource`

I'll file a separate issue for expanding our unit tests.

PR #2170, opened 2018-04-30
Fix #2169 by correcting shutdown timeout

### LINKED ISSUE #2169: LongPollingTransport shutdown timer is configured incorrectly
https://github.com/aspnet/SignalR/blob/c53514fa19bbb776ee3855a6cfaa1c4706b7f446/clients/ts/signalr/src/LongPollingTransport.ts#L182-L185

[`setTimeout`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) takes the callback as the **first** argument

anurse:
Still thinking about how best to test the Long Polling shutdown timeout case. We don't have any test coverage here yet because of pending #2048 .

Fixes #2169

### DIFF
--- a/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -577,13 +577,8 @@ describe("hubConnection", () => {
                     const hubConnection = getConnectionBuilder(transportType).build();
                     hubConnection.serverTimeoutInMilliseconds = 100;
 
-                    const timeout = setTimeout(200, () => {
-                        fail("Server timeout did not fire within expected interval");
-                    });
-
                     hubConnection.start().then(() => {
                         hubConnection.onclose((error) => {
-                            clearTimeout(timeout);
                             expect(error).toEqual(new Error("Server timeout elapsed without receiving a message from the server."));
                             done();
                         });
--- a/clients/ts/signalr/spec/LongPollingTransport.spec.ts
+++ b/clients/ts/signalr/spec/LongPollingTransport.spec.ts
@@ -1,0 +1,67 @@
+import { HttpResponse } from "../src/HttpClient";
+import { LogLevel } from "../src/ILogger";
+import { TransferFormat } from "../src/ITransport";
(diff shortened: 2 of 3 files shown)


JamesNK:
Can we add a unit test the verifies that shutdown timeout works? 😸

anurse:
It's never called in a unit test right now. I think I can actually bash out a test for this. Stand by

anurse:
:up: :date: Unit test!

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -577,13 +577,8 @@ describe("hubConnection", () => {
                     const hubConnection = getConnectionBuilder(transportType).build();
                     hubConnection.serverTimeoutInMilliseconds = 100;
 
-                    const timeout = setTimeout(200, () => {
```
anurse: This was an attempt to get the test to fail sooner than Jasmine's own timeout, but I think it's cleaner to just let Jasmine do it's thing.

clients/ts/signalr/spec/LongPollingTransport.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+                    const abort = new Promise((resolve, reject) => r.abortSignal.onabort = resolve);
+                    await abort;
+
+                    // Signal that the poll has completed.
+                    pollCompleted.resolve();
+                    return new HttpResponse(204);
+                }
+            })
+            .on("DELETE", (r) => new HttpResponse(202));
+        const transport = new LongPollingTransport(client, null, NullLogger.instance, false, 100);
```
JamesNK: Do we know this 100 ms timeout is honored because the default 5000 ms timeout will cause the test to timeout?

anurse: Before I added this parameter, the test would flake out. Sometimes it would pass, sometimes it wouldn't, and it was based on the jasmine 5000 ms timeout. It's very consistent with this change so we know it should be honored.

PR #2172, opened 2018-04-30
fix #2171 by adding HttpTransportType.None

### LINKED ISSUE #2171: JavaScript Client: Selecting WebSockets transport explicitly doesn't behave correctly when server doesn't support WebSockets
If you set the `transport` property on the options object to `HttpTransportType.WebSockets`, we now send a negotiate response (due to the new "skip negotiate" flag). If the server doesn't support WebSockets, we should fail, without falling back. However, `HttpTransportType.WebSockets` has a value of `0`, which means when we hit this code path ....

https://github.com/aspnet/SignalR/blob/c53514fa19bbb776ee3855a6cfaa1c4706b7f446/clients/ts/signalr/src/HttpConnection.ts#L265

... `requestedTransport` is `0`, which means it always short-circuits and goes into the `if`.

The fix is simple: Make `HttpTransportType` in JS match the C# one, add a `None` options with a value of `0` and bump everything up one.

anurse:
#2171

### DIFF
--- a/clients/ts/signalr/spec/HttpConnection.spec.ts
+++ b/clients/ts/signalr/spec/HttpConnection.spec.ts
@@ -223,27 +223,35 @@ describe("HttpConnection", () => {
     });
 
     eachTransport((requestedTransport: HttpTransportType) => {
-        // OPTIONS is not sent when WebSockets transport is explicitly requested
-        if (requestedTransport === HttpTransportType.WebSockets) {
-            return;
-        }
-        it(`cannot be started if requested ${HttpTransportType[requestedTransport]} transport not available on server`, async (done) => {
+        it(`cannot be started if requested ${HttpTransportType[requestedTransport]} transport not available on server`, async () => {
+            const negotiateResponse = {
+                availableTransports: [
+                    { transport: "WebSockets", transferFormats: [ "Text", "Binary" ] },
+                    { transport: "ServerSentEvents", transferFormats: [ "Text" ] },
+                    { transport: "LongPolling", transferFormats: [ "Text", "Binary" ] },
+                ],
+                connectionId: "abc123",
+            };
+
+            // Remove the requested transport from the response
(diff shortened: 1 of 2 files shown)


clients/ts/signalr/spec/HttpConnection.spec.ts
```diff
@@ -223,27 +223,35 @@ describe("HttpConnection", () => {
     });
 
     eachTransport((requestedTransport: HttpTransportType) => {
-        // OPTIONS is not sent when WebSockets transport is explicitly requested
-        if (requestedTransport === HttpTransportType.WebSockets) {
-            return;
-        }
-        it(`cannot be started if requested ${HttpTransportType[requestedTransport]} transport not available on server`, async (done) => {
+        it(`cannot be started if requested ${HttpTransportType[requestedTransport]} transport not available on server`, async () => {
```
anurse: This test now exhibits the error if run without the fix below. Once the fix is applied, it passes. Previously it wasn't running for WebSockets.

clients/ts/signalr/src/ITransport.ts
```diff
@@ -2,9 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 export enum HttpTransportType {
-    WebSockets,
-    ServerSentEvents,
-    LongPolling,
+    None = 0,
+    WebSockets = 1,
+    ServerSentEvents = 2,
+    LongPolling = 3,
```
JamesNK: C# version:

    [Flags]
    public enum HttpTransportType
    {
        None = 0,
        WebSockets = 1,
        ServerSentEvents = 2,
        LongPolling = 4,
    }

Should LongPolling be 4 for consistency and to allow bitwise in the future?

davidfowl: Yass

anurse:
:up: :date: aligned with C# enum values.

PR #2173, opened 2018-04-30
fix #2140 by ensuring the access token flows to WebSocketTransport

### LINKED ISSUE #2140: NegotiateResponse with access token broken in .NET Client for WebSocket transport
We're using the wrong access token provider for websockets. The DefaultTransportFactory stashes the AccessTokenProvider on the options which we never update directly https://github.com/aspnet/SignalR/blob/43fc0854c38bbde1cb495d1b5e4f3e594da79e50/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs#L110.

anurse:
Right now I do it by rewriting the access token factory on the HttpConnectionOptions. I'm open to other options, we could flow it through the factory but it gets a bit funky because we either have to deal with the fact that the access token provider is async but `ITransportFactory` is not (currently).

Fixes #2140


davidfowl:
Doesn't this break Stop then Start?

src/Microsoft.AspNetCore.Http.Connections.Client/Internal/AccessTokenHttpMessageHandler.cs
```diff
@@ -11,16 +10,16 @@ namespace Microsoft.AspNetCore.Http.Connections.Client.Internal
...
+        public AccessTokenHttpMessageHandler(HttpMessageHandler inner, HttpConnectionOptions httpConnectionOptions) : base(inner)
         {
-            _httpConnection = httpConnection;
+            _httpConnectionOptions = httpConnectionOptions;
         }
 
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
-            var accessToken = await _httpConnection.GetAccessTokenAsync();
+            var accessToken = (_httpConnectionOptions.AccessTokenProvider != null) ? await _httpConnectionOptions.AccessTokenProvider() : null;
```
JamesNK: Either `_httpConnectionOptions.AccessTokenProvider?.Invoke` or assign to local variable. Avoid race issues with another thread setting AccessTokenProvider to null.

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/Startup.cs
```diff
@@ -69,6 +70,14 @@ public void Configure(IApplicationBuilder app)
                     await context.Response.WriteAsync(GenerateJwtToken());
                     return;
                 }
+                else if(context.Request.Path.StartsWithSegments("/redirect"))
```
JamesNK: > **else if(**

!!! TRIGGERED !!!

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Negotiate.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.AspNetCore.Http.Connections.Client;
 using Microsoft.AspNetCore.Http.Connections.Client.Internal;
 using Microsoft.AspNetCore.SignalR.Tests;
+using Microsoft.AspNetCore.Testing.xunit;
```
JamesNK: Needed?

test/Microsoft.AspNetCore.SignalR.Client.FunctionalTests/HubConnectionTests.cs
```diff
@@ -797,6 +797,34 @@ bool ExpectedErrors(WriteContext writeContext)
             }
         }
 
+        [Theory]
+        [MemberData(nameof(TransportTypes))]
+        public async Task ClientCanUseJwtBearerTokenForAuthenticationWhenRedirected(HttpTransportType transportType)
+        {
+            using (StartVerifableLog(out var loggerFactory, $"{nameof(ClientCanUseJwtBearerTokenForAuthentication)}_{transportType}"))
```
JamesNK: `nameof(ClientCanUseJwtBearerTokenForAuthenticationWhenRedirected)`

anurse:
> Doesn't this break Stop then Start?

Yeah, it probably would. I'll plumb a delegate that pulls the right field off the `HttpConnection` through instead then

anurse:
:up: :date: Now I flow a delegate to `HttpConnection.GetAccessTokenAsync()` down through to the transport so it can invoke and `await` it right before starting the WebSocket transport. The other transports use the HttpMessageHandler, so they're good.

PR #2178, opened 2018-05-01
New Benchmark Scenario 

mikaelm12:
Adding an "EchoIdle" scenario in preparation for long running tests

### DIFF
--- a/benchmarkapps/BenchmarkServer/signalr.json
+++ b/benchmarkapps/BenchmarkServer/signalr.json
@@ -20,5 +20,8 @@
   "SignalREchoAll": {
     "ClientProperties": { "Scenario": "echoAll" },
     "Warmup": 0
+  },
+  "SignalREchoIdle": {
+    "ClientProperties": { "Scenario": "echoIdle"}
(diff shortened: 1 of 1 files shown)


benchmarkapps/BenchmarkServer/signalr.json
```diff
@@ -20,5 +20,8 @@
   "SignalREchoAll": {
     "ClientProperties": { "Scenario": "echoAll" },
     "Warmup": 0
+  },
+  "SignalREchoIdle": {
+    "ClientProperties": { "Scenario": "echoIdle"}
```
BrennanConroy: nit: Space after `}`

BrennanConroy: Do we want to set any other defaults for the idle scenario?

mikaelm12: I figured we could just keep the defaults for now maybe with the exception of the duration.  The built in default send period is every 10 minutes for the idle scenario.

PR #2179, opened 2018-05-01
Make HttpTransportType a flags enum in TypeScript

### LINKED ISSUE #1961: Allow multiple HttpTransportTypes to be specified in TS client
https://github.com/aspnet/SignalR/pull/1925#discussion_r180876421

anurse:
This is based on #2172 because it builds on that code and I didn't want to add this straight to that branch yet. I'll rebase on `release/2.1`.

It was pointed out in #2172 that we should make the `HttpTransportType` enum use powers of two so it could be made a flags enum in future (#1961 was initially punted). That got me thinking that this is kinda trivial, so what if we just do it now.

Fixes #1961

### DIFF
--- a/clients/ts/signalr/spec/HttpConnection.spec.ts
+++ b/clients/ts/signalr/spec/HttpConnection.spec.ts
@@ -254,6 +254,66 @@ describe("HttpConnection", () => {
                 expect(e.message).toBe("Unable to initialize any of the available transports.");
             }
         });
+
+        for (const [val, name] of [[null, "null"], [undefined, "undefined"], [0, "0"]]) {
+            it(`can be started using ${HttpTransportType[requestedTransport]} transport when transport mask is ${name}`, async () => {
+                const negotiateResponse = {
+                    availableTransports: [
+                        { transport: "WebSockets", transferFormats: [ "Text", "Binary" ] },
+                        { transport: "ServerSentEvents", transferFormats: [ "Text" ] },
+                        { transport: "LongPolling", transferFormats: [ "Text", "Binary" ] },
+                    ],
+                    connectionId: "abc123",
+                };
+
+                const options: IHttpConnectionOptions = {
+                    ...commonOptions,
+                    httpClient: new TestHttpClient()
+                        .on("POST", (r) => negotiateResponse)
+                        .on("GET", (r) => new HttpResponse(204)),
+                    transport: val,
+                } as IHttpConnectionOptions;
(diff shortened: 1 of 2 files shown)


clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -282,7 +282,7 @@ export class HttpConnection implements IConnection {
     }
 
     private isITransport(transport: any): transport is ITransport {
-        return typeof (transport) === "object" && "connect" in transport;
+        return transport && typeof (transport) === "object" && "connect" in transport;
```
anurse: This was needed because it fails if `transport` is `null`. That was true before my change too. `typeof null` is `object` (see below) but you can't use `in` with `null`, nor can you check for a method. There's a similar place we do this with `ILogger` that I'll fix as well.

image
image

image

clients/ts/signalr/spec/HttpConnection.spec.ts
```diff
@@ -223,27 +223,95 @@ describe("HttpConnection", () => {
...
                 await connection.start(TransferFormat.Text);
-                fail();
-                done();
+                fail("Expected connection.start to throw!");
+            } catch (e) {
+                expect(e.message).toBe("Unable to initialize any of the available transports.");
+            }
+        });
+
+        for (const [val, name] of [[null, "null"], [undefined, "undefined"], [0, "0"]]) {
```
JamesNK: name -> transportName
val -> transportValue

anurse:
> Worth having an end-to-end functional test

Not sure what you mean. A NodeJS-based functional test? I tried to limit this to just porting the existing tests, we can iterate on more testing in 2.2.0-mq. We still have the browser-based functional tests that run in Jasmine in the browser.

anurse:
I think the unit tests give us pretty broad coverage here. It's harder (but certainly possible) to build a functional test because we have to produce endpoints that properly limit the allowed transports.

JamesNK:
FYI this needs to be merged tomorrow (Friday). Please take some time to give feedback @davidfowl @BrennanConroy @mikaelm12

clients/ts/signalr/src/HttpConnection.ts
davidfowl: Put this comment in the code here https://github.com/aspnet/SignalR/pull/2179/files#diff-11502bbf017bbae00a723c7de894f310R350

Issue #2165, opened 2018-04-30
flaky test: StreamDoesNotStartIfTokenAlreadyCanceled_json_ServerSentEvents_default


davidfowl:
Seen on this PR https://github.com/aspnet/SignalR/pull/2162

https://ci3.dot.net/job/aspnet_SignalR/job/dev/job/windows-Configuration_Release_prtest/621/

```
20:53:23    System.Exception : 2 error(s) logged.
20:53:23   Microsoft.AspNetCore.Server.Kestrel - ApplicationError - Connection id "0HLDE9NGO7HGV", Request id "0HLDE9NGO7HGV:00000003": An unhandled exception was thrown by the application.
20:53:23   ===================
20:53:23   System.InvalidOperationException: Writing is not allowed after writer was completed.
20:53:23      at System.IO.Pipelines.Pipe.GetMemory(Int32 sizeHint)
20:53:23      at System.IO.Pipelines.Pipe.DefaultPipeWriter.GetSpan(Int32 sizeHint)
20:53:23      at System.IO.Pipelines.PipeWriter.WriteAsync(ReadOnlyMemory`1 source, CancellationToken cancellationToken)
20:53:23      at System.IO.Pipelines.PipeWriterStream.WriteCoreAsync(ReadOnlyMemory`1 source, CancellationToken cancellationToken)
20:53:23      at System.IO.Pipelines.PipeWriterStream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
20:53:23      at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.MessageBody.<CopyToAsync>d__22.MoveNext()
20:53:23   --- End of stack trace from previous location where exception was thrown ---
20:53:23      at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
20:53:23      at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
20:53:23      at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.<CopyToAsyncInternal>d__23.MoveNext()
20:53:23   --- End of stack trace from previous location where exception was thrown ---
20:53:23      at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #2180, opened 2018-05-01
Gracefully handle disposing while writing

JamesNK:
https://github.com/aspnet/SignalR/issues/2165


src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
```diff
@@ -180,7 +180,7 @@ public void TickHeartbeat()
         {
             var disposeTask = Task.CompletedTask;
 
-            await StateLock.WaitAsync();
+            await Task.WhenAll(StateLock.WaitAsync(), WriteLock.WaitAsync());
```
BrennanConroy: There was a reason we didn't take both locks here
I think this might deadlock if there is backpressure at https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs#L482

davidfowl: Don't we have a test for this?

JamesNK: So the issue is that because of back pressure the writer will wait until it can write again? Are you worried that we don't want the dispose to hang around, waiting on the write lock to become available?

Would a solution be to have a cancellation token that the connection is closing. That will break users of the connection out of waiting on locks and allow a more graceful shutdown.

davidfowl: We should call CancelPendingFlush to yield the backpressure (not sure why we aren’t doing this already).

Does this mean we can go back to a single lock? I need to think about this a bit

JamesNK: I was only talking about the implications of the dispose scenario if both locks were taken in it. CancelPendingFlush might help there.

JamesNK:
🆙 📅

JamesNK:

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1797,6 +1797,137 @@ bool ExpectedErrors(WriteContext writeContext)
...
+                var dispatcher = new HttpConnectionDispatcher(manager, loggerFactory);
+
+                var services = new ServiceCollection();
+                services.AddSingleton<TestConnectionHandler>();
+                var builder = new ConnectionBuilder(services.BuildServiceProvider());
+                builder.UseConnectionHandler<TestConnectionHandler>();
+                var app = builder.Build();
+                var options = new HttpConnectionDispatcherOptions();
+
+                TaskCompletionSource<bool> tcs = new TaskCompletionSource<bool>();
```
anurse: nit: `var`

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1797,6 +1797,137 @@ bool ExpectedErrors(WriteContext writeContext)
...
+                    context.Request.Query = qs;
+                    var buffer = Encoding.UTF8.GetBytes("Hello, world");
+                    requestBody.Write(buffer, 0, buffer.Length);
+                    requestBody.Seek(0, SeekOrigin.Begin);
+
+                    // Write. This will wait on the TCS inside ApplicationStream.CopyToAsync
+                    var sendTask1 = dispatcher.ExecuteAsync(context, options, app);
+
+                    // Extra check to make sure ExecuteAsync is waiting on the TCS
+                    if (sendTask1.Status != TaskStatus.WaitingForActivation)
```
anurse: FYI this is what `SyncPoint` in the `Tests.Utils` assembly is for. It's a pair of TCSes that let you do basically the same thing.

You would give the SyncPoint to the `ControllableMemoryStream` and have it call and `await` `.WaitToContinue()` in `CopyToAsync`. Then outside the code, after calling `ExecuteAsync` you await `.WaitForSyncPoint()`.

`WaitForSyncPoint` is released when `WaitToContinue` is called. Then `Continue` releases `WaitToContinue`.

JamesNK: I love it

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.Log.cs
```diff
@@ -46,6 +46,9 @@ private static class Log
             private static readonly Action<ILogger, Exception> _terminatingConnection =
                 LoggerMessage.Define(LogLevel.Trace, new EventId(12, "TerminatingConection"), "Terminating Long Polling connection due to a DELETE request.");
 
+            private static readonly Action<ILogger, string, Exception> _connectionDisposedWhileWriteInProgress =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(13, "ConnectionDisposedWhileWriteInProgress"), "Connection Id {TransportConnectionId} was disposed while a write was in progress.");
```
davidfowl: "Connection {TransportConnectionId}" remove the "Id"

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
```diff
@@ -267,6 +267,9 @@ public void TickHeartbeat()
                 {
                     Log.ShuttingDownTransportAndApplication(_logger, TransportType);
 
+                    // Cancel any pending flushes from back pressure
+                    Application?.Output.CancelPendingFlush();
```
davidfowl: Do we need this if we're going to be calling complete?

davidfowl: Ah, I see where we use this. It does affect other code though. I guess the only place that we do this is websockets. Have you looked there to make sure we handle cancellation ok?

JamesNK: I don't know where there is. Could you point me in the right direction.

I can say tests all pass 🙃

davidfowl: https://github.com/aspnet/SignalR/blob/028d0b233e7ee515e66c6dc04ee5a3a42d2956a1/src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/WebSocketsTransport.cs#L179

src/Common/PipeWriterStream.cs
```diff
@@ -83,7 +83,16 @@ private ValueTask WriteCoreAsync(ReadOnlyMemory<byte> source, CancellationToken
 
             return default;
 
-            async ValueTask WriteSlowAsync(ValueTask<FlushResult> flushTask) => await flushTask;
+            async ValueTask WriteSlowAsync(ValueTask<FlushResult> flushTask)
+            {
+                var flushResult = await flushTask;
+
+                // Cancellation can be triggered by PipeWriter.CancelPendingFlush
+                if (flushResult.IsCanceled)
```
davidfowl: Are we handling this on the client properly?

https://github.com/aspnet/SignalR/blob/f632330d7f5232053ef183c8c78b7c367087ff37/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/PipeReaderFactory.cs#L33

https://github.com/aspnet/SignalR/blob/9d8b83e6b34f294cc5aba4f94fdd74cb8e5bc0d7/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs#L187

test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs
```diff
@@ -1797,6 +1797,134 @@ bool ExpectedErrors(WriteContext writeContext)
...
+                    await connection.ApplicationStream.WriteAsync(buffer, 0, buffer.Length).OrTimeout();
+
+                    // Write. This will take the WriteLock and block because of back pressure
+                    var sendTask = dispatcher.ExecuteAsync(context, options, app);
+
+                    // Start disposing. This will take the StateLock and attempt to take the WriteLock
+                    // Dispose will cancel pending flush and should unblock WriteLock
+                    await connection.DisposeAsync().OrTimeout();
+
+                    // Sends were unblock
```
BrennanConroy: unblocked

src/Common/PipeWriterStream.cs
JamesNK: https://github.com/aspnet/SignalR/blob/f632330d7f5232053ef183c8c78b7c367087ff37/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/PipeReaderFactory.cs#L33

Already handles OperationCanceledException.

https://github.com/aspnet/SignalR/blob/9d8b83e6b34f294cc5aba4f94fdd74cb8e5bc0d7/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs#L187

Hmmm... the FlushAsync after CopyToAsync seems odd. CopyToAsync already flushes, no?

This is nested inside OperationCanceledException catch so is functionally similar to breaking out of the loop

PR #2190, opened 2018-05-02
fix #2187 by passing min size through to buffer writer when encoding

### LINKED ISSUE #2187: Exception when sending data from server to client: Failed writing message. System.ArgumentException: The output byte buffer is too small to contain the encoded data, encoding 'Unicode (UTF-8)' fallback 'System.Text.EncoderReplacementFallback'. Parameter name: bytes
Hey.
my system:
Server: on .net core 2.1 preview 2 running on windows using kestel (also tried on azure using app service - same result)
Client - Java script client version: "@aspnet/signalr": "^1.0.0-preview2-final"
Protocol: Json

Scenario: when sending data from the server to the client (using the command: await m_Client.SendAsync("setTripGeneralData",generalData)) in several cases (of data) the following exception is thrown:

```
2018-05-02 16:01:43.815 +03:00|0HLDG5GH9MK8I:00000010|Debug|9|Microsoft.AspNetCore.SignalR.HubConnectionContext.||Failed writing message.
System.ArgumentException: The output byte buffer is too small to contain the encoded data, encoding 'Unicode (UTF-8)' fallback 'System.Text.EncoderReplacementFallback'.
Parameter name: bytes
   at System.Text.Encoding.ThrowBytesOverflow()
   at System.Text.Encoding.ThrowBytesOverflow(EncoderNLS encoder, Boolean nothingEncoded)
   at System.Text.UTF8Encoding.GetBytes(Char* chars, Int32 charCount, Byte* bytes, Int32 byteCount, EncoderNLS baseEncoder)
   at System.Text.EncoderNLS.Convert(Char* chars, Int32 charCount, Byte* bytes, Int32 byteCount, Boolean flush, Int32& charsUsed, Int32& bytesUsed, Boolean& completed)
   at System.Text.Encoder.Convert(ReadOnlySpan`1 chars, Span`1 bytes, Boolean flush, Int32& charsUsed, Int32& bytesUsed, Boolean& completed)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.Utf8BufferTextWriter.WriteInternal(ReadOnlySpan`1 buffer)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.Utf8BufferTextWriter.Write(String value)
   at Newtonsoft.Json.Utilities.JavaScriptUtils.WriteEscapedJavaScriptString(TextWriter writer, String s, Char delimiter, Boolean appendDelimiters, Boolean[] charEscapeFlags, StringEscapeHandling stringEscapeHandling, IArrayPool`1 bufferPool, Char[]& writeBuffer)
   at Newtonsoft.Json.JsonTextWriter.WriteEscapedString(String value, Boolean quote)
   at Newtonsoft.Json.JsonTextWriter.WriteValue(String value)
   at Newtonsoft.Json.JsonWriter.WriteValue(JsonWriter writer, PrimitiveTypeCode typeCode, Object value)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializePrimitive(JsonWriter writer, Object value, JsonPrimitiveContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeList(JsonWriter writer, IEnumerable values, JsonArrayContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeList(JsonWriter writer, IEnumerable values, JsonArrayContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.Serialize(JsonWriter jsonWriter, Object value)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.JsonHubProtocol.WriteArguments(Object[] arguments, JsonTextWriter writer)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.JsonHubProtocol.WriteInvocationMessage(InvocationMessage message, JsonTextWriter writer)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.JsonHubProtocol.WriteMessageCore(HubMessage message, IBufferWriter`1 stream)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.JsonHubProtocol.WriteMessage(HubMessage message, IBufferWriter`1 output)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.HubProtocolExtensions.WriteToArray(IHubProtocol hubProtocol, HubMessage message)
   at Microsoft.AspNetCore.SignalR.Internal.Protocol.HubMessage.WriteMessage(IHubProtocol protocol)
   at Microsoft.AspNetCore.SignalR.HubConnectionContext.WriteCore(HubMessage message)
```

this didn't happened with the alpha release of signalR core (and .net core 2) with the same data cases.
although the exception implies that the problem is with the size of the data, the data size is not big and in other cases i am sending bigger data amounts with success. i tried investigating the data to pin the difference but i couldn't found nothing sensuous. after saying that, i did tried sending parts of the problematic data and succeeded. only when sending all the problematic data together (its just a list of objects) i am receiving the exception, so maybe it does related to size - i don't know.
i couldn't find a way to enlarge the buffer so i cant test it.
i know that the json.net version was updated in this release of .net core, maybe its related?

any clew anybody?

anurse:
This fixes an error that occurs when a multi-byte character is about to be written by Utf8BufferTextWriter, but there isn't enough space for it in the buffer. We use [`Encoder.Convert`](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoder.convert?view=netstandard-2.0#System_Text_Encoder_Convert_System_Char__System_Int32_System_Byte__System_Int32_System_Boolean_System_Int32__System_Int32__System_Boolean__) which writes as many characters as it can fit into the provided buffer. However, if there isn't even enough space for a single character, it throws. 

Consider this example of the current behavior, where the buffers are in 4-byte segments (of course the real buffers are like 2048 bytes, but for illustration...):

1. The buffer has 4 bytes of space: `[ ] [ ] [ ] [ ]`
2. We write `aい`, which takes up 4 bytes: `[0x61] [0xE3] [0x81] [0x84]`
3. We call `GetMemory(0)` and since the buffer is full, we get a new 4 byte buffer: `[ ] [ ] [ ] [ ]`
4. We try to write `はに`. Only the first character fits, so we fill 3 bytes: `[0xE3] [0x81] [0xAF] [ ]`
5. We call `GetMemory(0)` but since the buffer is not full, we get the same buffer: `[0xE3] [0x81] [0xAF] [ ]`
6. We try to write `に` (since it wasn't written before). The buffer is only 1 byte in size, so the encoder throws:

```
System.ArgumentException: The output byte buffer is too small to contain the encoded data, encoding 'Unicode (UTF-8)' fallback 'System.Text.EncoderReplacementFallback'.
Parameter name: bytes
```

The new behavior is similar, but it specifies a `sizeHint` value of `4` (the maximum number of bytes needed to encoded a UTF-8 character to `IBufferWriter.GetMemory`, which means it will force a new segment to be allocated if there isn't enough space in the current segment. So the updated logic for the same input as above is:

1. The buffer has 4 bytes of space: `[ ] [ ] [ ] [ ]`
2. We write `aい`, which takes up 4 bytes: `[0x61] [0xE3] [0x81] [0x84]`
3. We call `GetMemory(4)`; since the buffer is full, we get a new 4 byte buffer: `[ ] [ ] [ ] [ ]`
4. We try to write `はに`. Only the first character fits, so we fill 3 bytes: `[0xE3] [0x81] [0xAF] [ ]`
5. We call `GetMemory(4)`; since the buffer doesn't have 4 bytes left, we get a new buffer: `[ ] [ ] [ ] [ ]`
6. We try to write `に` (since it wasn't written before). And it fits: `[0xE3] [0x81] [0xAB] [ ]`

I'll run the hub protocol benchmarks and post the results here.

Fixes #2187

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/HubProtocolBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/HubProtocolBenchmark.cs
@@ -18,7 +18,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
         [Params(Message.NoArguments, Message.FewArguments, Message.ManyArguments, Message.LargeArguments)]
         public Message Input { get; set; }
 
-        [Params(/*Protocol.MsgPack,*/ Protocol.Json)]
+        [Params(Protocol.MsgPack, Protocol.Json)]
         public Protocol HubProtocol { get; set; }
 
         [GlobalSetup]
--- a/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/HubProtocolBenchmark.cs
+++ b/benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/HubProtocolBenchmark.cs
@@ -18,7 +18,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
         [Params(Message.NoArguments, Message.FewArguments, Message.ManyArguments, Message.LargeArguments)]
         public Message Input { get; set; }
 
-        [Params(Protocol.MsgPack, Protocol.Json)]
+        [Params(/*Protocol.MsgPack,*/ Protocol.Json)]
         public Protocol HubProtocol { get; set; }
 
         [GlobalSetup]
--- a/src/Common/MemoryBufferWriter.cs
+++ b/src/Common/MemoryBufferWriter.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -6,7 +6,6 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -24,7 +23,7 @@ namespace Microsoft.AspNetCore.Internal
         private readonly int _minimumSegmentSize;
         private int _bytesWritten;
 
-        private List<BufferSegment> _completedSegments;
+        private List<CompletedBuffer> _completedSegments;
         private byte[] _currentSegment;
         private int _position;
 
@@ -126,7 +125,7 @@ namespace Microsoft.AspNetCore.Internal
                 var count = _completedSegments.Count;
                 for (var i = 0; i < count; i++)
                 {
-                    destination.Write(_completedSegments[i].Memory.Span);
+                    destination.Write(_completedSegments[i].Span);
                 }
             }
 
@@ -168,10 +167,13 @@ namespace Microsoft.AspNetCore.Internal
                 // We're adding a segment to the list
                 if (_completedSegments == null)
                 {
-                    _completedSegments = new List<BufferSegment>();
+                    _completedSegments = new List<CompletedBuffer>();
                 }
 
-                _completedSegments.Add(new BufferSegment(_currentSegment, _position));
+                // Position might be less than the segment length if there wasn't enough space to satisfy the sizeHint when
+                // GetMemory was called. In that case we'll take the current segment and call it "completed", but need to
+                // ignore any empty space in it.
+                _completedSegments.Add(new CompletedBuffer(_currentSegment, _position));
             }
 
             // Get a new buffer using the minimum segment size, unless the size hint is larger than a single segment.
@@ -213,8 +215,8 @@ namespace Microsoft.AspNetCore.Internal
                 for (var i = 0; i < count; i++)
                 {
                     var segment = _completedSegments[i];
-                    segment.Memory.Span.CopyTo(result.AsSpan(totalWritten));
-                    totalWritten += segment.Length;
+                    segment.Span.CopyTo(result.AsSpan(totalWritten));
+                    totalWritten += segment.Span.Length;
                 }
             }
 
@@ -242,8 +244,8 @@ namespace Microsoft.AspNetCore.Internal
                 for (var i = 0; i < count; i++)
                 {
                     var segment = _completedSegments[i];
-                    segment.Memory.Span.CopyTo(span.Slice(totalWritten));
-                    totalWritten += segment.Length;
+                    segment.Span.CopyTo(span.Slice(totalWritten));
+                    totalWritten += segment.Span.Length;
                 }
             }
 
@@ -317,17 +319,17 @@ namespace Microsoft.AspNetCore.Internal
         /// <summary>
         /// Holds a byte[] from the pool and a size value. Basically a Memory but guaranteed to be backed by an ArrayPool byte[], so that we know we can return it.
         /// </summary>
-        private struct BufferSegment
+        private readonly struct CompletedBuffer
         {
             public byte[] Buffer { get; }
             public int Length { get; }
 
-            public ReadOnlyMemory<byte> Memory => Buffer.AsMemory(0, Length);
+            public ReadOnlySpan<byte> Span => Buffer.AsSpan(0, Length);
 
-            public BufferSegment(byte[] buffer, int size)
+            public CompletedBuffer(byte[] buffer, int length)
             {
                 Buffer = buffer;
-                Length = size;
+                Length = length;
             }
 
             public void Return()
--- a/src/Common/MemoryBufferWriter.cs
+++ b/src/Common/MemoryBufferWriter.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.Internal
         private readonly int _minimumSegmentSize;
         private int _bytesWritten;
 
-        private List<byte[]> _fullSegments;
+        private List<BufferSegment> _completedSegments;
         private byte[] _currentSegment;
         private int _position;
 
@@ -78,14 +78,14 @@ namespace Microsoft.AspNetCore.Internal
 
         public void Reset()
         {
-            if (_fullSegments != null)
+            if (_completedSegments != null)
             {
-                for (var i = 0; i < _fullSegments.Count; i++)
+                for (var i = 0; i < _completedSegments.Count; i++)
                 {
-                    ArrayPool<byte>.Shared.Return(_fullSegments[i]);
+                    _completedSegments[i].Return();
                 }
 
-                _fullSegments.Clear();
+                _completedSegments.Clear();
             }
 
             if (_currentSegment != null)
@@ -120,13 +120,13 @@ namespace Microsoft.AspNetCore.Internal
 
         public void CopyTo(IBufferWriter<byte> destination)
         {
-            if (_fullSegments != null)
+            if (_completedSegments != null)
             {
-                // Copy full segments
-                var count = _fullSegments.Count;
+                // Copy completed segments
+                var count = _completedSegments.Count;
                 for (var i = 0; i < count; i++)
                 {
-                    destination.Write(_fullSegments[i]);
+                    destination.Write(_completedSegments[i].Memory.Span);
                 }
             }
 
@@ -135,9 +135,9 @@ namespace Microsoft.AspNetCore.Internal
 
         public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
         {
-            if (_fullSegments == null)
+            if (_completedSegments == null)
             {
-                // There is only one segment so write without async
+                // There is only one segment so write without awaiting.
                 return destination.WriteAsync(_currentSegment, 0, _position);
             }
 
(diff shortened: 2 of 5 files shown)


src/Common/MemoryBufferWriter.cs
```diff
@@ -307,5 +313,27 @@ protected override void Dispose(bool disposing)
                 Reset();
             }
         }
+
+        /// <summary>
+        /// Holds a byte[] from the pool and a size value. Basically a Memory but guaranteed to be backed by an ArrayPool byte[], so that we know we can return it.
+        /// </summary>
+        private struct BufferSegment
```
davidfowl: Why do we need this?

anurse:
No obvious change in perf:

```
Before:

  ReadSingleMessage |   FewArguments |        Json |  1,588.4 ns |   6.770 ns |   6.001 ns |   629,557.9 | 0.0095 |     976 B |
 WriteSingleMessage |   FewArguments |        Json |  1,482.0 ns |   2.552 ns |   2.131 ns |   674,769.8 | 0.0076 |     824 B |
  ReadSingleMessage | LargeArguments |        Json | 60,528.2 ns | 197.791 ns | 165.164 ns |    16,521.2 | 0.7324 |   58872 B |
 WriteSingleMessage | LargeArguments |        Json | 36,588.6 ns |  80.589 ns |  71.440 ns |    27,330.9 | 0.3052 |   24792 B |
  ReadSingleMessage |  ManyArguments |        Json |  2,847.1 ns |   9.412 ns |   8.804 ns |   351,231.4 | 0.0153 |    1536 B |
 WriteSingleMessage |  ManyArguments |        Json |  2,781.2 ns |   5.741 ns |   5.090 ns |   359,558.5 | 0.0229 |    2080 B |
  ReadSingleMessage |    NoArguments |        Json |    859.3 ns |   3.244 ns |   2.876 ns | 1,163,745.3 | 0.0057 |     600 B |
 WriteSingleMessage |    NoArguments |        Json |    716.0 ns |   1.591 ns |   1.411 ns | 1,396,726.0 | 0.0038 |     392 B |


After:

  ReadSingleMessage |   FewArguments |        Json |  1,597.0 ns |  39.848 ns |  35.324 ns |   626,192.5 | 0.0095 |     976 B |
 WriteSingleMessage |   FewArguments |        Json |  1,535.0 ns |   5.302 ns |   4.428 ns |   651,461.4 | 0.0095 |     824 B |
  ReadSingleMessage | LargeArguments |        Json | 60,830.7 ns | 271.260 ns | 253.737 ns |    16,439.1 | 0.7324 |   58872 B |
 WriteSingleMessage | LargeArguments |        Json | 36,514.7 ns |  75.660 ns |  67.070 ns |    27,386.2 | 0.3052 |   24792 B |
  ReadSingleMessage |  ManyArguments |        Json |  2,808.3 ns |   9.053 ns |   8.025 ns |   356,082.0 | 0.0153 |    1536 B |
 WriteSingleMessage |  ManyArguments |        Json |  2,780.2 ns |   7.092 ns |   6.634 ns |   359,692.2 | 0.0229 |    2080 B |
  ReadSingleMessage |    NoArguments |        Json |    828.6 ns |   1.787 ns |   1.395 ns | 1,206,807.9 | 0.0067 |     600 B |
 WriteSingleMessage |    NoArguments |        Json |    729.8 ns |   2.421 ns |   2.264 ns | 1,370,279.2 | 0.0038 |     392 B |
```

src/Common/MemoryBufferWriter.cs
anurse: Previously, the list was of "Full" segments. However, now a segment can be retired and put into the chain when only partially full (because there wasn't enough space in the buffer). We can't just track the array, we need to track how much of the array is used. I considered using `ReadOnlyMemory<byte>` in the list but that meant returning it to the pool required calling `TryGetArray`.

src/Common/Utf8BufferTextWriter.cs
```diff
@@ -139,15 +140,16 @@ private Span<byte> GetBuffer()
 
         private void EnsureBuffer()
         {
-            if (_memoryUsed == _memory.Length)
+            var remaining = _memory.Length - _memoryUsed;
+            if (remaining < MaximumBytesPerUtf8Char)
```
JamesNK: Add a comment why this is done

src/Common/MemoryBufferWriter.cs
```diff
@@ -146,43 +146,49 @@ public override Task CopyToAsync(Stream destination, int bufferSize, Cancellatio
...
                 // We're adding a segment to the list
-                if (_fullSegments == null)
+                if (_completedSegments == null)
                 {
-                    _fullSegments = new List<byte[]>();
+                    _completedSegments = new List<BufferSegment>();
                 }
 
-                _fullSegments.Add(_currentSegment);
+                _completedSegments.Add(new BufferSegment(_currentSegment, _position));
```
JamesNK: Add a comment explaining why/when the position might be less than the complete segment length

src/Common/MemoryBufferWriter.cs
```diff
@@ -307,5 +313,27 @@ protected override void Dispose(bool disposing)
...
+        {
+            public byte[] Buffer { get; }
+            public int Length { get; }
+
+            public ReadOnlyMemory<byte> Memory => Buffer.AsMemory(0, Length);
+
+            public BufferSegment(byte[] buffer, int size)
+            {
+                Buffer = buffer;
+                Length = size;
```
JamesNK: Kind of weird that names are different

JamesNK: What about renaming BufferSegment to CompletedSegment? Length can be renamed to Used.

JamesNK:
Doesn't perf look the same as before? But yeah, no need to go byte[] -> Memory -> Span. Skip the memory step.

src/Common/MemoryBufferWriter.cs
davidfowl: It should just be Length, the names aren't too strange (look at memory and span).

src/Common/MemoryBufferWriter.cs
```diff
@@ -307,5 +313,27 @@ protected override void Dispose(bool disposing)
                 Reset();
             }
         }
+
+        /// <summary>
+        /// Holds a byte[] from the pool and a size value. Basically a Memory but guaranteed to be backed by an ArrayPool byte[], so that we know we can return it.
+        /// </summary>
+        private struct BufferSegment
+        {
+            public byte[] Buffer { get; }
```
davidfowl: Why are these public? Seems like you'd only ever need to access the Span or Memory

src/Common/MemoryBufferWriter.cs
davidfowl: > I considered using ReadOnlyMemory<byte> in the list but that meant returning it to the pool required calling TryGetArray.

Yea, I was wondering if this is better. Returning happens at the end and it might be more efficient than another struct.

src/Common/MemoryBufferWriter.cs
anurse: Derp, I just renamed the property and didn't catch the parameter.

src/Common/MemoryBufferWriter.cs
anurse: There is one place where it saves LOC by making these public. In `CopyToSlowAsync(Stream)` we don't have to do the crazy `#ifdef`s to use the Stream Span APIs. We know this is always an array though so I guess I can try unwrapping it.

anurse: Except that requires exposing a `ReadOnlyMemory<T>` and exposing both a `ROM` and a `ROS` just to avoid making fields on a private nested struct public seems unncessary.

benchmarks/Microsoft.AspNetCore.SignalR.Microbenchmarks/HubProtocolBenchmark.cs
```diff
@@ -18,7 +18,7 @@ public class HubProtocolBenchmark
         [Params(Message.NoArguments, Message.FewArguments, Message.ManyArguments, Message.LargeArguments)]
         public Message Input { get; set; }
 
-        [Params(Protocol.MsgPack, Protocol.Json)]
+        [Params(/*Protocol.MsgPack,*/ Protocol.Json)]
```
davidfowl: UNDO

anurse: derp

PR #2193, opened 2018-05-02
switch tests to 'jest'

anurse:
* Jasmine is still needed for the Functional Tests, Jest doesn't work in the browser. Though an interesting side note is that Jest actually uses Jasmine as the low-level runner, Jest is just a higher-level wrapper that provides features like running each file in a separate node process (for parallel execution) and more.
* There are some small product code changes that have no impact on any of our exported types; I added comments to those inline
* I reorganized and cleaned-up our tsconfig files. We were specifying "ES2016" as the target in the tsconfig.json but then every invocation of 'tsc' overrides that down to "ES5", so I just moved that setting to the tsconfig.json.
* I changed how the FunctionalTests "import" the signalr/signalr-protocol-msgpack library. Previously they used a bunch of hacks to yank the necessary files from outside the directory. That resulted in TypeScript actually re-compiling the code, which wasn't really the intent. Instead, I ran `npm install ../signalr` and `npm install ../signalr-protocol-msgpack` which create hard-linked directories. So `clients/ts/FunctionalTests/node_modules/@aspnet/signalr/` is a hard-link to `clients/ts/signalr`, etc. NPM re-creates the hard-link when you `npm install` so it should "Just Work" (CI is there to test this out for us :)). This works **without** requiring admin access on Windows.

The improvement to the test time is intense. I timed the whole process, including TypeScript compilation of the tests (which Jest, via `ts-jest` does automatically for us), to about 14 seconds. With this change, the whole run of unit tests takes...

image

Since this is test-only (see comments for some extremely low risk changes to product code that we could probably hack out if we needed to), we talked about putting this in RTM, so I'm putting the PR up targeting `release/2.1`. We can always change to `dev`.


clients/ts/FunctionalTests/package.json
```diff
@@ -4,12 +4,17 @@
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
-  "dependencies": {},
+  "dependencies": {
+    "@aspnet/signalr": "file:../signalr",
+    "@aspnet/signalr-protocol-msgpack": "file:../signalr-protocol-msgpack"
```
anurse: This is the magic that makes NPM create hard-links

clients/ts/FunctionalTests/rollup.config.js
```diff
@@ -8,7 +8,7 @@ import commonjs from 'rollup-plugin-commonjs'
 import resolve from 'rollup-plugin-node-resolve'
 
 export default {
-    input: path.join(__dirname, "obj", "js", "FunctionalTests", "ts", "index.js"),
+    input: path.join(__dirname, "obj", "js", "index.js"),
```
anurse: Now that we aren't compiling any of `@aspnet/signalr` during the build of the functional tests, TypeScript doesn't generate the extra folders.

clients/ts/FunctionalTests/ts/ConnectionTests.ts
```diff
@@ -6,7 +6,7 @@ import { eachTransport, ECHOENDPOINT_URL } from "./Common";
 import { TestLogger } from "./TestLogger";
 
 // We want to continue testing HttpConnection, but we don't export it anymore. So just pull it in directly from the source file.
-import { HttpConnection } from "../../signalr/src/HttpConnection";
+import { HttpConnection } from "@aspnet/signalr/dist/esm/HttpConnection";
```
anurse: This syntax allows us to import a specific path within a module, which means TypeScript won't go over to the source and try to compile it again.

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -116,9 +116,9 @@ describe("hubConnection", () => {
                     done();
                 });
 
-                const received = [];
+                const received: string[] = [];
                 hubConnection.start().then(() => {
-                    hubConnection.stream("Stream").subscribe({
+                    hubConnection.stream<string>("Stream").subscribe({
```
anurse: Our other tsconfig files set `noImplicitAny` to `true` which disallows any implicit `any` types (i.e. things the inference engine couldn't infer). This one didn't, but I brought it in line and had to add some explicit type annotations

clients/ts/FunctionalTests/ts/index.ts
```diff
@@ -4,6 +4,7 @@
 console.log("SignalR Functional Tests Loaded");
 
 import "es6-promise/dist/es6-promise.auto.js";
+import "./ConnectionTests";
```
anurse: So... uh.. when I brought the tests back I forgot to bring back the import. These haven't been running since the IConnection refactorings in TypeScript. They all still pass though, so no worries.

clients/ts/FunctionalTests/tsconfig.json
```diff
@@ -1,27 +1,12 @@
 {
-  "compileOnSave": false,
+  "extends": "../tsconfig.base.json",
```
anurse: We now share all compilerOptions from the common `tsconfig.base.json` in `clients\ts`, and only override the things we have to in the individual folders.

clients/ts/package.json
```diff
@@ -5,21 +5,45 @@
...
+    "globals": {
+      "ts-jest": {
+        "tsConfigFile": "./tsconfig.jest.json"
+      }
+    },
+    "transform": {
+      "^.+\\.tsx?$": "ts-jest"
+    },
+    "testEnvironment": "node",
+    "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$",
```
anurse: This is The Official Regex for `ts-jest`, and it's just an update to The Official Regex for `jest` to support `ts` and `tsx` files. The patterns are:

* Any file in a `__tests__` subfolder anywhere in the project directory.
* Any `.js`, `.jsx`, `.ts`, or `.tsx` file with the suffix `.test` or `.spec` anywhere in the project directory.

clients/ts/package.json
```diff
@@ -5,21 +5,45 @@
...
+    "moduleNameMapper": {
+      "^@aspnet/signalr$": "<rootDir>/signalr/src/index.ts"
```
anurse: Required so that the `signalr-protocol-msgpack` unit tests can load `@aspnet/signalr`

clients/ts/package.json
```diff
@@ -5,21 +5,45 @@
...
+    },
+    "moduleFileExtensions": [
+      "ts",
+      "tsx",
+      "js",
+      "jsx",
+      "json",
+      "node"
```
anurse: Only needs to be specified to add `.ts` and `.tsx` to the list. The rest are the default value which needs to be "re-specified" when you set this value.

clients/ts/signalr-protocol-msgpack/package.json
```diff
@@ -14,12 +14,11 @@
     "clean": "node ../node_modules/rimraf/bin.js ./dist",
     "build": "npm run clean && npm run build:lint && npm run build:esm && npm run build:cjs && npm run build:browser && npm run build:uglify",
     "build:lint": "node ../node_modules/tslint/bin/tslint -c ../tslint.json -p ./tsconfig.json",
-    "build:esm": "node ../node_modules/typescript/bin/tsc --project ./tsconfig.json --module es2015 --outDir ./dist/esm --target ES5 -d",
-    "build:cjs": "node ../node_modules/typescript/bin/tsc --project ./tsconfig.json --module commonjs --outDir ./dist/cjs --target ES5",
+    "build:esm": "node ../node_modules/typescript/bin/tsc --project ./tsconfig.json --module es2015 --outDir ./dist/esm -d",
+    "build:cjs": "node ../node_modules/typescript/bin/tsc --project ./tsconfig.json --module commonjs --outDir ./dist/cjs",
```
anurse: `--target` is now in the `tsconfig.base.json`

clients/ts/signalr-protocol-msgpack/package.json
```diff
@@ -14,12 +14,11 @@
...
     "build:browser": "node ../node_modules/rollup/bin/rollup -c",
     "build:uglify": "node ../node_modules/uglify-js/bin/uglifyjs --source-map \"url='signalr-protocol-msgpack.min.js.map',content='./dist/browser/signalr-protocol-msgpack.js.map'\" --comments -o ./dist/browser/signalr-protocol-msgpack.min.js ./dist/browser/signalr-protocol-msgpack.js",
-    "pretest": "node ../node_modules/rimraf/bin.js ./spec/obj && node ../node_modules/typescript/bin/tsc --project ./spec/tsconfig.json && cd ./spec/obj && npm init -y && npm install ../../../signalr",
-    "test": "node ../node_modules/jasmine/bin/jasmine.js ./spec/obj/spec/**/*.spec.js"
+    "test": "echo \"Run 'npm test' in the 'clients\\ts' folder to test this package\" && exit 1"
```
anurse: We no longer support `npm test` in a single package folder. It's faster to run them all together in one parallel run and it allows the config to be common.

clients/ts/signalr-protocol-msgpack/tests/MessagePackHubProtocol.test.ts
```diff
@@ -146,16 +146,16 @@ describe("MessageHubProtocol", () => {
...
+        ["message with no payload", [0x00], "Invalid payload."],
+        ["message with empty array", [0x01, 0x90], "Invalid payload."],
+        ["message without outer array", [0x01, 0xc2], "Invalid payload."],
+        ["message with invalid headers", [0x03, 0x92, 0x01, 0x05], "Invalid headers."],
+        ["Invocation message with invalid invocation id", [0x03, 0x92, 0x01, 0x80], "Invalid payload for Invocation message."],
+        ["StreamItem message with invalid invocation id", [0x03, 0x92, 0x02, 0x80], "Invalid payload for StreamItem message."],
+        ["Completion message with invalid invocation id", [0x04, 0x93, 0x03, 0x80, 0xa0], "Invalid payload for Completion message."],
+        ["Completion message with missing result", [0x05, 0x94, 0x03, 0x80, 0xa0, 0x01], "Invalid payload for Completion message."],
+        ["Completion message with missing error", [0x05, 0x94, 0x03, 0x80, 0xa0, 0x03], "Invalid payload for Completion message."],
+    ] as Array<[string, number[], string]>).forEach(([name, payload, expectedError]) =>
```
anurse: Jest's `.toThrow` matcher expects a `string` or `RegExp` that it matches against the message, rather than an `Error` object.

clients/ts/signalr-protocol-msgpack/tests/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
```
anurse: The tests need their own `tsconfig.json` for VSCode IntelliSense and squigglies to work. The `tsconfig.json` in the root of the package directory excludes the tests because we don't want them ending up in the npm package, so VSCode dutifully considers them not part of any project and uses the default settings which can't find `@aspnet/signalr`, etc.

clients/ts/package.json
```diff
@@ -5,21 +5,47 @@
...
     "tslint": "^5.9.1",
     "typescript": "^2.7.1",
     "uglify-js": "^3.3.5"
+  },
+  "jest": {
+    "globals": {
+      "ts-jest": {
+        "tsConfigFile": "./tsconfig.jest.json",
+        "skipBabel": true,
+        "enableTsDiagnostics": true
```
anurse: Without this, TypeScript semantic errors (i.e. all the type-checking) won't cause the test to fail.

clients/ts/signalr-protocol-msgpack/tests/BinaryMessageFormatter.test.ts
```diff
@@ -18,16 +18,16 @@ describe("Binary Message Formatter", () => {
...
+        [[0x80], "Cannot read message size."],
+        [[0x02, 0x01, 0x80, 0x80], "Cannot read message size."],
+        [[0x07, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x80], "Cannot read message size."], // the size of the second message is cut
+        [[0x07, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x01], "Incomplete message."], // second message has only size
+        [[0xff, 0xff, 0xff, 0xff, 0xff], "Messages bigger than 2GB are not supported."],
+        [[0x80, 0x80, 0x80, 0x80, 0x08], "Messages bigger than 2GB are not supported."],
+        [[0x80, 0x80, 0x80, 0x80, 0x80], "Messages bigger than 2GB are not supported."],
+        [[0x02, 0x00], "Incomplete message."],
+        [[0xff, 0xff, 0xff, 0xff, 0x07], "Incomplete message."],
+    ] as Array<[number[], string]>).forEach(([payload, expectedError]) => {
```
anurse: Jest's `.toThrow` matcher expects a `string` or `RegExp` that it matches against the message, rather than an `Error` object (which is what Jasmine expected).

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -17,14 +17,14 @@ const enum ConnectionState {
     Disconnected,
 }
 
-interface INegotiateResponse {
-    connectionId: string;
-    availableTransports: IAvailableTransport[];
-    url: string;
-    accessToken: string;
+export interface INegotiateResponse {
```
anurse: These are used by some tests now. This file is not exported from `index.ts` though so this has no effect on the public API. Plus, the only changes are to interfaces which don't actually show up in the generated JavaScript.

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -147,6 +147,14 @@ export class LongPollingTransport implements ITransport {
                 }
             }
         } finally {
+            // Mark the poll aborted.
+            this.pollAbort.abort();
+
+            // If we were on the shutdown timer, stop it.
+            if (this.shutdownTimer) {
+                clearTimeout(this.shutdownTimer);
+            }
```
anurse: I noticed this bug while running the tests and can fork it off into a separate PR if we don't take this for RTM. We need to mark the poll "aborted" when we shutdown normally so we don't start the shutdownTimer. And, if the shutdownTimer is running, we need to terminate it because we have shut down.

This does not cause a functional issue, but it manifested because Jest waits for all timers to elapse before shutting down. You'll see this manifest as a pause after Jest says the test suite has finished but before it actually terminates. It's useful that it does that, because we'd otherwise miss things like this.

clients/ts/signalr/tests/AbortSignal.test.ts
```diff
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 import { AbortController } from "../src/AbortController";
-import { asyncit as it } from "./Utils";
```
anurse: Jest's Jasmine supports Promises natively!

clients/ts/tsconfig.base.json
```diff
@@ -14,6 +14,15 @@
         "suppressImplicitAnyIndexErrors": true,
         "noEmitOnError": true,
         "stripInternal": true,
-        "lib": [ "es5", "es2015.promise", "es2015.iterable", "dom" ]
+        "lib": [ "es5", "es2015.promise", "es2015.iterable", "dom" ],
+        "baseUrl": ".",
+        "paths": {
```
anurse: This section provides mappings so TypeScript can find the packages. This is really just used by `signalr-protocol-msgpack` to find `signalr`. This has **no effect** on the generated code, it's just so TypeScript can find the definition files for the package.

clients/ts/package.json
JamesNK: I would say add a comment but JSON so yeah...

clients/ts/signalr/src/LongPollingTransport.ts
JamesNK: For visibility I think it should be a separate PR anyway.

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -337,31 +305,30 @@ describe("HttpConnection", () => {
                     return "";
                 }
                 return new HttpResponse(204, "No Content", "");
-            });
+            })
+            .on("DELETE", (r) => new HttpResponse(202));
```
JamesNK: Was the test http client's lack of handling deletes not showing up in the old test framework?

clients/ts/package.json
anurse: Yeah... There are tricks to use, since most of the package.json settings and subsettings ignore unknown properties (`"_comment": "A comment"`), but I don't think it really works. I may throw a README.md in `clients/ts` and put some notes in there.

anurse:
> Also I would like a wiki page documenting how to run and debug the tests in VSCode.

I'm planning to do one better and make sure the `launch.json` is set up and working for this, but I'm saving that work for MQ.

clients/ts/signalr/tests/HttpConnection.test.ts
anurse: No, I added `try { ... } finally { await connection.stop(); }` to clean up background things during tests and it meant `DELETE` requests started being issued.

clients/ts/signalr/src/LongPollingTransport.ts
anurse: You sound like my conscience :). Yeah, that's the right thing to do.

JamesNK:
FYI this needs to be merged tomorrow (Friday). Please take some time to give feedback @davidfowl @BrennanConroy @mikaelm12

clients/ts/signalr/spec/HubConnection.spec.ts
```diff
@@ -1,931 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
BrennanConroy: ?

anurse: Tests were moved from `spec` directory to `tests` directory, but this one was modified a bit, so it appears this way.

PR #2195, opened 2018-05-03
clean up the shutdown timer

anurse:
Found during #2193, forking off into a separate PR. We should take this for RTM even if we don't take #2193.

When `connection.stop` is called, we check the `pollAborted` "AbortSignal" (JavaScript equivalent of a CancellationToken) to see the poll is still running and if it is, we set a timer to see if the server will shut it down (we already sent the `DELETE` which should stop the server). However, if the poll completes normally before `connection.stop` is called, we don't set that token so we set up the timer anyway, logging an inaccurate log message 5 seconds after we shut down. There's no functional "break", but it's bad to log something that isn't true. Also, in Node JS applications (or tests), leaving this timeout on the queue can cause delays in shutting down.

This can also happen if we **are** mid-poll but the server **does** stop the poll before the timeout elapses. In that case, the timeout fires because we never clean it up.

Testing this requires asserting a timer callback never fires, which is a little tricky but I hacked something up. Jest has a set of fake timers so once #2193 is in and we start revamping our tests, we can look at using that to make it cleaner

### DIFF
--- a/clients/ts/signalr/spec/LongPollingTransport.spec.ts
+++ b/clients/ts/signalr/spec/LongPollingTransport.spec.ts
@@ -6,7 +6,8 @@ import { LongPollingTransport } from "../src/LongPollingTransport";
 import { ConsoleLogger } from "../src/Utils";
 
 import { TestHttpClient } from "./TestHttpClient";
-import { asyncit as it, PromiseSource } from "./Utils";
+import { asyncit as it, PromiseSource, delay } from "./Utils";
+import { HttpError, TimeoutError } from "../src/Errors";
 
 describe("LongPollingTransport", () => {
     it("shuts down poll after timeout even if server doesn't shut it down on receiving the DELETE", async () => {
@@ -48,6 +49,8 @@ describe("LongPollingTransport", () => {
                     return new HttpResponse(200);
                 } else {
                     await deleteReceived.promise;
+                    // Force the shutdown timer to be registered by not returning inline
+                    await delay(10);
                     pollCompleted.resolve();
                     return new HttpResponse(204);
                 }
@@ -56,7 +59,8 @@ describe("LongPollingTransport", () => {
                 deleteReceived.resolve();
                 return new HttpResponse(202);
             });
-        const transport = new LongPollingTransport(client, null, NullLogger.instance, false);
+        const logMessages: string[] = [];
+        const transport = new LongPollingTransport(client, null, NullLogger.instance, false, 100);
 
         await transport.connect("http://example.com", TransferFormat.Text);
         await transport.stop();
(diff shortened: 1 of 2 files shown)


clients/ts/signalr/spec/LongPollingTransport.spec.ts
```diff
@@ -56,12 +59,65 @@ describe("LongPollingTransport", () => {
                 deleteReceived.resolve();
                 return new HttpResponse(202);
             });
-        const transport = new LongPollingTransport(client, null, NullLogger.instance, false);
+        const logMessages: string[] = [];
```
JamesNK: I believe this isn't needed

clients/ts/signalr/spec/LongPollingTransport.spec.ts
```diff
@@ -56,12 +59,65 @@ describe("LongPollingTransport", () => {
...
+            await pollCompleted.promise;
+
+            // Wait for the shutdown timeout to elapse
+            // This can be much cleaner when we port to Jest because it has a built-in set of
+            // fake timers!
+            await delay(150);
+
+            expect(logMessages)
+                .not
+                .toContain("(LongPolling transport) server did not terminate after DELETE request, canceling poll.");
```
JamesNK: eeeeehhhhh, so flakey.

JamesNK: Instead of the logs, what about using private reflection and wrapping AbortController (so casting the transport to any and then setting the private field) with our test version, and inspecting it?

anurse: `LongPollingTransport` isn't exported from `index` so we can even just have a public property for this. I think I'm going to go simpler and just have a "stopped" flag. It's also cleaner to set a flag like that instead of checking the poll aborted token when stopping.

anurse:
:up: :date: less flaky tests

clients/ts/signalr/spec/LongPollingTransport.spec.ts
```diff
@@ -56,12 +60,63 @@ describe("LongPollingTransport", () => {
...
+            // This should complete, because the DELETE request triggers it to stop.
+            await pollCompleted.promise;
+
+            // Wait for the shutdown timeout to elapse
+            // This can be much cleaner when we port to Jest because it has a built-in set of
+            // fake timers!
+            await delay(150);
+
+            // The pollAbort token should be left unaborted because we shut down gracefully.
+            expect(transport.pollAborted).toBe(false);
```
JamesNK: Not `transport.pollAborted()`? Is the method called by the test framework automatically?

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -22,6 +23,12 @@ export class LongPollingTransport implements ITransport {
     private shutdownTimer: any; // We use 'any' because this is an object in NodeJS. But it still gets passed to clearTimeout, so it doesn't really matter
     private shutdownTimeout: number;
     private running: boolean;
+    private stopped: boolean;
+
+    // This is an internal type, not exported from 'index' so this is really just internal.
+    public get pollAborted() {
```
JamesNK: getPollAborted?

clients/ts/signalr/spec/LongPollingTransport.spec.ts
anurse: `get foo()` is ECMAScript/TypeScript property syntax. So `pollAborted` is a property, not a method.

clients/ts/signalr/src/LongPollingTransport.ts
anurse: See above comment. This is a property getter definition

Issue #2133, opened 2018-04-23
NegotiateProtocol doesn't handle responses with null AvailableTransports


davidfowl:

qinyuanpei:
Yeah, I meet same question like you. I am too sad to say anything.

anurse:
Re-closing, we're going to open new issues for patch work.

PR #2202, opened 2018-05-04
Handle null collections when writing NegotiateResponse

JamesNK:
https://github.com/aspnet/SignalR/issues/2133


test/Microsoft.AspNetCore.Http.Connections.Tests/NegotiateProtocolTests.cs
```diff
@@ -50,5 +54,35 @@ public void ParsingNegotiateResponseMessageThrowsForInvalid(string payload, stri
 
             Assert.Equal(expectedMessage, exception.InnerException.Message);
         }
+
+        [Fact]
+        public void WriteNegotiateResponseWithNullAvailableTransports()
+        {
+            MemoryBufferWriter writer = new MemoryBufferWriter();
```
BrennanConroy: Wrap in a using or use the `Get` `Return` pattern

Issue #2127, opened 2018-04-21
Expose a connection state from HubConnection


davidfowl:
We should expose a ConnectionState enum from the HubConnection so that it's possible to avoid throwing from SendAsync if the HubConnection wasn't connected. Today it requires you storing your own bool and setting it in the Closed event.

JamesNK:
Also need state on JS client.

moozzyk:
Only that it can still throw if the connection gets disconnected after the check and before actually sending data...

davidfowl:
Yep.

MagicAndre1981:
when will this be released? Microsoft.AspNetCore.SignalR.Client Version 1.0.2 doesn't include this.

I assume the 2.2.0 means Asp.Net Core 2.2? Why not sync SignalR versions to match Asp.Net Core to easily see which version has the fix?

davidfowl:
> when will this be released? Microsoft.AspNetCore.SignalR.Client Version 1.0.2 doesn't include this.

1.1, which will ship with ASP.NET Core 2.2

> I assume the 2.2.0 means Asp.Net Core 2.2? Why not sync SignalR versions to match Asp.Net Core to easily see which version has the fix?

We'll sync it in 3.0

davidfowl:
That's correct. There are packages on the myget feed if you want to try them out right now but the final build will be with ASP.NET Core 2.2

MagicAndre1981:
so I have to use [this version](https://dotnet.myget.org/feed/aspnetcore-dev/package/nuget/Microsoft.AspNetCore.SignalR/1.1.0-preview1-34892)?

MagicAndre1981:
the version works and I see the State property. How reliable are the 1.1 previews? Will we see radical API changes or only for 3 builds? 

btw, it would be better to have 2 feeds (1 feed for 3.0 and 1 for 2.2/1.1).

MagicAndre1981:
I've tested the nightly versions a bit more, but it doesn't work the same like in the old SignalR. In my case 3 time the state is disconnected and I start 3 connections not only 1 like in old SignalR.

```
Application started. Press Ctrl+C to shut down.
dbug: Microsoft.AspNetCore.Server.Kestrel[1]
      Connection id "0HLG2H0UTOJ3O" started.
dbug: Microsoft.AspNetCore.Server.Kestrel[1]
      Connection id "0HLG2H0UTOJ3P" started.
[...]

PR #2204, opened 2018-05-04
[WIP] Add HubConnection.Status

JamesNK:
https://github.com/aspnet/SignalR/issues/2127

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -20,6 +20,13 @@ using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
+    // TODO: Documentation
+    public enum HubConnectionStatus
+    {
+        Started,
+        Stopped
+    }
+
     /// <summary>
     /// A connection used to invoke hub methods on a SignalR Server.
     /// </summary>
@@ -56,6 +63,22 @@ namespace Microsoft.AspNetCore.SignalR.Client
         /// </summary>
         public TimeSpan ServerTimeout { get; set; } = DefaultServerTimeout;
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
+
+        // TODO: Documentation
+        public HubConnectionStatus Status
+        {
+            get
+            {
+                // Local copy for thread-safety
+                var connectionState = _connectionState;
+                if (connectionState == null || connectionState.Stopped)
+                {
+                    return HubConnectionStatus.Stopped;
+                }
+
+                return HubConnectionStatus.Started;
+            }
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="HubConnection"/> class.
@@ -986,6 +1009,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 get => _stopping;
                 set => _stopping = value;
             }
+
+            public bool Stopped => _stopTcs?.Task.Status == TaskStatus.RanToCompletion;
 
             public ConnectionState(ConnectionContext connection, HubConnection hubConnection)
             {
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
@@ -54,8 +54,11 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
                 var testConnection = new TestConnection();
                 await AsyncUsing(CreateHubConnection(testConnection), async connection =>
                 {
+                    Assert.Equal(HubConnectionStatus.Stopped, connection.Status);
+
                     await connection.StartAsync();
                     Assert.True(testConnection.Started.IsCompleted);
+                    Assert.Equal(HubConnectionStatus.Started, connection.Status);
                 });
             }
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -20,6 +20,13 @@
 
 namespace Microsoft.AspNetCore.SignalR.Client
 {
+    // TODO: Documentation
+    public enum HubConnectionStatus
+    {
+        Started,
```
mikaelm12: Hmm, would connected  make more sense here?

JamesNK: Could do! I went with Started and Stopped because of StartAsync and StopAsync.

🚲 🏠 @davidfowl @anurse

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -57,6 +64,22 @@ public partial class HubConnection
         public TimeSpan ServerTimeout { get; set; } = DefaultServerTimeout;
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
 
+        // TODO: Documentation
+        public HubConnectionStatus Status
+        {
+            get
+            {
+                // Local copy for thread-safety
+                var connectionState = _connectionState;
```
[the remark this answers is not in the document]
JamesNK: Yes but the field can be set to null by another thread in-between checking it is not null and getting IsStopped.

Copy is the wrong word. A better comment would be "Local reference for thread-safety".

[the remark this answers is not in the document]
acjh: Ah. Thanks for the clarification 🙂

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: I like `Started` and `Stopped` as well, given the method names.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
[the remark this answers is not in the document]
anurse: `?.` will do this for you ([source](https://sharplab.io/#v2:D4AQDABCCMDcCwAoEBmKAmCBhCBvJEhEADgE4CWAbgIYAuAptgPYB2L9AxreawMq11GAfQDOAhgkREIBIqmZtO3PuICuIiPzrq8s6YQDm9Wrqn7z5AGYAKUePoB+AHT8mxYvQAmEBw4i1SVXoASlNzcLkAdgV2Lh4WLVp1F1o3D09JCOkAXz0skGisVljlBLURFOpSBgy8nLzcs0I8+RBMIsU4lUEw8PkAIyYmABtNVPcvPCNaWEb6ppkF+XoWVQBbGKV4xJ18Bektaq8AGjqiVwnPBqRsoA))

It's mostly just a clever trick though, so I don't really mind either way. Was mostly just curious if if would do the Right Thing.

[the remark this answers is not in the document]
JamesNK: I like taking a local variable for thread-safety issues because it gives a good opportunity to comment and make it obvious that it is a thing future readers of the code need to think about.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
JamesNK: Hmmm, the methods are Started and Stopped but they're on a class with connection in the name.

I think I prefer `Disconnected` and `Connected`.

JamesNK:
🚲 🏠

JamesNK:

JamesNK:
Hey @davidfowl @anurse, is there any value in having a Connecting and Disconnecting state?

anurse:
@JamesNK Yeah, I guess that makes sense. Automatic Reconnect is also going to to throw a complexity in there (do we need a `Reconncting` state? or do we just transition back to `Connecting`)

JamesNK:
@davidfowl said two is fine. It's an enum so no reason more can't be added as needed.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -215,6 +224,58 @@ Task<ConnectionContext> ConnectionFactory(TransferFormat format)
                 });
             }
 
+            [Fact]
+            public async Task StatusIsNotStartedUntilStartAsyncIsFinished()
```
mikaelm12: The title should be "StatusIsNotConnected..."

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -215,6 +224,58 @@ Task<ConnectionContext> ConnectionFactory(TransferFormat format)
...
+
+                    // Wait for start to finish
+                    await startTask;
+
+                    Assert.Equal(HubConnectionState.Connected, connection.State);
+                });
+            }
+
+            [Fact]
+            public async Task StatusIsStoppedInCloseEvent()
```
mikaelm12: And here replace Stopped with Disconnected

PR #2205, opened 2018-05-04
 Throw error on unexpected close

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2184,6 +2184,38 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler).OrTimeout();
+
+                await client.Connection.Application.Output.WriteAsync(Encoding.UTF8.GetBytes(new[] { '{' })).OrTimeout();
+
+                // Close connection
+                client.Connection.Application.Output.Complete();
+
+                // Ignore message from OnConnectedAsync
+                _ = await client.ReadAsync().OrTimeout();
```
JamesNK: No need for OrTimeout if Task is ignored

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -643,7 +644,8 @@ private void CheckDisposed()
                                     break;
                                 }
                             }
-                            else if (result.IsCompleted)
+
+                            if (result.IsCompleted)
```
davidfowl: Won't this throw if there was a fully parsed handshake, and the pipe was completed? I'd just get rid of the `buffer.IsEmpty` check, or don't throw if the buffer is empty.

davidfowl: Or we do like the server and just return directly out of the method.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
davidfowl: Why is there any underscore.

JamesNK: +1. We assign to underscore with Task because there is a warning if you don't do anything with a returned task. Ignoring other types from a method shouldn't create a warning.

Issue #2130, opened 2018-04-22
 Embed version in the JavaScript client 


JamesNK:
We should log the version of the client (.NET/JS/whatever) and the version of the server to help diagnose issues when devs create issues with attached logs.

JamesNK:
Can pull version in .NET from assembly version attribute. Is the version embedded anywhere in JS?

JamesNK:
How do we embed the version number in the TS file? It is not something I've done before.

Lodash does it but I couldn't track down their build scripts.

Thoughts, ideas?

anurse:
We'd basically do it the same way we do it in any other file. We'd have to just do a simple find-replace for a string in one of our build scripts, such as `build/repo.targets`

muratg:
Updated the title to reflect how we approach this: https://github.com/aspnet/SignalR/pull/2229

anurse:
https://github.com/aspnet/SignalR/commit/f005e80b14055a4ca629a17413c4c8647ccf70a5

PR #2229, opened 2018-05-08
Embed version in the JavaScript client

anurse:
Related to https://github.com/aspnet/SignalR/issues/2130

The way this works for now is that it takes advantage of the fact that we update the `package.json` version from MSBuild during the packaging step. This happens **after** the build phase, so I added a script to the `prepack` phase in NPM that scans all files under `dist` (build outputs) and replaces a token with the actual version in `package.json`

### DIFF
--- a/clients/ts/build/embed-version.js
+++ b/clients/ts/build/embed-version.js
@@ -13,7 +13,7 @@ for (let i = 0; i < args.length; i++) {
 }
 
 function processDirectory(dir) {
-    if(verbose) {
+    if (verbose) {
         console.log(`processing ${dir}`);
     }
     for (const item of fs.readdirSync(dir)) {
--- a/clients/ts/build/embed-version.js
+++ b/clients/ts/build/embed-version.js
@@ -1,0 +1,26 @@
+const path = require("path");
+const fs = require("fs");
+const pkg = require(path.resolve(process.cwd(), "package.json"));
+
+function processDirectory(dir) {
+    for (const item of fs.readdirSync(dir)) {
+        const stats = fs.statSync(item);
+        if (stats.isDirectory()) {
+            processDirectory(item);
+        } else if (stats.isFile()) {
+            processFile(item);
+        }
+    }
+}
+
+const SEARCH_STRING = "0.0.0-DEV_VERSION";
+/**
+ * @param {string} file 
+ */
+function processFile(file) {
+    if (file.endsWith(".js") || file.endsWith(".ts")) {
+        let content = fs.readFileSync(file);
+        content = content.toString().replace(SEARCH_STRING, pkg.version);
+        fs.writeFileSync(file, content);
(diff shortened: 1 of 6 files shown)


clients/ts/build/embed-version.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const SEARCH_STRING = "0.0.0-DEV_BUILD";
+function processFile(file) {
+    if (file.endsWith(".js") || file.endsWith(".ts")) {
+        if (verbose) {
+            console.log(`processing ${file}`);
+        }
+        let content = fs.readFileSync(file);
+        content = content.toString().replace(SEARCH_STRING, pkg.version);
```
BrennanConroy: `pkg.version` just works?!

anurse: Yep :)

clients/ts/signalr-protocol-msgpack/src/index.ts
```diff
@@ -1,4 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+// Version token that will be replaced by the prepack command
+export const VERSION = "0.0.0-DEV_BUILD";
```
JamesNK: JS docs

clients/ts/signalr/src/index.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+// Version token that will be replaced by the prepack command
+export const VERSION = "0.0.0-DEV_BUILD";
```
JamesNK: JS docs

anurse:
> What can we do to unit test this?

We can't unit test this really, since the version isn't burned in until we package (which is after unit testing). I think for RTM this will just have to be a manual verification item.

anurse:
https://github.com/aspnet/SignalR/issues/2256 filed for automation

anurse:
:up: :date: Added a version number header to the UMD build. I didn't originally think it was necessary but it's useful for the minified version, where it's hard to see the version number. I used JSDoc "official" tags for both the version and the library title

clients/ts/signalr-protocol-msgpack/src/index.ts
```diff
@@ -1,4 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+// Version token that will be replaced by the prepack command
+/** The version of the SignalR client. */
```
JamesNK: Doc should say the version number is for the SignalR msgpack library

PR #2231, opened 2018-05-09
Update Issue Template

anurse:
We changed some of the logging APIs and I decided it would be better to just write a Wiki page with a full diagnostics guide to refer to from the template. We can also link to the guide when we need to ask for more information (logs, traces, etc.).

See the diagnostics guide here: https://github.com/aspnet/SignalR/wiki/Diagnostics-Guide - At some point it can be elevated to an official "Doc" on aspnet/Docs.

The Wiki doesn't have any review mechanism, so feel free to offer suggestions for that guide here or via other communication media :).


anurse:

Issue #2215, opened 2018-05-06
Additional tests for Utf8 writer


JamesNK:
Ensure that the exception here has been fixed and unit tested https://github.com/aspnet/SignalR/issues/2187#issuecomment-386871789

BrennanConroy:
@JamesNK Can this be closed?

PR #2238, opened 2018-05-09
Test combining multi-byte character writes with individual chars

JamesNK:
https://github.com/aspnet/SignalR/issues/2215

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/Utf8BufferTextWriterTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/Utf8BufferTextWriterTests.cs
@@ -301,18 +301,18 @@ namespace Microsoft.AspNetCore.SignalR.Common.Tests.Internal.Protocol
             const string testString = "aいbろcdはにeほfへどghi\uD800\uDC00";
             const int iterations = 5;
 
-            var bufferWriter1 = new TestMemoryBufferWriter(segmentSize);
+            var testBufferWriter = new TestMemoryBufferWriter(segmentSize);
             var sb = new StringBuilder();
 
-            using (var writer = new Utf8BufferTextWriter())
+            using (var textWriter = new Utf8BufferTextWriter())
             {
-                writer.SetWriter(bufferWriter1);
+                textWriter.SetWriter(testBufferWriter);
 
                 for (int i = 0; i < iterations; i++)
                 {
-                    writer.Write('"');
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/Utf8BufferTextWriterTests.cs
```diff
@@ -292,11 +292,47 @@ private void WriteMultiByteCharactersToSmallBuffers()
             Assert.Equal(testString, Encoding.UTF8.GetString(bufferWriter.ToArray()));
         }
 
+        public static IEnumerable<object[]> SegmentSizes => Enumerable.Range(4, 36).Select(segmentSize => new object[] { segmentSize });
+
+        [Theory]
+        [MemberData(nameof(SegmentSizes))]
+        public void WriteUnicodeStringAndCharsWithVaringSegmentSizes(int segmentSize)
```
anurse: So what is this testing that the existing test doesn't handle? Is it just adding some fuzzing to help ensure we don't encounter other issues? The existing test was designed to verify the scenario we encountered (trying to write to a buffer without enough capacity for a single character).

I'm all for fuzzing, just trying to understand if there's a case we didn't cover in the earlier testing.

JamesNK: The exception I linked to came from a call to `Write(char)` which ended up out of bounds somehow.

I wasn't able to recreate the failure so the some fuzzing is just to make sure it doesn't come up in future.

As I typed this it occured to me that `Write(char)` could have a multibyte which should also be tested.

BrennanConroy: nit: varying

anurse: Couldn't we just have a variant of the test immediately above this that iterated through the string writing individual chars? I'd prefer we explicitly trigger the situation we are trying to prevent rather than relying on fuzzing segments sizes. Maybe I'm misunderstanding the situation though :)

JamesNK: I've split tests up via parameters

PR #2242, opened 2018-05-09
Add Content-Type header in HttpClient.ts

YourFriendBmo:
This change prevents an error in react-native when running on the Android platform. If the Content-Type header is not set, an error is thrown.

Setting the value to `text/plain;charset=UTF-8` matches what Chrome sends without the header being specified.

Error w/out change:
image

When setting the `LogLevel` to `Debug`,  it would fail after:

`Debug: Sending negotiation request: ...`

When debugging the `xhr.onerror` function in HttpClient.ts, the details on the `xhr` object shows the underlying error from react-native is "Payload is set but no content-type header specified".

With this proposed change, the error is not thrown and results in the connection established and working as expected.

Thanks.

### DIFF
--- a/clients/ts/signalr/src/HttpClient.ts
+++ b/clients/ts/signalr/src/HttpClient.ts
@@ -163,6 +163,7 @@ export class DefaultHttpClient extends HttpClient {
             xhr.open(request.method, request.url, true);
             xhr.withCredentials = true;
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
+            xhr.setRequestHeader("Content-Type", "text/plain;charset=UTF-8");
 
             if (request.headers) {
                 Object.keys(request.headers)
--- a/clients/ts/signalr/src/HttpClient.ts
+++ b/clients/ts/signalr/src/HttpClient.ts
@@ -163,7 +163,7 @@ export class DefaultHttpClient extends HttpClient {
             xhr.open(request.method, request.url, true);
             xhr.withCredentials = true;
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
-            // Explicitly setting the Content-Type header is need for React Native on Android platform. 
+            // Explicitly setting the Content-Type header for React Native on Android platform. 
             xhr.setRequestHeader("Content-Type", "text/plain;charset=UTF-8");
 
             if (request.headers) {
--- a/clients/ts/signalr/src/HttpClient.ts
+++ b/clients/ts/signalr/src/HttpClient.ts
@@ -163,7 +163,7 @@ export class DefaultHttpClient extends HttpClient {
             xhr.open(request.method, request.url, true);
             xhr.withCredentials = true;
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
-            // Explicitly setting the Content-Type header for React Native on Android platform. 
+            // Explicitly setting the Content-Type header for React Native on Android platform.
             xhr.setRequestHeader("Content-Type", "text/plain;charset=UTF-8");
 
             if (request.headers) {
--- a/clients/ts/signalr/src/HttpClient.ts
+++ b/clients/ts/signalr/src/HttpClient.ts
@@ -163,6 +163,7 @@ export class DefaultHttpClient extends HttpClient {
             xhr.open(request.method, request.url, true);
             xhr.withCredentials = true;
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
+            // Explicitly setting the Content-Type header is need for React Native on Android platform. 
             xhr.setRequestHeader("Content-Type", "text/plain;charset=UTF-8");
 
             if (request.headers) {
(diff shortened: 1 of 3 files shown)


YourFriendBmo:
@anurse adding a test as described above won't work for this scenario because the method we're testing is not used for the portion that gets executed once inside of the `TestHub.cs` file. 

image

^^ the first HTTP call in the screenshot is the one that gets to the `TestHub.cs` file, that comes from `eventsource` and not `xhr` so it does not pass through the `HttpClient.ts` file.

I can add a more generic test of invoking the `send` method of `HttpClient.ts` and creating a middleware listening for a specific url path and return the `Content-Type` header that is being sent from xhr. Thoughts?

Thanks.

anurse:
Ah yeah, but that's only when using Server-Sent Events or WebSockets. If you use Long Polling, the HttpClient is used for the poll request, which is what makes it to `TestHub`.

YourFriendBmo:
I supposedly am using Long Polling:

```js
it("populates the Content-Type header when sending XMLHttpRequest", async (done) => {
        const hubConnection = getConnectionBuilder(undefined, TESTHUB_NOWEBSOCKETS_ENDPOINT_URL)
            .withHubProtocol(new JsonHubProtocol())
            .build();

        try {
            await hubConnection.start();

            // Check what transport was used by asking the server to tell us.
            expect(await hubConnection.invoke("GetActiveTransportName")).toEqual("LongPolling");
            expect(await hubConnection.invoke("GetContentTypeHeader")).toEqual("text/plain;charset=UTF-8");
            done();
        } catch (e) {
            fail(e);
        }
```

But it sends `eventsource` for the actual communication to the hub. Those 2 xhr POSTS do route through `HttpClient.ts` but do not hit the actual method inside of `TestHub.cs` - verified  by setting breakpoint in `TestHub.cs` which only hits on the `GET` from `eventsource`.

image


Also, I can confirm that those 2 xhr requests are getting to the modified code:
image

anurse:
Jasmine doesn't stop the test just because an assert fails. So the first assert is also failing, but you're not seeing it (Usually it should show both though, so it's surprising that you're only seeing one).

If you're seeing an EventSource request you definitely aren't using LongPolling. The code bears that out too

```javascript
        const hubConnection = getConnectionBuilder(undefined, TESTHUB_NOWEBSOCKETS_ENDPOINT_URL)
            .withHubProtocol(new JsonHubProtocol())
            .build();
```

The first parameter to `getConnectionBuilder` is a `TransportType` to use to restrict the transport used by the client. Just pass `HttpTransportType.LongPolling` in there to force Long Polling.

YourFriendBmo:
*facepalm* It is because Chrome was caching the js file. Passing `HttpTransportType.LongPolling` indeed does the trick. Thanks for the assistance @anurse ! 

image

anurse:
> It is because Chrome was caching the js file

That happens all the time :). So annoying! Glad you got it worked out, I'll take a look at the test!

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -163,6 +163,7 @@ export class DefaultHttpClient extends HttpClient {
             xhr.open(request.method, request.url, true);
             xhr.withCredentials = true;
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
+            xhr.setRequestHeader("Content-Type", "text/plain;charset=UTF-8");
```
anurse: Pop a comment on here indicating that this is needed by React-Native on Android, so we don't accidentally remove it later.

anurse:
Travis failure wasn't related to the change. I restarted the build.

clients/ts/signalr/src/HttpClient.ts
YourFriendBmo: Will do. Do you prefer at end of line or before the line?

JamesNK: Before the line please

JamesNK: Our TS linter is pretty strict. I think you need to remove the space at the end of the comment

> EXEC : error : /home/travis/build/aspnet/SignalR/clients/ts/signalr/src/HttpClient.ts[166, 96]: trailing whitespace

JamesNK:
Merged. Thanks for reporting the issue and contributing a PR for it.

Shhzdmrz:
Is this PR merged in new signalr 1.0.0 version?

anurse:
Yes, it should be. Are you seeing an issue with it?

JamesNK:
Isn't this 2.2?

Shhzdmrz:
still facing the same #2254

Shhzdmrz:
Adding ` xhr.setRequestHeader("Content-Type", "text/plain;charset=UTF-8"); to @aspnet/signalr/dist/browser/signalr.js resolved the issue.

anurse:
Ah yes, this didn't make the `1.0.0` release, we have the fix in and are targeting the `1.0.1` patch release.

PR #2246, opened 2018-05-10
fix #2225 by changing namespace for UseHub

### LINKED ISSUE #2225: UseHub is in the wrong namespace
UseHub is in the Microsoft.AspNetCore.SignalR namespace and it should be in the Microsoft.AspNetCore.Connections namespace (matching the IConnectionBuilder namespace).

/cc @muratg @anurse

anurse:
fixes #2225


anurse:
Declined by Shiproom and we can't take it in `dev`.

PR #2261, opened 2018-05-11
Use correct TaskCompletionSource ctor

### LINKED ISSUE #2262: AckHandler uses the wrong TCS constructor
It's passing in TaskContinuationOptions when it should be TaskCreationOptions. Otherwise it goes into the state field and has no effect. https://github.com/aspnet/SignalR/blob/5ff5c38325db23392accd17892a064252f34d44e/src/Microsoft.AspNetCore.SignalR.Redis/Internal/AckHandler.cs#L94

https://msdn.microsoft.com/en-us/library/dd449174(v=vs.110).aspx

Compare to: https://github.com/aspnet/KestrelHttpServer/pull/2567

halter73:
https://github.com/dotnet/corefx/issues/29664

Fixes #2262


davidfowl:
LGTM

PR #2264, opened 2018-05-13
NTP for benchmarking

BrennanConroy:
See https://github.com/aspnet/benchmarks/pull/464


benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -44,5 +44,10 @@ public void SendPayload(string payload)
         {
             // Dump the payload, we don't care
         }
+
+        public DateTime NTP()
+        {
+            return DateTime.UtcNow;
```
anurse: As I mentioned in the other PR, let's call this something like "GetCurrentTime" rather than "NTP" to avoid confusion.

Issue #2243, opened 2018-05-10
Closing one tab/connection kills notifications to a user using user id when the user have multiple connections from different tabs/browsers. Redis backplane


Boychenko:
Hi
I've found an issue with `Microsoft.AspNetCore.SignalR.Redis` as backplane.

Environment:
Microsoft.AspNetCore.SignalR: 1.0.0-rc1-final
Microsoft.AspNetCore.SignalR.Redis: 1.0.0-rc1-final
Javascript js file: 1.0.0-rc1-update1
Server: Kestrel / IIS
OS on server and client: Windows 10
Browser: Firefox (latest) and Chrome (latest)

Step to reproduce:
1 Open two tabs with the same user logged in
2 Close one tab

Expected: The app in the existing tab should be able to get notifications using user id
Actual: No notifications sent using user id delivered to this user anymore.

I made some research. There are my finding
* Everything works fine if I don't use Redis as a backplane.
* According to my understanding [unsubscribe](https://github.com/aspnet/SignalR/blob/6f6382799b00936521452cc6c5c516a5bb7f3ce8/src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs#L87) causing the problem. While it works fine for connection subscriptions because i'st closed anyway. For user subscriptions this is wrong. A user can have more than one connection to the same server.

Boychenko:
There is a [repository](https://github.com/Boychenko/SignalR.Redis_issue_sample/) with minimal app to reproduce.

anurse:
I've verified that this occurs. I'm working on a fix. We'll post back here with more information on scheduling/private builds/etc.

One thing you could try as a workaround (I realize it's not ideal) is using Groups instead of Users here. You have to manually manage the group, but it should not have this issue. We're going to look at fixing this issue properly, but until we have the fix released you could use that as a temporary workaround.

davidfowl:
The workaround would look like this:

```C#
public override Task OnConnectedAsync()
{
    Groups.AddToGroupAsync(Context.ConnectionId, "u-" + Context.UserIdentifier);
     return base.OnConnectedAsync();
}
```
Then you can use `Group("u" + userId)` to invoke instead of `User`. The u prefix is to avoid conflicting with group names.

Boychenko:
@anurse @davidfowl Thanks!
@davidfowl, workaround works fine. But I've improved it a bit.
```
public override Task OnConnectedAsync()
{
[...]

PR #2272, opened 2018-05-14
Fix multiple connections for a single user in Redis

BrennanConroy:
@anurse 
Let's start reviewing

https://github.com/aspnet/SignalR/issues/2243

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
@@ -1,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.SignalR.Redis.Internal
+{
+    internal class RedisSubscriptionManager
+    {
+        private class SubscriptionData
+        {
+            public readonly SemaphoreSlim Lock = new SemaphoreSlim(1, 1);
+            public readonly HubConnectionStore Connections = new HubConnectionStore();
+        }
+
+        // TODO: Investigate "memory leak" entries never get removed
+        private readonly ConcurrentDictionary<string, SubscriptionData> _subscriptions = new ConcurrentDictionary<string, SubscriptionData>(StringComparer.Ordinal);
+
+        public async Task AddSubscriptionAsync(string id, HubConnectionContext connection, Func<string, HubConnectionStore, Task> subscribeMethod)
+        {
+            var subscription = _subscriptions.GetOrAdd(id, _ => new SubscriptionData());
+
+            await subscription.Lock.WaitAsync();
+
+            try
+            {
+                subscription.Connections.Add(connection);
+
+                // Subscribe once
+                if (subscription.Connections.Count > 1)
+                {
+                    return;
+                }
+
+                await subscribeMethod(id, subscription.Connections);
+            }
+            finally
+            {
+                subscription.Lock.Release();
+            }
+        }
+
+        public async Task RemoveSubscriptionAsync(string id, HubConnectionContext connection, Func<string, Task> unsubscribeMethod)
+        {
+            if (!_subscriptions.TryGetValue(id, out var subscription))
+            {
+                return;
+            }
+
+            await subscription.Lock.WaitAsync();
+
+            try
+            {
+                if (subscription.Connections.Count > 0)
+                {
+                    subscription.Connections.Remove(connection);
+
+                    if (subscription.Connections.Count == 0)
+                    {
+                        await unsubscribeMethod(id);
+                    }
+                }
+            }
+            finally
+            {
+                subscription.Lock.Release();
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
@@ -30,12 +30,10 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
                 subscription.Connections.Add(connection);
 
                 // Subscribe once
-                if (subscription.Connections.Count > 1)
+                if (subscription.Connections.Count == 0)
                 {
-                    return;
+                    await subscribeMethod(id, subscription.Connections);
                 }
-
-                await subscribeMethod(id, subscription.Connections);
             }
             finally
             {
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
@@ -30,7 +30,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
                 subscription.Connections.Add(connection);
 
                 // Subscribe once
-                if (subscription.Connections.Count == 0)
+                if (subscription.Connections.Count == 1)
                 {
                     await subscribeMethod(id, subscription.Connections);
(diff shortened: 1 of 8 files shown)


davidfowl:
Can we make a separate structure now and just change groups and users to use it? It would likely be cleaner.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/Startup.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Castle.Core.Internal;
```
JamesNK: remove

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs
```diff
@@ -91,11 +90,77 @@ public RedisEndToEndTests(RedisServerFixture<Startup> serverFixture, ITestOutput
...
+        {
+            // Regression test:
+            // When multiple connections from the same user were connected and one left, it used to unsubscribe from the user channel
+            // Now we keep track of users connections and only unsubscribe when no users are listening
+            using (StartVerifableLog(out var loggerFactory, testName:
+                $"{nameof(CanSendAndReceiveUserMessagesWhenOneConnectionWithUserDisconnects)}_{transportType.ToString()}_{protocolName}"))
+            {
+                var protocol = HubProtocolHelpers.GetHubProtocol(protocolName);
+
+                var connection = CreateConnection(_serverFixture.FirstServer.Url + "/echo", transportType, protocol, loggerFactory, userName: "userA");
```
JamesNK: nit: firstConnection

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -461,17 +497,49 @@ private Task SubscribeToConnection(HubConnectionContext connection, HashSet<stri
...
+            await user.Lock.WaitAsync();
+
+            try
             {
-                var invocation = _protocol.ReadInvocation((byte[])data);
-                await connection.WriteAsync(invocation.Message);
-            });
+                user.Connections.Add(connection);
+
+                // Subscribe once
```
JamesNK: Is there a way to unit test this?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -365,6 +371,36 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
             await ack;
         }
 
+        private async Task RemoveUserAsync(HubConnectionContext connection)
+        {
+            var userChannel = _channels.User(connection.UserIdentifier);
```
JamesNK: Why is this called at the top of the method? Should the call be moved to where it is used?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -22,6 +22,7 @@ public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposab
         private readonly HubConnectionStore _connections = new HubConnectionStore();
         // TODO: Investigate "memory leak" entries never get removed
         private readonly ConcurrentDictionary<string, GroupData> _groups = new ConcurrentDictionary<string, GroupData>(StringComparer.Ordinal);
+        private readonly ConcurrentDictionary<string, UserData> _users = new ConcurrentDictionary<string, UserData>(StringComparer.Ordinal);
```
JamesNK: Should this be cleared in `Dispose`? Likewise groups?

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -365,6 +371,36 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
             await ack;
         }
 
+        private async Task RemoveUserAsync(HubConnectionContext connection)
+        {
+            var userChannel = _channels.User(connection.UserIdentifier);
+
+            if (!_users.TryGetValue(connection.UserIdentifier, out var user))
+            {
+                return;
```
JamesNK: Add a comment why we exit here

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -595,6 +664,12 @@ private class GroupData
             public readonly HubConnectionStore Connections = new HubConnectionStore();
         }
 
+        private class UserData
```
JamesNK: Maybe instead of duplicating GroupData with a different name, rename GroupData and reuse. SubscriptionConnections?

BrennanConroy:
Latest commit adds a `RedisSubscriptionManager` to remove duplicate code

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -20,8 +19,8 @@ namespace Microsoft.AspNetCore.SignalR.Redis
     public class RedisHubLifetimeManager<THub> : HubLifetimeManager<THub>, IDisposable where THub : Hub
     {
         private readonly HubConnectionStore _connections = new HubConnectionStore();
-        // TODO: Investigate "memory leak" entries never get removed
-        private readonly ConcurrentDictionary<string, GroupData> _groups = new ConcurrentDictionary<string, GroupData>(StringComparer.Ordinal);
+        private readonly RedisSubscriptionManager _groups = new RedisSubscriptionManager();
```
BrennanConroy: These could use the same `RedisSubscriptionManager`

src/Microsoft.AspNetCore.SignalR.Redis/RedisHubLifetimeManager.cs
```diff
@@ -102,6 +101,11 @@ public override Task OnDisconnectedAsync(HubConnectionContext connection)
                 }
             }
 
+            if (!string.IsNullOrEmpty(connection.UserIdentifier))
```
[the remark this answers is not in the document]
davidfowl: That's good!

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
```diff
@@ -0,0 +1,73 @@
...
+            var subscription = _subscriptions.GetOrAdd(id, _ => new SubscriptionData());
+
+            await subscription.Lock.WaitAsync();
+
+            try
+            {
+                subscription.Connections.Add(connection);
+
+                // Subscribe once
+                if (subscription.Connections.Count > 1)
```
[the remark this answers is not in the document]
BrennanConroy: If there is one subscription we just added it so we need to subscribe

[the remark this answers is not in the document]
BrennanConroy: The code is obvious
```
subscription.Connections.Add(connection);

// Subscribe once
if (subscription.Connections.Count > 1)
```
I think @anurse just didn't see the add directly above the if `subscription.Connections.Add(connection);`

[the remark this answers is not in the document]
anurse: Agreed, this doesn't actually seem that confusing, I just mis-read one line.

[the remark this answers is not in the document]
anurse: Inverting the `if` might actually make it clearer without introducing a comment that can get stale ;)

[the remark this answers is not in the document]
anurse: Plus it would be more consistent with how you handle remove

PR #2273, opened 2018-05-14
Fix multiple connections for a single user in Redis and MemoryLeak

BrennanConroy:
Version of https://github.com/aspnet/SignalR/pull/2272 with the memory leak fixed. ~~(TODO: Still need to manually verify that the leak is gone)~~

Used the `_connectionLock` because it was only used for connecting the Redis and the areas I added it shouldn't be under high contention. Can change it to separate locks if we want.

@anurse

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
@@ -21,7 +21,7 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
                 subscription = _subscriptions.GetOrAdd(id, _ => new HubConnectionStore());
 
                 // Subscribe once
-                if (subscription.Connections.Count == 1)
+                if (subscription.Count == 1)
                 {
                     firstSubscription = true;
                 }
--- a/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
@@ -16,55 +16,65 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Internal
             public readonly HubConnectionStore Connections = new HubConnectionStore();
         }
 
-        // TODO: Investigate "memory leak" entries never get removed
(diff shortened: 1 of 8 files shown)


BrennanConroy:
Confirmed memory leak gone, creating unique connections and closing them over and over again, capped at ~184MB, attaching debugger shows only current connection in the dictionary

anurse:
I'm not sure about this lock. We have to take it through the whole add/remove action. The only place we don't take it is when the subscription fires, in which you capture the `UserData` object anyway...

davidfowl:
Why do we have 2 different fixes here? The RedisSubscriptionManager should be in the release branch. It's just a better overall fix with less duplicate code.

BrennanConroy:
Changed the base to be the other PR. We have 2 PR's because this one is focused on fixing the memory leak and the other is fixing the bug. If shiproom is ok with the bug but not the memory leak fix then we have the change ready.

src/Microsoft.AspNetCore.SignalR.Redis/Internal/RedisSubscriptionManager.cs
```diff
@@ -16,57 +16,74 @@ private class SubscriptionData
...
-        // TODO: Investigate "memory leak" entries never get removed
         private readonly ConcurrentDictionary<string, SubscriptionData> _subscriptions = new ConcurrentDictionary<string, SubscriptionData>(StringComparer.Ordinal);
+        private readonly SemaphoreSlim _lock = new SemaphoreSlim(1, 1);
 
         public async Task AddSubscriptionAsync(string id, HubConnectionContext connection, Func<string, HubConnectionStore, Task> subscribeMethod)
         {
-            var subscription = _subscriptions.GetOrAdd(id, _ => new SubscriptionData());
-
-            await subscription.Lock.WaitAsync();
+            await _lock.WaitAsync();
```
JamesNK: A global lock wrapping all logic in add and remove subscription could easily become a bottleneck.

anurse: Yep, that's the main concern with doing this fix this way. However, the lock should, I believe, only be needed during connection/disconnection, which means it's not on the hottest of hot paths.

anurse:
I changed the base back to `release/2.1`, so we can review this as the complete change.

I think the lock is limited enough that the throughput impact would be minimal. It is only take during connection/disconnection, so throughput would be mostly affected by scenarios where lots of connections are being established/terminated rapidly. Most SignalR usage involves medium-to-long-running connections.

Issue #2163, opened 2018-04-29
Flaky test: hubConnection over LongPolling it sends DELETE request and waits for poll to terminate


davidfowl:
Seen in this PR https://github.com/aspnet/SignalR/pull/2162

```
not ok 90 hubConnection over LongPolling it sends DELETE request and waits for poll to terminate
    ---
      - messages:
        - message: Failed: Not Found
      - logs: 
        - level: Information
          timestamp: 2018-04-29T21:20:21.069Z
          message: Normalizing '/testhub' to 'http://127.0.0.1:4437/testhub'.
        - level: Debug
          timestamp: 2018-04-29T21:20:21.069Z
          message: Starting HubConnection.
        - level: Debug
          timestamp: 2018-04-29T21:20:21.069Z
          message: Starting connection with transfer format 'Text'.
        - level: Debug
          timestamp: 2018-04-29T21:20:21.069Z
          message: Sending negotiation request: http://127.0.0.1:4437/testhub/negotiate
        - level: Debug
          timestamp: 2018-04-29T21:20:21.074Z
          message: Skipping transport 'WebSockets' because it was disabled by the client.
        - level: Debug
          timestamp: 2018-04-29T21:20:21.074Z
          message: Skipping transport 'ServerSentEvents' because it was disabled by the client.
        - level: Debug
          timestamp: 2018-04-29T21:20:21.074Z
          message: Selecting transport 'LongPolling'
        - level: Trace
          timestamp: 2018-04-29T21:20:21.075Z
          message: (LongPolling transport) Connecting
        - level: Trace
          timestamp: 2018-04-29T21:20:21.075Z
          message: (LongPolling transport) polling: http://127.0.0.1:4437/testhub?id=FfLyq9LR2aS8iZYv_YA1dQ&_=1525036821075
        - level: Trace
          timestamp: 2018-04-29T21:20:21.080Z
          message: (LongPolling transport) polling: http://127.0.0.1:4437/testhub?id=FfLyq9LR2aS8iZYv_YA1dQ&_=1525036821080
        - level: Debug
          timestamp: 2018-04-29T21:20:21.080Z
          message: Sending handshake request.
        - level: Trace
          timestamp: 2018-04-29T21:20:21.080Z
          message: (LongPolling transport) sending data. String data of length 32. Content: '{"protocol":"json","version":1}'..
        - level: Trace
          timestamp: 2018-04-29T21:20:21.084Z
[...]

PR #2292, opened 2018-05-17
Refactor LongPollingTransport stop in TS client to align with C# client

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2163


clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -77,13 +71,13 @@ export class LongPollingTransport implements ITransport {
...
-            closeError = new HttpError(response.statusText, response.statusCode);
+            this.closeError = new HttpError(response.statusText, response.statusCode);
             this.running = false;
         } else {
             this.running = true;
         }
 
-        this.poll(this.url, pollOptions, closeError);
+        this.receiving = this.poll(this.url, pollOptions);
         return Promise.resolve();
```
JamesNK: What is this return for? Is it needed since this method has `async` on it?

anurse: I don't think it's needed, an `async` function returns a promise already.

clients/ts/signalr/tests/LongPollingTransport.test.ts
```diff
@@ -29,96 +29,95 @@ describe("LongPollingTransport", () => {
...
             .on("GET", async (r) => {
                 if (firstPoll) {
                     firstPoll = false;
                     return new HttpResponse(200);
                 } else {
-                    await deleteReceived.promise;
-                    // Force the shutdown timer to be registered by not returning inline
-                    await delay(10);
-                    pollCompleted.resolve();
+                    //
```
anurse: you forgot to

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -177,29 +164,37 @@ export class LongPollingTransport implements ITransport {
...
-        // Send a DELETE request to stop the poll
+        this.logger.log(LogLevel.Trace, `(LongPolling transport) Stopping polling.`);
+
+        // Tell receiving loop to stop, abort any current request, and then wait for it to finish
+        this.running = false;
+        this.pollAbort.abort();
+
         try {
-            this.running = false;
+            await this.receiving;
```
davidfowl: I recall having problem making this work well because transports were reusable. You should probably copy all of the state to locals before doing anything.

clients/ts/signalr/tests/LongPollingTransport.test.ts
```diff
@@ -29,96 +29,95 @@ describe("LongPollingTransport", () => {
...
                     return new HttpResponse(204);
                 }
             })
-            .on("DELETE", (r) => {
-                deleteReceived.resolve();
-                return new HttpResponse(202);
-            });
-        const transport = new LongPollingTransport(client, null, NullLogger.instance, false, 100);
+            .on("DELETE", (r) => new HttpResponse(202));
```
anurse: Should there be a `DELETE` request in this case? Maybe it's better to not expect it, since I don't think a `204` should result in a `DELETE` request (not 100% sure that the C# client behaves this way though, and if it sends the `DELETE` in the case of a `204`, the JS client should too)

anurse: Yeah, we only call `SendDeleteRequest` if the sender is trigger for stopping in C#: https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/LongPollingTransport.cs#L106

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -136,36 +130,29 @@ export class LongPollingTransport implements ITransport {
...
-            }
+            this.logger.log(LogLevel.Trace, "(LongPolling transport) Polling complete.");
 
-            // Fire our onclosed event
-            if (this.onclose) {
-                this.logger.log(LogLevel.Trace, `(LongPolling transport) Firing onclose event. Error: ${closeError || "<undefined>"}`);
-                this.onclose(closeError);
+            // We will reach here with pollAborted==false when the server returned a response causing
+            // the transport to close. If pollAborted==true then client initiated the stop
+            // and the stop method will raise the close event.
```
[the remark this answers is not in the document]
JamesNK: raiseOnClose shouldn't happen until after the delete request is completed. Could move the delete logic here as well but it feels a weird place to have it

clients/ts/signalr/tests/LongPollingTransport.test.ts
JamesNK: is copy pasta

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -177,29 +162,39 @@ export class LongPollingTransport implements ITransport {
...
+
+            // Raise close event here instead of in polling
+            // It needs to happen after the DELETE request is sent
+            this.raiseOnClose();
+        }
+    }
+
+    private raiseOnClose() {
+        if (this.onclose) {
+            this.logger.log(LogLevel.Trace, `(LongPolling transport) Firing onclose event. Error: ${this.closeError || "<undefined>"}`);
```
BrennanConroy: Graceful closes will log "Error: \<undefined\>"?

clients/ts/signalr/tests/LongPollingTransport.test.ts
```diff
@@ -29,96 +29,94 @@ describe("LongPollingTransport", () => {
...
             })
             .on("DELETE", (r) => new HttpResponse(202));
-        const transport = new LongPollingTransport(client, null, NullLogger.instance, false, 100);
+        const transport = new LongPollingTransport(client, null, NullLogger.instance, false);
 
         await transport.connect("http://example.com", TransferFormat.Text);
-        await transport.stop();
+        const stopPromise = transport.stop();
 
         // This should complete within the shutdown timeout
```
BrennanConroy: no more shutdown timeout

clients/ts/signalr/tests/LongPollingTransport.test.ts
```diff
@@ -29,96 +29,94 @@ describe("LongPollingTransport", () => {
...
             .on("GET", async (r) => {
                 if (firstPoll) {
                     firstPoll = false;
                     return new HttpResponse(200);
                 } else {
-                    await deleteReceived.promise;
-                    // Force the shutdown timer to be registered by not returning inline
-                    await delay(10);
-                    pollCompleted.resolve();
+                    // A 204 response will stop of the long polling transport
```
BrennanConroy: "will stop of"?

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -158,6 +158,12 @@ export class DefaultHttpClient extends HttpClient {
     /** @inheritDoc */
     public send(request: HttpRequest): Promise<HttpResponse> {
         return new Promise<HttpResponse>((resolve, reject) => {
+            // Check that abort was not signaled before calling send
+            if (request.abortSignal && request.abortSignal.aborted) {
+                reject(new AbortError());
```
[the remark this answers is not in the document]
JamesNK: It's like `OperationCanceledException`. It's just thrown to indicate an abort.

clients/ts/signalr/src/LongPollingTransport.ts
JamesNK: Yeah. That's what it currently logs.

JamesNK:

clients/ts/signalr/src/LongPollingTransport.ts
davidfowl: What happens if you don't await stop

JamesNK: A transport isn't reused when a HubConnection is started and stopped. And the HttpConnection state is immediately set to Disconnected when stop is called.

davidfowl: Ah! I forgot @anurse did that work

clients/ts/signalr/src/LongPollingTransport.ts
BrennanConroy: Could we change it to not have "Error: \<undefined\>" for clean close

clients/ts/signalr/tests/Utils.ts
```diff
@@ -38,3 +38,26 @@ export class PromiseSource<T = void> implements Promise<T> {
         return this.promise.catch(onrejected);
     }
 }
+
+export class SyncPoint {
```
davidfowl: Fancy!

anurse: SyncPoint continues it's journey towards world domination.

PR #2300, opened 2018-05-17
Enable analyzers

pakrym:


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -46,7 +46,10 @@ public class HubConnectionHandler<THub> : ConnectionHandler where THub : Hub
                                     IOptions<HubOptions<THub>> hubOptions,
                                     ILoggerFactory loggerFactory,
                                     IUserIdProvider userIdProvider,
-                                    HubDispatcher<THub> dispatcher)
+#pragma warning disable PUB0001 // Pubternal type in public API
+                                    HubDispatcher<THub> dispatcher
```
davidfowl: Oh snap. Nice find!

anurse: I'm glad it's only the one place :)

PR #2309, opened 2018-05-17
Dygray/ping timer

### LINKED ISSUE #2175: Send client to server pings from C# client
Depending on how connections die it can take an extremely long time for OnDisconnectedAsync to fire on the server. See the issue below for examples: 

https://github.com/aspnet/SignalR/issues/1290

The implementation would be the same as the client, we would abort the connection after some configured interval of inactivity.

### LINKED ISSUE #2307: Send client to server pings from TS client
Just like #2175, but for TypeScript :).

DylanDmitri:
Fixes #2307 
Fixes #2175 

```HubConnection``` now pings server every <15> seconds.

Extracted logic for ```TimeoutTimer``` into a class, to re-use the code for the ```PingTimer``` which makes the pings.

Changed signature on ```SendMessage``` to take more generic parameters, needed some logging changes.

### DIFF
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -6,8 +6,8 @@ import { IConnection } from "./IConnection";
 import { CancelInvocationMessage, CompletionMessage, IHubProtocol, InvocationMessage, MessageType, StreamInvocationMessage, StreamItemMessage } from "./IHubProtocol";
 import { ILogger, LogLevel } from "./ILogger";
 import { IStreamResult } from "./Stream";
-import { Arg, Subject } from "./Utils";
 import { TextMessageFormat } from "./TextMessageFormat";
+import { Arg, Subject } from "./Utils";
 
 const DEFAULT_TIMEOUT_IN_MS: number = 30 * 1000;
 const DEFAULT_PING_INTERVAL_IN_MS: number = 15 * 1000;
@@ -74,7 +74,7 @@ export class HubConnection {
         this.id = 0;
         this.connectionState = HubConnectionState.Disconnected;
 
-        this.cachedPingMessage = this.protocol.writeMessage({ type: MessageType.Ping, });
+        this.cachedPingMessage = this.protocol.writeMessage({ type: MessageType.Ping });
     }
 
     /** Indicates the state of the {@link HubConnection} to the server. */
@@ -379,7 +379,7 @@ export class HubConnection {
 
         return remainingData;
     }
-    
+
     private resetPingInterval() {
         this.cleanupPingTimer();
         this.pingServerHandle = setTimeout(() => this.connectionSend(this.cachedPingMessage), this.pingIntervalInMilliseconds);
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -41,6 +41,11 @@ export class HubConnection {
      */
     public serverTimeoutInMilliseconds: number;
 
+    /** Default interval at which to ping the server.
+     *
+     * The default value is 15,000 milliseconds (15 seconds).
+     * Allows the server to detect hard disconnects (like when a client unplugs their computer).
+     */
     public pingIntervalInMilliseconds: number;
 
     /** @internal */
@@ -100,7 +105,7 @@ export class HubConnection {
 
         this.logger.log(LogLevel.Debug, "Sending handshake request.");
 
-        await this.connectionSend(this.handshakeProtocol.writeHandshakeRequest(handshakeRequest));
+        await this.sendMessage(this.handshakeProtocol.writeHandshakeRequest(handshakeRequest));
 
         this.logger.log(LogLevel.Information, `Using HubProtocol '${this.protocol.name}'.`);
 
@@ -140,7 +145,7 @@ export class HubConnection {
 
             delete this.callbacks[invocationDescriptor.invocationId];
 
-            return this.connectionSend(cancelMessage);
+            return this.sendMessage(cancelMessage);
         });
 
         this.callbacks[invocationDescriptor.invocationId] = (invocationEvent: CompletionMessage | StreamItemMessage, error?: Error) => {
@@ -162,7 +167,7 @@ export class HubConnection {
 
         const message = this.protocol.writeMessage(invocationDescriptor);
 
-        this.connectionSend(message)
+        this.sendMessage(message)
             .catch((e) => {
                 subject.error(e);
                 delete this.callbacks[invocationDescriptor.invocationId];
@@ -171,7 +176,7 @@ export class HubConnection {
         return subject;
     }
 
-    private connectionSend(message: any) {
+    private sendMessage(message: any) {
         this.resetPingInterval();
         return this.connection.send(message);
     }
@@ -190,7 +195,7 @@ export class HubConnection {
 
         const message = this.protocol.writeMessage(invocationDescriptor);
 
-        return this.connectionSend(message);
+        return this.sendMessage(message);
     }
 
     /** Invokes a hub method on the server using the specified name and arguments.
@@ -227,7 +232,7 @@ export class HubConnection {
 
             const message = this.protocol.writeMessage(invocationDescriptor);
 
-            this.connectionSend(message)
+            this.sendMessage(message)
                 .catch((e) => {
                     reject(e);
                     delete this.callbacks[invocationDescriptor.invocationId];
@@ -382,7 +387,7 @@ export class HubConnection {
 
     private resetPingInterval() {
         this.cleanupPingTimer();
-        this.pingServerHandle = setTimeout(() => this.connectionSend(this.cachedPingMessage), this.pingIntervalInMilliseconds);
+        this.pingServerHandle = setTimeout(() => this.sendMessage(this.cachedPingMessage), this.pingIntervalInMilliseconds);
     }
 
     private resetTimeoutPeriod() {
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -7,19 +7,20 @@ import { CancelInvocationMessage, CompletionMessage, IHubProtocol, InvocationMes
 import { ILogger, LogLevel } from "./ILogger";
 import { IStreamResult } from "./Stream";
 import { Arg, Subject } from "./Utils";
+import { TextMessageFormat } from "./TextMessageFormat";
 
 const DEFAULT_TIMEOUT_IN_MS: number = 30 * 1000;
+const DEFAULT_PING_INTERVAL_IN_MS: number = 15 * 1000;
 
 /** Describes the current state of the {@link HubConnection} to the server. */
 export enum HubConnectionState {
-    /** The hub connection is disconnected. */
     Disconnected,
-    /** The hub connection is connected. */
     Connected,
 }
 
 /** Represents a connection to a SignalR Hub. */
 export class HubConnection {
+    private readonly cachedPingMessage: string | ArrayBuffer;
     private readonly connection: IConnection;
     private readonly logger: ILogger;
     private protocol: IHubProtocol;
@@ -29,6 +30,7 @@ export class HubConnection {
     private id: number;
     private closedCallbacks: Array<(error?: Error) => void>;
     private timeoutHandle: NodeJS.Timer;
+    private pingServerHandle: NodeJS.Timer;
     private receivedHandshakeResponse: boolean;
     private connectionState: HubConnectionState;
 
@@ -38,6 +40,8 @@ export class HubConnection {
      * The default timeout value is 30,000 milliseconds (30 seconds).
      */
     public serverTimeoutInMilliseconds: number;
+
+    public pingIntervalInMilliseconds: number;
 
     /** @internal */
     // Using a public static factory method means we can have a private constructor and an _internal_
@@ -54,6 +58,7 @@ export class HubConnection {
         Arg.isRequired(protocol, "protocol");
 
         this.serverTimeoutInMilliseconds = DEFAULT_TIMEOUT_IN_MS;
+        this.pingIntervalInMilliseconds = DEFAULT_PING_INTERVAL_IN_MS;
 
         this.logger = logger;
         this.protocol = protocol;
@@ -68,6 +73,8 @@ export class HubConnection {
         this.closedCallbacks = [];
         this.id = 0;
         this.connectionState = HubConnectionState.Disconnected;
+
+        this.cachedPingMessage = this.protocol.writeMessage({ type: MessageType.Ping, });
     }
 
     /** Indicates the state of the {@link HubConnection} to the server. */
@@ -93,13 +100,14 @@ export class HubConnection {
 
         this.logger.log(LogLevel.Debug, "Sending handshake request.");
 
-        await this.connection.send(this.handshakeProtocol.writeHandshakeRequest(handshakeRequest));
+        await this.connectionSend(this.handshakeProtocol.writeHandshakeRequest(handshakeRequest));
 
         this.logger.log(LogLevel.Information, `Using HubProtocol '${this.protocol.name}'.`);
 
         // defensively cleanup timeout in case we receive a message from the server before we finish start
         this.cleanupTimeout();
-        this.configureTimeout();
+        this.resetTimeoutPeriod();
+        this.resetPingInterval();
 
         this.connectionState = HubConnectionState.Connected;
     }
@@ -112,6 +120,7 @@ export class HubConnection {
         this.logger.log(LogLevel.Debug, "Stopping HubConnection.");
 
         this.cleanupTimeout();
+        this.cleanupPingTimer();
         return this.connection.stop();
     }
 
@@ -131,7 +140,7 @@ export class HubConnection {
 
             delete this.callbacks[invocationDescriptor.invocationId];
 
-            return this.connection.send(cancelMessage);
+            return this.connectionSend(cancelMessage);
         });
 
         this.callbacks[invocationDescriptor.invocationId] = (invocationEvent: CompletionMessage | StreamItemMessage, error?: Error) => {
@@ -153,13 +162,18 @@ export class HubConnection {
 
         const message = this.protocol.writeMessage(invocationDescriptor);
 
-        this.connection.send(message)
+        this.connectionSend(message)
             .catch((e) => {
                 subject.error(e);
                 delete this.callbacks[invocationDescriptor.invocationId];
             });
 
         return subject;
+    }
+
+    private connectionSend(message: any) {
+        this.resetPingInterval();
+        return this.connection.send(message);
     }
 
     /** Invokes a hub method on the server using the specified name and arguments. Does not wait for a response from the receiver.
@@ -176,7 +190,7 @@ export class HubConnection {
 
         const message = this.protocol.writeMessage(invocationDescriptor);
 
-        return this.connection.send(message);
+        return this.connectionSend(message);
     }
 
     /** Invokes a hub method on the server using the specified name and arguments.
@@ -213,7 +227,7 @@ export class HubConnection {
 
             const message = this.protocol.writeMessage(invocationDescriptor);
 
-            this.connection.send(message)
+            this.connectionSend(message)
                 .catch((e) => {
                     reject(e);
                     delete this.callbacks[invocationDescriptor.invocationId];
@@ -297,6 +311,7 @@ export class HubConnection {
 
     private processIncomingData(data: any) {
         this.cleanupTimeout();
+        this.cleanupPingTimer();
 
         if (!this.receivedHandshakeResponse) {
             data = this.processHandshakeResponse(data);
@@ -337,7 +352,7 @@ export class HubConnection {
             }
         }
 
-        this.configureTimeout();
+        this.resetTimeoutPeriod();
     }
 
     private processHandshakeResponse(data: any): any {
@@ -364,8 +379,13 @@ export class HubConnection {
 
         return remainingData;
     }
+    
+    private resetPingInterval() {
+        this.cleanupPingTimer();
+        this.pingServerHandle = setTimeout(() => this.connectionSend(this.cachedPingMessage), this.pingIntervalInMilliseconds);
+    }
 
-    private configureTimeout() {
+    private resetTimeoutPeriod() {
         if (!this.connection.features || !this.connection.features.inherentKeepAlive) {
             // Set the timeout timer
             this.timeoutHandle = setTimeout(() => this.serverTimeout(), this.serverTimeoutInMilliseconds);
@@ -406,8 +426,15 @@ export class HubConnection {
             });
 
         this.cleanupTimeout();
+        this.cleanupPingTimer();
 
         this.closedCallbacks.forEach((c) => c.apply(this, [error]));
+    }
+
+    private cleanupPingTimer(): void {
+        if (this.pingServerHandle) {
+            clearTimeout(this.pingServerHandle);
+        }
     }
 
     private cleanupTimeout(): void {
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -316,7 +316,6 @@ export class HubConnection {
 
     private processIncomingData(data: any) {
         this.cleanupTimeout();
-        this.cleanupPingTimer();
 
         if (!this.receivedHandshakeResponse) {
             data = this.processHandshakeResponse(data);
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -14,7 +14,9 @@ const DEFAULT_PING_INTERVAL_IN_MS: number = 15 * 1000;
 
 /** Describes the current state of the {@link HubConnection} to the server. */
 export enum HubConnectionState {
+    /** The hub connection is disconnected. */
     Disconnected,
+    /** The hub connection is connected. */
     Connected,
 }
 
--- a/clients/ts/signalr/tests/HubConnection.test.ts
+++ b/clients/ts/signalr/tests/HubConnection.test.ts
@@ -65,27 +65,6 @@ describe("HubConnection", () => {
                 await hubConnection.stop();
             }
         });
-
-        it("doesn't ping while sending other messages", async () => {
-            const connection = new TestConnection();
-            const hubConnection = createHubConnection(connection);
-
-            hubConnection.pingIntervalInMilliseconds = 150;
-
-            try {
-                await hubConnection.start();
-
-                for (let i = 0; i < 30; i++) {
-                    await hubConnection.send("some message");
-                    await delay(10);
-                }
-
-                const numPings = connection.sentData.filter((s) => JSON.parse(s).type === MessageType.Ping).length;
-                expect(numPings).toBe(0);
-            } finally {
-                await hubConnection.stop();
-            }
-        });
     });
 
     describe("stop", () => {
--- a/clients/ts/signalr/tests/HubConnection.test.ts
+++ b/clients/ts/signalr/tests/HubConnection.test.ts
@@ -76,7 +76,7 @@ describe("HubConnection", () => {
                 await hubConnection.start();
 
                 for (let i = 0; i < 30; i++) {
-                    hubConnection.send("some message");
+                    await hubConnection.send("some message");
                     await delay(10);
                 }
 
--- a/clients/ts/signalr/tests/HubConnection.test.ts
+++ b/clients/ts/signalr/tests/HubConnection.test.ts
@@ -53,11 +53,11 @@ describe("HubConnection", () => {
             const connection = new TestConnection();
             const hubConnection = createHubConnection(connection);
 
-            hubConnection.pingIntervalInMilliseconds = 150;
+            hubConnection.pingIntervalInMilliseconds = 5;
 
             try {
                 await hubConnection.start();
-                await delay(322);
+                await delay(32);
 
                 const numPings = connection.sentData.filter((s) => JSON.parse(s).type === MessageType.Ping).length;
                 expect(numPings).toBeGreaterThanOrEqual(2);
--- a/clients/ts/signalr/tests/HubConnection.test.ts
+++ b/clients/ts/signalr/tests/HubConnection.test.ts
@@ -52,15 +52,15 @@ describe("HubConnection", () => {
         it("automatically sends multiple pings", async () => {
             const connection = new TestConnection();
             const hubConnection = createHubConnection(connection);
-    
+
             hubConnection.pingIntervalInMilliseconds = 60;
-    
+
             try {
                 await hubConnection.start();
                 await delay(150);
-    
-                var numPings = connection.sentData.filter(s => JSON.parse(s).type==MessageType.Ping).length;
-                expect(numPings).toBeGreaterThanOrEqual(2); 
+
+                const numPings = connection.sentData.filter((s) => JSON.parse(s).type === MessageType.Ping).length;
+                expect(numPings).toBeGreaterThanOrEqual(2);
             } finally {
                 await hubConnection.stop();
             }
@@ -75,12 +75,12 @@ describe("HubConnection", () => {
             try {
                 await hubConnection.start();
 
-                for (var i=0; i<15; i++) {
+                for (let i = 0; i < 15; i++) {
                     hubConnection.send("some message");
                     await delay(10);
                 }
 
-                var numPings = connection.sentData.filter(s => JSON.parse(s).type==MessageType.Ping).length;
+                const numPings = connection.sentData.filter((s) => JSON.parse(s).type === MessageType.Ping).length;
                 expect(numPings).toBe(0);
             } finally {
                 await hubConnection.stop();
@@ -910,7 +910,7 @@ describe("HubConnection", () => {
                 hubConnection.onclose((e) => state = hubConnection.state);
                 // Typically this would be called by the transport
                 connection.onclose();
-                
+
                 expect(state).toBe(HubConnectionState.Disconnected);
             } finally {
                 hubConnection.stop();
--- a/clients/ts/signalr/tests/HubConnection.test.ts
+++ b/clients/ts/signalr/tests/HubConnection.test.ts
@@ -53,11 +53,11 @@ describe("HubConnection", () => {
             const connection = new TestConnection();
             const hubConnection = createHubConnection(connection);
 
-            hubConnection.pingIntervalInMilliseconds = 60;
+            hubConnection.pingIntervalInMilliseconds = 150;
 
             try {
                 await hubConnection.start();
-                await delay(150);
+                await delay(322);
 
                 const numPings = connection.sentData.filter((s) => JSON.parse(s).type === MessageType.Ping).length;
                 expect(numPings).toBeGreaterThanOrEqual(2);
@@ -70,12 +70,12 @@ describe("HubConnection", () => {
             const connection = new TestConnection();
             const hubConnection = createHubConnection(connection);
 
-            hubConnection.pingIntervalInMilliseconds = 60;
+            hubConnection.pingIntervalInMilliseconds = 150;
 
             try {
                 await hubConnection.start();
 
-                for (let i = 0; i < 15; i++) {
+                for (let i = 0; i < 30; i++) {
                     hubConnection.send("some message");
                     await delay(10);
                 }
--- a/clients/ts/signalr/tests/HubConnection.test.ts
+++ b/clients/ts/signalr/tests/HubConnection.test.ts
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
         });
     });
 
+    describe("ping", () => {
+        it("automatically sends multiple pings", async () => {
+            const connection = new TestConnection();
+            const hubConnection = createHubConnection(connection);
+    
+            hubConnection.pingIntervalInMilliseconds = 60;
+    
+            try {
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs [hunk 1]
```diff
@@ -105,6 +125,60 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
 
             _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             _logger = _loggerFactory.CreateLogger<HubConnection>();
+
+            // closes this class, if it goes `15s` without hearing from server
+            // starts in receive loop
+            _timeoutTimer = new PeriodicConnectionTimer(
```
davidfowl: Any reason we aren't using timer awaitable? I dislike the fact that we have these callbacks that are in the middle of the constructor.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs
```diff
@@ -174,6 +174,14 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _removingHandlers =
                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(58, "RemovingHandlers"), "Removing handlers for client method '{MethodName}'.");
 
+            private static readonly Action<ILogger, string, Exception> _sendingMessageGeneric =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(59, "SendingMessageGeneric"), "Sending {MessageType} message.");
+
+            private static readonly Action<ILogger, string, Exception> _messageSentGeneric =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(6, "MessageSentGeneric"), "Sending {MessageType} message completed.");
+
+
```
mikaelm12: Remove these extra lines

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -55,9 +58,26 @@ public partial class HubConnection
...
+        /// Sending any message resets the timer to the start of the interval.
+        /// </remarks>
+        public TimeSpan PingInterval
+        {
+            get { return _pingTimer.Interval; }
+            set { _pingTimer.Interval = value; }
+        }
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
 
+
```
mikaelm12: remove this line

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -105,6 +125,60 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
...
+                    }
+
+                    if (!hub._connectionLock.Wait(0))
+                    {
+                        // if we can't get the lock, someone else is sending, so we don't need to ping
+                        return;
+                    }
+
+                    // if we got here, we successfully have the lock
+
```
mikaelm12: Remove one of these lines

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1152,5 +1191,39 @@ IReadOnlyList<Type> IInvocationBinder.GetParameterTypes(string methodName)
...
+            public PeriodicConnectionTimer(TimeSpan interval, object state, TimerCallback onTimeout)
+            {
+                Interval = interval;
+                _onTimeout = onTimeout;
+                _state = state;
+            }
+
+            public void Start()
+            {
+                _timer = new Timer(
```
mikaelm12: We generally don't use newlines in ctors

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1152,5 +1191,39 @@ IReadOnlyList<Type> IInvocationBinder.GetParameterTypes(string methodName)
...
+                    _onTimeout,
+                    _state,
+                    dueTime: Interval,
```
mikaelm12: And we generally don't use named args unless for some reasons we're passing in literal values.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +565,30 @@ public class Protocol
...
                 }
             }
+
+            [Fact]
+            public async Task ClientPingsOnInterval()
+            {
+                var connection = new TestConnection();
+                var hubConnection = CreateHubConnection(connection);
+
+                hubConnection.PingInterval = TimeSpan.FromMilliseconds(70);
```
mikaelm12: You should comment why you chose the delay values that you did

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +565,30 @@ public class Protocol
...
+
+                try
+                {
+                    await hubConnection.StartAsync().OrTimeout();
+                    await Task.Delay(100);
+
+                    var received = await connection.ReadSentTextMessageAsync().OrTimeout();
+                    Assert.Equal("{\"type\":6}", received);
+
```
mikaelm12: Remove newline

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -105,6 +125,60 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
...
+
+                    ((ConnectionState)state).CloseException = new TimeoutException(
+                        $"Server timeout ({_timeoutTimer.Interval.TotalMilliseconds:0.00}ms) elapsed without receiving a message from the server.");
+                    ((ConnectionState)state).Connection.Transport.Input.CancelPendingRead();
+                });
+
+            _pingTimer = new PeriodicConnectionTimer(
+                DefaultPingInterval,
+                state: this,
+                async (state) =>
```
mikaelm12: I don't think we're going to want to have this lambda in the constructor.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -105,6 +125,60 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
...
+                            return;
+                        }
+
+                        await hub.SendHubMessage(PingMessage.Instance);
+                    }
+                    finally
+                    {
+                        hub.ReleaseConnectionLock();
+                    }
+
```
mikaelm12: remove newline

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -105,6 +125,60 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
...
+                DefaultServerTimeout,
+                _connectionState,
+                (state) =>
+                {
+                    if (Debugger.IsAttached)
+                    {
+                        return;
+                    }
+
+                    ((ConnectionState)state).CloseException = new TimeoutException(
```
mikaelm12: Cast state to ConnectionState once and reuse 😄

anurse:
Had a chat with @DylanDmitri and we're going to refactor this a bit to use `TimerAwaitable` and a single timer that checks both ping and server timeout, so stand-by for updates :).

anurse:
Also we'll need to rebase on `dev` when the new iteration is ready so that we don't get all the test failures from the logging changes :)

DylanDmitri:
Refactored to use a single timer.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +565,62 @@ public class Protocol
...
+                    await Task.Delay(220);
+
+                    var firstPing = await connection.ReadSentTextMessageAsync().OrTimeout(TimeSpan.FromMilliseconds(5));
+                    Assert.Equal("{\"type\":6}", firstPing);
+
+                    var secondPing = await connection.ReadSentTextMessageAsync().OrTimeout(TimeSpan.FromMilliseconds(5));
+                    Assert.Equal("{\"type\":6}", secondPing);
+                    
+                    // with a 200ms delay, we should have enough time for 2 pings but not 3
+                    await Assert.ThrowsAsync<TimeoutException>(
```
anurse: Ah, this is what's causing the test failure. The problem is that `Task.Delay` doesn't guarantee that you'll run after exactly 220ms, it just means that you'll go back on the scheduler queue after 220ms. On slower machines (like our CI machines), when a lot of other tests are running in parallel, it might take longer than 220ms, so you may end up with more than 2 pings.

I don't think we need this assert anyway, the goal is to test that we get at least 2 pings, we're OK with getting more.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1152,5 +1210,23 @@ IReadOnlyList<Type> IInvocationBinder.GetParameterTypes(string methodName)
                 throw new InvalidOperationException($"There are no callbacks registered for the method '{methodName}'");
             }
         }
+
+        private class PeriodicConnectionTimer
```
anurse: I think this class may be a little overkill 😄 . The only things we need to store are the interval (which we already have on HubConnection) and the next activation (which we could just put straight on HubConnection), and the `Reset` method is pretty simple, so we can just inline it wherever it's used.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1152,5 +1210,23 @@ IReadOnlyList<Type> IInvocationBinder.GetParameterTypes(string methodName)
...
+        {
+            public TimeSpan Interval { get; set; }
+            public DateTime NextActivation { get; set; }
```
anurse: Actually, instead of a `DateTime`, we should store a `long` and use the value from [`Stopwatch.GetTimestamp()`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.gettimestamp?view=netstandard-2.0). 

`GetTimestamp()` is much faster than `DateTime.Now` since it just reads the current value of the hardware timer, which stores the number of "ticks" that have elapsed since the machine started. Since the interval is in seconds though, you also need to look at [`Stopwatch.Frequency`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.frequency?view=netstandard-2.0) to figure out how many "ticks" there are in a second (since it differs from machine to machine).

`Stopwatch.GetTimestamp` is generally faster when doing *relative* timing like this (where you don't care about the actual Date, just the time elapsed since some other event.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -703,7 +736,10 @@ private void CheckDisposed()
 
             Log.ReceiveLoopStarting(_logger);
 
-            var timeoutTimer = StartTimeoutTimer(connectionState);
+            // performs periodic tasks -- here sending pings and checking timeout
+            // disposed with `timer.Stop()` in the finally block below
+            TimerAwaitable timer = new TimerAwaitable(TickRate, TickRate);
```
davidfowl: `var timer`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -814,61 +852,81 @@ private void CheckDisposed()
...
+                    return;
+                }
+
+                await SendHubMessage(PingMessage.Instance);
             }
-            else
+            finally
             {
-                Log.NotUsingServerTimeout(_logger);
+                ReleaseConnectionLock();
```
BrennanConroy: There wont be a matching "Acquired connection lock" message currently

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +565,62 @@ public class Protocol
...
+
+                hubConnection.TickRate = TimeSpan.FromMilliseconds(30);
+                hubConnection.PingInterval = TimeSpan.FromMilliseconds(80);
+
+                try
+                {
+                    await hubConnection.StartAsync().OrTimeout();
+                    await Task.Delay(120);
+
+                    var received = await connection.ReadSentTextMessageAsync().OrTimeout(TimeSpan.FromMilliseconds(100));
```
BrennanConroy: nit: `var sent`

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +565,62 @@ public class Protocol
...
+                var connection = new TestConnection();
+                var hubConnection = CreateHubConnection(connection);
+
+                hubConnection.TickRate = TimeSpan.FromMilliseconds(30);
+                hubConnection.PingInterval = TimeSpan.FromMilliseconds(80);
+
+                try
+                {
+                    await hubConnection.StartAsync().OrTimeout();
+                    await Task.Delay(120);
```
BrennanConroy: I don't think we should have this

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
BrennanConroy: Make it `OrTimeout(TimeSpan.FromSeconds(1))`

It's hard to write tests for small wait durations, so we try to err on the conservative side. For this test we know we'll get a ping before 1 second, if we don't, something went wrong and we'll timeout

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +565,62 @@ public class Protocol
...
+                }
+                finally
+                {
+                    await hubConnection.DisposeAsync().OrTimeout();
+                    await connection.DisposeAsync().OrTimeout();
+                }
+            }
+
+            [Fact]
+            public async Task ClientPingsMultipleTimes()
```
BrennanConroy: These 2 tests can be merged

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -44,20 +45,41 @@ public partial class HubConnection
...
+        private readonly PeriodicConnectionTimer _timeoutTimer;
+        private readonly PeriodicConnectionTimer _pingTimer;
         private bool _disposed;
 
         // Transient state to a connection
         private ConnectionState _connectionState;
 
         public event Func<Exception, Task> Closed;
 
+        internal TimeSpan TickRate { get; set; } = TimeSpan.FromSeconds(1);
```
BrennanConroy: // Internal for testing purposes

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
anurse: Yeah, that's a good point, the `ReadSentTextMessageAsync` call will block until the ping message arrives, so there's no need for the `Task.Delay` :)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Ooh, yeah good point :). Because you're not calling `WaitConnectionLock` and instead calling `Wait` directly on the semaphore we don't get the log message.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
DylanDmitri: |                Method |      Mean (ns) |     Error (ns) |    StdDev (ns) |         Op/s | Allocated |
|---------------------- |----------:|----------:|----------:|-------------:|----------:|
|           DatetimeNow | 202.06 | 0.4520 | 0.3529 |  4,949,002.8 |       0 B |
|        DatetimeNowUTC |  21.16 | 0.0301 | 0.0235 | 47,248,226.7 |       0 B |
| StopwatchGetTimestamp |  18.82 | 0.4083 | 0.4193 | 53,147,234.1 |       0 B |

Switching to ```DateTime.UtcNow()``` because it's almost as fast while avoiding tick frequency math.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -44,18 +45,31 @@ public partial class HubConnection
...
 
         public event Func<Exception, Task> Closed;
 
+        // internal for testing purposes
+        internal TimeSpan TickRate { get; set; } = TimeSpan.FromSeconds(1);
+
+        /// <summary>
+        /// Gets or sets the server timeout interval for the connection. 
+        /// </summary>
+        public TimeSpan ServerTimeoutInterval { get; set; } = TimeSpan.FromSeconds(15);
```
BrennanConroy: Don't think we can change this name, it would be a breaking change and we generally don't do those except during major releases

BrennanConroy: It did, look at the failing windows build

davidfowl: Oh wait, we turned off appveyor 😆

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -39,6 +41,8 @@ export class HubConnection {
      */
     public serverTimeoutInMilliseconds: number;
 
+    public pingIntervalInMilliseconds: number;
```
anurse: Can you add JSDoc comments to this, similar to the XML docs for C#? Use the comments above `serverTimeoutInMilliseconds` as a guide. The first sentence is a brief summary, followed by a blank line, followed by a more detailed explanation.

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -162,6 +171,11 @@ export class HubConnection {
         return subject;
     }
 
+    private connectionSend(message: any) {
```
anurse: Let's call this `sendMessage`, it's a little clearer than `connectionSend` IMO

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs
```diff
@@ -174,6 +174,12 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _removingHandlers =
                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(58, "RemovingHandlers"), "Removing handlers for client method '{MethodName}'.");
 
+            private static readonly Action<ILogger, string, Exception> _sendingMessageGeneric =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(59, "SendingMessageGeneric"), "Sending {MessageType} message.");
```
anurse: @davidfowl Have we decided what constitutes a "breaking change" with Logging messages? In theory, we could make `SendingMessage` and `SendingMessageGeneric` the same EventID, but with different delegates, one that takes an Invocation ID and one that does not.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -44,18 +45,31 @@ public partial class HubConnection
...
         /// <summary>
-        /// Gets or sets the server timeout interval for the connection. Changes to this value
-        /// will not be applied until the Keep Alive timer is next reset.
+        /// Gets or sets the interval at which the client sends ping messages.
         /// </summary>
-        public TimeSpan ServerTimeout { get; set; } = DefaultServerTimeout;
+        /// <remarks>
+        /// Sending any message resets the timer to the start of the interval.
+        /// </remarks>
+        public TimeSpan SendPingInterval { get; set; } = TimeSpan.FromSeconds(15);
```
anurse: Let's be consistent with the JavaScript naming and use `PingInterval` here.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -465,7 +481,7 @@ private Task CloseAsync(ConnectionContext connection)
             }
         }
 
-        private async Task SendHubMessage(HubInvocationMessage hubMessage, CancellationToken cancellationToken = default)
+        internal async Task SendHubMessage(HubMessage hubMessage, CancellationToken cancellationToken = default)
```
anurse: This should be able to stay `private` I think.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -814,61 +838,82 @@ private void CheckDisposed()
...
+            if (_disposed || !_connectionLock.Wait(0))
+            {
+                // if we're disposed or can't get the lock, we don't need to ping
+                return;
+            }
 
-            // We use '!== true' because it could be null, which we treat as false.
-            if (connectionState.Connection.Features.Get<IConnectionInherentKeepAliveFeature>()?.HasInherentKeepAlive != true)
+            // if we got here, we successfully have the lock
+            // TODO: log lock larceny
```
anurse: Don't forget to add the logging here :)

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +565,65 @@ public class Protocol
...
+
+                try
+                {
+                    await hubConnection.StartAsync().OrTimeout();
+
+                    var firstPing = await connection.ReadSentTextMessageAsync().OrTimeout(TimeSpan.FromMilliseconds(200));
+                    Assert.Equal("{\"type\":6}", firstPing);
+
+                    //var secondPing = await connection.ReadSentTextMessageAsync().OrTimeout(TimeSpan.FromMilliseconds(120));
+                    //Assert.Equal("{\"type\":6}", secondPing);
```
anurse: Does this still need to be commented out?

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -146,6 +146,28 @@ public Task ReceiveBytesAsync(byte[] bytes)
             }
         }
 
+        public async Task<IList<string>> ReadAllSentMessagesAsync()
+        {
+            if (!_disposed.Task.IsCompleted)
```
anurse: You can actually just write this as `Disposed.IsCompleted` because we already define the `Disposed` property above as `_disposed.Task` :)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
DylanDmitri: Changed it intending to use it in the unit tests, because it's where ```PingInterval``` is actually reset. Ending up using the less generic ```SendAsync``` to activate the reset, but forgot to switch it back.

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -6,20 +6,21 @@ import { IConnection } from "./IConnection";
...
 import { IStreamResult } from "./Stream";
+import { TextMessageFormat } from "./TextMessageFormat";
 import { Arg, Subject } from "./Utils";
 
 const DEFAULT_TIMEOUT_IN_MS: number = 30 * 1000;
+const DEFAULT_PING_INTERVAL_IN_MS: number = 15 * 1000;
 
 /** Describes the current state of the {@link HubConnection} to the server. */
 export enum HubConnectionState {
-    /** The hub connection is disconnected. */
```
JamesNK: How come these JSDoc comments are removed?

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
         });
     });
 
+    describe("ping", () => {
+        it("automatically sends multiple pings", async () => {
+            const connection = new TestConnection();
+            const hubConnection = createHubConnection(connection);
+
+            hubConnection.pingIntervalInMilliseconds = 60;
```
JamesNK: Increase to 150

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
...
+
+            try {
+                await hubConnection.start();
+                await delay(150);
```
JamesNK: Increase to 300

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
...
+            } finally {
+                await hubConnection.stop();
+            }
+        });
+
+        it("doesn't ping while sending other messages", async () => {
+            const connection = new TestConnection();
+            const hubConnection = createHubConnection(connection);
+
+            hubConnection.pingIntervalInMilliseconds = 60;
```
JamesNK: Increase to 150

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
...
+
+            try {
+                await hubConnection.start();
+
+                for (let i = 0; i < 15; i++) {
```
JamesNK: 15 -> 30

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -44,18 +45,31 @@ public partial class HubConnection
...
+        /// Gets or sets the server timeout interval for the connection. 
+        /// </summary>
+        public TimeSpan ServerTimeout { get; set; } = TimeSpan.FromSeconds(15);
+
         /// <summary>
-        /// Gets or sets the server timeout interval for the connection. Changes to this value
-        /// will not be applied until the Keep Alive timer is next reset.
+        /// Gets or sets the interval at which the client sends ping messages.
         /// </summary>
-        public TimeSpan ServerTimeout { get; set; } = DefaultServerTimeout;
```
JamesNK: Why no longer use `DefaultServerTimeout`?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
JamesNK: Why not use `DefaultPingInterval`?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -476,6 +492,9 @@ private Task CloseAsync(ConnectionContext connection)
             // REVIEW: If a token is passed in and is canceled during FlushAsync it seems to break .Complete()...
             await _connectionState.Connection.Transport.Output.FlushAsync();
 
+            // we've sent a message, so don't ping for a while
```
JamesNK: nit: In SignalR we start comments with capitals. Don't ask me why, it was the standard when I started 😄

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -703,7 +722,10 @@ private void CheckDisposed()
 
             Log.ReceiveLoopStarting(_logger);
 
-            var timeoutTimer = StartTimeoutTimer(connectionState);
+            // performs periodic tasks -- here sending pings and checking timeout
```
JamesNK: nit: Capital letters in comments. Update all other comments in PR.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: I'm not even sure we're that consistent :)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -814,61 +838,82 @@ private void CheckDisposed()
...
+            // get them roughly synced
+            timer.Start();
 
-            try
-            {
-                Log.InvokingClosedEventHandler(_logger);
-                await closed.Invoke(closeException);
-            }
-            catch (Exception ex)
+            _nextActivationSendPing = DateTime.UtcNow + SendPingInterval;
```
BrennanConroy: nit: Use the same `UtcNow` value

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -297,6 +316,7 @@ export class HubConnection {
 
     private processIncomingData(data: any) {
         this.cleanupTimeout();
+        this.cleanupPingTimer();
```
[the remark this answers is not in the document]
DylanDmitri: oops yeah

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
DylanDmitri: I wrote a bad comment, the lines intend to initialize the timers rather than sync them. As long as they're within a couple seconds it's okay. Changing the comment.

clients/ts/signalr/src/HubConnection.ts
DylanDmitri: The comments seemed redundant, but I guess I don't know how/where users are viewing these. Will add the comments back.

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
...
+            const connection = new TestConnection();
+            const hubConnection = createHubConnection(connection);
+
+            hubConnection.pingIntervalInMilliseconds = 150;
+
+            try {
+                await hubConnection.start();
+
+                for (let i = 0; i < 30; i++) {
+                    hubConnection.send("some message");
```
anurse: This should be `await`ed. Because of the `delay` it should be fine, but it's good hygiene to make sure we `await` it.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
anurse: Don't forget to swap this back to `private` ;)

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -44,18 +46,37 @@ public partial class HubConnection
...
+        /// </remarks>
         public TimeSpan ServerTimeout { get; set; } = DefaultServerTimeout;
+
+        /// <summary>
+        /// Gets or sets the interval at which the client sends ping messages.
+        /// </summary>
+        /// <remarks>
+        /// Sending any message resets the timer to the start of the interval.
+        /// </remarks>
+        public TimeSpan SendPingInterval { get; set; } = DefaultPingInterval;
```
anurse: Can you rename this to `PingInterval` so it matches the JavaScript?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
anurse: This should be uncommented, right?

anurse:
```
Microsoft.AspNetCore.SignalR.Client.Tests.HubConnectionTests+Protocol.ClientPingsStopIfSendingOtherMessages [FAIL]
  Failed   Microsoft.AspNetCore.SignalR.Client.Tests.HubConnectionTests+Protocol.ClientPingsStopIfSendingOtherMessages
RUNDOTNET : error Message:  [/home/travis/.dotnet/buildtools/korebuild/2.2.0-preview1-17060/KoreBuild.proj]
   Assert.NotEqual() Failure
  Expected: Not "{\"type\":6}"
  Actual:   "{\"type\":6}"
```

Looks like maybe the "client doesn't send pings if it's sending messages fast enough" test might just be too flaky for our build server. If things get slow, it can end up sending a ping even though we didn't expect it to.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -565,6 +566,66 @@ public class Protocol
...
+                }
+                finally
+                {
+                    await hubConnection.DisposeAsync().OrTimeout();
+                    await connection.DisposeAsync().OrTimeout();
+                }
+            }
+
+            [Fact]
+            public async Task ClientPingsStopIfSendingOtherMessages()
```
anurse: As we discussed in person, let's just remove this test. It's too flaky and it doesn't test a high-value scenario (if we end up pinging despite other messages being sent, it's no big deal)

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
...
+                await delay(322);
+
+                const numPings = connection.sentData.filter((s) => JSON.parse(s).type === MessageType.Ping).length;
+                expect(numPings).toBeGreaterThanOrEqual(2);
+            } finally {
+                await hubConnection.stop();
+            }
+        });
+
+        it("doesn't ping while sending other messages", async () => {
```
anurse: Let's chop this test as well, it could be subject to the same flakiness as the C# one

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -44,18 +46,37 @@ public partial class HubConnection
         private readonly IServiceProvider _serviceProvider;
         private readonly IConnectionFactory _connectionFactory;
         private readonly ConcurrentDictionary<string, InvocationHandlerList> _handlers = new ConcurrentDictionary<string, InvocationHandlerList>(StringComparer.Ordinal);
+        private DateTime _nextActivationServerTimeout;
+        private DateTime _nextActivationSendPing;
```
anurse: Need to have a quick chat about tearing and 32-bit :).

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -48,6 +48,46 @@ describe("HubConnection", () => {
         });
     });
 
+    describe("ping", () => {
+        it("automatically sends multiple pings", async () => {
+            const connection = new TestConnection();
+            const hubConnection = createHubConnection(connection);
+
+            hubConnection.pingIntervalInMilliseconds = 150;
```
BrennanConroy: Set this to 1 and then wait a couple hundred milliseconds below. We don't care if we get 1000 pings, as long as it tests that "multiple" pings are sent

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs
```diff
@@ -174,6 +174,18 @@ private static class Log
...
+
+            private static readonly Action<ILogger, string, Exception> _messageSentGeneric =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(60, "MessageSentGeneric"), "Sending {MessageType} message completed.");
+
+            private static readonly Action<ILogger, Exception> _acquiredConnectionLockForPing =
+                LoggerMessage.Define(LogLevel.Trace, new EventId(61, "AcquiredConnectionLockForPing"), "Have the connection lock, about to ping the server.");
```
BrennanConroy: nit: "Acquired the Connection Lock for writing the ping message."

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs
```diff
@@ -174,6 +174,18 @@ private static class Log
...
+
+            private static readonly Action<ILogger, Exception> _unableToAcquireConnectionLockForPing =
+                LoggerMessage.Define(LogLevel.Trace, new EventId(62, "UnableToAcquireConnectionLockForPing"), "Tried to get connection lock, but it's in use. No need to ping because another message is being sent.");
```
BrennanConroy: nit: EventName: "SkipPing" Message: "Skipping ping because a send is already in progress."

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -53,11 +53,11 @@ describe("HubConnection", () => {
             const connection = new TestConnection();
             const hubConnection = createHubConnection(connection);
 
-            hubConnection.pingIntervalInMilliseconds = 150;
+            hubConnection.pingIntervalInMilliseconds = 5;
 
             try {
                 await hubConnection.start();
-                await delay(322);
+                await delay(32);
```
BrennanConroy: Too small of a delay, on windows the minimum timer tick is 15 milliseconds, so this can be flaky. Just make it 200

Issue #2311, opened 2018-05-18
JsonHubProtocol does not correctly count extra arguments for complex objects


JamesNK:
```
{'type':1,'invocationId':'42','target':'foo','arguments':[1,'',{'1':1,'2':2}]}
```

Error message says there are 8 arguments instead of 3 because the content is not correctly skipped.

anurse:
Reopening to track as a candidate for 2.1.1 patch.

anurse:
Re-closing, we're going to open new issues for patch work.

PR #2312, opened 2018-05-18
Fix JsonHubProtocol error message when too many arguments

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2311


JamesNK:
Yes

Issue #2310, opened 2018-05-18
Flakey test: StreamDoesNotStartIfTokenAlreadyCanceled


JamesNK:
https://ci.appveyor.com/project/aspnetci/signalr/build/1.0.4713

```
Microsoft.AspNetCore.SignalR.Client.FunctionalTests.HubConnectionTests.StreamDoesNotStartIfTokenAlreadyCanceled(protocolName: "json", transportType: ServerSentEvents, path: "/hubT")
RUNDOTNET : error Message:  [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17051\KoreBuild.proj]
   System.Exception : 1 error(s) logged.
  SERVER Microsoft.AspNetCore.Server.Kestrel - ApplicationError - Connection id "0HLDSAJQ34UGH", Request id "0HLDSAJQ34UGH:00000002": An unhandled exception was thrown by the application.
  ===================
  Microsoft.AspNetCore.Server.Kestrel.Core.BadHttpRequestException: Unexpected end of request content.
     at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1MessageBody.<PumpAsync>d__4.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
     at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
     at System.IO.Pipelines.Pipe.GetReadAsyncResult()
     at System.IO.Pipelines.Pipe.DefaultPipeReader.GetResult(Int16 token)
     at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.MessageBody.<CopyToAsync>d__22.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.<CopyToAsyncInternal>d__23.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #2314, opened 2018-05-18
Handle incoming HTTP requests being canceled gracefully

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2310

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.Log.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.Log.cs
@@ -49,6 +49,9 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
             private static readonly Action<ILogger, string, Exception> _connectionDisposedWhileWriteInProgress =
                 LoggerMessage.Define<string>(LogLevel.Debug, new EventId(13, "ConnectionDisposedWhileWriteInProgress"), "Connection {TransportConnectionId} was disposed while a write was in progress.");
 
+            private static readonly Action<ILogger, string, Exception> _badHttpRequest =
+                LoggerMessage.Define<string>(LogLevel.Information, new EventId(14, "BadHttpRequest"), "Connection {TransportConnectionId} received a bad HTTP request.");
+
             public static void ConnectionDisposed(ILogger logger, string connectionId)
             {
                 _connectionDisposed(logger, connectionId, null);
@@ -112,6 +115,11 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
             public static void ConnectionDisposedWhileWriteInProgress(ILogger logger, string connectionId, Exception ex)
             {
                 _connectionDisposedWhileWriteInProgress(logger, connectionId, ex);
+            }
+
+            public static void BadHttpRequest(ILogger logger, string connectionId, Exception ex)
+            {
+                _badHttpRequest(logger, connectionId, ex);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.Log.cs
```diff
@@ -49,6 +49,9 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _connectionDisposedWhileWriteInProgress =
                 LoggerMessage.Define<string>(LogLevel.Debug, new EventId(13, "ConnectionDisposedWhileWriteInProgress"), "Connection {TransportConnectionId} was disposed while a write was in progress.");
 
+            private static readonly Action<ILogger, string, Exception> _badHttpRequest =
```
davidfowl: I wouldn't call this bad request. It's an IO exception.

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.Log.cs
```diff
@@ -49,6 +49,9 @@ private static class Log
...
+                LoggerMessage.Define<string>(LogLevel.Information, new EventId(14, "BadHttpRequest"), "Connection {TransportConnectionId} received a bad HTTP request.");
```
davidfowl: This should be debug.

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
```diff
@@ -505,6 +505,15 @@ private static bool ServerHasWebSockets(IFeatureCollection features)
                         context.Response.ContentType = "text/plain";
                         return;
                     }
+                    catch (IOException ex)
+                    {
+                        // Can occur when the HTTP request is canceled by the client
+                        Log.BadHttpRequest(_logger, connection.ConnectionId, ex);
+
+                        context.Response.StatusCode = StatusCodes.Status400BadRequest;
```
davidfowl: The interesting thing about this is that the client is already gone and that's why you got an IOException happened in the first place. The chances of the response going somewhere is 0

JamesNK: It occurred to me when I wrote it. Doesn't hurt 🤷‍♂️

davidfowl: I'd remove it.

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.Log.cs
davidfowl: Failed to read request data

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
JamesNK: Can there be an IOException from Kestrel for something other than a FIN? It could be seen in that case.

davidfowl: a fin usually results in a graceful close (steam returns 0). I guess setting headers doesn’t hurt as long as we don’t explicitly try to write anything.

JamesNK:
🆙 📅

Issue #2305, opened 2018-05-17
DateTime.Kind no longer honored in SignalR RPC


sonicmouse:
I upgraded to `v1.0.0-rc1-final` and I am seeing new behavior.

From the server, I call a method on the client. The parameter object is of type `DateTime`. The `DateTime` object has the `Kind` property set to `Utc`. When it gets to the client method, the `DateTime` object has a `Kind` property of `Local`.

In previous versions of SignalR it would keep the `Kind` as `Utc` across RPCs.

Is this expected now? Do i really have to do a UTC conversion in all SignalR RPCs even if it was originally sent at UTC?

I believe this is an issue that should be changed back to how it previously worked.

davidfowl:
> I upgraded to v1.0.0-rc1-final and I am seeing new behavior.

I'm trying to unpack this statement. Maybe the next step is to send us a repro project showing exactly how you configured the application.

sonicmouse:
Take a test SignalR server/client. Add this to the client:

```
            _hub.On("testDateTimeUTC", new Type[] { typeof(DateTime) }, (o) =>
            {
                System.Diagnostics.Debug.Assert(((DateTime)o[0]).Kind == DateTimeKind.Utc);
                return Task.CompletedTask;
            });
```

Then in the server, call this:
`await proxy.SendAsync("testDateTimeUTC", DateTime.UtcNow);`

The assertion will be hit.

sonicmouse:
```
            _hub = new HubConnectionBuilder()
                .WithUrl(_envSettings.CreatePlayerServiceHubPath(), (x) =>
                {
                    x.Headers.Add("RegistrationKey", regCode);
                    x.Headers.Add("DeviceName", deviceName);
                    x.Headers.Add("TimeZoneOffset", Services.TimeService.TimeZoneOffset.ToString());
                })
                .Build();
```
There are some properties I am adding in that are pertinent to the client connection as header objects. Just string objects. For example `regCode` is a typical registration code "12345-12345-12345", `deviceName` is a simple name like "ABCDEFG-12345", and timezone offset looks like "-5:00"

davidfowl:
@JamesNK any ideas?

JamesNK:
Hmmm, I think parameters are still going through the serializer. I'm not sure what else could have changed to cause this off the top of my head.

sonicmouse:
Are you guys seeing this? Did you try it? Or is it just me?

davidfowl:
[...]

PR #2317, opened 2018-05-19
Fix losing UTC DateTimeKind on ISO8601 UTC values

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2305

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
@@ -182,7 +182,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                                             // If we have an invocation id already we can parse the end result
                                             var returnType = binder.GetReturnType(invocationId);
 
-                                            if (!ReadForType(reader, returnType))
+                                            if (!JsonUtils.ReadForType(reader, returnType))
                                             {
                                                 throw new JsonReaderException("Unexpected end when reading JSON");
                                             }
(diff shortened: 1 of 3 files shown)


davidfowl:
Can we do this as 2 separate fixes? The date fixes should be one and the end of array fixes should be another one. Don’t we also need to handle return types as well as arguments for dates.

src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
```diff
@@ -599,14 +605,48 @@ private HubMessage BindInvocationMessage(string invocationId, string target, obj
...
+            else if (type == typeof(DateTimeOffset) || type == typeof(DateTimeOffset?))
+            {
+                reader.ReadAsDateTimeOffset();
+            }
+            else
+            {
+                reader.Read();
+            }
+
+            return reader.TokenType != JsonToken.None;
```
davidfowl: Comment on what this means.

src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
```diff
@@ -599,14 +605,48 @@ private HubMessage BindInvocationMessage(string invocationId, string target, obj
...
 
+        private static bool ReadForType(JsonTextReader reader, Type type)
+        {
+            // Explicity read values as dates from JSON with reader.
+            // We do this because otherwise dates are read as strings
+            // and the JsonSerializer will use a conversion method that won't
+            // preserve UTC in DateTime.Kind for UTC ISO8601 dates
+            if (type == typeof(DateTime) || type == typeof(DateTime?))
+            {
+                reader.ReadAsDateTime();
```
davidfowl: This changes the token type to Date right? Do we have anything else like this?

src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
```diff
@@ -599,14 +605,48 @@ private HubMessage BindInvocationMessage(string invocationId, string target, obj
             return new InvocationMessage(invocationId, target, arguments);
         }
 
+        private static bool ReadForType(JsonTextReader reader, Type type)
```
davidfowl: Move this to JsonUtils

src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
JamesNK: Yes.

What?

davidfowl: I was looking at `JsonTokenType` and JsonSerializerSettings to see if there are any other things in this category. The only one I see is FloatParseHandling.

JamesNK: Only date reading was disable in the reader.

PR #2319, opened 2018-05-19
 Fix not reading to end of argument JSON on binding error

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2321

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
@@ -608,36 +608,6 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
             return new InvocationMessage(invocationId, target, arguments);
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.SignalR.Protocols.Json/Protocol/JsonHubProtocol.cs
```diff
@@ -225,6 +226,14 @@ private HubMessage ParseMessage(Utf8BufferTextReader textReader, IInvocationBind
                                             catch (Exception ex)
                                             {
                                                 argumentBindingException = ExceptionDispatchInfo.Capture(ex);
+
+                                                // Could be at any point in argument array JSON when an error is thrown
+                                                // Read until the end of the argument JSON array
+                                                while (reader.Depth == initialDepth && reader.TokenType == JsonToken.StartArray ||
```
davidfowl: This matters because we catch the error right? I guess today the stream is corrupted at this point since we don't consume everything? 

I'm trying to figure out why this is the only place we need to do this.

PR #2345, opened 2018-05-22
Clarify record separator usage

anurse:
The record separator was previously only mentioned in a section on JSON payloads, which may not have been clear that it's required for all messages.


davidfowl:
LGTM

Issue #1948, opened 2018-04-11
Consider providing an xunit test suite for HubLifetimeManager implementations


davidfowl:
DI and EF has used this technique to great success. Basically for any provider model, you provide a suite of tests that need to pass in order for to validate your behavior assumptions. We provide the harness, the user plugs in their HubLifetimeManager and we validate our expectations about their implementation.

anurse:
Sounds great for 2.2.0-mq ;)

davidfowl:
Moving to the backlog

davidfowl:
We should do this in 2.2.0-mq: See

https://github.com/DarkCow/SignalR.Orleans/tree/ef5b9f3f8c2f9eb1f475a8285977b5fd5b8d9d9d/test/SignalR.Orleans.Tests

anurse:
A lot of this would just be converting our existing tests for `DefaultHubLifetimeManager` and `RedisHubLifetimeManager` into an abstract base class and shipping it in an assembly.

anurse:
@mikaelm12 you were mentioning you were ready for some new MQ work, want to take a look at this?

davidfowl:
Yes, it'll also help the service.

mikaelm12:
648705f6484ffc0ceefab7c571747af895a389d4

anurse:
Reopening this as the packaging needs to be fixed. See @natemcmaster 's comments in the #2353

mikaelm12:
Fixed: dd64e4d9a5b84c6e3a5ee850ab20336daf4faa14

PR #2353, opened 2018-05-23
Add A  Public Test Suite for HubLifetimeManager

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/1948

So here I created a base class for HubLifetimeManager testing losely based on how EF goes about this called  `HubLifetimeManagerTestsBase`. 
I put some of the simpler testing scenarios here and I still need to see what else will fit there (apply to all HubLifetimeManagers). I put it in common for now but it'll need to be it's own project. Named something like `Microsoft.AspNetCore.SignalR.Specification.Tests`.

Also there's the question of if this should live under `src` instead of `test` since it's something that we would want to ship. Here you can see that EF does put their spec test project under `src `https://github.com/aspnet/EntityFrameworkCore/tree/dev/src/EFCore.Relational.Specification.Tests

I just wanted to open this for discussion on how we want to go about exposing this. Still cleaning this up.


test/Microsoft.AspNetCore.SignalR.Common.Tests/RedisCustomHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,484 @@
...
+
+        private async Task AssertMessageAsync(TestClient client)
+        {
+            var message = Assert.IsType<InvocationMessage>(await client.ReadAsync().OrTimeout());
+            Assert.Equal("Hello", message.Target);
+            Assert.Single(message.Arguments);
+            Assert.Equal("World", (string)message.Arguments[0]);
+        }
+        [Fact]
+        public async Task InvokeAllAsyncWithMultipleServersWritesToAllConnectionsOutput()
```
davidfowl: What's redis specific about all of these tests?

mikaelm12: So there was actually a gap in the testing across the HubLifetimeManagers. Redis had a lot more tests than Default but they weren't all specific to the Redis scenario. Part of the clean up that I'm going to do.

davidfowl:
A few pieces of feedback:

- It should be under src
- We need to test one other real lifetime manager implementation (we can get the orleans impl and we can also try out the azure signalr impl).

test/Microsoft.AspNetCore.SignalR.Common.Tests/RedisCustomHubLifetimeManagerTests.cs
BrennanConroy: This shouldn't be redis specific, it should be scaleout specific. I think we should add another class that inherits from the generic lifetime manager test suite and provides scaleout specific tests.

test/Microsoft.AspNetCore.SignalR.Common.Tests/HubLifetimeManagerTestsBase.cs
```diff
@@ -0,0 +1,142 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR.Protocol;
+using Microsoft.AspNetCore.SignalR.Tests;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Common.Tests
+{
+    public abstract class HubLifeTimeManagerTestsBase<THub> where THub : Hub
```
anurse: nit: `Lifetime` (lowercase `t`)

test/Microsoft.AspNetCore.SignalR.Common.Tests/HubLifetimeManagerTestsBase.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+        public HubLifetimeManager<THub> Manager;
+
+        public abstract HubLifetimeManager<THub> GetNewHubLifetimeManager();
```
anurse: nit: `CreateHubLifetimeManager`

test/Microsoft.AspNetCore.SignalR.Common.Tests/HubLifetimeManagerTestsBase.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR.Protocol;
+using Microsoft.AspNetCore.SignalR.Tests;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Common.Tests
+{
+    public abstract class HubLifeTimeManagerTestsBase<THub> where THub : Hub
+    {
+        public HubLifetimeManager<THub> Manager;
```
anurse: Property ;)

test/Microsoft.AspNetCore.SignalR.Common.Tests/HubLifetimeManagerTestsBase.cs
davidfowl: LIFETIME

anurse: lIfETyMe

JamesNK: life_time

test/Microsoft.AspNetCore.SignalR.Common.Tests/RedisCustomHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,484 @@
...
+        }
+
+        private async Task AssertMessageAsync(TestClient client)
+        {
+            var message = Assert.IsType<InvocationMessage>(await client.ReadAsync().OrTimeout());
+            Assert.Equal("Hello", message.Target);
+            Assert.Single(message.Arguments);
+            Assert.Equal("World", (string)message.Arguments[0]);
+        }
+        [Fact]
```
JamesNK: nit: new line to separate methods

test/Microsoft.AspNetCore.SignalR.Common.Tests/RedisCustomHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,484 @@
...
+                await AssertMessageAsync(client2);
+
+                // Disconnect one connection for the user
+                await manager.OnDisconnectedAsync(connection1).OrTimeout();
+                await manager.SendUserAsync("userA", "Hello", new object[] { "World" }).OrTimeout();
+                await AssertMessageAsync(client2);
+            }
+        }
+
+
```
JamesNK: nit: remove extra line

anurse:
This looks good, I think the pattern is what I would expect. Let's break this into two chunks though so we don't get too bogged down. Let's get the infrastructure in place and make sure we have an issue filed to fill out the tests in the next iteration. Same with verifying the infrastructure on an existing third-party HubLifetimeManager. Those are all important aspects of this, but don't have to block this PR.

davidfowl:
We need to figure out what public API this exposes.

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
```diff
@@ -0,0 +1,38 @@
...
+    
+      
+    <ItemGroup>
+    <Content Include="..\..\test\xunit.runner.json" Link="xunit.runner.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.SignalR.Common\Microsoft.AspNetCore.SignalR.Common.csproj" />
```
davidfowl: I think these are problematic, we need to narrow down all of the dependencies in here

davidfowl: @mikaelm12 what are you doing about these?

src/Microsoft.AspNetCore.SignalR.Specification.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
BrennanConroy: These should continue living in the test folder and just use the `Specification.Tests` project

src/Microsoft.AspNetCore.SignalR.Specification.Tests/DefaultHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
BrennanConroy: These should continue living in the test folder and just use the `Specification.Tests` project

src/Microsoft.AspNetCore.SignalR.Specification.Tests/HubLifetimeManagerTestBase.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR.Protocol;
+using Microsoft.AspNetCore.SignalR.Tests;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Specification.Tests
+{
+    public abstract class HubLifetimeManagerTestsBase<THub> where THub : Hub
+    {
+        public HubLifetimeManager<THub> Manager { get; set; }
```
BrennanConroy: Remove

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.Testing" Version="$(MicrosoftAspNetCoreTestingPackageVersion)" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPackageVersion)" />
+    <PackageReference Include="Moq" Version="$(MoqPackageVersion)" />
+    <PackageReference Include="xunit" Version="$(XunitPackageVersion)" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitRunnerVisualStudioPackageVersion)" />
+  </ItemGroup>
+    
+      
+    <ItemGroup>
+    <Content Include="..\..\test\xunit.runner.json" Link="xunit.runner.json">
```
BrennanConroy: Remove

src/Microsoft.AspNetCore.SignalR.Specification.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        private TestRedisServer _server;
+
+        public override HubLifetimeManager<MyHub> CreateNewHubLifetimeManager()
+        {
+            _server =  new TestRedisServer();
+
+            return CreateLifetimeManager(_server);
+        }
+
+        public override HubLifetimeManager<MyHub> CreateNewHubLifetimeManagerFromExistingServer()
```
BrennanConroy: Seems odd, why not just use the `CreateNewHubLifetimeManager` function and let the user decide if they want to cache an internal server or not

anurse:
When you say public API are you referring to the things that derived test classes need to implement? That's what I mean by the infrastructure. Let's try to close on that and get this in, then we can file an issue to track reviewing the tests themselves.

src/Microsoft.AspNetCore.SignalR.Specification.Tests/ScaleoutHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,412 @@
...
+using Microsoft.AspNetCore.SignalR.Protocol;
+using Microsoft.AspNetCore.SignalR.Tests;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Specification.Tests
+{
+    public abstract class ScaleoutHubLifetimeManagerTests : HubLifetimeManagerTestsBase<MyHub>
+    {
+        public abstract HubLifetimeManager<MyHub> CreateNewHubLifetimeManagerFromExistingServer();
```
anurse: If we break this up a little, I think we can have a cleaner API. I think we should split this into an API for creating a "Backplane" (i.e. TestRedisServer in the case of the Redis one). Then for our tests that want to test two "servers" connected via a common backplane, the test would call this API, then pass the Backplane object in to each `CreateHubLifetimeManager` method.

Something like this:

```csharp
public async Task TestMultipleServersConnectedToTheSameBackplane()
{
    var backplane = CreateBackplane();
    var server1 = CreateHubLifetimeManager(backplane);
    var server2 = CreateHubLifetimeManager(backplane);
    // The test...
}

public async Task TestOneServer()
{
    var backplane = CreateBackplane();
    var server = CreateHubLifetimeManager(backplane);
    // The test...
}
```

This also allows the test code to get an explicit call to initialize however they want.

davidfowl:
> When you say public API are you referring to the things that derived test classes need to implement? That's what I mean by the infrastructure. Let's try to close on that and get this in, then we can file an issue to track reviewing the tests themselves.

I don't want to start producing packages for this that depend on our testing infrastructure that we never intended to make public in the first place.

https://github.com/aspnet/SignalR/pull/2353/files/6cf1e4f3482b00ebcded6a33b1adafbe6e80acc1..92b323e4815d727f4c5519ef106de45d96762a54#diff-718a409f6d0400b4f5bd996f9ad585feR34

anurse:
Ah yeah, I agree on that. We could look at source-including those things in this package (as `internal`)

mikaelm12:
>We could look at source-including those things in this package (as internal)

Is that the approach we want to take?

anurse:
I think that's the best approach. We need to make sure they are brought in as `internal` though.

anurse:
What's the status on this? MQ is coming to a close, so we want to get things resolved ASAP.

mikaelm12:
Made some changes, like conditionally making things internal to fix issues with ambiguous references with source linked files as per @pranavkm's  suggestion. 
Need to fix the failing test though

src/Directory.Build.props
```diff
@@ -11,4 +11,8 @@
     <PackageReference Include="Internal.AspNetCore.Sdk" PrivateAssets="All" Version="$(InternalAspNetCoreSdkPackageVersion)" />
     <PackageReference Include="Internal.AspNetCore.Analyzers" PrivateAssets="All" Version="$(InternalAspNetCoreAnalyzersPackageVersion)" />
   </ItemGroup>
-</Project>
+
+  <!--<PropertyGroup>
+    <EnableApiCheck>false</EnableApiCheck>
+  </PropertyGroup>-->
```
anurse: What's this about? Bad merge or something intentional?

src/Microsoft.AspNetCore.SignalR.Specification.Tests/HubLifetimeManagerTestBase.cs
```diff
@@ -0,0 +1,142 @@
...
+                var connection1 = HubConnectionContextUtils.Create(client1.Connection);
+                var connection2 = HubConnectionContextUtils.Create(client2.Connection);
+
+                await manager.OnConnectedAsync(connection1).OrTimeout();
+                await manager.OnConnectedAsync(connection2).OrTimeout();
+
+                await manager.AddToGroupAsync(connection1.ConnectionId, "gunit").OrTimeout();
+                await manager.AddToGroupAsync(connection2.ConnectionId, "gunit").OrTimeout();
+
+                await manager.SendGroupExceptAsync("gunit", "Hello", new object[] { "World" }, new[] { connection2.ConnectionId }).OrTimeout();
```
anurse: Maybe now that we're going to "ship" these tests as things we want our customers to use we should just use boring old names like "Group1"... 😢

src/Microsoft.AspNetCore.SignalR.Specification.Tests/ScaleoutHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,399 @@
...
+                await manager2.RemoveFromGroupAsync(connection.ConnectionId, "name").OrTimeout();
+
+                await manager2.SendGroupAsync("name", "Hello", new object[] { "World" }).OrTimeout();
+
+                Assert.Null(client.TryRead());
+            }
+        }
+
+        [Fact]
+        public async Task InvokeConnectionAsyncForLocalConnectionDoesNotPublishToRedis()
```
anurse: nit: InvokeConnectionAsyncForLocalConnectionDoesNotPublishTo~~Redis~~**Backplane**

src/Microsoft.AspNetCore.SignalR.Specification.Tests/ScaleoutHubLifetimeManagerTests.cs
```diff
@@ -0,0 +1,399 @@
...
+            using (var client = new TestClient())
+            {
+                // Force an exception when writing to connection
+                var connectionMock = HubConnectionContextUtils.CreateMock(client.Connection);
+                connectionMock.Setup(m => m.WriteAsync(It.IsAny<HubMessage>(), It.IsAny<CancellationToken>())).Throws(new Exception());
+                var connection = connectionMock.Object;
+
+                await manager2.OnConnectedAsync(connection).OrTimeout();
+
+                // This doesn't throw because there is no connection.ConnectionId on this server so it has to publish to redis.
```
anurse: nit: this server so it has to publish to ~~redis~~ **the backplane**.

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
```diff
@@ -1,558 +1,58 @@
...
-                await AssertMessageAsync(client1);
-
-                Assert.Null(client2.TryRead());
-            }
+            return CreateLifetimeManager(_server);
         }
 
-        [Fact]
-        public async Task InvokeConnectionAsyncOnServerWithoutConnectionWritesOutputToConnection()
+        class TestObject
```
anurse: Is this still needed?

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,14 @@
...
+////[assembly: InternalsVisibleTo("Microsoft.AspNetCore.SignalR.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
+////[assembly: InternalsVisibleTo("Microsoft.AspNetCore.SignalR.Redis.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
anurse: Remove these

test/Microsoft.AspNetCore.SignalR.Tests.Utils/Microsoft.AspNetCore.SignalR.Tests.Utils.csproj
```diff
@@ -1,8 +1,9 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>$(StandardTestTfms)</TargetFrameworks>
     <RootNamespace>Microsoft.AspNetCore.SignalR.Tests</RootNamespace>
+  <DefineConstants>$(DefineConstants);TESTUTILS</DefineConstants>
```
anurse: spacing

test/Microsoft.AspNetCore.SignalR.Tests.Utils/TaskExtensions.cs
```diff
@@ -7,7 +7,12 @@
 
 namespace System.Threading.Tasks
 {
-    public static class TaskExtensions
+#if TESTUTILS
```
anurse: Kinda ugly, but it works, so let's just get it done ;)

src/Microsoft.AspNetCore.SignalR.Specification.Tests/HubLifetimeManagerTestBase.cs
mikaelm12: LOL. @BrennanConroy made these names. 
I'll change them

src/Directory.Build.props
mikaelm12: I was running into an issue with it for some reason a little while ago but it's gone now.. 
Removing this

test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisHubLifetimeManagerTests.cs
mikaelm12: It's used in one of the tests. I didn't really want to change the inner workings of the tests

anurse:

natemcmaster:
We need to change the dependencies on this project. cref http://xunit.github.io/releases/2.3

> As a reminder: If you're extending xUnit.net and want to publish your extension as a NuGet package, you should import xunit.extensibility.core and/or xunit.extensibility.execution, not xunit or xunit.core. If you do this wrong, you might have problems generating your NuGet package via dotnet pack.

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
```diff
@@ -0,0 +1,40 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Tests for users to verify their own implementations of SignalR types</Description>
+    <DeveloperBuildTestTfms>netcoreapp2.2</DeveloperBuildTestTfms>
```
natemcmaster: Change this to use `<TargetFramework>` like a normal class library project. Also, consider making this netstandard2.0.

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <Compile Include="..\..\test\Microsoft.AspNetCore.SignalR.Tests.Utils\TestClient.cs" Link="TestClient.cs" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Core\Microsoft.AspNetCore.SignalR.Core.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Common\Microsoft.AspNetCore.SignalR.Common.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Protocols.MessagePack\Microsoft.AspNetCore.SignalR.Protocols.MessagePack.csproj" />
+  </ItemGroup>
+  
+
+  <ItemGroup>
+    <PackageReference Include="Internal.AspNetCore.Sdk" PrivateAssets="All" Version="$(InternalAspNetCoreSdkPackageVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Testing" Version="$(MicrosoftAspNetCoreTestingPackageVersion)" />
```
natemcmaster: This is not a shipping package. These tests cannot rely on this.

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPackageVersion)" />
+    <PackageReference Include="Moq" Version="$(MoqPackageVersion)" />
```
natemcmaster: We should be very careful about this. This would be a new 3rd party dependency for our packaage consumers. At the moment, Moq is only an "internal" dependency of our test code. cc @Eilon

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <PackageReference Include="xunit" Version="$(XunitPackageVersion)" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitRunnerVisualStudioPackageVersion)" />
```
natemcmaster: Remove. This shouldn't be in the generated nuspec.

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.SignalR.Protocols.MessagePack\Microsoft.AspNetCore.SignalR.Protocols.MessagePack.csproj" />
+  </ItemGroup>
+  
+
+  <ItemGroup>
+    <PackageReference Include="Internal.AspNetCore.Sdk" PrivateAssets="All" Version="$(InternalAspNetCoreSdkPackageVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Testing" Version="$(MicrosoftAspNetCoreTestingPackageVersion)" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPackageVersion)" />
+    <PackageReference Include="Moq" Version="$(MoqPackageVersion)" />
+    <PackageReference Include="xunit" Version="$(XunitPackageVersion)" />
```
natemcmaster: Class libraries that ship tests should use `xunit.execution.core` and/or `xunit.extensibility.execution` instead of the `xunit` metapackage

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
anurse: I'm not sure we actually need to use Moq. But we can also look to remove it if we do.

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
mikaelm12: This can actually just be removed

src/Microsoft.AspNetCore.SignalR.Specification.Tests/Microsoft.AspNetCore.SignalR.Specification.Tests.csproj
mikaelm12: It looks like it's only used in HubConnectionContextUtils which is source linked

mikaelm12:
Addressing the new comments here: https://github.com/aspnet/SignalR/pull/2451

PR #2357, opened 2018-05-23
[2.1.1] Fix losing UTC DateTimeKind on ISO8601 UTC values (#2317)

anurse:
Port #2317 to `release/2.1`


anurse:
Approved for 2.1.1 **but the branch is not yet open**

PR #2360, opened 2018-05-23
[2.1.1] Fix not reading to end of argument JSON on binding error (#2319)

anurse:
Port #2319 to `release/2.1`


anurse:
Approved for 2.1.1 **but the branch is not yet open**

PR #2363, opened 2018-05-23
[2.1.1] Bump package versions

natemcmaster:


natemcmaster:
I'll merge this PR once release/2.1 branches are open for 2.1.1 changes. Until then, leaving this open as "SHP: pending approval".

natemcmaster:
@anurse is any manual action required to also update the npm package to 1.0.1?

anurse:
Maybe. Let me take a look

anurse:
Looks like we're good. We use the `PackageVersion` msbuild variable to fill it in, so as long as that is set correctly we're good. Probably not a bad idea to update the `package.json` files but the build process overwrites the value within the `package.json` anyway.

image

anurse:
I pushed a commit updating the `package.json` versions, just because it makes it all consistent. As I said, it's not mandatory to do thing in order to make the package versions correct.

natemcmaster:
Thanks

Issue #1637, opened 2018-03-18
Optimize HubConnectionContext.WriteAsync


davidfowl:
![image](https://user-images.githubusercontent.com/95136/37564094-13232e04-2a4c-11e8-84bf-592f99041dd3.png)

SignalR controls a couple of things here:
- The calls to GetTimeStamp ~19.7%
- The cost of WriteMessage  ~11.6%
- The use of the SemaphoreSlim ~4.5%

We also have 2 layers of state machines that can be folded into 1. SafeWriteAsync and WriteAsync. We can move the error handling into WriteAsync and make that a part of the contract of HubConnectionContext.WriteAsync (it should never throw).

We should also try to completely get rid of the state machine until there is back pressure.

davidfowl:
We made progress here but I want to keep looking at this in the rc so I'll move it there.

PR #2364, opened 2018-05-23
Dygray/ping catcher

### LINKED ISSUE #2289: Use Client-to-Server Pings to mark clients disconnected
Prerequisite: #2175 

Once we have client to server ping messages, we want to use them to have the client marked as disconnected when it hasn't been heard from within a certain interval, much like we do with the server.

DylanDmitri:
Fixes #2289

Server catches pings from clients. Silent clients are timed out and disconnected.

Clients auto-ping on startup to signal that it's a pinging client. This first ping is caught in ```DefaultHubDispatcher```, which registers a timeout method to the ```HubConnectionContext``` heartbeat. 

The timeout window is reset whenever a message is received (per connection, under ```DispatchMessagesAsync``` in ```HubConnectionHandler```).

On timeout the connection is closed.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -884,7 +884,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private async Task TimerLoop(TimerAwaitable timer)
         {
-            // tell the server you intend to ping
+            // tell the server we intend to ping
             await SendHubMessage(PingMessage.Instance);
 
             // initialize the timers
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -885,7 +885,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private async Task TimerLoop(TimerAwaitable timer)
         {
             // Tell the server we intend to ping
-            // Pld clients never ping, and shouldn't be timed out
+            // Old clients never ping, and shouldn't be timed out
             // So ping to tell the server that we should be timed out if we stop
             await SendHubMessage(PingMessage.Instance);
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -884,7 +884,9 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private async Task TimerLoop(TimerAwaitable timer)
         {
-            // tell the server we intend to ping
+            // Tell the server we intend to ping
+            // Pld clients never ping, and shouldn't be timed out
+            // So ping to tell the server that we should be timed out if we stop
             await SendHubMessage(PingMessage.Instance);
 
             // initialize the timers
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -886,12 +886,11 @@ namespace Microsoft.AspNetCore.SignalR.Client
         {
             // tell the server you intend to ping
             await SendHubMessage(PingMessage.Instance);
-            //await SendCoreAsyncCore(PingMessage.Instance);
 
             // initialize the timers
-            ResetSendPing();
-            ResetTimeout();
             timer.Start();
+            ResetTimeout();
+            ResetSendPing();
 
             using (timer)
             {
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -884,11 +884,15 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         private async Task TimerLoop(TimerAwaitable timer)
         {
+            // tell the server you intend to ping
+            await SendHubMessage(PingMessage.Instance);
+            //await SendCoreAsyncCore(PingMessage.Instance);
+
             // initialize the timers
-            timer.Start();
             ResetSendPing();
             ResetTimeout();
-            
+            timer.Start();
+
             using (timer)
             {
                 // await returns True until `timer.Stop()` is called in the `finally` block of `ReceiveLoop`
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -34,7 +34,7 @@ namespace Microsoft.AspNetCore.SignalR
 
         private long _lastSendTimeStamp = DateTime.UtcNow.Ticks;
         private long _lastReceivedTimeStamp = DateTime.UtcNow.Ticks;
-        internal bool _receivedMessageThisInterval = false;
+        private bool _receivedMessageThisInterval = false;
         private ReadOnlyMemory<byte> _cachedPingMessage;
         private bool _clientTimeoutActive;
 
@@ -506,6 +506,11 @@ namespace Microsoft.AspNetCore.SignalR
                 // Communicate the fact that we're finished triggering abort callbacks
                 connection._abortCompletedTcs.TrySetResult(null);
             }
+        }
+
+        internal void ResetClientTimeout()
+        {
+            _receivedMessageThisInterval = true;
         }
 
         private static class Log
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -35,7 +35,6 @@ namespace Microsoft.AspNetCore.SignalR
         private long _lastSendTimestamp = DateTime.UtcNow.Ticks;
         private long _lastReceivedTimestamp = DateTime.UtcNow.Ticks;
         private ReadOnlyMemory<byte> _cachedPingMessage;
-        private bool _clientTimeoutActive;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="HubConnectionContext"/> class.
@@ -388,6 +387,9 @@ namespace Microsoft.AspNetCore.SignalR
                                         Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext)state).KeepAliveTick(), this);
                                     }
 
+                                    // only assign this if the 
+                                    Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext)state).CheckClientTimeout(), this);
+
                                     Log.HandshakeComplete(_logger, Protocol.Name);
                                     await WriteHandshakeResponseAsync(HandshakeResponseMessage.Empty);
                                     return true;
@@ -449,16 +451,6 @@ namespace Microsoft.AspNetCore.SignalR
 
                 Interlocked.Exchange(ref _lastSendTimestamp, currentTime);
             }
-        }
-
-        public void StartPeriodicallyCheckingForClientTimeout()
-        {
-            if (_clientTimeoutActive)
-            {
-                return;
-            }
-            _clientTimeoutActive = true;
-            Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext)state).CheckClientTimeout(), this);
         }
 
         private void CheckClientTimeout()
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -121,7 +121,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
 
             // Otherwise, release the lock acquired when entering WriteAsync
-            DoneWithSend();
+            _writeLock.Release();
 
             return default;
         }
@@ -151,7 +151,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
 
             // Otherwise, release the lock acquired when entering WriteAsync
-            DoneWithSend();
+            _writeLock.Release();
 
             return default;
         }
@@ -204,7 +204,7 @@ namespace Microsoft.AspNetCore.SignalR
             finally
             {
                 // Release the lock acquired when entering WriteAsync
-                DoneWithSend();
+                _writeLock.Release();
             }
         }
 
@@ -223,7 +223,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                DoneWithSend();
+                _writeLock.Release();
             }
         }
 
@@ -242,7 +242,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                DoneWithSend();
+                _writeLock.Release();
             }
         }
 
@@ -272,7 +272,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                DoneWithSend();
+                _writeLock.Release();
             }
         }
 
@@ -296,7 +296,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                DoneWithSend();
+                _writeLock.Release();
             }
         }
 
@@ -456,6 +456,8 @@ namespace Microsoft.AspNetCore.SignalR
                 // adding a Ping message when the transport is full is unnecessary since the
                 // transport is still in the process of sending frames.
                 _ = TryWritePingAsync();
+
+                Interlocked.Exchange(ref _lastSendTimestamp, currentTime);
             }
         }
 
@@ -499,12 +501,6 @@ namespace Microsoft.AspNetCore.SignalR
                 // Communicate the fact that we're finished triggering abort callbacks
                 connection._abortCompletedTcs.TrySetResult(null);
             }
-        }
-
-        private void DoneWithSend()
-        {
-            _writeLock.Release();
-            Volatile.Write(ref _lastSendTimestamp, DateTime.UtcNow.Ticks);
         }
 
         private static class Log
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -115,7 +115,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
 
             // Otherwise, release the lock acquired when entering WriteAsync
-            _writeLock.Release();
+            DoneWithSend();
 
             return default;
         }
@@ -145,7 +145,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
 
             // Otherwise, release the lock acquired when entering WriteAsync
-            _writeLock.Release();
+            DoneWithSend();
 
             return default;
         }
@@ -198,7 +198,7 @@ namespace Microsoft.AspNetCore.SignalR
             finally
             {
                 // Release the lock acquired when entering WriteAsync
-                _writeLock.Release();
+                DoneWithSend();
             }
         }
 
@@ -217,7 +217,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                _writeLock.Release();
+                DoneWithSend();
             }
         }
 
@@ -236,7 +236,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                _writeLock.Release();
+                DoneWithSend();
             }
         }
 
@@ -266,7 +266,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                _writeLock.Release();
+                DoneWithSend();
             }
         }
 
@@ -290,7 +290,7 @@ namespace Microsoft.AspNetCore.SignalR
             }
             finally
             {
-                _writeLock.Release();
+                DoneWithSend();
             }
         }
 
@@ -449,8 +449,6 @@ namespace Microsoft.AspNetCore.SignalR
                 // adding a Ping message when the transport is full is unnecessary since the
                 // transport is still in the process of sending frames.
                 _ = TryWritePingAsync();
-
-                Interlocked.Exchange(ref _lastSendTimestamp, currentTime);
             }
         }
 
@@ -494,6 +492,12 @@ namespace Microsoft.AspNetCore.SignalR
                 // Communicate the fact that we're finished triggering abort callbacks
                 connection._abortCompletedTcs.TrySetResult(null);
             }
+        }
+
+        private void DoneWithSend()
+        {
+            _writeLock.Release();
+            Interlocked.Exchange(ref _lastSendTimestamp, DateTime.UtcNow.Ticks);
         }
 
         private static class Log
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -32,8 +32,8 @@ namespace Microsoft.AspNetCore.SignalR
         private readonly long _clientTimeoutInterval;
         private readonly SemaphoreSlim _writeLock = new SemaphoreSlim(1);
 
-        private int _heartbeatsFromLastSend = 0;
-        private int _heartbeatsFromLastReceive = 0;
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -857,11 +857,15 @@ public void ResetTimeout()
 
         private async Task TimerLoop(TimerAwaitable timer)
         {
+            // tell the server you intend to ping
+            await SendHubMessage(PingMessage.Instance);
+            //await SendCoreAsyncCore(PingMessage.Instance);
```
JamesNK: remove

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -857,11 +857,15 @@ public void ResetTimeout()
...
+
             // initialize the timers
-            timer.Start();
             ResetSendPing();
             ResetTimeout();
-            
+            timer.Start();
```
JamesNK: Is there a reason you moved Start afterwards? Add a one sentence comment saying why it should happen after the reset methods.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -28,24 +28,29 @@ public class HubConnectionContext
         private readonly ILogger _logger;
         private readonly CancellationTokenSource _connectionAbortedTokenSource = new CancellationTokenSource();
         private readonly TaskCompletionSource<object> _abortCompletedTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
-        private readonly long _keepAliveDuration;
+        private readonly TimeSpan _keepAliveInterval;
+        private readonly TimeSpan _clientTimeoutInterval;
         private readonly SemaphoreSlim _writeLock = new SemaphoreSlim(1);
 
-        private long _lastSendTimestamp = Stopwatch.GetTimestamp();
+        private long _lastSendTimestamp = DateTime.UtcNow.Ticks;
```
JamesNK: How come you're using UtcNow.Ticks instead of GetTimestamp?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -28,24 +28,29 @@ public class HubConnectionContext
...
         /// <summary>
         /// Initializes a new instance of the <see cref="HubConnectionContext"/> class.
         /// </summary>
         /// <param name="connectionContext">The underlying <see cref="ConnectionContext"/>.</param>
-        /// <param name="keepAliveInterval">The keep alive interval.</param>
+        /// <param name="keepAliveInterval">How often this client is pinged.</param>
         /// <param name="loggerFactory">The logger factory.</param>
-        public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory)
+        /// <param name="clientTimeoutInterval">Clients we haven't heard from in this interval are assumed to have disconnected.</param>
+        public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory, TimeSpan? clientTimeoutInterval = null)
```
JamesNK: Changing the public ctor is an API breaking change. What's the policy here @anurse ?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -428,24 +433,48 @@ internal Task AbortAsync()
...
+            {
+                return;
+            }
+            _clientTimeoutActive = true;
+            Features.Get<IConnectionHeartbeatFeature>()?.OnHeartbeat(state => ((HubConnectionContext)state).CheckClientTimeout(), this);
+        }
+
+        private void CheckClientTimeout()
+        {
+            // if it's been too long since we've heard from the client, then close this
```
JamesNK: `If it's been too long` 😄

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -428,24 +433,48 @@ internal Task AbortAsync()
...
+            if (DateTime.UtcNow.Ticks - Interlocked.Read(ref _lastReceivedTimestamp) > _clientTimeoutInterval.Ticks)
+            {
+                Abort();
+            }
+        }
+
+        public void ResetClientTimeout()
+        {
+            _lastReceivedTimestamp = DateTime.UtcNow.Ticks;
```
JamesNK: Some places the ticks are being called time and others it is timestamp. I don't have a preference, just use one consistently.

src/Microsoft.AspNetCore.SignalR.Core/HubOptions.cs
```diff
@@ -26,6 +26,11 @@ public class HubOptions
         /// </summary>
         public TimeSpan? KeepAliveInterval { get; set; } = null;
 
+        /// <summary>
+        /// Gets or sets the interval used by the server to
```
JamesNK: to....

src/Microsoft.AspNetCore.SignalR.Core/Internal/HubOptionsSetup.cs
```diff
@@ -14,26 +14,23 @@ public class HubOptionsSetup : IConfigureOptions<HubOptions>
 
         internal static TimeSpan DefaultKeepAliveInterval => TimeSpan.FromSeconds(15);
 
-        private readonly List<string> _protocols = new List<string>();
+        public static TimeSpan DefaultClientTimeoutInterval => TimeSpan.FromSeconds(30);
```
JamesNK: The other defaults here are internal, not public.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -580,10 +580,10 @@ public class Protocol
                 {
                     await hubConnection.StartAsync().OrTimeout();
 
-                    var firstPing = await connection.ReadSentTextMessageAsync().OrTimeout(TimeSpan.FromMilliseconds(200));
+                    var firstPing = await connection.ReadSentTextMessageAsync(ignorePings:false).OrTimeout(TimeSpan.FromMilliseconds(200));
```
JamesNK: nit: space after :

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -580,10 +580,10 @@ public class Protocol
...
                     Assert.Equal("{\"type\":6}", firstPing);
 
-                    var secondPing = await connection.ReadSentTextMessageAsync().OrTimeout(TimeSpan.FromMilliseconds(200));
+                    var secondPing = await connection.ReadSentTextMessageAsync(ignorePings:false).OrTimeout(TimeSpan.FromMilliseconds(200));
```
JamesNK: nit: space after :

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -116,9 +116,23 @@ public Task ReceiveBytesAsync(byte[] bytes)
         {
             return Application.Output.WriteAsync(bytes).AsTask();
         }
-        public async Task<string> ReadSentTextMessageAsync()
+        public async Task<string> ReadSentTextMessageAsync(bool ignorePings=true)
```
JamesNK: nit: spaces around =

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -116,9 +116,23 @@ public Task ReceiveBytesAsync(byte[] bytes)
...
         {
             // Read a single text message from the Application Input pipe
+
+            while (true)
+            {
+                var result = await ReadSentTextMessageAsyncInner();
+                if (ignorePings && result == "{\"type\":6}")
```
JamesNK: Add a `IsPingMessage` method for testing whether content is a ping or not.

Load string into a `JObject`, then test if the "type" property is 6. That will be more resilient to any unit tests that change the formatting to include spaces or line breaks in JSON messages.

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -144,7 +158,7 @@ public Task ReceiveBytesAsync(byte[] bytes)
             }
         }
 
-        public async Task<IList<string>> ReadAllSentMessagesAsync()
+        public async Task<IList<string>> ReadAllSentMessagesAsync(bool ignorePings=true)
```
JamesNK: nit: spaces around =

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
JamesNK: And new line in-between methods.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2019,6 +2019,73 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+            using (var client = new TestClient(new JsonHubProtocol()))
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler);
+                await client.Connected.OrTimeout();
+                // This is a fake client -- it doesn't auto-ping to signal
+
+                await Task.Delay(120);
+                client.TickHeartbeat();
+
+                await Assert.ThrowsAsync<TimeoutException>(() => connectionHandlerTask.OrTimeout(60));
```
JamesNK: What is being tested by this timeout? Could you add a comment

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2019,6 +2019,73 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+                   options.ClientTimeoutInterval = TimeSpan.FromMilliseconds(100)));
+            var connectionHandler = serviceProvider.GetService<HubConnectionHandler<MethodHub>>();
+
+            using (var client = new TestClient(new JsonHubProtocol()))
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler);
+                await client.Connected.OrTimeout();
+                // This is a fake client -- it doesn't auto-ping to signal
+
+                await Task.Delay(120);
```
JamesNK: I'm guessing it is important that the delay be 120 to be over the ClientTimeoutInterval? Could you add some more comments to the tests explaining what is being simulated.

DylanDmitri:
It's complaining about the refactoring of ```HubConnectionContext``` constructor.

```
// before
HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory)

// after -- added an optional parameter
HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory, TimeSpan? clientTimeoutInterval = null)
```
I want to add this occurrence to the "list of breaking changes".

More generally, there's weirdness in the ```HubConnectionContext``` constructor, because some logic (coalescing default values, setup calls) happens in ```HubConnectionHandler```. Maybe refactor the constructor?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
anurse: Kestrel actually uses `DateTime.UtcNow`, as does the client timeout logic. `DateTime.UtcNow` has basically the same performance, so we decided consistency was better here.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
anurse: Ah yeah, we're going to need to talk about breaking changes :).

anurse:
Yeah, we can't have a breaking change in our API. We can chat about it tomorrow, it's pretty straightforward to avoid the breaking change here.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
DylanDmitri: The order of these lines is purely visual / a matter of taste. Mostly wanted the two ping-related lines to be adjacent because it seemed nicer to read.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
DylanDmitri: checks that the client remains connected (ie that the handler task times out)
only clients that ping (new version) should be timed out

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2019,6 +2019,75 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler);
+                await client.Connected.OrTimeout();
+                // This is a fake client -- it doesn't auto-ping to signal
+
+                // We go over the 100 ms timeout interval...
+                await Task.Delay(120);
+                client.TickHeartbeat();
+
+                // but client should still be open, since it never pinged to activate the timeout checking
+                await Assert.ThrowsAsync<TimeoutException>(() => connectionHandlerTask.OrTimeout(60));
```
JamesNK: Using a timeout to test that the client is still open is weird. Is there no flag on TestClient that says the state of the connection?

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2019,6 +2019,75 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+            using (var client = new TestClient(new JsonHubProtocol()))
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler);
+                await client.Connected.OrTimeout();
+                await client.SendHubMessageAsync(PingMessage.Instance);
+
+                await Task.Delay(120);
+                client.TickHeartbeat();
+
+                await connectionHandlerTask.OrTimeout(60);
```
JamesNK: This short a timeout will cause flakiness. I would just use the default timeout unless there is a good reason not to.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2019,6 +2019,75 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+                await client.SendHubMessageAsync(PingMessage.Instance);
+
+                for (int i = 0; i < 5; i++)
+                {
+                    await Task.Delay(20);
+                    client.TickHeartbeat();
+                    await client.SendHubMessageAsync(PingMessage.Instance);
+                }
+
+                await Assert.ThrowsAsync<TimeoutException>(() => connectionHandlerTask.OrTimeout(50));
```
JamesNK: Using a timeout to test that the client is still open is weird. Is there no flag on TestClient that says the state of the connection?

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -428,24 +436,46 @@ internal Task AbortAsync()
...
             {
                 // Haven't sent a message for the entire keep-alive duration, so send a ping.
                 // If the transport channel is full, this will fail, but that's OK because
                 // adding a Ping message when the transport is full is unnecessary since the
                 // transport is still in the process of sending frames.
                 _ = TryWritePingAsync();
+            }
+        }
+
+        public void StartPeriodicallyCheckingForClientTimeout()
```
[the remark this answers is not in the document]
anurse: Also, make this `internal`. This class is `public` so that certain user components can invoke methods on it, but I don't think we want them messing with Ping frames like this.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -428,24 +436,46 @@ internal Task AbortAsync()
...
+        private void CheckClientTimeout()
+        {
+            // If it's been too long since we've heard from the client, then close this
+            if (DateTime.UtcNow.Ticks - Interlocked.Read(ref _lastReceivedTimestamp) > _clientTimeoutInterval)
+            {
+                Abort();
             }
         }
 
+        public void ResetClientTimeout()
```
anurse: Make this `internal` as well.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
anurse: Yeah, you can just check that the task has not completed here. The `Task.Delay(120)` followed by manually ticking the heartbeat means we know at least `120` milliseconds have elapsed (if not more)

```
Assert.False(connectionHandlerTask.IsCompleted)
```

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
anurse: Yep, just use `.OrTimeout()`

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
anurse: Same as above, use `Assert.False(connectionHandlerTask.IsCompleted)` to detect this. We know the `TickHeartbeat` is the thing that would shut it down, and we know it has run.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -856,11 +856,14 @@ public void ResetTimeout()
 
         private async Task TimerLoop(TimerAwaitable timer)
         {
+            // tell the server you intend to ping
```
anurse: nit: we generally use "we" to refer to what the code will do "tell the server we intend to ping"

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -464,6 +494,12 @@ private static void AbortConnection(object state)
             }
         }
 
+        private void DoneWithSend()
```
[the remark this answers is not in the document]
davidfowl: Doesn't this re-introduce this performance problem https://github.com/aspnet/SignalR/issues/1637

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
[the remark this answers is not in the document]
DylanDmitri: For the name, I was trying to imply that the *checking* for potential timeouts was starting, rather than the timeout process itself.

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -282,7 +296,7 @@ private ValueTask TryWritePingAsync()
             }
             finally
             {
-                _writeLock.Release();
+                DoneWithSend();
```
anurse: Missed renaming the call sites :). Remember that VS can do the rename for you and apply it to all the places the method is called! Just use that light bulb/Ctrl-. shortcut ;)

DylanDmitri: aaaaaaaaahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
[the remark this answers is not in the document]
anurse: Ah, so this was removed intentionally? That probably should have been documented somewhere, like in a comment...

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -428,24 +442,52 @@ internal Task AbortAsync()
...
+            // If it's been too long since we've heard from the client, then close this
+            if (DateTime.UtcNow.Ticks - Volatile.Read(ref _lastReceivedTimeStamp) > _clientTimeoutInterval)
+            {
+                Abort();
             }
         }
 
+        internal void ResetClientTimeout()
+        {
+            Volatile.Write(ref _lastReceivedTimeStamp, DateTime.UtcNow.Ticks);
```
BrennanConroy: Last update I heard, we were going to set a bool here and do the needful in the heartbeat etc.

DylanDmitri: Ideally this would happen. The problem is converting from a known time interval (15s) to some number of heartbeats. There's no clean way to see the heartbeat interval, and adding a field to the ```IHeartBeat``` interface would be a breaking change. Without a way to control the interval relative to what the client expects, the pattern is kinda stuck.

And so it's back to the original implementation, of just calling every 15s regardless.

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -2019,6 +2019,75 @@ public int Serialize(ref byte[] bytes, int offset, T value, IFormatterResolver f
...
+                await connectionHandlerTask.OrTimeout();
+            }
+        }
+
+        [Fact]
+        public async Task ReceivingMessagesPreventsConnectionTimeoutFromOccuring()
+        {
+            var serviceProvider = HubConnectionHandlerTestUtils.CreateServiceProvider(services =>
+                services.Configure<HubOptions>(options =>
+                     options.ClientTimeoutInterval = TimeSpan.FromMilliseconds(100)));
```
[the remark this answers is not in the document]
DylanDmitri: When checking this test -- turns out the changed timeout behavior (with the boolean flag) means the window needs to be more than twice the time elapsed to ensure a disconnect (i.e. this test would pass if the behavior was broken). Changed the values to a `300ms` window and a `10x100ms` delay loop, which should be resilient.

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -116,9 +117,26 @@ public Task ReceiveBytesAsync(byte[] bytes)
...
+
+        public async Task<string> ReadSentTextMessageAsync(bool ignorePings = true)
         {
             // Read a single text message from the Application Input pipe
+
+            while (true)
+            {
+                var result = await ReadSentTextMessageAsyncInner();
+
+                var receivedMessageType = (int)JObject.Parse(result).Property("type").Value;
```
anurse: You can get even fancier 😸 . `JObject` can be accessed like a dictionary: `(int)JObject.Parse(result)["type"]`

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
anurse: Don't forget this! :)

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -197,6 +200,8 @@ HubDispatcher<THub> dispatcher
 
                     if (!buffer.IsEmpty)
                     {
+                        connection._receivedMessageThisInterval = true;
```
anurse: We generally avoid exposing internal flags like this across class boundaries. Let's wrap this up in an `internal` method on `HubConnectionContext` (something like `ResetClientTimeout`)

mikaelm12:
Can you rebase this on dev so the diff doesn't look so crazy 😄 
It'll make it a lot easier to review.

anurse:
Actually it looks like this was merged again after it was rebased. I'll work with @DylanDmitri tomorrow to clean it up.

davidfowl:
glad to see this got merged, there's no performance regression right?

anurse:
Yep, @DylanDmitri @BrennanConroy and I reviewed the perf dashboard today. No regression

davidfowl:
@anurse hmm, can we also look at the microbenchmarks for writing

PR #2365, opened 2018-05-23
Collect latency values by default for benchmark runs

BrennanConroy:


BrennanConroy:
Hopefully ;)

###### (yes, tested it)

Issue #2367, opened 2018-05-24
Flakey ClientPingsMultipleTimes


JamesNK:
https://travis-ci.org/aspnet/SignalR/jobs/382984813

```
  Failed   Microsoft.AspNetCore.SignalR.Client.Tests.HubConnectionTests+Protocol.ClientPingsMultipleTimes
RUNDOTNET : error Message:  [/home/travis/.dotnet/buildtools/korebuild/2.2.0-preview1-17060/KoreBuild.proj]
   System.TimeoutException : Operation in ClientPingsMultipleTimes timed out at /_/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs:586
  Stack Trace:
     at System.Threading.Tasks.TaskExtensions.OrTimeout[T](Task`1 task, TimeSpan timeout, String memberName, String filePath, Nullable`1 lineNumber)
     at Microsoft.AspNetCore.SignalR.Client.Tests.HubConnectionTests.Protocol.ClientPingsMultipleTimes() in /_/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs:line 586
     at Microsoft.AspNetCore.SignalR.Client.Tests.HubConnectionTests.Protocol.ClientPingsMultipleTimes() in /_/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs:line 592
  --- End of stack trace from previous location where exception was thrown ---
  Application startup exception: System.NotSupportedException: Duplicate definitions of 'OverloadedMethod'. Overloading is not supported.
     at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1.DiscoverHubMethods() in /_/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs:line 417
     at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1..ctor(IServiceScopeFactory serviceScopeFactory, IHubContext`1 hubContext, IOptions`1 hubOptions, IOptions`1 globalHubOptions, ILogger`1 logger) in /_/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs:line 38
  --- End of stack trace from previous location where exception was thrown ---
[...]

PR #2368, opened 2018-05-24
Fix flaky ClientPingsMultipleTimes test

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2367


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -34,7 +34,6 @@ public partial class HubConnection
         public static readonly TimeSpan DefaultServerTimeout = TimeSpan.FromSeconds(30); // Server ping rate is 15 sec, this is 2 times that.
         public static readonly TimeSpan DefaultHandshakeTimeout = TimeSpan.FromSeconds(15);
         public static readonly TimeSpan DefaultPingInterval = TimeSpan.FromSeconds(15);
-        public static readonly TimeSpan DefaultTickRate = TimeSpan.FromSeconds(1);
```
JamesNK: This is only used internally. Changed to not be public.

PR #2384, opened 2018-05-25
PingInterval -> KeepAliveInterval

DylanDmitri:


clients/ts/signalr/src/HubConnection.ts
```diff
@@ -114,7 +114,7 @@ export class HubConnection {
         // defensively cleanup timeout in case we receive a message from the server before we finish start
         this.cleanupTimeout();
         this.resetTimeoutPeriod();
-        this.resetPingInterval();
+        this.resetkeepAliveInterval();
```
anurse: `resetKeepAliveInterval` (upper-case the `k` here)

JamesNK:
These are API breaking changes?

anurse:
Nope, this was added in 2.2

Issue #2326, opened 2018-05-20
Turn strict mode on in TypeScript client


davidfowl:
Because Anders said so. Also it'll probably find more bugs in our code.

anurse:
Yeah, I was noticing that too :). Let's do it in MQ.

PR #2388, opened 2018-05-26
 TypeScript strict mode

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2326


clients/ts/signalr-protocol-msgpack/src/MessagePackHubProtocol.ts
```diff
@@ -173,24 +185,27 @@ export class MessagePackHubProtocol implements IHubProtocol {
...
         }
 
-        const completionMessage = {
-            error: null as string,
-            headers,
-            invocationId: properties[2],
-            result: null as any,
-            type: MessageType.Completion,
-        };
+        let error: string | undefined;
```
moozzyk: Can it just be `string?`?

JamesNK: Only with properties and parameters.

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -11,14 +11,20 @@ export class ServerSentEventsTransport implements ITransport {
...
+    private url?: string;
 
-    constructor(httpClient: HttpClient, accessTokenFactory: () => string | Promise<string>, logger: ILogger, logMessageContent: boolean) {
+    public onreceive: ((data: string | ArrayBuffer) => void) | null;
+    public onclose: ((error?: Error) => void) | null;
+
+    constructor(httpClient: HttpClient, accessTokenFactory: (() => string | Promise<string>) | undefined, logger: ILogger, logMessageContent: boolean) {
         this.httpClient = httpClient;
-        this.accessTokenFactory = accessTokenFactory || (() => null);
+        this.accessTokenFactory = accessTokenFactory || (() => "");
```
davidfowl: I think this should be null and accessTokenFactory should be:

accessTokenFactory: (() => string | Promise<string> | null)

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -46,8 +52,9 @@ export class WebSocketTransport implements ITransport {
                 resolve();
             };
 
-            webSocket.onerror = (event: ErrorEvent) => {
-                reject(event.error);
+            webSocket.onerror = (event: Event) => {
+                const error = (event instanceof ErrorEvent) ? event.error : null;
+                reject(error);
```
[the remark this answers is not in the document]
JamesNK: But `Event` doesn't have an error property.

clients/ts/signalr/src/ServerSentEventsTransport.ts
JamesNK: accessTokenFactory is passed in from the end user, and the end user would never have a reason to return null from it.

The result from the factory is always checked to not be an empty string so it shouldn't matter.
Alternatively the code could be updated to not set a lambda at all.

clients/ts/signalr/src/WebSocketTransport.ts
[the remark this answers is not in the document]
davidfowl: I see.

clients/ts/signalr/src/ServerSentEventsTransport.ts
davidfowl: > Alternatively the code could be updated to not set a lambda at all.

Sure, lets do this then since we're still null checking it.

clients/ts/signalr-protocol-msgpack/src/MessagePackHubProtocol.ts
```diff
@@ -37,7 +37,19 @@ export class MessagePackHubProtocol implements IHubProtocol {
...
             logger = NullLogger.instance;
         }
-        return BinaryMessageFormat.parse(input).map((m) => this.parseMessage(m, logger));
+
+        const messages = BinaryMessageFormat.parse(input);
+
+        const hubMessages = [];
+        for (const message of messages) {
+            const parsedMessage = this.parseMessage(message, logger);
+            // Can be null for an unknown message
```
anurse: Add note that `parseMessage` already logged about this. Was about to ask for you to log this condition here ;)

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -201,11 +205,14 @@ export class HttpConnection implements IConnection {
         }
     }
 
-    private createConnectUrl(url: string, connectionId: string) {
+    private createConnectUrl(url: string, connectionId: string | null | undefined) {
```
anurse: Pretty sure parameters can have `connectionId?`

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -201,11 +205,14 @@ export class HttpConnection implements IConnection {
...
+        if (!connectionId) {
+            return url;
+        }
         return url + (url.indexOf("?") === -1 ? "?" : "&") + `id=${connectionId}`;
     }
 
-    private async createTransport(url: string, requestedTransport: HttpTransportType | ITransport, negotiateResponse: INegotiateResponse, requestedTransferFormat: TransferFormat): Promise<void> {
+    private async createTransport(url: string, requestedTransport: HttpTransportType | ITransport | undefined, negotiateResponse: INegotiateResponse, requestedTransferFormat: TransferFormat): Promise<void> {
```
anurse: `requestedTransport?`

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -246,17 +253,17 @@ export class HttpConnection implements IConnection {
     private constructTransport(transport: HttpTransportType) {
         switch (transport) {
             case HttpTransportType.WebSockets:
-                return new WebSocketTransport(this.accessTokenFactory, this.logger, this.options.logMessageContent);
+                return new WebSocketTransport(this.accessTokenFactory, this.logger, this.options.logMessageContent || false);
```
anurse: Nice, this is much clearer, glad strict mode pushes us to do that.

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -246,17 +253,17 @@ export class HttpConnection implements IConnection {
...
             case HttpTransportType.LongPolling:
-                return new LongPollingTransport(this.httpClient, this.accessTokenFactory, this.logger, this.options.logMessageContent);
+                return new LongPollingTransport(this.httpClient, this.accessTokenFactory, this.logger, this.options.logMessageContent || false);
             default:
                 throw new Error(`Unknown transport: ${transport}.`);
         }
     }
 
-    private resolveTransport(endpoint: IAvailableTransport, requestedTransport: HttpTransportType, requestedTransferFormat: TransferFormat): HttpTransportType | null {
+    private resolveTransport(endpoint: IAvailableTransport, requestedTransport: HttpTransportType | undefined, requestedTransferFormat: TransferFormat): HttpTransportType | null {
```
anurse: `requestedTransport?`

anurse: If TypeScript complains about the argument ordering, then we should be using `null` anyway, or reorder the arguments. You aren't really supposed to use `undefined` like a value. Syntax like `this.resolveTransport(foo, undefined, bar)` is generally considered a code smell (says the person who probably refactored this to put that pattern in place ;P)

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -11,28 +11,36 @@ import { Arg, getDataDetail, sendMessage } from "./Utils";
...
-    private pollAbort: AbortController;
+    private url?: string;
+    private pollXhr?: XMLHttpRequest;
     private running: boolean;
-    private receiving: Promise<void>;
-    private closeError: Error;
+    private receiving?: Promise<void>;
+    private closeError?: Error;
+
+    public onreceive: ((data: string | ArrayBuffer) => void) | null;
```
anurse: `onreceive?`

clients/ts/signalr/src/LongPollingTransport.ts
```diff
@@ -11,28 +11,36 @@ import { Arg, getDataDetail, sendMessage } from "./Utils";
...
+    public onclose: ((error?: Error) => void) | null;
```
anurse: `onclose?`

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -8,17 +8,23 @@ import { Arg, getDataDetail, sendMessage } from "./Utils";
 
 export class ServerSentEventsTransport implements ITransport {
     private readonly httpClient: HttpClient;
-    private readonly accessTokenFactory: () => string | Promise<string>;
+    private readonly accessTokenFactory: (() => string | Promise<string>) | undefined;
```
anurse: `accessTokenFactory?`

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -8,17 +8,23 @@ import { Arg, getDataDetail, sendMessage } from "./Utils";
...
-    private eventSource: EventSource;
-    private url: string;
+    private eventSource?: EventSource;
+    private url?: string;
 
-    constructor(httpClient: HttpClient, accessTokenFactory: () => string | Promise<string>, logger: ILogger, logMessageContent: boolean) {
+    public onreceive: ((data: string | ArrayBuffer) => void) | null;
+    public onclose: ((error?: Error) => void) | null;
+
+    constructor(httpClient: HttpClient, accessTokenFactory: (() => string | Promise<string>) | undefined, logger: ILogger, logMessageContent: boolean) {
```
anurse: `accessTokenFactory?` or use `null`?

clients/ts/signalr/src/Utils.ts
```diff
@@ -22,19 +22,19 @@ export class Arg {
 }
 
 export function getDataDetail(data: any, includeContent: boolean): string {
-    let length: string = null;
+    let detail = "";
```
anurse: 👍

clients/ts/signalr/src/Utils.ts
```diff
@@ -51,13 +51,15 @@ export function formatArrayBuffer(data: ArrayBuffer): string {
     return str.substr(0, str.length - 1);
 }
 
-export async function sendMessage(logger: ILogger, transportName: string, httpClient: HttpClient, url: string, accessTokenFactory: () => string | Promise<string>, content: string | ArrayBuffer, logMessageContent: boolean): Promise<void> {
+export async function sendMessage(logger: ILogger, transportName: string, httpClient: HttpClient, url: string, accessTokenFactory: (() => string | Promise<string>) | undefined, content: string | ArrayBuffer, logMessageContent: boolean): Promise<void> {
```
anurse: `accessTokenFactory?` or `null`?

clients/ts/signalr/src/WebSocketTransport.ts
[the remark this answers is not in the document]
anurse: Yeah, that syntax would work in JS but TypeScript will squiggle unless you actually assert the type properly.

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -165,8 +164,8 @@ describe("HttpConnection", () => {
             stop(): Promise<void> {
                 return Promise.resolve();
             },
-            onclose: undefined,
-            onreceive: undefined,
+            onclose: null,
+            onreceive: null,
```
anurse: If these were `onclose?` and `onreceive?` then you would just omit them entirely.

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -622,12 +622,12 @@ describe("HubConnection", () => {
             try {
                 connection.receiveHandshakeResponse();
 
-                hubConnection.on(null, undefined);
-                hubConnection.on(undefined, null);
-                hubConnection.on("message", null);
-                hubConnection.on("message", undefined);
-                hubConnection.on(null, () => { });
-                hubConnection.on(undefined, () => { });
+                hubConnection.on(null!, undefined!);
```
anurse: `null!` amuses me :)

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -1029,21 +1041,22 @@ class TestConnection implements IConnection {
...
     public receiveBinary(data: ArrayBuffer) {
-        this.onreceive(data);
+        this.invokeOnReceive(data);
     }
 
-    public onreceive: (data: string | ArrayBuffer) => void;
-    public onclose: (error?: Error) => void;
-    public sentData: any[];
-    public lastInvocationId: string;
+    invokeOnReceive(data: string | ArrayBuffer) {
```
anurse: We're probably turning TSLint on for tests soon, and it will get grumpy if this doesn't say `private` ;)

clients/ts/signalr/src/AbortController.ts
```diff
@@ -8,7 +8,7 @@
 // Not exported from index.
 export class AbortController implements AbortSignal {
     private isAborted: boolean = false;
-    public onabort: () => void;
+    public onabort: (() => void) | null = null;
```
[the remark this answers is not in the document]
JamesNK: I felt that on publicly exposed members we should always have an explicit value, i.e. either be set to a real function or set to null.

clients/ts/signalr/src/AbortController.ts
```diff
@@ -33,5 +33,5 @@ export interface AbortSignal {
     /** Indicates if the request has been aborted. */
     aborted: boolean;
     /** Set this to a handler that will be invoked when the request is aborted. */
-    onabort: () => void;
+    onabort: (() => void) | null;
```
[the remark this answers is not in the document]
JamesNK: I felt that on publicly exposed members we should always have an explicit value, i.e. either be set to a real function or set to null.

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -166,15 +166,27 @@ export class DefaultHttpClient extends HttpClient {
...
             xhr.open(request.method, request.url, true);
             xhr.withCredentials = true;
             xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
             // Explicitly setting the Content-Type header for React Native on Android platform.
             xhr.setRequestHeader("Content-Type", "text/plain;charset=UTF-8");
 
-            if (request.headers) {
-                Object.keys(request.headers)
-                    .forEach((header) => xhr.setRequestHeader(header, request.headers[header]));
+            const headers = request.headers;
```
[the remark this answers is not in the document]
JamesNK: Required by strict mode. Type system didn't like `request.headers` nested in the delegete.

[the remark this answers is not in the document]
anurse: Ah yeah, it's because it doesn't know when the function is actually invoked. Makes sense.

clients/ts/signalr/src/HttpConnection.ts
JamesNK: It is undefined because the property is uses can be undefined. I could coalesce to null when it is called but it is code internal to the class so I didn't see the need.

Issue #2288, opened 2018-05-16
Move C++ library source out of aspnet/SignalR


JamesNK:
Is there a way to link to all the source code from libraries that the C++ client uses?

![image](https://user-images.githubusercontent.com/303201/40144498-fd0366b6-59b2-11e8-8a95-b778c2920eee.png)

mikaelm12:
Or at least start with moving the gtest source out of our repo

davidfowl:
I 10000% agree

anurse:
Even submodules would be a step in the right direction :). Or having the build script download the gtest sources (**only** when it knows for sure it's going to run the tests)

anurse:
Propose for 2.2.0-mq

PR #2391, opened 2018-05-27
Submodule gtest

BrennanConroy:
**The only new/changed files are at the very top and bottom of the diff. The middle is all deleted gtest files.**

#2288


BrennanConroy:
>Will GitHub stop counting the submoduled files against our language breakdown

Looks like it, based off of Universe being 24% powershell and having mostly submodules

Issue #2153, opened 2018-04-26
Add ConnectionId to logging scope in .NET Client logging


anurse:
We should start a scope with the connection ID in each of `Invoke`, `Send`, `Stream`, and when invoking handlers, etc.

PR #2394, opened 2018-05-28
Add connection ID scope to HubConnection

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2153

There is also the same logging scope in HttpConnection. Should it be removed?

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
@@ -1,0 +1,70 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace Microsoft.AspNetCore.SignalR.Client.Internal
+{
+    internal class ConnectionLogScope : IReadOnlyList<KeyValuePair<string, object>>
+    {
+        private string _cachedToString;
+        private string _connectionId;
+
+        public string ConnectionId
+        {
+            get => _connectionId;
+            set
+            {
+                _cachedToString = null;
+                _connectionId = value;
+            }
+        }
+
+        public KeyValuePair<string, object> this[int index]
+        {
+            get
+            {
+                if (Count == 1 && index == 0)
+                {
+                    return new KeyValuePair<string, object>("ClientConnectionId", ConnectionId);
+                }
+
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+        }
+
+        public int Count => string.IsNullOrEmpty(ConnectionId) ? 0 : 1;
+
+        public IEnumerator<KeyValuePair<string, object>> GetEnumerator()
+        {
+            for (var i = 0; i < Count; ++i)
+            {
+                yield return this[i];
+            }
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public override string ToString()
+        {
+            if (_cachedToString == null)
+            {
+                if (!string.IsNullOrEmpty(ConnectionId))
+                {
+                    _cachedToString = string.Format(
+                        CultureInfo.InvariantCulture,
+                        "ClientConnectionId:{0}",
+                        ConnectionId);
+                }
+            }
+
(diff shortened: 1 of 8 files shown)


davidfowl:
> Unfortunately this isn't how scopes work with logging because internally it uses a ThreadStatic

Yes, they never worked on the client and they should be removed... They don't use a ThreadStatic, they use async locals (I know you mean that but better to call it out).

> An example of what a fixed version looks like:

👍

BrennanConroy:
>They don't use a ThreadStatic, they use async locals

Haha, whoops

JamesNK:
🆙 📅

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public override string ToString()
+        {
+            if (_cachedToString == null)
+            {
+                if (!string.IsNullOrEmpty(ConnectionId))
+                {
+                    _cachedToString = string.Format(
+                        CultureInfo.InvariantCulture,
+                        "ClientConnectionId:{0}",
```
davidfowl: Why not just ConnectionId

test/Microsoft.AspNetCore.SignalR.Tests.Utils/LogSinkProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Testing;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    // TestSink doesn't seem to be thread-safe :(.
```
davidfowl: It should be now.

/cc @JunTaoLuo

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
```diff
@@ -0,0 +1,70 @@
...
+            return GetEnumerator();
+        }
+
+        public override string ToString()
+        {
+            if (_cachedToString == null)
+            {
+                if (!string.IsNullOrEmpty(ConnectionId))
+                {
+                    _cachedToString = string.Format(
```
anurse: `$"Y U NO {interpolate}"`?

anurse: `FormattableString.Invariant($"ConnectionId: {ConnectionId}")` if you really need invariant culture.

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
JamesNK: ConnectionLogScope is copied and pasted from HttpConnection. ClientConnectionId is used there as well.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/LogSinkProvider.cs
JamesNK: One thing that LogSinkProvider does that TestSink probably doesn't is raise an event when an item is added for it.

I can update the comment.

Would TestSink having an event make sense? It is just used internally by aspnet team right?

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
JamesNK: Because I copy pasted from other code that no interpolate. I can update both.

anurse: Noice

src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
anurse: Hrm, I now remember why we don't use `ConnectionId` on the server. It conflicts with the HTTP `ConnectionId` hosting puts in. The client is less likely to have that problem... unless you're using the client on an ASP.NET server. Maybe `ClientConnectionId` was better in the first place. So what I'm saying is... I'm flip-flopping ;)

Issue #2155, opened 2018-04-27
Remove log for scanning connections


davidfowl:
The scanning connections log is just spam and should be removed...

PR #2397, opened 2018-05-29
Remove scanning connections log messages

JamesNK:
Fixes https://github.com/aspnet/SignalR/issues/2155

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionManager.Log.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionManager.Log.cs
@@ -25,14 +25,8 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
             private static readonly Action<ILogger, string, Exception> _connectionTimedOut =
                 LoggerMessage.Define<string>(LogLevel.Trace, new EventId(5, "ConnectionTimedOut"), "Connection {TransportConnectionId} timed out.");
 
-            private static readonly Action<ILogger, Exception> _scanningConnections =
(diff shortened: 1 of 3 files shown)


davidfowl:
What is up with this spanish OS thing.

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionManager.Log.cs
```diff
@@ -25,15 +25,9 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _connectionTimedOut =
                 LoggerMessage.Define<string>(LogLevel.Trace, new EventId(5, "ConnectionTimedOut"), "Connection {TransportConnectionId} timed out.");
 
-            private static readonly Action<ILogger, Exception> _scanningConnections =
```
davidfowl: Should we leave a comment in saying what these were?

JamesNK: I like that. Can do

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionsEventSource.cs
```diff
@@ -84,17 +75,5 @@ public void ConnectionTimedOut(string connectionId)
                 }
             }
         }
-
-        [Event(eventId: 4, Level = EventLevel.Verbose, Message = "Scanning connections.")]
-        public void ScanningConnections()
```
anurse: Put a comment here as well.

PR #2408, opened 2018-05-30
Add WebSocket and EventSource polyfills to options

BrennanConroy:
#2048


clients/ts/signalr/src/IHttpConnectionOptions.ts
```diff
@@ -38,4 +39,16 @@ export interface IHttpConnectionOptions {
      * Negotiation can only be skipped when the {@link transport} property is set to 'HttpTransportType.WebSockets'.
      */
     skipNegotiation?: boolean;
+
+    /** A constructor that can be used to create a WebSocket.
```
anurse: Put a comment above these indicating that these are for unit testing or super intrepid code spelunkers. We'll be providing automatic polyfills in Node JS

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -55,6 +56,18 @@ export class HttpConnection implements IConnection {
         options = options || {};
         options.accessTokenFactory = options.accessTokenFactory || (() => null);
         options.logMessageContent = options.logMessageContent || false;
+        options.WebSocket = options.WebSocket;
+        options.EventSource = options.EventSource;
+
+        if (typeof window !== "undefined") {
```
anurse: I have yet a third way to do this that might work better and won't need this funky `window` or `global` thing ;)

```javascript
if(typeof WebSocket !== "undefined") {
    options.WebSocket = options.WebSocket || WebSocket
}
if(typeof EventSource !== "undefined") {
    options.EventSource = options.EventSource || EventSource 
}
```

Then your tests can go back (yet again) to `global["WebSocket"] = blah`

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -578,6 +581,119 @@ describe("HttpConnection", () => {
...
             expect(() => new (HttpConnection as any)()).toThrowError("The 'url' argument is required.");
         });
+
+        it("uses global WebSocket if defined", async () => {
+            // tslint:disable-next-line:no-string-literal
+            global["WebSocket"] = class WebSocket {
+                    constructor(url: string, protocols?: string | string[]) {
+                        throw new Error("WebSocket constructor called.");
+                    }
+                };
```
anurse: indentation is funky. I'd prefer one of:

```typescript
            global["WebSocket"] = class WebSocket {
                constructor(url: string, protocols?: string | string[]) {
                    throw new Error("WebSocket constructor called.");
                }
            };
```

or:

```typescript
            global["WebSocket"] = 
                class WebSocket {
                    constructor(url: string, protocols?: string | string[]) {
                        throw new Error("WebSocket constructor called.");
                    }
                };
```

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -578,6 +581,119 @@ describe("HttpConnection", () => {
...
+                .rejects
+                .toThrow("Unable to initialize any of the available transports.");
+
+            expect(eventSourceConstructorCalled).toEqual(true);
+
+            // tslint:disable-next-line:no-string-literal
+            delete global["EventSource"];
+        });
+
+        it("uses EventSource option if passed in", async () => {
```
anurse: nit: `uses EventSource constructor from options if provided`

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -578,6 +581,119 @@ describe("HttpConnection", () => {
...
+            const connection = new HttpConnection("http://tempuri.org", options);
+
+            await expect(connection.start(TransferFormat.Text))
+                .rejects
+                .toThrow("Unable to initialize any of the available transports.");
+
+            expect(eventSourceConstructorCalled).toEqual(true);
+        });
+
+        it("uses WebSocket option if passed in", async () => {
```
anurse: `uses WebSocket constructor from options if provided`

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -4,29 +4,32 @@
...
 import { ITransport, TransferFormat } from "./ITransport";
+import { EventSourceConstructor } from "./Polyfills";
 import { Arg, getDataDetail, sendMessage } from "./Utils";
 
 export class ServerSentEventsTransport implements ITransport {
     private readonly httpClient: HttpClient;
     private readonly accessTokenFactory: () => string | Promise<string>;
     private readonly logger: ILogger;
     private readonly logMessageContent: boolean;
+    private readonly eventSourceConstructor: EventSourceConstructor;
```
JamesNK: `eventSourceConstructor?`

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -4,29 +4,32 @@
...
     private eventSource: EventSource;
     private url: string;
 
-    constructor(httpClient: HttpClient, accessTokenFactory: () => string | Promise<string>, logger: ILogger, logMessageContent: boolean) {
+    constructor(httpClient: HttpClient, accessTokenFactory: () => string | Promise<string>, logger: ILogger, logMessageContent: boolean, eventSourceConstructor: EventSourceConstructor) {
```
JamesNK: `eventSourceConstructor: EventSourceConstructor | undefined`

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -4,29 +4,32 @@
...
+        this.eventSourceConstructor = eventSourceConstructor;
     }
 
     public async connect(url: string, transferFormat: TransferFormat): Promise<void> {
         Arg.isRequired(url, "url");
         Arg.isRequired(transferFormat, "transferFormat");
         Arg.isIn(transferFormat, TransferFormat, "transferFormat");
 
-        if (typeof (EventSource) === "undefined") {
+        if (!this.eventSourceConstructor) {
```
JamesNK: You might need to assign this to a local variable before doing this test to keep the TS strictness happy on line 50.

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -3,26 +3,29 @@
...
 import { ILogger, LogLevel } from "./ILogger";
 import { ITransport, TransferFormat } from "./ITransport";
+import { WebSocketConstructor } from "./Polyfills";
 import { Arg, getDataDetail } from "./Utils";
 
 export class WebSocketTransport implements ITransport {
     private readonly logger: ILogger;
     private readonly accessTokenFactory: () => string | Promise<string>;
     private readonly logMessageContent: boolean;
+    private readonly webSocketConstructor: WebSocketConstructor;
```
JamesNK: `webSocketConstructor?`

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -3,26 +3,29 @@
...
     private webSocket: WebSocket;
 
-    constructor(accessTokenFactory: () => string | Promise<string>, logger: ILogger, logMessageContent: boolean) {
+    constructor(accessTokenFactory: () => string | Promise<string>, logger: ILogger, logMessageContent: boolean, webSocketConstructor: WebSocketConstructor) {
```
JamesNK: `webSocketConstructor: WebSocketConstructor | undefined`

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -3,26 +3,29 @@
...
+        this.webSocketConstructor = webSocketConstructor;
     }
 
     public async connect(url: string, transferFormat: TransferFormat): Promise<void> {
         Arg.isRequired(url, "url");
         Arg.isRequired(transferFormat, "transferFormat");
         Arg.isIn(transferFormat, TransferFormat, "transferFormat");
 
-        if (typeof (WebSocket) === "undefined") {
+        if (!this.webSocketConstructor) {
```
JamesNK: See previous comment about assigning to local variable

clients/ts/signalr/src/Polyfills.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+// Not exported from index
+
+export interface EventSourceConstructor {
```
davidfowl: Why is this so weird? Is this a ts/js thing? Why not just make it an interface?

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -32,7 +36,7 @@ export class ServerSentEventsTransport implements ITransport {
         Arg.isRequired(transferFormat, "transferFormat");
         Arg.isIn(transferFormat, TransferFormat, "transferFormat");
 
-        if (typeof (EventSource) === "undefined") {
+        if (!this.eventSourceConstructor || this.eventSourceConstructor === undefined) {
```
JamesNK: `!this.eventSourceConstructor` does not also cover `undefined`?

JamesNK: Why not do this check in the constructor? Then eventSourceConstructor field doesn't need to be optional.

JamesNK: Actually you could do it before the constructor is called in `HttpConnection`.

Ditto parameter for `WebSocketTransport`.

anurse: We just have to make sure the fallback behavior works the same when we do that.

BrennanConroy: Fallback works, we check the constructors in `resolveTransport` and skip them if they don't exist

BrennanConroy:

clients/ts/signalr/src/Polyfills.ts
davidfowl: So this is just describing the browser's EventSource and WS as a typescript interface?

BrennanConroy:

clients/ts/signalr/src/Polyfills.ts
anurse: Kinda, it's describing the **constructor function** as a TypeScript Type so it can be used as the type of a property. In TypeScript an interface can represent anything, including a function. This is the syntax for making an interface that describes a constructor function (and in the case of `WebSocketConstructor`, it also includes static properties).

The TypeScript built-in DOM typings already include interfaces for `WebSocket` and `EventSource` **instances** but [the way they represent the constructor function itself](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L4662) isn't usable as a type, they just declare it as a global variable with an in-line type. This is basically just taking that in-line type and making it into a named type so the global `WebSocket` constructor function can be passed in.

build/buildpipeline/windows.groovy
```diff
@@ -2,7 +2,7 @@
 
 // 'node' indicates to Jenkins that the enclosed block runs on a node that matches
 // the label 'windows-with-vs'
-simpleNode('Windows.10.Enterprise.RS3.ASPNET') {
+simpleNode('Windows.10.Amd64.EnterpriseRS3.ASPNET.Open') {
```
BrennanConroy: @pranavkm FYI in case we have other repos using this

PR #2409, opened 2018-05-30
Add test coverage to TS Client tests

BrennanConroy:
Tests now take ~30 seconds from ~8 seconds. We might consider adding a flag to enable this?

Sample of what it looks like:
codecov

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ ipch/
 *.aps
 *.opendb
 *.db
+coverage/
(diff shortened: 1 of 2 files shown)


anurse:
I think I'm actually leaning towards just making this a separate command: `npm run coverage`. Running it on the CI doesn't seem super useful yet. It's not a metric we're set up to track anywhere. It's useful to have on an ad-hoc basis for devs to use when expanding test coverage or testing new features, but unless we have a way to track the numbers I don't think it needs to be in the main build.

anurse:
It does look nice in Travis though :)

image

Issue #1804, opened 2018-03-31
Add Redis benchmark to BenchmarkServer


davidfowl:
There should be an option to run the benchmarks against redis. We'll also need to be able to point to an existing redis instance from the application.

PR #2418, opened 2018-06-01
Add Redis support to E2E benchmarks

BrennanConroy:
Part of https://github.com/aspnet/SignalR/issues/1804
The rest of the work will be infrastructure etc.

### DIFF
--- a/benchmarkapps/BenchmarkServer/Startup.cs
+++ b/benchmarkapps/BenchmarkServer/Startup.cs
@@ -4,19 +4,32 @@
 using BenchmarkServer.Hubs;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace BenchmarkServer
 {
     public class Startup
     {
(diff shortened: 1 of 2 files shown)


benchmarkapps/BenchmarkServer/Startup.cs
```diff
@@ -4,19 +4,32 @@
...
+        private readonly IConfiguration _config;
+        public Startup(IConfiguration configuration)
+        {
+            _config = configuration;
+        }
+
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddSignalR(o =>
+            var signalrBuilder = services.AddSignalR(o =>
```
davidfowl: Part of me hates that we don't have builder methods that can noop themselves when not fully configured.

anurse: You mean something like `.AddRedis(connectionString)` being a no-op when `connectionString` is `null`? I'm not opposed to the idea.

We currently have `.AddRedis()` which defaults to `localhost`, and making `.AddRedis(null)` do something different might be funky, but it's not too bad. We should make sure we log an `info` for that or something though (`Redis configuration skipped; provided connection string was null`). And this is all probably breaking, so would have to wait for 3.0

PR #2440, opened 2018-06-06
Add VSTS ci pipeline

ryanbrandenburg:


anurse:
🎉

ryanbrandenburg:
Looks like there are some [failures](https://dotnet.visualstudio.com/public/public%20Team/_build/results?buildId=699&view=ms.vss-test-web.test-result-details) here. Do they make sense to you @anurse?

BrennanConroy:
They have docker for windows and don't have linux container support I'm guessing?

ryanbrandenburg:
Not sure exactly what you mean, but I'll leave the debugging of it to you. If the result is that we need to ensure certain things are installed then 1) let's make it produce a better  error, earlier 2) I'll add demands to our queue setup and talk to our VSTS agent people about making sure they're met. Otherwise I imagine the solution is to conventionalize the tests more.

ryanbrandenburg:
I found a snippet to help us ensure node 8+:
```yml
steps:
- task: NodeTool@0
  displayName: Use Node 8.x
  inputs:
    versionSpec: 8.x
```

.vsts-pipelines/builds/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+
+phases:
+- phase: Windows
+  queue: Hosted VS2017
+  steps:
+  - checkout: self
+    clean: true
+  - script: .\build.cmd -ci
+    displayName: Run build.cmd
+  - task: NodeTool@0
```
natemcmaster: This probably needs to happen before build.cmd is executed ;)

anurse:
So, @BrennanConroy and I did some investigation and I think we've identified the issue. It looks like the VSTS agent runs inside a Docker container itself, but gives us access to the **host** Docker socket. This means we can create containers that can run alongside the agent. Our test code was presuming it was running on the docker **host** though, so it creates a Redis container, publishes the port and then just connects to `localhost:[port]` assuming it will work.

Unfortunately, that doesn't work when we're **also** in a container, we need to create a docker network, attach the VSTS container we're running in to the network, and start the Redis container on the same network. Then use the address of the Redis container when connecting. It's not going to be a trivial change :(

anurse:
Looks like we might be able to do this by just querying for the IP address of the Redis container and using that: `docker inspect --format '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' container_name_or_id`

BrennanConroy:
@anurse Review please!

anurse:
I think I figured out how to re-run it...

.vsts-pipelines/builds/ci-public.yml
```diff
@@ -0,0 +1,40 @@
+trigger:
```
anurse: Can these two files import a common template as well somehow? Just wondering if we can further reduce duplication by putting the "NodeTool" custom step stuff put in a single file.

ryanbrandenburg: That's something we can experiment with later, once we've seen a few more special cases. I'm not sure we know enough about the ways things will be different between repos to be useful yet.

BrennanConroy:
Are we good to merge?

PR #2459, opened 2018-06-08
fix #2407 by tslinting tests before running

### LINKED ISSUE #2407: Enable TSLint for tests
We should run TSLint over our unit tests. Particularly rules like [no-floating-promises](https://palantir.github.io/tslint/rules/no-floating-promises/) which ensure you properly `await` a promise. It's easy to mess that up and get weird behavior in a test.

I'd suggest doing this by adding a new script to each of `clients\ts\signalr\package.json` and `clients\ts\signalr-protocol-msgpack\package.json` that lints the tests, and then have the `test` script in `clients\ts\package.json` run those commands before running `jest`.

anurse:
I also saw some extra logging that we intend to suppress and ~~noticed a weird compilation error (might not be happening on CI because of a compiler version difference or something, but it's benign to fix)~~ nothing else. [Everything is fine. We're all fine here. How are you?](https://www.youtube.com/watch?v=KYAbFqkvzQA)


clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -351,24 +351,24 @@ describe("HttpConnection", () => {
...
-            })
-            .on("GET", (r) => {
-                if (firstPoll) {
-                    firstPoll = false;
-                    return "";
-                }
-                return new HttpResponse(204, "No Content", "");
-            })
-            .on("DELETE", (r) => new HttpResponse(202));
+            .on("POST", /negotiate$/, () => {
```
anurse: TSLint doesn't like unused parameters.

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -644,16 +644,16 @@ describe("HttpConnection", () => {
...
                     eventSourceConstructorCalled = true;
                     throw new Error("EventSource constructor called.");
                 }
             };
 
             const options: IHttpConnectionOptions = {
                 ...commonOptions,
-                EventSource: customEventSourceType as EventSourceConstructor,
-                httpClient: new TestHttpClient().on("POST", (r) => {
+                EventSource: customEventSourceType as any,
```
anurse: TS compiler was failing because our custom EventSource type defined in this test doesn't actually implement all the things in EventSource. I guess casting doesn't really satisfy it, but casting to `any` is basically like telling the TS compiler to leave us alone.

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -674,15 +676,17 @@ describe("HttpConnection", () => {
...
+                constructor(_url: string, _protocols?: string | string[]) {
                     throw new Error("WebSocket constructor called.");
                 }
-            };
+            }
 
             const options: IHttpConnectionOptions = {
                 ...commonOptions,
-                WebSocket: customWebSocketType as WebSocketConstructor,
+                WebSocket: TestWebSocket,
```
anurse: Not sure why, but this doesn't need the `as` cast.

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -643,17 +643,19 @@ describe("HttpConnection", () => {
         it("uses EventSource constructor from options if provided", async () => {
             let eventSourceConstructorCalled: boolean = false;
 
-            const customEventSourceType = class EventSource {
-                constructor(url: string, eventSourceInitDict?: EventSourceInit) {
+            class TestEventSource {
```
anurse: These were mostly a style change rather than the linter. Since the class can just be local to the function and have it's own name, there's no need to assign it to a variable.

BrennanConroy:
You forgot "no-floating-promises"

clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -643,17 +643,19 @@ describe("HttpConnection", () => {
...
+                // The "_" prefix tell TypeScript not to worry about unused parameter, but tslint doesn't like it.
+                // tslint:disable-next-line:variable-name
+                constructor(_url: string, _eventSourceInitDict: EventSourceInit) {
```
BrennanConroy: `"no-unused-variable": [true, {"ignore-pattern": "^_"}]`?

anurse:
:up: :date: added `no-floating-promises` and fixed the things. The only product changes were places we were intentionally ignoring the Promise return value (because we had no need to `await` it).

Issue #2439, opened 2018-06-06
Throwing HubException from server should bypass EnableDetailedErrors flag


BrennanConroy:
Just like ASP.NET SignalR

anurse:
Yep, that makes sense. We should do it.

MiszczProgramowania:
I have better idea... implement possibility to extend/overide some functionalities of exceptions parsing  and handling.

It will open possibility for implementation of:
Throwing HubException from server should bypass EnableDetailedErrors flag
and a at least hundred of other strategies...

davidfowl:
I don’t think we’re going to be providing any extensibility here. We’re going to do what ASP.NEt SignalR does and see if satisfies 90% of users. We have no desire for pre-mature extensibility that a majority of users don’t need.

mikaelm12:
Done here: c7ebae47ea3569183a27c1edd745d79da7c8f159

Issue #2437, opened 2018-06-06
Throwing HubException on server and getting the error message on typescript client


IonutArhire:
Here is the back-end method I'm invoking:

```
public async Task SendResources(string gameKey) {
      throw new HubException("oops");
      ...
}
```
In my typescript client I have:

```
public connected(data): void {    
  this._hubConnection.invoke('SendResources', this._gameKey).catch(this.errorHandler);
}

public errorHandler(err): void {
  console.log(err);
}
```
The execution flow enters 'errorHandler' with 'err' being of type Error. The problem is that I can't access the message of the exception. All info I get from the 'err' variable is this:

![capture](https://user-images.githubusercontent.com/26301880/41033716-2a5d3b04-6990-11e8-8def-53f696a2e881.PNG)

I followed what the docs said: https://docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-2.1#handle-errors

Any Idea how to handle error messages correctly with signalR?

Versions:
.net core: 2.1
angular: 5
@aspnet/signalr: 1.0.0

BrennanConroy:
Error messages aren't propagated to the client by default as they can have sensitive information. You can set the `EnableDetailedErrors` property to true when setting up SignalR and that will allow the messages through.

I've opened https://github.com/aspnet/SignalR/issues/2439 because we probably will want `HubException` to ignore the `EnableDetailedErrors` flag.

anurse:
I'm going to close this since @BrennanConroy opened a new issue for the actionable work here.

mikaelm12:
This is resolved here c7ebae47ea3569183a27c1edd745d79da7c8f159

PR #2461, opened 2018-06-09
Show detailed error message if exception is of type HubException

mikaelm12:
This PR  enables showing a detailed error message on the client regardless of the state of the `EnabledDetailedErrors `flag  if a `HubException `was thrown on the server side. 
This is the functionality in ASP.NET SignalR.

Issue: https://github.com/aspnet/SignalR/issues/2439

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/ErrorMessageHelper.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/ErrorMessageHelper.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
         {
             if (exception is HubException || includeExceptionDetails)
             {
-                return message + $" {exception.GetType().Name}: {exception.Message}";
+                return $"{message} {exception.GetType().Name}: {exception.Message}";
 
             }
 
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/ErrorMessageHelper.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/ErrorMessageHelper.cs
@@ -9,12 +9,13 @@ namespace Microsoft.AspNetCore.SignalR.Internal
     {
         internal static string BuildErrorMessage(string message, Exception exception, bool includeExceptionDetails)
         {
-            if (!includeExceptionDetails)
+            if (exception is HubException || includeExceptionDetails)
             {
-                return message;
+                return message + $" {exception.GetType().Name}: {exception.Message}";
+
             }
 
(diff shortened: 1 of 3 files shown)


mikaelm12:
Helps avoid issues like: https://github.com/aspnet/SignalR/issues/2437 for context

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -540,6 +540,35 @@ public void FailsToLoadInvalidTypedHubClient()
...
+
+            var connectionHandler = serviceProvider.GetService<HubConnectionHandler<MethodHub>>();
+
+            using (var client = new TestClient())
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler);
+
+                var message = await client.InvokeAsync(methodName).OrTimeout();
+
+                {
```
BrennanConroy: ?

src/Microsoft.AspNetCore.SignalR.Core/Internal/ErrorMessageHelper.cs
```diff
@@ -9,12 +9,13 @@ internal static class ErrorMessageHelper
     {
         internal static string BuildErrorMessage(string message, Exception exception, bool includeExceptionDetails)
         {
-            if (!includeExceptionDetails)
+            if (exception is HubException || includeExceptionDetails)
             {
-                return message;
+                return message + $" {exception.GetType().Name}: {exception.Message}";
```
BrennanConroy: nit: `message` can go in the string interpolation too

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
```diff
@@ -540,6 +540,35 @@ public void FailsToLoadInvalidTypedHubClient()
             }
         }
 
+        [Fact]
+        public async Task DetailedExceptionEvenWhenNotExplicitlySet()
+        {
+            var serviceProvider = HubConnectionHandlerTestUtils.CreateServiceProvider(builder =>
+            {
+                builder.AddSignalR();
```
BrennanConroy: remove, this is already called in `CreateServiceProvider`

test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs
mikaelm12: These used to be an if/else block here lol

mikaelm12:
Updated

davidfowl:
👍

PR #2469, opened 2018-06-11
Add build definition for deploying samples

anurse:
Adds a build script that deploys the samples and functional tests to an
Azure App Service. The specific App Service name is parameterized and
kept internal.

This is also where some of our Browser Functional Test automation (via SauceLabs) and Xamarin automation will live, but that's not ready yet.


anurse:
What's going on with VSTS cancelling the builds randomly?

Issue #2447, opened 2018-06-07
Flaky Test: CanStartConnectionFromClosedEvent


BrennanConroy:
```
Failed   Microsoft.AspNetCore.SignalR.Client.FunctionalTests.HubConnectionTests.CanStartConnectionFromClosedEvent(protocolName: "json", transportType: WebSockets, path: "/default")
RUNDOTNET : error Message:  [/home/travis/.dotnet/buildtools/korebuild/2.2.0-preview1-17067/KoreBuild.proj]
   System.Exception : 2 error(s) logged.
  Microsoft.AspNetCore.SignalR.Client.HubConnection - ErrorDuringClosedEvent - An exception was thrown in the handler for the Closed event.
  ===================
  System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
  Parameter name: index
     at Microsoft.AspNetCore.SignalR.Client.Internal.ConnectionLogScope.get_Item(Int32 index) in /_/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs:line 38
     at Microsoft.AspNetCore.SignalR.Client.Internal.ConnectionLogScope.GetEnumerator()+MoveNext() in /_/src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs:line 46
     at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
     at Microsoft.AspNetCore.SignalR.Tests.WrappingLoggerFactory.WrappingLogger.<>c__DisplayClass6_0.<GetConnectionId>b__0(Object scope, Object s) in /_/test/Microsoft.AspNetCore.SignalR.Tests.Utils/WrappingLoggerFactory.cs:line 107
     at Microsoft.Extensions.Logging.LoggerExternalScopeProvider.ForEachScope[TState](Action`2 callback, TState state)
     at Microsoft.AspNetCore.SignalR.Tests.WrappingLoggerFactory.WrappingLogger.GetConnectionId() in /_/test/Microsoft.AspNetCore.SignalR.Tests.Utils/WrappingLoggerFactory.cs:line 113
[...]

PR #2470, opened 2018-06-11
Return empty string for connectionId in some cases

BrennanConroy:
#2447


src/Microsoft.AspNetCore.SignalR.Client.Core/Internal/ConnectionLogScope.cs
```diff
@@ -30,7 +30,7 @@ public string ConnectionId
         {
             get
             {
-                if (Count == 1 && index == 0)
+                if (index == 0)
```
anurse: I think this is fine. The only case where you actually get `null` here is if the value is `null`ed out in between when the consumer calls `.Count` and when the consumer dereferences the indexer.

Issue #2405, opened 2018-05-30
Expand Unit Tests using mocks


anurse:
Prerequisite #2048 

Expand our unit tests using the mocks added in #2048.

BrennanConroy:
Unblocked

PR #2484, opened 2018-06-13
Add JS unit tests for SSE

BrennanConroy:
#2405 
This is just for SSE. WebSockets will be a different PR.

Pay special attention to the actual SSE transport, I changed a couple things because they were wrong

### DIFF
--- a/clients/ts/signalr/src/ServerSentEventsTransport.ts
+++ b/clients/ts/signalr/src/ServerSentEventsTransport.ts
@@ -38,6 +38,7 @@ export class ServerSentEventsTransport implements ITransport {
 
         this.logger.log(LogLevel.Trace, "(SSE transport) Connecting");
 
+        // set url before accessTokenFactory because this.url is only for send and we set the auth header instead of the query string for send
         this.url = url;
 
         if (this.accessTokenFactory) {
--- a/clients/ts/signalr/src/ServerSentEventsTransport.ts
+++ b/clients/ts/signalr/src/ServerSentEventsTransport.ts
@@ -38,6 +38,8 @@ export class ServerSentEventsTransport implements ITransport {
 
         this.logger.log(LogLevel.Trace, "(SSE transport) Connecting");
 
+        this.url = url;
+
         if (this.accessTokenFactory) {
             const token = await this.accessTokenFactory();
             if (token) {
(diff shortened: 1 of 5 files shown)


clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -38,18 +38,20 @@ export class ServerSentEventsTransport implements ITransport {
 
         this.logger.log(LogLevel.Trace, "(SSE transport) Connecting");
 
+        this.url = url;
```
anurse: Can you put a comment in about why we do this before updating the URL with the query string? (Because `this.url` is only used by the send, which can set headers and thus doesn't need the query string goop)

clients/ts/signalr/tests/ServerSentEventsTransport.test.ts
```diff
@@ -0,0 +1,177 @@
...
+import { TestHttpClient } from "./TestHttpClient";
+
+describe("ServerSentEventsTransport", () => {
+    it("does not allow non-text formats", async () => {
+        await VerifyLogger.run(async (logger) => {
+            const sse = new ServerSentEventsTransport(new TestHttpClient(), undefined, logger, true, TestEventSource);
+
+            await expect(sse.connect("", TransferFormat.Binary))
+                .rejects
+                .toMatchObject({ message: "The Server-Sent Events transport only supports the 'Text' transfer format" });
```
anurse: I think `.toEqual(new Error("..."))` will work; according to a quick test I did in node:

```
> expect(new Error("foo")).toEqual(new Error("bar"))
Error: expect(received).toEqual(expected)

Expected value to equal:
  [Error: bar]
Received:
  [Error: foo]

Difference:

- Expected
+ Received

- [Error: bar]
+ [Error: foo]
> expect(new Error("foo")).toEqual(new Error("foo"))
undefined
```

clients/ts/signalr/tests/ServerSentEventsTransport.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        });
+    });
+
+    it("connect waits for EventSource to be connected", async () => {
+        await VerifyLogger.run(async (logger) => {
+            const sse = new ServerSentEventsTransport(new TestHttpClient(), undefined, logger, true, TestEventSource);
+
+            let connectPromise = sse.connect("http://example.com", TransferFormat.Text);
```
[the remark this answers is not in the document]
BrennanConroy: It's reset at line 32 ;)

BrennanConroy:
🆙 📅

Issue #2482, opened 2018-06-13
Java Client Work Plan


mikaelm12:
## Getting Started

- [ ]  Decide on Maven or Gradle
- [ ] Get an empty project into the repo under clients/java
- [ ] HubConnection - (not going to worry about HttpConnection for now)
    - [ ] 	Handshake - send hardcoded request
    - [ ] 	Only going to support JSON protocol for now
    - [ ] 	Basic Sending Invocations, no streaming to start
    - [ ] 	Register `On` method handlers to HubConnection
    - [ ] 	Stop HubConnection
- [ ]  Basic "Happy Path" tests
- [ ] Transports - Just websockets for now

## Next Steps

- [ ]  Get this integrated with our build infrastructure
- [ ] More thorough testing - End to End + unit testing 
- [ ] HttpConnection 
- [ ] InvokeAsync
- [ ] Ensure functionality with SignalR Service
- [ ] Configuration
- [ ] Streaming
- [ ] SSE
- [ ]  LongPolling
- [ ]  Transport Fallback

mikaelm12:
cc @anurse @BrennanConroy @muratg

anurse:
Looks good. Can you file an issue for Getting Started, we can create issues for the Next Steps after we start iterating.

PR #2485, opened 2018-06-13
Add Empty Java Client Project

mikaelm12:
Getting started on the Java Client work list.
Here: https://github.com/aspnet/SignalR/issues/2482

Let's just check something in to begin


.gitignore
```diff
@@ -28,6 +28,11 @@ nuget.exe
 project.lock.json
 runtimes/
 .build/
+#Gradle
```
anurse: Spacing :). And put them at the bottom.

```
# Gradle files
...
```

clients/java/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -0,0 +1,8 @@
+auto.sync=false
+build.scans.enabled=false
+connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)
+connection.project.dir=
+eclipse.preferences.version=1
+gradle.user.home=
+offline.mode=false
+override.workspace.settings=true
```
anurse: I think we shouldn't check this file in. When I opened the project in VS Code it modified the file. Seems like it might be a per-user thing...

clients/java/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+    api 'org.apache.commons:commons-math3:3.6.1'
+
+    // This dependency is used internally, and not exposed to consumers on their own compile classpath.
+    implementation 'com.google.guava:guava:23.0'
+
+    // Use JUnit test framework
+    testImplementation 'junit:junit:4.12'
+
+    //Java JSON Library 
+    implementation 'com.google.code.gson:gson:2.8.5'
```
anurse: Are these dependencies required to build or just sample dependencies that come with the template? It looks like the latter to me (`commons-math`, `guava`, `gson` don't seem to be used)

clients/java/.classpath
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
anurse: I think these can be auto-generated by a Gradle plugin: https://docs.gradle.org/current/userguide/eclipse_plugin.html

Maybe we should configure that and exclude this and the `.project` file from Git

clients/java/build.gradle
mikaelm12: I added this one and forgot to remove it. The rest came out of the box. Commented them out and it still builds successfully 🤷‍♂️

mikaelm12: Except JUnit of course

BrennanConroy:
Email Eilon about dependencies

clients/java/src/main/java/HubConnection.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
BrennanConroy: `clients/java/src/main/java/filename.java` seems a bit much, maybe `clients/java/src/filename.java`?

clients/ts/package-lock.json
```diff
@@ -1029,7 +1029,6 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
-        "fsevents": "1.2.2",
```
BrennanConroy: probably want to revert this change?

clients/java/src/main/java/HubConnection.java
mikaelm12: `src/main/java` is the standard convention for Java projects

mikaelm12:
Opening a new PR

PR #2491, opened 2018-06-14
Bind the IObservable subscription in the AsChannelReader extension method to the connection livetime

### LINKED ISSUE #2456: The AsChannelReader sample doesn't handle the case when the connection is closed
I'm using the Git 1.0.0 tag, run build.cmd and then run the `samples\SignalRSamples` application.

In a Browser, I choose the streaming examples, `Connect` and then choose `From Observable`.

When the observable completes, the channel properly completes and the observable's subscription is disposed. However, when I `Disconnect` before the observable completes, the channel in turn doesn't ever complete and the subscription remains active. Unbounded channels then effectively become a memory leak.

The latter is what I'm seeing in my application, where the observables don't ever complete by design and browsers just connect and disconnect to a Hub.

Tragetaschen:
These are the changes I made to the extension method in my project to bind the observable subscription to the connection livetime.

Fixes #2456

### DIFF
--- a/samples/SignalRSamples/Hubs/Streaming.cs
+++ b/samples/SignalRSamples/Hubs/Streaming.cs
@@ -17,7 +17,7 @@ namespace SignalRSamples.Hubs
                              .Select((_, index) => index)
                              .Take(count);
 
-            return observable.AsChannelReader();
+            return observable.AsChannelReader(Context.ConnectionAborted);
         }
 
         public ChannelReader<int> ChannelCounter(int count, int delay)
--- a/samples/SignalRSamples/ObservableExtensions.cs
+++ b/samples/SignalRSamples/ObservableExtensions.cs
@@ -3,13 +3,18 @@
 
 using System;
 using System.Reactive.Linq;
+using System.Threading;
 using System.Threading.Channels;
 
 namespace SignalRSamples
 {
     public static class ObservableExtensions
     {
-        public static ChannelReader<T> AsChannelReader<T>(this IObservable<T> observable, int? maxBufferSize = null)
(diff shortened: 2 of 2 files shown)


samples/SignalRSamples/Hubs/Streaming.cs
```diff
@@ -17,7 +17,7 @@ public ChannelReader<int> ObservableCounter(int count, int delay)
                              .Select((_, index) => index)
                              .Take(count);
 
-            return observable.AsChannelReader();
+            return observable.AsChannelReader(Context);
```
davidfowl: Pass the token in directly

samples/SignalRSamples/ObservableExtensions.cs
```diff
@@ -9,7 +10,11 @@ namespace SignalRSamples
 {
     public static class ObservableExtensions
     {
-        public static ChannelReader<T> AsChannelReader<T>(this IObservable<T> observable, int? maxBufferSize = null)
+        public static ChannelReader<T> AsChannelReader<T>(
+            this IObservable<T> observable,
+            HubCallerContext context,
```
davidfowl: Accept a cancellation token directly

samples/SignalRSamples/ObservableExtensions.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.SignalR;
```
BrennanConroy: nit: sort

PR #2495, opened 2018-06-14
Add JS unit tests for WebSockets

BrennanConroy:
#2405

### DIFF
--- a/clients/ts/signalr/tests/ServerSentEventsTransport.test.ts
+++ b/clients/ts/signalr/tests/ServerSentEventsTransport.test.ts
@@ -33,12 +33,12 @@ describe("ServerSentEventsTransport", () => {
 
             await TestEventSource.eventSource.openSet;
 
-            expect(connectComplete).toEqual(false);
+            expect(connectComplete).toBe(false);
 
             TestEventSource.eventSource.onopen(new TestMessageEvent());
 
             await connectPromise;
-            expect(connectComplete).toEqual(true);
+            expect(connectComplete).toBe(true);
         });
     });
 
@@ -49,7 +49,7 @@ describe("ServerSentEventsTransport", () => {
                 await VerifyLogger.run(async (logger) => {
                     await createAndStartSSE(logger, input, () => "secretToken");
 
-                    expect(TestEventSource.eventSource.url).toEqual(expected);
+                    expect(TestEventSource.eventSource.url).toBe(expected);
                 });
             });
         });
@@ -66,8 +66,8 @@ describe("ServerSentEventsTransport", () => {
 
             await sse.send("");
 
-            expect(request!.headers!.Authorization).toEqual("Bearer secretToken");
-            expect(request!.url).toEqual("http://example.com");
+            expect(request!.headers!.Authorization).toBe("Bearer secretToken");
+            expect(request!.url).toBe("http://example.com");
         });
     });
 
@@ -83,7 +83,7 @@ describe("ServerSentEventsTransport", () => {
 
             await sse.send("send data");
 
-            expect(request!.content).toEqual("send data");
+            expect(request!.content).toBe("send data");
         });
     });
 
@@ -100,8 +100,8 @@ describe("ServerSentEventsTransport", () => {
             message.data = "receive data";
             TestEventSource.eventSource.onmessage(message);
 
-            expect(typeof received!).toEqual("string");
-            expect(received!).toEqual("receive data");
+            expect(typeof received!).toBe("string");
+            expect(received!).toBe("receive data");
         });
     });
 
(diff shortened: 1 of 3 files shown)


clients/ts/signalr/tests/WebSocketTransport.test.ts
```diff
@@ -0,0 +1,203 @@
...
+import { WebSocketTransport } from "../src/WebSocketTransport";
+import { VerifyLogger } from "./Common";
+import { TestMessageEvent } from "./TestEventSource";
+import { TestCloseEvent, TestEvent, TestWebSocket } from "./TestWebSocket";
+
+describe("WebSocketTransport", () => {
+    [[TransferFormat.Text, "blob"],
+    [TransferFormat.Binary, "arraybuffer"]]
+        .forEach(([input, expected]) => {
+            it(`sets websocket binarytype based on transferformat: ${TransferFormat[input]}`, async () => {
```
anurse: The logic isn't to set the binary type to `blob` when sending `Text` and to `arraybuffer` when sending `Binary`. It's actually just that we always want `binaryType` to be `arraybuffer` because that's how we send binary frames. However, we only care about setting it when the `TransferFormat` is `Binary` and we'll never send any binary frames when it's `Text` so we don't bother setting it.

What I'm saying is, I don't think we need to test that the `binaryType` value is `blob` when using `TransferMode.Text`.

clients/ts/signalr/tests/WebSocketTransport.test.ts
```diff
@@ -0,0 +1,203 @@
...
+
+        TestWebSocket.webSocket.onerror(new TestEvent());
+
+        await expect(connectPromise)
+            .rejects;
+        expect(connectComplete).toEqual(false);
+    });
+
+    [["http://example.com", "ws://example.com?access_token=secretToken"],
+    ["http://example.com?value=null", "ws://example.com?value=null&access_token=secretToken"]]
```
anurse: Throw some tests in that don't use an access_token to also verify the conversion from `http(s)` to `ws(s)`

anurse: Maybe two separate "theories" (one with an access_token and one without) so the names don't get too funky:

* `generates correct WebSocket URL for ${input}`
* `generates correct WebSocket URL for ${input} with access_token`

clients/ts/signalr/tests/WebSocketTransport.test.ts
```diff
@@ -0,0 +1,203 @@
...
+            };
+
+            const message = new TestCloseEvent();
+            message.wasClean = true;
+            message.code = 1000;
+            message.reason = "success";
+            TestWebSocket.webSocket.onclose(message);
+
+            expect(closeCalled).toEqual(true);
+            expect(error!).toEqual(undefined);
```
anurse: `.toBeUndefined()`

clients/ts/signalr/tests/WebSocketTransport.test.ts
```diff
@@ -0,0 +1,203 @@
...
+                    const webSocket = new WebSocketTransport(undefined, logger, true, TestWebSocket);
+
+                    const connectPromise = webSocket.connect("http://example.com", input as TransferFormat);
+
+                    await TestWebSocket.webSocket.openSet;
+                    TestWebSocket.webSocket.onopen(new TestEvent());
+
+                    await connectPromise;
+
+                    expect(TestWebSocket.webSocket.binaryType).toEqual(expected);
```
anurse: When checking primitives, we should use `.toBe`. It's equivalent to `===`. I forgot to note that on the SSE version...

anurse: Just do a scan through this and the SSE file for `.toEqual` and replace it with `.toBe` where it looks appropriate (primitive values)

clients/ts/signalr/tests/WebSocketTransport.test.ts
```diff
@@ -0,0 +1,203 @@
...
+            expect(closeCalled).toEqual(true);
+            expect(error!).toEqual(undefined);
+
+            await expect(webSocket.send(""))
+                .rejects
+                .toThrow("WebSocket is not in the OPEN state");
+        });
+    });
+
+    it("can send data", async () => {
```
anurse: add one for binary data

anurse: really just means sending a `Uint8Array` instead of a string and watching it go through

clients/ts/signalr/tests/ServerSentEventsTransport.test.ts
```diff
@@ -33,12 +33,12 @@ describe("ServerSentEventsTransport", () => {
 
             await TestEventSource.eventSource.openSet;
 
-            expect(connectComplete).toEqual(false);
+            expect(connectComplete).toBe(false);
```
anurse: To be, or not to be: that is the question: Whether 'tis nobler in the mind to suffer The slings and arrows of outrageous fortune, Or to take arms against a sea of troubles, And by opposing end them?

PR #2496, opened 2018-06-14
Switch UMD build to WebPack

anurse:
Switch to WebPack. It works better on later versions of node (or it should; we'll see with the CI :))

* Functional Tests use the UMD build (which is what changed as a result of this) and still all pass
* Size is similar:
  * With Rollup:
    * signalr.js - 131KB
    * signalr.min.js - 68KB
    * signalr-protocol-msgpack.js - 74KB
    * signalr-protocol-msgpack.min.js - 48KB
  * Web WebPack:
    * signalr.js - 163KB **See Note**
    * signalr.min.js - 87KB
    * signalr-protocol-msgpack.js - 76KB
    * signalr-protocol-msgpack.min.js - 48KB
* Note: WebPack brings in a couple more polyfills to make things more resilient, also it brings in slightly larger module-loading goop. Hence the larger file size for `signalr.js` (since that uses a bunch of modules).


clients/ts/FunctionalTests/webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+    resolve: {
+        extensions: [".ts", ".js"]
+    },
+    output: {
+        filename: 'signalr-functional-tests.js',
+        path: path.resolve(__dirname, "wwwroot", "dist"),
+    },
+    externals: {
+        "@aspnet/signalr": "signalR",
+        "@aspnet/signalr-protocols-msgpack": "signalR.protocols.msgpack",
```
BrennanConroy: `@aspnet/signalr-protocol-msgpack`

clients/ts/signalr-protocol-msgpack/webpack.config.js
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+const path = require("path");
+const baseConfig = require("../webpack.config.base");
+module.exports = baseConfig(__dirname, "signalr-protocol-msgpack", {
+    msgpack5: "msgpack5",
```
BrennanConroy: string?

anurse: meh

anurse:
Only failure is the Spanish build, which isn't related to this PR (but will be my next focus ;)

davidfowl:
Maybe we should try https://parceljs.org

anurse:
I did a spike after I learned about it (which might have been right around when you did ;)) and it was a fairly minimal change but we could consider it. WebPack has a big advantage over Parcel and Rollup: It's what literally everyone else uses. That's not always a good enough reason, but with all else (relatively) equal, it's a pretty nice feature. There's a lot more support, guidance, etc. out there for WebPack.

I spent almost a week getting Rollup working right for us, digging through code because there weren't enough guides for our specific scenarios. With WebPack it took me one day ;).

PR #2499, opened 2018-06-15
Add Empty Java Project

mikaelm12:
This one is slightly cleaner


clients/java/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id 'java'
+}
+
+group 'signalr'
```
anurse: We need to decide on our groupId (a "namespace" concept in Maven). Azure SDKs are under `com.microsoft.azure`, and there are similar org-specific hives under `com.microsoft`: https://mvnrepository.com/artifact/com.microsoft . I think `com.microsoft.aspnet` or `com.microsoft.aspnetcore` would work.

Also, it looks like the artifactId (Maven's packageID equivalent, but scoped within the groupId) is taken from the folder name by default, meaning this one will be `java`, so our full package ID right now is `signalr:java`. Not sure that's what we want ;P. Since we probably need samples too, we could create a folder structure like: `clients/java/signalr` (for the client) and `clients/java/signalr-client-sample` for a sample client?

MyGet supports publishing Maven packages, so we can do some test publishing to a rando MyGet feed to make sure we're happy with all those bits and bobs.

clients/java/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+version '0.1.0'
+
+sourceCompatibility = 1.8
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testCompile group: 'junit', name: 'junit', version: '4.12'
```
anurse: Looks like `testCompile` has been deprecated in favor of `testImplementation`: https://docs.gradle.org/current/userguide/java_plugin.html#tab:configurations

clients/java/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id 'java'
```
anurse: For others reviewing this. Docs on the "java" plugin: https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_project_layout

.gitignore
```diff
@@ -56,3 +56,9 @@ ipch/
 *.opendb
 *.db
 coverage/
+# Gradle Files
++.gradletasknamecache
++.gradle/
++build/
++bin/
++.settings/
```
anurse: The output directory appears to be called `out`. You've got the `.class` files in the commit right now :)

anurse: You can put `.gitignore` files at any level to apply their patterns only to files below that directory. I wonder if we should move these (and add `out/`) to a `gitignore` in `clients/java`?

BrennanConroy:
Remove gradlewrapper.jar and the *.class files?

mikaelm12:
I just put our group id as `com.microsoft.aspnetcore`

mikaelm12:
Would like to get this in quickly to avoid having to branch off of this branch to continue working on it 😄

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
anurse: Is this the way this is usually done in JUnit? C# has an equivalent feature (`using static`) that we generally avoid using and I'd initially suggest we align with that. But if this is how JUnit tests are generally written, I'm fine with that :)

anurse:
I think the gradle-wrapper.jar is required:

```
ps1› ./gradlew
Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain
```

It sounds like it **is** expected that you'll check-in the `gradle-wrapper.jar` file:

>Note: To make the Wrapper files available to other developers and execution environments you’ll need to check them into version control. All Wrapper files including the JAR file are very small in size. Adding the JAR file to version control is expected. Some organizations do not allow projects to submit binary files to version control. At the moment there are no alternative options to the approach.

https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper

I'm OK with doing that. It's not ideal, but it's apparently The Gradle Way(TM)

clients/java/signalr/src/test/java/HubConnectionTest.java
jkotalik: https://www.tutorialspoint.com/junit/junit_test_framework.htm seems to be fairly standard.

PR #2501, opened 2018-06-15
Corrige los fallos cuando las pruebas se ejecutan en máquinas que usan castellano como lenguaje.

### LINKED ISSUE #1967: Run build and tests on Spanish image
Consider running SignalR build and tests on a Spanish image to catch culture related bugs.

Roslyn already does this - https://github.com/aspnet/SignalR/issues/1905#issuecomment-380648062

// @jaredpar @pranavkm

anurse:
... I hope Google Translate didn't make me look too stupid there ;).

This should fix the issues with the Spanish build. Just a few places where we hard-coded asserts of things that end up getting localized.

Fixes #1967

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
@@ -508,11 +508,9 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
                 try
                 {
                     var startTask = hubConnection.StartAsync(cts.Token);
-                    var exception = await Assert.ThrowsAnyAsync<OperationCanceledException>(() => startTask.OrTimeout());
+                    await Assert.ThrowsAnyAsync<OperationCanceledException>(() => startTask.OrTimeout());
 
-                    // REVIEW: Alternatively, we could just not assert the message...
-                    // We need the localized version of the default message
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -509,7 +509,10 @@ Task<ConnectionContext> ConnectionFactory(TransferFormat format)
                 {
                     var startTask = hubConnection.StartAsync(cts.Token);
                     var exception = await Assert.ThrowsAnyAsync<OperationCanceledException>(() => startTask.OrTimeout());
-                    Assert.Equal("The operation was canceled.", exception.Message);
+
+                    // REVIEW: Alternatively, we could just not assert the message...
```
anurse: Thoughts on this? If we don't care about the exact message, we should just not assert it. If we care that it's the "Default" message, we should use the constructor to get the actual string so it's localized.

mikaelm12: Yeah, I think in this case the exact message is not important. This seems like a fine fix to me

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ExceptionMessageExtensions.cs
```diff
@@ -0,0 +1,22 @@
+using System;
```
mikaelm12: Header  😄

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.ConnectionLifecycle.cs
BrennanConroy: I think we don't need this message

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ExceptionMessageExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    public static class ExceptionMessageExtensions
+    {
+        public static string GetLocalizationSafeMessage(this ArgumentException argex)
```
mikaelm12: nit camel case `argex` to `argEx`

anurse:
I'm probably going to change the commit message when I squash it to remove the terrible spanish btw ;)

javiercn:
@anurse @davidfowl I fixed the PR title for you :)

anurse:
Filed a bug for the flaky test

PR #2512, opened 2018-06-19
fix #2474 by adding some transitive dependencies as top-level

### LINKED ISSUE #2474: Pull up System.Threading.Tasks.Extensions and System.Runtime.CompilerServices.Unsafe
In order to fully fix #2406 we need to pull our transitive reference of `System.Threading.Tasks.Extensions` and `System.Runtime.CompilerServices.Unsafe` up to `4.5.1`. We get those through `System.IO.Pipelines` but corefx doesn't flow transitive dependencies up.

This is only **required** in the client, but it is a Common package that references `System.IO.Pipelines`. There's no reason we couldn't just reference these packages directly in the `Microsoft.AspNetCore.SignalR.Client.Core` package though, even though that package doesn't directly depend upon `System.IO.Pipelines`.

cc @davidfowl

anurse:
Fixes #2474 

This is a fix being considered for 2.1.3. The 2.1.1 patch contained fixes for these `System.*` packages to make them work properly on UWP and Xamarin. However, since we get them transitively through `System.Threading.Channels` users don't see that they need to update them, and updating our package alone is insufficient. NuGet doesn't advertise transitive updates, so they have no way of knowing they need the fix.

This change adds these dependencies (which are already present transitively) as direct dependencies so that when users update our package (which NuGet should advertise as users will be directly depending on our package), they get the fix.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/Microsoft.AspNetCore.SignalR.Client.Core.csproj
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/Microsoft.AspNetCore.SignalR.Client.Core.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Client for ASP.NET Core SignalR</Description>
@@ -21,6 +21,8 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjectionPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingPackageVersion)" />
     <PackageReference Include="System.Threading.Channels" Version="$(SystemThreadingChannelsPackageVersion)" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="$(SystemThreadingTasksExtensionsPackageVersion)" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="$(SystemRuntimeCompilerServicesUnsafePackageVersion)" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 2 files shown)


anurse:
⚠️ ⚠️ ⚠️  Note `release/2.1` is not open for merges yet ⚠️ ⚠️ ⚠️

src/Microsoft.AspNetCore.SignalR.Client.Core/Microsoft.AspNetCore.SignalR.Client.Core.csproj
```diff
@@ -21,6 +21,8 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjectionPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingPackageVersion)" />
     <PackageReference Include="System.Threading.Channels" Version="$(SystemThreadingChannelsPackageVersion)" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="$(SystemThreadingTasksExtensionsPackageVersion)" />
```
davidfowl: This shouldn't be back ported to 2.2

anurse: Agreed. It's targeting `release/2.1`

anurse:
I'm following up on some things here still. It turns out System.Threading.Tasks.Extensions is not present in Xamarin yet but was marked in-box in 4.5.1, so updating might actually make things worse for now.

anurse:
Ok, `release/2.1` is open for 2.1.3 fixes and this should be ready to go.

BrennanConroy:
> so updating might actually make things worse for now

What's the status of this?

anurse:
1. This only needs to be System.Threading.Tasks.Extensions, and it's only required for UWP apps targetting Windows 10, version 1803 as the minimum
2. Adding this reference (System.Threading.Tasks.Extensions) will now break Xamarin, since that package is now marked in-box in Xamarin but it's not actually in box.

anurse:
So... no, don't merge yet I guess

Issue #2361, opened 2018-05-23
Consider protocol "minor" versioning strategy


anurse:
For client-to-server pings, we need a way to express to the server that the client supports some optional feature so the server knows it can adjust behavior. This is just one example of possible future needs here. For now, the server only marks the client as eligible for being timed out when it has sent a single ping, and we're going to make sure the client always sends a single ping. However, I want to revisit this in 2.2 and consider a possible "minor versioning" strategy for protocols that can be expressed during the handshake.

The 2.1 RTM server was designed to ignore extra properties in the `HandshakeRequest` payload which means we can use that to add an additional `minorVersion` field to the payload which is used to detect minor protocol changes. We should consider adding something like that and then consider changing the server behavior around pings to use that.

I'm filing this because I don't want to block the client-to-server ping feature on this stuff right now.

anurse:
This is going to be needed to support client-to-server pinging.

PR #2520, opened 2018-06-20
Dygray/handshake versioning

DylanDmitri:
After negotiating handshake, server replies with a `minorVersion` field. This is the minor version of the agreed upon protocol. This is a declaration by the server, not a negotiation. The minor version field is stored on the client, and will be used to gate features that are added later (for example, does this server support streaming?).

In general, if no minor version number is given (old server) it defaults to 0, because 0 is a low number.

No adding fields to `IHubProtocol`, so made a new `IHubProtocolWithMinorVersion`. Feels bad, but oh well.

Message caching:
 - `HandshakeProtocol` used to cache a single success message (empty error) in the `public static ReadOnlyMemory<byte> SuccessHandshakeData`. Eliminating this is technically a breaking change, so instead it's marked as obsolete and hardcoded to same series of bytes the default message would've serialized to previously.

- With the addition of a `minorVersion` field, there's multiple possible return messages. Fortunately there's not many (no more than the number of protocols your server supports). Added a lookup dictionary for this caching. 

- Checked `HubConnectionContextBenchmark`; the additional dictionary overhead makes the handshake response ~2% slower (from `2.05 us` to `2.09 us`). For reference, adding caching makes the response ~20 times faster.

Fixes https://github.com/aspnet/SignalR/issues/2361

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -53,7 +53,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         // Transient state to a connection
         private ConnectionState _connectionState;
-        private int _serverProtocolMinorVersion;
+        internal int _serverProtocolMinorVersion;
 
         public event Func<Exception, Task> Closed;
 
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -53,6 +53,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         // Transient state to a connection
         private ConnectionState _connectionState;
+        private int _serverProtocolMinorVersion;
 
         public event Func<Exception, Task> Closed;
 
@@ -719,6 +720,8 @@ namespace Microsoft.AspNetCore.SignalR.Client
                                             $"Unable to complete handshake with the server due to an error: {message.Error}");
                                     }
 
+                                    _serverProtocolMinorVersion = message.MinorVersion;
+
                                     break;
                                 }
                             }
@@ -737,11 +740,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
                     }
                 }
             }
+            
+            // shutdown if we're unable to read handshake
             // Ignore HubException because we throw it when we receive a handshake response with an error
-            // And we don't need to log that the handshake failed
+            // And because we already have the error, we don't need to log that the handshake failed
             catch (Exception ex) when (!(ex is HubException))
             {
-                // shutdown if we're unable to read handshake
                 Log.ErrorReceivingHandshakeResponse(_logger, ex);
                 throw;
             }
@@ -817,7 +821,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                     finally
                     {
                         // The buffer was sliced up to where it was consumed, so we can just advance to the start.
-                        // We mark examined as buffer.End so that if we didn't receive a full frame, we'll wait for more data
+                        // We mark examined as buffer. End so that if we didn't receive a full frame, we'll wait for more data
                         // before yielding the read again.
                         connectionState.Connection.Transport.Input.AdvanceTo(buffer.Start, buffer.End);
                     }
@@ -866,7 +870,6 @@ namespace Microsoft.AspNetCore.SignalR.Client
             // There is no need to start a new task if there is no Closed event registered
             if (closed != null)
             {
-
                 // Fire-and-forget the closed event
                 _ = RunClosedEvent(closed, connectionState.CloseException);
             }
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -821,7 +821,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                     finally
                     {
                         // The buffer was sliced up to where it was consumed, so we can just advance to the start.
-                        // We mark examined as buffer. End so that if we didn't receive a full frame, we'll wait for more data
+                        // We mark examined as `buffer.End` so that if we didn't receive a full frame, we'll wait for more data
                         // before yielding the read again.
                         connectionState.Connection.Transport.Input.AdvanceTo(buffer.Start, buffer.End);
                     }
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -53,7 +53,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
         // Transient state to a connection
         private ConnectionState _connectionState;
-        internal int _serverProtocolMinorVersion;
+        private int _serverProtocolMinorVersion;
 
         public event Func<Exception, Task> Closed;
 
--- a/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Buffers;
+using System.Collections.Generic;
 using System.IO;
 using Microsoft.AspNetCore.Internal;
 using Microsoft.AspNetCore.SignalR.Internal;
@@ -17,27 +18,38 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
     {
         private const string ProtocolPropertyName = "protocol";
         private const string ProtocolVersionPropertyName = "version";
+        private const string MinorVersionName = "minorVersion";
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
         /// <summary>
         /// The serialized representation of a success handshake.
         /// </summary>
-        public static ReadOnlyMemory<byte> SuccessHandshakeData;
+        public static Dictionary<IHubProtocol, ReadOnlyMemory<byte>> messageCache = new Dictionary<IHubProtocol, ReadOnlyMemory<byte>>();
 
-        static HandshakeProtocol()
+        public static ReadOnlySpan<byte> GetCachedSuccessMessage(IHubProtocol protocol)
         {
+
+            if (messageCache.TryGetValue(protocol, out var message))
+            { 
+                return message.Span;
+            }
+
             var memoryBufferWriter = MemoryBufferWriter.Get();
             try
             {
-                WriteResponseMessage(HandshakeResponseMessage.Empty, memoryBufferWriter);
-                SuccessHandshakeData = memoryBufferWriter.ToArray();
+                WriteResponseMessage(new HandshakeResponseMessage(protocol), memoryBufferWriter);
+                messageCache.Add(protocol, memoryBufferWriter.ToArray());
             }
             finally
             {
                 MemoryBufferWriter.Return(memoryBufferWriter);
             }
+
+            messageCache.TryGetValue(protocol, out var result);
+            return result.Span;
         }
+
 
         /// <summary>
         /// Writes the serialized representation of a <see cref="HandshakeRequestMessage"/> to the specified writer.
@@ -87,6 +99,9 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                         writer.WriteValue(responseMessage.Error);
                     }
 
+                    writer.WritePropertyName(MinorVersionName);
+                    writer.WriteValue(responseMessage.MinorVersion);
+
                     writer.WriteEndObject();
                     writer.Flush();
                 }
@@ -107,6 +122,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
         /// <returns>A value that is <c>true</c> if the <see cref="HandshakeResponseMessage"/> was successfully parsed; otherwise, <c>false</c>.</returns>
         public static bool TryParseResponseMessage(ref ReadOnlySequence<byte> buffer, out HandshakeResponseMessage responseMessage)
         {
+
             if (!TextMessageParser.TryParseMessage(ref buffer, out var payload))
             {
                 responseMessage = null;
@@ -122,6 +138,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                     JsonUtils.CheckRead(reader);
                     JsonUtils.EnsureObjectStart(reader);
 
+                    int? minorVersion = null;
                     string error = null;
 
                     var completed = false;
@@ -141,6 +158,9 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                                     case ErrorPropertyName:
                                         error = JsonUtils.ReadAsString(reader, ErrorPropertyName);
                                         break;
+                                    case MinorVersionName:
+                                        minorVersion = (int)JsonUtils.ReadAsInt32(reader, MinorVersionName);
+                                        break;
                                     default:
                                         reader.Skip();
                                         break;
@@ -154,7 +174,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                         }
                     };
 
-                    responseMessage = (error != null) ? new HandshakeResponseMessage(error) : HandshakeResponseMessage.Empty;
+                    responseMessage = new HandshakeResponseMessage(minorVersion, error);
                     return true;
                 }
             }
@@ -163,6 +183,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                 Utf8BufferTextReader.Return(textReader);
             }
         }
+
 
         /// <summary>
         /// Creates a new <see cref="HandshakeRequestMessage"/> from the specified serialized representation.
--- a/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
@@ -23,12 +23,6 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
-        /// <summary>
-        /// The serialized representation of a successful handshake with no minor version value.
-        /// </summary>
-        [Obsolete("A successful handshake should include a minor version, but the value here defaults to 0.")]
-        public static ReadOnlyMemory<byte> SuccessHandshakeData = Encoding.UTF8.GetBytes("{\"minorVersion\":0}"); 
-
         private static ConcurrentDictionary<IHubProtocol, ReadOnlyMemory<byte>> _messageCache = new ConcurrentDictionary<IHubProtocol, ReadOnlyMemory<byte>>();
 
         public static ReadOnlySpan<byte> GetSuccessfulHandshake(IHubProtocol protocol)
@@ -123,7 +117,6 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
         /// <returns>A value that is <c>true</c> if the <see cref="HandshakeResponseMessage"/> was successfully parsed; otherwise, <c>false</c>.</returns>
         public static bool TryParseResponseMessage(ref ReadOnlySequence<byte> buffer, out HandshakeResponseMessage responseMessage)
         {
-
             if (!TextMessageParser.TryParseMessage(ref buffer, out var payload))
             {
                 responseMessage = null;
@@ -184,7 +177,6 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                 Utf8BufferTextReader.Return(textReader);
             }
         }
-
 
         /// <summary>
         /// Creates a new <see cref="HandshakeRequestMessage"/> from the specified serialized representation.
--- a/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Buffers;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -19,17 +20,20 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
     {
         private const string ProtocolPropertyName = "protocol";
         private const string ProtocolVersionPropertyName = "version";
-        private const string MinorVersionName = "minorVersion";
+        private const string MinorVersionPropertyName = "minorVersion";
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
+        /// <summary>
+        /// Previously used to cache the errorless return message.
+        /// Now, return messages need a minor version number, so multiple are cached in a dictionary.
+        /// </summary>
         [Obsolete()]
         public static ReadOnlyMemory<byte> SuccessHandshakeData = new byte[] { 123, 34, 109, 105, 110, 111, 114, 86, 101, 114, 115, 105, 111, 110, 34, 58, 48, 125, 30 }; 
 
+        private static ConcurrentDictionary<IHubProtocol, ReadOnlyMemory<byte>> _messageCache = new ConcurrentDictionary<IHubProtocol, ReadOnlyMemory<byte>>();
 
-        private static Dictionary<IHubProtocol, ReadOnlyMemory<byte>> _messageCache = new Dictionary<IHubProtocol, ReadOnlyMemory<byte>>();
-
-        public static ReadOnlySpan<byte> GetCachedSuccessMessage(IHubProtocol protocol)
+        public static ReadOnlySpan<byte> GetCachedSuccessMessageData(IHubProtocol protocol)
         {
             if (_messageCache.TryGetValue(protocol, out var message))
             {
@@ -40,7 +44,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
             try
             {
                 WriteResponseMessage(new HandshakeResponseMessage(protocol), memoryBufferWriter);
-                _messageCache.Add(protocol, memoryBufferWriter.ToArray());
+                _messageCache.TryAdd(protocol, memoryBufferWriter.ToArray());
             }
             finally
             {
@@ -50,7 +54,6 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
             _messageCache.TryGetValue(protocol, out var result);
             return result.Span;
         }
-
 
         /// <summary>
         /// Writes the serialized representation of a <see cref="HandshakeRequestMessage"/> to the specified writer.
@@ -100,7 +103,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                         writer.WriteValue(responseMessage.Error);
                     }
 
-                    writer.WritePropertyName(MinorVersionName);
+                    writer.WritePropertyName(MinorVersionPropertyName);
                     writer.WriteValue(responseMessage.MinorVersion);
 
                     writer.WriteEndObject();
@@ -159,8 +162,8 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                                     case ErrorPropertyName:
                                         error = JsonUtils.ReadAsString(reader, ErrorPropertyName);
                                         break;
-                                    case MinorVersionName:
-                                        minorVersion = (int)JsonUtils.ReadAsInt32(reader, MinorVersionName);
+                                    case MinorVersionPropertyName:
+                                        minorVersion = (int)JsonUtils.ReadAsInt32(reader, MinorVersionPropertyName);
                                         break;
                                     default:
                                         reader.Skip();
--- a/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
@@ -29,9 +29,8 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
 
         public static ReadOnlySpan<byte> GetCachedSuccessMessage(IHubProtocol protocol)
         {
-
             if (messageCache.TryGetValue(protocol, out var message))
-            { 
+            {
                 return message.Span;
             }
 
--- a/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                 var memoryBufferWriter = MemoryBufferWriter.Get();
                 try
                 {
-                    WriteResponseMessage(new HandshakeResponseMessage(protocol.GetMinorVersion()), memoryBufferWriter);
+                    WriteResponseMessage(new HandshakeResponseMessage(protocol.MinorVersion), memoryBufferWriter);
                     result = memoryBufferWriter.ToArray();
                     _messageCache.TryAdd(protocol, result);
                 }
--- a/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
@@ -43,8 +43,8 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
                 try
                 {
                     WriteResponseMessage(new HandshakeResponseMessage(protocol.GetMinorVersion()), memoryBufferWriter);
-                    _messageCache.TryAdd(protocol, memoryBufferWriter.ToArray());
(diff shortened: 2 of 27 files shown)


src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,28 +19,39 @@ public static class HandshakeProtocol
...
         private const string ProtocolVersionPropertyName = "version";
+        private const string MinorVersionName = "minorVersion";
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
-        /// <summary>
-        /// The serialized representation of a success handshake.
-        /// </summary>
-        public static ReadOnlyMemory<byte> SuccessHandshakeData;
+        [Obsolete()]
```
JamesNK: Add a reason to the attribute.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,28 +19,39 @@ public static class HandshakeProtocol
     {
         private const string ProtocolPropertyName = "protocol";
         private const string ProtocolVersionPropertyName = "version";
+        private const string MinorVersionName = "minorVersion";
```
JamesNK: MinorVersionPropertyName

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,28 +19,39 @@ public static class HandshakeProtocol
...
+        [Obsolete()]
+        public static ReadOnlyMemory<byte> SuccessHandshakeData = new byte[] { 123, 34, 109, 105, 110, 111, 114, 86, 101, 114, 115, 105, 111, 110, 34, 58, 48, 125, 30 }; 
+
+
+        private static Dictionary<IHubProtocol, ReadOnlyMemory<byte>> _messageCache = new Dictionary<IHubProtocol, ReadOnlyMemory<byte>>();
 
-        static HandshakeProtocol()
+        public static ReadOnlySpan<byte> GetCachedSuccessMessage(IHubProtocol protocol)
         {
+            if (_messageCache.TryGetValue(protocol, out var message))
```
JamesNK: A regular dictionary isn't threadsafe. If a message is added at the same time as one is fetched then an error can occur.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,28 +19,39 @@ public static class HandshakeProtocol
...
-        /// </summary>
-        public static ReadOnlyMemory<byte> SuccessHandshakeData;
+        [Obsolete()]
+        public static ReadOnlyMemory<byte> SuccessHandshakeData = new byte[] { 123, 34, 109, 105, 110, 111, 114, 86, 101, 114, 115, 105, 111, 110, 34, 58, 48, 125, 30 }; 
+
+
+        private static Dictionary<IHubProtocol, ReadOnlyMemory<byte>> _messageCache = new Dictionary<IHubProtocol, ReadOnlyMemory<byte>>();
 
-        static HandshakeProtocol()
+        public static ReadOnlySpan<byte> GetCachedSuccessMessage(IHubProtocol protocol)
```
JamesNK: GetCachedSuccessMessageData

You're getting the serialized data about the message, not the message itself.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,28 +19,39 @@ public static class HandshakeProtocol
...
-                SuccessHandshakeData = memoryBufferWriter.ToArray();
+                WriteResponseMessage(new HandshakeResponseMessage(protocol), memoryBufferWriter);
+                _messageCache.Add(protocol, memoryBufferWriter.ToArray());
             }
             finally
             {
                 MemoryBufferWriter.Return(memoryBufferWriter);
             }
+
+            _messageCache.TryGetValue(protocol, out var result);
```
JamesNK: Return a span from `memoryBufferWriter.ToArray()` up above. Avoids adding the data to the cache then getting it again for no reason.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
```diff
@@ -9,23 +9,42 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
     public class HandshakeResponseMessage : HubMessage
     {
         /// <summary>
+        /// Depreciated.
         /// An empty response message with no error.
         /// </summary>
-        public static readonly HandshakeResponseMessage Empty = new HandshakeResponseMessage(null);
+        public static readonly HandshakeResponseMessage Default = new HandshakeResponseMessage(error: null);
```
JamesNK: Renaming this is an API breaking change?

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
```diff
@@ -9,23 +9,42 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HandshakeResponseMessage"/> class.
+        /// If you are intending to send back an error, the minor version doesn't matter.
+        /// </summary>
+        /// <param name="error"></param>
+        public HandshakeResponseMessage(string error) : this(null, error) { }
+
+        public HandshakeResponseMessage(int minorVersion) : this(minorVersion, null) { }
+
+        public HandshakeResponseMessage(IHubProtocol protocol)
```
JamesNK: This ctor feels a little weird. IMO the response message shouldn't know about the IHubProcotol.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/IHubProtocol.cs
```diff
@@ -57,4 +57,9 @@ public interface IHubProtocol
         /// <returns>A value indicating whether the protocol supports the specified version.</returns>
         bool IsVersionSupported(int version);
     }
+
+    public interface IHubProtocolWithMinorVersion : IHubProtocol
```
JamesNK: `IHubProtocol2`? Adding a number is how interfaces are recommended to be versioned. Prevents in the future:

`IHubProtocolWithMinorVersionAndSomethingElseWeJustAddedFiveYearsFromNow`

src/Microsoft.AspNetCore.SignalR.Protocols.MessagePack/Protocol/MessagePackHubProtocol.cs
```diff
@@ -29,13 +29,17 @@ public class MessagePackHubProtocol : IHubProtocol
 
         private static readonly string ProtocolName = "messagepack";
         private static readonly int ProtocolVersion = 1;
+        private static readonly int ProtocolMinorVersion = 0;
+
```
JamesNK: nit: too liney

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -78,7 +78,7 @@ public TestConnection(Func<Task> onStart = null, Func<Task> onDispose = null, bo
             return this;
         }
 
-        public async Task<string> ReadHandshakeAndSendResponseAsync()
+        public async Task<string> ReadHandshakeAndSendResponseAsync(int minorVersion=0)
```
JamesNK: nit: `minorVersion = 0`

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
anurse: Yeah, we'll have to rename it back :'(

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
anurse: Yeah, we should use a `ConcurrentDictionary` here, or only access the dictionary from a lock.

DylanDmitri: yeah it seems this is breaking the CI tests.

Will change it to a `ConcurrentDictionary`.

Ideally, would cache messages on startup from a list of "all supported protocols", rather than having to check and maybe cache on sending a response. Not sure where to find an "all supported protocols" list though, it seems each hub has its own. Could try something like caching any new protocols on hub creation, but that seems like more trouble than it's worth.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
DylanDmitri: 😞

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
DylanDmitri: This would look like `return (ReadOnlySpan<byte>)memoryBufferWriter.ToArray();`

I kind of like the previous way, because the symmetry between the two returns makes the function easier to read, at least to me. This is run 2-3 times total per server, so perf concerns are minimal.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
DylanDmitri: I agree that it's weird, but I felt bad writing the same "try to cast, default to 0" logic in two different places. Is there a better place (than a ctor) for holding this logic?

anurse: If you need to do that logic in multiple places, you could write an extension method:

```csharp
public static class HubProtocolMinorVersionExtensions
{
    public static int GetMinorVersion(this IHubProtocol protocol)
    {
        // Do the magic.
    }
}
```

Then it could be called from somewhere else just like a regular method: `protocol.GetMinorVersion()`.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/IHubProtocol.cs
anurse: This is new territory for us. We don't have any `InterfaceName2` interfaces around. But I do generally agree.

anurse:
Spanish build failure is https://github.com/aspnet/SignalR/issues/2537

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -53,6 +53,7 @@ public partial class HubConnection
 
         // Transient state to a connection
         private ConnectionState _connectionState;
+        internal int _serverProtocolMinorVersion;
```
anurse: This doesn't need to be `internal` any more, right?

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,26 +20,39 @@ public static class HandshakeProtocol
...
+
+            if(!_messageCache.TryGetValue(protocol, out result))
             {
-                MemoryBufferWriter.Return(memoryBufferWriter);
+                var memoryBufferWriter = MemoryBufferWriter.Get();
+                try
+                {
+                    WriteResponseMessage(new HandshakeResponseMessage(protocol.GetMinorVersion()), memoryBufferWriter);
+                    _messageCache.TryAdd(protocol, memoryBufferWriter.ToArray());
+                    result = (ReadOnlyMemory<byte>)memoryBufferWriter.ToArray();
```
anurse: Just call `ToArray` once and store it in a variable. In fact, you can just store it in `result` first and then use `result` to add to the `_messageCache`. `ToArray` has to allocate new memory to store the data, so we don't want to call it twice.

Also, you don't need an explicit cast to `ReadOnlyMemory<byte>`. The following compiles just file: `ReadOnlyMemory<byte> foo = memoryBufferWriter.ToArray()`

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
```diff
@@ -11,21 +11,36 @@ public class HandshakeResponseMessage : HubMessage
...
         /// <summary>
         /// Initializes a new instance of the <see cref="HandshakeResponseMessage"/> class.
+        /// If you are intending to send back an error, the minor version doesn't matter.
         /// </summary>
-        /// <param name="error">An optional response error message. A <c>null</c> error message indicates a succesful handshake.</param>
-        public HandshakeResponseMessage(string error)
+        /// <param name="error"></param>
+        public HandshakeResponseMessage(string error) : this(null, error) { }
+
+        public HandshakeResponseMessage(int minorVersion) : this(minorVersion, null) { }
```
anurse: Add some documentation comments to this as well. Including a `<param>` tag for the `minorVersion` that explains it's purpose.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
```diff
@@ -11,21 +11,36 @@ public class HandshakeResponseMessage : HubMessage
...
         public string Error { get; }
 
+        /// <summary>
+        /// Highest minor protocol version that the server supports.
+        /// </summary>
+        public int MinorVersion { get; }
+
         /// <summary>
         /// Initializes a new instance of the <see cref="HandshakeResponseMessage"/> class.
+        /// If you are intending to send back an error, the minor version doesn't matter.
```
anurse: Suggested rephrase: An error response does not include a minor version, since the handshake has failed.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HubProtocolExtensions.cs
```diff
@@ -29,5 +29,10 @@ public static byte[] GetMessageBytes(this IHubProtocol hubProtocol, HubMessage m
                 MemoryBufferWriter.Return(writer);
             }
         }
+
+        public static int GetMinorVersion(this IHubProtocol protocol)
+        {
+            return (protocol is IHubProtocol2) ? ((IHubProtocol2)protocol).MinorVersion : 0;
```
anurse: You can make this even sleeker:

```csharp
return (protocol is IHubProtocol2 protocol2) ? protocol2.MinorVersion : 0;
```

The `is` statement can create a new variable where the type has been changed without requiring another cast.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HubProtocolExtensions.cs
```diff
@@ -29,5 +29,10 @@ public static byte[] GetMessageBytes(this IHubProtocol hubProtocol, HubMessage m
                 MemoryBufferWriter.Return(writer);
             }
         }
+
+        public static int GetMinorVersion(this IHubProtocol protocol)
```
anurse: We should add documentation comments here:

```csharp
/// <summary>
/// Gets the minor version for an <see cref="IHubProtocol" />. Returns 0 if the protocol does not specify a minor version.
/// </summary>
```

src/Microsoft.AspNetCore.SignalR.Protocols.MessagePack/Protocol/MessagePackHubProtocol.cs
```diff
@@ -29,13 +29,16 @@ public class MessagePackHubProtocol : IHubProtocol
...
-
+        private static readonly int ProtocolMinorVersion = 0;
+        
         /// <inheritdoc />
         public string Name => ProtocolName;
 
         /// <inheritdoc />
         public int Version => ProtocolVersion;
 
+        public int MinorVersion => ProtocolMinorVersion;
```
anurse: Add the `/// <inheritdoc />` line the other properties have. That tells our documentation generator to copy the docs from the `IHubProtocol2.MinorVersion` property here.

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Protocol.cs
```diff
@@ -63,6 +63,25 @@ public class Protocol
...
+            public async Task ClientIsOkayReceivingMinorVersionInHandshake()
+            {
+                var connection = new TestConnection(autoHandshake: false);
+                var hubConnection = CreateHubConnection(connection);
+                try
+                {
+                    var startTask = hubConnection.StartAsync();
+                    var message = await connection.ReadHandshakeAndSendResponseAsync(56);
+
+                    await startTask;
```
anurse: Add a comment here: "We're just testing that the client doesn't fail when a minor version is added to the handshake. The client doesn't actually use that version anywhere yet so there's nothing else to test at this time"

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
DylanDmitri: oops

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,26 +20,39 @@ public static class HandshakeProtocol
...
         private const string ProtocolPropertyName = "protocol";
         private const string ProtocolVersionPropertyName = "version";
+        private const string MinorVersionPropertyName = "minorVersion";
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
         /// <summary>
-        /// The serialized representation of a success handshake.
+        /// Previously used to cache the errorless return message.
+        /// Now, return messages need a minor version number, so multiple are cached in a dictionary.
```
anurse: We generally don't want to talk about previous versions or future versions in docs as that quickly becomes out of date. How about:

The serialized representation of a successful handshake with no minor version value.

It's descriptive and the obsolete attribute does a better job of telling you not to use it :)

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
```diff
@@ -11,21 +11,41 @@ public class HandshakeResponseMessage : HubMessage
...
+        /// </summary>
+        /// <param name="minorVersion"></param>
+        public HandshakeResponseMessage(int minorVersion) : this(minorVersion, null) { }
+
         /// <summary>
         /// Initializes a new instance of the <see cref="HandshakeResponseMessage"/> class.
         /// </summary>
-        /// <param name="error">An optional response error message. A <c>null</c> error message indicates a succesful handshake.</param>
-        public HandshakeResponseMessage(string error)
+        public HandshakeResponseMessage(int? minorVersion, string error)
```
anurse: I'd rather this took an `int` and that the caller of the constructor does the `?? 0` logic. As a general rule I think it's good when constructor parameters map directly to the type and name of the property, unless there's a really good reason to do that. It means the caller has to do some more work, but that's actually a good thing here I think because it lets the caller decide what a "missing" minor version represents.

anurse:
I think we're going to move this to 3.0 since the client-to-server streaming feature is moving to 3.0. That would also mean that this doesn't have to do the ugly `IHubProtocol2` thing. Thoughts @davidfowl ?

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,26 +20,39 @@ public static class HandshakeProtocol
...
         /// </summary>
-        public static ReadOnlyMemory<byte> SuccessHandshakeData;
+        [Obsolete("A successful handshake should include a minor version, but the value here does not.")]
+        public static ReadOnlyMemory<byte> SuccessHandshakeData = new byte[] { 123, 34, 109, 105, 110, 111, 114, 86, 101, 114, 115, 105, 111, 110, 34, 58, 48, 125, 30 };
```
BrennanConroy: `= Encoding.UTF8.GetBytes("string");`

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,26 +20,39 @@ public static class HandshakeProtocol
...
-            var memoryBufferWriter = MemoryBufferWriter.Get();
-            try
-            {
-                WriteResponseMessage(HandshakeResponseMessage.Empty, memoryBufferWriter);
-                SuccessHandshakeData = memoryBufferWriter.ToArray();
-            }
-            finally
+            ReadOnlyMemory<byte> result;
+
+            if(!_messageCache.TryGetValue(protocol, out result))
```
BrennanConroy: nit: space

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -107,6 +126,7 @@ public static void WriteResponseMessage(HandshakeResponseMessage responseMessage
         /// <returns>A value that is <c>true</c> if the <see cref="HandshakeResponseMessage"/> was successfully parsed; otherwise, <c>false</c>.</returns>
         public static bool TryParseResponseMessage(ref ReadOnlySequence<byte> buffer, out HandshakeResponseMessage responseMessage)
         {
+
```
BrennanConroy: remove

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
```diff
@@ -11,21 +11,41 @@ public class HandshakeResponseMessage : HubMessage
...
         {
-            // Note that a response with an empty string for error in the JSON is considered an errored response
+            // MinorVersion defaults to 0, because old servers don't send a minor version 
+            MinorVersion = minorVersion ?? 0;
             Error = error;
         }
+
```
BrennanConroy: remove

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -164,6 +188,7 @@ public static bool TryParseResponseMessage(ref ReadOnlySequence<byte> buffer, ou
             }
         }
 
+
```
BrennanConroy: remove

test/Microsoft.AspNetCore.SignalR.Client.Tests/TestConnection.cs
```diff
@@ -1,6 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Connections;
```
BrennanConroy: revert

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HubProtocolExtensions.cs
```diff
@@ -30,9 +30,15 @@ public static byte[] GetMessageBytes(this IHubProtocol hubProtocol, HubMessage m
             }
         }
 
+        /// <summary>
+        /// Gets the minor version for an <see cref="IHubProtocol"/>.
+        /// Defaults to 0 if the protocol does not specify a minor version.
+        /// </summary>
+        /// <param name="protocol"></param>
```
JamesNK: The protocol

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HubProtocolExtensions.cs
```diff
@@ -30,9 +30,15 @@ public static byte[] GetMessageBytes(this IHubProtocol hubProtocol, HubMessage m
...
+        /// <returns></returns>
```
JamesNK: The protocol's minor version. Defaults to 0 if the protocol does not specify a minor version.

anurse:
Let's do this in 3.0. I'll update the bug and @DylanDmitri you can remove `IHubProtocol2` and just add the `MinorVersion` field to the `IHubProtocol` interface :)

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,26 +19,37 @@ public static class HandshakeProtocol
...
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
         /// <summary>
-        /// The serialized representation of a success handshake.
+        /// The serialized representation of a successful handshake with no minor version value.
         /// </summary>
-        public static ReadOnlyMemory<byte> SuccessHandshakeData;
+        [Obsolete("A successful handshake should include a minor version, but the value here defaults to 0.")]
+        public static ReadOnlyMemory<byte> SuccessHandshakeData = Encoding.UTF8.GetBytes("{\"minorVersion\":0}");
```
anurse: We can just remove this if this is going in to 3.0 ;)

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs
anurse: Ping on this comment :)

src/Microsoft.AspNetCore.SignalR.Common/Protocol/IHubProtocol.cs
```diff
@@ -56,5 +56,10 @@ public interface IHubProtocol
         /// <param name="version">The version.</param>
         /// <returns>A value indicating whether the protocol supports the specified version.</returns>
         bool IsVersionSupported(int version);
+
+        /// <summary>
+        /// Gets the minor version of the protocol.
+        /// </summary>
+        int MinorVersion { get; }
```
anurse: nit: move this property up to be right under `Version`

anurse:
Hey @DylanDmitri if you can respond to the active feedback here we can get this merged pretty soon :)

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
anurse: Still haven't deleted this?

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
anurse: still a blank line here that isn't needed

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
anurse: still a blank line here that isn't needed

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -817,7 +821,7 @@ private void CheckDisposed()
                     finally
                     {
                         // The buffer was sliced up to where it was consumed, so we can just advance to the start.
-                        // We mark examined as buffer.End so that if we didn't receive a full frame, we'll wait for more data
+                        // We mark examined as buffer. End so that if we didn't receive a full frame, we'll wait for more data
```
anurse: We actually don't want a space here since `buffer.End` is a property, so the period isn't ending the sentence ;)

anurse:
Ok, things look good, but we need to get the breaking changes approved. I'll get that process going.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/IHubProtocol.cs
```diff
@@ -18,10 +18,15 @@ public interface IHubProtocol
...
         /// <summary>
-        /// Gets the version of the protocol.
+        /// Gets the major version of the protocol.
         /// </summary>
         int Version { get; }
 
+        /// <summary>
+        /// Gets the minor version of the protocol.
+        /// </summary>
+        int MinorVersion { get; }
```
anurse: @muratg for breaking change approval in 3.0. This is a new property that *hub protocols* will need to implement.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -17,26 +19,31 @@ public static class HandshakeProtocol
...
         private const string ProtocolPropertyName = "protocol";
         private const string ProtocolVersionPropertyName = "version";
+        private const string MinorVersionPropertyName = "minorVersion";
         private const string ErrorPropertyName = "error";
         private const string TypePropertyName = "type";
 
-        /// <summary>
-        /// The serialized representation of a success handshake.
-        /// </summary>
-        public static ReadOnlyMemory<byte> SuccessHandshakeData;
```
anurse: @muratg for breaking change approval in **3.0**. We're removing this since it's not really appropriate any more. It used to be the case that a successful handshake is always the exact same content. However, now it can vary based on the minor version provided. So, it's no longer appropriate to have a static field with the handshake content.

We could look at keeping this in as `Obsolete` but this is an extremely esoteric API that normal users shouldn't really be using (it should have been `internal` in the first place).

anurse:
@natemcmaster @ryanbrandenburg What's the process for updating baselines in `master` when breaking changes are made for 3.0? I tried the `generate api-baselines` command as described in https://github.com/aspnet/BuildTools/blob/dev/docs/KoreBuild.md#korebuild-commands but it didn't update the baselines, it just reported the same error we're seeing in the CI builds right now (that the API check failed because breaking changes were added)

natemcmaster:
You don't update baselines until we have a minor release to base it one. Follow the instructions in the error message and add a breakingchanges.netcore.json file. When we produce a new baseline (e.g. when 2.2 or 3.0 release) those breaking changes will get rolled into the new baseline.

```
2018-07-05T17:53:09.6389138Z   The following types have one or more members removed from them.
2018-07-05T17:53:09.6389339Z   
2018-07-05T17:53:09.6389534Z       public static class Microsoft.AspNetCore.SignalR.Protocol.HandshakeProtocol
2018-07-05T17:53:09.6389801Z           public static System.ReadOnlyMemory<System.Byte> SuccessHandshakeData
2018-07-05T17:53:09.6390326Z   
2018-07-05T17:53:09.6390462Z   
2018-07-05T17:53:09.6390576Z   
2018-07-05T17:53:09.6390748Z   The following interfaces have one or more members added to them.
2018-07-05T17:53:09.6390951Z       public interface Microsoft.AspNetCore.SignalR.Protocol.IHubProtocol
2018-07-05T17:53:09.6391150Z           System.Int32 get_MinorVersion()
2018-07-05T17:53:09.6391287Z   
2018-07-05T17:53:09.6391413Z   
2018-07-05T17:53:09.6391529Z   
2018-07-05T17:53:09.6391747Z   Following is the list of exclusions that either need to be added to the list of breaking changes, or the breaking changes themselves need to be reverted:
2018-07-05T17:53:09.7051045Z   [
2018-07-05T17:53:09.7052639Z     {
2018-07-05T17:53:09.7052962Z       "TypeId": "public static class Microsoft.AspNetCore.SignalR.Protocol.HandshakeProtocol",
2018-07-05T17:53:09.7053267Z       "MemberId": "public static System.ReadOnlyMemory<System.Byte> SuccessHandshakeData",
2018-07-05T17:53:09.7053553Z       "Kind": "Removal"
2018-07-05T17:53:09.7053746Z     },
2018-07-05T17:53:09.7053938Z     {
2018-07-05T17:53:09.7054177Z       "TypeId": "public interface Microsoft.AspNetCore.SignalR.Protocol.IHubProtocol",
2018-07-05T17:53:09.7054456Z       "MemberId": "System.Int32 get_MinorVersion()",
2018-07-05T17:53:09.7054682Z       "Kind": "Addition"
2018-07-05T17:53:09.7054889Z     }
2018-07-05T17:53:09.7055066Z   ]
2018-07-05T17:53:09.7055258Z   
2018-07-05T17:53:09.7055520Z   The process for breaking changes is described in: https://github.com/aspnet/Home/wiki/Engineering-guidelines#breaking-changes
2018-07-05T17:53:09.7055872Z   The process to add an exclusion to this tool is described in: https://github.com/aspnet/BuildTools/wiki/Api-Check#apicheck-exceptions
```

src/Microsoft.AspNetCore.SignalR.Common/Protocol/IHubProtocol.cs
muratg: :shipit: (in 3.0) :)

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
muratg: I'm OK with removing in 3.0.

Question is, should we `Obsolete` it in 2.2?

anurse:
Ok, I wasn't clear on that since it looked like that was referring to the process for breaking changes during a minor release. Thanks!

Issue #515, opened 2017-06-03
Add endpoint base classes to support JSON over websockets for ReadableChannel<T> and IObservable<T>


davidfowl:
The basic idea here is that in the most common use case for websockets is streaming data over the socket using JSON. The reason there's no need for dispatch in this case is because that's already handled by the route itself (urls dispatch vs message dispatch). In these simple cases, we provide all of the plumbing and connection tracking so that websockets are efficient and connection tracking is good. These 2 derived endpoints are just samples of what it could look like (the below is hacked together). Obviously we'd need serializer extensibility and we could consider supporting non JSON but here's that's the gist.

```C#
app.UseSockets(routes =>
{
    routes.MapEndPoint<Stocks>("stocks");
});
```

### IObservable implementation

```C#
public class Stocks : ObservableEndPoint<long>
{
    public override IObservable<long> Subscribe(ConnectionContext connection)
    {
        return Observable.Interval(TimeSpan.FromSeconds(1));
    }
}
```

### Channels implementation

```C#
public class Stocks : ChannelEndPoint<long>
{
    public override ReadableChannel<long> Subscribe(ConnectionContext connection)
    {
        var channel = Channel.CreateUnbounded<long>();

        async Task Produce()
        {
            long number = 0;
            while (!connection.Transport.Input.Completion.IsCompleted)
            {
                await channel.Out.WriteAsync(number++);

                await Task.Delay(1000);
            }
        }

        _ = Produce();

        return channel.In;
    }
}
```

### Implementations

```C#
public abstract class ChannelEndPoint<T> : EndPoint
{
    public abstract ReadableChannel<T> Subscribe(ConnectionContext connection);

    public override async Task OnConnectedAsync(ConnectionContext connection)
    {
        var input = Subscribe(connection);

        while (await input.WaitToReadAsync())
        {
            while (input.TryRead(out var item))
            {
                // Inefficient, don't do this in real life
                var data = Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(item));

                connection.Transport.Output.TryWrite(new Message(data, MessageType.Text));
            }
        }
    }
}

public abstract class ObservableEndPoint<T> : EndPoint
{
    public abstract IObservable<T> Subscribe(ConnectionContext connection);

    public override Task OnConnectedAsync(ConnectionContext connection)
    {
        var tcs = new TaskCompletionSource<object>();
        var disposable = Subscribe(connection).Subscribe(item =>
        {
[...]

PR #2523, opened 2018-06-22
Added support for a raw upgrade stream transport

davidfowl:
- This transport flows frames to the ConnectionHandler itself so that it can parse the websocket frames itself.
- This is extremely useful for preserving websocket frames and for proxying and building higher level components on top that
need to frames.
- This also enables users to build **pure** WebSocket applications on top of the ConnectionHandler API directly.
- It also makes things like this https://github.com/aspnet/SignalR/issues/515 *much* easier as the underlying framing is preserved (though the implementation would be different).

~PS: Looking into adding a test~


src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/HttpUpgradeTransport.cs
```diff
@@ -0,0 +1,274 @@
...
+#if NETCOREAPP2_2
+                    var read = await stream.ReadAsync(memory, CancellationToken.None);
+#else
+                    var isArray = MemoryMarshal.TryGetArray<byte>(memory, out var arraySegment);
+                    Debug.Assert(isArray);
+
+                    // Exceptions are handled above where the send and receive tasks are being run.
+                    var read = await stream.ReadAsync(arraySegment.Array, arraySegment.Offset, arraySegment.Count, CancellationToken.None);
+#endif
+                    // Need to check again for NetCoreApp2.2 because a close can happen between a 0-byte read and the actual read
```
BrennanConroy: Copy pasta.
So what happens if the 0-byte read returns from a close then we try the real read?
I think before when we didn't check it threw when you tried to read again, but I don't know the behavior of the stream

davidfowl:
Finally updated this.

anurse:
Lets chat about this on Wednesday and decide where this should land. It looks lightweight enough that I'm ok with taking it in 2.2 as long as it's well tested. My concern isn't with the code itself though, but with the maintenance and documentation needs. I'm also not a huge fan of the way things plugs in, since it couples the Connection Handler to the transport (so it requires proper configuration and nothing enforces that)

davidfowl:
This doesn’t need documentation it’s pretty specific. As for bleeding details to the ConnectionHandler, that’s the entire feature. Primarily I’d like to build features on top of this and tbh it isn’t a SignalR feature, it’s a transport feature. It’s a building block for building other features described in the original PR.

test/Microsoft.AspNetCore.SignalR.Tests/WebSocketConnectionHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    public class WebSocketConnectionHandler : ConnectionHandler
+    {
+        public override async Task OnConnectedAsync(ConnectionContext connection)
+        {
+            using (var websocket = WebSocketProtocol.CreateFromStream(new PipeDuplexStream(connection.Transport), isServer: true, subProtocol: null, keepAliveInterval: TimeSpan.FromMinutes(2)))
```
davidfowl: Writing this test reveled that we need to flow the subprotocol here.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/HttpUpgradeTransport.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        private async Task StartReceiving(Stream stream)
+        {
+            try
+            {
+                while (true)
+                {
+#if NETCOREAPP2_2
+                    // Do a 0 byte read so that idle connections don't allocate a buffer when waiting for a read
+                    await stream.ReadAsync(Memory<byte>.Empty, CancellationToken.None);
```
BrennanConroy: I think the issue is, you do a 0-byte read here and there was a 0 byte message from the pipe (the close), then when you do the actual read below you wait for a message that was already read so it will "hang"

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/HttpUpgradeTransport.cs
```diff
@@ -0,0 +1,285 @@
...
+                            break;
+                        }
+                    }
+                    finally
+                    {
+                        _application.Input.AdvanceTo(buffer.End);
+                    }
+                }
+            }
+            catch (WebSocketException ex) when (ex.WebSocketErrorCode == WebSocketError.ConnectionClosedPrematurely)
```
anurse: When would this ever happen? You're writing to a stream now, not a WebSocket.

src/Microsoft.AspNetCore.Http.Connections/WebSocketOptions.cs
```diff
@@ -10,6 +10,11 @@ public class WebSocketOptions
     {
         public TimeSpan CloseTimeout { get; set; } = TimeSpan.FromSeconds(5);
 
+        /// <summary>
+        /// Gets or sets a value that determines if websocket frames should be be preserved.
```
anurse: If this is set in `MapHub`, you're going to have a very bad time. We should a) document that it can't be set when using Hubs and b) have SignalR throw if the feature is present and indicates that raw frames are being used.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/HttpUpgradeTransport.cs
davidfowl: You could argue this should be handled in the application layer and not in the transport.

src/Microsoft.AspNetCore.Http.Connections/WebSocketOptions.cs
davidfowl: Yep. Will do.

davidfowl: So this throws a wrench into things a bit. I think it makes sense to see if it's possible to wire this up some other way without it being an option on WebSockets.

anurse: That was kinda my initial thought, and it was the SignalR problem that really clarified it for me.

davidfowl:
I'm going to close this for now.

PR #2524, opened 2018-06-22
Adding the HubProtocol Layer to the Java Client

mikaelm12:
Adding the HubProtocol Layer. The `HubConnection `automatically uses the `JsonHubProtocol `for now but this change is the first step for supporting other protocols later.
It still only really handles `InvocationMessage `message types. I'll add support for the other message types in another PR.

### DIFF
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -22,7 +22,7 @@ public class HubConnection {
 
             // message will be null if we receive any message other than an invocation.
             // Adding this to avoid getting error messages on pings for now.
-            for (InvocationMessage message: messages) {
+            for (InvocationMessage message : messages) {
                 if (message != null && handlers.containsKey(message.target)) {
                     handlers.get(message.target).invoke(message.arguments[0]);
                 }
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -47,8 +47,8 @@ public class HubConnection {
         connected = false;
     }
 
-    public void send(String method, Object arg1) {
-        InvocationMessage invocationMessage = new InvocationMessage(method, new Object[]{ arg1 });
+    public void send(String method, Object...args) {
+        InvocationMessage invocationMessage = new InvocationMessage(method, args);
         String message = protocol.writeMessage(invocationMessage);
         _transport.send(message);
     }
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -47,7 +47,7 @@ public class HubConnection {
         connected = false;
     }
 
-    public void send(String method, Object...args) {
+    public void send(String method, Object... args) {
         InvocationMessage invocationMessage = new InvocationMessage(method, args);
         String message = protocol.writeMessage(invocationMessage);
         _transport.send(message);
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -6,10 +6,10 @@ import java.util.HashMap;
 
 public class HubConnection {
     private String _url;
-    private ITransport _transport;
+    private Transport _transport;
     private OnReceiveCallBack callback;
     private HashMap<String, Action> handlers = new HashMap<>();
-    private IHubProtocol protocol;
+    private HubProtocol protocol;
 
     public Boolean connected = false;
 
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -1,9 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
-
 import java.net.URISyntaxException;
 import java.util.HashMap;
 
@@ -12,7 +9,7 @@ public class HubConnection {
     private ITransport _transport;
     private OnReceiveCallBack callback;
     private HashMap<String, Action> handlers = new HashMap<>();
-    private JsonParser jsonParser = new JsonParser();
+    private IHubProtocol protocol;
     private static final String RECORD_SEPARATOR = "\u001e";
 
     public Boolean connected = false;
@@ -28,9 +25,17 @@ public class HubConnection {
                 if (splitMessage.length() == 2) {
                     continue;
                 }
-                processMessage(splitMessage);
+
+                InvocationMessage message = protocol.parseMessage(splitMessage);
+
+                // message will be null if we receive any message other than an invocation.
+                // Adding this to avoid getting error messages on pings for now.
+                if (message != null && handlers.containsKey(message.target)) {
+                    handlers.get(message.target).invoke(message.arguments[0]);
+                }
             }
         };
+        protocol = new JsonHubProtocol();
 
         try {
             _transport = new WebSocketTransport(_url);
@@ -39,43 +44,7 @@ public class HubConnection {
         }
     }
 
-    private void processMessage(String message) {
-        JsonObject jsonMessage = jsonParser.parse(message).getAsJsonObject();
-        String messageType = jsonMessage.get("type").toString();
-        switch(messageType) {
-            case "1":
-                //Invocation Message
-                String target = jsonMessage.get("target").getAsString();
-                if (handlers.containsKey(target)) {
-                    handlers.get(target).invoke(jsonMessage.get("arguments"));
-                }
-                break;
-            case "2":
-                //Stream item
-                //Don't care yet
-                break;
-            case "3":
-                //Completion
-                //Don't care yet
-                break;
-            case "4":
-                //Stream invocation
-                //Don't care yet;
-                break;
-            case "5":
-                //Cancel invocation
-                //Don't care yet
-                break;
-            case "6":
-                //Ping
-                //Don't care yet
-                break;
-            case "7":
-                // Close message
-                //Don't care yet;
-                break;
-        }
-    }
+
 
     public void start() throws InterruptedException {
         _transport.setOnReceive(this.callback);
@@ -89,7 +58,8 @@ public class HubConnection {
     }
 
     public void send(String method, Object arg1) {
-        InvocationMessage message = new InvocationMessage(method, new Object[]{ arg1 });
+        InvocationMessage invocationMessage = new InvocationMessage(method, new Object[]{ arg1 });
+        String message = protocol.writeMessage(invocationMessage);
         _transport.send(message);
     }
 
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -10,32 +10,24 @@ public class HubConnection {
     private OnReceiveCallBack callback;
     private HashMap<String, Action> handlers = new HashMap<>();
     private IHubProtocol protocol;
-    private static final String RECORD_SEPARATOR = "\u001e";
 
     public Boolean connected = false;
 
     public HubConnection(String url) {
         _url = url;
+        protocol = new JsonHubProtocol();
         callback = (payload) -> {
-            String[] messages = payload.split(RECORD_SEPARATOR);
 
-            for (String splitMessage : messages) {
+            InvocationMessage[] messages = protocol.parseMessages(payload);
 
-                // Empty handshake response "{}". We can ignore it
-                if (splitMessage.length() == 2) {
-                    continue;
-                }
-
-                InvocationMessage message = protocol.parseMessage(splitMessage);
-
-                // message will be null if we receive any message other than an invocation.
-                // Adding this to avoid getting error messages on pings for now.
+            // message will be null if we receive any message other than an invocation.
+            // Adding this to avoid getting error messages on pings for now.
+            for (InvocationMessage message: messages) {
                 if (message != null && handlers.containsKey(message.target)) {
                     handlers.get(message.target).invoke(message.arguments[0]);
                 }
             }
         };
-        protocol = new JsonHubProtocol();
 
         try {
             _transport = new WebSocketTransport(_url);
@@ -43,8 +35,6 @@ public class HubConnection {
             e.printStackTrace();
         }
     }
-
-
 
     public void start() throws InterruptedException {
         _transport.setOnReceive(this.callback);
--- a/clients/java/signalr/src/main/java/IHubProtocol.java
+++ b/clients/java/signalr/src/main/java/IHubProtocol.java
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-import com.google.gson.JsonObject;
-
 public interface IHubProtocol {
     String name();
     int version();
--- a/clients/java/signalr/src/main/java/IHubProtocol.java
+++ b/clients/java/signalr/src/main/java/IHubProtocol.java
@@ -5,7 +5,7 @@ public interface IHubProtocol {
     String name();
     int version();
     TransferFormat transferFormat();
-    InvocationMessage parseMessage(String message);
+    InvocationMessage[] parseMessages(String message);
     String writeMessage(InvocationMessage message);
 }
 
--- a/clients/java/signalr/src/main/java/IHubProtocol.java
+++ b/clients/java/signalr/src/main/java/IHubProtocol.java
@@ -1,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import com.google.gson.JsonObject;
+
+public interface IHubProtocol {
+    String name();
+    int version();
+    TransferFormat transferFormat();
+    InvocationMessage parseMessage(String message);
+    String writeMessage(InvocationMessage message);
+}
+
--- a/clients/java/signalr/src/main/java/JsonHubProtocol.java
+++ b/clients/java/signalr/src/main/java/JsonHubProtocol.java
@@ -34,7 +34,6 @@ public class JsonHubProtocol implements HubProtocol {
         String[] messages = payload.split(RECORD_SEPARATOR);
         List<InvocationMessage> invocationMessages = new ArrayList<>();
         for (String splitMessage : messages) {
-
             // Empty handshake response "{}". We can ignore it
             if (splitMessage.equals("{}")) {
                 continue;
--- a/clients/java/signalr/src/main/java/JsonHubProtocol.java
+++ b/clients/java/signalr/src/main/java/JsonHubProtocol.java
@@ -31,18 +31,18 @@ public class JsonHubProtocol implements HubProtocol {
 
     @Override
     public InvocationMessage[] parseMessages(String payload) {
-
         String[] messages = payload.split(RECORD_SEPARATOR);
         List<InvocationMessage> invocationMessages = new ArrayList<>();
-        // Empty handshake response "{}". We can ignore it
-        for (String splitMessage: messages) {
+        for (String splitMessage : messages) {
+
+            // Empty handshake response "{}". We can ignore it
             if (splitMessage.equals("{}")) {
                 continue;
             }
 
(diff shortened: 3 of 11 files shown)


clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -12,7 +9,7 @@
     private ITransport _transport;
     private OnReceiveCallBack callback;
     private HashMap<String, Action> handlers = new HashMap<>();
-    private JsonParser jsonParser = new JsonParser();
+    private IHubProtocol protocol;
     private static final String RECORD_SEPARATOR = "\u001e";
```
davidfowl: Remove this.

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -28,9 +25,17 @@ public HubConnection(String url) {
                 if (splitMessage.length() == 2) {
                     continue;
                 }
-                processMessage(splitMessage);
+
+                InvocationMessage message = protocol.parseMessage(splitMessage);
```
davidfowl: You should pass the payload directly to the protocol.

clients/java/signalr/src/main/java/IHubProtocol.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public interface IHubProtocol {
+    String name();
+    int version();
+    TransferFormat transferFormat();
+    InvocationMessage parseMessage(String message);
```
davidfowl: I think this needs to return an array.

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -12,7 +9,7 @@
     private ITransport _transport;
     private OnReceiveCallBack callback;
     private HashMap<String, Action> handlers = new HashMap<>();
-    private JsonParser jsonParser = new JsonParser();
+    private IHubProtocol protocol;
```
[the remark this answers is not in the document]
mikaelm12: The java standard I believe is to use camelCase and  `this.` I was planning on updating it everywhere in a style check pr after this 😄 
https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names

clients/java/signalr/src/main/java/IHubProtocol.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public interface IHubProtocol {
```
davidfowl: We need to model this after the JS client APIs since we're not going to be handling partial messages.

clients/java/signalr/src/main/java/IHubProtocol.java
mikaelm12: Yeah, if we pass the entire payload into the protocol than this should be an array

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+        String[] messages = payload.split(RECORD_SEPARATOR);
+        ArrayList<InvocationMessage> invocationMessages = new ArrayList<>();
+        // Empty handshake response "{}". We can ignore it
+        for (String splitMessage: messages) {
+            if (splitMessage.length() == 2) {
+                continue;
+            }
+
+            JsonObject jsonMessage = jsonParser.parse(splitMessage).getAsJsonObject();
+            String messageType = jsonMessage.get("type").toString();
```
moozzyk: make an enum and parse it back?

mikaelm12:
Updated

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+        return 1;
+    }
+
+    @Override
+    public TransferFormat transferFormat() {
+        return TransferFormat.Text;
+    }
+
+    @Override
+    public InvocationMessage[] parseMessages(String payload) {
```
moozzyk: rant: I _love_ how Java insists on a curly bracket at the end and then you insert a new line  to make code readable.

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+
+        String[] messages = payload.split(RECORD_SEPARATOR);
+        ArrayList<InvocationMessage> invocationMessages = new ArrayList<>();
```
moozzyk: should be `List<InvocationMessage>` on the left side. Java 10 has `var` (and, honestly, nothing more for simple folks like me)

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+                    break;
+                case "7":
+                    // Close message
+                    //Don't care yet;
+                    break;
+            }
+        }
+        InvocationMessage[] copiedMessages = new InvocationMessage[invocationMessages.size()];
+        invocationMessages.toArray(copiedMessages);
+        return copiedMessages;
```
moozzyk: I know that you gain extra point in Java if you can write a one-liner in 5 lines but it can be just:

```
return invocationMessages.toArray(new InvocationMessage[invocationMessage.size()]);
```

(and passing the array here in is super weird and I believe has to do either with type erasure or because non-generic `.toArray` was already taken or both).

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final String RECORD_SEPARATOR = "\u001e";
+
+    @Override
+    public String name() {
+        return "json";
+    }
+
+    @Override
+    public int version() {
+        return 1;
```
moozzyk: Wait... You're just returning a value you need instead of creating a nested static factory class that returns a VersionManager that returns a value of a private a final static UPPER_CASED variable containing the version (or an enum)? Dude, your break from Java was too long...

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+import java.util.ArrayList;
+
+public class JsonHubProtocol implements IHubProtocol{
+    private JsonParser jsonParser = new JsonParser();
```
moozzyk: this could be `final` (as well as almost any local variable below and function input parameters)

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -12,23 +9,23 @@
...
+            InvocationMessage[] messages = protocol.parseMessages(payload);
 
-                // Empty handshake response "{}". We can ignore it
-                if (splitMessage.length() == 2) {
-                    continue;
+            // message will be null if we receive any message other than an invocation.
+            // Adding this to avoid getting error messages on pings for now.
+            for (InvocationMessage message: messages) {
+                if (message != null && handlers.containsKey(message.target)) {
+                    handlers.get(message.target).invoke(message.arguments[0]);
```
moozzyk: how about methods taking more than one arguments?

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -89,11 +48,12 @@ public void stop(){
     }
 
     public void send(String method, Object arg1) {
-        InvocationMessage message = new InvocationMessage(method, new Object[]{ arg1 });
+        InvocationMessage invocationMessage = new InvocationMessage(method, new Object[]{ arg1 });
```
moozzyk: can you just take `final @NonNull Object... args` to avoid lotsa overloads?

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public InvocationMessage[] parseMessages(String payload) {
+
+        String[] messages = payload.split(RECORD_SEPARATOR);
+        ArrayList<InvocationMessage> invocationMessages = new ArrayList<>();
+        // Empty handshake response "{}". We can ignore it
+        for (String splitMessage: messages) {
+            if (splitMessage.length() == 2) {
```
moozzyk: nit: you could also `!"{}".equals(message)` for readability

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        InvocationMessage[] copiedMessages = new InvocationMessage[invocationMessages.size()];
+        invocationMessages.toArray(copiedMessages);
+        return copiedMessages;
+    }
+
+    @Override
+    public String writeMessage(InvocationMessage invocationMessage) {
+        Gson gson = new Gson();
+        return gson.toJson(invocationMessage) + RECORD_SEPARATOR;
```
moozzyk: just because there is not string interpolation...

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,86 @@
...
+            }
+        }
+        InvocationMessage[] copiedMessages = new InvocationMessage[invocationMessages.size()];
+        invocationMessages.toArray(copiedMessages);
+        return copiedMessages;
+    }
+
+    @Override
+    public String writeMessage(InvocationMessage invocationMessage) {
+        Gson gson = new Gson();
```
moozzyk: feels expensive to instantiate this for each message.

moozzyk:
I am also missing some tests...

clients/java/signalr/src/main/java/JsonHubProtocol.java
mikaelm12: Yeah, that's a better approach and eventually what I'll do. This is still very early stages. Just started on this this week 😄

clients/java/signalr/src/main/java/JsonHubProtocol.java
mikaelm12: Yeah, that's better. More exact and readable

davidfowl: You might want to stub that out by adding a HandshakeProtocol next and putting that outside of  the protocol implementation itself.

clients/java/signalr/src/main/java/JsonHubProtocol.java
mikaelm12: Yeah, you're right. Will move this up with the JsonParser

clients/java/signalr/src/main/java/HubConnection.java
mikaelm12: Haven't added support for that just yet 😄

clients/java/signalr/src/main/java/HubConnection.java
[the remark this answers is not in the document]
anurse: Let's keep the style consistent now rather than a batch update later. Otherwise we're likely to forget.

clients/java/signalr/src/main/java/IHubProtocol.java
anurse: Also, Java naming style is to **not** use a prefix for interface names. So this should just be called `HubProtocol`

clients/java/signalr/src/main/java/IHubProtocol.java
```diff
@@ -0,0 +1,11 @@
...
+    String name();
+    int version();
```
anurse: Is this the naming convention for this kind of thing? I vaguely remember using the `getVersion()` style way back when I last touched Java but maybe that's changed.

clients/java/signalr/src/main/java/JsonHubProtocol.java
anurse: Needs moar `JsonHubProtocolVersionIdentifierFactory`

clients/java/signalr/src/main/java/IHubProtocol.java
mikaelm12: Oh yeah. No prefix

clients/java/signalr/src/main/java/IHubProtocol.java
mikaelm12: Should be `getVersion`

clients/java/signalr/src/main/java/IHubProtocol.java
anurse: I think I prefer that over the .NET style, which smells too much like ye-olde-COM. But there's no changing .NET style now 😄

clients/java/signalr/src/test/java/JsonHubProtocolTest.java
```diff
@@ -0,0 +1,97 @@
...
+        String expectedResult = "{\"type\":1,\"target\":\"test\",\"arguments\":[\"42\"]}\u001E";
+        assertEquals(expectedResult, result);
+    }
+
+    @Test
+    public void ParseSingleMessage(){
+        String stringifiedMessage = "{\"type\":1,\"target\":\"test\",\"arguments\":[42]}\u001E";
+        InvocationMessage[] messages = jsonHubProtocol.parseMessages(stringifiedMessage);
+        //We know it's only one message
+        InvocationMessage message = messages[0];
```
mikaelm12: Should assert the length and add a new line here

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -1,34 +1,31 @@
...
 
-            for (String splitMessage : messages) {
+            InvocationMessage[] messages = protocol.parseMessages(payload);
 
-                // Empty handshake response "{}". We can ignore it
-                if (splitMessage.length() == 2) {
-                    continue;
+            // message will be null if we receive any message other than an invocation.
+            // Adding this to avoid getting error messages on pings for now.
+            for (InvocationMessage message: messages) {
```
BrennanConroy: space before `:`

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,85 @@
...
+        return TransferFormat.Text;
+    }
+
+    @Override
+    public InvocationMessage[] parseMessages(String payload) {
+
+        String[] messages = payload.split(RECORD_SEPARATOR);
+        List<InvocationMessage> invocationMessages = new ArrayList<>();
+        // Empty handshake response "{}". We can ignore it
+        for (String splitMessage: messages) {
```
BrennanConroy: space before `:`

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,85 @@
...
+    public TransferFormat getTransferFormat() {
+        return TransferFormat.Text;
+    }
+
+    @Override
+    public InvocationMessage[] parseMessages(String payload) {
+
+        String[] messages = payload.split(RECORD_SEPARATOR);
+        List<InvocationMessage> invocationMessages = new ArrayList<>();
+        // Empty handshake response "{}". We can ignore it
```
BrennanConroy: move the comment

clients/java/signalr/src/main/java/WebSocketTransport.java
```diff
@@ -8,7 +8,7 @@
 import java.net.URI;
 import java.net.URISyntaxException;
 
-public class WebSocketTransport implements ITransport {
+public class WebSocketTransport implements Transport {
     private static final String RECORD_SEPARATOR = "\u001e";
```
BrennanConroy: remove

clients/java/signalr/src/test/java/JsonHubProtocolTest.java
```diff
@@ -0,0 +1,97 @@
...
+        assertEquals("json", jsonHubProtocol.getName());
+    }
+
+    @Test
+    public void checkVersionNumber(){
+        assertEquals(1,jsonHubProtocol.getVersion());
+    }
+
+    @Test
+    public void checkTransferFormat(){
```
BrennanConroy: casing check?

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,85 @@
...
+                continue;
+            }
+
+            JsonObject jsonMessage = jsonParser.parse(splitMessage).getAsJsonObject();
+            String messageType = jsonMessage.get("type").toString();
+            switch(messageType) {
+                case "1":
+                    //Invocation Message
+                    String target = jsonMessage.get("target").getAsString();
+                    JsonElement args = jsonMessage.get("arguments");
```
BrennanConroy: Write a TODO to parse args into types?

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public TransferFormat getTransferFormat() {
+        return TransferFormat.Text;
+    }
+
+    @Override
+    public InvocationMessage[] parseMessages(String payload) {
+
```
BrennanConroy: remove newline

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,85 @@
...
+        List<InvocationMessage> invocationMessages = new ArrayList<>();
+        // Empty handshake response "{}". We can ignore it
+        for (String splitMessage: messages) {
+            if (splitMessage.equals("{}")) {
+                continue;
+            }
+
+            JsonObject jsonMessage = jsonParser.parse(splitMessage).getAsJsonObject();
+            String messageType = jsonMessage.get("type").toString();
+            switch(messageType) {
```
BrennanConroy: nit: space after `switch`

clients/java/signalr/src/test/java/JsonHubProtocolTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private JsonHubProtocol jsonHubProtocol = new JsonHubProtocol();
+    private static final String RECORD_SEPARATOR = "\u001e";
+
+    @Test
+    public void checkProtocolName() {
+        assertEquals("json", jsonHubProtocol.getName());
+    }
+
+    @Test
+    public void checkVersionNumber(){
```
BrennanConroy: add space before all these curly braces

clients/java/signalr/src/test/java/JsonHubProtocolTest.java
```diff
@@ -0,0 +1,97 @@
+import com.google.gson.JsonArray;
```
BrennanConroy: copyright

clients/java/signalr/src/test/java/JsonHubProtocolTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private static final String RECORD_SEPARATOR = "\u001e";
+
+    @Test
+    public void checkProtocolName() {
+        assertEquals("json", jsonHubProtocol.getName());
+    }
+
+    @Test
+    public void checkVersionNumber(){
+        assertEquals(1,jsonHubProtocol.getVersion());
```
BrennanConroy: nit: missing space

mikaelm12:
Updated

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -88,12 +47,13 @@ public void stop(){
         connected = false;
     }
 
-    public void send(String method, Object arg1) {
-        InvocationMessage message = new InvocationMessage(method, new Object[]{ arg1 });
+    public void send(String method, Object...args) {
```
BrennanConroy: a space is needed on one side of the `...`

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -0,0 +1,85 @@
...
+    public TransferFormat getTransferFormat() {
+        return TransferFormat.Text;
+    }
+
+    @Override
+    public InvocationMessage[] parseMessages(String payload) {
+        String[] messages = payload.split(RECORD_SEPARATOR);
+        List<InvocationMessage> invocationMessages = new ArrayList<>();
+        for (String splitMessage : messages) {
+
```
BrennanConroy: newline

clients/java/signalr/src/main/java/JsonHubProtocol.java
moozzyk: looks like something a Java linter would detect and flag...

PR #2530, opened 2018-06-22
Ping Cleanup

DylanDmitri:
Changed a potentially misleading comment on the c# client, moved the `ResetPing` call to a more obvious location.

Both client and server skip sending pings.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -12,7 +12,6 @@ using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
-using Microsoft.AspNetCore.Connections.Features;
 using Microsoft.AspNetCore.Internal;
 using Microsoft.AspNetCore.SignalR.Client.Internal;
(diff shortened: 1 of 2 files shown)


anurse:
@dotnet-bot test Windows Release x64 Build (Spanish language image) please

anurse:
@dotnet-bot test Spanish language image please

anurse:
@pranavkm is there a way to re-run the Spanish image build? Trying the whole string just ends up running the non-spanish build again because it's a substring

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -926,6 +925,11 @@ private void OnServerTimeout()
 
         private async Task PingServer()
         {
+            if (Debugger.IsAttached)
```
anurse: Hrm, why are we not sending the Pings while debugging? I get why we don't want to time-out when debugging but I think we should still be sending the pings.

pranavkm:
Hmm. Maybe it needs to be a distinct name? Roslyn does names in the format `windows_debug_spanish_unit32_prtest`. Perhaps changing https://github.com/aspnet/SignalR/blob/dev/build/buildpipeline/pipeline.groovy#L15 to be somewhat similar may help?

PR #2536, opened 2018-06-25
update some packages

anurse:
Results of `npm audit fix` on the Functional Tests. These are **test only** dependencies. There are no changes to product dependencies here. The product dependencies have no `npm audit` warnings.


anurse:
macOS is failing due to docker issues? @BrennanConroy do we know if this these images have what we need?

BrennanConroy:
It looks like some of the OSX images have docker and fail and some don't have docker, or at least hit our skip condition if they have docker? Ignore it for now and I'll look into it

PR #2538, opened 2018-06-25
Try renaming spanish build

anurse:
dotnetbot doesn't like it when you have builds that have a matching substring at the start, it's hard to re-run them. This should make the spanish build different enough.

### DIFF
--- a/build/buildpipeline/pipeline.groovy
+++ b/build/buildpipeline/pipeline.groovy
@@ -12,7 +12,7 @@ def parameters = [
 windowsPipeline.triggerPipelineOnEveryGithubPR("Windows ${configuration} x64 Build", parameters)
 windowsPipeline.triggerPipelineOnGithubPush(parameters)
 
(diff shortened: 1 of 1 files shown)


build/buildpipeline/pipeline.groovy
```diff
@@ -12,7 +12,7 @@ def parameters = [
 windowsPipeline.triggerPipelineOnEveryGithubPR("Windows ${configuration} x64 Build", parameters)
 windowsPipeline.triggerPipelineOnGithubPush(parameters)
 
-windowsESPipeline.triggerPipelineOnEveryGithubPR("Windows ${configuration} x64 Build (Spanish language image) ", parameters)
+windowsESPipeline.triggerPipelineOnEveryGithubPR("Windows ${configuration} (Spanish Language) x64 Build", parameters)
```
pranavkm: I'd remove the parantheses to be extra safe. Unfortunately the name change won't show up until it's merged in to dev

anurse: > Unfortunately the name change won't show up until it's merged in to dev

Yep, I was expecting that.

PR #2542, opened 2018-06-26
StreamInvocationMessage => StreamInvocation Message

windperson:
There's a typo in Hub Protocol spec document, 

"**StreamInvocationMessage**" should be "**StreamInvocation** message".


anurse:
Thanks for fixing this!

PR #2544, opened 2018-06-26
Update build.gradle to not use depricated compile

mikaelm12:
`compile` is now depricated 
From the [gradle docs](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph)

> The compile, testCompile, runtime and testRuntime configurations inherited from the Java plugin are still available but are deprecated. You should avoid using them, as they are only kept for backwards compatibility.


mikaelm12:
cc @Eilon  Updated this based off your email

Eilon:
Nice!

Issue #2540, opened 2018-06-26
Refactor HubConnection to make it unit testable


mikaelm12:
Start by creating a constructor that takes an ITransport. Then we can create a MockTransport and use that to fake a connection and then write unit tests.

anurse:
https://github.com/aspnet/SignalR/commit/13a3dfe8dcb07e3433d449ce176de54dc387bda1

PR #2550, opened 2018-06-27
Adding HubConnection Tests

mikaelm12:
Adding some testing for `HubConnection` in the Java client.
Had to add another ctor to `HubConnection` to accept a transport so I could pass in a `MockTransport`
Issue: #2540

### DIFF
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -35,8 +35,7 @@ public class HubConnection {
             } catch (URISyntaxException e) {
                 e.printStackTrace();
             }
-        }
-        else {
+        } else {
             this.transport = transport;
         }
     }
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -5,18 +5,18 @@ import java.net.URISyntaxException;
 import java.util.HashMap;
 
 public class HubConnection {
-    private String _url;
-    private Transport _transport;
+    private String url;
+    private Transport transport;
     private OnReceiveCallBack callback;
     private HashMap<String, Action> handlers = new HashMap<>();
     private HubProtocol protocol;
 
     public Boolean connected = false;
 
-    public HubConnection(String url) {
-        _url = url;
-        protocol = new JsonHubProtocol();
-        callback = (payload) -> {
+    public HubConnection(String url, Transport transport){
+        this.url = url;
+        this.protocol = new JsonHubProtocol();
+        this.callback = (payload) -> {
 
             InvocationMessage[] messages = protocol.parseMessages(payload);
 
@@ -29,28 +29,37 @@ public class HubConnection {
             }
         };
 
-        try {
-            _transport = new WebSocketTransport(_url);
-        } catch (URISyntaxException e) {
-            e.printStackTrace();
+        if (transport == null){
+            try {
+                this.transport = new WebSocketTransport(this.url);
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
+            }
+        }
+        else {
+            this.transport = transport;
         }
     }
 
+    public HubConnection(String url) {
+        this(url, null);
+    }
+
     public void start() throws InterruptedException {
-        _transport.setOnReceive(this.callback);
-        _transport.start();
+        transport.setOnReceive(this.callback);
+        transport.start();
         connected = true;
     }
 
     public void stop(){
-        _transport.stop();
+        transport.stop();
         connected = false;
     }
 
     public void send(String method, Object... args) {
         InvocationMessage invocationMessage = new InvocationMessage(method, args);
         String message = protocol.writeMessage(invocationMessage);
-        _transport.send(message);
+        transport.send(message);
     }
 
     public void On(String target, Action callback) {
--- a/clients/java/signalr/src/test/java/HubConnectionTest.java
+++ b/clients/java/signalr/src/test/java/HubConnectionTest.java
@@ -4,6 +4,9 @@
 import com.google.gson.JsonArray;
 import org.junit.Test;
 
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
 import static org.junit.Assert.*;
 
 public class HubConnectionTest {
@@ -20,13 +23,13 @@ public class HubConnectionTest {
 
     @Test
     public void SendWithNoParamsTriggersOnHandler() throws InterruptedException {
-        TestObject obj = new TestObject();
+        AtomicInteger value = new AtomicInteger(0);
         Transport mockTransport = new MockEchoTransport();
         HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
 
         Action callback = (param) -> {
-            assertEquals(0, obj.counter);
-            obj.incrementCounter();
+            assertEquals(0, value.get());
+            value.set(value.incrementAndGet());
         };
         hubConnection.On("inc", callback);
 
@@ -34,19 +37,18 @@ public class HubConnectionTest {
         hubConnection.send("inc");
 
         // Confirming that our handler was called and that the counter property was incremented.
-        assertEquals(1, obj.counter);
+        assertEquals(1, value.get());
     }
 
     @Test
     public void SendWithParamTriggersOnHandler() throws InterruptedException {
-        TestObject obj = new TestObject();
+        AtomicReference<String> value = new AtomicReference<>();
         Transport mockTransport = new MockEchoTransport();
         HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
 
         Action callback = (param) -> {
-            assertNull(obj.message);
-            String message = ((JsonArray) param).get(0).getAsString();
-            obj.message = message;
+            assertNull(value.get());
+            value.set(((JsonArray) param).get(0).getAsString());
         };
         hubConnection.On("inc", callback);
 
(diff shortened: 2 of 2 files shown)


clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -29,28 +29,37 @@ public HubConnection(String url) {
...
-        } catch (URISyntaxException e) {
-            e.printStackTrace();
+        if (transport == null){
+            try {
+                this.transport = new WebSocketTransport(this.url);
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
+            }
+        }
+        else {
```
BrennanConroy: no line break

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -29,28 +29,37 @@ public HubConnection(String url) {
...
     }
 
+    public HubConnection(String url) {
+        this(url, null);
+    }
+
     public void start() throws InterruptedException {
-        _transport.setOnReceive(this.callback);
-        _transport.start();
+        transport.setOnReceive(this.callback);
```
BrennanConroy: `this.`

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
 import static org.junit.Assert.*;
 
 public class HubConnectionTest {
-    // Coming soon!
     @Test
-    public void testEmptyCollection() {
+    public void checkHubConnectionState() throws InterruptedException {
+        Transport mockTransport = new MockEchoTransport();
+        HubConnection hubConnection = new HubConnection("test", mockTransport);
+            hubConnection.start();
```
BrennanConroy: nit: indenting

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
+            assertTrue(hubConnection.connected);
+
+            hubConnection.stop();
+            assertFalse(hubConnection.connected);
+    }
+
+    @Test
+    public void SendWithNoParamsTriggersOnHandler() throws InterruptedException {
+        TestObject obj = new TestObject();
```
BrennanConroy: this should just be an int

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
+        Action callback = (param)-> {
+            assertEquals(0, obj.counter);
+            obj.incrementCounter();
+        };
+        hubConnection.On("inc", callback);
+
+        hubConnection.start();
+        hubConnection.send("inc");
+
+        // Confirming that our handlers was called and that the counter property was incremented.
```
BrennanConroy: handler*

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
+            assertFalse(hubConnection.connected);
+    }
+
+    @Test
+    public void SendWithNoParamsTriggersOnHandler() throws InterruptedException {
+        TestObject obj = new TestObject();
+        Transport mockTransport = new MockEchoTransport();
+        HubConnection hubConnection = new HubConnection("placeholder", mockTransport);
+
+        Action callback = (param)-> {
```
BrennanConroy: space before `->`

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
 
 import static org.junit.Assert.*;
 
 public class HubConnectionTest {
-    // Coming soon!
     @Test
-    public void testEmptyCollection() {
+    public void checkHubConnectionState() throws InterruptedException {
+        Transport mockTransport = new MockEchoTransport();
+        HubConnection hubConnection = new HubConnection("test", mockTransport);
```
BrennanConroy: "http://example.com", I prefer a fake url to be more explicit

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
+        hubConnection.start();
+        hubConnection.send("inc");
+
+        // Confirming that our handlers was called and that the counter property was incremented.
+        assertEquals(1, obj.counter);
+    }
+
+    @Test
+    public void SendWithParamTriggersOnHandler() throws InterruptedException {
+        TestObject obj = new TestObject();
```
BrennanConroy: This can just be a string

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
+        hubConnection.On("inc", callback);
+
+        hubConnection.start();
+        hubConnection.send("inc", "Hello World");
+
+        // Confirming that our handler was called and the correct message was passed in.
+        assertEquals("Hello World", obj.message);
+    }
+
+    private class MockEchoTransport implements Transport{
```
BrennanConroy: nit: space

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
+        private OnReceiveCallBack onReceiveCallBack;
+
+        @Override
+        public void start() throws InterruptedException {return;}
```
BrennanConroy: `{ }`?

clients/java/signalr/src/test/java/HubConnectionTest.java
mikaelm12: So it actually can't just be a string here. It would have to be declared as final, which won't work for this, or as an `AtomicReference<String>`.  I just oped to use the `TestObject ` instead

clients/java/signalr/src/test/java/HubConnectionTest.java
mikaelm12: Same here: https://github.com/aspnet/SignalR/pull/2550#discussion_r198643231

mikaelm12:
Updated

clients/java/signalr/src/main/java/HubConnection.java
BrennanConroy: I think we should do something

clients/java/signalr/src/test/java/HubConnectionTest.java
BrennanConroy: Ok, make a local object then? `var wrapper = new Object(){ int counter = 0; };`

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
 import static org.junit.Assert.*;
 
 public class HubConnectionTest {
-    // Coming soon!
     @Test
-    public void testEmptyCollection() {
+    public void checkHubConnectionState() throws InterruptedException {
+        Transport mockTransport = new MockEchoTransport();
+        HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
+            hubConnection.start();
```
BrennanConroy: didn't fix indenting :(

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,92 @@
...
+            assertTrue(hubConnection.connected);
```
BrennanConroy: I hope `connected` is going to go away in later PRs

mikaelm12: Planning on replacing this with an enum like the other clients

mikaelm12:
Updated

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -5,18 +5,18 @@
 import java.util.HashMap;
 
 public class HubConnection {
-    private String _url;
-    private Transport _transport;
+    private String url;
+    private Transport transport;
```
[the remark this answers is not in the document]
anurse: nice and tidy

mikaelm12:
Can I merge this?  Doesn't look like there are anymore comments. Working in a branch off a branch rn 😅

clients/java/signalr/src/test/java/HubConnectionTest.java
mikaelm12: Java doesn't support anonymous types in the same way C# does. That wouldn't work.

mikaelm12:
Ping

clients/java/signalr/src/test/java/HubConnectionTest.java
BrennanConroy: My concern here is that `TestObject` will become this ugly monolith with 30 different properties/methods to support all tests. It's already bad looking with 2 properties used for different tests. Is there a reason `AtomicReference<T>` wasn't used?

mikaelm12: I thought there  would be push back against using `AtomicReference<T>` but you make a good point. The pr updated and `TestObject` has been removed

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -1,13 +1,86 @@
...
+
+    @Test
+    public void SendWithNoParamsTriggersOnHandler() throws InterruptedException {
+        AtomicInteger value = new AtomicInteger(0);
+        Transport mockTransport = new MockEchoTransport();
+        HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
+
+        Action callback = (param) -> {
+            assertEquals(0, value.get());
+            value.set(value.incrementAndGet());
```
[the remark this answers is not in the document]
mikaelm12: Yup, you're right. Good catch

PR #2564, opened 2018-06-29
[automated] Update branding for 3.0

natemcmaster:
Part of https://github.com/aspnet/Home/issues/3278.
This should updates the branding to the next major version

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,7 +1,7 @@
-<Project>
+﻿<Project>
   <PropertyGroup>
-    <VersionPrefix>1.1.0</VersionPrefix>
-    <VersionSuffix>preview1</VersionSuffix>
+    <VersionPrefix>3.0.0</VersionPrefix>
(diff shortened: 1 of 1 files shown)


muratg:
@DamianEdwards for the versioning question. When it's 2.0, we'll have some customer confusion in our hands (ASP.NET SignalR 2.x vs ASP.NET Core SignalR 2.x) 

cc @anurse

anurse:
I think we had decided to just bite the bullet and jump to `3.0` for SignalR. I still also believe `1.0` was the right version number at release for the same reason (avoiding confusion with ASP.NET SignalR 2.x).

natemcmaster:
Ok, i'll update the PR to change branding to 3.0. I'll get positive confirmation from @DamianEdwards before merging.

natemcmaster:
Confirmed in person with @anurse. 3.0.0 here we go!

PR #2568, opened 2018-06-29
On Overloads for Java Client

mikaelm12:
Making the client more usable!
Adding on overloads for the java client so you can register callbacks that can take more than one argument from an invocation.

One thing to note
In Java there's no way to get the runtime type of a generic. So in C#  where we do
```csharp
hubConnection.On<String>("Foo",  message => {...}
```
We don't have to pass in the class because we can extract that information from the generic. Since we can't do that in Java, the code looks like this
```java
hubConnection.On("Foo", (message) -> { ... }, String.class);
```
It's a little different but ultimately to the user, they still only declare each type once.


clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -55,13 +55,39 @@ public void stop(){
...
     }
 
-    public void send(String method, Object... args) {
+    public void send(String method, Object... args) throws Exception {
         InvocationMessage invocationMessage = new InvocationMessage(method, args);
         String message = protocol.writeMessage(invocationMessage);
         transport.send(message);
     }
 
     public void On(String target, Action callback) {
```
anurse: I think I made this comment on the earlier PR. This should be `camelCased` like the other methods (so `on` instead of `On`) 😸

clients/java/signalr/src/main/java/ActionBase.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public void invoke(Object[] args) throws Exception {
+        if(this.action != null){
+            this.action.invoke();
+            return;
+        }
+
+        // At this point we know we have params sos we initialize an array to store the types to deserialize.
+        ArrayList<Object> types = gson.fromJson((JsonArray)args[0], (new ArrayList<Object>()).getClass());
```
BrennanConroy: What about other protocols

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -55,13 +55,39 @@ public void stop(){
...
-        handlers.put(target, callback);
+        ActionBase actionBase = new ActionBase(callback);
+        handlers.put(target, actionBase);
+    }
+
+    public <T1> void On(String target, Action1<T1> callback, Class<T1> param1) {
```
anurse: Just so I understand, to call this I'd have to write:

```java
connection.on("Foo", (a, b) -> {
    // do stuff with a, b
}, String.class, SomeCustomType.class);
```

Correct? Type Erasure sucks 😢

anurse: I should've just kept reading the PR ;P

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -55,13 +55,39 @@ public void stop(){
...
+        ActionBase actionBase = new ActionBase(callback, param1, param2, param3);
+        handlers.put(target, actionBase);
+    }
+
+    public <T1, T2, T3, T4> void On(String target, Action4<T1,T2, T3, T4> callback, Class<T1> param1, Class<T2> param2, Class<T3> param3, Class<T4> param4) {
+        ActionBase actionBase = new ActionBase(callback, param1, param2, param3, param4);
+        handlers.put(target, actionBase);
+    }
+
+    public <T1, T2, T3, T4,T5> void On(String target, Action5<T1,T2, T3, T4, T5> callback, Class<T1> param1, Class<T2> param2, Class<T3> param3, Class<T4> param4, Class<T5> param5) {
```
anurse: nit spacing after `T4` in `public <T1, T2, T3, T4,T5>`

clients/java/signalr/src/main/java/ActionBase.java
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+
+import java.util.ArrayList;
+
+public class ActionBase {
```
anurse: This seems really gross to me... Is there no other way we can do this? Does Java have something like reflection that would let us just apply an array of arguments to a function?

anurse: Could we build code that wraps any of `Action1` through `Action5` in a different lambda that takes an `Object[]` and then just store that wrapper instance in the list rather than this big class-full-o-pointers?

mikaelm12:
@anurse I did the investigation that you suggested and this currently doesn't support custom types. I'm looking into that and I'm going to file a new work item for that.

mikaelm12:
This is a lot cleaner than before. I replace the `ActionBase` class with an interface that uses variable arguments.

clients/java/signalr/src/main/java/Chat.java
```diff
@@ -1,23 +1,19 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-import com.google.gson.JsonArray;
-
-import java.net.URISyntaxException;
 import java.util.Scanner;
 
 public class Chat {
-    public static void main(String[] args) throws URISyntaxException, InterruptedException {
+    public static void main(String[] args) throws Exception {
```
anurse: This is also know in Java circles as "Fine! I give up!" :trollface:

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -22,35 +20,33 @@ public void checkHubConnectionState() throws InterruptedException {
     }
 
     @Test
-    public void SendWithNoParamsTriggersOnHandler() throws InterruptedException {
-        AtomicInteger value = new AtomicInteger(0);
+    public void SendWithNoParamsTriggersOnHandler() throws Exception {
+        AtomicReference<Double> value = new AtomicReference<Double>(0.0);
```
anurse: Curious why this was changed to an atomic double instead of being left as an integer.

mikaelm12: Gson actually has trouble deserializing directly into `Integer` type. It defaults to `Double`. [Random SO post about the issue](https://stackoverflow.com/questions/17090589/gson-deserialize-integers-as-integers-and-not-as-doubles). 
One work aroud that I don't like would be to check if the desired cast type is Integer and then do the needful to convert the Double into an int but that feels wrong. Going to file an issue for this and custom type deserialization as well.

anurse: Ah, ok. Maybe throw a quick comment in there, but I'm OK with this given the explaination.

Issue #2570, opened 2018-06-29
Format HubConnection Urls for Websocket connections


mikaelm12:
Currently you have to manually format a url for a weboscket connection to start with `ws:` or `wss:` instead of `http:` and `https:`. This should be handled internally in the Websockets transport.

PR #2571, opened 2018-06-29
Formatting urls properly when using websockets

mikaelm12:
The web sockets transport should format urls to establish web sockets connections properly.

Issue: https://github.com/aspnet/SignalR/issues/2570

### DIFF
--- a/clients/java/signalr/src/test/java/WebSocketTransportTest.java
+++ b/clients/java/signalr/src/test/java/WebSocketTransportTest.java
@@ -1,0 +1,33 @@
+import org.junit.Test;
+
+import java.net.URISyntaxException;
+
+import static org.junit.Assert.*;
+
+public class WebSocketTransportTest {
+
+    @Test
+    public void checkThatWsIsUnchanged() throws URISyntaxException {
+        WebSocketTransport webSocketTransport = new WebSocketTransport("ws://example.com");
+        assertEquals("ws://example.com", webSocketTransport.getUrl().toString());
+    }
+
+    @Test
+    public void checkThatWssIsUnchanged() throws URISyntaxException {
+        WebSocketTransport webSocketTransport = new WebSocketTransport("wss://example.com");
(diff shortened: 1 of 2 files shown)


clients/java/signalr/src/test/java/WebSocketTransportTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.Test;
+
+import java.net.URISyntaxException;
+
+import static org.junit.Assert.*;
+
+public class WebSocketTransportTest {
+
+    @Test
+    public void checkThatWsIsUnchanged() throws URISyntaxException {
```
anurse: Looks like junit has parameterized tests. It's parameterized by class instead of by method but that could work fine here (no reason we can't use a nested class like `WebSocketTransportTests.UrlTests`)

https://github.com/junit-team/junit4/wiki/parameterized-tests

mikaelm12: Cool, I'll look into this

mikaelm12:
Updated to use Junit's parametrized tests feature

clients/java/signalr/src/test/java/WebSocketTransportTest.java
```diff
@@ -0,0 +1,35 @@
...
+    private String expectedUrl;
+
+    public WebSocketTransportTest(String url, String expectedProtocol){
+        this.url = url;
+        this.expectedUrl = expectedProtocol;
+    }
+
+    @Parameterized.Parameters
+    public static Collection protocols(){
+        return Arrays.asList(new String[][] {{"http://example.com", "ws://example.com"}, {"https://example.com", "wss://example.com"},
```
anurse: nit: maybe format these out into a line per parameter set?

```java
return Arrays.asList(new String[][] {
    { "http://example.com", "ws://example.com" },
    // ...
});
```

mikaelm12: Sure

Issue #2561, opened 2018-06-28
[2.1.3] Port TryAddSingleton for services #2432


BrennanConroy:
#2432

In 2.1.X we override user services with ours if they set them before `services.AddSignalR();`.
Can be pretty frustrating and difficult to figure out. Simple fix, low risk
@anurse

mikaelm12:
Link to my commit: e41764c161574983c49362cfb29ce23eb5941b74
It's a fairly small code change

muratg:
Approved for 2.1.3.

BrennanConroy:
Done

PR #2434, opened 2018-06-06
Use TryAddSingleton over AddSingleton

mikaelm12:
Quick fix for issue: https://github.com/aspnet/SignalR/issues/2432

PR #2581, opened 2018-07-03
[2.1.3] Update AddSignalRCore to respect user registered services (#2434)

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2561
Porting change from release/2.2

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Tests/AddSignalRTests.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Tests/AddSignalRTests.cs
@@ -1,0 +1,188 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


mikaelm12:
You add a test be sure to also add it to all appropriate branches

BrennanConroy:
Test added

BrennanConroy:
Looking for approval again now that tests are in

test/Microsoft.AspNetCore.SignalR.Tests/AddSignalRTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            Assert.IsType<CustomHubActivator<CustomHub>>(serviceProvider.GetRequiredService<IHubActivator<CustomHub>>());
+            Assert.IsType<CustomHubContext<CustomHub>>(serviceProvider.GetRequiredService<IHubContext<CustomHub>>());
+        }
+    }
+
+    public class CustomHub : Hub
+    {
+    }
+
+    public class CustomTHub : Hub<string>
```
mikaelm12: Hmm, should there be a another test for `DynamicHub ` here as well?

PR #2583, opened 2018-07-03
Adding Support for On Overloads in Java Client (#2568)

mikaelm12:
release/2.2

### DIFF
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -1,15 +1,20 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+
 import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.HashMap;
 
 public class HubConnection {
     private String url;
     private Transport transport;
     private OnReceiveCallBack callback;
-    private HashMap<String, Action> handlers = new HashMap<>();
+    private HashMap<String, ActionBase> handlers = new HashMap<>();
     private HubProtocol protocol;
+    private Gson gson = new Gson();
 
     public Boolean connected = false;
 
@@ -24,7 +29,9 @@ public class HubConnection {
             // Adding this to avoid getting error messages on pings for now.
             for (InvocationMessage message : messages) {
                 if (message != null && handlers.containsKey(message.target)) {
-                    handlers.get(message.target).invoke(message.arguments[0]);
+                    ArrayList<Object> types = gson.fromJson((JsonArray)message.arguments[0], (new ArrayList<Object>()).getClass());
+
+                    handlers.get(message.target).invoke(types.toArray());
                 }
             }
         };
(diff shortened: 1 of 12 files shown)


clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -55,13 +62,48 @@ public void stop(){
...
+        handlers.put(target, ac);
+    }
+
+    public <T1> void on(String target, Action1<T1> callback, Class<T1> param1) {
+        ActionBase ac = params -> callback.invoke(param1.cast(params[0]));
+        handlers.put(target, ac);
+    }
+
+    public <T1, T2> void on(String target, Action2<T1,T2> callback, Class<T1> param1, Class<T2> param2) {
+        ActionBase action = params -> {
```
moozzyk: curious - why the above did not need separate scope for `callback.invoke`?

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -55,13 +62,48 @@ public void stop(){
...
+            callback.invoke(param1.cast(params[0]), param2.cast(params[1]));
+        };
+        handlers.put(target, action);
+    }
+
+    public <T1, T2, T3> void on(String target, Action3<T1,T2, T3> callback, Class<T1> param1, Class<T2> param2, Class<T3> param3) {
```
moozzyk: nit: formatting (`T1,T2, T3`)

clients/java/signalr/src/main/java/Transport.java
```diff
@@ -3,8 +3,8 @@
 
 public interface Transport {
     void start() throws InterruptedException;
-    void send(String message);
+    void send(String message) throws Exception;
```
moozzyk: Doesn't it require catching base exception? Can/should this Exception be of more specific type (possibly even SignalRException)?

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -59,14 +57,134 @@ public void SendWithParamTriggersOnHandler() throws InterruptedException {
...
+    public void SendWithTwoParamsTriggersOnHandler() throws Exception {
+        AtomicReference<String> value1 = new AtomicReference<>();
+        AtomicReference<Double> value2 = new AtomicReference<>();
+
+        Transport mockTransport = new MockEchoTransport();
+        HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
+
+        hubConnection.on("inc", (param1, param2) ->{
+            assertNull(value1.get());
+            assertNull((value2.get()));
```
moozzyk: why these extra brackets?

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -21,36 +19,36 @@ public void checkHubConnectionState() throws InterruptedException {
...
 
     @Test
-    public void SendWithParamTriggersOnHandler() throws InterruptedException {
+    public void SendWithParamTriggersOnHandler() throws Exception {
         AtomicReference<String> value = new AtomicReference<>();
         Transport mockTransport = new MockEchoTransport();
         HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
 
-        Action callback = (param) -> {
+        hubConnection.on("inc", (param) ->{
```
moozzyk: nit: format

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -59,14 +57,134 @@ public void SendWithParamTriggersOnHandler() throws InterruptedException {
...
+
+        // Confirming that our handler was called and the correct message was passed in.
+        assertEquals("A", value1.get());
+        assertEquals("B", value2.get());
+        assertEquals("C", value3.get());
+    }
+
+    @Test
+    public void SendWithFourParamsTriggersOnHandler() throws Exception {
+        AtomicReference<String> value1 = new AtomicReference<>();
```
moozzyk: Is one of the reasons of using AtomicReference here to avoid an error saying that a variable _is not final or effectively final_?

clients/java/signalr/src/main/java/Transport.java
```diff
@@ -3,8 +3,8 @@
 
 public interface Transport {
     void start() throws InterruptedException;
```
moozzyk: I think the InterruptedException here is specific to how WebSockets transport is started. Http based transports might throw different exceptions which you may need to list here. If someone comes up with a different transport (unlikely) that needs to throw yet another exception they won't be able to implement this interface. Similarly if you decide to use a different WebSockets library of Http client you may not be able to switch without breaking users.

clients/java/signalr/src/test/java/HubConnectionTest.java
mikaelm12: Yeah, that's the reason. is there a better way?

moozzyk: I personally find this Java _feature_ limiting. What actually happens if an assert fails inside the lambda? Will the test fail correctly or will it crash the entire suite? I think it will not fail the test because you likely would have already left the method by the time you receive the callback. I wonder if you can actually assert the value inside the `on` method. 

Also, I find this test incorrect. I am not sure how this test does not fail because you never ensure that the callback was actually called before verifying the values. You also never stop the connection which you should. To properly stop the connection (since Java obviously does not have async test methods because it does not have async/await/Task) you need to kind of do what we did in the old SignalR. You use a `countdownLatch` which you release in the callback and which you await (`CountDownLatch.await`) with timeout before stopping the connection. You stop the connection in a `finally` because when the timeout expires it throws and you want to stop the connection regardless. If asserts inside the callback work as expected (i.e. a failed assert just fails the test) and you use the latch then you probably don't need to use AtomicReferences. You will not however be able to verify it the hub method was invoked more than once. You kind of do it now but this is really by accident because you don't close the connection and I don't really know what/how this would fail if you received the second call after the test has finished but the connection is still running on a background thread.

Or maybe AtomicReferences here are just fine (only I believe you still need the latch to block verification and closing the connection until you are really receive the method because what you have now either does not work at all or is super racy.

Issue #2529, opened 2018-06-22
Get JavaScript API docs posted


anurse:
We have JSDoc comments, but the API docs don't actually go anywhere right now :(. There are JavaScript API docs on docs.microsoft.com: https://docs.microsoft.com/en-us/javascript/api/, but I'm not sure exactly how to get them there :). We should follow up with the necessary folks to get the docs posted.

TODO:
* [ ] Review and fix errors in docs
  * [@aspnet/signalr](https://review.docs.microsoft.com/en-us/javascript/api/@aspnet/signalr/?branch=smoke-test)
  * [@aspnet/signalr-protocol-msgpack](https://review.docs.microsoft.com/en-us/javascript/api/@aspnet/signalr-protocol-msgpack/?branch=smoke-test)

BrennanConroy:
Reopening to track merging into release/2.1

Eilon:
## Description

TODO: Describe the bug

## Customer Impact

TODO: Describe how it affects customers. Is there a workaround? How many customers reported?

## Regression from?

TODO: Fill in when it worked, when it stopped working

## Risk

High/Medium/Low/Very Low
TODO: Fill in explanation of risk.

Eilon:
Approved for 2.1.4.

BrennanConroy:
Done in https://github.com/aspnet/SignalR/pull/2605

PR #2605, opened 2018-07-06
Update TS docs for docs.microsoft.com generation

BrennanConroy:


BrennanConroy:
It's already been pushed live
https://docs.microsoft.com/en-us/javascript/api/?view=signalr-js-latest

You have nothing to compare to since we didn't have the docs before.

davidfowl:
Any details about why you had to add @aspnet/signalr everywhere?

BrennanConroy:
Something about how docfx creates the xref links, it needs the fullname otherwise it displays `<xref: HttpResponse>` for example instead of a link.

BrennanConroy:
@Eilon for 2.1.4 consideration

BrennanConroy:
@Eilon https://github.com/aspnet/SignalR/issues/2529

BrennanConroy:
@anurse Could you do a quick pass again before I merge? I changed some things back to internal now that it's supported.

PR #2617, opened 2018-07-11
HubConnection _pendingCalls Change to ConcurrentDictionary

vpaulino:
As the documentation says [here](https://msdn.microsoft.com/en-us/library/dd997373%28v=vs.110%29.aspx): 
"
In general, use a System.Collections.Concurrent.ConcurrentDictionary in any scenario where you are adding and updating keys or values concurrently from multiple threads. In scenarios that involve frequent updates and relatively few reads, the ConcurrentDictionary generally offers modest benefits. In scenarios that involve many reads and many updates, the ConcurrentDictionary generally is significantly faster on computers that have any number of cores.
"

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -1145,11 +1145,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             {
                 Log.CancelingOutstandingInvocations(_hubConnection._logger);
 
-                // https://msdn.microsoft.com/en-us/library/dd287131.aspx
-                // From the documentation:
-                // The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary,
-                // however it does not represent a moment-in-time snapshot of the dictionary.
-                // The contents exposed through the enumerator may contain modifications made to the dictionary after GetEnumerator was called.
+               
                 foreach (var outstandingCallKeyPair in _pendingCalls)
                 {
                     Log.RemovingInvocation(_hubConnection._logger, outstandingCallKeyPair.Value.InvocationId);
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -742,7 +742,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                     }
                 }
             }
-            
+
             // shutdown if we're unable to read handshake
             // Ignore HubException because we throw it when we receive a handshake response with an error
             // And because we already have the error, we don't need to log that the handshake failed
@@ -1129,43 +1129,27 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             public void AddInvocation(InvocationRequest irq)
             {
-                 
-                    
                 if (!_pendingCalls.TryAdd(irq.InvocationId, irq))
                 {
                     Log.InvocationAlreadyInUse(_hubConnection._logger, irq.InvocationId);
                     throw new InvalidOperationException($"Invocation ID '{irq.InvocationId}' is already in use.");
                 }
-                     
-                 
             }
 
-            public bool TryGetInvocation(string invocationId, out InvocationRequest irq)
-            {
-                 
-                    return _pendingCalls.TryGetValue(invocationId, out irq);
-                 
-            }
+            public bool TryGetInvocation(string invocationId, out InvocationRequest irq) => _pendingCalls.TryGetValue(invocationId, out irq);
 
-            public bool TryRemoveInvocation(string invocationId, out InvocationRequest irq)
-            {
-                
+            public bool TryRemoveInvocation(string invocationId, out InvocationRequest irq) => _pendingCalls.TryRemove(invocationId, out irq);
 
-                    return  _pendingCalls.TryRemove(invocationId, out irq);
-                     
-                
-            }
 
             public void CancelOutstandingInvocations(Exception exception)
             {
                 Log.CancelingOutstandingInvocations(_hubConnection._logger);
 
-
-                //https://msdn.microsoft.com/en-us/library/dd287131.aspx
+                // https://msdn.microsoft.com/en-us/library/dd287131.aspx
                 // From the documentation:
-                //The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary,
-                //however it does not represent a moment-in-time snapshot of the dictionary.
-                //The contents exposed through the enumerator may contain modifications made to the dictionary after GetEnumerator was called.
+                // The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary,
+                // however it does not represent a moment-in-time snapshot of the dictionary.
+                // The contents exposed through the enumerator may contain modifications made to the dictionary after GetEnumerator was called.
                 foreach (var outstandingCallKeyPair in _pendingCalls)
                 {
                     Log.RemovingInvocation(_hubConnection._logger, outstandingCallKeyPair.Value.InvocationId);
@@ -1176,7 +1160,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                     outstandingCallKeyPair.Value.Dispose();
                 }
                 _pendingCalls.Clear();
-                
+
             }
 
             public Task StopAsync()
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -1103,7 +1103,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             private TaskCompletionSource<object> _stopTcs;
             private readonly object _lock = new object();
-            private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>(StringComparer.Ordinal);
+            private readonly ConcurrentDictionary<string, InvocationRequest> _pendingCalls = new ConcurrentDictionary<string, InvocationRequest>(StringComparer.Ordinal);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
 
             public void AddInvocation(InvocationRequest irq)
             {
-                lock (_lock)
+
```
mikaelm12: Remove both of these new lines

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
-                        throw new InvalidOperationException($"Invocation ID '{irq.InvocationId}' is already in use.");
-                    }
-                    else
-                    {
-                        _pendingCalls.Add(irq.InvocationId, irq);
-                    }
+                    Log.InvocationAlreadyInUse(_hubConnection._logger, irq.InvocationId);
+                    throw new InvalidOperationException($"Invocation ID '{irq.InvocationId}' is already in use.");
                 }
+
```
mikaelm12: Remove new lines here

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
             }
 
             public bool TryGetInvocation(string invocationId, out InvocationRequest irq)
             {
-                lock (_lock)
-                {
+                 
                     return _pendingCalls.TryGetValue(invocationId, out irq);
-                }
+
```
mikaelm12: Remove new line

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
-                    {
-                        _pendingCalls.Remove(invocationId);
-                        return true;
-                    }
-                    else
-                    {
-                        return false;
-                    }
-                }
+
```
mikaelm12: Remove new lines

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
+                     
+                
             }
 
             public void CancelOutstandingInvocations(Exception exception)
             {
                 Log.CancelingOutstandingInvocations(_hubConnection._logger);
 
-                lock (_lock)
+
```
mikaelm12: Remove new line

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
-                    }
-                    else
-                    {
-                        return false;
-                    }
-                }
+                
+
+                    return  _pendingCalls.TryRemove(invocationId, out irq);
+
```
mikaelm12: Remove new line

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
+                
             }
 
             public void CancelOutstandingInvocations(Exception exception)
             {
                 Log.CancelingOutstandingInvocations(_hubConnection._logger);
 
-                lock (_lock)
+
+                //https://msdn.microsoft.com/en-us/library/dd287131.aspx
```
mikaelm12: Our convention is to put a space after `//` in our comments

mikaelm12: But hmm, this comment while helpful for review, might not be necessary

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
-                            outstandingCall.Fail(exception);
-                        }
-                        outstandingCall.Dispose();
+                        outstandingCallKeyPair.Value.Fail(exception);
                     }
-                    _pendingCalls.Clear();
+                    outstandingCallKeyPair.Value.Dispose();
                 }
+                _pendingCalls.Clear();
+
```
mikaelm12: Remove new line

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,54 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
                 Log.CancelingOutstandingInvocations(_hubConnection._logger);
 
-                lock (_lock)
+
+                //https://msdn.microsoft.com/en-us/library/dd287131.aspx
+                // From the documentation:
+                //The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary,
+                //however it does not represent a moment-in-time snapshot of the dictionary.
+                //The contents exposed through the enumerator may contain modifications made to the dictionary after GetEnumerator was called.
+                foreach (var outstandingCallKeyPair in _pendingCalls)
```
mikaelm12: Is there a reason that you iterate through the key pairs instead of just the values of `_pendingCalls`. It looks like you access the `Value` field in every reference to `outstandingCallKeyPair`.

vpaulino:
Hi there!

Glad to ear feedback! I will not be able to review this topics today ( I have limited internet access today ) but tomorrow I will answer all your questions! :)

mikaelm12:
@vpaulino Awesome!

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
vpaulino: Should I leave it for now or remove it and added to the PR message ?

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
vpaulino: The only reason to access the keypair instead of the collection of values is because the documentation doesn't give any guarantees about the Values thread safety  Enumerator being return. They only  give assurance about the concurrent Dictionary enumerator  being return. 
If you use the values enumerator, that is a not threadsafe copy of the state of those values, that may not represent the reallity of that dictionary value objects wich can be a problem if you call Dispose over an object that may  already  be removed from that dictionary.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -1129,61 +1129,38 @@ public ConnectionState(ConnectionContext connection, HubConnection hubConnection
...
-                    }
-                }
-            }
 
             public void CancelOutstandingInvocations(Exception exception)
             {
                 Log.CancelingOutstandingInvocations(_hubConnection._logger);
 
-                lock (_lock)
+                // https://msdn.microsoft.com/en-us/library/dd287131.aspx
```
mikaelm12: Lets go ahead and remove this 👍

davidfowl: The behavior here has slightly changed. Even though it's safe to enumerate the dictionary itself in a lock, it was being synchronized with other operations (like StopAsync).

vpaulino: I will try to demonstrate what you are saying, but the scope for that unit tests  is not very simple because that is a inner private class.  
Some questions regarding the usage of ConnectionState: 
An instance of ConnectionState exists per connection that lives in a Lock Sync of one unit ( method StartAsyncCore of HubConnection class ) that means that you will have only one thread executing that scope of code at a time no ? So from where will it come more threads that may be executing that code concurrently ?

anurse:
We originally considered a ConcurrentDictionary but decided against it because this is a fairly write-heavy dictionary and as you cited:

> In scenarios that involve frequent updates and relatively few reads, the ConcurrentDictionary generally offers modest benefits.

Do you have data showing a concurrency problem in your scenario?

vpaulino:
Hi there,
So I have made some tests and it looks like lock is more less time consuming than ConcurrentDictionary. 
I have implemented a small console application that prints to the output the time consuming per iteration that compares booth approaches:

https://github.com/vpaulino/Concurrency-Playground

My conclusions are based on the Stopwatch that measure each operations and reports that I generate with VS extension: Concurrency Visualizer 

Looks like bigger are the number of records that we want to manipulate on the dictionary the more time it takes to execute the operations concurrently.

Fell free to comment the project or add some ideas of more use cases to test. if you find some logic error on that test some input is welcome ! :) 

For me if you can reject this PR because it clearly degrades performance. Sorry about that.

anurse:
> For me if you can reject this PR because it clearly degrades performance. Sorry about that.

To be clear, you're good with withdrawing this PR? I'm going to close it since it seems like that's the case, apologies if I misinterpreted your comment.

Thanks for showing interest in improving the product! We're certainly happy to keep exploring PRs like this when we've got some data to show a benefit. If you need help collecting some data, feel free to open an issue to discuss a possible improvement, or just send me a direct email (`[my github username]@microsoft.com`) and I'm happy to help give some guidance on how to collect the necessary data and explore improvements.

Issue #2589, opened 2018-07-03
Parse Mesage Types in HubProtocol


mikaelm12:
Each of the message types needs to be parsed  and handled properly in the HubProtocol. 
`HandshakeResponse`, `InvocationMessage`, `StreamInvocation`, `StreamItem`, `CancelInvocation`, `Completion`, `Error`, `Ping`, `Close`, `HandshakeResponse`.

The `HubMessage` abstraction needs to be introduced here.

mikaelm12:
Done: 25c65a5d85f11d498bbace628f72388c44905943

PR #2620, opened 2018-07-11
Add HubMessage and parse messages from the server into different message types

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/2589
I've added the `HubMessage `abstraction and Enums for the different message types.

### DIFF
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -25,8 +25,6 @@ public class HubConnection {
 
             HubMessage[] messages = protocol.parseMessages(payload);
 
-            // message will be null if we receive any message other than an invocation.
-            // Adding this to avoid getting error messages on pings for now.
             for (HubMessage message : messages) {
                 switch (message.getMessageType()) {
                     case INVOCATION:
@@ -46,11 +44,9 @@ public class HubConnection {
                     case STREAM_ITEM:
                     case CANCEL_INVOCATION:
                     case COMPLETION:
-                    case ERROR:
-                        throw  new UnsupportedOperationException("Received an unsupported message type.");
                     case PING:
-                        //We don't need to do anything in the case of a ping message.
-                        //The other message types aren't supported
+                        // We don't need to do anything in the case of a ping message.
+                        // The other message types aren't supported
                         break;
                 }
             }
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -23,19 +23,35 @@ public class HubConnection {
         this.protocol = new JsonHubProtocol();
         this.callback = (payload) -> {
 
-            InvocationMessage[] messages = protocol.parseMessages(payload);
+            HubMessage[] messages = protocol.parseMessages(payload);
 
             // message will be null if we receive any message other than an invocation.
             // Adding this to avoid getting error messages on pings for now.
-            for (InvocationMessage message : messages) {
-                if (message != null && handlers.containsKey(message.target)) {
-                    ArrayList<Object> args = gson.fromJson((JsonArray)message.arguments[0], (new ArrayList<Object>()).getClass());
-                    List<ActionBase> actions = handlers.get(message.target);
-                    if (actions != null) {
-                        for (ActionBase action: actions) {
-                            action.invoke(args.toArray());
+            for (HubMessage message : messages) {
+                switch (message.getMessageType()) {
+                    case INVOCATION:
+                        InvocationMessage invocationMessage = (InvocationMessage)message;
+                        if (message != null && handlers.containsKey(invocationMessage.target)) {
+                            ArrayList<Object> args = gson.fromJson((JsonArray)invocationMessage.arguments[0], (new ArrayList<Object>()).getClass());
+                            List<ActionBase> actions = handlers.get(invocationMessage.target);
+                            if (actions != null) {
+                                for (ActionBase action: actions) {
+                                    action.invoke(args.toArray());
+                                }
+                            }
                         }
-                    }
+                        break;
+                    case CLOSE:
+                    case STREAM_INVOCATION:
+                    case STREAM_ITEM:
+                    case CANCEL_INVOCATION:
+                    case COMPLETION:
+                    case ERROR:
+                        throw  new UnsupportedOperationException("Received an unsupported message type.");
+                    case PING:
+                        //We don't need to do anything in the case of a ping message.
+                        //The other message types aren't supported
+                        break;
                 }
             }
         };
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -39,9 +39,10 @@ public class HubConnection {
                             }
                         }
                         break;
-                    case CLOSE:
                     case STREAM_INVOCATION:
                     case STREAM_ITEM:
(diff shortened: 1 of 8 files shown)


BrennanConroy:
Shouldn't this target release/2.2

mikaelm12:
>Shouldn't this target release/2.2

Yes, yes it should

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -23,19 +23,35 @@ public HubConnection(String url, Transport transport){
...
+                            }
                         }
-                    }
+                        break;
+                    case CLOSE:
+                    case STREAM_INVOCATION:
+                    case STREAM_ITEM:
+                    case CANCEL_INVOCATION:
+                    case COMPLETION:
+                    case ERROR:
```
BrennanConroy: I'm not familiar with the `ERROR` message type

clients/java/signalr/src/main/java/HubMessageType.java
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public enum HubMessageType {
+    INVOCATION,
```
BrennanConroy: Assign values to these

clients/java/signalr/src/main/java/JsonHubProtocol.java
```diff
@@ -46,35 +46,34 @@ public TransferFormat getTransferFormat() {
                     //Invocation Message
                     String target = jsonMessage.get("target").getAsString();
                     JsonElement args = jsonMessage.get("arguments");
-                    invocationMessages.add(new InvocationMessage(target, new Object[] {args}));
+                    hubMessages.add(new InvocationMessage(target, new Object[] {args}));
+
                     break;
                 case "2":
```
BrennanConroy: Can we use the enum values now?

clients/java/signalr/src/main/java/InvocationMessage.java
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-public class InvocationMessage {
+public class InvocationMessage extends HubMessage {
     int type = 1;
```
BrennanConroy: This should now be of `HubMessageType`

mikaelm12: `HubMessageType` is just the enum tied to each of the specific message types (InvocationMessage, StreamItem etc). The actual base class is HubMessage

BrennanConroy: ? `type` should be `HubMessageType` instead of `int`

clients/java/signalr/src/main/java/HubConnection.java
mikaelm12: Lol. Got a little trigger happy with message types and created a new one

clients/java/signalr/src/main/java/InvocationMessage.java
mikaelm12: Oh, my comment was referring to line 4, now realizing you were talking about line 5

mikaelm12: Yeah, was going to give a long explanation about why this is the way it is, but I think I'll be able to change it once I give the enums values. That might not work how I expect though since enums in java are slightly more complex

mikaelm12: Essentially I had the type as an int field for serialization. Then an abstract method on the base type `HubMessage` which is used at the `HubConnection` level

clients/java/signalr/src/main/java/JsonHubProtocol.java
mikaelm12: So here in the HubProtocol level I just pull the value right from the JSON and construct the specific HubMessage type but in a list of the base type `HubMessage`. 
The enum values are used in the HubConnection layer when we actually need to cast to the specific message type and need to decide what to do with each message of each type

mikaelm12:
Updated

mikaelm12:
Ping

clients/java/signalr/src/main/java/InvocationMessage.java
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-public class InvocationMessage {
-    int type = 1;
+public class InvocationMessage extends HubMessage {
+    int type = HubMessageType.INVOCATION.value;
```
BrennanConroy: JsonHubProtocol.writeMessage will have to be rewritten eventually and when that happens you can change this to HubMessageType then

mikaelm12:
Updated

mikaelm12:
😄

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -23,19 +23,32 @@ public HubConnection(String url, Transport transport){
...
+                            }
                         }
-                    }
+                        break;
+                    case STREAM_INVOCATION:
+                    case STREAM_ITEM:
+                        throw new UnsupportedOperationException("Streaming is not yet supported");
+                    case CLOSE:
+                    case CANCEL_INVOCATION:
+                    case COMPLETION:
```
jkotalik: Shouldn't these 3 enum values `throw UnsupportedOperationException()`?

jkotalik: Or at least log something to say these message type are unsupported.

clients/java/signalr/src/main/java/InvocationMessage.java
jkotalik: I don't see where type is used.

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -23,19 +23,32 @@ public HubConnection(String url, Transport transport){
...
-                if (message != null && handlers.containsKey(message.target)) {
-                    ArrayList<Object> args = gson.fromJson((JsonArray)message.arguments[0], (new ArrayList<Object>()).getClass());
-                    List<ActionBase> actions = handlers.get(message.target);
-                    if (actions != null) {
-                        for (ActionBase action: actions) {
-                            action.invoke(args.toArray());
+            HubMessage[] messages = protocol.parseMessages(payload);
+
+            for (HubMessage message : messages) {
+                switch (message.getMessageType()) {
```
jkotalik: There may be a performance or design reason behind this, but why have this enum when you can just call a generic method on each message type, like:
```java
    message.Handle(params);
}
```

And then in the class itself invoke the actions.

jkotalik: I see that this is how it was done in the C# client, so you can probably dismiss this comment.

clients/java/signalr/src/main/java/InvocationMessage.java
mikaelm12: We need this field for when we serialize the invocation message and send it.

clients/java/signalr/src/main/java/HubConnection.java
mikaelm12: Yeah, I guess they should all throw `UnsupportedOperationException`s for now

PR #2622, opened 2018-07-12
Remove test dependency on Microsoft.Extensions.Buffers.Testing.Sources

natemcmaster:
This source code use to live in common. This puts a copy of the source code here instead of pulling it via package reference.

cref https://github.com/aspnet/Common/pull/386


natemcmaster:
Closing in favor of this: https://github.com/aspnet/KestrelHttpServer/pull/2717/commits/128eefdef3825e1a2b8313f76b235c465b10b682, which means the .Sources package will still be produced, but it comes from kestrel instead of common.

PR #2627, opened 2018-07-13
HubConnection state enum

mikaelm12:
Changing how we expose the HubConnection connection status from a bool to an enum

### DIFF
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -16,7 +16,7 @@ public class HubConnection {
     private HubProtocol protocol;
     private Gson gson = new Gson();
 
-    public Boolean connected = false;
+    public HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
 
     public HubConnection(String url, Transport transport){
         this.url = url;
@@ -71,12 +71,12 @@ public class HubConnection {
     public void start() throws InterruptedException {
         transport.setOnReceive(this.callback);
         transport.start();
-        connected = true;
+        connectionState = HubConnectionState.CONNECTED;
     }
 
     public void stop(){
         transport.stop();
-        connected = false;
+        connectionState = HubConnectionState.DISCONNECTED;
     }
 
     public void send(String method, Object... args) throws Exception {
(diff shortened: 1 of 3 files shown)


clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -16,7 +16,7 @@
     private HubProtocol protocol;
     private Gson gson = new Gson();
 
-    public Boolean connected = false;
+    public HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
```
jkotalik: AFAIK, in Java it is more common to expose public members through getters and setters, like `getConnectionState()` and `setConnectionState()`

clients/java/signalr/src/main/java/HubConnectionState.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public enum HubConnectionState {
+    CONNECTED,
```
jkotalik: Will there be other ConnectionStates besides these two? Otherwise I don't see the motivation for this change.

clients/java/signalr/src/main/java/HubConnection.java
BrennanConroy: Also, this shouldn't be public, we don't expose connection state in any of the other clients

clients/java/signalr/src/main/java/HubConnectionState.java
BrennanConroy: There should be in the future

clients/java/signalr/src/main/java/HubConnection.java
mikaelm12: > it is more common to expose public members through getters and setters, like getConnectionState() and setConnectionState()
Yup, I'll change it to use a getter/setter

mikaelm12: > we don't expose connection state in any of the other clients

We do in the C# client
https://github.com/aspnet/SignalR/blob/2d4fb0af6fd2ef2e3a059a2a15a56484d8800d35/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L82-L98

BrennanConroy: Oh, that's new. Since 2.2 I guess we do...

mikaelm12:
Merged the base PR and this is now pointing to release/2.2 with the requested changes.

Issue #2615, opened 2018-07-10
Add ability to remove handlers from HubConnection


mikaelm12:
The title pretty much says it all 😄 
We just need to add the functionality to the `CallbackMap` type and then add the API to the `HubConnection`
The equivalent in the C# client is 
https://github.com/aspnet/SignalR/blob/46c8ba5fdc881f19a0384613510e71290719503d/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L212-L221

mikaelm12:
Done: 8919d3897ed854287f4602cddbbaa4f671f1a37d and additionally: e9151906e1668ce679ffacebeb422f3d57a7418c

PR #2628, opened 2018-07-13
Remove handlers from HubConnection

mikaelm12:
This pr lets users remove all handlers from a HubConnection for a certain target. 
I'll add the ability to remove specific handlers with a `Subscription` type similar to what I added in the C# where where we dispose that object to remove the associated single handler.
Issue: https://github.com/aspnet/SignalR/issues/2615

### DIFF
--- a/clients/java/signalr/src/main/java/CallbackMap.java
+++ b/clients/java/signalr/src/main/java/CallbackMap.java
@@ -24,5 +24,8 @@ public class CallbackMap {
     public List<ActionBase> get(String key){
         return handlers.get(key);
     }
+
+    public void remove(String key) { handlers.remove(key); }
+
 }
 
--- a/clients/java/signalr/src/main/java/CallbackMap.java
+++ b/clients/java/signalr/src/main/java/CallbackMap.java
@@ -17,11 +17,11 @@ public class CallbackMap {
         handlers.computeIfAbsent(target, (ac) -> new ArrayList<>(Arrays.asList(action)));
     }
 
-    public Boolean containsKey(String key){
+    public Boolean containsKey(String key) {
         return handlers.containsKey(key);
     }
 
-    public List<ActionBase> get(String key){
+    public List<ActionBase> get(String key) {
         return handlers.get(key);
     }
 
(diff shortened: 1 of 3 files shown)


clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -39,6 +39,76 @@ public void RegisteringMultipleHandlersAndBothGetTriggered() throws Exception {
         assertEquals(2, value.get(), 0);
     }
 
+    @Test
+    public void RemoveHandlerByName() throws Exception {
+
+        AtomicReference<Double> value = new AtomicReference<>(0.0);
```
jkotalik: Can you put this all in a `@BeforeTest` to avoid duplication?

clients/java/signalr/src/main/java/CallbackMap.java
```diff
@@ -24,5 +24,8 @@ public Boolean containsKey(String key){
     public List<ActionBase> get(String key){
         return handlers.get(key);
     }
+
+    public void remove(String key) { handlers.remove(key); }
```
jkotalik: Nit: Stay consistent with indentation.

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -39,6 +39,76 @@ public void RegisteringMultipleHandlersAndBothGetTriggered() throws Exception {
...
+        Transport mockTransport = new MockEchoTransport();
+        HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
+        Action action = () -> value.getAndUpdate((val) -> val + 1);
+        Action secondAction = () -> value.getAndUpdate((val) -> val + 2);
+
+        hubConnection.on("inc", action);
+        hubConnection.on("inc", secondAction);
+
+
```
jkotalik: nit: extra line.

clients/java/signalr/src/test/java/HubConnectionTest.java
mikaelm12: `@Before` runs before every test method and the method have different set up needs so a one size fits all `@Before` doesn't really work here.

jkotalik: Maybe put them in separate classes? Besides that's I don' see much either.

mikaelm12: Yeah, thought about that but I don't really think it's worth it tbh

PR #2644, opened 2018-07-16
Remove handlers from HubConnection (#2628)

mikaelm12:
Merging back from 2.2


mikaelm12:
(History looks cleaner if I remember and just do it instead of merging the bots pr.)

Issue #2257, opened 2018-05-11
If the HubProtocol throws while writing it can leave partial messages in the transport pipe which will get sent on the next successful write


BrennanConroy:
See: https://github.com/aspnet/SignalR/issues/2255#issuecomment-388415398

Because we pass the pipe directly to the HubProtocol it will start writing a message and if an error occurs then the pipe will have the unfinished message. The next write that comes in will start writing where the failed message left off and send both the partial and new message, which the client wont understand.

anurse:
From discussions with @BrennanConroy last week, we think there isn't really a way we can solve this. Even if we could, the connection is in a broken state. We failed to send a message that the client may be expecting (it could be the completion message that would unblock the client's `Task`, for example). We should just Abort the whole connection in this case.

PR #2654, opened 2018-07-17
Abort connection on protocol error

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2257

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -36,6 +36,7 @@ namespace Microsoft.AspNetCore.SignalR
         private bool _receivedMessageThisInterval = false;
         private ReadOnlyMemory<byte> _cachedPingMessage;
         private bool _clientTimeoutActive;
+        private bool _connectedAborted = true;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="HubConnectionContext"/> class.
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionHandler.cs
```diff
@@ -202,7 +203,7 @@ HubDispatcher<THub> dispatcher
                     {
                         connection.ResetClientTimeout();
 
-                        while (protocol.TryParseMessage(ref buffer, _dispatcher, out var message))
+                        while (!abortedToken.IsCancellationRequested && protocol.TryParseMessage(ref buffer, _dispatcher, out var message))
```
davidfowl: Why do we need this?

BrennanConroy: We chatted and we're going to check the cancellation inside HubConnectionContext and noop writes instead of using the token in multiple places

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -312,6 +331,8 @@ public virtual void Abort()
                 return;
             }
 
+            _connectedAborted = false;
```
anurse: This is backwards now right? If the name is `_connectionAborted`, it should be being set to `true` here :)

src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -36,6 +36,7 @@ public class HubConnectionContext
         private bool _receivedMessageThisInterval = false;
         private ReadOnlyMemory<byte> _cachedPingMessage;
         private bool _clientTimeoutActive;
+        private bool _connectedAborted = true;
```
davidfowl: Why is this true by default?

Issue #2621, opened 2018-07-11
Properly construct and parse Handshake request and response in the Java client


mikaelm12:
Right now this is kind of hacked together. The client doesn't properly wait for the handshake response and there is no way to configure the handshake message.

anurse:
https://github.com/aspnet/SignalR/commit/60c5392fd5ac6b5ec2b3bda7957daac1f5aecea2

PR #2656, opened 2018-07-18
Parse HandshakeResponse

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/2621
Properly parsing the handshake response.


clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -71,10 +78,12 @@ public HubConnectionState getConnectionState() {
         return connectionState;
     }
 
-    public void start() throws InterruptedException {
+    public void start() throws Exception {
         transport.setOnReceive(this.callback);
         transport.start();
-        connectionState = HubConnectionState.CONNECTED;
+        HandshakeRequestMessage handshakeRequest = new HandshakeRequestMessage(protocolName, protocolVersion);
+        String handshakeMessage = protocol.writeMessage(handshakeRequest);
```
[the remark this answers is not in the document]
mikaelm12: True

clients/java/signalr/src/main/java/HandshakeRequestMessage.java
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public class HandshakeRequestMessage extends HubMessage{
```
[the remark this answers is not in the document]
mikaelm12: It does in the C# client.
https://github.com/aspnet/SignalR/blob/056da5114aeb201e805e8aa03eb9f6e32bdb9bf1/src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeResponseMessage.cs#L9

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -17,6 +17,10 @@
     private Gson gson = new Gson();
     private HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
 
+    // Only json is supported currently
+    private String protocolName = "json";
```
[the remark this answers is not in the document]
mikaelm12: Well it's hard coded now but later this will be configurable and set in the HubConnection ctor

clients/java/signalr/src/main/java/HandshakeRequestMessage.java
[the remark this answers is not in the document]
anurse: Hrm, that's a bit sketchy since it isn't part of the Hub Protocol. I'm not sure why it does in C#, but I don't think we should continue that trend unless there's a very good reason to do it.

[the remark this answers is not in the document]
anurse: also nit: spacing before `{`

clients/java/signalr/src/main/java/HandshakeResponseMessage.java
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public class HandshakeResponseMessage extends HubMessage{
```
anurse: nit: spacing before `{`s in this file

clients/java/signalr/src/main/java/HandshakeResponseMessage.java
```diff
@@ -0,0 +1,19 @@
...
+    public HandshakeResponseMessage(){
+        this(null);
+    }
+
+    public HandshakeResponseMessage(String error){
+        this.error = error;
+    }
+
+    @Override
+    HubMessageType getMessageType() {
```
anurse: It shouldn't have a "Message Type" because it's not part of the hub protocol

clients/java/signalr/src/main/java/HubConnection.java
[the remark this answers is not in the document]
anurse: It's not part of the Hub Protocol at all. If you look at the C# code, it's a separate "HandshakeProtocol" that does the work.

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -40,14 +44,17 @@ public HubConnection(String url, Transport transport) {
...
-                        throw new UnsupportedOperationException("Streaming is not yet supported");
                     case CLOSE:
                     case CANCEL_INVOCATION:
                     case COMPLETION:
+                        throw new UnsupportedOperationException("The message type" + message.getMessageType() + " is not supported yet.");
                     case PING:
                         // We don't need to do anything in the case of a ping message.
                         // The other message types aren't supported
                         break;
+                    case HANDSHAKE_RESPONSE:
```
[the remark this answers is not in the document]
anurse: These should be handled in a separate process. The callback should probably look something like:

```java
this.handshakeCompleted = false;
    if(!this.handshakeCompleted) {
        // parses using HandshakeProtocol and sets this.handshakeCompleted
        processHandshakeResponse(payload);
    } else {
        // basically just what you have right now
        processHubMessages(payload);
    }
}
```

mikaelm12:
The HandshakeResponse being a sub class of HubMessage in C# kind of threw off my thinking of how this should look. I'm going to re work it

mikaelm12:
Updated

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -15,21 +15,38 @@
...
+
     public HubConnection(String url, Transport transport) {
         this.url = url;
         this.protocol = new JsonHubProtocol();
         this.callback = (payload) -> {
 
+            if (!handshakeReceived) {
+                int handshakeLength = payload.indexOf(RECORD_SEPARATOR) + 1;
+                payload = payload.substring(handshakeLength);
+                handshakeReceived = true;
```
BrennanConroy: You're ignoring the handshake response, it could be an error

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -15,21 +15,38 @@
     private CallbackMap handlers = new CallbackMap();
     private HubProtocol protocol;
     private Gson gson = new Gson();
+    private Boolean handshakeReceived = false;
+    private static final String RECORD_SEPARATOR = "\u001e";
     private HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
 
+    // Only json is supported currently
+    private String protocolName = "json";
```
BrennanConroy: These belong on the protocol class, see https://github.com/aspnet/SignalR/blob/release/2.2/clients/ts/signalr/src/JsonHubProtocol.ts#L16-L18

clients/java/signalr/src/test/java/JsonHubProtocolTest.java
```diff
@@ -84,19 +84,6 @@ public void ParseSingleUnsupportedStreamInvocationMessage() {
         HubMessage[] messages = jsonHubProtocol.parseMessages(stringifiedMessage);
     }
 
-    @Test
-    public void ParseHandshakeResponsePlusMessage() {
```
BrennanConroy: You should add a HubConnection test for receiving a HandshakeResponse + a message

clients/java/signalr/src/main/java/HandshakeProtocol.java
```diff
@@ -0,0 +1,14 @@
...
+import com.google.gson.Gson;
+
+public class HandshakeProtocol {
+    public static Gson gson = new Gson();
+    private static final String RECORD_SEPARATOR = "\u001e";
+
+    public static String createHandshakeRequestMessage(HandshakeRequestMessage message){
+        // The handshake request is always in the JSON format
+        return gson.toJson(message) + RECORD_SEPARATOR;
+    }
```
anurse: We should also add parsing logic to parse a `String` into `HandshakeResponseMessage`

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -15,21 +15,38 @@
...
+
             HubMessage[] messages = protocol.parseMessages(payload);
 
             for (HubMessage message : messages) {
                 switch (message.getMessageType()) {
                     case INVOCATION:
                         InvocationMessage invocationMessage = (InvocationMessage)message;
                         if (message != null && handlers.containsKey(invocationMessage.target)) {
-                            ArrayList<Object> args = gson.fromJson((JsonArray)invocationMessage.arguments[0], (new ArrayList<Object>()).getClass());
+                            ArrayList<Object> args = gson.fromJson((JsonArray)invocationMessage.arguments[0], (new ArrayList<>()).getClass());
```
anurse: Curious why `new ArrayList<Object>` changed to `new ArrayList<>`. Is this intentional?

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -40,10 +57,10 @@ public HubConnection(String url, Transport transport) {
                         break;
                     case STREAM_INVOCATION:
                     case STREAM_ITEM:
-                        throw new UnsupportedOperationException("Streaming is not yet supported");
                     case CLOSE:
                     case CANCEL_INVOCATION:
                     case COMPLETION:
+                        throw new UnsupportedOperationException("The message type" + message.getMessageType() + " is not supported yet.");
```
anurse: nit: you need a space inside the string after `the message type`

clients/java/signalr/src/main/java/HubConnection.java
mikaelm12: Yeah, the `Object` on the other side is not needed so I just took it out.  Nothing to do with this change specifically

anurse: Cool, sounds good.

clients/java/signalr/src/main/java/HandshakeProtocol.java
mikaelm12: Added a `parseHandshakeResponse` method here. Adding unit tests for the HandshakeProtocol now

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -15,21 +15,39 @@
...
         this.url = url;
         this.protocol = new JsonHubProtocol();
         this.callback = (payload) -> {
 
+            if (!handshakeReceived) {
+                int handshakeLength = payload.indexOf(RECORD_SEPARATOR) + 1;
+                String handshakeResponseString = payload.substring(0, handshakeLength - 1);
+                HandshakeResponseMessage handshakeResponse = HandshakeProtocol.parseHandshakeResponse(handshakeResponseString);
+                if (handshakeResponse.error != null){
+                    throw new Exception("Error in handshake" + handshakeResponse.error);
```
anurse: nit: space after "handshake", otherwise you'll get "Error in handshakeerror message" 😄

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -15,21 +15,39 @@
...
+                }
+                handshakeReceived = true;
+
+                // We only received the handshake response ,so we can return.
```
anurse: mega-nit: spacing around "so" ;)

clients/java/signalr/src/main/java/HubConnection.java
anurse: This is why string interpolation is so much better ;P. #javajabs

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -369,14 +370,39 @@ public void SendWithFiveParamsTriggersOnHandler() throws Exception {
         assertEquals(12, value5.get(), 0);
     }
 
+    @Test
+    public void ReceiveHandshakeResponseAndMessage() throws Exception {
```
anurse: nit: `receiveHandshakeResponseAndMessage` (Java methods are `camelCase` usually, and the others are ;))

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -393,4 +419,35 @@ public void onReceive(String message) throws Exception {
         @Override
         public void stop() {return;}
     }
+
+    private class HandshakeTransport implements Transport {
```
anurse: I think it would be cleaner and more reusable to just build a proper `MockTransport` like we have in C# (in C# it's `TestConnection` that does this work). It would store a list of all the messages that have been sent so far and provide a method that could trigger `onReceive`. So a test would look something like:

```java
MockTransport transport = new MockTransport();
HubConnection hubConnection = new HubConnection("http://example.com", transport);

hubConnection.on("inc", () ->{
    assertEquals(0.0, value.get(), 0);
    value.getAndUpdate((val) -> val + 1);
});

hubConnection.start();

assertEqual(transport.getSentMessages()[0], "{... handshake request ...}");

transport.receiveMessage("{ ... handshake response ... }" + RECORD_SEPARATOR + "{ ... invocation ... }" + RECORD_SEPARATOR);
```

This `HandshakeTransport` is working fine, but seems a little funky, with the message in `send` being ignored, and it's kinda just a single-purpose component.

mikaelm12: Yeah, the public `receiveMessage ` on the transport is what I need 😄

mikaelm12: That cleans up some of the trickery in the test transports

mikaelm12: I'll add the method in the MockTransport and change the new tests to use it for now. I'll update the rest of the tests in a later PR.

mikaelm12:
This is updated with a refactored `MockTransport`.

mikaelm12:
I realize we're all busy with the hackathon but it would be **awesome** if I could get a last review and get this in, because I could also merge https://github.com/aspnet/SignalR/pull/2660 and then have a another pr ready  for when we're back on our regular work schedules 😄

clients/java/signalr/src/main/java/HandshakeResponseMessage.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public class HandshakeResponseMessage {
+    public String error;
+
+    public HandshakeResponseMessage(){
```
anurse: nit: `{` spacing. Do IntelliJ/VSCode have auto formatters?

clients/java/signalr/src/main/java/HandshakeResponseMessage.java
```diff
@@ -0,0 +1,14 @@
...
+        this(null);
+    }
+
+    public HandshakeResponseMessage(String error){
```
anurse: nit: `{` spacing

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -15,21 +15,39 @@
...
+                String handshakeResponseString = payload.substring(0, handshakeLength - 1);
+                HandshakeResponseMessage handshakeResponse = HandshakeProtocol.parseHandshakeResponse(handshakeResponseString);
+                if (handshakeResponse.error != null){
+                    throw new Exception("Error in handshake " + handshakeResponse.error);
+                }
+                handshakeReceived = true;
+
+                payload = payload.substring(handshakeLength);
+                // The payload only contained the handshake response so we can return.
+                if (payload.length() == 0){
```
anurse: nit: `{` spacing

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -15,21 +15,39 @@
...
     public HubConnection(String url, Transport transport) {
         this.url = url;
         this.protocol = new JsonHubProtocol();
         this.callback = (payload) -> {
 
+            if (!handshakeReceived) {
+                int handshakeLength = payload.indexOf(RECORD_SEPARATOR) + 1;
+                String handshakeResponseString = payload.substring(0, handshakeLength - 1);
+                HandshakeResponseMessage handshakeResponse = HandshakeProtocol.parseHandshakeResponse(handshakeResponseString);
+                if (handshakeResponse.error != null){
```
anurse: nit: `{` spacing

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -369,15 +395,39 @@ public void SendWithFiveParamsTriggersOnHandler() throws Exception {
...
+            value.getAndUpdate((val) -> val + 1);
+        });
+
+        // On start we're going to receive the handshake response and also an invocation in the same payload.
+        hubConnection.start();
+
+        String expectedSentMessage  = "{\"protocol\":\"json\",\"version\":1}" + RECORD_SEPARATOR;
+        assertEquals(expectedSentMessage, transport.getSentMessages()[0]);
+
+        transport.receiveMessage("{}" + RECORD_SEPARATOR +"{\"type\":1,\"target\":\"inc\",\"arguments\":[]}" + RECORD_SEPARATOR);
```
anurse: nit: `+` spacing (the second one)

definitely seems like an auto-formatter would save a bunch of time here ;P

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -32,7 +35,9 @@ public void RegisteringMultipleHandlersAndBothGetTriggered() throws Exception {
         assertEquals(0.0, value.get(), 0);
 
         hubConnection.start();
-        hubConnection.send("inc");
+
+        mockTransport.receiveMessage("{}" + RECORD_SEPARATOR);
+        mockTransport.receiveMessage("{\"type\":1,\"target\":\"inc\",\"arguments\":[]}" + RECORD_SEPARATOR);
 
         // Confirming that our handler was called and that the counter property was incremented.
         assertEquals(2, value.get(), 0);
```
anurse: Looks like we're only testing what the client _receives_ here, and not asserting what the client _sends_ to make sure it's correct. The new handshake test does this, but the existing tests haven't been modified to do that. Previously the echo transport did a reasonable job of this because it would only echo back what was actually sent. Now, the client could send absolutely nothing and the test would still pass.

mikaelm12: I'll get the sent message and verify that as well

PR #2679, opened 2018-07-25
Parse HandshakeResponse (#2656)

mikaelm12:
Merge from release/2.2


clients/java/signalr/src/main/java/HandshakeProtocol.java
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import com.google.gson.Gson;
+
+public class HandshakeProtocol {
+    public static Gson gson = new Gson();
```
moozzyk: Why is this public? Seems like implementation details that should have at most package access.

mikaelm12: This should be private

PR #2690, opened 2018-07-26
Test and package Java client with build.cmd

mikaelm12:
Don't merge yet. 
Starting the process of integrating the Java client in our build process. Step 1 is just getting tests running with `build.cmd`.
This doesn't check yet if the machine has java installed before trying to run. Doing that should be the next step.


build/repo.targets
```diff
@@ -71,6 +76,11 @@
     <Exec Command="npm run build" WorkingDirectory="$(RepositoryRoot)clients/ts/FunctionalTests" IgnoreStandardErrorWarningFormat="true" />
   </Target>
 
+<Target Name="BuildJavaClient">
+   <Message Text="Building Java client" />
```
anurse: Add `Importance="high"` like the others, otherwise it doesn't show at the default MSBuild verbosity.

anurse:
We may not want to wait on getting Java installed on Jenkins since that's non-trivial. For VSTS it's apparently a simple change to the YML file.

anurse:
For reviewers; CI failures are because none of the PR CI configurations have Java yet :)

mikaelm12:
After hours of fiddling with this, I think I got what we want.
I got `signalr-client-0.1.0-sources.jar`, `signalr-client-0.1.0-javadoc.jar` (with empty template docs)  and the POM files to generate with `build.cmd`. I also renamed the root project name to `signalr-client`

clients/java/signalr/build.gradle
```diff
@@ -16,3 +17,32 @@ dependencies {
...
+            inceptionYear '2018'
+            licenses {
+                license {
+                    name 'The Apache Software License, Version 2.0'
+                    url 'http://www.apache.org/licenses/LICENSE-2.0.txt'
+                    distribution 'repo'
+                }
+            }
+        }
+    }.writeTo("pom.xml")
```
mikaelm12: Do we want to give this a different name/put this in a different place?

build/repo.targets
```diff
@@ -83,4 +93,12 @@
     <Move SourceFiles="%(NPMPackage.OutputTar)" DestinationFiles="%(NPMPackage.ArtifactPath)" />
     <Move SourceFiles="%(NPMPackage.PackageJson).bak" DestinationFiles="%(NPMPackage.PackageJson)" />
   </Target>
+
+<Target Name="PackJavaClient">
+  <Message Text="Packing Java client" Importance="high" />
+  <Exec Command="gradlew jar" WorkingDirectory="$(RepositoryRoot)clients/java/signalr" />
+  <Exec Command="gradlew sourceJar" WorkingDirectory="$(RepositoryRoot)clients/java/signalr" />
+  <Exec Command="gradlew javadocJar" WorkingDirectory="$(RepositoryRoot)clients/java/signalr" />
+  <Exec Command="gradlew generatePOM" WorkingDirectory="$(RepositoryRoot)clients/java/signalr" />
```
anurse: I believe you can run multiple tasks by specifying them on a single command line `gradlew jar sourceJar javadocJar generatePOM`.

build/repo.targets
```diff
@@ -71,8 +76,13 @@
     <Exec Command="npm run build" WorkingDirectory="$(RepositoryRoot)clients/ts/FunctionalTests" IgnoreStandardErrorWarningFormat="true" />
   </Target>
 
+<Target Name="BuildJavaClient">
```
anurse: Need some indentation clean up in this file 😸

clients/java/signalr/build.gradle
```diff
@@ -16,3 +17,32 @@ dependencies {
...
+}
+
+task generatePOM {
+    pom {
+        project {
+            groupId 'com.microsoft.aspnetcore'
+            artifactId 'signalr'
+            version '0.1.0'
+
+            inceptionYear '2018'
```
anurse: Top Spinning from the movie "Inception"

clients/java/signalr/build.gradle
anurse: Call it `signalr-client-0.1.0-preview1.pom` (like the JAR files) and have the gradle build drop it into the `build/libs` directory along side them (so we have a directory that can just be copied over to `artifacts`. Then, add an MSBuild [Copy](https://msdn.microsoft.com/en-us/library/3e54c37h.aspx) Task after the `Exec` tasks to copy them to `$(BuildDir)`.

That should cause the build to start failing due to "undeclared artifacts", which is good, it means the artifacts are in the right place. Then you just need to define `ArtifactInfo` MSBuild items for them. @natemcmaster can help out a bit there too.

Take a look at the `GetNpmArtifactInfo` target. It's wired in to the `GetArtifactInfo` phase that KoreBuild defines. We'll want a `GetJavaArtifactInfo` target that works the same I think. I'm not sure if KoreBuild will need any updates to handle that, since we'll want to use a new ArtifactType.

clients/java/signalr/build.gradle
```diff
@@ -16,3 +17,32 @@ dependencies {
...
+    classifier "javadoc"
+    from javadoc.destinationDir
+}
+
+task generatePOM {
+    pom {
+        project {
+            groupId 'com.microsoft.aspnetcore'
+            artifactId 'signalr'
+            version '0.1.0'
```
anurse: We should use `0.1.0-preview1`.

clients/java/signalr/build.gradle
mikaelm12: > Call it signalr-client-0.1.0-preview1.pom (like the JAR files

I think this might actually need to be named `pom.xml`

anurse: It's required to be `pom.xml` when you're using Maven to build, but we're using Gradle. When publishing to Maven Central it should be `signalr-client-0.1.0-preview1.pom`. The [Maven Central Requirements doc](https://central.sonatype.org/pages/requirements.html) seems to indicate that in the examples.

mikaelm12: Ah okay cool. I was looking at some different sources. That link is very clear 👍

anurse: > Yes, please define an <ArtifactInfo> item with ArtifactType="ICanHasJavaPackageStuff"

I suggest `MavenPomFile` for the `.pom` and `JavaJar` for the `.jar` files.

mikaelm12:
Okay, so I've got gradle producing the `pom` and `jars` that we need with the correct names, then MSBuild copying the files into the right directory and declaring the new artifacts.

mikaelm12:
So I've added conditions to only run the java targets if the `JAVA_HOME` environment is set because this should indicate that java is installed on the machine

mikaelm12:
So far it from VSTS at least it seems like this strategy works on the Windows builds but not Mac or Linux. Ideas @anurse?

anurse:
Ah, that's a different issue. It's because you're running `gradlew` instead of `./gradlew`

mikaelm12:
Rip. I'll get on it

build/repo.targets
```diff
@@ -57,12 +62,32 @@
     </ItemGroup>
 
     <ItemGroup>
-      <FilesToExcludeFromSigning Include="%(NPMPackage.ArtifactPath)" />
+      <FilesToExcludeFromSigning Include="%(NPMPackage.ArtifactPath);" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="GetJavaArtifactInfo" Condition="'$(JAVA_HOME)' != ''">
```
anurse: Looks like we should also check for `Exists('$(JAVA_HOME)/lib/tools.jar')` since `JAVA_HOME` can be set when a JRE with no build tools is installed.

anurse:
The Jenkins OSX build is failing with Redis docker issues that we've been hitting for the past couple months. Can we just drop the Jenkins builds now that we have VSTS? We could keep the Spanish one since we don't have an image for that on VSTS. Thoughts @davidfowl @natemcmaster @ryanbrandenburg ?

natemcmaster:
Feel free to drop Jenkins/AppVeyor/Travis whenever you feel confident enough in the coverage you get from VSTS.

anurse:
Ok, I vote for dropping all but the Spanish build. Will sync up with @davidfowl and @pranavkm offline to avoid spamming this thread.

mikaelm12:
Green Builds 💚 💚 💚

build/repo.targets
```diff
@@ -83,4 +113,24 @@
     <Move SourceFiles="%(NPMPackage.OutputTar)" DestinationFiles="%(NPMPackage.ArtifactPath)" />
     <Move SourceFiles="%(NPMPackage.PackageJson).bak" DestinationFiles="%(NPMPackage.PackageJson)" />
   </Target>
+
+  <ItemGroup>
+    <JavaBuildFiles Include="signalr-client-$(JavaClientVersion)-preview1.jar;signalr-client-$(JavaClientVersion)-preview1-javadoc.jar;signalr-client-$(JavaClientVersion)-preview1-sources.jar;signalr-client-$(JavaClientVersion)-preview1.pom"/>
```
natemcmaster: Can you make the `preview1` part of the file name programmatic too? This is the kind of stuff we typically forget to update when we do preview2.

PR #2692, opened 2018-07-27
TypeScript improvements and tslib dep

### LINKED ISSUE #2410: Module not found: Can't resolve 'tslib' in '...@aspnet\signalr\dist\esm'
Hello,
i get the following error using "@aspnet/signalr": "1.0.0", and creat-react-app 

```
Failed to compile.

./node_modules/@aspnet/signalr/dist/esm/Errors.js
Module not found: Can't resolve 'tslib' in 'D:\VSTS\LSNG\node_modules\@aspnet\signalr\dist\esm'
```

anurse:
Isolates the modules used for dev tools into a sub-folder so they don't pollute the node resolution scope. This was causing the tests to pass despite the fact that the signalr and signalr-protocol-msgpack packages required the 'tslib' module but it wasn't specified as a dependency.

A few changes were needed to allow those tools to properly resolve things out of `clients/ts/common/node_modules` even though they are run in the `clients/ts` directory.

Also TSLint has deprecated their "unused variables" lint in favor of the built-in support in the compiler. So I updated to use that. Suppressions are now done via `// @ts-ignore: [Justification]` comments. Unlike TSLint's suppressions, `@ts-ignore` just does a blanket suppression of all "strictness" checks for the next line, so the commentary after the `:` is just a justification, rather than being a specific rule to disable.

Fixes #2410


clients/ts/signalr-protocol-msgpack/package.json
```diff
@@ -43,7 +43,8 @@
     "@aspnet/signalr": "^1.0.0-preview3"
   },
   "dependencies": {
-    "msgpack5": "^4.0.2"
+    "msgpack5": "^4.0.2",
+    "tslib": "^1.9.3"
```
anurse: FYI @Eilon @davidfowl @muratg  : new dependency on both `signalr` and `signalr-protocol-msgpack`. This is a library shipped by the TypeScript team that contains all the helpers they would normally inject into a compiled JS file. This is a recommended way to use the helpers as it allows multiple libraries to share the helpers in a single app rather than having the same helpers injected into every single file.

Unlike other dependencies we've added so far, this is a **runtime dependency** and not just for build. We can chat on Monday if you want to discuss it.

As a precedent: [`@angular/core`](https://www.npmjs.com/package/@angular/core) has this same dependency

clients/ts/signalr/package.json
```diff
@@ -37,5 +37,8 @@
   ],
   "devDependencies": {
     "es6-promise": "^4.2.2"
+  },
+  "dependencies": {
+    "tslib": "^1.9.3"
```
anurse: See above re: `tslib` dependency.

clients/ts/signalr-protocol-msgpack/package.json
davidfowl: Lets discuss monday

anurse:
We decided to just go to embedding the helpers.

anurse:
Ping?

clients/ts/FunctionalTests/ts/ConnectionTests.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+// This code uses a lot of `.then` instead of `await` and TSLint doesn't like it.
+// tslint:disable:no-floating-promises
```
BrennanConroy: Whats wrong with using await?

And wasn't this option enabled partially because we wanted to find possibly flaky tests because of floating promises?

anurse: These tests need to be fully converted to use promises properly. They currently use `done` instead. I didn't want to go through and completely rework them (including adding the necessary helpers so jasmine works) in this PR.

clients/ts/webpack.config.base.js
```diff
@@ -17,6 +17,10 @@ module.exports = function (modulePath, browserBaseName, options) {
             process: false,
             Buffer: false,
         },
+        resolveLoader: {
+            // Special resolution rules for loaders (which are in the 'common' directory)
+            modules: [ path.resolve(__dirname, "common", "node_modules") ],
```
BrennanConroy: This is the main change for the PR?

anurse: Yeah, the restructuring and then this change was needed to allow the WebPack loaders to resolve out of the other directory.

PR #2697, opened 2018-07-30
Fix race when connection is canceled and new poll comes in

BrennanConroy:
#2700
There was a race where we could set the connection state to `Inactive` even though there was an active poll which could cause the connection to be closed by the server.

The fix was to add a "task queue" so all new poll requests would wait until the previous poll finished all state updates before starting the new poll state updates.

Details:
Browser could abort the http request (normal timeout behavior) and then send a new poll, on the server we might start processing the poll and see that there is an active poll which we could proceed to cancel and wait for the transport task to finish. The transport task would get an operation canceled exception and see that the http request was aborted so would just return normally, Then it would proceed to wait for the state lock which the new poll is currently holding, the new poll would see the transport task is finished and set the status as active and start the new transport task and then release the lock. Now the previous poll can aquire the lock, see that the connection is active (which it assumed was from when it polled) and set it to inactive. Then our heartbeat would come along 5 seconds later and clean the connection up.

TLDR:
Poll is canceled by client, client sends new poll, race between previous poll and new poll where status could be set as inactive and heartbeat could clean up connection.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
@@ -217,6 +217,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
 
                         // Always wait for the previous request to drain
                         await connection.PreviousPollTask;
+                        connection.PreviousPollTask = currentRequestTcs.Task;
                     }
 
                     // Mark the connection as active
@@ -256,8 +257,6 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
 
                         // Start the transport
                         connection.TransportTask = longPolling.ProcessRequestAsync(context, tokenSource.Token);
-
-                        connection.PreviousPollTask = currentRequestTcs.Task;
 
                         // Start the timeout after we return from creating the transport task
                         timeoutSource.CancelAfter(options.LongPolling.PollTimeout);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
```diff
@@ -256,6 +257,8 @@ public HttpConnectionDispatcher(HttpConnectionManager manager, ILoggerFactory lo
                         // Start the transport
                         connection.TransportTask = longPolling.ProcessRequestAsync(context, tokenSource.Token);
 
+                        connection.PreviousPollTask = currentRequestTcs.Task;
```
anurse: We should move this outside the `if`, just in case a client re-polls during the first request. It would be rare because the first poll is "instant" but it could happen.

davidfowl: Good catch.

src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/LongPollingTransport.cs
```diff
@@ -90,6 +90,8 @@ public LongPollingTransport(CancellationToken timeoutToken, PipeReader applicati
             catch (Exception ex)
             {
                 Log.LongPollingTerminated(_logger, ex);
+                context.Response.ContentType = "text/plain";
```
davidfowl: This part of the change conflicts with https://github.com/aspnet/SignalR/commit/3d6e1e69fd03d0aed50f713ece482e7df65434a6. Revert it?

BrennanConroy: Beat me to it, was just updating this

PR #2703, opened 2018-07-31
Using the version suffix variable

mikaelm12:
Using the VersionSuffix variable instead of hard coding "preview1"

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -115,21 +115,21 @@
   </Target>
 
   <ItemGroup>
-    <JavaBuildFiles Include="signalr-client-$(JavaClientVersion)-$(VersionSuffix).jar;signalr-client-$(JavaClientVersion)-$(VersionSuffix)-javadoc.jar;signalr-client-$(JavaClientVersion)-$(VersionSuffix)-sources.jar;signalr-client-$(JavaClientVersion)-$(VersionSuffix).pom"/>
+    <JavaBuildFiles Include="signalr-client-$(JavaClientVersion).jar;signalr-client-$(JavaClientVersion)-javadoc.jar;signalr-client-$(JavaClientVersion)-sources.jar;signalr-client-$(JavaClientVersion).pom"/>
   </ItemGroup>
   
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


natemcmaster:
I don't know the answer to this, but is there a way to flow this into gradle too?

https://github.com/aspnet/SignalR/blob/e03166252578dca884f9b9d5c1d8a632de97754a/clients/java/signalr/build.gradle#L7

build/repo.targets
```diff
@@ -115,21 +115,21 @@
   </Target>
 
   <ItemGroup>
-    <JavaBuildFiles Include="signalr-client-$(JavaClientVersion)-preview1.jar;signalr-client-$(JavaClientVersion)-preview1-javadoc.jar;signalr-client-$(JavaClientVersion)-preview1-sources.jar;signalr-client-$(JavaClientVersion)-preview1.pom"/>
+    <JavaBuildFiles Include="signalr-client-$(JavaClientVersion)-$(VersionSuffix).jar;signalr-client-$(JavaClientVersion)-$(VersionSuffix)-javadoc.jar;signalr-client-$(JavaClientVersion)-$(VersionSuffix)-sources.jar;signalr-client-$(JavaClientVersion)-$(VersionSuffix).pom"/>
```
natemcmaster: FYI - this is going to result in `signalr-client-0.1.0-preview1-t000.jar` because VersionSuffix is appended with `$(BuildNumber)` in version.props.

An alternative to this is to make `$(JavaClientVersion)` contain the full version string, so all you need here is `signalr-client-$(JavaClientVersion).jar`


https://github.com/aspnet/SignalR/blob/e03166252578dca884f9b9d5c1d8a632de97754a/version.props#L11

mikaelm12:
> I don't know the answer to this, but is there a way to flow this into gradle too?

Hmm, I don't think so...

anurse:
Looks like we can set [project properties](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties) when we invoke gradle using `-P`. So if we invoked gradle with something like: `./gradlew -Pversionsuffix=$(VersionSuffix) blah blah blah` that should result in a `versionsuffix` variable in Gradle we can access.

PR #2706, opened 2018-07-31
remove Jenkins configs already covered by VSTS

anurse:
Removing macOS Jenkins config for PR builds. It seems to have an issue with Docker causing our Redis end-to-end tests to fail.

1. We have coverage of those end-to-end scenarios from other images (Jenkins Windows, for example).
2. We have macOS coverage from VSTS.

It's my understanding that it's **only** the version of these files in `master` that matters (cc @pranavkm) so there's no need to port this. Also you won't see the build disabled in this PR because it won't take effect until it's merged.


pranavkm:
The build's failing because it uses the checked in copy of the pipeline.groovy file, but the results of osx.groovy from the current PR.

anurse:
macOS failure is because of the removed groovy script, and thus expected.

anurse:
Travis seems to be weird network issues. I've asked @Eilon to also disable that since we have sufficient coverage there as well.

Eilon:
Travis CI disabled for this repo.

anurse:
I'm going to disable Ubuntu as well because it doesn't actually have a build script :)

PR #2708, opened 2018-07-31
Move .pom file in gradle build task

mikaelm12:
We're now generating .pom files. Adding it to the .gitignore file at the Java client level

### DIFF
--- a/clients/java/signalr/.gitignore
+++ b/clients/java/signalr/.gitignore
@@ -7,5 +7,6 @@ bin/
 out/
 *.class
 *.jar
+*.pom
 # We need the gradle-wrapper.jar file so that others can use the gradle wrapper
 !gradle/wrapper/gradle-wrapper.jar
--- a/clients/java/signalr/.gitignore
+++ b/clients/java/signalr/.gitignore
@@ -7,6 +7,5 @@ bin/
 out/
 *.class
 *.jar
-*.pom
 # We need the gradle-wrapper.jar file so that others can use the gradle wrapper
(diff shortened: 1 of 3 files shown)


clients/java/signalr/.gitignore
```diff
@@ -7,5 +7,6 @@ bin/
 out/
 *.class
 *.jar
+*.pom
```
anurse: Maybe we can just ignore the whole `build` folder?

mikaelm12: So we're already ignoring the entire build folder. The problem was that if you just run the gradle build task, the `signalr-client-0.1.0-preview1.pom` file would get created in the root of the java client. It was only getting moved by MSBuild. I updated the PR to have gradle move the .pom file to the build directory immediately after creating it, and then removed the move task that was in the MSBuild code.

BrennanConroy: Can this be removed then?

clients/java/signalr/build.gradle
```diff
@@ -45,4 +45,7 @@ task generatePOM {
             }
         }
     }.writeTo("signalr-client-0.1.0-preview1.pom")
-}
\ No newline at end of file
+
+    ant.move file: "signalr-client-0.1.0-preview1.pom",
```
BrennanConroy: Are these versions going to not be hardcoded in the future?

clients/java/signalr/.gitignore
anurse: Ah, good. Sounds fine to me then. Thanks for clarifying!

Issue #2638, opened 2018-07-16
Add logging for Java Client


mikaelm12:
The Java client basically has no logging at the moment. 
Will need to add
- Logger abstraction
- Make log level configurable
- Have console logger implementation

PR #2720, opened 2018-08-02
Add logging infrastructure to the Java Client

mikaelm12:
Starting to add logging on the Java client
Issue: https://github.com/aspnet/SignalR/issues/2638

### DIFF
--- a/clients/java/signalr/src/main/java/ConsoleLogger.java
+++ b/clients/java/signalr/src/main/java/ConsoleLogger.java
@@ -1,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public class ConsoleLogger implements Logger {
+    private LogLevel logLevel;
+    public ConsoleLogger(LogLevel logLevel) {
+            this.logLevel = logLevel;
+    }
+
+    @Override
+    public void log(LogLevel logLevel, String message) {
+        if(logLevel.value >= this.logLevel.value){
+            switch (logLevel) {
+                case Debug:
+                case Information:
+                case Warning:
+                    System.out.println(message);
+                    break;
+                case Error:
+                case Critical:
+                    System.err.println(message);
+                    break;
+            }
+        }
+    }
+}
--- a/clients/java/signalr/src/main/java/ConsoleLogger.java
+++ b/clients/java/signalr/src/main/java/ConsoleLogger.java
@@ -13,12 +13,29 @@ public class ConsoleLogger implements Logger {
             switch (logLevel) {
                 case Debug:
                 case Information:
-                case Warning:
                     System.out.println(message);
                     break;
+                case Warning:
                 case Error:
                 case Critical:
                     System.err.println(message);
+                    break;
+            }
+        }
+    }
+
+    @Override
+    public void log(LogLevel logLevel, String formattedMessage, Object... args) {
+        if (logLevel.value >= this.logLevel.value) {
+            switch (logLevel) {
+                case Debug:
+                case Information:
+                    System.out.printf(formattedMessage, args);
+                    break;
+                case Warning:
+                case Error:
+                case Critical:
+                    System.err.printf(formattedMessage, args);
                     break;
             }
         }
--- a/clients/java/signalr/src/main/java/ConsoleLogger.java
+++ b/clients/java/signalr/src/main/java/ConsoleLogger.java
@@ -27,6 +27,7 @@ public class ConsoleLogger implements Logger {
     @Override
     public void log(LogLevel logLevel, String formattedMessage, Object... args) {
         if (logLevel.value >= this.logLevel.value) {
+            formattedMessage = formattedMessage + "%n";
             switch (logLevel) {
                 case Debug:
                 case Information:
--- a/clients/java/signalr/src/main/java/HubConnection.java
+++ b/clients/java/signalr/src/main/java/HubConnection.java
@@ -18,10 +18,12 @@ public class HubConnection {
     private Boolean handshakeReceived = false;
     private static final String RECORD_SEPARATOR = "\u001e";
     private HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
+    private Logger logger;
 
-    public HubConnection(String url, Transport transport) {
+    public HubConnection(String url, Transport transport, Logger logger){
         this.url = url;
         this.protocol = new JsonHubProtocol();
+        this.logger = logger;
         this.callback = (payload) -> {
 
             if (!handshakeReceived) {
@@ -45,6 +47,7 @@ public class HubConnection {
             for (HubMessage message : messages) {
                 switch (message.getMessageType()) {
                     case INVOCATION:
+                        logger.log("Recevied message of type INVOCATION");
                         InvocationMessage invocationMessage = (InvocationMessage)message;
                         if (message != null && handlers.containsKey(invocationMessage.target)) {
                             ArrayList<Object> args = gson.fromJson((JsonArray)invocationMessage.arguments[0], (new ArrayList<>()).getClass());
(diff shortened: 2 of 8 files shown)


clients/java/signalr/src/main/java/ConsoleLogger.java
```diff
@@ -0,0 +1,26 @@
...
+    private LogLevel logLevel;
+    public ConsoleLogger(LogLevel logLevel) {
+            this.logLevel = logLevel;
+    }
+
+    @Override
+    public void log(LogLevel logLevel, String message) {
+        if(logLevel.value >= this.logLevel.value){
+            switch (logLevel) {
+                case Debug:
```
mikaelm12: Just  copied the log levels from the TS client
https://github.com/aspnet/SignalR/blob/32d380a05801bda255bbf9d758298b8285e52ea5/clients/ts/signalr/src/ILogger.ts#L9-L24

mikaelm12: We can probably just consolidate these to Info and Error

clients/java/signalr/src/main/java/LogLevel.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public enum LogLevel {
+    Debug(1),
```
mikaelm12: refer to above comment

clients/java/signalr/src/main/java/ConsoleLogger.java
anurse: We wouldn't want to consolidate them, then we'd lose the ability to have more granular messages.

mikaelm12:
Updated.

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -81,36 +87,50 @@ public HubConnection(String url, Transport transport) {
...
         InvocationMessage invocationMessage = new InvocationMessage(method, args);
         String message = protocol.writeMessage(invocationMessage);
+        logger.log(LogLevel.Debug, "Sending message");
         transport.send(message);
     }
 
     public Subscription on(String target, Action callback) {
         ActionBase action = args -> callback.invoke();
         handlers.put(target, action);
+        logger.log(LogLevel.Trace, "Handler added to target: " + target);
```
anurse: Can you go through [HubConnection.Log.cs](https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs) and make sure that you have similar messages here when they're relevant and that the text matches (unless it would be super weird).

We should make the logging consistent between those clients. We also should've done the same with TypeScript but better late than never ;P

clients/java/signalr/src/main/java/Logger.java
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+public interface Logger {
+    void log(LogLevel logLevel, String message);
```
anurse: Can we add an overload to this that takes format specifiers that we flow through `String.format`?

```java
void log(LogLevel logLevel, String format, Object... args);
```

That means we can make interpolated log strings a little cleaner:

```java
logger.log(LogLevel.Trace, "Some message (some value = %d)", someValue);
```

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -125,6 +145,7 @@ public Subscription on(String target, Action callback) {
             callback.invoke(param1.cast(params[0]), param2.cast(params[1]));
         };
         handlers.put(target, action);
+        logger.log(LogLevel.Trace, "Handler added to target: " + target);
```
anurse: We should use `String.format` (via the overload I suggested below for Logger) to concatenate strings.

clients/java/signalr/src/main/java/WebSocketTransport.java
```diff
@@ -37,8 +39,10 @@ else if(url.startsWith(HTTP)){
 
     @Override
     public void start() throws InterruptedException {
+        logger.log(LogLevel.Debug, "Starting Websocket connection");
```
anurse: Again, take a quick look at the WebSocket transport in C# and make sure the levels and messages align where they should

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -18,17 +18,20 @@
...
         this.protocol = new JsonHubProtocol();
+        this.logger = logger;
         this.callback = (payload) -> {
 
             if (!handshakeReceived) {
                 int handshakeLength = payload.indexOf(RECORD_SEPARATOR) + 1;
                 String handshakeResponseString = payload.substring(0, handshakeLength - 1);
                 HandshakeResponseMessage handshakeResponse = HandshakeProtocol.parseHandshakeResponse(handshakeResponseString);
                 if (handshakeResponse.error != null) {
+                    logger.log(LogLevel.Error, "There was an error in the handshake responsed");
```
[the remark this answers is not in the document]
mikaelm12: Makes sense

clients/java/signalr/src/main/java/HubConnection.java
mikaelm12: Yeah, these should be consistent

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -45,6 +48,7 @@ public HubConnection(String url, Transport transport) {
             for (HubMessage message : messages) {
                 switch (message.getMessageType()) {
                     case INVOCATION:
+                        logger.log(LogLevel.Debug,"Recevied message of type INVOCATION");
                         InvocationMessage invocationMessage = (InvocationMessage)message;
                         if (message != null && handlers.containsKey(invocationMessage.target)) {
```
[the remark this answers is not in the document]
mikaelm12: Good catch

clients/java/signalr/src/main/java/ConsoleLogger.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    @Override
+    public void log(LogLevel logLevel, String formattedMessage, Object... args) {
+        if (logLevel.value >= this.logLevel.value) {
+            switch (logLevel) {
+                case Debug:
+                case Information:
+                    System.out.printf(formattedMessage, args);
```
anurse: Does this print a newline at the end? I don't think it does from the [docs](https://docs.oracle.com/javase/7/docs/api/java/io/PrintStream.html). I think `System.out.println(String.format(formattedMessage, args))` is what we want here (and the same below for `System.err`)

clients/java/signalr/src/main/java/HubConnection.java
```diff
@@ -43,36 +47,41 @@ public HubConnection(String url, Transport transport) {
             HubMessage[] messages = protocol.parseMessages(payload);
 
             for (HubMessage message : messages) {
+                logger.log(LogLevel.Debug,"Received message of type %s%n", message.getMessageType());
```
anurse: Ah, I see, you're putting `%n` in the log messages themselves. I don't think we should require that the caller provide the newlines. The newline isn't part of the "log message" it's part of what the Console Logger has to do in order to render it to the console. If we were logging to a database or some other storage, you probably wouldn't want a newline here

clients/java/signalr/src/main/java/ConsoleLogger.java
mikaelm12: No new line in `printf` which is why I added them in the messages. Makes more sense to add it to the console logger 👍

clients/java/signalr/src/main/java/ConsoleLogger.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class ConsoleLogger implements Logger {
+    private LogLevel logLevel;
+    public ConsoleLogger(LogLevel logLevel) {
+            this.logLevel = logLevel;
+    }
+
+    @Override
+    public void log(LogLevel logLevel, String message) {
+        if(logLevel.value >= this.logLevel.value){
```
moozzyk: Any custom logger will have to have this logic which is a smell (coupling). An idea I came up with in the Swift client was to create an internal filtering logger that wraps the actual logger (https://github.com/moozzyk/SignalR-Client-Swift/commit/afc00220754babf78054a8f66abc82aa1ef08e1d). This way none of the implementations of the logger interface needs to worry about min log level.

PR #2722, opened 2018-08-02
Log when closing connection due to timeout

BrennanConroy:
Was debugging and took a long time to figure out why the connection was closing. This should be logged.

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
@@ -502,6 +502,7 @@ namespace Microsoft.AspNetCore.SignalR
             {
                 if (!_receivedMessageThisInterval)
                 {
+                    Log.ClientTimeout(_logger);
                     Abort();
                 }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs
```diff
@@ -560,6 +561,9 @@ private static class Log
             private static readonly Action<ILogger, Exception> _abortFailed =
                 LoggerMessage.Define(LogLevel.Trace, new EventId(8, "AbortFailed"), "Abort callback failed.");
 
+            private static readonly Action<ILogger, Exception> _clientTimeout =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(9, "ClientTimeout"), "Client has not sent a message in too long. Closing connection.");
```
mikaelm12: Maybe "Client has been inactive for too long"
or "Client hasn't sent a message in the Timeout period". I kind of want to reference the timeout but it's not a big deal

anurse: Don't use "too long" ;)

Rephrase this (and if possible, put the timeout value in) as something like "Client timeout interval ({value}) elapsed without receiving a message. Closing connection."

Issue #2717, opened 2018-08-02
.NET Client expects server pings on LongPolling connections which will close idle connections


BrennanConroy:
LongPolling has an inherent keep alive system and so we correctly turn off server-side pinging. However the .NET Client is still expecting the pings, so after 30 seconds it will close the connection.

The TS client already checks for inherent keep alive, so it doesn't have this issue.

PR #2727, opened 2018-08-03
Honor InherentKeepAliveFeature for server timeout

BrennanConroy:
#2717 

Not a fan of using a delay for a timeout test, but couldn't think of a better way...

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -12,6 +12,7 @@ using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Connections.Features;
 using Microsoft.AspNetCore.Internal;
 using Microsoft.AspNetCore.SignalR.Client.Internal;
 using Microsoft.AspNetCore.SignalR.Protocol;
@@ -893,12 +894,14 @@ namespace Microsoft.AspNetCore.SignalR.Client
             ResetTimeout();
             ResetSendPing();
 
+            var inherentKeepAlive = _connectionState.Connection.Features.Get<IConnectionInherentKeepAliveFeature>()?.HasInherentKeepAlive ?? false;
+
             using (timer)
             {
                 // await returns True until `timer.Stop()` is called in the `finally` block of `ReceiveLoop`
                 while (await timer)
                 {
-                    if (DateTime.UtcNow.Ticks > Volatile.Read(ref _nextActivationServerTimeout))
+                    if (!inherentKeepAlive && DateTime.UtcNow.Ticks > Volatile.Read(ref _nextActivationServerTimeout))
                     {
                         OnServerTimeout();
                     }
--- a/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
@@ -48,6 +48,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private long _nextActivationServerTimeout;
         private long _nextActivationSendPing;
         private bool _disposed;
+        private bool _hasInherentKeepAlive;
 
         private readonly ConnectionLogScope _logScope;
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -119,6 +120,33 @@ public async Task ConnectionTerminatedIfServerTimeoutIntervalElapsesWithNoMessag
...
+                await hubConnection.StartAsync().OrTimeout();
+
+                var closeTcs = new TaskCompletionSource<Exception>();
+                hubConnection.Closed += ex =>
+                {
+                    closeTcs.TrySetResult(ex);
+                    return Task.CompletedTask;
+                };
+
+                await Task.Delay(1000);
```
DylanDmitri: For flaky test avoidance reasons, maybe "fake" this passage of time by manually decreasing `_nextActivationServerTimeout` by 1000 ms.

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
```diff
@@ -893,12 +894,14 @@ private async Task TimerLoop(TimerAwaitable timer)
...
 
+            var inherentKeepAlive = _connectionState.Connection.Features.Get<IConnectionInherentKeepAliveFeature>()?.HasInherentKeepAlive ?? false;
+
             using (timer)
             {
                 // await returns True until `timer.Stop()` is called in the `finally` block of `ReceiveLoop`
                 while (await timer)
                 {
-                    if (DateTime.UtcNow.Ticks > Volatile.Read(ref _nextActivationServerTimeout))
+                    if (!inherentKeepAlive && DateTime.UtcNow.Ticks > Volatile.Read(ref _nextActivationServerTimeout))
```
DylanDmitri: Should the server not ping connections with inherent keep alive? or is this more trouble than it's worth?

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
```diff
@@ -119,6 +120,33 @@ public async Task ConnectionTerminatedIfServerTimeoutIntervalElapsesWithNoMessag
             Assert.Equal($"Server timeout ({hubConnection.ServerTimeout.TotalMilliseconds:0.00}ms) elapsed without receiving a message from the server.", exception.Message);
         }
 
+        [Fact]
+        public async Task InherentKeepAliveFeatureCanPreventServerTimeout()
```
anurse: nit: `ServerTimeoutIsDisabledWhenUsingTransportWithInherentKeepAlive`

src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs
BrennanConroy: We already don't send the pings
https://github.com/aspnet/SignalR/blob/810c7cf9ad70d2a12708826100127a085b556a7f/src/Microsoft.AspNetCore.SignalR.Core/HubConnectionContext.cs#L415-L419

test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.cs
BrennanConroy: It's still using a `Timer` internally, which has a ~15ms minimum delay already, so setting the `ServerTimeout` to 1ms should guarantee that if the timer runs then the server timeout code should run.

BrennanConroy: But that gives me an idea, I could wait for a ping to be sent out by the client which means that the timer ran and therefore that the server timeout ran

davidfowl: Instead of doing this, can we do what we do on the server:

https://github.com/aspnet/SignalR/blob/3d6e1e69fd03d0aed50f713ece482e7df65434a6/test/Microsoft.AspNetCore.Http.Connections.Tests/HttpConnectionDispatcherTests.cs#L451

Call the method that runs per tick to force behavior instead of waiting for the timer?

BrennanConroy: We're using your fancy `TimerAwaitable`, so there is no function we can call
https://github.com/aspnet/SignalR/blob/f0c4170e42e72462c05cf1ac14b820e38731da90/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L899-L901

davidfowl: Make a function.

davidfowl:
👏

PR #2731, opened 2018-08-03
Solve LongPolling races by moving Cts disposal to connection disposal

BrennanConroy:
When we changed the code to be cancel we started hitting https://github.com/dotnet/corefx/issues/31587

But we realized we could move the cts disposal to only dispose on connection closure because new polls already dispose the previous token.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionManager.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionManager.cs
@@ -194,6 +194,8 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
         {
             try
             {
+                connection.Cancellation?.Dispose();
+
                 await connection.DisposeAsync(closeGracefully);
             }
             catch (IOException ex)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionManager.cs
```diff
@@ -194,6 +194,8 @@ public async Task DisposeAndRemoveAsync(HttpConnectionContext connection, bool c
         {
             try
             {
+                connection.Cancellation?.Dispose();
```
davidfowl: Make it part of DisposeAsync

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionContext.cs
```diff
@@ -182,6 +182,8 @@ public async Task DisposeAsync(bool closeGracefully = false)
         {
             Task disposeTask;
 
+            Cancellation?.Dispose();
```
davidfowl: The race still exists if you don’t put it in the lock right? It’s just extremely unlikely.

BrennanConroy: It's very unlikely and would result in a "Failed disposing connection" log.

A new poll would need to be calling cancel while the connection is being disposed.

davidfowl: Yes I'm less concerned about that.

PR #2740, opened 2018-08-06
Skip Redis tests on macOS

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
```diff
@@ -50,6 +50,13 @@ private static Docker Create()
 
         private static string GetDockerLocation()
         {
+            // OSX + Docker + Redis don't play well together for some reason. We already have these tests covered on Linux and Windows
```
davidfowl: tsk... 😄 I'll see if I can figure it out.

davidfowl:
Why are the windows tests failing

BrennanConroy:
The spanish build I think is because of the LongPolling issue. And the other Windows test is a flaky thing that happens sometimes. Don't know how.

PR #2742, opened 2018-08-06
Avoid ODE in tests with Server logging

BrennanConroy:


test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerLogScope.cs
```diff
@@ -49,6 +50,11 @@ private void ServerFixtureOnServerLogged(LogRecord logRecord)
             // https://github.com/aspnet/Logging/issues/810
```
JamesNK: This issue has been fixed. Move the the `logger = _serverLoggers.GetOrAdd(...)` call out of the lock and remove the comment.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerLogScope.cs
```diff
@@ -49,6 +50,11 @@ private void ServerFixtureOnServerLogged(LogRecord logRecord)
...
             lock (_lock)
             {
+                if (_disposed)
```
JamesNK: The disposed check should probably be at the start of the method.

test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerLogScope.cs
BrennanConroy: Keeping it in the lock as that's what is preventing a race where we would otherwise get an ODE.

But removed the comment :D

Issue #2735, opened 2018-08-04
(Server shutdown) Writing is not allowed after writer was completed


benaadams:
When shutting down server always receive the following if a client is connected
```
System.InvalidOperationException: Writing is not allowed after writer was completed.
   at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException_NoWritingAllowed()
   at System.IO.Pipelines.Pipe.GetMemory(Int32 sizeHint)
   at System.IO.Pipelines.Pipe.DefaultPipeWriter.GetMemory(Int32 sizeHint)
   at Microsoft.AspNetCore.Internal.Utf8BufferTextWriter.EnsureBuffer()
   at Microsoft.AspNetCore.Internal.Utf8BufferTextWriter.Write(Char value)
   at Newtonsoft.Json.JsonTextWriter.WriteStartObject()
   at Microsoft.AspNetCore.SignalR.Protocol.JsonHubProtocol.WriteMessageCore(HubMessage message, IBufferWriter`1 stream)
   at Microsoft.AspNetCore.SignalR.Protocol.JsonHubProtocol.WriteMessage(HubMessage message, IBufferWriter`1 output)
   at Microsoft.AspNetCore.SignalR.HubConnectionContext.WriteCore(HubMessage message)
```

In `OnDisconnectedAsync` I do remove the connection form Groups; so that might cause the issue?

However, it doesn't occur while the server is running and clients disconnect.

* NuGet Packages: `"Microsoft.AspNetCore.App" Version="2.1.1"`
* Versions of Client-Side NuGet/NPM Packages:
* JavaScript client: `"@aspnet/signalr": "^1.0.2"`
* Server Kestrel 
* Server Windows 10
* Client Windows 10
* Chrome Version 68.0.3440.84 (Official Build) (64-bit)

```
dbug: Microsoft.AspNetCore.Server.Kestrel[9]
      Connection id "0HLFQ9QH9B610" completed keep alive response.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 3234.9771ms 200 application/json; charset=utf-8
Application is shutting down...
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[5]
      Hosting shutdown
dbug: Microsoft.AspNetCore.Http.Connections.Internal.Transports.WebSocketsTransport[7]
      Waiting for the client to close the socket.
dbug: Microsoft.AspNetCore.SignalR.HubConnectionContext[6]
[...]

PR #2748, opened 2018-08-07
Make un-graceful shutdown a bit more graceful

davidfowl:
- Change how graceful shutdown is done to ensure there are no errors on shutdown.
- Avoid closing pipes from under other components that own those pipes.
- Close the application output so that the application's read loop ends gracefully. For websockets, we cancel the token, for other transports, we close the output directly.

Fixes https://github.com/aspnet/SignalR/issues/2735


davidfowl:

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
```diff
@@ -166,6 +166,9 @@ private async Task ExecuteAsync(HttpContext context, ConnectionDelegate connecti
 
                 Log.EstablishedConnection(_logger);
 
+                // Allow the reads to be cancelled
```
BrennanConroy: nit: canceled

PR #2751, opened 2018-08-07
fix jdk detection by detecting javac instead of tools.jar

anurse:
Turns out Java 9 dropped `tools.jar`, which is what we were using to detect the JDK. Gradle works fine on Java 9+ but because tools.jar was removed, our MSBuild thinks the JDK isn't installed.

Changed to use the Java Compiler binary itself to detect this.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -43,7 +43,7 @@
     <Exec Command="npm run test:inner -- --no-color --configuration $(Configuration)" WorkingDirectory="$(RepositoryRoot)clients/ts/FunctionalTests" IgnoreStandardErrorWarningFormat="true" />
   </Target>
 
-  <Target Name="RunJavaTests" Condition="'$(JAVA_HOME)' != '' And Exists('$(JAVA_HOME)/lib/tools.jar')">
+  <Target Name="RunJavaTests" Condition="'$(HaveJava)' == 'true'">
     <Message Text="Running Java client tests" Importance="high" />
     <Exec Command="./gradlew test" WorkingDirectory="$(RepositoryRoot)clients/java/signalr" IgnoreStandardErrorWarningFormat="true" />
   </Target> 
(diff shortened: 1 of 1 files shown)


build/repo.targets
```diff
@@ -10,6 +10,14 @@
     </NPMPackage>
   </ItemGroup>
 
+  <PropertyGroup Condition="'$(JAVA_HOME)' != ''">
+    <!-- Determine the path we need to search to find javac.exe -->
+    <JavacPath Condition="'$(OS)' == 'Windows_NT'">$(JAVA_HOME)\bin\javac.exe</JavacPath>
+    <JavacPath Condition="'$(OS)' != 'Windows_NT'">$(JAVA_HOME)/bin/javac</JavacPath>
+
+    <HaveJdk Condition="Exists('$(JavacPath)')">true</HaveJdk>
```
mikaelm12: You should also update the condition for the `RunJavaTests` target

anurse: whoops!

build/repo.targets
```diff
@@ -35,7 +43,7 @@
     <Exec Command="npm run test:inner -- --no-color --configuration $(Configuration)" WorkingDirectory="$(RepositoryRoot)clients/ts/FunctionalTests" IgnoreStandardErrorWarningFormat="true" />
   </Target>
 
-  <Target Name="RunJavaTests" Condition="'$(JAVA_HOME)' != '' And Exists('$(JAVA_HOME)/lib/tools.jar')">
+  <Target Name="RunJavaTests" Condition="'$(HaveJava)' == 'true'">
```
campersau: `HaveJdk`?

mikaelm12:
nit: Maybe change `Have` to `Has`

build/repo.targets
anurse: Nice catch :)

Issue #2684, opened 2018-07-26
Flaky test: ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=515720&buildTypeId=XPlat_Windows_Win2008r2_Universe) occasionally with the following error:
```
System.TimeoutException : Operation in ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport timed out at /_/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs:360
   at System.Threading.Tasks.TaskExtensions.OrTimeout(Task task, TimeSpan timeout, String memberName, String filePath, Nullable`1 lineNumber)
   at Microsoft.AspNetCore.SignalR.Tests.EndToEndTests.ConnectionCanSendAndReceiveDifferentMessageSizesWebSocketsTransport(String message) in /_/test/Microsoft.AspNetCore.SignalR.Tests/EndToEndTests.cs:line 361
--- End of stack trace from previous location where exception was thrown ---
***** Text was not loaded fully because its' size exceeds 2 MB, see full log for the whole text *****
```

CC @Eilon,@muratg,@mkArtakMSFT

This issue was made automatically. If there is a problem contact @ryanbrandenburg.

ryanbrandenburg:
Please use this workflow to address this flaky test issue, including checking applicable checkboxes and filling in the applicable "TODO" entries:

* Is this test failure caused by product code flakiness? (Either this product, or another product this test depends on.)
  * [ ] File a bug against the product (TODO: Link to other bug)
  * Is it possible to change the test to avoid the flakiness?
    * Yes? Go to the "Change the test!" section.
    * No?
      * [ ] Disable the test (TODO: Link to PR/commit)
      * [ ] Wait for other bug to be resolved
      * [ ] Wait for us to get build that has the fix
      * [ ] Re-enable our test (TODO: Link to PR/commit)
      * [ ] Close this bug

* Is it that the test itself is flaky? This includes external transient problems (e.g. remote server problems, file system race condition, etc.)
  * Is there is a way to change our test to avoid this flakiness?
    * Yes? Change the test!
      * [ ] Change the test to avoid the flakiness, for example by using a different test strategy, or by adding retries w/ timeouts (TODO: Link to PR/commit)
      * [ ] Run the test 100 times locally as a sanity check.
      * [ ] Close this bug
    * No?
[...]

PR #2752, opened 2018-08-07
Log Started Transport for WebSockets

BrennanConroy:


davidfowl:
What's this for?

anurse:
It's about the only small bit of reasonable changes I can think of to get more data for https://github.com/aspnet/SignalR/issues/2684 short of diving really deep to figure out why `ConnectAsync` seems to be taking 10 mins.

davidfowl:
@BrennanConroy next time link the bug.

PR #2759, opened 2018-08-08
disable the gradle daemon on the CI

anurse:
Going to merge when I get green builds since this is currently blocking UpdateUniverse in 2.2

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -16,6 +16,9 @@
     <JavacPath Condition="'$(OS)' != 'Windows_NT'">$(JAVA_HOME)/bin/javac</JavacPath>
 
     <HasJdk Condition="Exists('$(JavacPath)')">true</HasJdk>
+
+    <!-- Disable gradle daemon on CI since the CI seems to try to wait for the daemon to shut down, which it doesn't do :) -->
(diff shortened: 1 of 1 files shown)


build/repo.targets
```diff
@@ -16,6 +16,9 @@
     <JavacPath Condition="'$(OS)' != 'Windows_NT'">$(JAVA_HOME)/bin/javac</JavacPath>
 
     <HasJdk Condition="Exists('$(JavacPath)')">true</HasJdk>
+
+    <!-- Disable gradle daemon on CI since the CI seems to try to wait for the daemon to shut down, which it doesn't do :) -->
```
anurse: We only disable when `CI` is `true` because we'd still like the Daemon to be in use on dev machines.

ryanbrandenburg: By "the CI" do you mean VSTS? If so it sounds like the same thing we had a thread about this with the VSTS folks, and we have an issue filed with them that I'll link you to.

anurse: No, TeamCity. VSTS works fine.

anurse: It's either a TeamCity thing, a Universe thing (since it only occurs in TeamCity when building Universe/UpdateUniverse) or some combination of both.

Issue #2761, opened 2018-08-08
Move classes in to 'com.microsoft.aspnet.signalr' package


anurse:
Our Java classes are in the default package. We should move them into `com.microsoft.aspnet.signalr`.

We need a `package com.microsoft.aspnet.signalr;` directive at the top of each file and probably need to put them in a folder structure (`com/microsoft/aspnet/signalr`).

mikaelm12:
Done here: c9104d4932b85ab7dfa784f3314421584a3882e1

Issue #2676, opened 2018-07-25
Consider adding "Spotless" to Java build


anurse:
We should use something like [Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-to-java-source-google-java-format) to enforce and apply code style to our Java code. I think we should do similar things in C#, but at least with C# we're all quite familiar with the code style and can enforce the major ones in PRs. Plus people get a little wary when you start suggesting automated C# style enforcement 😛 

I suggest we just pick an existing Java style and go with it. [Google Java Style](https://google.github.io/styleguide/javaguide.html) seems fine and is [supported by Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-to-java-source-google-java-format)

mikaelm12:
Cool, I'll take a look

mikaelm12:
Done here: 39e6c6ff2f91d41564615efdd2df04063d499250

PR #2763, opened 2018-08-08
Move Java client into com.microsoft.aspnet.signalr

mikaelm12:
This is also causing an issue when trying to use the client from a jar. Because it was in the default package I was running into an issue of having classes signed by different sources in the same package which isn't allowed. Moving the client into the `com.microsoft.aspnet.signalr ` package should fix that.
```
Exception in thread "main" java.lang.SecurityException: class "HubConnection"'s signer information does not match signer information of other classes in the same package
```
Issue: https://github.com/aspnet/SignalR/issues/2761


clients/java/signalr/src/test/java/HandshakeProtocolTest.java
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+import com.microsoft.aspnet.signalr.HandshakeProtocol;
```
mikaelm12: Need to decide on the ordering of import statements

anurse: https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

#2676 should be how we resolve this in general.

Issue #2744, opened 2018-08-06
Changes for Maven publishing


anurse:
We need to make sure our POM includes a few things. See below for what I modified the POM file to include. We need to make sure the file that comes out of our build has these values:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.microsoft.aspnet</groupId>
  <artifactId>signalr</artifactId>
  <version>0.1.0-alpha1</version>
  <inceptionYear>2018</inceptionYear>
  <description>ASP.NET Core SignalR Client for Java applications</description>
  <url>https://github.com/aspnet/SignalR</url>
  <name>${project.groupId}:${project.artifactId}</name>
  <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
  <scm>
    <connection>scm:git:git://github.com/aspnet/SignalR.git</connection>
    <developerConnection>scm:git:git://github.com/aspnet/SignalR.git</developerConnection>
    <url>http://github.com/aspnet/SignalR/tree/master</url>
  </scm>
  <developers>
    <developer>
      <id>microsoft</id>
      <name>Microsoft</name>
    </developer>
  </developers>
  <dependencies>
    <!-- snipped, these were fine -->
  </dependencies>
</project>
```

anurse:
Note that we reserved the groupId `com.microsoft.aspnet`, so we need to change the groupId to match that.

anurse:
We also need the `<developer>` section filled in. Stand-by for updated blurb above.

anurse:
Also the output file names should match the artifactId. Currently the artifactId is `signalr`, but the files are `signalr-client-0.1.0-preview1.jar`. We should rename the files to `signalr-0.1.0-preview1.jar,pom,etc.`

mikaelm12:
Alright, working on this now

mikaelm12:
Done here: 5843b5415b56367de53f78376d94cac61f6fc0a7

PR #2765, opened 2018-08-08
Update Java client build scripts for Maven publishing

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/2744

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -126,15 +126,9 @@
   </Target>
 
   <ItemGroup>
-    <JavaBuildFiles Include="signalr-$(JavaClientVersion).jar;signalr-$(JavaClientVersion)-javadoc.jar;signalr-$(JavaClientVersion)-sources.jar;signalr-$(JavaClientVersion).pom"/>
-  </ItemGroup>
-  
-  <ItemGroup>
     <Jars Include="signalr-$(JavaClientVersion).jar;signalr-$(JavaClientVersion)-javadoc.jar;signalr-$(JavaClientVersion)-sources.jar;" />
-  </ItemGroup>
-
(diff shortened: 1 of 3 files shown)


clients/java/signalr/build.gradle
```diff
@@ -36,16 +36,30 @@ task generatePOM {
             version '0.1.0-preview1'
```
anurse: There's a `groupId` field on line 34 that's incorrect. I think you don't need to specify `groupId`, `artifactId`, or `version` and it will pull them from the Gradle top-level properties. I did a quick test removing them and ran `gradlew generatePOM` and the generated POM still had the right values. That would let us remove duplication :).

mikaelm12:
Updated. Scanned the pom in the generated artifacts of the VSTS build looks good

build/repo.targets
```diff
@@ -126,15 +126,15 @@
   </Target>
 
   <ItemGroup>
-    <JavaBuildFiles Include="signalr-client-$(JavaClientVersion).jar;signalr-client-$(JavaClientVersion)-javadoc.jar;signalr-client-$(JavaClientVersion)-sources.jar;signalr-client-$(JavaClientVersion).pom"/>
+    <JavaBuildFiles Include="signalr-$(JavaClientVersion).jar;signalr-$(JavaClientVersion)-javadoc.jar;signalr-$(JavaClientVersion)-sources.jar;signalr-$(JavaClientVersion).pom"/>
```
BrennanConroy: Could you define the `Jars` and `PomFile` above here, then use those variables for `JavaBuildFiles` so you don't have to duplicate the names

Issue #2313, opened 2018-05-18
Unhelpful error message when binding to a method that doesn't exist


JamesNK:
If you are sending to a hub method that doesn't exist, and you are sending arguments, rather than getting a message that the name of the method is bad you get an error message about the number of arguments:

```
System.IO.InvalidDataException: Invocation provides 1 argument(s) but target expects 0.
   at Microsoft.AspNetCore.SignalR.Protocol.JsonHubProtocol.BindArguments(JsonTextReader reader, IReadOnlyList`1 paramTypes) in C:\Development\Source\SignalR\src\Microsoft.AspNetCore.SignalR.Protocols.Json\Protocol\JsonHubProtocol.cs:line 618
   at Microsoft.AspNetCore.SignalR.Protocol.JsonHubProtocol.ParseMessage(Utf8BufferTextReader textReader, IInvocationBinder binder) in C:\Development\Source\SignalR\src\Microsoft.AspNetCore.SignalR.Protocols.Json\Protocol\JsonHubProtocol.cs:line 223
```

This is because DefaultHubDispatcher returns an empty list for an unknown method name:

```
public override IReadOnlyList<Type> GetParameterTypes(string methodName)
{
    if (!_methods.TryGetValue(methodName, out var descriptor))
    {
        return Type.EmptyTypes;
    }
    return descriptor.ParameterTypes;
}
```

anurse:
Looks straightforward enough :).

mikaelm12:
I was taking an initial look at this issue this past Saturday and I think it's a little more involved than we initially thought. I'll take it after clearing whats on my plate now but I'm going to bump it up a level

anurse:
Let's get it on the queue for next week

anurse:
We also want to avoid the error message on the client when a method that does not exist is called.

We could reuse `InvocationBindingErrorMessage` (or create a similar synthetic message) to carry information like this out of the protocol so that the HubConnection/HubConnectionHandler can handle it the way it wants to.

Jelledb:
Exactly this, I just stumbled upon this page and kept looking for any errors in my amount of arguments, only to find out thanks to this page that my methods were inconsistently named.

PR #2768, opened 2018-08-08
Provide a better error message when invoking a non-existant hub method

BrennanConroy:
#2313

### DIFF
--- a/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
+++ b/src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
@@ -136,7 +136,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
         {
             if (!_methods.TryGetValue(methodName, out var descriptor))
             {
-                throw new HubException("Method doesn't exist.");
+                throw new HubException("Method does not exist.");
             }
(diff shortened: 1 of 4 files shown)


BrennanConroy:
I'll add a test for invoking a non-existant hub with arguments

src/Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubDispatcher.cs
```diff
@@ -136,7 +136,7 @@ public override IReadOnlyList<Type> GetParameterTypes(string methodName)
         {
             if (!_methods.TryGetValue(methodName, out var descriptor))
             {
-                return Type.EmptyTypes;
+                throw new HubException("Method doesn't exist.");
```
mikaelm12: Include the name of the method that doesn't exist

anurse: Let's do "does not"

anurse: It's already in the message that gets generated. See the tests.

mikaelm12: Derp

Issue #2714, opened 2018-08-01
[2.1.5] LongPolling: Setting connection to inactive while there is still an active poll


BrennanConroy:
@Eilon For 2.1.4 consideration

PR for change in release/2.2 https://github.com/aspnet/SignalR/pull/2697
Original Issue: #2700

Eilon:
@BrennanConroy - please fill in the template:

## Description

TODO: Describe the bug

## Customer Impact

TODO: Describe how it affects customers. Is there a workaround? How many customers reported?

## Regression from?

TODO: Fill in when it worked, when it stopped working

## Risk

High/Medium/Low/Very Low
TODO: Fill in explanation of risk.

Eilon:
@BrennanConroy can you provide info on whether there are customer reports of this?

BrennanConroy:
This issue was hitting the bug https://github.com/aspnet/SignalR/issues/2672

Eilon:
@BrennanConroy putting a reminder here for you to update the customer impact section.

natemcmaster:
There was no representative in shiproom today for this one, so this bug is not yet approved for 2.1.x.

cc @muratg

mkArtakMSFT:
Not approved for 2.1.X in Shiproom.

muratg:
Closing this. We'll reconsider if there's customer interest.

anurse:
We need to make sure someone from the team is at the shiproom for this since we need to understand why this is not being considered for a servicing release. It is a customer-reported issue that causes connections to be dropped randomly, as described in the impact above. My understanding has been that this meets the bar.

vivmishra:
Approved for 2.1.5

anurse:
@BrennanConroy please open a PR porting this change to `release/2.1` and link it here.

natemcmaster:
Fixed in https://github.com/aspnet/SignalR/pull/2769

PR #2769, opened 2018-08-08
Backport LongPolling race to 2.1.5

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2714

@anurse Take over if we need this in in the next couple days.


anurse:
This was approved in Shiproom based on the `release/2.2` PR

PR #2776, opened 2018-08-09
add version number to java client build

anurse:
Embed the build number into the Java client packages.

image

### DIFF
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -4,6 +4,8 @@ plugins {
 }
 
 group 'com.microsoft.aspnet'
+
+// If we're run from outside MSBuild, just assign a bogus dev version.
 version project.findProperty('packageVersion') ?: "99.99.99-dev"
 
 sourceCompatibility = 1.8
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -4,7 +4,7 @@ plugins {
 }
 
 group 'com.microsoft.aspnet'
-version '0.1.0-preview1'
+version project.findProperty('packageVersion') ?: "99.99.99-dev"
 
 sourceCompatibility = 1.8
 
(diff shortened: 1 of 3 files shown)


clients/java/signalr/build.gradle
```diff
@@ -54,8 +56,7 @@ task generatePOM {
                 }
             }
         }
-    }.writeTo("signalr-0.1.0-preview1.pom")
-
-    ant.move file: "signalr-0.1.0-preview1.pom",
-            todir: "${buildDir}/libs"
+    }.writeTo("${buildDir}/libs/signalr-${project.version}.pom")
```
mikaelm12: This is better 👍

version.props
```diff
@@ -1,11 +1,14 @@
 ﻿<Project>
   <PropertyGroup>
     <VersionPrefix>1.1.0</VersionPrefix>
-    <JavaClientVersion>0.1.0-preview1</JavaClientVersion>
+    <JavaVersionPrefix>0.1.0</JavaVersionPrefix>
     <VersionSuffix>preview1</VersionSuffix>
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' == 'rtm' ">$(VersionPrefix)</PackageVersion>
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' != 'rtm' ">$(VersionPrefix)-$(VersionSuffix)-final</PackageVersion>
     <BuildNumber Condition="'$(BuildNumber)' == ''">t000</BuildNumber>
+    <JavaPackageVersion>$(JavaVersionPrefix)-$(VersionSuffix)-$(BuildNumber)</JavaPackageVersion>
```
mikaelm12: So we'll get the `t000` suffix when there's no build number here, right? 
I remember running into issues with that,

mikaelm12: Oh, I remember now. I added the `preview1` to the `JavaClientVersion` because using the version suffix variable was appending the placeholder build number.
Disregard the previous comment

clients/java/signalr/build.gradle
anurse: Yeah, it was causing the POM file to appear even when just running the `compileJava` task. Not sure why but removing the `ant.move` and having the POM file written directly to the right place fixed it.

Issue #2750, opened 2018-08-07
Include timestamps in console log messages


anurse:
We don't include timestamps in the console log messages we write from the client. It would make correlation with server diagnostics much easier if we did.

* [x] TypeScript
* [ ] Java

anurse:
TypeScript and Java need this added. C# does it via the logger itself adding the timestamp.

anurse:
Recommended format:

```
[Timestamp] [Level] Message
```

Where `Timestamp` is the ISO date format: `YYYY-MM-ddTHH:MM:ssZ` (i.e. `2018-08-07T10:10:00Z`)

mikaelm12:
I'll assign this to myself and close it out with the Java log timestamps.

mikaelm12:
Done here: https://github.com/aspnet/SignalR/commit/6310cffd7c7a8e9beef5d7d774f557870e08b401

PR #2777, opened 2018-08-09
Add timestamps to log messages

mikaelm12:
Looks like: `[2018-08-09T12:13Z] [Debug] Received message of type PING`
Issue: https://github.com/aspnet/SignalR/issues/2750

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/ConsoleLogger.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/ConsoleLogger.java
@@ -9,7 +9,7 @@ import java.util.Date;
 
 public class ConsoleLogger implements Logger {
     private LogLevel logLevel;
-    private DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
+    private DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mmZ");
 
     public ConsoleLogger(LogLevel logLevel) {
             this.logLevel = logLevel;
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/ConsoleLogger.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/ConsoleLogger.java
@@ -9,10 +9,11 @@ import java.util.Date;
 
 public class ConsoleLogger implements Logger {
     private LogLevel logLevel;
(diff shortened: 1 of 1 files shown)


mikaelm12:
Ping 😄

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/ConsoleLogger.java
```diff
@@ -3,15 +3,23 @@
 
 package com.microsoft.aspnet.signalr;
 
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
 public class ConsoleLogger implements Logger {
     private LogLevel logLevel;
+    private DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
```
anurse: I think we want to use `Z` without quotes. If we use quotes, it's a literal `Z`, but if we don't use quotes it expands to the time zone in the correct format. So for Redmond you'd get `2018-08-10T12:54-0800`. If we know for sure it's always UTC, we can use `Z`, but I actually think logs should be in local time anyway.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/ConsoleLogger.java
```diff
@@ -29,7 +37,8 @@ public void log(LogLevel logLevel, String message) {
     @Override
     public void log(LogLevel logLevel, String formattedMessage, Object... args) {
         if (logLevel.value >= this.logLevel.value) {
-            formattedMessage = formattedMessage + "%n";
+            String timeStamp = dateFormat.format(new Date());
```
anurse: Is this in local time or UTC?

mikaelm12: Local time

mikaelm12:
Updated 👍

anurse:
Can you update your example above? With the change you made it should no longer be printing a `Z` at the end (or if it is, we'll need to figure out how to fix that because the `Z` indicates UTC which is inaccurate :))

mikaelm12:
`[2018-08-11T23:00-0700] [Debug] Received message of type PING`

anurse:
> Why is this inaccurate? Also, it's what we did for TS client

Because the TS client renders a UTC time, but the Java client is rendering a local time.

PR #2780, opened 2018-08-09
Add HubConnectionBuilder to Java client

mikaelm12:
Adding `HubConnectionBuilder` to Java client.
It currently lets you configure the url, transport, loglevel, or logger.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -24,7 +24,13 @@ public class HubConnection {
     public HubConnection(String url, Transport transport, Logger logger){
         this.url = url;
         this.protocol = new JsonHubProtocol();
-        this.logger = logger;
+
+        if(logger != null) {
+            this.logger = logger;
+        } else {
+            this.logger = new NullLogger();
+        }
+
         this.callback = (payload) -> {
 
             if (!handshakeReceived) {
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -25,7 +25,7 @@ public class HubConnection {
         this.url = url;
         this.protocol = new JsonHubProtocol();
 
-        if(logger != null) {
+        if (logger != null) {
             this.logger = logger;
         } else {
             this.logger = new NullLogger();
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
@@ -1,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+public class HubConnectionBuilder {
+    private boolean built = false;
+    private String url;
+    private Transport transport;
+    private Logger logger;
+
+    public HubConnectionBuilder setUrl(String url){
+        this.url = url;
+        return this;
+    }
+
+    public HubConnectionBuilder setTransport(Transport transport) {
+        this.transport = transport;
+        return this;
+    }
+
+    public HubConnectionBuilder setLogLevel(LogLevel logLevel) {
+        this.logger = new ConsoleLogger(logLevel);
+        return this;
+    }
+
+    public HubConnectionBuilder setLogger(Logger logger){
+        this.logger = logger;
+        return this;
+    }
+
+    public HubConnection build() throws Exception {
+        if(!built){
+            built = true;
+            return new HubConnection(url , transport, logger);
+        }
+        throw new Exception("HubConnectionBuilder allows creation only of a single instance of HubConnection.");
+    }
+}
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
@@ -14,7 +14,8 @@ public class HubConnectionBuilder {
         return this;
     }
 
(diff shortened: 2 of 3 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.microsoft.aspnet.signalr;
+
+public class HubConnectionBuilder {
+    private boolean built = false;
+    private String url;
+    private Transport transport;
+    private Logger logger;
+
+    public HubConnectionBuilder setUrl(String url){
```
anurse: We should match the .NET API. So `withUrl`. Also, can we put the transport as a parameter here so that we're matching the behavior of the C# client?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+        this.url = url;
+        return this;
+    }
+
+    public HubConnectionBuilder setTransport(Transport transport) {
+        this.transport = transport;
+        return this;
+    }
+
+    public HubConnectionBuilder setLogLevel(LogLevel logLevel) {
```
anurse: Make this `configureLogging`, like the JS client.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+        this.logger = new ConsoleLogger(logLevel);
+        return this;
+    }
+
+    public HubConnectionBuilder setLogger(Logger logger){
```
anurse: Make this an overload of `configureLogging`, again like the JS client.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+        this.logger = logger;
+        return this;
+    }
+
+    public HubConnection build() throws Exception {
+        if(!built){
+            built = true;
+            return new HubConnection(url , transport, logger);
```
anurse: nit: remove extra space between `url` and `,`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
mikaelm12: I personally hate the `withUrl` api we have in the C# client 😄

mikaelm12: I'm in favor of using excplicit calls to `withUrl`, and `withTransport`

anurse: How will that work with things like raw tcp transports? The idea of the `WithUrl` call in C# is that that is the call that sets up the "HTTP Transport" layer, which needs a URL and (optionally) the kind of Http Transport to use.

In C#, the TCP code would be something like:

```
    .WithEndpoint("example.com", port: 1234)
    .Build()
```

So there wouldn't be an option to declare a transport. If we have `withTransport` separately, then you could get in to trouble with things like:

```java
HubConnection connection = new HubConnectionBuilder()
    .withEndpoint("example.com", port: 1234)
    .withTransport(LongPolling) // What does this line mean now?
    .Build()
```

mikaelm12: *sigh Okay

mikaelm12: I still don't like `withUrl` though lol

anurse: I don't entirely disagree with that :).

mikaelm12:
Updated

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -24,7 +24,13 @@
     public HubConnection(String url, Transport transport, Logger logger){
         this.url = url;
         this.protocol = new JsonHubProtocol();
-        this.logger = logger;
+
+        if(logger != null) {
```
BrennanConroy: format

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.microsoft.aspnet.signalr;
+
+public class HubConnectionBuilder {
+    private boolean built = false;
+    private String url;
+    private Transport transport;
+    private Logger logger;
+
+    public HubConnectionBuilder withUrl(String url){
```
BrennanConroy: format file

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+public class HubConnectionBuilder {
+    private boolean built = false;
```
BrennanConroy: Default value should be false so don't need to set this

anurse:
> You should figure out if there are some auto-formatting settings or extensions you can use

#2676

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/2676 is on my todo list

mikaelm12:
Rebased this to include commit:https://github.com/aspnet/SignalR/commit/82cef1a0c6324d6f5f5485e1de152db1be93f32f so the VSTS builds should pass now

PR #2786, opened 2018-08-10
Update Gradle to 4.9

mikaelm12:
Running into an issue described here: https://github.com/gradle/gradle/issues/4931
Upgrading to a version greater than or equal to 4.8 (4.9 in this case) should fix it.

### DIFF
--- a/clients/java/signalr/gradle/wrapper/gradle-wrapper.properties
+++ b/clients/java/signalr/gradle/wrapper/gradle-wrapper.properties
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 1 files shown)


ryanbrandenburg:
Did we get OSS approval for the new version of Gradle?

mikaelm12:
> Did we get OSS approval for the new version of Gradle?

Do we have to do this for new versions? I wasn't aware.
@muratg Can we get this process started?

anurse:
Yes, we should register this. FYI though, this change fixes build failures, so it's fairly high priority.

mikaelm12:
@muratg Does that mean this has been OSS approved?

muratg:
@mikaelm12 No, perhaps @anurse or @Eilon can help.

mikaelm12:
@anurse and I submitted Gradle 4.9 in the OSS approval tool. Going to merge this now 👍

Issue #2788, opened 2018-08-10
Failure on Linux Jobs on VSTS


mikaelm12:
Looks like it can't start Chrome headless with out the `--no-sandbox` flag

````
2018-08-10T22:47:30.2380916Z   Checking for ChromeHeadless at /usr/bin/google-chrome...
2018-08-10T22:47:30.2403701Z   Located ChromeHeadless at /usr/bin/google-chrome.
2018-08-10T22:47:30.2420687Z   Checking for ChromiumHeadless at undefined...
2018-08-10T22:47:30.2436042Z   Unable to locate ChromiumHeadless. Skipping.
2018-08-10T22:47:30.2452499Z   Checking for FirefoxHeadless at firefox...
2018-08-10T22:47:30.2467369Z   Unable to locate FirefoxHeadless. Skipping.
2018-08-10T22:47:31.2537932Z   10 08 2018 22:47:31.250:INFO [karma]: Writing browser console to file: /opt/vsts/work/1/s/artifacts/logs/browserlogs.console.2018-08-10T22-47-31-004Z
2018-08-10T22:47:31.2768183Z   10 08 2018 22:47:31.275:INFO [karma]: Karma v2.0.4 server started at http://0.0.0.0:9876/
2018-08-10T22:47:31.2789273Z   10 08 2018 22:47:31.276:INFO [launcher]: Launching browser ChromeHeadless with unlimited concurrency
2018-08-10T22:47:31.2839481Z   10 08 2018 22:47:31.282:INFO [launcher]: Starting browser ChromeHeadless
2018-08-10T22:47:32.5276176Z   10 08 2018 22:47:32.525:ERROR [launcher]: Cannot start ChromeHeadless
2018-08-10T22:47:32.5299402Z   	[0810/224732.423558:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
2018-08-10T22:47:32.5314950Z   
2018-08-10T22:47:32.5329290Z   10 08 2018 22:47:32.525:ERROR [launcher]: ChromeHeadless stdout: 
2018-08-10T22:47:32.5345303Z   10 08 2018 22:47:32.525:ERROR [launcher]: ChromeHeadless stderr: [0810/224732.423558:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
2018-08-10T22:47:32.5362224Z   
[...]

PR #2794, opened 2018-08-11
[TEST] Testing ChromeHeadless Fix

mikaelm12:
We're seeing failures on linux because of issues with Chrome headless and our functional tests. I'm testing fixes
Issue: https://github.com/aspnet/SignalR/issues/2788


anurse:
I'll take this over :). Thanks for getting the core solution worked out!

Issue #2781, opened 2018-08-09
Throw if there was an error connecting with the WebsocketTransport


mikaelm12:
Currently what happens when start is called is that `connect` gets called on the underlying websocket but instead of throwing, it returns false if there is an error. So what happens is we don't see a failure until the handshake message is sent directly afterwards but fails because we couldn't connect. 

We should instead verify the result of connect and proceed the handshake step conditionally.

mikaelm12:
Done here:https://github.com/aspnet/SignalR/commit/85c934dcdd492290871803d7257948a594330d44

PR #2801, opened 2018-08-13
Throw if the Websocket cant connect

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/2781
Currently the Websockets transport doesn't fail immediately if it can't connect. It fails right after when it tries to send the handshake response. This ensures that the websocket was able to connect before continuing.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
@@ -40,10 +40,10 @@ public class WebSocketTransport implements Transport {
 
     @Override
     public void start() throws Exception {
-        logger.log(LogLevel.Debug, "Starting Websocket connection");
+        logger.log(LogLevel.Debug, "Starting Websocket connection.");
         webSocketClient = createWebSocket();
         if (!webSocketClient.connectBlocking()) {
-            String errorMessage = "There was an error starting the Websockets transport";
+            String errorMessage = "There was an error starting the Websockets transport.";
             logger.log(LogLevel.Debug, errorMessage);
             throw new Exception(errorMessage);
         }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
@@ -39,10 +39,14 @@ public class WebSocketTransport implements Transport {
     }
 
     @Override
-    public void start() throws InterruptedException {
+    public void start() throws Exception {
         logger.log(LogLevel.Debug, "Starting Websocket connection");
         webSocketClient = createWebSocket();
-        webSocketClient.connectBlocking();
+        if (!webSocketClient.connectBlocking()) {
+            String errorMessage = "There was an error starting the Websockets transport";
+            logger.log(LogLevel.Debug, errorMessage);
+            throw new Exception(errorMessage);
+        }
         logger.log(LogLevel.Information, "WebSocket transport connected to: %s", webSocketClient.getURI());
     }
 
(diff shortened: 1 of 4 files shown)


clients/java/signalr/src/test/java/WebSocketTransportTest.java
```diff
@@ -2,38 +2,22 @@
...
-    }
+    @Rule
+    public ExpectedException expectedEx = ExpectedException.none();
 
     @Test
-    public void checkWebsocketUrlProtocol() throws URISyntaxException {
-        WebSocketTransport webSocketTransport = new WebSocketTransport(this.url, new NullLogger());
-        assertEquals(this.expectedUrl, webSocketTransport.getUrl().toString());
+    public void WebsocketThrowsIIfItCantConnect() throws Exception {
+        expectedEx.expect(Exception.class);
```
anurse: We generally prefer using something like `Assert.Throws`. It looks like JUnit 5.x has [`assertThrows`](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertThrows-java.lang.Class-org.junit.jupiter.api.function.Executable-java.lang.String-) that we could use instead of `ExpectedException`.

mikaelm12: Do we want to take a new dependency for that though?

anurse: Isn't it just updating junit to the latest version? We need to register it, but since it's a test-only dependency, I don't think @Eilon will have any concerns. We should be using the latest version anyway.

mikaelm12: Changing to junit 5 breaks a lot of things and idk if it's worth it atm tbh

anurse: Ok, but it'll only get harder moving forward :). If it's a lot of work we don't need to do it in this PR.

mikaelm12: Yeah, I like the JUnit 5 assertThrows syntax more and it also provides a better way to do parametrized tests but I started making the changes and it was just a little to involved to be a side effect of this change imo. I'll file an issue and plan to do this in the preview2 wave.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -39,10 +39,14 @@ else if(url.startsWith(HTTP)){
     }
 
     @Override
-    public void start() throws InterruptedException {
+    public void start() throws Exception {
         logger.log(LogLevel.Debug, "Starting Websocket connection");
         webSocketClient = createWebSocket();
-        webSocketClient.connectBlocking();
+        if (!webSocketClient.connectBlocking()) {
+            String errorMessage = "There was an error starting the Websockets transport";
```
BrennanConroy: nit: full stop

clients/java/signalr/src/test/java/WebSocketTransportTest.java
```diff
@@ -2,38 +2,22 @@
...
-                {"wss://example.com", "wss://example.com"}});
-    }
+    @Rule
+    public ExpectedException expectedEx = ExpectedException.none();
 
     @Test
-    public void checkWebsocketUrlProtocol() throws URISyntaxException {
-        WebSocketTransport webSocketTransport = new WebSocketTransport(this.url, new NullLogger());
-        assertEquals(this.expectedUrl, webSocketTransport.getUrl().toString());
+    public void WebsocketThrowsIIfItCantConnect() throws Exception {
```
BrennanConroy: Extra `I` in test name

clients/java/signalr/src/test/java/WebSocketTransportUrlFormatTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.runners.Parameterized;
+
+import static org.junit.Assert.*;
+
+@RunWith(Parameterized.class)
+public class WebSocketTransportUrlFormatTest {
+    private String url;
+    private String expectedUrl;
+
+    public WebSocketTransportUrlFormatTest(String url, String expectedProtocol){
```
BrennanConroy: nit: space before `{`

Issue #2783, opened 2018-08-10
HubConnection.start fails after previous stop


ArtemAvramenko:
let connection = new signalR.HubConnectionBuilder()
        .withUrl('http://localhost:65432/myHub')
        .build();

    connection.start()
        .then(() => console.log('Connected'))
        .then(() => connection.stop())
        .then(() => console.log('Disconnected'))
        .then(() => connection.start()) // <--- fails here
        .then(() => console.log('Reconnected'))
        .catch(e => console.error(e.message));

The reconnect fails in any browser. The log shows this:
> Cannot send data if the connection is not in the 'Connected' State.

* Versions of Server-Side NuGet Packages: 2.3.0
* Versions of Client-Side NPM Packages: 1.0.2
* Are you using the C# client or the JavaScript client: Javascript
* The Server you are using: IIS/IIS Express (AspNetCoreModule)
* The Operating System on the Server: Windows 10
* The Operating System on the Client: Windows 10
* The Browser on the client, if using the JavaScript client: any (IE/Chrome/Edge/Firefox):

ArtemAvramenko:
Server log:

    Hosting environment: Development
    Content root path: C:\....
    Now listening on: http://localhost:2530
    Application started. Press Ctrl+C to shut down.


Client log:

    Debug: Starting HubConnection.
    Debug: Starting connection with transfer format 'Text'.
    Debug: Sending negotiation request: http://localhost:65432/myHub/negotiate
    Debug: Selecting transport 'WebSockets'
    Trace: (WebSockets transport) Connecting
    Information: WebSocket connected to ws://localhost:65432/myHub?id=...
    Debug: Sending handshake request.
    Trace: (WebSockets transport) sending data. String data of length 32.
    Information: Using HubProtocol 'json'.
    Connected
    Debug: Stopping HubConnection.
    Disconnected
    Debug: Starting HubConnection.
    Debug: Starting connection with transfer format 'Text'.
    Debug: Sending negotiation request: http://localhost:65432/appHub/negotiate
    Trace: (WebSockets transport) socket closed.
    Information: Connection disconnected.
    Debug: Sending handshake request.
    Error: Cannot send data if the connection is not in the 'Connected' State.

BrennanConroy:
@ArtemAvramenko Could you collect server logs with a more verbose log level? https://github.com/aspnet/SignalR/wiki/Diagnostics-Guide#server-side-logging

[...]

Issue #2557, opened 2018-06-28
Server Timeout in JavaScript Client doesnt work as expected


ottohak:
Hi guys,

I am currently developing an internal application that needs to know when the connection to my SignalR server is lost. My goal is to react to a lost connection in up to 3 seconds.

So I set the `serverTimeoutInMilliseconds` to 3000 and the 'KeepAliveInterval' to `TimeSpan.FromSeconds(1)`.

But this doesn't work as expected.
For a test I send every second a message to the client, when I walk far enough away from my WiFi 
access point my Surface stops getting messages for 20+ seconds and then I am getting a onclose event.

If test this with my Android Phone running Chrome the onclose message is almost immediately fired when the signal is lost.

Maybe this this the way Windows handles lost network connections, but if I set the `serverTimeoutInMilliseconds` this shouldn't be a problem, right?

Thanks for this great project!

### Please include as much of the following as you can in your bug report

* Versions of Server-Side NuGet Packages: tried 1.0.0 and 1.0.1
* Versions of Client-Side NuGet/NPM Packages: 1.0.0
* Are you using the C# client or the JavaScript client: Javascript
* The Server you are using (Kestrel/HttpSysServer/IIS/IIS Express/Azure Web App/etc.): Visual Studio Debug so I think Kestrel
* The Operating System on the Server (Windows/Linux/macOS): Windows
* The Operating System on the Client (Windows/Linux/macOS): Windows 10
* The Browser on the client, if using the JavaScript client (IE/Chrome/Edge/Firefox/etc.): Firefox 60 and Chrome 67
* Please use the Diagnostics Guide at https://github.com/aspnet/SignalR/wiki/Diagnostics-Guide to collect server-side and client-side logs and network traces if possible. Please **do not** put the traces and logs in the Issue text as they are quite large. The Diagnostics Guide describes how to attach them to the issue.
[Chrome_IP_Removed.txt](https://github.com/aspnet/SignalR/files/2144983/Chrome_IP_Removed.txt)

When in doubt, feel free to file the issue, we're happy to help answer questions. We also suggest using the `asp.net-core-signalr` tag on StackOverflow to ask questions.

BrennanConroy:
The HAR file you gave didn't complete the handshake request so it's not even running the code that makes use of `serverTimeoutInMilliseconds`. Could you try your scenario again but collect logs from a connection that actually connects?

ottohak:
[...]

PR #2804, opened 2018-08-14
[TS Client] Properly wait for WebSocket close

BrennanConroy:
There was a race between the `WebSocket.onclose` callback being called and `HubConnection.start` being called after `HubConnection.stop` returned. This was because we didn't wait for the `onclose` callback, which called `HttpConnection.stopConnection`, to finish before we returned from `WebSocket.close` which completed the `HubConnection.stop` call.

The fix is to call the `onclose` callback manually if `stop` is called, which conveniently fixed another issue we had.

Original issue: https://github.com/aspnet/SignalR/issues/2783
Bonus issue: https://github.com/aspnet/SignalR/issues/2557

No tests because we don't test race conditions.

### DIFF
--- a/clients/ts/signalr/src/WebSocketTransport.ts
+++ b/clients/ts/signalr/src/WebSocketTransport.ts
@@ -72,7 +72,7 @@ export class WebSocketTransport implements ITransport {
                 // webSocket will be null if the transport did not start successfully
                 this.logger.log(LogLevel.Trace, "(WebSockets transport) socket closed.");
                 if (this.onclose) {
-                    if (event.wasClean === false || event.code !== 1000) {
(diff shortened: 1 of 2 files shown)


clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -94,9 +94,20 @@ export class WebSocketTransport implements ITransport {
 
     public stop(): Promise<void> {
         if (this.webSocket) {
+            // Capture close callback and invoke it manually after clearing it
+            const close = this.webSocket.onclose;
```
anurse: As discussed in person, use a function instead.

Issue #2756, opened 2018-08-08
[2.1.x] Module not found: Can't resolve 'tslib' in '...@aspnet\signalr\dist\esm'


anurse:
Propose taking https://github.com/aspnet/SignalR/issues/2410 in 2.1.x branch

## Description

A missing dependency in `package.json` for SignalR's JavaScript client causes a compilation error when using SignalR from WebPack, NodeJS and other module loaders. The `<script>` tag reference style is not affected.

SignalR expects the `tslib` library to be available when running. This library provides runtime support for TypeScript features (`async`/`await`, class inheritance, etc.) when running on older ECMAScript runtimes. Normally, these helpers are automatically injected into the generated JavaScript, but we had configured TypeScript to import these from the `tslib` module instead. Unfortunately we neglected to put the dependency on `tslib` into our `package.json`, leading to runtime errors.

This was missed in testing because our tests run within a directory structure that already has `tslib`. We fixed this by changing TypeScript to **embed the helper code** rather than depending on the `tslib` library

## Customer Impact

When using the SignalR JavaScript client from an environment that uses JavaScript modules such as WebPack or NodeJS, the user will see compilation or runtime errors: `Can't resolve 'tslib' in '...'` because the script is trying to load a node module that isn't present.

This only occurs when **both** of the below are true:

1. The user is using a module loader or NodeJS. It does not reproduce when using a `<script>` tag to reference SignalR because that pattern uses a different generated script that has the helpers built in.

2. The user is **not** using Angular or some other library that brings in TypeScript. Most TypeScript-based libraries already depend upon `tslib` which means if the user is using Angular (for example), the `tslib` module will be present even though SignalR doesn't have the dependency (because Angular brought it in)

**Workaround**: There is a workaround. The consumer can add a dependency on `tslib` to their own application.

## Regression from?

None. This was in the initial release.

## Risk

**Low**: We are using TypeScript's built-in ability to inject the helpers. Also our tests now cover this usage scenario much better.

**Mitigations**: We restructured our tests to ensure they avoid accidentally leaking build-time dependencies when running tests.

Previously we had our node modules structured in a way that hid this issue. We had a top-level folder that brought in build-time depedencies (the TypeScript compiler, for example) and then subfolders for each package. Because node resolves modules from all parent directories, this meant that we didn't detect the missing dependency during testing because our tests ran from within this directory structure, which had `tslib` because the root directory had components that **did** bring in `tslib`.

```
clients/ts/
    node_modules/
[...]

PR #2808, opened 2018-08-14
TypeScript improvements and tslib dep (#2692)

BrennanConroy:
Issue: https://github.com/aspnet/SignalR/issues/2756


anurse:
Previously approved in shiproom via the linked issue.

clients/ts/signalr/src/Errors.ts
```diff
@@ -43,3 +45,23 @@ export class TimeoutError extends Error {
         this.__proto__ = trueProto;
     }
 }
+
+/** Error thrown when an action is aborted. */
+export class AbortError extends Error {
```
anurse: This wasn't part of the original change, looks like something from 2.2 that's leaking over.

clients/ts/FunctionalTests/ts/LogBannerReporter.ts
```diff
@@ -0,0 +1,19 @@
+export class LogBannerReporter implements jasmine.CustomReporter {
```
anurse: This looks like something from 2.2 leaking over.

PR #2813, opened 2018-08-15
Updated the lock files

davidfowl:
Not sure why nobody is updating these but it changes whenever I build.

### DIFF
--- a/clients/ts/FunctionalTests/package-lock.json
+++ b/clients/ts/FunctionalTests/package-lock.json
@@ -20,21 +20,21 @@
     "@aspnet/signalr-protocol-msgpack": {
       "version": "file:../signalr-protocol-msgpack",
       "requires": {
-        "msgpack5": "^4.0.2"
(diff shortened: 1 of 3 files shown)


anurse:
What version of npm did you do this with? If we're going to do it, we should do it at the latest version or they're just going to keep bouncing around. Looks like the latest is `6.3.0`. I ran the restore with that build and get a completely different diff from you (though they are still out of date).

Also, everybody should update their NPM version to the latest (cc @BrennanConroy @mikaelm12) because that's what gets us in this situation ;).

anurse:
Also it's important to know what can cause this file to change:

* A change in NPM version that includes new lock file content
* Addition/modification of a dependency in package.json
* Release of a new version of a dependency if that version becomes the new match for our condition in package.json (i.e. if we reference `^1.0.0` and a new patch release is made). NPM selects the highest matching version of packages rather than the lowest (which NuGet uses)

davidfowl:
Done!

Issue #2728, opened 2018-08-03
Add useful error message when connecting to ASP.NET SignalR server


anurse:
The ASP.NET Core SignalR Client is not compatible with the ASP.NET SignalR server. We should detect this in the clients and add a useful error message.

PR #2820, opened 2018-08-15
Detect ASP.NET SignalR server connection attempt

BrennanConroy:
#2728 
Wanted to get the ball rolling on this and see if anyone had a different idea of how to detect this.

I did a quick test and this did what I expected when connecting to an ASP.NET SignalR Server.

### DIFF
--- a/clients/ts/signalr/src/HttpConnection.ts
+++ b/clients/ts/signalr/src/HttpConnection.ts
@@ -147,7 +147,7 @@ export class HttpConnection implements IConnection {
                     }
 
                     if ((negotiateResponse as any).ProtocolVersion) {
-                        throw Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET *Core* SignalR Server.");
+                        throw Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server. See https://aka.ms/signalr-core-differences for details.");
                     }
 
                     if (negotiateResponse.url) {
--- a/clients/ts/signalr/src/HttpConnection.ts
+++ b/clients/ts/signalr/src/HttpConnection.ts
@@ -147,7 +147,7 @@ export class HttpConnection implements IConnection {
                     }
 
                     if ((negotiateResponse as any).ProtocolVersion) {
-                        throw Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server.");
+                        throw Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET *Core* SignalR Server.");
                     }
 
                     if (negotiateResponse.url) {
--- a/clients/ts/signalr/src/HttpConnection.ts
+++ b/clients/ts/signalr/src/HttpConnection.ts
@@ -146,6 +146,10 @@ export class HttpConnection implements IConnection {
                         return;
                     }
 
+                    if ((negotiateResponse as any).ProtocolVersion) {
+                        throw Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server.");
+                    }
+
                     if (negotiateResponse.url) {
                         url = negotiateResponse.url;
                     }
--- a/src/Microsoft.AspNetCore.Http.Connections.Common/NegotiateProtocol.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Common/NegotiateProtocol.cs
@@ -137,7 +137,7 @@ namespace Microsoft.AspNetCore.Http.Connections
                                         }
                                         break;
                                     case TryWebSocketsPropertyName:
-                                        throw new InvalidOperationException($"It seems like you are trying to connect to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server.");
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Http.Connections.Common/NegotiateProtocol.cs
```diff
@@ -18,6 +18,8 @@ public static class NegotiateProtocol
         private const string AvailableTransportsPropertyName = "availableTransports";
         private const string TransportPropertyName = "transport";
         private const string TransferFormatsPropertyName = "transferFormats";
+        // Used to detect ASP.NET SignalR Server connection attempt
+        private const string TryWebSocketsPropertyName = "TryWebSockets";
```
anurse: Can we use `ProtocolVersion` instead? It's still unique to ASP.NET SignalR but it's a little more of an obvious property to use when detecting.

src/Microsoft.AspNetCore.Http.Connections.Common/NegotiateProtocol.cs
```diff
@@ -134,6 +136,8 @@ public static NegotiationResponse ParseResponse(Stream content)
                                             }
                                         }
                                         break;
+                                    case TryWebSocketsPropertyName:
+                                        throw new InvalidOperationException($"It seems like you are trying to connect to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server.");
```
anurse: I just created https://aka.ms/signalr-core-differences that links to the SignalR Core "differences" doc. Maybe we should put that in the exception message? aka.ms links can be modified at any time, so it's like an FWLink in that we can redirect it when necessary. cc @muratg @Eilon @davidfowl

test/Microsoft.AspNetCore.Http.Connections.Tests/NegotiateProtocolTests.cs
```diff
@@ -55,6 +55,27 @@ public void ParsingNegotiateResponseMessageThrowsForInvalid(string payload, stri
...
+                "\"ProtocolVersion\":\"1.5\"," +
+                "\"TransportConnectTimeout\":30.0," +
+                "\"LongPollDelay\":0.0}";
+
+            var responseData = Encoding.UTF8.GetBytes(payload);
+            var ms = new MemoryStream(responseData);
+
+            var exception = Assert.Throws<InvalidDataException>(() => NegotiateProtocol.ParseResponse(ms));
+
+            Assert.Equal("It seems like you are trying to connect to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server.", exception.InnerException.Message);
```
davidfowl: seems? LOL. What is this clippy? "It looks like"

BrennanConroy: Because I want word-smithing! Give me suggestions!

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -146,6 +146,10 @@ export class HttpConnection implements IConnection {
                         return;
                     }
 
+                    if ((negotiateResponse as any).ProtocolVersion) {
+                        throw Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET *Core* SignalR Server.");
```
anurse: Remove the `*`, this isn't markdown :). Add the `aka.ms/signalr-core-differences` link as well

src/Microsoft.AspNetCore.Http.Connections.Common/NegotiateProtocol.cs
```diff
@@ -134,6 +136,8 @@ public static NegotiationResponse ParseResponse(Stream content)
                                             }
                                         }
                                         break;
+                                    case ProtocolVersionPropertyName:
+                                        throw new InvalidOperationException($"Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET *Core* SignalR Server.");
```
anurse: obvs. align the messages once you've updated them

clients/ts/signalr/src/HttpConnection.ts
anurse: Also I'm updating the doc to provide some more helpful content here: https://github.com/aspnet/Docs/pull/8148

src/Microsoft.AspNetCore.Http.Connections.Common/NegotiateProtocol.cs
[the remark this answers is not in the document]
anurse: It's not markdown, and we don't really do that kind of thing in our messages as it makes them harder to localize and use in search engines. I think we should just leave it as-is.

davidfowl:
We also need ASP.NET SignalR client -> ASP.NET Core SignalR Server

BrennanConroy:
https://github.com/SignalR/SignalR/issues/4160

test/Microsoft.AspNetCore.Http.Connections.Tests/NegotiateProtocolTests.cs
```diff
@@ -55,6 +55,27 @@ public void ParsingNegotiateResponseMessageThrowsForInvalid(string payload, stri
             Assert.Equal(expectedMessage, exception.InnerException.Message);
         }
 
+        [Fact]
+        public void ParsingClassicSignalRResponseThrowsError()
```
anurse: nit: `ParsingAspNetSignalRResponseThrowsError`

Issue #2803, opened 2018-08-14
Add support for registering OnClosed callbacks for the Java client


mikaelm12:
We need to allow users to author callbacks that run when a HubConnection is closed.

mikaelm12:
Done here: 8c9093332b90e8a64497ebb8bc06ed84a99b1615

PR #2821, opened 2018-08-16
Support for Adding OnClosed Logic

mikaelm12:
Lets users register a callback to run when a `HubConnection` is stopped.
Issue:https://github.com/aspnet/SignalR/issues/2803

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -165,8 +165,9 @@ public class HubConnection {
         connectionState = HubConnectionState.DISCONNECTED;
         logger.log(LogLevel.Information, "HubConnection stopped.");
         if (onClosedCallbackList != null){
+            HubException hubException = new HubException(errorMessage);
             for (Consumer<Exception> callback: onClosedCallbackList) {
-                callback.accept(new HubException(errorMessage));
+                callback.accept(hubException);
             }
         }
     }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -20,6 +20,7 @@ public class HubConnection {
     private static final String RECORD_SEPARATOR = "\u001e";
     private HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
     private Logger logger;
+    private Action1<Exception> onClosedCallback;
 
     public HubConnection(String url, Transport transport, Logger logger){
         this.url = url;
@@ -162,6 +163,9 @@ public class HubConnection {
         transport.stop();
         connectionState = HubConnectionState.DISCONNECTED;
         logger.log(LogLevel.Information, "HubConnection stopped.");
+        if (onClosedCallback != null){
+            onClosedCallback.invoke(new Exception(errorMessage));
+        }
     }
 
     /**
@@ -405,5 +409,9 @@ public class HubConnection {
     public void remove(String name) {
         handlers.remove(name);
         logger.log(LogLevel.Trace, "Removing handlers for client method %s" , name);
+    }
+
+    public void onClosed(Action1<Exception> callback) {
+        this.onClosedCallback = callback;
     }
 }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -8,6 +8,7 @@ import com.google.gson.JsonArray;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -162,6 +163,9 @@ private void stop(String errorMessage) {
         transport.stop();
         connectionState = HubConnectionState.DISCONNECTED;
         logger.log(LogLevel.Information, "HubConnection stopped.");
+        if (onClosedCallback != null){
+            onClosedCallback.invoke(new Exception(errorMessage));
```
mikaelm12: Now I'm just creating a new Exception her but eventually I'll pass in a HubException.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -20,6 +20,7 @@
     private static final String RECORD_SEPARATOR = "\u001e";
     private HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
     private Logger logger;
+    private Action1<Exception> onClosedCallback;
```
mikaelm12: I just reused one of the `Action` types I defined earlier. Should I make a specific `OnClosedCallback `interface though? Seems unnecessary imo

anurse: I really don't know what the Java Way(TM) is here :). It looks like there is a type called [`Consumer<T>`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html) which is basically `Action<T>`. Since Java types can't differ by generic parameter count alone, we still need our custom handlers for `.on` but we could just use `Consumer<T>` here I guess.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: Will you do that update in this PR? Seems simple enough, just needs the new class.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -406,4 +410,8 @@ public void remove(String name) {
         handlers.remove(name);
         logger.log(LogLevel.Trace, "Removing handlers for client method %s" , name);
     }
+
+    public void onClosed(Action1<Exception> callback) {
+        this.onClosedCallback = callback;
```
anurse: Should we support multiple listeners? TypeScript and .NET do.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: Sure I can do that here

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: Yeah, you're right

mikaelm12: Curious to know if anyone actually registers more than one OnClosed handlers though

mikaelm12:
Updated

mikaelm12:
Actually, need another test for checking that multiple  OnClosed handlers run.

mikaelm12:
Updated now

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -162,6 +164,11 @@ private void stop(String errorMessage) {
         transport.stop();
         connectionState = HubConnectionState.DISCONNECTED;
         logger.log(LogLevel.Information, "HubConnection stopped.");
+        if (onClosedCallbackList != null){
+            for (Consumer<Exception> callback: onClosedCallbackList) {
+                callback.accept(new HubException(errorMessage));
```
BrennanConroy: Move `new HubException(errorMessage)` above for loop

davidfowl: What happens if you call start in the close callback?

mikaelm12: It will try to restart the connection

davidfowl:
@mikaelm12 my comment never made it I don't think. What happens when you call start from the closed callback?

mikaelm12:
It will try to restart the connection. Should I make the `HubConnection` explicitly non re-startable for the time being?

davidfowl:
> It will try to restart the connection. Should I make the HubConnection explicitly non re-startable for the time being?

No it's fine. I'm more worried about being in an unknown state when stop is being called with more user code running. As long as we make sure we design for user code running then it should be fine.

PR #2823, opened 2018-08-16
Update build to Node 10 and use `npm ci`

### LINKED ISSUE #2817: Update build to use `npm ci` to improve reliability
https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

This is not yet available in Node 8.x (see https://github.com/nodejs/node/pull/21593).

natemcmaster:
Resolves #2817 

Changes:
* Set minimum node version to 10.9 (requires TeamCity updates)
* Use `npm ci` in CI builds instead of `npm install`


anurse:
@BrennanConroy this looks like the orphaned ping timer issue we saw earlier. This should have been resolved earlier right?

BrennanConroy:
Yeah, we added the connected check before the ping. So there may be another bug here

natemcmaster:
ok. re-triggered the PR with latest changes from 2.2

natemcmaster:
Still getting the same errors.

BrennanConroy:
Yeah super strange, seems to consistently repro on VSTS when you upgrade node. Can't hit it locally with 10.9.0

natemcmaster:
I'm closing this for now because I don't want to deal with the hassle of updating TeamCity agents. Will re-address when we finish the VSTS move.

PR #2824, opened 2018-08-17
Check connection state before start stop and send

mikaelm12:
The PR makes sure that the client is in the correct connection state before doing the operations in the start stop and send methods.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -196,7 +196,7 @@ public class HubConnection {
      */
     public void send(String method, Object... args) throws Exception {
         if (connectionState != HubConnectionState.CONNECTED) {
-            throw new Exception("The 'send' method cannot be called if the connection is not active");
+            throw new HubException("The 'send' method cannot be called if the connection is not active");
         }
 
         InvocationMessage invocationMessage = new InvocationMessage(method, args);
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -142,6 +142,10 @@ public class HubConnection {
      * @throws Exception An error occurred while connecting.
      */
     public void start() throws Exception {
+        if (connectionState != HubConnectionState.DISCONNECTED) {
+            return;
+        }
+        
         logger.log(LogLevel.Debug, "Starting HubConnection");
         transport.setOnReceive(this.callback);
         transport.start();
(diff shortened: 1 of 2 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -179,6 +187,10 @@ public void stop() {
      * @throws Exception If there was an error while sending.
      */
     public void send(String method, Object... args) throws Exception {
+        if (connectionState != HubConnectionState.CONNECTED) {
+            throw new Exception("Cannot send data if the connection is not in the 'Connected' State.");
```
mikaelm12: This will be `HubException` once https://github.com/aspnet/SignalR/pull/2821 is merged.

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -597,6 +599,34 @@ public void ReceiveHandshakeResponseAndMessage() throws Exception {
...
+        hubConnection.stop();
+        assertEquals(HubConnectionState.DISCONNECTED, hubConnection.getConnectionState());
+    }
+
+    @Rule
+    public ExpectedException exceptionRule = ExpectedException.none();
+
+    @Test
+    public void CantSendBeforeStart() throws Exception {
+        exceptionRule.expect(Exception.class);
```
mikaelm12: This will be `HubExcecption` after https://github.com/aspnet/SignalR/pull/2821

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
BrennanConroy: Should the message be `Connected` or `CONNECTED` since that's the enum name

clients/java/signalr/src/test/java/HubConnectionTest.java
```diff
@@ -597,6 +599,34 @@ public void ReceiveHandshakeResponseAndMessage() throws Exception {
...
+
+        hubConnection.stop();
+        assertEquals(HubConnectionState.DISCONNECTED, hubConnection.getConnectionState());
+    }
+
+    @Rule
+    public ExpectedException exceptionRule = ExpectedException.none();
+
+    @Test
+    public void CantSendBeforeStart() throws Exception {
```
BrennanConroy: nit: Cannot

anurse: nit: UnableToTransmitInvocationMessagePriorToConnectionEstablishmentUsingTheStartMethod :trollface:

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: Okay, we should update the TS client for consistency

mikaelm12: https://github.com/aspnet/SignalR/blob/8c9093332b90e8a64497ebb8bc06ed84a99b1615/clients/ts/signalr/src/HttpConnection.ts#L95-L98

anurse: 💯

mikaelm12:
Nate said we can merge again 👍

PR #2840, opened 2018-08-20
Ignore failures from sending ping

BrennanConroy:
There is a race in the ping timer where the connection can close after the `sendMessage` call goes async and then it will throw inside `sendMessage` with a connection closed error and that can trigger unhandledexception handlers which is bad.

Should fix PR #2823


BrennanConroy:
FYI @natemcmaster

PR #2847, opened 2018-08-21
Fix TS msgpack on IE and Ping breaking connection on server

BrennanConroy:
Simpler fix than adding a new polyfill 😄

### DIFF
--- a/clients/ts/signalr-protocol-msgpack/src/MessagePackHubProtocol.ts
+++ b/clients/ts/signalr-protocol-msgpack/src/MessagePackHubProtocol.ts
@@ -12,7 +12,8 @@ import { BinaryMessageFormat } from "./BinaryMessageFormat";
 
 // constant encoding of the ping message
 // see: https://github.com/aspnet/SignalR/blob/dev/specs/HubProtocol.md#ping-message-encoding-1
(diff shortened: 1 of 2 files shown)


clients/ts/signalr-protocol-msgpack/src/MessagePackHubProtocol.ts
```diff
@@ -12,7 +12,8 @@ import { BinaryMessageFormat } from "./BinaryMessageFormat";
 
 // constant encoding of the ping message
 // see: https://github.com/aspnet/SignalR/blob/dev/specs/HubProtocol.md#ping-message-encoding-1
-const SERIALIZED_PING_MESSAGE: ArrayBuffer = Uint8Array.from([0x91, MessageType.Ping]).buffer;
+// Don't use Uint8Array.from as IE does not support it
+const SERIALIZED_PING_MESSAGE: ArrayBuffer = new Uint8Array([0x91, MessageType.Ping]).buffer;
```
anurse: This doesn't fix the missing length prefix thing though does it? Let's add that here.

BrennanConroy:
Ping (pun intended)

PR #2851, opened 2018-08-21
remove extraneous file

anurse:
This file isn't needed and was left behind when we moved to Karma.


mikaelm12:
After green builds of course 😄

Issue #2694, opened 2018-07-28
Introduce CompletableFuture & Async Await to the Java Client 


mikaelm12:
Currently everything is synchronous. We can stop blocking everywhere with Java's `CompletableFuture` type and then we can make it nice with EA's async await library

CompletableFuture - https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html
EA Async Await library for Java - https://github.com/electronicarts/ea-async

mikaelm12:
There are no milestones for previews but this is planned to go into preview2

anurse:
Also consider [RxJava](https://github.com/ReactiveX/RxJava)

anurse:
We should look at this now. I think it's going to be the place where we're going to need the most feedback.

mikaelm12:
Yeah, been thinking about this for sure but I guess I'm waiting for Preview1 to go out the door first

anurse:
Let's do futures and start by just wrapping the blocking stuff in Futures.

mikaelm12:
Done here: f27df1d61e39d5c9995d65edd2212fa36e24ea94

PR #2858, opened 2018-08-23
[WIP] Start messing around with async APIs

mikaelm12:
I'm starting work on issue: https://github.com/aspnet/SignalR/issues/2694 to add Async APIs to the Java client.

Just getting started adding the APIs. This is an experimental pr

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Chat.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Chat.java
@@ -4,6 +4,7 @@
 package com.microsoft.aspnet.signalr;
 
 import java.util.Scanner;
+import java.util.concurrent.CompletableFuture;
 
 public class Chat {
     public static void main(String[] args) throws Exception {
@@ -27,9 +28,10 @@ public class Chat {
             });
 
             //This is a blocking call
-            hubConnection.start();
+            hubConnection.startAsync().get();
+            hubConnection.send("Send", "RACE CONDITION");
 
-            while (!input.equals("leave")){
+        while (!input.equals("leave")){
                 // Scans the next token of the input as an int.
                 input = reader.nextLine();
                 hubConnection.send("Send", input);
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -8,6 +8,7 @@ import com.google.gson.JsonArray;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 
 public class HubConnection {
@@ -156,6 +157,33 @@ public class HubConnection {
     }
 
     /**
+     * Starts a connection to the server asynchronously.
+     * @throws Exception An error occurred while connecting.
+     */
+    public CompletableFuture startAsync() throws Exception {
+        if (connectionState != HubConnectionState.DISCONNECTED) {
+            return CompletableFuture.completedFuture(null);
+        }
+
+        logger.log(LogLevel.Debug, "Starting HubConnection");
+        transport.setOnReceive(this.callback);
+        CompletableFuture transportStartFuture = transport.startAsync();
+        String handshake = HandshakeProtocol.createHandshakeRequestMessage(new HandshakeRequestMessage(protocol.getName(), protocol.getVersion()));
+        CompletableFuture startFuture = transportStartFuture.thenApply((future) -> {
+            CompletableFuture sendFuture = null;
+            try {
+                sendFuture = transport.sendAsync(handshake);
+                connectionState = HubConnectionState.CONNECTED;
+                logger.log(LogLevel.Information, "HubConnected started.");
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+            return  sendFuture;
+        });
+        return startFuture;
+    }
+
+    /**
      * Stops a connection to the server.
      */
     private void stop(String errorMessage) {
@@ -203,6 +231,24 @@ public class HubConnection {
         String message = protocol.writeMessage(invocationMessage);
         logger.log(LogLevel.Debug, "Sending message");
         transport.send(message);
+    }
+
+    /**
+     * Invokes a hub method on the server using the specified method name.
+     * Does not wait for a response from the receiver.
+     * @param method The name of the server method to invoke.
+     * @param args The arguments to be passed to the method.
+     * @throws Exception If there was an error while sending.
+     */
+    public CompletableFuture sendAsync(String method, Object... args) throws Exception {
+        if (connectionState != HubConnectionState.CONNECTED) {
+            throw new HubException("The 'send' method cannot be called if the connection is not active");
+        }
+
+        InvocationMessage invocationMessage = new InvocationMessage(method, args);
+        String message = protocol.writeMessage(invocationMessage);
+        logger.log(LogLevel.Debug, "Sending message");
+        return transport.sendAsync(message);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Chat.java
```diff
@@ -27,9 +28,10 @@ public static void main(String[] args) throws Exception {
             });
 
             //This is a blocking call
-            hubConnection.start();
+            hubConnection.startAsync().get();
```
mikaelm12: Without "awaiting" this call to start we have a race between start and send and the send fails

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -205,6 +233,24 @@ public void send(String method, Object... args) throws Exception {
         transport.send(message);
     }
 
+    /**
+     * Invokes a hub method on the server using the specified method name.
+     * Does not wait for a response from the receiver.
+     * @param method The name of the server method to invoke.
+     * @param args The arguments to be passed to the method.
+     * @throws Exception If there was an error while sending.
```
mikaelm12: Returns CompletableFuture

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Transport.java
```diff
@@ -3,9 +3,13 @@
 
 package com.microsoft.aspnet.signalr;
 
+import java.util.concurrent.CompletableFuture;
+
 public interface Transport {
     void start() throws Exception;
+    CompletableFuture startAsync() throws Exception;
```
mikaelm12: Just adding the async APIs for now so I don't have to change everything.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -155,6 +156,33 @@ public void start() throws Exception {
...
+        transport.setOnReceive(this.callback);
+        CompletableFuture transportStartFuture = transport.startAsync();
+        String handshake = HandshakeProtocol.createHandshakeRequestMessage(new HandshakeRequestMessage(protocol.getName(), protocol.getVersion()));
+        CompletableFuture startFuture = transportStartFuture.thenApply((future) -> {
+            CompletableFuture sendFuture = null;
+            try {
+                sendFuture = transport.sendAsync(handshake);
+                connectionState = HubConnectionState.CONNECTED;
+                logger.log(LogLevel.Information, "HubConnected started.");
+            } catch (Exception e) {
```
anurse: This should probably be a `.handle` call rather than a `try...catch`. Otherwise we miss asynchronous errors.

Also, to match the synchronous behavior, the `connectionState` change and the log statement have to be chained after the `sendAsync` with `.thenApply` because they should be written when the handshake is finished. Although, I don't think we need to print the stack trace of the error anyway, so we can probably skip the `.handle` part

Something like this:

```java
return transport.sendAsync(handshake)
    .thenApply((future) -> {
        connectionState = HubConnectionState.CONNECTED;
        logger.log(LogLevel.Information, "HubConnection started.");
    });
```

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -205,6 +233,24 @@ public void send(String method, Object... args) throws Exception {
...
     }
 
+    /**
+     * Invokes a hub method on the server using the specified method name.
+     * Does not wait for a response from the receiver.
+     * @param method The name of the server method to invoke.
+     * @param args The arguments to be passed to the method.
+     * @throws Exception If there was an error while sending.
+     */
+    public CompletableFuture sendAsync(String method, Object... args) throws Exception {
```
anurse: We shouldn't actually return `CompletableFuture`, we should return [`CompletionStage`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html).

`CompletionStage` is what consumers need to chain other code to run when the Future completes. `CompletableFuture` gives the caller the ability to "complete" the task, which isn't their job :).

From what I can tell:

* `CompletableFuture` (Java) <==> `TaskCompletionSource` (.NET)
* `CompletionStage` (Java) <==> `Task` (.NET)

`CompletableFuture` just implements `CompletionStage` so we can just return the future instance, I just think we should change the type of the return value.

The one thing that gets slightly harder is synchronously blocking since `CompletionStage` doesn't have the `.get` method. However, it does have `.toCompletableFuture()` so if you want to be synchronous you can call `.sendAsync().toCompletableFuture().get()`

anurse: That also strongly discourages people from using `.get()` which is the equivalent of `.Wait()` or `.Result` on `Task`, both of which should be avoided in .NET.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -50,11 +53,29 @@ public void start() throws Exception {
         logger.log(LogLevel.Information, "WebSocket transport connected to: %s", webSocketClient.getURI());
     }
 
+    @Override
+    public CompletableFuture startAsync() throws Exception {
+        return CompletableFuture.runAsync(() -> {
```
anurse: So I assume this is the equivalent of `Task.Run`... not ideal at all. I think maybe we should build a simple wrapper around the WebSocket client that uses futures. It looks like this should be straightforward enough. Instead of `connectBlocking`, we just create a `CompletableFuture`, call `connect` and return the future. In `onOpen` you complete the future.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -155,6 +156,33 @@ public void start() throws Exception {
         logger.log(LogLevel.Information, "HubConnected started.");
     }
 
+    /**
+     * Starts a connection to the server asynchronously.
+     * @throws Exception An error occurred while connecting.
+     */
+    public CompletableFuture startAsync() throws Exception {
```
davidfowl: Is the convention to use the async suffix in java?

anurse: I'm not entirely sure. For example, the [Azure Key Vault Client](https://github.com/Azure/azure-keyvault-java/blob/dev/azure-keyvault/src/main/java/com/microsoft/azure/keyvault/KeyVaultClientCustom.java#L152) and [Azure Service Bus Client](https://github.com/Azure/azure-service-bus-java/blob/dev/azure-servicebus/src/main/java/com/microsoft/azure/servicebus/QueueClient.java#L409) both do this, and has sync/async versions using the `Async` suffix to distinguish them. I presume other Azure SDKs do it, but I haven't checked them all.

However, the `Async` suffix [has a different meaning in `CompletableFuture` itself](https://stackoverflow.com/a/47489654/29813). In that case, `thenApply` indicates that the continuation should be run in-line on the same thread, whereas `thenApplyAsync` indicates that the continuation should be run on a new thread-pool thread.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: Looking at some Azure SDKs, it looks like just returning `CompletableFuture` is the common pattern. For example, the [Azure Service Bus SDK for Java](https://github.com/Azure/azure-service-bus-java/blob/dev/azure-servicebus/src/main/java/com/microsoft/azure/servicebus/QueueClient.java#L176).

So I guess just go with what they do.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: It doesn't look like it, but I'm just prototyping here so I didn't want to mess with the existing APIs

mikaelm12:
Closing in favor of https://github.com/aspnet/SignalR/pull/2944

Issue #2875, opened 2018-08-28
NullReferenceException in HttpConnectionManager if application has already been started


creativbox:
When using SignalR in a modular application like OrchardCore, the HttpConnectionManager will throw a NRE in Start() because the application has already been started (i.e. IApplicationLifeTime.ApplicationStarted is already signaled and _nextHeartbeat gets set too late). 

Microsoft.AspNetCore.SignalR version 1.0.3
https://github.com/OrchardCMS/OrchardCore/issues/2268

BrennanConroy:
Fixed in https://github.com/aspnet/SignalR/pull/2878

PR #2878, opened 2018-08-28
Initialize HttpConnectionManager before registering callbacks

BrennanConroy:
#2875


davidfowl:
Is this backport worthy?

anurse:
So far we've only had a report of it occurring in Orchard scenarios, I don't know that it meets the bar unless we hear more feedback.

Issue #2879, opened 2018-08-28
Re-enable logging for SSE tests


BrennanConroy:
Because of https://github.com/dotnet/coreclr/issues/19698 (.NET Core) / https://devdiv.visualstudio.com/DevDiv/_workitems/edit/672340 (.NET Framework) 2 of our tests are flaky when logging is enabled. We are removing the logging until the issue has been fixed. This is tracking the fix.

See [commit](https://github.com/aspnet/SignalR/commit/f314362ea5d239c38a0c45e212212737b151a42b) for the logging removed.

BrennanConroy:
Github "fixes" is flaky... https://github.com/aspnet/SignalR/pull/3320

Issue #2835, opened 2018-08-19
Test failure: SSETransportStopsWithErrorIfSendingMessageFails


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=535030&buildTypeId=XPlat_Windows_Win7_Universe) occasionally with the following error:
```
Assert.Throws() Failure
Expected: typeof(System.Net.Http.HttpRequestException)
Actual:   typeof(System.TimeoutException): Operation in SSETransportStopsWithErrorIfSendingMessageFails timed out at /_/test/Microsoft.AspNetCore.SignalR.Client.Tests/ServerSentEventsTransportTests.cs:211
   at System.Threading.Tasks.TaskExtensions.<OrTimeout>d__6`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
------- Stdout: -------
| [0.000s] TestLifetime Information: Starting test SSETransportStopsWithErrorIfSendingMessageFails at 2018-08-19T08:21:48
| [0.009s] TestLifetime Information: Starting test SSETransportStopsWithErrorIfSendingMessageFails at 2018-08-19T08:21:48
| [0.010s] Microsoft.AspNetCore.Http.Connections.Client.Internal.ServerSentEventsTransport Information: Starting transport. Transfer mode: Text.
| [4.979s] TestLifetime Information: Finished test SSETransportStopsWithErrorIfSendingMessageFails in 4.8914184s
| [4.991s] TestLifetime Information: Finished test SSETransportStopsWithErrorIfSendingMessageFails in 4.9097814s
```
Other tests within that build may have failed with a similar message, but they are not listed here. Check the link above for more info.

This test failed on master.

CC @muratg

This issue was made automatically. If there is a problem contact ryanbrandenburg.

ryanbrandenburg:
Please use this workflow to address this flaky test issue, including checking applicable checkboxes and filling in the applicable "TODO" entries:

* Is this actually a flaky test?
  * No, this is a regular test failure, fix the test/product (TODO: Link to commit/PR)
  * Yes, proceed below...

* Is this test failure caused by product code flakiness? (Either this product, or another product this test depends on.)
  * [ ] File a bug against the product (TODO: Link to other bug)
[...]

Issue #2832, opened 2018-08-18
Test failure: SSEWontStartIfSuccessfulConnectionIsNotEstablished


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=534598&buildTypeId=XPlat_Windows_Win2008r2_Universe) occasionally with the following error:
```
System.TimeoutException : Operation in WithConnectionAsync timed out at /_/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Helpers.cs:69
   at System.Threading.Tasks.TaskExtensions.<OrTimeout>d__2.MoveNext() in /_/test/Microsoft.AspNetCore.SignalR.Tests.Utils/TaskExtensions.cs:line 36
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.SignalR.Client.Tests.HttpConnectionTests.<WithConnectionAsync>d__9.MoveNext() in /_/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Helpers.cs:line 69
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.SignalR.Client.Tests.HttpConnectionTests.ConnectionLifecycle.<SSEWontStartIfSuccessfulConnectionIsNotEstablished>d__11.MoveNext() in /_/test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.ConnectionLifecycle.cs:line 346
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #2880, opened 2018-08-28
Remove logging from flaky SSE tests

BrennanConroy:
#2832 #2835 

https://github.com/aspnet/SignalR/issues/2879 tracking re-enabling logging

### DIFF
--- a/clients/ts/signalr-protocol-msgpack/package-lock.json
+++ b/clients/ts/signalr-protocol-msgpack/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@aspnet/signalr-protocol-msgpack",
-  "version": "1.1.0-preview1-t000",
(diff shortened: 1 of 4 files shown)


clients/ts/signalr-protocol-msgpack/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aspnet/signalr-protocol-msgpack",
-  "version": "1.1.0-preview1-t000",
+  "version": "1.1.0-preview2-t000",
```
anurse: This was missed in the branding update @natemcmaster did (our bad though, for not catching it in review). Maybe we should like... fail the build on the CI if this file changes. We need a way to make sure changes get checked in.

Issue #2871, opened 2018-08-27
Support Negotiation on the Java client


mikaelm12:
We currently only support 1 transport (connecting directly with websockets), but we need to support the Negotiate step for adding support for the SignalR Service.
That issue here is tracked here: https://github.com/aspnet/SignalR/issues/2862

mikaelm12:
Fixed here: dc6088bf21307df7c10c6eab4f9beaf3ffc91cc7

Issue #2862, opened 2018-08-24
Add support for the Azure SignalR Service


mikaelm12:
We need to add support for using the Java client with the SignalR service.

mikaelm12:
I tied this with the Negotiate pr
Fixed here: dc6088bf21307df7c10c6eab4f9beaf3ffc91cc7

Issue #2859, opened 2018-08-24
SignalR Client should preserve query string when getting a redirect negotiation reponse


xscript:
Right now, when client getting a redirect response (with `url` and `accessToken` properties), the returned URL will be used and query string in the original request URL is abandoned, see [JavaScript client](https://github.com/aspnet/SignalR/blob/release/2.2/clients/ts/signalr/src/HttpConnection.ts#L153-L155) and [C# client](https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs#L318-L321).
Per offline talk with @davidfowl , SignalR client should preserve the original query strings.

BrennanConroy:
We may want to consider taking this in a 2.1.X patch too.

anurse:
I think this would be a breaking change though, and might cause unintentional information disclosure. Need to talk with @davidfowl about this.

BrennanConroy:
We might punt this because of https://github.com/Azure/azure-signalr/pull/184

anurse:
If that's the way this is being resolved, then yeah, we don't need to make a change here. Depends on if @davidfowl still believes this to be a bug in general

BrennanConroy:
@davidfowl Thoughts?

PR #2882, opened 2018-08-28
Negotiate + SignalR Service Support for the Java client

mikaelm12:
Added support for the negotiation dance between the client and server and also added support for the SignalR service.

Was originally going to make these two different PRs but I really wanted to see the Java client working with the service 😄 
Issue: https://github.com/aspnet/SignalR/issues/2871
and issue: https://github.com/aspnet/SignalR/issues/2862

Need to add tests

### DIFF
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -19,6 +19,7 @@ dependencies {
     implementation "org.java-websocket:Java-WebSocket:1.3.8"
     implementation 'com.google.code.gson:gson:2.8.5'
     implementation 'org.apache.httpcomponents:httpclient:4.5.6'
+    implementation 'com.squareup.okhttp3:okhttp:3.11.0'
 }
 
 task sourceJar(type: Jar) {
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -18,7 +18,7 @@ dependencies {
     testImplementation group: 'junit', name: 'junit', version: '4.12'
     implementation "org.java-websocket:Java-WebSocket:1.3.8"
     implementation 'com.google.code.gson:gson:2.8.5'
-    compile 'org.apache.httpcomponents:httpclient:4.5.6'
+    implementation 'org.apache.httpcomponents:httpclient:4.5.6'
 }
 
 task sourceJar(type: Jar) {
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -18,6 +18,7 @@ dependencies {
     testImplementation group: 'junit', name: 'junit', version: '4.12'
     implementation "org.java-websocket:Java-WebSocket:1.3.8"
     implementation 'com.google.code.gson:gson:2.8.5'
+    compile 'org.apache.httpcomponents:httpclient:4.5.6'
 }
 
 task sourceJar(type: Jar) {
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -18,7 +18,6 @@ dependencies {
     testImplementation group: 'junit', name: 'junit', version: '4.12'
     implementation "org.java-websocket:Java-WebSocket:1.3.8"
     implementation 'com.google.code.gson:gson:2.8.5'
-    implementation 'org.apache.httpcomponents:httpclient:4.5.6'
     implementation 'com.squareup.okhttp3:okhttp:3.11.0'
 }
 
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -101,7 +101,7 @@ public class HubConnection {
             this.transport = transport;
         }
     }
-    
+
     public HubConnection(String url, Transport transport, Logger logger){
         this(url, transport, logger, true);
     }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -7,7 +7,9 @@ import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Consumer;
 
 public class HubConnection {
@@ -25,6 +27,7 @@ public class HubConnection {
     private boolean skipNegotiate = false;
     private NegotiateResponse negotiateResponse;
     private String accessToken;
+    private Map<String, String> headers = new HashMap<>();
 
     private static int MAX_NEGOTIATE_ATTEMPTS = 100;
 
@@ -163,12 +166,20 @@ public class HubConnection {
                 this.accessToken = (this.negotiateResponse == null) ? null : this.negotiateResponse.getAccessToken();
                 this.negotiateResponse = Negotiate.processNegotiate(url, this.accessToken);
 
-                if (this.accessToken != null) {
-                    this.url = this.url + "&id=" + negotiateResponse.getConnectionId() + "&access_token=" + this.accessToken;
+                if (this.negotiateResponse.getConnectionId() != null) {
+                    if (this.url.contains("?")) {
+                        this.url = this.url + "&id=" + negotiateResponse.getConnectionId();
+                    } else {
+                        this.url = this.url + "?id=" + negotiateResponse.getConnectionId();
+                    }
+                }
+
+                if (negotiateResponse.getAccessToken() != null) {
+                    this.headers.put("Authorization", "Bearer " + negotiateResponse.getAccessToken());
                 }
 
                 if (this.negotiateResponse.getRedirectUrl() != null) {
-                        this.url = this.negotiateResponse.getRedirectUrl();
+                    this.url = this.negotiateResponse.getRedirectUrl();
                 }
 
                 negotiateAttempts++;
@@ -180,8 +191,9 @@ public class HubConnection {
 
         logger.log(LogLevel.Debug, "Starting HubConnection");
         if (transport == null) {
-            transport = new WebSocketTransport(this.url, this.logger);
+            transport = new WebSocketTransport(this.url, this.logger, this.headers);
         }
+
         transport.setOnReceive(this.callback);
         transport.start();
         String handshake = HandshakeProtocol.createHandshakeRequestMessage(new HandshakeRequestMessage(protocol.getName(), protocol.getVersion()));
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -28,7 +28,7 @@ public class HubConnection {
 
     private static int MAX_NEGOTIATE_ATTEMPTS = 100;
 
-    public HubConnection(String url, Transport transport, Logger logger, boolean skipNegotiate){
+    public HubConnection(String url, Transport transport, Logger logger, boolean skipNegotiate) {
         this.url = url;
         this.protocol = new JsonHubProtocol();
 
@@ -61,16 +61,16 @@ public class HubConnection {
             HubMessage[] messages = protocol.parseMessages(payload);
 
             for (HubMessage message : messages) {
-                logger.log(LogLevel.Debug,"Received message of type %s", message.getMessageType());
+                logger.log(LogLevel.Debug, "Received message of type %s", message.getMessageType());
                 switch (message.getMessageType()) {
                     case INVOCATION:
-                        InvocationMessage invocationMessage = (InvocationMessage)message;
+                        InvocationMessage invocationMessage = (InvocationMessage) message;
                         if (handlers.containsKey(invocationMessage.target)) {
-                            ArrayList<Object> args = gson.fromJson((JsonArray)invocationMessage.arguments[0], (new ArrayList<>()).getClass());
+                            ArrayList<Object> args = gson.fromJson((JsonArray) invocationMessage.arguments[0], (new ArrayList<>()).getClass());
                             List<ActionBase> actions = handlers.get(invocationMessage.target);
                             if (actions != null) {
                                 logger.log(LogLevel.Debug, "Invoking handlers for target %s", invocationMessage.target);
-                                for (ActionBase action: actions) {
+                                for (ActionBase action : actions) {
                                     action.invoke(args.toArray());
                                 }
                             }
@@ -80,7 +80,7 @@ public class HubConnection {
                         break;
                     case CLOSE:
                         logger.log(LogLevel.Information, "Close message received from server.");
-                        CloseMessage closeMessage = (CloseMessage)message;
+                        CloseMessage closeMessage = (CloseMessage) message;
                         stop(closeMessage.getError());
                         break;
                     case PING:
@@ -97,22 +97,23 @@ public class HubConnection {
             }
         };
 
-        if (transport != null){
+        if (transport != null) {
             this.transport = transport;
         }
     }
 
-    public HubConnection(String url, Transport transport, Logger logger){
+    public HubConnection(String url, Transport transport, Logger logger) {
         this(url, transport, logger, false);
     }
 
-    public HubConnection(String url, Transport transport, boolean skipNegotiate){
+    public HubConnection(String url, Transport transport, boolean skipNegotiate) {
         this(url, transport, new NullLogger(), skipNegotiate);
     }
 
     /**
      * Initializes a new instance of the {@link HubConnection} class.
-     * @param url The url of the SignalR server to connect to.
+     *
+     * @param url       The url of the SignalR server to connect to.
      * @param transport The {@link Transport} that the client will use to communicate with the server.
      */
     public HubConnection(String url, Transport transport) {
@@ -121,6 +122,7 @@ public class HubConnection {
 
     /**
      * Initializes a new instance of the {@link HubConnection} class.
+     *
      * @param url The url of the SignalR server to connect to.
      */
     public HubConnection(String url) {
@@ -129,15 +131,17 @@ public class HubConnection {
 
     /**
      * Initializes a new instance of the {@link HubConnection} class.
-     * @param url The url of the SignalR server to connect to.
+     *
+     * @param url      The url of the SignalR server to connect to.
      * @param logLevel The minimum level of messages to log.
      */
-    public HubConnection(String url, LogLevel logLevel){
+    public HubConnection(String url, LogLevel logLevel) {
         this(url, null, new ConsoleLogger(logLevel));
     }
 
     /**
      * Indicates the state of the {@link HubConnection} to the server.
+     *
      * @return HubConnection state enum.
      */
     public HubConnectionState getConnectionState() {
@@ -146,6 +150,7 @@ public class HubConnection {
 
     /**
      * Starts a connection to the server.
+     *
      * @throws Exception An error occurred while connecting.
      */
     public void start() throws Exception {
@@ -155,14 +160,13 @@ public class HubConnection {
         if (!skipNegotiate) {
             int negotiateAttempts = 0;
             do {
-                if(negotiateAttempts > 0 && this.negotiateResponse.accessToken != null){
+                if (negotiateAttempts > 0 && this.negotiateResponse.accessToken != null) {
                     this.accessToken = negotiateResponse.accessToken;
                     this.negotiateResponse = Negotiate.processNegotiate(url, this.negotiateResponse.accessToken);
                     this.url = this.url + "&id=" + negotiateResponse.connectionId + "&access_token=" + this.accessToken;
-                }
-                else {
+                } else {
                     this.negotiateResponse = Negotiate.processNegotiate(url);
-                    if (this.negotiateResponse.redirectUrl != null){
+                    if (this.negotiateResponse.redirectUrl != null) {
                         this.url = this.negotiateResponse.redirectUrl;
                     }
                 }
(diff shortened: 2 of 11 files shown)


clients/java/signalr/build.gradle
```diff
@@ -18,6 +18,7 @@ dependencies {
     testImplementation group: 'junit', name: 'junit', version: '4.12'
     implementation "org.java-websocket:Java-WebSocket:1.3.8"
     implementation 'com.google.code.gson:gson:2.8.5'
+    compile 'org.apache.httpcomponents:httpclient:4.5.6'
```
mikaelm12: New dependency for an implementation of an HttpClient

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -22,8 +22,13 @@
     private HubConnectionState connectionState = HubConnectionState.DISCONNECTED;
     private Logger logger;
     private List<Consumer<Exception>> onClosedCallbackList;
+    private boolean skipNegotiate = true;
+    private NegotiateResponse negotiateResponse;
+    private String accessToken;
 
-    public HubConnection(String url, Transport transport, Logger logger){
+    private static int MAX_NEGOTIATE_ATTEMPTS = 5;
```
mikaelm12: Just arbitrarily put 5 for the max. I think it's much higher for TS

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -145,8 +148,32 @@ public void start() throws Exception {
         if (connectionState != HubConnectionState.DISCONNECTED) {
             return;
         }
+        if (!skipNegotiate) {
+            int negotiateAttempts = 0;
+            do {
+                if(negotiateAttempts > 0 && this.negotiateResponse.accessToken != null){
+                    this.accessToken = negotiateResponse.accessToken;
+                    this.negotiateResponse = Negotiate.processNegotiate(url, this.negotiateResponse.accessToken);
+                    this.url = this.url + "&id=" + negotiateResponse.connectionId + "&access_token=" + this.accessToken;
```
mikaelm12: Should I just store the `accessToken`  in the `Authorization` header instead?

BrennanConroy: Yes

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,47 @@
...
+        HttpClient client = HttpClientBuilder.create().build();
+        HttpPost post = new HttpPost(url + "/negotiate");
+        HttpResponse response = client.execute(post);
+        String result = getResponseResult(response);
+        return new NegotiateResponse(result);
+    }
+
+    public static NegotiateResponse processNegotiate(String url, String accessTokenHeader) throws IOException {
+        HttpClient client = HttpClientBuilder.create().build();
+        url = url.substring(0, url.indexOf('?')) + "negotiate" + url.substring(url.indexOf('?'));
```
BrennanConroy: This assumes the url ends with '/'

BrennanConroy: Is there some library code you can use instead of trying to correctly parse urls?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class Negotiate {
+
+    public static NegotiateResponse processNegotiate(String url) throws IOException {
+        HttpClient client = HttpClientBuilder.create().build();
+        HttpPost post = new HttpPost(url + "/negotiate");
```
BrennanConroy: This assumes there is no query string present

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
```diff
@@ -0,0 +1,39 @@
...
+    public Set<String> availableTransports = new HashSet<>();
+    public boolean shouldRedirect;
+    public String redirectUrl;
+    public String accessToken;
+
+
+    public NegotiateResponse(String negotiatePayload) {
+        JsonObject negotiateResponse= jsonParser.parse(negotiatePayload).getAsJsonObject();
+        if (negotiateResponse.has("url")) {
+            this.shouldRedirect = true;
```
BrennanConroy: unneeded, just check if the `redirectUrl` is non-empty

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
```diff
@@ -0,0 +1,39 @@
...
+            this.redirectUrl = negotiateResponse.get("url").getAsString();
+            if (negotiateResponse.has("accessToken")) {
+                this.accessToken = negotiateResponse.get("accessToken").getAsString();
+            }
+            return;
+        }
+        this.connectionId = negotiateResponse.get("connectionId").getAsString();
+        negotiateResponse.get("availableTransports");
```
BrennanConroy: dupe code

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -13,15 +16,20 @@
...
     private static final String HTTPS = "https";
     private static final String WS = "ws";
     private static final String WSS = "wss";
+    private static final String ACCESS_TOKEN_PREFIX = "access_token=";
 
     public WebSocketTransport(String url, Logger logger) throws URISyntaxException {
         this.url = formatUrl(url);
         this.logger = logger;
+        if (url.indexOf(ACCESS_TOKEN_PREFIX) > 0){
+            this.accessToken = url.substring(url.indexOf(ACCESS_TOKEN_PREFIX) + ACCESS_TOKEN_PREFIX.length());
```
BrennanConroy: What is this for?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
mikaelm12: This was debugging

clients/java/signalr/build.gradle
mikaelm12: This has been approved 👍

anurse: Looks like [`compile` has been deprecated](https://docs.gradle.org/current/userguide/java_plugin.html#tab:configurations) and `implementation` should be used instead (like the other dependencies)

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: It's 100 in .NET: https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs#L28

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,47 @@
...
+        String result = getResponseResult(response);
+        return new NegotiateResponse(result);
+    }
+
+    public static NegotiateResponse processNegotiate(String url, String accessTokenHeader) throws IOException {
+        HttpClient client = HttpClientBuilder.create().build();
+        url = url.substring(0, url.indexOf('?')) + "negotiate" + url.substring(url.indexOf('?'));
+        HttpPost post = new HttpPost(url);
+        post.setHeader("Authorization", "Bearer " + accessTokenHeader);
+        HttpResponse response = client.execute(post);
```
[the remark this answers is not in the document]
mikaelm12: It looks like there's an Async flavor of this HttpClient https://hc.apache.org/httpcomponents-asyncclient-ga/httpasyncclient/apidocs/org/apache/http/nio/client/HttpAsyncClient.html
I can update the Async issue to note that we should be using this once we start using `CompletableFuture`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: This will just keep appending `&id` and `&access_token` as the client keeps getting redirected. I'm not sure the logic needs to be different based on which attempt we're on. The loop [in the C# client
](https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs#L314) doesn't really do that. The initial URL and initial Access Token aren't relevant at all once the negotiate response comes in (https://github.com/aspnet/SignalR/issues/2859 notwithstanding, but we should come back to that)

mikaelm12:
So I was testing this on Android and it turns out that the Apache HttpClient library isn't supported.
https://developer.android.com/about/versions/marshmallow/android-6.0-changes#behavior-apache-http-client

I'm looking into using Squares [OkHttp ](https://github.com/square/okhttp)Library as a replacement.

davidfowl:
OkHttp is the pretty common android replacement. Grpc uses it https://github.com/grpc/grpc-java/tree/master/okhttp

mikaelm12:
>OkHttp is the pretty common android replacement.

Yeah, should have just started there. Hindsight is 20/20 I guess.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,77 @@
...
+        StringBuffer result = new StringBuffer();
+        String line = "";
+        while ((line = rd.readLine()) != null) {
+            result.append(line);
+        }
+
+        return result.toString();
+    }
+
+    private static String resolveNegotiateUrl(String url){
```
davidfowl: nit: the formatting in this method

clients/java/signalr/build.gradle
```diff
@@ -19,6 +19,7 @@ dependencies {
     implementation "org.java-websocket:Java-WebSocket:1.3.8"
     implementation 'com.google.code.gson:gson:2.8.5'
     implementation 'org.apache.httpcomponents:httpclient:4.5.6'
+    implementation 'com.squareup.okhttp3:okhttp:3.11.0'
```
davidfowl: We need to do an OSS request 

cc @Eilon

mikaelm12: Already sent an email before updating the PR

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,64 @@
...
+        String result = response.body().string();
+        return new NegotiateResponse(result);
+    }
+
+    private static String resolveNegotiateUrl(String url){
+        String negotiateUrl = "";
+
+        // Check if we have a query string. If we do then we ignore it for now.
+        int queryStringIndex = url.indexOf('?');
+        if(queryStringIndex > 0){
```
davidfowl: you have what I call @anurse virus. The `if(` always touching 😄

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
```diff
@@ -0,0 +1,35 @@
...
+        JsonObject negotiateResponse= jsonParser.parse(negotiatePayload).getAsJsonObject();
+        if (negotiateResponse.has("url")) {
+            this.redirectUrl = negotiateResponse.get("url").getAsString();
+            if (negotiateResponse.has("accessToken")) {
+                this.accessToken = negotiateResponse.get("accessToken").getAsString();
+            }
+            return;
+        }
+        this.connectionId = negotiateResponse.get("connectionId").getAsString();
+        JsonArray transports = ((JsonArray)negotiateResponse.get("availableTransports"));
```
davidfowl: nit: Get rid of the extra parenthesis

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
davidfowl: This looks like it should be removed.

mikaelm12:
Okay, this now works on Android and I no longer store the accessToken in the url. Moved to headers. And the formatting should be good too

mikaelm12:
Updated

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -13,35 +16,47 @@
...
     private URI formatUrl(String url) throws URISyntaxException {
-        if(url.startsWith(HTTPS)){
+        if (url.startsWith(HTTPS)) {
             url = WSS + url.substring(HTTPS.length());
-        }
-        else if(url.startsWith(HTTP)){
+        } else if (url.startsWith(HTTP)) {
             url = WS + url.substring(HTTP.length());
         }
+        if (url.indexOf(ACCESS_TOKEN_PREFIX) > 0) {
```
davidfowl: Why does this exist?

mikaelm12: We parse the access token from the url so we can put it in a header when we construct the underlying WebSocketClient. We actually know the accessToken before we even contruct the transport. I can add a ctor that takes headers and get rid of this

davidfowl: That would be much cleaner.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -92,20 +100,23 @@ public HubConnection(String url, Transport transport, Logger logger){
...
-            } catch (URISyntaxException e) {
-                e.printStackTrace();
-            }
-        } else {
+        if (transport != null) {
             this.transport = transport;
         }
     }
 
+    public HubConnection(String url, Transport transport, Logger logger) {
```
davidfowl: Why did you add 2 constructors?

mikaelm12: For testing. Once I add the skipNegotiate option to the builder I can clean this up

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,63 @@
...
+public class Negotiate {
+
+    public static NegotiateResponse processNegotiate(String url) throws IOException {
+        return processNegotiate(url, null);
+    }
+
+    public static NegotiateResponse processNegotiate(String url, String accessTokenHeader) throws IOException {
+        url = resolveNegotiateUrl(url);
+        OkHttpClient client = new OkHttpClient();
+        RequestBody body = RequestBody.create(null, new byte[]{});
```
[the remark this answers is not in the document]
mikaelm12: >What are the create arguments?

`create(MediaType contentType, byte[] content)`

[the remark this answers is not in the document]
mikaelm12: >Why aren’t we copying all of the headers here?

This seems to be the way we do this in the TS client

mikaelm12:
Ping

clients/java/signalr/src/test/java/NegotiateResponseTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.microsoft.aspnet.signalr.NegotiateResponse;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class NegotiateResponseTest {
+
+    @Test
+    public void VerifyNegoitateResponse() {
```
BrennanConroy: nit: Negotiate

clients/java/signalr/src/test/java/NegotiateResponseTest.java
```diff
@@ -0,0 +1,38 @@
...
+        assertTrue(negotiateResponse.getAvailableTransports().contains("WebSockets"));
+        assertTrue(negotiateResponse.getAvailableTransports().contains("ServerSentEvents"));
+        assertTrue(negotiateResponse.getAvailableTransports().contains("LongPolling"));
+        assertNull(negotiateResponse.getAccessToken());
+        assertNull(negotiateResponse.getRedirectUrl());
+        assertEquals("bVOiRPG8-6YiJ6d7ZcTOVQ", negotiateResponse.getConnectionId());
+    }
+
+    @Test
+    public void VerifyRedirectNegoitateResponse() {
```
BrennanConroy: nit: Negotiate

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -420,18 +477,23 @@ public Subscription on(String target, Action callback) {
...
     public void onClosed(Consumer<Exception> callback) {
-        if (onClosedCallbackList == null){
+        if (onClosedCallbackList == null) {
             onClosedCallbackList = new ArrayList<>();
         }
 
         onClosedCallbackList.add(callback);
     }
+
+    public void setSkipNegotiate(boolean skipNegotiate) {
```
BrennanConroy: Remove

mikaelm12:
> There are no tests that check the redirect url behavior.

I wanted to add this but this requires adding infrastructure to allow passing in custom HttpClients for testing. That should happen eventually but not in this PR imo. Trying to stop this PR from growing any more

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -72,31 +82,31 @@ public void stop() {
...
-                     onReceive(message);
-                 } catch (Exception e) {
-                     e.printStackTrace();
-                 }
-             }
-
-             @Override
-             public void onClose(int code, String reason, boolean remote) {
+    private WebSocketClient createWebSocket(Map<String, String> headers) {
+        return new WebSocketClient(this.url, headers) {
```
BrennanConroy: Why are you now using `this` when accessing class members?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        Request request = requestBuilder.build();
+
+        Response response = client.newCall(request).execute();
+        String result = response.body().string();
+        return new NegotiateResponse(result);
+    }
+
+    private static String resolveNegotiateUrl(String url) {
```
BrennanConroy: Add some tests for this please. See https://github.com/aspnet/SignalR/blob/release/2.2/clients/ts/signalr/tests/Common.ts#L17 and https://github.com/aspnet/SignalR/blob/release/2.2/clients/ts/signalr/tests/HttpConnection.test.ts#L220 for some reference

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -139,14 +153,47 @@ public HubConnectionState getConnectionState() {
...
      * @throws Exception An error occurred while connecting.
      */
     public void start() throws Exception {
         if (connectionState != HubConnectionState.DISCONNECTED) {
             return;
         }
+        if (!skipNegotiate) {
+            int negotiateAttempts = 0;
+            do {
+                this.accessToken = (this.negotiateResponse == null) ? null : this.negotiateResponse.getAccessToken();
```
BrennanConroy: this this this

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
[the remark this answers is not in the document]
anurse: It's a little hidden, but we do copy all the headers in the JavaScript client:

https://github.com/aspnet/SignalR/blob/201c058b9379b584317b912a4928a83ad7b569bf/clients/ts/signalr/src/HttpClient.ts#L184-L190

anurse:
Formatting changes made a bunch of noise in HubConnection unfortunately :(

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
mikaelm12: Good point. 👍

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
mikaelm12: I actually want to move towards using `this`. Since it's not the convention to use an underscore prefix I've started liking the idea of using `this`

mikaelm12:
Final reviews?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Negotiate.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        Request request = requestBuilder.build();
+
+        Response response = client.newCall(request).execute();
+        String result = response.body().string();
+        return new NegotiateResponse(result);
+    }
+
+    public static String resolveNegotiateUrl(String url) {
```
BrennanConroy: We'll need to look at the public API later

PR #2888, opened 2018-08-29
Use gradle to detect java

anurse:
Fixes https://github.com/aspnet/Coherence-Signed/issues/1004


anurse:
Weird. I rebuilt it: https://dotnet.visualstudio.com/public/public%20Team/_build/results?buildId=16505 and it didn't update the GitHub status.

PR #2890, opened 2018-08-29
fix chrome headless lag on Win7

anurse:
Apparently, adding these switches makes Chrome Headless not super slow... I have no idea why but [the internet](https://github.com/Codeception/CodeceptJS/issues/561#issuecomment-373666779) said so.

Fixes https://github.com/aspnet/Coherence-Signed/issues/916


anurse:
Merging to unblock build.

PR #2902, opened 2018-08-31
Add Spotless google-java check to build

mikaelm12:
This adds Spotless configured to use the google-java format. 
A couple issues with this. The line length limit messes with the file head and it's configured to use 2 space indenting instead of 4.
Personally I'd rather write a custom set of rules. If people would rather that than I'll close this PR in favor of that.
Issue: https://github.com/aspnet/SignalR/issues/2676


mikaelm12:
Closing this in favor of https://github.com/aspnet/SignalR/pull/2912

Issue #2893, opened 2018-08-30
"can't find variable: ErrorEvent" when running in React-Native


Kimmen:
We are using Signalr in our React-Native app for pushing data through websockets. Sometimes we get a websocket error (perhaps if you run start twice, haven't looked at it). But when that happens we get a crash stating "**can't find variable: ErrorEvent**", I've attached a screenshot (See below)

The code that emits this error can be found in WebSocketTransport.ts on line 60: https://github.com/aspnet/SignalR/blob/f91ba389077b43fcab85a9a31e5678229093295a/clients/ts/signalr/src/WebSocketTransport.ts#L60

This might correlate to [JavaScriptCore runtime](https://trac.webkit.org/wiki/JavaScriptCore) that is used in React-Native. Maybe effects Safari as well?

* Versions of Server-Side NuGet Packages: 1.0.0-preview1-10015
* Versions of Client-Side NuGet/NPM Packages: 1.1.0-preview1-35029
* Are you using the C# client or the JavaScript client: Javascript
* The Server you are using (Kestrel/HttpSysServer/IIS/IIS Express/Azure Web App/etc.): Azure Web App
* The Operating System on the Server (Windows/Linux/macOS): Windows
* The Operating System on the Client (Windows/Linux/macOS): Android
* The Browser on the client, if using the JavaScript client (IE/Chrome/Edge/Firefox/etc.): React-Native / JavaScriptCore runtime according to [React-Natives documentation](https://facebook.github.io/react-native/docs/javascript-environment.html).

![error-event](https://user-images.githubusercontent.com/1217354/44834466-94863b00-ac31-11e8-8518-6c7effd117cd.png)

Issue #2535, opened 2018-06-25
Add NodeJS support to client


anurse:
* Take dependencies on WebSocket and SSE polyfills
* Build HttpClient based on Node `http` module
* Auto-detect node and use the polyfill and Node HttpClient.

BrennanConroy:
Done https://github.com/aspnet/SignalR/commit/a51eec7a1d76272721da7437a273b3c6c74c581a

PR #2910, opened 2018-09-04
Add NodeJS in-box support

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2535

Also fixes https://github.com/aspnet/SignalR/issues/2893


BrennanConroy:
@Eilon Adding a couple new dependencies https://github.com/aspnet/SignalR/pull/2910/files#diff-8c7936f2b30b4097f0fac9abf544b282R43

clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -59,11 +60,21 @@ export class HttpConnection implements IConnection {
         options = options || {};
         options.logMessageContent = options.logMessageContent || false;
 
+        // Access module with require to allow browser builds to remove websocket dependency
```
davidfowl: I dunno what this means, but i'll ask tomorrow

clients/ts/signalr/src/Polyfills.ts
```diff
@@ -14,3 +20,110 @@ export interface WebSocketConstructor {
     readonly CONNECTING: number;
     readonly OPEN: number;
 }
+
+export class WebSocketWrapper implements WebSocket {
```
davidfowl: Add a comment why this is needed.

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -57,7 +57,10 @@ export class WebSocketTransport implements ITransport {
             };
 
             webSocket.onerror = (event: Event) => {
-                const error = (event instanceof ErrorEvent) ? event.error : null;
+                let error: any = null;
+                if (typeof ErrorEvent !== "undefined" && event instanceof ErrorEvent) {
```
davidfowl: Why?

clients/ts/signalr/src/Utils.ts
```diff
@@ -68,9 +68,11 @@ export async function sendMessage(logger: ILogger, transportName: string, httpCl
 
     logger.log(LogLevel.Trace, `(${transportName} transport) sending data. ${getDataDetail(content, logMessageContent)}.`);
 
+    const responseType = content instanceof ArrayBuffer ? "arraybuffer" : "text";
```
davidfowl: ArrayBuffer or Buffer?

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -229,4 +239,78 @@ export class DefaultHttpClient extends HttpClient {
...
+                    dataLength += chunk.length;
+                });
+
+                res.on("end", () => {
+                    if (request.abortSignal) {
+                        request.abortSignal.onabort = null;
+                    }
+
+                    if (res.statusCode && res.statusCode >= 200 && res.statusCode < 300) {
+                        let resp: string | ArrayBuffer;
```
davidfowl: Never a Buffer here?

clients/ts/FunctionalTests/func.jest.config.js
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+module.exports = {
+    transformIgnorePatterns: [
+        ".*/node_modules/(?!@aspnet)"
```
anurse: Since this is a JS file, not a JSON file, can you put a comment in? Something like "We reference the ESM output directly in our tests and need it to run through the TS compiler in order to work in NodeJS"

clients/ts/FunctionalTests/ts/Common.ts
```diff
@@ -22,19 +22,28 @@ if ((window as any).__karma__) {
     // Running in Karma? Need to use an absolute URL
     ENDPOINT_BASE_URL = server;
     console.log(`Using SignalR Server: ${ENDPOINT_BASE_URL}`);
-} else {
+} else if (typeof document !== "undefined") {
     ENDPOINT_BASE_URL = `${document.location.protocol}//${document.location.host}`;
+} else if (process.env.SERVER_URL) {
```
anurse: let's be paranoid: `else if (process && process.env && process.env.SERVER_URL)`

clients/ts/FunctionalTests/ts/Common.ts
```diff
@@ -51,7 +60,7 @@ export function eachTransportAndProtocol(action: (transport: HttpTransportType,
     const protocols: IHubProtocol[] = [new JsonHubProtocol()];
     // IE9 does not support XmlHttpRequest advanced features so disable for now
```
anurse: Clarify this comment: Only run messagepack tests if in Node, or if the browser supports binary content (indicated by the `responseType` property existing)

clients/ts/FunctionalTests/ts/Common.ts
```diff
@@ -65,3 +74,7 @@ export function eachTransportAndProtocol(action: (transport: HttpTransportType,
         });
     });
 }
+
+export function windowOrglobal(): any {
```
anurse: Suggested rename: `getGlobalObject()`.

clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -606,8 +606,8 @@ describe("hubConnection", () => {
     it("transport falls back from WebSockets to SSE or LongPolling", async (done) => {
```
anurse: This function is not relevant in node, you can just disable it. And then you can use `window` directly.

clients/ts/FunctionalTests/ts/WebSocketTests.ts
```diff
@@ -1,33 +1,37 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+import { WebSocketWrapper } from "@aspnet/signalr/dist/esm/Polyfills";
```
anurse: Remove this and use `require` below (see comments below). We don't ever want to run `require('w3cwebsocket')` in  a browser environment.

clients/ts/FunctionalTests/ts/WebSocketTests.ts
```diff
@@ -1,33 +1,37 @@
...
+describe("WebSockets", () => {
+    it("can be used to connect to SignalR", (done) => {
+        const message = "message";
 
-            const webSocket = new WebSocket(ECHOENDPOINT_URL.replace(/^http/, "ws"));
+        let webSocket: WebSocket;
+        if (typeof WebSocket !== "undefined") {
+            webSocket = new WebSocket(ECHOENDPOINT_URL.replace(/^http/, "ws"));
+        } else {
+            webSocket = new WebSocketWrapper(ECHOENDPOINT_URL.replace(/^http/, "ws"));
```
anurse: Use `require` to import the module **inside** the `if` condition that guarantees we're in node.

clients/ts/FunctionalTests/ts/WebSocketTests.ts
```diff
@@ -1,33 +1,37 @@
...
-    describe("WebSockets", () => {
-        it("can be used to connect to SignalR", (done) => {
-            const message = "message";
+describe("WebSockets", () => {
+    it("can be used to connect to SignalR", (done) => {
+        const message = "message";
 
-            const webSocket = new WebSocket(ECHOENDPOINT_URL.replace(/^http/, "ws"));
+        let webSocket: WebSocket;
+        if (typeof WebSocket !== "undefined") {
```
anurse: We should gracefully handle the case of running the tests in a browser that doesn't have WebSockets. Right now that will fall through this condition to the else and you'll have a Bad Problem(TM).

Change this condition to `if(window)` and then test `typeof WebSocket` inside the if and if it's undefined just no-op the test.

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+import * as HTTP from "http";
```
anurse: nit: lowercase `http` for the bound variable too.

clients/ts/signalr/package.json
```diff
@@ -36,6 +36,12 @@
     "src/**/*"
   ],
   "devDependencies": {
-    "es6-promise": "^4.2.2"
+    "es6-promise": "^4.2.2",
+    "@types/websocket": "^0.0.40",
+    "@types/eventsource": "^1.0.2"
+  },
+  "dependencies": {
```
Eilon: @BrennanConroy do these become something that we re-ship embedded in our own JS file? Or just used at build time, and customers would get them via NPM?

anurse: These are runtime dependencies acquired by the **user** when they install our package. We do not embed them in any of our JavaScript files.

Eilon: OK cool then I'll wait for @anurse 's OSS integration thingy to add these to our OSS list. No immediate action required here.

clients/ts/signalr/package.json
```diff
@@ -36,6 +36,12 @@
     "src/**/*"
   ],
   "devDependencies": {
-    "es6-promise": "^4.2.2"
+    "es6-promise": "^4.2.2",
+    "@types/websocket": "^0.0.40",
+    "@types/eventsource": "^1.0.2"
```
BrennanConroy: Since we use require, I might not need the @types anymore

BrennanConroy:
Ping

Kamikadze4GAME:
Hi all.
When will this changes on npmjs? And some examples or docs?

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -57,7 +57,11 @@ export class WebSocketTransport implements ITransport {
             };
 
             webSocket.onerror = (event: Event) => {
-                const error = (event instanceof ErrorEvent) ? event.error : null;
+                let error: any = null;
+                // ErrorEvent is a browser only type we need to check if the type exists before using it
+                if (typeof ErrorEvent !== "undefined" && event instanceof ErrorEvent) {
+                    error = event.error;
+                }
                 reject(error);
```
anurse: This will just be `null` in Node right? That seems non-ideal. Is there no error parameter to this callback in Node?

clients/ts/signalr/webpack.config.js
```diff
@@ -2,4 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 const baseConfig = require("../webpack.config.base");
-module.exports = baseConfig(__dirname, "signalr");
\ No newline at end of file
+module.exports = baseConfig(__dirname, "signalr", {
+    // These are only used in Node environments
```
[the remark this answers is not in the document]
anurse: Ok. I'd like to review the generated WebPack bundle before you merge, on Monday.

clients/ts/signalr/src/WebSocketTransport.ts
BrennanConroy: The value of `event` in node:
`{"AT_TARGET": 0, "BUBBLING_PHASE": 0, "CAPTURING_PHASE": 0, "NONE": 0, "bubbles": false, "cancelBubble": false,"cancelable": false, "defaultPrevented": false, "eventPhase": 0, "isTrusted": false, "returnValue": false, "scoped": false, "timeStamp": 0, "type": ""}`

clients/ts/signalr/package.json
BrennanConroy: @Eilon ?

Eilon: We need a 3PN in any case where we redistribute 3rd party code. This can be either because we've checked in 3rd party code into a GitHub repo we own (because we own the repo, we are the ones redist'ing the code), or this can be because we produce a package (e.g. NUPKG, NPM, etc.) that includes 3rd party code (e.g. our template NUPKGs contain jQuery).

If the user acquires the 3rd party code via another mechanism (e.g. directly from NuGet.org via a dependency), then 3PN is not needed.

anurse: Cool, thanks for the clarification! No 3pn needed here then, we don't embed anything new.

clients/ts/signalr/src/WebSocketTransport.ts
anurse: :( ok then

BrennanConroy:
>So just so I'm clear: the only substantial new content is the NodeHttpClient (i.e. the polyfills just work as-is) and it isn't present at all in the browser build, right?

Ouch, belittling my work! Yeah, browser builds have the polyfills + NodeHttpClient stripped.

PR #2912, opened 2018-09-04
Spotless with custom formatting rules for the Java client

mikaelm12:
I created a custom rule set that covers our conventions. I like this more than the google java formatter because of their apparently intentionally unchangeable line length limit's and their 2 space indentations.

Issue: https://github.com/aspnet/SignalR/issues/2676

### DIFF
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -24,12 +24,15 @@ dependencies {
 spotless {
     java {
         licenseHeader '// Copyright (c) .NET Foundation. All rights reserved.\n' +
-        '// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.\n' // License header
+        '// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.\n\n' // License header
 
         importOrder 'java', 'javax', 'org', 'com', 'com.diffplug', ''	// A sequence of package names
 
         replace      'Not enough space after if', 'if(', 'if ('
         replace      'Not enough space after )', '){', ') {'
+        replace      'Not enough space after for', 'for(', 'for ('
+        replace      'Not enough space after while', 'while (', 'while ('
+        replace      'Not enough space after switch', 'switch(', 'switch ('
         replaceRegex 'Too much space after if', 'if +\\(', 'if ('
         trimTrailingWhitespace()
         indentWithSpaces(4)
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -1,6 +1,7 @@
 plugins {
     id 'java'
     id 'maven'
+    id "com.diffplug.gradle.spotless" version "3.14.0"
 }
 
 group 'com.microsoft.aspnet'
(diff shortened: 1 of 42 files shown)


clients/java/signalr/build.gradle
```diff
@@ -20,6 +21,22 @@ dependencies {
...
 }
 
+spotless {
+    java {
+        licenseHeader '// Copyright (c) .NET Foundation. All rights reserved.\n' +
+        '// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.\n' // License header
+
+        importOrder 'java', 'javax', 'org', 'com', 'com.diffplug', ''	// A sequence of package names
+
+        replace      'Not enough space after if', 'if(', 'if ('
```
BrennanConroy: for loops and while loops too

clients/java/signalr/build.gradle
```diff
@@ -20,6 +21,22 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.8.5'
 }
 
+spotless {
+    java {
+        licenseHeader '// Copyright (c) .NET Foundation. All rights reserved.\n' +
+        '// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.\n' // License header
```
BrennanConroy: Add an extra newline

clients/java/signalr/build.gradle
```diff
@@ -20,6 +21,22 @@ dependencies {
...
+
+        importOrder 'java', 'javax', 'org', 'com', 'com.diffplug', ''	// A sequence of package names
+
+        replace      'Not enough space after if', 'if(', 'if ('
+        replace      'Not enough space after )', '){', ') {'
+        replaceRegex 'Too much space after if', 'if +\\(', 'if ('
+        trimTrailingWhitespace()
```
BrennanConroy: Couple rules that would be nice but might be difficult to do:
* no spaces before a comma (not 100% sure about this one)
* single space around all the comparison operators (sounds hard)

clients/java/signalr/build.gradle
anurse: This is my concern about custom rules... 😟

clients/java/signalr/build.gradle
anurse: It's disappointing that Spotless doesn't have a way to configure specific rulesets... I'd hoped we could still use standard rules and just turn some of them off.

clients/java/signalr/build.gradle
BrennanConroy: And `switch`...

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -163,7 +164,7 @@ private void stop(String errorMessage) {
             return;
         }
 
-        if(errorMessage != null) {
+        if (errorMessage != null) {
             logger.log(LogLevel.Error , "HubConnection disconnected with an error %s.", errorMessage);
```
BrennanConroy: Here is an example of the `,` problem

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -24,15 +25,15 @@ public WebSocketTransport(String url, Logger logger) throws URISyntaxException {
...
         return url;
     }
 
     private URI formatUrl(String url) throws URISyntaxException {
-        if(url.startsWith(HTTPS)){
+        if (url.startsWith(HTTPS)) {
             url = WSS + url.substring(HTTPS.length());
         }
-        else if(url.startsWith(HTTP)){
+        else if (url.startsWith(HTTP)) {
```
BrennanConroy: `else if` and `else` are supposed to be on the same line as `}`

mikaelm12:
Closing in favor of https://github.com/aspnet/SignalR/pull/2924

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
mikaelm12: I'll apply this to the other pr

mikaelm12: https://github.com/aspnet/SignalR/pull/2924

BrennanConroy: :/

PR #2923, opened 2018-09-06
Cancel previous request on close

BrennanConroy:


BrennanConroy:
## Description

There is a race when a LongPolling connection is re-polling and the connection is being closed where there could be a deadlock on the server.

## Customer Impact

LongPolling connections could accidentally deadlock when closing and use server resources.

## Regression from?

Previous 2.1.5 fix for a different issue.

## Risk

High/Medium/Low/Very Low
**Low**:
The race is obvious if you know it is there, and the fix is simple.

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
```diff
@@ -215,8 +215,18 @@ private async Task ExecuteAsync(HttpContext context, ConnectionDelegate connecti
...
                         connection.Cancellation?.Cancel();
 
-                        // Wait for the previous request to drain
-                        await connection.PreviousPollTask;
+                        try
+                        {
+                            // Wait for the previous request to drain
+                            await connection.PreviousPollTask;
+                        }
+                        catch
```
davidfowl: `catch (OperationCancelledException)`

anurse:
A g-g-g-g-g-g-g-ghost!!

BrennanConroy:
I'm doubtful since it's a race and I don't see any obvious way to inject code to force the race to happen when we want it to happen.

Eilon:
Approved for 2.1.5.

PR #2924, opened 2018-09-06
Spotless for the Java Client

mikaelm12:
Updated version of: https://github.com/aspnet/SignalR/pull/2912/files


mikaelm12:
This like the old PR also moves all the tests into a `com.microsoft.aspnet.signalr.test` package

Issue #2917, opened 2018-09-06
Test failure: CanSendAndReceiveUserMessagesFromMultipleConnectionsWithSameUser


aspnet-hello:
This test [fails](http://aspnetci/viewLog.html?buildId=545199&buildTypeId=Releases_22xPublic_UbuntuUniverse) occasionally with the following error:
```
System.TimeoutException : Operation in CanSendAndReceiveUserMessagesFromMultipleConnectionsWithSameUser timed out at /_/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs:118
   at System.Threading.Tasks.TaskExtensions.OrTimeout[T](Task`1 task, TimeSpan timeout, String memberName, String filePath, Nullable`1 lineNumber) in /_/test/Microsoft.AspNetCore.SignalR.Tests.Utils/TaskExtensions.cs:line 67
   at Microsoft.AspNetCore.SignalR.Redis.Tests.RedisEndToEndTests.CanSendAndReceiveUserMessagesFromMultipleConnectionsWithSameUser(HttpTransportType transportType, String protocolName) in /_/test/Microsoft.AspNetCore.SignalR.Redis.Tests/RedisEndToEnd.cs:line 118
--- End of stack trace from previous location where exception was thrown ---
------- Stdout: -------
| [0.005s] TestLifetime Information: Starting test CanSendAndReceiveUserMessagesFromMultipleConnectionsWithSameUser-ServerSentEvents-json at 2018-09-06T02:14:24
| [0.003s] TestLifetime Information: Starting test CanSendAndReceiveUserMessagesFromMultipleConnectionsWithSameUser_ServerSentEvents_json at 2018-09-06T02:14:24
| [0.023s] Microsoft.AspNetCore.Http.Connections.Client.Internal.ServerSentEventsTransport Information: Pj4XujS5DVHaOxCiFcZzcQ - Starting transport. Transfer mode: Text.
| [0.032s] Microsoft.AspNetCore.Http.Connections.Client.HttpConnection Information: Pj4XujS5DVHaOxCiFcZzcQ - HttpConnection Started.
| [0.033s] Microsoft.AspNetCore.SignalR.Client.HubConnection Information: Pj4XujS5DVHaOxCiFcZzcQ - Using HubProtocol 'json v1'.
[...]

PR #2936, opened 2018-09-10
Add more redis logs

BrennanConroy:
Doing this because of https://github.com/aspnet/SignalR/issues/2917

Added one log in production https://github.com/aspnet/SignalR/pull/2936/files#diff-43747ccefdb3a6d796ece10fe906a62eR438 and added the `docker logs redisContainer` command on close to the test to collect any logs from the redis instance.

Also removed a bunch of unnecessary `async`

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
@@ -93,6 +93,9 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Tests
 
         public void Stop(ILogger logger)
         {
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
```diff
@@ -93,6 +93,9 @@ public void Start(ILogger logger)
 
         public void Stop(ILogger logger)
         {
+            // Get logs from Redis container before stopping the container
+            RunProcessAndThrowIfFailed(_path, $"logs {_dockerContainerName}", logger, TimeSpan.FromSeconds(5));
```
davidfowl: Nice!

Issue #2587, opened 2018-07-03
Move De-serialization logic into HubProtocol layer


mikaelm12:
Currently in the Java client the deserialization from JSON happens in the HubConnection layer via
https://github.com/aspnet/SignalR/blob/896c6692b4fdaee16e2fbbb26b86821091c32f71/clients/java/signalr/src/main/java/HubConnection.java#L32

This should be handled via the HubProtocol so that it can be done with different protocols.

BrennanConroy:
https://github.com/aspnet/SignalR/commit/55597adf616757aaec0bc7a611754430b9ef4fd8

Issue #2582, opened 2018-07-03
Java Client De-Serializing Custom Types


mikaelm12:
Currently the Java client doesn't support de-serializing custom types. Gson stores the types as a `LinkedTreeMap` which we can't directly cast to arbitrary types. Need to do some investigation into how to do this in a clean way.

anurse:
I'd expect we can use the same pattern we use in .NET here and have an "invocation binder" interface that flows into the Hub Protocol and allows the protocol to get the `Class` objects representing the destination types.

BrennanConroy:
https://github.com/aspnet/SignalR/commit/55597adf616757aaec0bc7a611754430b9ef4fd8

Issue #2610, opened 2018-07-09
Better Exception Handling in Java Client


mikaelm12:
Currently the exception handling is minimal and general method signatures just declare that general exceptions could be thrown. Work needs to be done to make this more specific and catch these and fail at the right levels of the client. 

Context
https://github.com/aspnet/SignalR/pull/2583#discussion_r200011358
https://github.com/aspnet/SignalR/pull/2583#discussion_r200013217

moozzyk:
Effective Java 70-72

anurse:
Consider wrapping some exceptions so the throws clauses aren't so broad. For example in HubProtocol

```java
HubMessage[] parseMessages(String message, InvocationBinder binder) throws HubProtocolException;
```

or even just

```java
HubMessage[] parseMessages(String message, InvocationBinder binder) throws SignalRException;
```

anurse:
It looks like we should just use `HubException` only for representing errors that come from `invoke` and use `RuntimeException` (and friends) for everything else.

anurse:
Handshake response error -> `RuntimeException`

PR #2940, opened 2018-09-10
[Java] De-Serializing Custom Types

BrennanConroy:
De-Serializing Custom Types #2582
Move De-serialization logic into HubProtocol #2587

Go "easy". First time writing Java.


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/StreamInvocationMessage.java
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+public class StreamInvocationMessage extends HubMessage {
```
mikaelm12: Can this inherit from InvocationMessage instead? Just update the references to `HubMessageType.INVOCATION` with `HubMessageType.STREAM_INVOCATION`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -492,4 +533,26 @@ public void onClosed(Consumer<Exception> callback) {
...
+        @Override
+        public Class<?> GetReturnType(String invocationId) {
+            return null;
+        }
+
+        @Override
+        public List<Class<?>> GetParameterTypes(String methodName) throws Exception {
+            List<InvocationHandler> handlers = connection.handlers.get(methodName);
+            if (handlers == null || handlers.size() == 0) {
+                throw new Exception(String.format("There are no callbacks registered for the method '%s'.", methodName));
```
davidfowl: Do we throw for this in other clients?

BrennanConroy: https://github.com/aspnet/SignalR/blob/adc5706bcb2b3cb85e98e303999eb9a778426f7c/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L1256

davidfowl: That's not the same though right. If there's no method, it logs. If there are no callbacks, it throws.

BrennanConroy: Ok, true

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -61,24 +58,22 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
...
-            HubMessage[] messages = protocol.parseMessages(payload);
+            HubMessage[] messages = protocol.parseMessages(payload, connectionState);
 
             for (HubMessage message : messages) {
-                logger.log(LogLevel.Debug, "Received message of type %s", message.getMessageType());
+                logger.log(LogLevel.Debug, "Received message of type %s.", message.getMessageType());
                 switch (message.getMessageType()) {
                     case INVOCATION:
                         InvocationMessage invocationMessage = (InvocationMessage) message;
                         if (handlers.containsKey(invocationMessage.target)) {
```
anurse: We don't need this check right? [`.get` returns `null` if the key doesn't exist](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#get-java.lang.Object-) so we're just checking twice.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -262,9 +259,9 @@ public void send(String method, Object... args) throws Exception {
      */
     public Subscription on(String target, Action callback) {
         ActionBase action = args -> callback.invoke();
-        handlers.put(target, action);
+        InvocationHandler handler = handlers.put(target, action, new ArrayList<>());
```
anurse: We can cache this `ArrayList` since we never modify it, right?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
```diff
@@ -9,21 +9,23 @@
 import java.util.concurrent.ConcurrentHashMap;
 
 public class CallbackMap {
-    private ConcurrentHashMap<String, List<ActionBase>> handlers = new ConcurrentHashMap<>();
+    private ConcurrentHashMap<String, List<InvocationHandler>> handlers = new ConcurrentHashMap<>();
 
-    public void put(String target, ActionBase action) {
+    public InvocationHandler put(String target, ActionBase action, ArrayList<Class<?>> classes) {
+        InvocationHandler handler = new InvocationHandler(action, classes);
```
anurse: I suggest passing `classes` through [`Collection.unmodifiableList`](https://docs.oracle.com/javase/1.5.0/docs/api/java/util/Collections.html#unmodifiableList(java.util.List)) to make sure we can't accidentally modify it when we didn't expect to be able to.

anurse: Java doesn't seem to have an equivalent for `IReadOnlyList` but `unmodifiableList` is what came up when I googled read only lists ;)

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -298,9 +297,12 @@ public Subscription on(String target, Action callback) {
         ActionBase action = params -> {
             callback.invoke(param1.cast(params[0]), param2.cast(params[1]));
         };
-        handlers.put(target, action);
+        ArrayList<Class<?>> classes = new ArrayList<>();
```
anurse: nit: pass the capacity in so it can preallocate the right array size.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -492,4 +533,26 @@ public void onClosed(Consumer<Exception> callback) {
...
+
+    private class ConnectionState implements InvocationBinder {
+        HubConnection connection;
+
+        public ConnectionState(HubConnection connection) {
+            this.connection = connection;
+        }
+
+        @Override
+        public Class<?> GetReturnType(String invocationId) {
```
anurse: your C# is showing ;P

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -492,4 +533,26 @@ public void onClosed(Consumer<Exception> callback) {
...
+            return null;
+        }
+
+        @Override
+        public List<Class<?>> GetParameterTypes(String methodName) throws Exception {
```
anurse: `camelCaseItBro`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -492,4 +533,26 @@ public void onClosed(Consumer<Exception> callback) {
 
         onClosedCallbackList.add(callback);
     }
+
+    private class ConnectionState implements InvocationBinder {
```
anurse: How about `ConnectionBinder` instead of `ConnectionState` to avoid the name collision?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubProtocol.java
```diff
@@ -16,7 +18,7 @@
      * @param message A string representation of one or more {@link HubMessage}s.
      * @return A list of {@link HubMessage}s.
      */
-    HubMessage[] parseMessages(String message);
+    HubMessage[] parseMessages(String message, InvocationBinder binder) throws Exception;
```
anurse: Checked exceptions...

Eye roll

anurse: I think we should have a wrapper exception type here, but that can wait for The Great Exception Reckoning (#2610)

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/InvocationBinder.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+import java.util.List;
+
+public interface InvocationBinder {
+    Class<?> GetReturnType(String invocationId);
+    List<Class<?>> GetParameterTypes(String methodName) throws Exception;
```
anurse: Camels

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/InvocationHandler.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.ArrayList;
+
+public class InvocationHandler {
+    InvocationHandler(ActionBase action, ArrayList<Class<?>> classes) {
+        this.action = action;
+        this.classes = classes;
+    }
+
+    public ArrayList<Class<?>> classes;
+    public ActionBase action;
```
anurse: Public fields? We should use the same pattern we use in C# (but Javaized) and have private fields with `getClasses`/`setClasses` (as needed), etc.

anurse: We do it for completely private types in C#, so I don't see why we shouldn't in Java.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/StreamInvocationMessage.java
```diff
@@ -0,0 +1,47 @@
...
+
+    int type = HubMessageType.STREAM_INVOCATION.value;
+    String invocationId;
```
anurse: Remember, unlike in C#, the default access for Java is "package" which is more like `internal` than `private`. These are truly `private` so we need to explicitly make them `private`

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/test/HubConnectionTest.java
```diff
@@ -595,6 +619,38 @@ public void SendWithEightParamsTriggersOnHandler() throws Exception {
...
     }
 
+    private class Custom {
+        public int number;
+        public String str;
+        public boolean[] bools;
+    }
+
+    @Test
+    public void SendWithCustomObjectTriggersOnHandler() throws Exception {
```
anurse: Can you just fix all the camel-casing in the tests once and for all? ... it's ... bothering me 😸

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java
```diff
@@ -32,41 +35,105 @@ public TransferFormat getTransferFormat() {
...
+            JsonArray argumentsToken = null;
+
+            JsonReader reader = new JsonReader(new StringReader(str));
+            reader.beginObject();
+
+            do {
+                String name = reader.nextName();
+                switch (name) {
+                    case "type":
+                        messageType = HubMessageType.values()[reader.nextInt() - 1];
```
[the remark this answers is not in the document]
anurse: Turns out [this is the "right way" to do this](https://stackoverflow.com/a/5878986/29813) :(.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java
```diff
@@ -32,41 +35,105 @@ public TransferFormat getTransferFormat() {
...
                     break;
+                case STREAM_INVOCATION:
                 case STREAM_ITEM:
                 case COMPLETION:
-                case STREAM_INVOCATION:
                 case CANCEL_INVOCATION:
                     throw new UnsupportedOperationException(String.format("The message type %s is not supported yet.", messageType));
                 case PING:
-                    //Ping
                     hubMessages.add(new PingMessage());
```
anurse: While you're here can you quickly bash up a singleton for this? [Looks like](https://www.journaldev.com/1377/java-singleton-design-pattern-best-practices-examples) the java pattern would be: 

```java
class PingMessage
{
    private static final PingMessage instance = new PingMessage();
    
    private PingMessage()
    {
    }

    public static PingMessage getInstance(){
        return instance;
    }
}

File: clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java (hunk 2 above)
Comment (anurse) [reply clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java]: Yeah, let's match .NET here

File: clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java (hunk 15 above)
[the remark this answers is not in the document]
Comment (BrennanConroy) [reply clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java]: Java 😞

File: clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java (hunk 9 above)
Comment (BrennanConroy) [reply clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java]: Because it's going to have more later https://github.com/aspnet/SignalR/blob/adc5706bcb2b3cb85e98e303999eb9a778426f7c/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L1106

File: clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java (hunk 16 above)
Comment (mikaelm12) [reply clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java]: I can do this as part of API review changes. No need to add it here

Comment (anurse) [reply clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java]: As long as it's tracked somewhere

File: clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/test/JsonHubProtocolTest.java [hunk 17]
```diff
@@ -161,23 +183,89 @@ public void ParseTwoMessages() {
...
-        assertEquals(42, messageResult.get(0).getAsInt());
-        assertEquals(24, messageResult.get(1).getAsInt());
+        int messageResult = (int) message.getArguments()[0];
+        int messageResult2 = (int) message.getArguments()[1];
+        assertEquals(42, messageResult);
+        assertEquals(24, messageResult2);
+    }
+
+    @Test
+    public void ParseMessageWithOutOfOrderProperties() throws Exception {
```
anurse: Don't make it worse Brennan! Put some camel in there :).

BrennanConroy: Ugh!!!!!

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/StreamInvocationMessage.java
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+class StreamInvocationMessage extends InvocationMessage {
```
mikaelm12: 👍

PR #2944, opened 2018-09-11
WIP starting to add Async APIs

mikaelm12:
Starting to phase in the Async APIs.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -3,10 +3,12 @@
 
 package com.microsoft.aspnet.signalr;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 
 import com.google.gson.Gson;
@@ -151,6 +153,45 @@ public class HubConnection {
         return connectionState;
     }
 
+
+    public CompletableFuture startAsync() throws Exception {
+        if (connectionState != HubConnectionState.DISCONNECTED) {
+            return CompletableFuture.completedFuture(null);
+        }
+        if (!skipNegotiate) {
+            int negotiateAttempts = 0;
+            do {
+                accessToken = (negotiateResponse == null) ? null : negotiateResponse.getAccessToken();
+                negotiateResponse = handleNegotiate();
+                negotiateAttempts++;
+            } while (negotiateResponse.getRedirectUrl() != null && negotiateAttempts < MAX_NEGOTIATE_ATTEMPTS);
+            if (!negotiateResponse.getAvailableTransports().contains("WebSockets")) {
+                throw new HubException("There were no compatible transports on the server.");
+            }
+        }
+
+        logger.log(LogLevel.Debug, "Starting HubConnection");
+        if (transport == null) {
+            transport = new WebSocketTransport(url, logger, headers);
+        }
+
+        transport.setOnReceive(this.callback);
+
+        return transport.startAsync().thenCompose((future) -> {
+            String handshake = HandshakeProtocol.createHandshakeRequestMessage(new HandshakeRequestMessage(protocol.getName(), protocol.getVersion()));
+            CompletableFuture handshakeFuture = null;
+            try {
+                 handshakeFuture = transport.sendAsync(handshake);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+            return handshakeFuture.thenRun(() -> {
+                connectionState = HubConnectionState.CONNECTED;
+                logger.log(LogLevel.Information, "HubConnected started.");
+            });
+        });
+    }
+
     /**
      * Starts a connection to the server.
      *
@@ -164,24 +205,7 @@ public class HubConnection {
             int negotiateAttempts = 0;
             do {
                 accessToken = (negotiateResponse == null) ? null : negotiateResponse.getAccessToken();
-                negotiateResponse = Negotiate.processNegotiate(url, accessToken);
-
-                if (negotiateResponse.getConnectionId() != null) {
-                    if (url.contains("?")) {
-                        url = url + "&id=" + negotiateResponse.getConnectionId();
-                    } else {
-                        url = url + "?id=" + negotiateResponse.getConnectionId();
-                    }
-                }
-
-                if (negotiateResponse.getAccessToken() != null) {
-                    this.headers.put("Authorization", "Bearer " + negotiateResponse.getAccessToken());
-                }
-
-                if (negotiateResponse.getRedirectUrl() != null) {
-                    url = this.negotiateResponse.getRedirectUrl();
-                }
-
+                negotiateResponse = handleNegotiate();
                 negotiateAttempts++;
             } while (negotiateResponse.getRedirectUrl() != null && negotiateAttempts < MAX_NEGOTIATE_ATTEMPTS);
             if (!negotiateResponse.getAvailableTransports().contains("WebSockets")) {
@@ -200,6 +224,29 @@ public class HubConnection {
         transport.send(handshake);
         connectionState = HubConnectionState.CONNECTED;
         logger.log(LogLevel.Information, "HubConnected started.");
+    }
+
+    private NegotiateResponse handleNegotiate() throws IOException {
+        accessToken = (negotiateResponse == null) ? null : negotiateResponse.getAccessToken();
+        negotiateResponse = Negotiate.processNegotiate(url, accessToken);
+
+        if (negotiateResponse.getConnectionId() != null) {
+            if (url.contains("?")) {
+                url = url + "&id=" + negotiateResponse.getConnectionId();
+            } else {
+                url = url + "?id=" + negotiateResponse.getConnectionId();
+            }
+        }
+
+        if (negotiateResponse.getAccessToken() != null) {
+            this.headers.put("Authorization", "Bearer " + negotiateResponse.getAccessToken());
+        }
+
+        if (negotiateResponse.getRedirectUrl() != null) {
+            url = this.negotiateResponse.getRedirectUrl();
+        }
+
+        return negotiateResponse;
     }
 
     /**
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -9,6 +9,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 
 import com.google.gson.Gson;
@@ -30,6 +32,7 @@ public class HubConnection {
     private NegotiateResponse negotiateResponse;
     private String accessToken;
     private Map<String, String> headers = new HashMap<>();
+    private Lock hubConnectionStateLock = new ReentrantLock();
 
     private static int MAX_NEGOTIATE_ATTEMPTS = 100;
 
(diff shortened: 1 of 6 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,6 +153,45 @@ public HubConnectionState getConnectionState() {
         return connectionState;
     }
 
+
+    public CompletableFuture startAsync() throws Exception {
```
anurse: I know this was originally done when prototyping, but if we're doing this for real now, we should call it `start` and replace the existing one. If that means updating the tests, let's get that done :)

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,6 +153,45 @@ public HubConnectionState getConnectionState() {
...
+        if (connectionState != HubConnectionState.DISCONNECTED) {
```
anurse: We're going to need a lock here (we probably needed it before too). If I call `connection.start()` twice without waiting for the first one, the second one should block until the first one finishes. See [HttpConnection.StartAsyncCore](https://github.com/aspnet/SignalR/blob/adc5706bcb2b3cb85e98e303999eb9a778426f7c/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs#L195)

anurse: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html ?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,6 +153,45 @@ public HubConnectionState getConnectionState() {
...
+            return CompletableFuture.completedFuture(null);
+        }
+        if (!skipNegotiate) {
+            int negotiateAttempts = 0;
+            do {
+                accessToken = (negotiateResponse == null) ? null : negotiateResponse.getAccessToken();
+                negotiateResponse = handleNegotiate();
```
anurse: Make sure there's a bug tracking async negotiation unless you're going to do it in this PR.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Transport.java
```diff
@@ -3,9 +3,13 @@
...
+import java.util.concurrent.CompletableFuture;
+
 public interface Transport {
     void start() throws Exception;
+    CompletableFuture startAsync() throws Exception;
     void send(String message) throws Exception;
+    CompletableFuture sendAsync(String message) throws Exception;
     void setOnReceive(OnReceiveCallBack callback);
     void onReceive(String message) throws Exception;
     void stop();
```
anurse: The interface should be async here, even if the implementation is fully sync and just returns a completed future.

anurse: `WebSocketClient.closeConnection` can probably block though so it should be wrapped in `CompletableFuture.runAsync` anyway.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/Transport.java
```diff
@@ -3,9 +3,13 @@
 
 package com.microsoft.aspnet.signalr;
 
+import java.util.concurrent.CompletableFuture;
+
 public interface Transport {
     void start() throws Exception;
+    CompletableFuture startAsync() throws Exception;
     void send(String message) throws Exception;
+    CompletableFuture sendAsync(String message) throws Exception;
```
anurse: As above, let's replace the non-async ones entirely.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -59,11 +60,29 @@ public void start() throws Exception {
...
     }
 
+    @Override
+    public CompletableFuture startAsync() {
+        return CompletableFuture.runAsync(() -> {
+            webSocketClient = createWebSocket(headers);
+            try {
+                webSocketClient.connectBlocking();
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
BrennanConroy: We rethrow or something

mikaelm12: Was having some trouble with that actually. This actually fails one of the tests because the exception gets smothered. But I wasn't able to re-throw. Figuring that out now

mikaelm12:
Currently theres a test failing. It checks the webSockets failure to start test. It's failing because I'm having trouble properly propagating errors from CompletableFutures from the transport

anurse:
Hrm, this isn't building in VSTS for some reason...

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
mikaelm12: Figured out that we have to throw Runtime Exceptions here

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,79 +156,97 @@ public HubConnectionState getConnectionState() {
...
+                transport = new WebSocketTransport(url, logger, headers);
+            }
+
+            transport.setOnReceive(this.callback);
+
+            return transport.start().thenCompose((future) -> {
+                String handshake = HandshakeProtocol.createHandshakeRequestMessage(new HandshakeRequestMessage(protocol.getName(), protocol.getVersion()));
+                CompletableFuture handshakeFuture= transport.send(handshake);
+                return handshakeFuture.thenRun(() -> {
+                    connectionState = HubConnectionState.CONNECTED;
```
BrennanConroy: The `connectionState` is changed here outside of a lock.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,79 +156,97 @@ public HubConnectionState getConnectionState() {
...
+            if (transport == null) {
+                transport = new WebSocketTransport(url, logger, headers);
+            }
+
+            transport.setOnReceive(this.callback);
+
+            return transport.start().thenCompose((future) -> {
+                String handshake = HandshakeProtocol.createHandshakeRequestMessage(new HandshakeRequestMessage(protocol.getName(), protocol.getVersion()));
+                CompletableFuture handshakeFuture= transport.send(handshake);
+                return handshakeFuture.thenRun(() -> {
```
BrennanConroy: just `.thenRun` on the `transport.send` no need for an extra variable.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,79 +156,97 @@ public HubConnectionState getConnectionState() {
...
 
-        logger.log(LogLevel.Debug, "Starting HubConnection");
-        if (transport == null) {
-            transport = new WebSocketTransport(url, logger, headers);
+        if (negotiateResponse.getAccessToken() != null) {
+            this.headers.put("Authorization", "Bearer " + negotiateResponse.getAccessToken());
+        }
+
+        if (negotiateResponse.getRedirectUrl() != null) {
+            url = this.negotiateResponse.getRedirectUrl();
```
BrennanConroy: nit: `this.url`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,79 +156,97 @@ public HubConnectionState getConnectionState() {
...
+        }
+    }
 
-                negotiateAttempts++;
-            } while (negotiateResponse.getRedirectUrl() != null && negotiateAttempts < MAX_NEGOTIATE_ATTEMPTS);
-            if (!negotiateResponse.getAvailableTransports().contains("WebSockets")) {
-                throw new HubException("There were no compatible transports on the server.");
+    private NegotiateResponse handleNegotiate() throws IOException {
+        accessToken = (negotiateResponse == null) ? null : negotiateResponse.getAccessToken();
+        negotiateResponse = Negotiate.processNegotiate(url, accessToken);
```
BrennanConroy: Shouldn't we make negotiate async in this PR too?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -151,79 +156,97 @@ public HubConnectionState getConnectionState() {
...
+            } else {
+                logger.log(LogLevel.Debug, "Stopping HubConnection.");
+            }
+
+            transport.stop();
+            connectionState = HubConnectionState.DISCONNECTED;
+            logger.log(LogLevel.Information, "HubConnection stopped.");
+            if (onClosedCallbackList != null) {
+                HubException hubException = new HubException(errorMessage);
+                for (Consumer<Exception> callback : onClosedCallbackList) {
```
BrennanConroy: This looks like good deadlock potential. It is common practice to try and restart the connection in the `onClosed` callbacks which are currently run inside this lock. And that would mean calling `start`, which trys to acquire the lock resulting in a deadlock.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -47,21 +48,24 @@ private URI formatUrl(String url) throws URISyntaxException {
     }
 
     @Override
-    public void start() throws Exception {
-        logger.log(LogLevel.Debug, "Starting Websocket connection.");
```
BrennanConroy: Add the logging back

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -47,21 +48,24 @@ private URI formatUrl(String url) throws URISyntaxException {
...
-        }
-        logger.log(LogLevel.Information, "WebSocket transport connected to: %s", webSocketClient.getURI());
+    public CompletableFuture start() {
+        return CompletableFuture.runAsync(() -> {
+            webSocketClient = createWebSocket(headers);
+            try {
+                if (!webSocketClient.connectBlocking()) {
+                    throw new RuntimeException("There was an error starting the Websockets transport");
+                }
+
```
BrennanConroy: nit: empty line

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/test/HubConnectionTest.java
```diff
@@ -708,11 +709,13 @@ public void CannotSendBeforeStart() throws Exception {
         private ArrayList<String> sentMessages = new ArrayList<>();
 
         @Override
-        public void start() {}
+        public CompletableFuture start() {return CompletableFuture.completedFuture(null);}
```
BrennanConroy: nit: formatting

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/test/HubConnectionTest.java
```diff
@@ -708,11 +709,13 @@ public void CannotSendBeforeStart() throws Exception {
...
 
         @Override
-        public void send(String message) {
+        public CompletableFuture send(String message) {
+
```
BrennanConroy: nit: empty line

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/test/HubConnectionTest.java
```diff
@@ -726,7 +729,7 @@ public void onReceive(String message) throws Exception {
         }
 
         @Override
-        public void stop() {}
+        public CompletableFuture stop() {return  CompletableFuture.completedFuture(null);}
```
BrennanConroy: nit: formatting

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/test/WebSocketTransportTest.java
```diff
@@ -21,6 +21,8 @@ public void WebsocketThrowsIfItCantConnect() throws Exception {
         expectedEx.expect(Exception.class);
         expectedEx.expectMessage("There was an error starting the Websockets transport");
         Transport transport = new WebSocketTransport("www.notarealurl12345.fake", new NullLogger());
-        transport.start();
+        CompletableFuture future = transport.start();
+        future.get();
+
```
BrennanConroy: nit: empty line

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/test/WebSocketTransportTest.java
```diff
@@ -21,6 +21,8 @@ public void WebsocketThrowsIfItCantConnect() throws Exception {
         expectedEx.expect(Exception.class);
         expectedEx.expectMessage("There was an error starting the Websockets transport");
         Transport transport = new WebSocketTransport("www.notarealurl12345.fake", new NullLogger());
-        transport.start();
+        CompletableFuture future = transport.start();
```
BrennanConroy: Why not just `.get` inline?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: I don't want to block invoke any longer than it needs to be 😄

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: Yeah, good eye. We just need to reduce the scope of the locked operations.

mikaelm12:
Closing this in favor of https://github.com/aspnet/SignalR/pull/2971
It addresses the outstanding comments here and is rebased

PR #2945, opened 2018-09-11
Making all the things Package Private in the Java Client

mikaelm12:
Part 1 of API review PRs
The default access level in Java if package private. So everything in the same package has access to it. This is the closest we can get to internal. 
I've made most of the types package private and moved the test code into the same package as the source, `com.microsoft.src`, but they are still in their own directories.
Also, Spotless is awesome 😄


clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HandshakeProtocolTest.java
```diff
@@ -1,16 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-package com.microsoft.aspnet.signalr.test;
+package com.microsoft.aspnet.signalr;
```
anurse: Unfortunate that this doesn't work with nested packages, but this works so 🤷‍♂️

PR #2946, opened 2018-09-11
Collect "redis MONITOR" logs

BrennanConroy:
Spinning up another container to run the "MONITOR" redis command against the redis server to collect more logs.


test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
```diff
@@ -97,28 +103,47 @@ public void Stop(ILogger logger)
...
+
+            // Need to WaitForExit without a timeout to guarantee the output stream has written everything
+            process.WaitForExit();
+
+            output = string.Join(Environment.NewLine, lines);
+
+            return process.ExitCode;
+        }
+
+        private static (Process, ConcurrentQueue<string>) RunProcess(string fileName, string arguments, ILogger logger)
```
anurse: Can you put a prefix into the log messages indicating where they come from? Might need to pass that in here (since the file name will be long and awkward). I'd like to know which lines come from redis and which from the monitor process.

BrennanConroy: Yeah, I was just thinking about that

anurse:
@dotnet-bot test Windows Release Spanish Language x64 Build please

Issue #2949, opened 2018-09-12
Randomly getting Missing required property 'protocol'.


drajvver:
Hello everyone.

I'm writing there, since I've googled, searched almost everywhere I could and probably exhausted all other resources to get the informations.
I'm using SignalR with AspNetCore App 2.1.1, using the newest version - 1.0.3.
The problem is that once in a while, without apparent pattern, this exception occures:

`System.IO.InvalidDataException: Missing required property 'protocol'.
   at Microsoft.AspNetCore.SignalR.Protocol.HandshakeProtocol.TryParseRequestMessage(ReadOnlySequence1& buffer, HandshakeRequestMessage& requestMessage)
   at Microsoft.AspNetCore.SignalR.HubConnectionContext.HandshakeAsync(TimeSpan timeout, IReadOnlyList1 supportedProtocols, IHubProtocolResolver protocolResolver, IUserIdProvider userIdProvider, Boolean enableDetailedErrors)`

All clients are connecting to the app from native iOS Swift app (here it's using the SignalR-Client-Swift) or native Android app. I wanted to log this issue and corelate it with, for example IP address or other request attributes that could. App creator claims that protocol is being sent correctly, but I don't have access to his codebase to make sure of that.
My bet is that someone else is trying to connect to the socket without sending proper payloads.
I've managed to catch info about payload in logs by setting the Verbose level, getting this:

> 2018-09-12 14:11:03 [Debug] Establishing new connection.
> 2018-09-12 14:11:03 [Debug] OnConnectedAsync started.
> 2018-09-12 14:11:03 [Debug] Socket opened using Sub-Protocol: 'null'.
> 2018-09-12 14:11:03 [Verbose] Message received. Type: Text, size: 32, EndOfMessage: True.
> 2018-09-12 14:11:03 [Debug] Found protocol implementation for requested protocol: "json".
> 2018-09-12 14:11:03 [Information] Completed connection handshake. Using HubProtocol '"json"'.

But how do I corelate this to other stuff that's going on in the request?

I've also tried creating a middleware, and got pretty far with it, but unfortunately after "catching" the request that I need, it stops rest of the app from working, since the request is already read and cannot be passed forward to SignalR to work with.

Anyone have any ideas, or maybe you already stumbled upon that issue and managed to fix it somehow? Or maybe this is completely random and I shouldn't worry much about it?

davidfowl:
[...]

PR #2953, opened 2018-09-13
Improve the handshake request parsing errors

davidfowl:
- Print out the raw handshake payload as utf8 text if it fails to parse as JSON or if we're missing properties. This should help flesh out potentially buggy clients.

#2949


test/Microsoft.AspNetCore.SignalR.Common.Tests/Internal/Protocol/HandshakeProtocolTests.cs
```diff
@@ -50,10 +50,10 @@ public void ParsingHandshakeRequestNotCompleteReturnsFalse()
         [InlineData("42\u001e", "Unexpected JSON Token Type 'Integer'. Expected a JSON Object.")]
         [InlineData("\"42\"\u001e", "Unexpected JSON Token Type 'String'. Expected a JSON Object.")]
         [InlineData("null\u001e", "Unexpected JSON Token Type 'Null'. Expected a JSON Object.")]
-        [InlineData("{}\u001e", "Missing required property 'protocol'.")]
+        [InlineData("{}\u001e", "Missing required property 'protocol'. {}")]
         [InlineData("[]\u001e", "Unexpected JSON Token Type 'Array'. Expected a JSON Object.")]
-        [InlineData("{\"protocol\":\"json\"}\u001e", "Missing required property 'version'.")]
-        [InlineData("{\"version\":1}\u001e", "Missing required property 'protocol'.")]
+        [InlineData("{\"protocol\":\"json\"}\u001e", "Missing required property 'version'. {\"protocol\":\"json\"}")]
+        [InlineData("{\"version\":1}\u001e", "Missing required property 'protocol'. {\"version\":1}")]
```
mikaelm12: Maybe add another test with a properly formatted json payload with an unexpected field.

davidfowl: Done

mikaelm12:
Hmm, it looks likes there's an issue with detecting the JRE and with spotless

mikaelm12:
Bad solution but try to pushing an empty commit to re-trigger the build. If that doesn't work I'll investigate further. The other PRs aren't failing so this might just be a one off.

davidfowl:
@dotnet-bot test

davidfowl:

anurse:

davidfowl:
This this good to go.

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -235,6 +236,13 @@ public static bool TryParseRequestMessage(ref ReadOnlySequence<byte> buffer, out
                 Utf8BufferTextReader.Return(textReader);
             }
 
+            // For error messages, we want to print the payload as text
+            string GetPayloadAsString()
+            {
+                // REVIEW: Should we show hex for binary charaters?
```
anurse: I mean, you're only doing this in the HandshakeProtocol which is always JSON, so .... no?

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -214,17 +215,17 @@ public static bool TryParseRequestMessage(ref ReadOnlySequence<byte> buffer, out
                                 completed = true;
                                 break;
                             default:
-                                throw new InvalidDataException($"Unexpected token '{reader.TokenType}' when reading handshake request JSON.");
+                                throw new InvalidDataException($"Unexpected token '{reader.TokenType}' when reading handshake request JSON. {GetPayloadAsString()}");
```
anurse: Put something useful around the payload goop otherwise it looks totally out of place and apropos of nothing in the message.

Suggest: `Blah blah an error occurred. Message content: '{...}'`

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
```diff
@@ -214,17 +215,17 @@ public static bool TryParseRequestMessage(ref ReadOnlySequence<byte> buffer, out
...
 
                     if (protocol == null)
                     {
-                        throw new InvalidDataException($"Missing required property '{ProtocolPropertyName}'.");
+                        throw new InvalidDataException($"Missing required property '{ProtocolPropertyName}'. {GetPayloadAsString()}");
                     }
                     if (protocolVersion == null)
                     {
-                        throw new InvalidDataException($"Missing required property '{ProtocolVersionPropertyName}'.");
+                        throw new InvalidDataException($"Missing required property '{ProtocolVersionPropertyName}'. {GetPayloadAsString()}");
```
BrennanConroy: nit: "Received: {GetPayloadAsString()}"

src/Microsoft.AspNetCore.SignalR.Common/Protocol/HandshakeProtocol.cs
davidfowl: If you sent some BS (like a message pack invocation before the handshake), it would be useful. But I don't think it matters much.

JamesNK:
Is there a security concern with echoing a message with sensitive data back to the client?

Or is it ok because the message will only be returned if detailed error messages are on?

davidfowl:
> Is there a security concern with echoing a message with sensitive data back to the client?

In general, there's nothing security sensitive here, this is echoing back what the client sent. 

> Or is it ok because the message will only be returned if detailed error messages are on?

That's also true, but unlike when we usually send error messages back to the client, this data came from that client in that specific first place. There's no new information being disclosed that the client didn't already have.

PR #2956, opened 2018-09-13
Java Client API review cleanup

mikaelm12:
Hitting a couple of things from the API review here. Part of https://github.com/aspnet/SignalR/issues/2957

- ALL_CAPS Transfer Format enum
- Move HubConnection methods
- Make PingMessage a singleton
- Spotless clean up

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/PingMessage.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/PingMessage.java
@@ -5,18 +5,13 @@ package com.microsoft.aspnet.signalr;
 
 class PingMessage extends HubMessage
 {
-    private static PingMessage instance;
+    private static PingMessage instance = new PingMessage();
 
     private PingMessage()
     {
     }
 
-    public static PingMessage getInstance() {
-        if (instance == null) {
-            instance = new PingMessage();
-        }
-        return instance;
-    }
+    public static PingMessage getInstance() {return instance;}
 
     @Override
     public HubMessageType getMessageType() {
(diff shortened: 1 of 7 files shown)


BrennanConroy:
LGTM

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/PingMessage.java
```diff
@@ -3,12 +3,23 @@
...
+{
+    private static PingMessage instance;
 
-    int type = HubMessageType.PING.value;
+    private PingMessage()
+    {
+    }
+
+    public static PingMessage getInstance() {
+        if (instance == null) {
```
BrennanConroy: This doesn't need to be lazy-initialized

anurse: Yeah, skip the lazy-init and just initialize it in the field.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/PingMessage.java
```diff
@@ -3,12 +3,18 @@
...
+class PingMessage extends HubMessage
+{
+    private static PingMessage instance = new PingMessage();
 
-    int type = HubMessageType.PING.value;
+    private PingMessage()
+    {
+    }
+
+    public static PingMessage getInstance() {return instance;}
```
davidfowl: nit: formatting

mikaelm12: Do we always want a new line after a bracket? I'll add a spotless rule

BrennanConroy: I think it's more about the spaces here

mikaelm12: Yeah I get that, but it's easier to normalize and write rules for if we just put newlines

Issue #2926, opened 2018-09-07
Support for InvokeAsync


anurse:
We only support `SendAsync`-style invocation in the Java Client right now. We need support for waiting for the server to finish the invocation.

anurse:
Sounds like this is the way we have to do it, so let's do it. I wonder if we should put the return type first?

```java
.invoke(String.class, "GetProductName", 123, 456);
```

vs

```java
.invoke("GetProductName", String.class, 123, 456);
```

I like the first one (return type first) because a) it looks closer to the C# and b) it keeps the method name and args together.

BrennanConroy:
https://github.com/aspnet/SignalR/commit/4b378692a4ea350091f1101e519169e28b8662b5

PR #2961, opened 2018-09-14
[Java] Add Invoke support

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2926

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CompletionMessage.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CompletionMessage.java
@@ -1,0 +1,38 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+class CompletionMessage extends HubMessage {
+    private int type = HubMessageType.COMPLETION.value;
+    private String invocationId;
+    private Object result;
+    private String error;
+
+    public CompletionMessage(String invocationId, Object result, String error) {
+        if (error != null && result != null)
+        {
+            throw new IllegalArgumentException("Expected either 'error' or 'result' to be provided, but not both");
+        }
+        this.invocationId = invocationId;
+        this.result = result;
+        this.error = error;
+    }
+
+    public Object getResult() {
+        return result;
+    }
+
+    public String getError() {
+        return error;
+    }
+
+    public String getInvocationId() {
+        return invocationId;
+    }
+
+    @Override
+    public HubMessageType getMessageType() {
+        return HubMessageType.values()[type - 1];
+    }
+}
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -9,6 +9,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
@@ -71,13 +72,13 @@ public class HubConnection {
                 switch (message.getMessageType()) {
                     case INVOCATION:
                         InvocationMessage invocationMessage = (InvocationMessage) message;
-                        List<InvocationHandler> handlers = this.handlers.get(invocationMessage.target);
+                        List<InvocationHandler> handlers = this.handlers.get(invocationMessage.getTarget());
                         if (handlers != null) {
                             for (InvocationHandler handler : handlers) {
-                                handler.getAction().invoke(invocationMessage.arguments);
+                                handler.getAction().invoke(invocationMessage.getArguments());
                             }
                         } else {
-                            logger.log(LogLevel.Warning, "Failed to find handler for %s method.", invocationMessage.target);
+                            logger.log(LogLevel.Warning, "Failed to find handler for %s method.", invocationMessage.getMessageType());
                         }
                         break;
                     case CLOSE:
@@ -88,10 +89,18 @@ public class HubConnection {
                     case PING:
                         // We don't need to do anything in the case of a ping message.
                         break;
+                    case COMPLETION:
+                        CompletionMessage completionMessage = (CompletionMessage)message;
+                        InvocationRequest irq = connectionState.tryRemoveInvocation(completionMessage.getInvocationId());
+                        if (irq == null) {
+                            logger.log(LogLevel.Warning, "Dropped unsolicited Completion message for invocation '%s'.", completionMessage.getInvocationId());
+                            continue;
+                        }
+                        irq.complete(completionMessage);
+                        break;
                     case STREAM_INVOCATION:
                     case STREAM_ITEM:
                     case CANCEL_INVOCATION:
-                    case COMPLETION:
                         logger.log(LogLevel.Error, "This client does not support %s messages.", message.getMessageType());
 
                         throw new UnsupportedOperationException(String.format("The message type %s is not supported yet.", message.getMessageType()));
@@ -220,6 +229,7 @@ public class HubConnection {
      * Stops a connection to the server.
      */
     private void stop(String errorMessage) {
+        HubException hubException = null;
         hubConnectionStateLock.lock();
         try {
             if (hubConnectionState == HubConnectionState.DISCONNECTED) {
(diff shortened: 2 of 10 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -244,9 +255,50 @@ public void send(String method, Object... args) throws Exception {
...
+            if (v != null) {
+                // This should never happen
+                throw new IllegalStateException("Invocation Id is already used");
+            }
+
+            InvocationRequest irq = new InvocationRequest(returnType);
+            CompletableFuture<Object> pendingCall = irq.getPendingCall();
+            // forward the invocation result to the user
+            pendingCall.thenAccept((returnValue) -> {
+                // Primitive types can't be cast with the Object cast function
```
anurse: nit: the function is "Class.cast" :)

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -244,9 +255,50 @@ public void send(String method, Object... args) throws Exception {
...
+        // Make sure the actual send is after setting up the future otherwise there is a race
+        // where the map doesn't have the future yet when the response is returned
+        sendHubMessage(invocationMessage);
+
+        return future;
+    }
+
+    private void sendHubMessage(HubMessage message) throws Exception {
+        String serializedMessage = protocol.writeMessage(message);
+        logger.log(LogLevel.Debug, "Sending message.");
```
anurse: Put a little more detail in here maybe? What kind of message are you sending and the invocation ID if there is one? (see https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.Log.cs#L28)

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -304,12 +306,141 @@ public void registeringMultipleHandlersThatTakeParamsAndBothGetTriggered() throw
...
+
+        mockTransport.receiveMessage("{\"type\":3,\"invocationId\":\"1\",\"result\":42}" + RECORD_SEPARATOR);
+        assertTrue(result.isDone());
+
+        assertEquals(42, result.get(), 0);
+        assertEquals("message", result2.get());
+    }
+
+    @Test
+    public void completionMessageValuesCanBeOutOfOrder() throws Exception {
```
anurse: nit: technically this is a JsonHubProtocol test. Do we have those already? Could you move this there and test it that way instead?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -244,9 +259,54 @@ public void send(String method, Object... args) throws Exception {
...
+        InvocationMessage invocationMessage = new InvocationMessage(method, args);
+        String id = connectionState.getNextId();
+        invocationMessage.setInvocationId(id);
+        CompletableFuture<T> future = new CompletableFuture<>();
+
+        connectionState.pendingInvocations.compute(id, (k, v) -> {
+            if (v != null) {
+                // This should never happen
+                throw new IllegalStateException("Invocation Id is already used");
+            }
```
davidfowl: If this happens, what state will the connection be in? Will it leak memory? Will things not get cleaned up?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -244,9 +259,54 @@ public void send(String method, Object... args) throws Exception {
...
+
+            InvocationRequest irq = new InvocationRequest(returnType);
+            CompletableFuture<Object> pendingCall = irq.getPendingCall();
+            // forward the invocation result to the user
+            pendingCall.thenAccept((returnValue) -> {
+                // Primitive types can't be cast with the Class cast function
+                if (returnType.isPrimitive()) {
+                    future.complete((T)returnValue);
+                } else {
```
davidfowl: Where do these continuations run? Inline? On a thread pool?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -534,15 +594,37 @@ public void onClosed(Consumer<Exception> callback) {
...
 
+        public String getNextId() {
+            int i = nextId.incrementAndGet();
+            return Integer.toString(i);
+        }
+
+        public void cancelOutstandingInvocations(Exception ex) {
+            pendingInvocations.forEach((k, v) -> {
+                if (ex == null) {
+                    v.dispose();
```
davidfowl: When you dispose an invocation what happens? Does the caller see an exception?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java
```diff
@@ -114,9 +119,14 @@ public TransferFormat getTransferFormat() {
                         hubMessages.add(new InvocationMessage(target, arguments.toArray()));
                     }
                     break;
+                case COMPLETION:
+                    if (resultToken != null) {
+                        result = gson.fromJson(resultToken, binder.getReturnType(invocationId));
```
davidfowl: Does this parser work if properties are out of order?

BrennanConroy: https://github.com/aspnet/SignalR/pull/2961/files#diff-40b6641977b917a809fd35253a9368ffR218

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
BrennanConroy: https://github.com/aspnet/SignalR/pull/2961/files#diff-5ec328a725285659deb33f61428b2a70R420
The caller gets a `CancellationException`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
BrennanConroy: Looks like they run on the same thread unless you use the Async suffix version

davidfowl: So the consumer has to chose.. hmmm I don’t love that.

anurse: Yeah, that's generally how it works. If we want, we can force it by running `.thenApplyAsync((f) -> f)` or something similar to force the future on to the thread pool (aka "fork-join pool" in Java).

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CompletionMessage.java
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+class CompletionMessage extends HubMessage {
+    int type = HubMessageType.COMPLETION.value;
```
anurse: This should also be `private`, right?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: > Throws: RuntimeException - or Error if the remappingFunction does so, in which case the mapping is unchanged

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#compute-K-java.util.function.BiFunction-

So the invocation will never be added and this will throw synchronously without having started any work or added anything to the invocation table.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
davidfowl: I think we need to do that (dispatch)

BrennanConroy: @mikaelm12 FYI for your async PR

BrennanConroy:
Async stuff should be good now.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -534,15 +596,39 @@ public Subscription on(String target, Action callback) {
...
+                } else {
+                    v.fail(ex);
+                }
+            });
+
+            pendingInvocations.clear();
```
anurse: There's technically a race here right? If I call invoke in between the forEach and the clear, it will hang.

anurse: Maybe we should just lock here.

BrennanConroy:
@dotnet-bot test

BrennanConroy:

BrennanConroy:

BrennanConroy:

PR #2974, opened 2018-09-18
Move Java chat sample to a samples package

mikaelm12:
Moving the chat sample into a samples package.
It's also good because now only the fully public types, not the package private types, are exposed to the sample.


mikaelm12:
@dotnet-bot  test

mikaelm12:
Looks like I'll need to make a small change since this is now in a samples package. Cool

Issue #2903, opened 2018-09-01
Javascript & Java clients mark state as connected before the handshake response has been received


davidfowl:
The .NET client waits for the handshake to complete before marking the connection as usable but this isn't the same for the javascript client.

https://github.com/aspnet/SignalR/blob/f42f4c56e61c5891a851907cf4639575dbbc8df4/clients/ts/signalr/src/HubConnection.ts#L123

davidfowl:
> I agree it's wrong, but what's the actual customer impact here? Won't the user sends be queued behind the handshake messages and still work?

Not sure, I was reviewing the java client and saw this bug and noticed it was in the javascript client as well. I'm guessing it can result in more failures than necessary since we're exposing the state now in 1.1.

anurse:
Ok cool. You're right, it's backwards and we should fix it, just wanted to understand it fully. I'll put it in 2.2 for now.

BrennanConroy:
Fixed in #3068

PR #2983, opened 2018-09-19
[TS] Wait for handshake response

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2903

### DIFF
--- a/clients/ts/signalr/src/HandshakeProtocol.ts
+++ b/clients/ts/signalr/src/HandshakeProtocol.ts
@@ -55,7 +55,11 @@ export class HandshakeProtocol {
 
         // At this point we should have just the single handshake message
         const messages = TextMessageFormat.parse(messageData);
-        responseMessage = JSON.parse(messages[0]);
+        const response = JSON.parse(messages[0]);
+        if (response.type) {
+            throw new Error("Handshake response should not have a 'type' value.");
+        }
+        responseMessage = response;
 
         // multiple messages could have arrived with handshake
         // return additional data to be parsed as usual, or null if all parsed
--- a/clients/ts/signalr/src/HubConnection.ts
+++ b/clients/ts/signalr/src/HubConnection.ts
@@ -31,6 +31,8 @@ export class HubConnection {
     private id: number;
     private closedCallbacks: Array<(error?: Error) => void>;
     private receivedHandshakeResponse: boolean;
+    private handshakeResolver!: (value?: PromiseLike<{}>) => void;
+    private handshakeRejecter!: (reason?: any) => void;
     private connectionState: HubConnectionState;
 
     // The type of these a) doesn't matter and b) varies when building in browser and node contexts
@@ -106,6 +108,11 @@ export class HubConnection {
         this.logger.log(LogLevel.Debug, "Starting HubConnection.");
 
         this.receivedHandshakeResponse = false;
+        // Set up the promise before any connection is started otherwise it could race with received messages
+        const handshakePromise = new Promise((resolve, reject) => {
+            this.handshakeResolver = resolve;
+            this.handshakeRejecter = reject;
+        });
 
         await this.connection.start(this.protocol.transferFormat);
 
@@ -120,6 +127,8 @@ export class HubConnection {
         this.resetTimeoutPeriod();
         this.resetKeepAliveInterval();
 
+        // Wait for the handshake to complete before marking connection as connected
+        await handshakePromise;
         this.connectionState = HubConnectionState.Connected;
     }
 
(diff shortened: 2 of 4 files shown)


clients/ts/signalr/src/HandshakeProtocol.ts
```diff
@@ -55,7 +55,11 @@ export class HandshakeProtocol {
 
         // At this point we should have just the single handshake message
         const messages = TextMessageFormat.parse(messageData);
-        responseMessage = JSON.parse(messages[0]);
+        const response = JSON.parse(messages[0]);
+        if (response.type) {
+            throw new Error("Handshake response should not have a 'type' value.");
```
davidfowl: What is this?

clients/ts/signalr/src/HubConnection.ts
```diff
@@ -330,6 +339,7 @@ export class HubConnection {
         if (!this.receivedHandshakeResponse) {
             data = this.processHandshakeResponse(data);
             this.receivedHandshakeResponse = true;
+            this.handshakeResolver();
```
davidfowl: Why not move this inside of processHandshakeResponse?

clients/ts/signalr/src/HandshakeProtocol.ts
BrennanConroy: This is checking if we receive a message that isn't a handshake response, we do the same thing in the .NET Client

anurse: Agreed. Perhaps `Expected a handshake response from the server`.

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -163,12 +163,14 @@ describe("HubConnection", () => {
...
 
-                const connection = new TestConnection();
+                const connection = new TestConnection(false);
                 const hubConnection = createHubConnection(connection, logger, mockProtocol);
                 try {
+                    const startPromise = hubConnection.start();
                     const data = "{}" + TextMessageFormat.RecordSeparator;
 
                     connection.receiveText(data);
+                    await startPromise;
```
anurse: For the tests that are directly about the handshake, I'd like to assert that the promise isn't resolved before the `receiveText` method is called.

You can do that with code like this:

```javascript
let startCompleted = false;
const startPromise = hubConnection.start().then(() => startCompleted = true);
...
expect(startCompleted).toBe(false);
connection.receiveText(data);
await startPromise;
```

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -187,13 +189,16 @@ describe("HubConnection", () => {
                     protocolCalled = true;
                 };
 
-                const connection = new TestConnection();
+                const connection = new TestConnection(false);
                 const hubConnection = createHubConnection(connection, logger, mockProtocol);
                 try {
+                    const startPromise = hubConnection.start();
```
anurse: assert that the promise isn't finished before `receiveBinary` below.

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -210,9 +215,10 @@ describe("HubConnection", () => {
                 const mockProtocol = new TestProtocol(TransferFormat.Binary);
                 mockProtocol.onreceive = (d) => receivedProcotolData = d as ArrayBuffer;
 
-                const connection = new TestConnection();
+                const connection = new TestConnection(false);
                 const hubConnection = createHubConnection(connection, logger, mockProtocol);
                 try {
+                    const startPromise = hubConnection.start();
```
anurse: assert that the promise isn't finished before `receiveBinary` below.

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -240,12 +247,14 @@ describe("HubConnection", () => {
                 const mockProtocol = new TestProtocol(TransferFormat.Text);
                 mockProtocol.onreceive = (d) => receivedProcotolData = d as string;
 
-                const connection = new TestConnection();
+                const connection = new TestConnection(false);
                 const hubConnection = createHubConnection(connection, logger, mockProtocol);
                 try {
+                    const startPromise = hubConnection.start();
```
anurse: assert that the promise isn't finished before `receiveText` below.

clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -1092,11 +1086,14 @@ class TestConnection implements IConnection {
     public sentData: any[];
     public lastInvocationId: string | null;
 
-    constructor() {
+    private sendHandshake: boolean | null;
+
+    constructor(sendHandshake: boolean | null = true) {
```
anurse: nit: we have the same logic in the .NET tests and we call this parameter `autoHandshake`. Also, I believe we don't want to permit `null` here either, since we have a default value.

clients/ts/signalr/src/HandshakeProtocol.ts
BrennanConroy: Shall I change the .NET error too then?

Issue #2948, opened 2018-09-12
Error accessing identity name with windows authentication


FTWinston:
* Using Microsoft.AspNetCore.SignalR 1.0.1 on .NET Core 2.1, with @aspnet/signalr 1.0.3.
* Javascript client, running in Chrome v68 & v69 on Windows 10.
* Hosting either in IIS 8 on Windows Server 2012 R2, or in IIS express on Windows 10.

Using windows authentication and long polling, I'm getting an ObjectDisposedException when accessing Context.User.Identity.Name from inside any hub methods, or from OnConnectedAsync.

This has only started happening since upgrading from .NET Core 2.0 and SignalR 1.0.0-alpha2-final.

It doesn't happen every time, but if you have connected succesfully once before, it always happens on your second connection attempt.

```
System.ObjectDisposedException: Safe handle has been closed
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
   at Interop.Advapi32.GetTokenInformation(SafeAccessTokenHandle TokenHandle, UInt32 TokenInformationClass, SafeLocalAllocHandle TokenInformation, UInt32 TokenInformationLength, UInt32& ReturnLength)
   at System.Security.Principal.WindowsIdentity.GetTokenInformation(SafeAccessTokenHandle tokenHandle, TokenInformationClass tokenInformationClass, Boolean nullOnInvalidParam)
   at System.Security.Principal.WindowsIdentity.get_User()
   at System.Security.Principal.WindowsIdentity.<GetName>b__46_0()
   at System.Security.Principal.WindowsIdentity.<>c__DisplayClass62_0.<RunImpersonatedInternal>b__0(Object <p0>)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Security.Principal.WindowsIdentity.RunImpersonatedInternal(SafeAccessTokenHandle token, Action action)
   at System.Security.Principal.WindowsIdentity.RunImpersonated(SafeAccessTokenHandle safeAccessTokenHandle, Action action)
   at System.Security.Principal.WindowsIdentity.GetName()
   at System.Security.Principal.WindowsIdentity.get_Name()
[...]

PR #2985, opened 2018-09-19
Clone Windows Identity in LongPolling connections

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2948

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
@@ -630,8 +630,9 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
 
                     // Don't copy the identity if it's a windows identity
                     // We specifically clone the identity on first poll if it's a windows identity
-                    // There would be a race here if we swapped the new User and the application tried to access the User
-                    if (context.User.Identity.GetType() != typeof(WindowsIdentity))
+                    // If we swapped the new User here we'd have to dispose the old identities which could race with the application
+                    // trying to access the identity.
+                    if (context.User.Identity is WindowsIdentity)
                     {
                         existing.User = context.User;
                     }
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
```diff
@@ -630,21 +627,49 @@ private async Task<bool> EnsureConnectionStateAsync(HttpConnectionContext connec
                 {
                     // Set the request trace identifier to the current http request handling the poll
                     existing.TraceIdentifier = context.TraceIdentifier;
-                    existing.User = context.User;
+
+                    // Don't copy the identity if it's a windows identity
+                    // We specifically clone the identity on first poll if it's a windows identity
+                    // There would be a race here if we swapped the new User and the application tried to access the User
```
anurse: To be fair, the race exists no matter what. It's `Dispose`ing the principal that caused the big problem.

anurse: And it's the race that makes me want to stop updating the User altogether, but that would be a breaking change, so it would have to wait until 3.0 at least

BrennanConroy: Oh right, it was dispose 😆

src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionDispatcher.cs
```diff
@@ -630,21 +627,49 @@ private async Task<bool> EnsureConnectionStateAsync(HttpConnectionContext connec
...
+                    if (context.User.Identity.GetType() != typeof(WindowsIdentity))
```
davidfowl: use is

anurse: Yeah, I was kinda wondering why you used `GetType`

BrennanConroy: Other languages crowding my head :/

anurse:
```
if(context.User.Identity.getType() instanceof WindowsIdentity.class)
```

Issue #2986, opened 2018-09-20
IE 11: Invalid operand to ‘instanceof’: Function expected


balkarov:
Hello. Thank you for your library!

I have problem in Internet Explorer 11. 
I have web application 1 which has signalR. And I have another application2 (without signalR) which has iframe with application 1. 
When I close iframe (remove from dom) in application 2 I get error: `Invalid operand to ‘instanceof’: Function expected` every ~10-15 seconds. 
After debug I saw that after close iframe ArrayBuffer is undefined in signalR. But I open iframe with app 1 ArrayBuffer is defined.
![123123123](https://user-images.githubusercontent.com/6030335/45815722-31447180-bce2-11e8-8eb9-e9f28eec68b2.png)

### Info

* "@aspnet/signalr": "1.0.3"
* Microsoft.AspNetCore.SignalR 1.0.3
* npm 6.1.0
* The JavaScript client
* IIS Express
* Windows 10
* IE 11

anurse:
@BrennanConroy we can probably be a little more defensive here.

BrennanConroy:
This fix will appear in the 1.1.0-preview3 release
Fixed via https://github.com/aspnet/SignalR/commit/ef3d3b1cddebf43e6f8bbfc81519a11356d2095a

Issue #2973, opened 2018-09-18
Change Websocket dependency in Node to be fully javascript


mikaelm12:
Ran into an error I've never seen before when building master to bring the 2.2 changes back to 3.0.

Brennan says it has to do with the Websockets dependency for Node.

```

  > websocket@1.0.26 install C:\github\exp\SignalR\clients\ts\FunctionalTests\node_modules\websocket
  > (node-gyp rebuild 2> builderror.log) || (exit 0)


  C:\github\exp\SignalR\clients\ts\FunctionalTests\node_modules\websocket>if not defined npm_config_node_gyp (node "C:\Program Files\nodejs\node_modules\npm\bin\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild )  else (node "" rebuild )
  Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
MSBUILD : error MSB3428: Could not load the Visual C++ component "VCBuild.exe". To fix this, 1) install the .NET Framework 2.0 SDK, 2) install Microsoft Visual Studio 2005 or 3) add the location of the component to the system path if it is installed elsewhere.  [C:\github\exp\SignalR\clients\ts\FunctionalTests\node_modules\websocket\build\binding.sln] [C:\Users\MikaelMengistu\.dotnet\buildtools\korebuild\3.0.0-alpha1-20180911.2\KoreBuild.proj]
MSBUILD : error MSB3428: Could not load the Visual C++ component "VCBuild.exe". To fix this, 1) install the .NET Framework 2.0 SDK, 2) install Microsoft Visual Studio 2005 or 3) add the location of the component to the system path if it is installed elsewhere.  [C:\github\exp\SignalR\clients\ts\FunctionalTests\node_modules\websocket\build\binding.sln] [C:\Users\MikaelMengistu\.dotnet\buildtools\korebuild\3.0.0-alpha1-20180911.2\KoreBuild.proj]
  npm WARN @aspnet/signalr-protocol-msgpack@1.1.0-preview1-t000 requires a peer of @aspnet/signalr@^1.0.0-preview3 but none is installed. You must install peer dependencies yourself.
  npm WARN karma-summary-reporter@1.5.0 requires a peer of karma@^1.2.0 || 2.x but none is installed. You must install peer dependencies yourself.
  npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules\fsevents):
[...]

PR #2989, opened 2018-09-20
[TS] Change websocket library in Node

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2973
Also, #2986

Inverted a couple `instanceof ArrayBuffer` checks because the `ArrayBuffer` returned from `ws` isn't an ArrayBuffer or something super odd, so checking for string instead and assuming buffer otherwise works.

See https://github.com/nodejs/node/issues/20978 for some context maybe? @anurse maybe you'll understand it better.

### DIFF
--- a/clients/ts/FunctionalTests/package.json
+++ b/clients/ts/FunctionalTests/package.json
@@ -32,7 +32,7 @@
     "karma-summary-reporter": "^1.5.0",
     "ts-node": "^4.1.0",
     "typescript": "^3.0.1",
-    "websocket": " ^1.0.26"
+    "ws": " ^6.0.0"
   },
   "scripts": {
     "clean": "node ../common/node_modules/rimraf/bin.js ./wwwroot/dist ./obj/js",
--- a/clients/ts/signalr-protocol-msgpack/src/Utils.ts
+++ b/clients/ts/signalr-protocol-msgpack/src/Utils.ts
@@ -6,6 +6,6 @@
 export function isArrayBuffer(val: any): val is ArrayBuffer {
     return val && typeof ArrayBuffer !== "undefined" &&
         (val instanceof ArrayBuffer ||
-        // Sometimes we get an ArrayBuffer that comes from a polyfill
+        // Sometimes we get an ArrayBuffer that doesn't satisfy instanceof
         (val.constructor && val.constructor.name === "ArrayBuffer"));
 }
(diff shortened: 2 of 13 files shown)


BrennanConroy:
We'll need to do the needful after this is merged.

anurse:
This is different runtime dependency, so it still needs @Eilon to approve

clients/ts/signalr-protocol-msgpack/src/Utils.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+// Copied from signalr/Utils.ts
+/** @private */
+export function isArrayBuffer(val: any): val is ArrayBuffer {
+    return val && typeof ArrayBuffer !== "undefined" &&
+        (val instanceof ArrayBuffer ||
+        // Sometimes we get an ArrayBuffer that comes from a polyfill
```
anurse: nit: I wrote this comment when I was randomly speculating :). Just say something like "Sometimes we get an ArrayBuffer that doesn't satisfy instanceof 🤷‍♂️ "

clients/ts/signalr/package-lock.json
```diff
@@ -104,21 +62,13 @@
...
+    "ws": {
+      "version": "6.0.0",
+      "resolved": "https://registry.npmjs.org/ws/-/ws-6.0.0.tgz",
+      "integrity": "sha512-c2UlYcAZp1VS8AORtpq6y4RJIkJ9dQz18W32SpR/qXGfLDZ2jU4y4wKvvZwqbi7U6gxFQTeE+urMbXU/tsDy4w==",
       "requires": {
-        "debug": "^2.2.0",
-        "nan": "^2.3.3",
-        "typedarray-to-buffer": "^3.1.2",
-        "yaeti": "^0.0.6"
+        "async-limiter": "~1.0.0"
```
anurse: Fewer extra deps. Nice :)

clients/ts/FunctionalTests/package.json
```diff
@@ -32,7 +32,7 @@
     "karma-summary-reporter": "^1.5.0",
     "ts-node": "^4.1.0",
     "typescript": "^3.0.1",
-    "websocket": " ^1.0.26"
+    "ws": " ^6.0.0"
```
Eilon: So I need to get approval for this one?

BrennanConroy:
Yes please. We're removing the websocket one too if that matters

clients/ts/FunctionalTests/package.json
anurse: Yes, we are swapping in a pure-JavaScript websocket library instead of the previous WebSocket library we were using which had a native component.

Eilon:
@BrennanConroy removals don't matter.

clients/ts/FunctionalTests/package.json
Eilon: Approved.

Issue #2988, opened 2018-09-20
Add "error" property to Negotiate response


anurse:
To make it easier to for older clients to report errors due to incompatibilities, we should allow the `/negotiate` response to include an `error` property. If present, the client will **always** cancel the connection and report the string provided as the error message.

In the JavaScript client, this error should be wrapped in an `Error` instance and the `HubConnection.start` promise should be rejected with that `Error`.

In the .NET and Java clients, this error should be wrapped in an `Exception` (or `HubException`?) and used to fault the `Task`/`CompletableFuture` returned by `HubConnection.Start`

PR #2998, opened 2018-09-20
Add error to negotiate

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2988

No httpclient testability in Java yet.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
@@ -15,10 +15,15 @@ class NegotiateResponse {
     private Set<String> availableTransports = new HashSet<>();
     private String redirectUrl;
     private String accessToken;
+    private String error;
     private JsonParser jsonParser = new JsonParser();
 
     public NegotiateResponse(String negotiatePayload) {
         JsonObject negotiateResponse = jsonParser.parse(negotiatePayload).getAsJsonObject();
+        if (negotiateResponse.has("error")) {
+            this.error = negotiateResponse.get("url").getAsString();
+            return;
+        }
         if (negotiateResponse.has("url")) {
             this.redirectUrl = negotiateResponse.get("url").getAsString();
             if (negotiateResponse.has("accessToken")) {
@@ -47,5 +52,9 @@ class NegotiateResponse {
 
     public String getAccessToken() {
         return accessToken;
+    }
+
+    public String getError() {
+        return error;
     }
 }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
@@ -3,12 +3,12 @@
 
 package com.microsoft.aspnet.signalr;
 
+import java.io.IOException;
+import java.io.StringReader;
 import java.util.HashSet;
 import java.util.Set;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
+import com.google.gson.stream.JsonReader;
 
 class NegotiateResponse {
     private String connectionId;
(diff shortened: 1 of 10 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
```diff
@@ -15,10 +15,15 @@
...
     private String redirectUrl;
     private String accessToken;
+    private String error;
     private JsonParser jsonParser = new JsonParser();
 
     public NegotiateResponse(String negotiatePayload) {
         JsonObject negotiateResponse = jsonParser.parse(negotiatePayload).getAsJsonObject();
+        if (negotiateResponse.has("error")) {
+            this.error = negotiateResponse.get("url").getAsString();
+            return;
```
anurse: I don't think the parsing logic should exit early like this. You're right that an `error` property generally means everything else should be ignored, but let the caller decide that.

src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs
```diff
@@ -428,6 +428,10 @@ private async Task<NegotiationResponse> NegotiateAsync(Uri url, HttpClient httpC
                         {
                             negotiateResponse = NegotiateProtocol.ParseResponse(responseStream);
                         }
+                        if (!string.IsNullOrEmpty(negotiateResponse.Error))
+                        {
+                            throw new InvalidOperationException(negotiateResponse.Error);
```
anurse: `InvalidOperationException` isn't really right here... We don't know anything about the underlying problem, it's no better than just throwing `Exception`.

specs/TransportProtocols.md
```diff
@@ -62,6 +62,19 @@ The `POST [endpoint-base]/negotiate` request is used to establish a connection b
   * The `url` which is the URL the client should connect to.
   * The `accessToken` which is an optional bearer token for accessing the specified url.
 
+
+3. A response that contains an `error` which should abort the connection attempt.
```
davidfowl: Change abort to stop.

davidfowl: Maybe we need a protocol version here @anurse

anurse: Yeah, it's not a bad idea. But do we want to do the two-part handshake that ASP.NET SignalR does? Where the client reports it's max supported version?

davidfowl: ERRR, I hope we don't need to change it that much.

BrennanConroy:
No

test/Microsoft.AspNetCore.SignalR.Client.Tests/HttpConnectionTests.Negotiate.cs
```diff
@@ -380,6 +382,37 @@ public async Task StartSkipsOverTransportsThatDoNotSupportTheRequredTransferForm
...
+                        writeContext.EventId.Name == "ErrorWithNegotiation";
+                }
+
+                var testHttpHandler = new TestHttpMessageHandler(autoNegotiate: false);
+                testHttpHandler.OnNegotiate((request, cancellationToken) =>
+                {
+                    return ResponseUtils.CreateResponse(HttpStatusCode.OK,
+                            JsonConvert.SerializeObject(new
+                            {
+                                error = "Negotiate failed."
```
davidfowl: nit: Change this error message to something that isn't about negotiation. This seems like something that could be ambiguous

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/NegotiateResponse.java
```diff
@@ -3,34 +3,74 @@
...
+                        }
+                        reader.endObject();
+                    }
+                    reader.endArray();
+                    break;
+                case "connectionId":
+                    this.connectionId = reader.nextString();
+                    break;
+                default:
+                    // Skip unknown property, allows new clients to still work with old protocols
```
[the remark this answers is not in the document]
JamesNK: +have a test case for skipping a simple value, e.g. "a string!", and a complex value, e.g. ["array", "of", "content"]

I'm guessing skipValue will skip all of the complex value but it is worth double checking. It is a common bug I've seen in code that manually reads JSON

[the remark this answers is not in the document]
BrennanConroy: We do now, both cases. Thanks James

BrennanConroy:

PR #2999, opened 2018-09-20
Wait for the Handshake Response Java

mikaelm12:
Waiting for the `Handshake Response` in the Java client before entering the `CONNECTED` state. I added an intermediate `CONNECTING` state as well.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -57,6 +57,7 @@ public class HubConnection {
                     throw new HubException(errorMessage);
                 }
                 handshakeReceived = true;
+                hubConnectionState = HubConnectionState.CONNECTED;
 
                 payload = payload.substring(handshakeLength);
                 // The payload only contained the handshake response so we can return.
@@ -214,7 +215,7 @@ public class HubConnection {
             return transport.send(handshake).thenRun(() -> {
                 hubConnectionStateLock.lock();
                 try {
-                    hubConnectionState = HubConnectionState.CONNECTED;
+                    hubConnectionState = HubConnectionState.CONNECTING;
                     connectionState = new ConnectionState(this);
                     logger.log(LogLevel.Information, "HubConnected started.");
                 } finally {
--- a/clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
+++ b/clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
@@ -19,9 +19,12 @@ class HubConnectionTest {
 
     @Test
     public void checkHubConnectionState() throws Exception {
-        Transport mockTransport = new MockTransport();
+        MockTransport mockTransport = new MockTransport();
         HubConnection hubConnection = new HubConnection("http://example.com", mockTransport, true);
         hubConnection.start();
+        assertEquals(HubConnectionState.CONNECTING, hubConnection.getConnectionState());
+
+        mockTransport.receiveMessage("{}" + RECORD_SEPARATOR);
         assertEquals(HubConnectionState.CONNECTED, hubConnection.getConnectionState());
 
         hubConnection.stop();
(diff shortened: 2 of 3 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -57,6 +57,7 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
                     throw new HubException(errorMessage);
                 }
                 handshakeReceived = true;
+                hubConnectionState = HubConnectionState.CONNECTED;
```
mikaelm12: Should probably lock here

anurse:
We should do what we do with .NET and have the `CompletableFuture` remain incomplete until the handshake finishes, and only then complete it.

mikaelm12:
Okay, the start future now waits for the transport to start, the handshake request to be sent and the handshake response to be received before changing the HubConnection state to connected and completing the future.

davidfowl:
What happens if the handshake is never sent from the server  (due to the connection dropping)?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -46,23 +47,28 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
         }
         this.skipNegotiate = skipNegotiate;
         this.callback = (payload) -> {
+            hubConnectionStateLock.lock();
```
BrennanConroy: Why is the lock being used here?

mikaelm12:
> What happens if the handshake is never sent from the server (due to the connection dropping)?

Good point. This should have a timeout. The returned future should be a "WhenAny" of a timeout and the "WhenAll"/"allOf" future of the connect process and receiving the handshake response.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: We used to change the HubConnectionState here. Good find

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -209,20 +210,27 @@ public CompletableFuture start() throws Exception {
...
+        })).thenRun(() -> {
+            hubConnectionStateLock.lock();
+            try {
+                hubConnectionState = HubConnectionState.CONNECTED;
+                connectionState = new ConnectionState(this);
+                logger.log(LogLevel.Information, "HubConnected started.");
+            } finally {
+                hubConnectionStateLock.unlock();
+            }
+        }).acceptEither(timeoutAfter(5, TimeUnit.SECONDS), startFuture -> {});
```
BrennanConroy: I don't think we want a timeout as the server is the one who sends a handshake response with an error if it times out.

And if the connection to the server is dropped then we should `completeExceptionally` the `handshakeResponseFuture` in the `stop` function. Which it looks like we don't wire up to the transport yet.

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -19,15 +17,31 @@
 
     @Test
     public void checkHubConnectionState() throws Exception {
-        Transport mockTransport = new MockTransport();
+        MockTransport mockTransport = new MockTransport();
         HubConnection hubConnection = new HubConnection("http://example.com", mockTransport, true);
         hubConnection.start();
+        assertEquals(HubConnectionState.DISCONNECTED, hubConnection.getConnectionState());
+
+        mockTransport.receiveMessage("{}" + RECORD_SEPARATOR);
```
BrennanConroy: Add "autoHandshake" to MockTransport similar to https://github.com/aspnet/SignalR/blob/release/2.2/clients/ts/signalr/tests/HubConnection.test.ts#L1103

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: So looks like we actually will need the timeout, and we'll have to set it to longer than what the server timeout is. If we sent the handshake and the server went offline we could potentially wait for the response forever

PR #3000, opened 2018-09-21
[TS] Reject handshake if connection closes before response

BrennanConroy:
Updated Jest reference because I was finding it impossible to debug a test failure and found similar issues filed against Jest that have been fixed in 23.X.X. Side-effect is rejects.toThrow no longer works in some of our tests because of another bug https://github.com/facebook/jest/issues/6675. I can revert the Jest change if we want now that I fixed the test.

### DIFF
--- a/clients/ts/common/package.json
+++ b/clients/ts/common/package.json
@@ -10,10 +10,10 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@types/jest": "^22.2.3",
+    "@types/jest": "^23.3.2",
     "@types/node": "^8.5.2",
     "@types/webpack": "^4.4.0",
-    "jest": "^22.4.3",
+    "jest": "^23.6.0",
     "jest-junit": "^5.1.0",
     "rimraf": "^2.6.2",
     "ts-jest": "^22.4.6",
--- a/clients/ts/signalr/tests/HubConnection.test.ts
+++ b/clients/ts/signalr/tests/HubConnection.test.ts
@@ -273,6 +273,27 @@ describe("HubConnection", () => {
             });
         });
 
+        it.only("start completes if connection closes and handshake not received yet", async () => {
+            await VerifyLogger.run(async (logger) => {
+                const mockProtocol = new TestProtocol(TransferFormat.Text);
+
+                const connection = new TestConnection(false);
+                const hubConnection = createHubConnection(connection, logger, mockProtocol);
+                try {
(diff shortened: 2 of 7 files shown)


clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -273,6 +273,27 @@ describe("HubConnection", () => {
...
+        it.only("start completes if connection closes and handshake not received yet", async () => {
+            await VerifyLogger.run(async (logger) => {
+                const mockProtocol = new TestProtocol(TransferFormat.Text);
+
+                const connection = new TestConnection(false);
+                const hubConnection = createHubConnection(connection, logger, mockProtocol);
+                try {
+                    let startCompleted = false;
+                    const startPromise = hubConnection.start().then(() => startCompleted = true);
+                    expect(startCompleted).toBe(false);
```
davidfowl: Shouldn't this assert after the await?

davidfowl:
```
Microsoft.AspNetCore.SignalR.Tests.HubConnectionHandlerTests.SendToOthersInGroup(hubType: typeof(Microsoft.AspNetCore.SignalR.Tests.HubT))
13:07:39   Error Message:
13:07:39    System.TimeoutException : Operation in SendToOthersInGroup timed out at /_/test/Microsoft.AspNetCore.SignalR.Tests/HubConnectionHandlerTests.cs:1327
```

davidfowl:

clients/ts/common/package.json
```diff
@@ -10,10 +10,10 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@types/jest": "^22.2.3",
+    "@types/jest": "^23.3.2",
     "@types/node": "^8.5.2",
     "@types/webpack": "^4.4.0",
-    "jest": "^22.4.3",
+    "jest": "^23.6.0",
```
anurse: FYI @Eilon. Updated `devDependencies` (internal use only, not distributed)

clients/ts/signalr/tests/HubConnection.test.ts
anurse: No, it's asserting that `startCompleted` is **false**. It will be `true` after the `await`. This is to ensure that the promise doesn't complete until the handshake is received.

clients/ts/common/package.json
Eilon: Up-roved.

PR #3010, opened 2018-09-25
[Java] Throw useful error on incorrect argument length

BrennanConroy:


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java
```diff
@@ -151,4 +141,42 @@ public TransferFormat getTransferFormat() {
...
         return gson.toJson(hubMessage) + RECORD_SEPARATOR;
     }
+
+    private ArrayList<Object> bindArguments(JsonArray argumentsToken, List<Class<?>> paramTypes) {
+        if (argumentsToken.size() != paramTypes.size()) {
+            throw new RuntimeException(String.format("Invocation provides %d argument(s) but target expects %d.", argumentsToken.size(), paramTypes.size()));
+        }
+
+        ArrayList<Object> arguments = null;
+        if (paramTypes != null && paramTypes.size() >= 1) {
```
mikaelm12: You would have already null ref'd in the above `paramTypes.size()`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/JsonHubProtocol.java
```diff
@@ -151,4 +141,42 @@ public TransferFormat getTransferFormat() {
...
+        int paramCount = paramTypes.size();
+        int argCount = 0;
+        ArrayList<Object> arguments = new ArrayList<>();
+        while (reader.peek() != JsonToken.END_ARRAY) {
+            if (argCount < paramCount) {
+                arguments.add(gson.fromJson(reader, paramTypes.get(argCount)));
+            } else {
+                reader.skipValue();
+            }
+            ++argCount;
```
mikaelm12: nit: `argCount++`
The order should matter since you're not directly using whats returned here but I think we usually put the `++` after.

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/JsonHubProtocolTest.java
```diff
@@ -219,6 +219,24 @@ public void parseCompletionMessageWithOutOfOrderProperties() throws Exception {
         assertEquals(42 , message.getResult());
     }
 
+    @Test
+    public void errorWhileParsingMismatchingArgumentLengthWithOutOfOrderProperties() throws Exception {
```
mikaelm12: Can you add a test for passing in fewer arguments than expected

Issue #2915, opened 2018-09-05
Java client should support using custom HttpClient for testing


mikaelm12:
This will open up the test-ability of the client by allowing us to test code paths that depend on network calls that we will then be able to simulate.

BrennanConroy:
https://github.com/aspnet/SignalR/commit/713e224591262ba52f871ae09ec66f95018e607b

PR #3011, opened 2018-09-25
[Java] Add HttpClient support for testing

BrennanConroy:
https://github.com/aspnet/SignalR/issues/2915

Wrote a single test for proof-of-concept. Will leave more tests to a later PR, want this change to be focused on getting the HttpClient code in.

The main part to focus on is the negotiate stuff, it seems a little hacky right now.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/DefaultHttpClient.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/DefaultHttpClient.java
@@ -12,20 +12,21 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import okhttp3.ResponseBody;
 
 class DefaultHttpClient extends HttpClient {
     private OkHttpClient client = new OkHttpClient();
 
     @Override
     public CompletableFuture<HttpResponse> send(HttpRequest httpRequest) {
-        RequestBody body = RequestBody.create(null, new byte[] {});
         Request.Builder requestBuilder = new Request.Builder().url(httpRequest.url);
         if (httpRequest.method == "GET") {
             requestBuilder.get();
         } else if (httpRequest.method == "POST") {
+            RequestBody body = RequestBody.create(null, new byte[] {});
             requestBuilder.post(body);
         } else if (httpRequest.method == "DELETE") {
-            requestBuilder.delete(body);
+            requestBuilder.delete();
         }
 
         Request request = requestBuilder.build();
@@ -40,9 +41,10 @@ class DefaultHttpClient extends HttpClient {
 
             @Override
             public void onResponse(Call call, Response response) throws IOException {
-                // TODO: Look at response.body().string() what if it's binary data?
-                HttpResponse httpResponse = new HttpResponse(response.code(), response.message(), response.body().string());
-                responseFuture.complete(httpResponse);
+                try (ResponseBody body = response.body()) {
+                    HttpResponse httpResponse = new HttpResponse(response.code(), response.message(), body.string());
+                    responseFuture.complete(httpResponse);
+                }
             }
 
         });
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/DefaultHttpClient.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/DefaultHttpClient.java
@@ -1,0 +1,53 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.aspnet.signalr;
+
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
+class DefaultHttpClient extends HttpClient {
+
+    @Override
+    public CompletableFuture<HttpResponse> send(HttpRequest httpRequest) {
+		OkHttpClient client = new OkHttpClient();
+        RequestBody body = RequestBody.create(null, new byte[] {});
+        Request.Builder requestBuilder = new Request.Builder().url(httpRequest.url);
+        if (httpRequest.method == "GET") {
+            requestBuilder.get();
+        } else if (httpRequest.method == "POST") {
+            requestBuilder.post(body);
+        } else if (httpRequest.method == "DELETE") {
+            requestBuilder.delete(body);
+        }
+
+        Request request = requestBuilder.build();
+
+        CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
+
+        client.newCall(request).enqueue(new Callback() {
+            @Override
+            public void onFailure(Call call, IOException e) {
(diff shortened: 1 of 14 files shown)


clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/TestHttpClient.java
```diff
@@ -0,0 +1,54 @@
...
+    public CompletableFuture<HttpResponse> send(HttpRequest request) {
+        this.sentRequests.add(request);
+        return this.handler.apply(request);
+    }
+
+    public List<HttpRequest> getSentRequests() {
+        return sentRequests;
+    }
+    /*
+     * public on(method: string | RegExp, url: string, handler: TestHttpHandler): TestHttpClient;
```
mikaelm12: Copy pasta from TS client?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -8,6 +8,7 @@
     private String url;
     private Transport transport;
     private Logger logger;
+    private HttpClient client;
```
mikaelm12: Maybe add another comment here that this is only for testing

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/DefaultHttpClient.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            @Override
+            public void onResponse(Call call, Response response) throws IOException {
+                try (ResponseBody body = response.body()) {
+                    HttpResponse httpResponse = new HttpResponse(response.code(), response.message(), body.string());
+                    responseFuture.complete(httpResponse);
+                }
+            }
+
```
mikaelm12: Remove newline

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HttpClient.java
```diff
@@ -0,0 +1,92 @@
...
+        this.statusText = statusText;
+        this.content = content;
+    }
+
+    public String getContent() {
+        return content;
+    }
+}
+
+abstract class HttpClient {
```
mikaelm12: IMO The HttpClient should be in it's own file. I'd be fine with keeping the request and response together though. Curious to hear other thoughts though.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -113,64 +124,35 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
         }
     }
 
-    private NegotiateResponse handleNegotiate() throws IOException {
-        accessToken = (negotiateResponse == null) ? null : negotiateResponse.getAccessToken();
-        negotiateResponse = Negotiate.processNegotiate(url, accessToken);
+    private CompletableFuture<NegotiateResponse> handleNegotiate() throws IOException, InterruptedException, ExecutionException {
+        HttpRequest request = new HttpRequest();
+        request.headers = this.headers;
```
mikaelm12: setter?

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -841,24 +842,43 @@ public void callingStartOnStartedHubConnectionNoOps() throws Exception {
...
-        HubConnection hubConnection = new HubConnection("http://example.com", mockTransport);
+        HubConnection hubConnection = new HubConnection("http://example.com", mockTransport, true);
         assertEquals(HubConnectionState.DISCONNECTED, hubConnection.getConnectionState());
 
         Throwable exception = assertThrows(HubException.class, () -> hubConnection.send("inc"));
         assertEquals("The 'send' method cannot be called if the connection is not active", exception.getMessage());
     }
 
+    @Test
+    public void negotiateSentOnStart() throws Exception {
```
mikaelm12: This is a good start. We're also going to want to add a test that verifies the redirect behavior.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -30,10 +31,16 @@ public HubConnectionBuilder configureLogging(Logger logger) {
         return this;
     }
 
+    // For testing purposes only
+    HubConnectionBuilder configureHttpClient(HttpClient client) {
```
mikaelm12: Why not just use the ctor you added?

BrennanConroy: Because I'm trying to steer the tests towards https://github.com/aspnet/SignalR/blob/release/2.2/test/Microsoft.AspNetCore.SignalR.Client.Tests/HubConnectionTests.Helpers.cs#L10

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -187,42 +169,82 @@ public HubConnectionState getConnectionState() {
      *
      * @throws Exception An error occurred while connecting.
      */
-    public CompletableFuture start() throws Exception {
+    public CompletableFuture<Void> start() throws Exception {
```
davidfowl: Why?

BrennanConroy: "CompletableFuture is a raw type. References to generic type CompletableFuture<T> should be parameterized"
This is what VSCode tells me with a green squiggle, so changing to `<Void>` gets rid of it

anurse: It's general cleanliness. I believe `CompletableFuture` is a shorthand for `CompletableFuture<?>` (remember Java doesn't allow multiple types with the same name but different generic parameters) which allows `start` to return **any** kind of `CompletableFuture`. This makes the type-checker ensure you return no value from your future.

anurse:
Depending on when https://github.com/aspnet/SignalR/pull/3012 merges, we may need to include WebSockets in this abstraction.

BrennanConroy:
🆙 📅 With websockets stuff

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -877,4 +880,84 @@ public void errorWhenReceivingInvokeWithIncorrectArgumentLength() throws Excepti
...
+            hubConnection.start().get(1000, TimeUnit.MILLISECONDS);
+        } catch(Exception ex) {}
+
+        List<HttpRequest> sentRequests = client.getSentRequests();
+        assertEquals(1, sentRequests.size());
+        assertEquals("http://example.com/negotiate", sentRequests.get(0).getUrl());
+    }
+
+    @Test
+    public void negotiateThatRedirectsForeverFailsAfter100Tries() throws InterruptedException, TimeoutException, Exception {
```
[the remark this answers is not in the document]
BrennanConroy: Test added https://github.com/aspnet/SignalR/pull/3011/files#diff-5ec328a725285659deb33f61428b2a70R951

BrennanConroy:

Issue #2792, opened 2018-08-11
Look into using a different websockets library for the Java client


mikaelm12:
We should look into switching to a Websockets library with a better async API.

anurse:
Might be necessary for #2694

mikaelm12:
Look at this: https://medium.com/square-corner-blog/web-sockets-now-shipping-in-okhttp-3-5-463a9eec82d1

anurse:
This also somewhat solves the problem with `.send` being blocking. OkHttp uses a background thread to do sending so `WebSocket.send` just queues the data to be sent and should return immediately. Because of how pipelines works that's not far from what .NET does. The difference is that the .NET logic will apply backpressure but the Java client won't be able to do that, we'll just keep accepting `.send`s. 🤷‍♂️

PR #3012, opened 2018-09-26
Change websockets library

mikaelm12:
Still early and hacky but just want to leave this out here.Started hacking on changing the websocket library to Square's OkHttp WebSocket implementation. 
The motivation for doing this now is that we aren't keeping track of cookies which is causing problems and this library provides an easier way to do that. Still need to verify that behavior

One issue I have now is I'm a little unsure how to deal with errors during connection with these websockets. 

Issue: https://github.com/aspnet/SignalR/issues/2792
Issue: https://github.com/aspnet/SignalR/issues/3013

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -3,8 +3,6 @@
 
 package com.microsoft.aspnet.signalr;
 
-import okhttp3.OkHttpClient;
-
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -15,6 +13,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
+
+import okhttp3.OkHttpClient;
 
 public class HubConnection {
     private String url;
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -85,7 +85,7 @@ public class HubConnection {
                     public List<Cookie> loadForRequest(HttpUrl url) {
                         List<Cookie> matchedCookies = new ArrayList<>();
                         List<Cookie> expiredCookies = new ArrayList<>();
-                        for (Cookie cookie: cookieList) {
+                        for (Cookie cookie : cookieList) {
                             if (cookie.expiresAt() < System.currentTimeMillis()) {
                                 expiredCookies.add(cookie);
                             } else if (cookie.matches(url)) {
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -75,7 +75,7 @@ public class HubConnection {
                                     break;
                                 }
                             }
-                            if(!replacedCookie) {
+                            if (!replacedCookie) {
                                 cookieList.add(cookie);
                             }
                         }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -4,10 +4,7 @@
 package com.microsoft.aspnet.signalr;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
@@ -60,26 +57,42 @@ public class HubConnection {
         }
 
         this.skipNegotiate = skipNegotiate;
+
         this.httpClient = new OkHttpClient.Builder()
                 .cookieJar(new CookieJar() {
-                    private final HashMap<String, List<Cookie>> cookieStore = new HashMap<>();
+                    private List<Cookie> cookieList = new ArrayList<>();
 
                     @Override
                     public void saveFromResponse(HttpUrl url, List<Cookie> cookies) {
-                        if(cookieStore.containsKey(url.host())) {
-                            cookieStore.get(url.host()).addAll(cookies);
-                        } else {
-                            cookieStore.put(url.host(), cookies);
+                        for (Cookie cookie : cookies) {
+                            for (Cookie innerCookie : cookieList) {
+                                if (cookie.name().equals(innerCookie.name()) && innerCookie.matches(url)) {
+                                    // We have a new cookie that matches an older one so we replace the older one.
+                                    cookieList.remove(innerCookie);
+                                }
+                            }
+                            cookieList.add(cookie);
                         }
                     }
 
                     @Override
                     public List<Cookie> loadForRequest(HttpUrl url) {
-                        List<Cookie> cookies = cookieStore.get(url.host());
-                        return cookies != null ? cookies : new ArrayList<>();
+                        List<Cookie> matchedCookies = new ArrayList<>();
+                        List<Cookie> expiredCookies = new ArrayList<>();
+                        for(Cookie cookie: cookieList) {
+                            if (cookie.expiresAt() < System.currentTimeMillis()) {
+                                expiredCookies.add(cookie);
+                            } else if (cookie.matches(url)) {
+                                matchedCookies.add(cookie);
+                            }
+                        }
+
+                        cookieList.removeAll(expiredCookies);
+                        return matchedCookies;
                     }
                 })
                 .build();
+
         this.callback = (payload) -> {
 
             if (!handshakeReceived) {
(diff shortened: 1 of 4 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -45,6 +51,26 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
             this.logger = new NullLogger();
         }
         this.skipNegotiate = skipNegotiate;
+        this.httpClient = new OkHttpClient.Builder()
+                .cookieJar(new CookieJar() {
```
BrennanConroy: @Tratcher Could you take a look at this cookie storage and see if there are any "gotchas" we should be aware of.

Tratcher: Yeah, don't write this code. Find a real implementation.

BrennanConroy: We had a nice chat about this. We'll be using the `Cookie` class from OkHttp and using `matches` and `expiresAt` to return the cookies.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -8,15 +8,17 @@
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
-import org.java_websocket.client.WebSocketClient;
-import org.java_websocket.handshake.ServerHandshake;
+import okhttp3.*;
 
 class WebSocketTransport implements Transport {
-    private WebSocketClient webSocketClient;
+    private WebSocket newWebSocketClient;
```
BrennanConroy: `websocket` or `websocketClient`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -8,15 +8,17 @@
...
+    private SignalRWebSocketListener  webSocketListener = new SignalRWebSocketListener();
```
BrennanConroy: nit: extra space

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -82,36 +79,55 @@ public void onReceive(String message) throws Exception {
     @Override
     public CompletableFuture stop() {
         return CompletableFuture.runAsync(() -> {
-            webSocketClient.closeConnection(0, "HubConnection Stopped");
+            newWebSocketClient.close(0, "HubConnection stopped.");
```
BrennanConroy: `(1000, empty string)`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -82,36 +79,55 @@ public void onReceive(String message) throws Exception {
...
+        public void onClosing(WebSocket webSocket, int code, String reason) {
+            logger.log(LogLevel.Information, "WebSocket connection stopping with " +
+                    "code %d and reason %d", code, reason);
+        }
+
+        @Override
+        public void onFailure(WebSocket webSocket, Throwable t, Response response) {
+            logger.log(LogLevel.Error, "Error : %d", t.getMessage());
+            // If the start future hasn't completed yet, then we need to complete it exceptionally.
+            if (!startFuture.isDone()) {
```
BrennanConroy: Do the same in `onClosing`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -54,6 +57,42 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
...
+                    private List<Cookie> cookieList = new ArrayList<>();
+
+                    @Override
+                    public void saveFromResponse(HttpUrl url, List<Cookie> cookies) {
+                        for (Cookie cookie : cookies) {
+                            for (Cookie innerCookie : cookieList) {
+                                if (cookie.name().equals(innerCookie.name()) && innerCookie.matches(url)) {
+                                    // We have a new cookie that matches an older one so we replace the older one.
+                                    cookieList.remove(innerCookie);
```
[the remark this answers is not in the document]
BrennanConroy: I meant `break;` not `continue;`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -82,36 +79,62 @@ public void onReceive(String message) throws Exception {
...
+                onReceive(message);
+            } catch (Exception e) {
+                e.printStackTrace();
             }
-        };
+        }
+
+        @Override
+        public void onClosing(WebSocket webSocket, int code, String reason) {
+            logger.log(LogLevel.Information, "WebSocket connection stopping with " +
```
BrennanConroy: `%s` for reason

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -8,15 +8,17 @@
...
 import java.util.concurrent.CompletableFuture;
 
-import org.java_websocket.client.WebSocketClient;
-import org.java_websocket.handshake.ServerHandshake;
+import okhttp3.*;
 
 class WebSocketTransport implements Transport {
-    private WebSocketClient webSocketClient;
+    private WebSocket websocketClient;
+    private SignalRWebSocketListener webSocketListener = new SignalRWebSocketListener();
```
BrennanConroy: nit: this doesn't need to be stored

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -54,6 +57,48 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
...
+                                cookieList.add(cookie);
+                            }
+                        }
+                    }
+
+                    @Override
+                    public List<Cookie> loadForRequest(HttpUrl url) {
+                        List<Cookie> matchedCookies = new ArrayList<>();
+                        List<Cookie> expiredCookies = new ArrayList<>();
+                        for (Cookie cookie: cookieList) {
```
BrennanConroy: nit: space after `cookie`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -54,6 +57,48 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
...
+                            if (cookie.expiresAt() < System.currentTimeMillis()) {
+                                expiredCookies.add(cookie);
+                            } else if (cookie.matches(url)) {
+                                matchedCookies.add(cookie);
+                            }
+                        }
+
+                        cookieList.removeAll(expiredCookies);
```
Tratcher: Do you have to worry about any concurrency here? Will there ever be overlapping requests?

BrennanConroy: Every implementation I've found looks like it assumes no overlaps, including one by one of the maintainers of the library https://gist.github.com/swankjesse/e283f2ddf264abd9a0d45bc80e7ec2d8

Tratcher: For your signalr scenarios is that a valid assumption? Or is concurrency managed by the calling httpclient? Java List certainly isn't thread safe.

BrennanConroy: We don't guarantee requests are called in sequence. We'll give this a look and see if we need a lock.

BrennanConroy: From some simple testing it looks like we need a lock

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -54,6 +57,61 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
...
+                                if (cookie.expiresAt() < System.currentTimeMillis()) {
+                                    expiredCookies.add(cookie);
+                                } else if (cookie.matches(url)) {
+                                    matchedCookies.add(cookie);
+                                }
+                            }
+
+                            cookieList.removeAll(expiredCookies);
+                            return matchedCookies;
+
```
BrennanConroy: nit: newline

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -54,6 +57,61 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
...
+                                    }
+                                }
+                                if (!replacedCookie) {
+                                    cookieList.add(cookie);
+                                }
+                            }
+                        } finally {
+                            cookieLock.unlock();
+                        }
+
```
BrennanConroy: nit: newline

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
muratg: Also, why is concat necessary?

mikaelm12: Just for line length.

Issue #3025, opened 2018-09-28
[Java] Add pings and timeouts


BrennanConroy:
Currently we don't close the connection if we fail to receive a message from the server in a specific time interval.

And we don't send pings to the server at regular intervals.

PR #3027, opened 2018-09-28
[Java] Add pings and server timeout

BrennanConroy:
https://github.com/aspnet/SignalR/issues/3025

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -45,25 +45,25 @@ public class HubConnection {
     private Timer pingTimer = null;
     private long nextServerTimeout = 0;
     private long nextPingActivation = 0;
-    private long keepAliveInterval = 15000;
-    private long serverTimeout = 30000;
+    private Duration keepAliveInterval = Duration.ofSeconds(15);
+    private Duration serverTimeout = Duration.ofSeconds(30);
     private long tickRate = 1000;
     private CompletableFuture<Void> handshakeResponseFuture;
     private Duration handshakeResponseTimeout = Duration.ofSeconds(15);
 
-    public void setServerTimeoutInMilliseconds(long serverTimeout) {
+    public void setServerTimeout(Duration serverTimeout) {
         this.serverTimeout = serverTimeout;
     }
 
-    public long getServerTimeoutInMilliseconds() {
+    public Duration getServerTimeout() {
         return this.serverTimeout;
     }
 
-    public void setKeepAliveIntervalInMilliseconds(long keepAliveInterval) {
+    public void setKeepAliveInterval(Duration keepAliveInterval) {
         this.keepAliveInterval = keepAliveInterval;
     }
 
-    public long getKeepAliveIntervalInMilliseconds() {
+    public Duration getKeepAliveInterval() {
         return this.keepAliveInterval;
     }
 
@@ -463,11 +463,11 @@ public class HubConnection {
     }
 
     private void resetServerTimeout() {
-        this.nextServerTimeout = System.currentTimeMillis() + serverTimeout;
+        this.nextServerTimeout = System.currentTimeMillis() + serverTimeout.toMillis();
     }
 
     private void resetKeepAlive() {
-        this.nextPingActivation = System.currentTimeMillis() + keepAliveInterval;
+        this.nextPingActivation = System.currentTimeMillis() + keepAliveInterval.toMillis();
     }
 
     /**
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -14,6 +14,7 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
@@ -43,8 +44,8 @@ public class HubConnection {
     private HttpClient httpClient;
     private String stopError;
     private Timer pingTimer = null;
-    private long nextServerTimeout = 0;
-    private long nextPingActivation = 0;
+    private AtomicLong nextServerTimeout = new AtomicLong();
+    private AtomicLong nextPingActivation = new AtomicLong();
     private Duration keepAliveInterval = Duration.ofSeconds(15);
     private Duration serverTimeout = Duration.ofSeconds(30);
     private long tickRate = 1000;
@@ -277,19 +278,25 @@ public class HubConnection {
                             connectionState = new ConnectionState(this);
                             logger.log(LogLevel.Information, "HubConnection started.");
 
-                        resetServerTimeout();
-                        this.pingTimer = new Timer();
-                        this.pingTimer.schedule(new TimerTask() {
-                            @Override
-                            public void run() {
-                                try {
-                                    if (System.currentTimeMillis() > nextServerTimeout) {
-                                        stop("Server timeout elapsed without receiving a message from the server.");
-                                        return;
-                                    }
+                            resetServerTimeout();
+                            this.pingTimer = new Timer();
+                            this.pingTimer.schedule(new TimerTask() {
+                                @Override
+                                public void run() {
+                                    try {
+                                        if (System.currentTimeMillis() > nextServerTimeout.get()) {
+                                            stop("Server timeout elapsed without receiving a message from the server.");
+                                            return;
+                                        }
 
-                                    if (System.currentTimeMillis() > nextPingActivation) {
-                                        sendHubMessage(PingMessage.getInstance());
+                                        if (System.currentTimeMillis() > nextPingActivation.get()) {
+                                            sendHubMessage(PingMessage.getInstance());
+                                        }
+                                    } catch (Exception e) {
+                                        logger.log(LogLevel.Warning, String.format("Error sending ping: %s", e.getMessage()));
+                                        // The connection is probably in a bad or closed state now, cleanup the timer so
+                                        // it stops triggering
+                                        pingTimer.cancel();
                                     }
                                 } catch (Exception e) {
                                     logger.log(LogLevel.Warning, String.format("Error sending ping: %s", e.getMessage()));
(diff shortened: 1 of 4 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -31,6 +34,33 @@
...
     private Map<String, String> headers = new HashMap<>();
     private ConnectionState connectionState = null;
+    private Timer pingTimer = null;
+    private long nextServerTimeout = 0;
+    private long nextPingActivation = 0;
+    private long keepAliveInterval = 15000;
+    private long serverTimeout = 30000;
+    private long tickRate = 1000;
+
+    public void setServerTimeoutInMilliseconds(long serverTimeout) {
```
mikaelm12: Can we move these getters and setters after the rest of the fields

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -289,15 +342,24 @@ public void send(String method, Object... args) throws Exception {
     }
 
     private void sendHubMessage(HubMessage message) throws Exception {
+        resetKeepAlive();
```
mikaelm12: nit: I think this should technically go after the call to send.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -185,9 +215,29 @@ public CompletableFuture start() throws Exception {
             return transport.send(handshake).thenRun(() -> {
                 hubConnectionStateLock.lock();
                 try {
+                    resetServerTimeout();
```
mikaelm12: I don't believe we should be resetting the server timeout here since we aren't receiving a message

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -185,9 +215,29 @@ public CompletableFuture start() throws Exception {
...
                     hubConnectionState = HubConnectionState.CONNECTED;
                     connectionState = new ConnectionState(this);
                     logger.log(LogLevel.Information, "HubConnected started.");
+
+                    this.pingTimer = new Timer();
+                    this.pingTimer.schedule(new TimerTask() {
+                        @Override
+                        public void run() {
+                            try {
+                                if (nextServerTimeout < System.currentTimeMillis()) {
```
mikaelm12: Super nit: I feel like this reads much better if you write it as  `System.currentTimeMillis() > nextServerTimeout`.
The check below as well.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -185,9 +215,29 @@ public CompletableFuture start() throws Exception {
...
+                                    stop("Server timeout elapsed without receiving a message from the server.");
```
mikaelm12: Do we also want to include the actual server timeout duration like the .NET Client?
https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnection.cs#L932

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -244,6 +273,26 @@ public CompletableFuture start() throws Exception {
...
+                        @Override
+                        public void run() {
+                            try {
+                                if (System.currentTimeMillis() > nextServerTimeout) {
+                                    stop("Server timeout elapsed without receiving a message from the server.");
+                                    return;
+                                }
+
+                                if (System.currentTimeMillis() > nextPingActivation) {
+                                    sendHubMessage(PingMessage.getInstance());
```
davidfowl: I assume sendHubMessage is doing the appropriate locking?

BrennanConroy: What locking is that?

davidfowl: I'm asking. There's no chance of overlapping sends right?

davidfowl: So the queue in the websocket library is locked and the library itself is thread safe?

BrennanConroy: I have confirmed (by reviewing the library code) that the websocket library is thread safe (at least for sends).

See https://github.com/square/okhttp/blob/ceb42f132220c71c7ad9a2948b24ee4f9e7df8e0/okhttp/src/main/java/okhttp3/internal/ws/RealWebSocket.java#L383

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -217,6 +247,26 @@ public HubConnectionState getConnectionState() {
...
+                                        if (System.currentTimeMillis() > nextServerTimeout) {
+                                            stop("Server timeout elapsed without receiving a message from the server.");
+                                            return;
+                                        }
+
+                                        if (System.currentTimeMillis() > nextPingActivation) {
+                                            sendHubMessage(PingMessage.getInstance());
+                                        }
+                                    } catch (Exception e) {
+                                        pingTimer.cancel();
```
davidfowl: what does this cancel?

davidfowl: We should log something here.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -32,10 +35,37 @@
...
 
+    public void setServerTimeoutInMilliseconds(long serverTimeout) {
+        this.serverTimeout = serverTimeout;
+    }
+
+    public long getServerTimeoutInMilliseconds() {
+        return this.serverTimeout;
+    }
+
+    public void setKeepAliveIntervalInMilliseconds(long keepAliveInterval) {
```
anurse: Can we use [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) so that we don't have to specify the units in the name? If we have to use `long`s, then I agree that the units should be here but I'd like to avoid `camelCaseSoupWithABunchOfWordsInIt`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -33,10 +37,37 @@
     private Map<String, String> headers = new HashMap<>();
     private ConnectionState connectionState = null;
     private HttpClient httpClient;
+    private Timer pingTimer = null;
+    private long nextServerTimeout = 0;
```
anurse: > Reads and writes are atomic for reference variables and for most primitive variables (all types **except long** and double). - https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html

You'll need to use [`AtomicLong`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicLong.html)

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: We do log the `Error sending ping` which uniquely indicates that the ping timer was cancelled.

BrennanConroy: Added the log after this comment ;)

BrennanConroy:
:up: 📅

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -38,6 +43,33 @@
...
     private HttpClient httpClient;
     private String stopError;
+    private Timer pingTimer = null;
+    private AtomicLong nextServerTimeout = new AtomicLong();
+    private AtomicLong nextPingActivation = new AtomicLong();
+    private Duration keepAliveInterval = Duration.ofSeconds(15);
+    private Duration serverTimeout = Duration.ofSeconds(30);
+    private Duration tickRate = Duration.ofSeconds(1);
+
+    public void setServerTimeout(Duration serverTimeout) {
```
mikaelm12: Are these things we want to make directly settable from the `HubConnection`? I guess we do it in the .NET client

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -371,11 +427,21 @@ public void send(String method, Object... args) throws Exception {
     private void sendHubMessage(HubMessage message) throws Exception {
         String serializedMessage = protocol.writeMessage(message);
         if (message.getMessageType() == HubMessageType.INVOCATION) {
-            logger.log(LogLevel.Debug, "Sending %d message '%s'.", message.getMessageType().value, ((InvocationMessage)message).getInvocationId());
+            logger.log(LogLevel.Debug, "Sending %s message '%s'.", message.getMessageType().name(), ((InvocationMessage)message).getInvocationId());
         } else {
-            logger.log(LogLevel.Debug, "Sending %d message.", message.getMessageType().value);
+            logger.log(LogLevel.Debug, "Sending %s message.", message.getMessageType().name());
```
mikaelm12: 👍

clients/java/signalr/src/test/java/com/microsoft/signalr/MockTransport.java
```diff
@@ -8,10 +8,21 @@
...
     private ArrayList<String> sentMessages = new ArrayList<>();
     private String url;
     private Consumer<String> onClose;
+    private boolean ignorePings;
+
+    public MockTransport() {
+        this(false);
+    }
+
+    public MockTransport(boolean ignorePings) {
```
mikaelm12: Yeah, this is gonna conflict with the autoHandshake ctor that I added. We can add another ctor that takes two bool args but that might look weird.

mikaelm12: Could just add a setter as well

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
anurse: Yeah, they were put on the HubConnection because they can be set during the connection (in theory) and we didn't want to overload the initial options.

BrennanConroy:

BrennanConroy:

PR #3029, opened 2018-09-28
Make Java client more reusable

BrennanConroy:


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -199,12 +203,12 @@ public CompletableFuture start() throws Exception {
     /**
      * Stops a connection to the server.
      */
-    private void stop(String errorMessage) {
+    private CompletableFuture<Void> stop(String errorMessage) {
```
mikaelm12: Should this happen in a `runAsync` call so that users could fire and forget the completion of what ever onClosed callbacks they have?

BrennanConroy: Mostly ignore this part as I changed it again in the HttpClient change

mikaelm12: Ah, okay. Yeah, I'm going over that now

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -182,17 +170,15 @@ public HubConnectionState getConnectionState() {
...
-            negotiate = CompletableFuture.completedFuture(null);
+            negotiate = CompletableFuture.completedFuture(baseUrl);
         }
 
-        return negotiate.thenCompose((response) -> {
-            // If we didn't skip negotiate and got a null response then exit start because we
-            // are probably disconnected
-            if (response == null && !skipNegotiate) {
+        return negotiate.thenCompose((url) -> {
+            if (url == null) {
```
mikaelm12: Should this case be handled exceptionally?

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -966,4 +966,42 @@ public void negotiateRedirectIsFollowed()
         assertEquals(HubConnectionState.CONNECTED, hubConnection.getConnectionState());
         hubConnection.stop();
     }
+
+    @Test
+    public void hubConnectionCanBeStartedAfterBeingStopped()
+            throws InterruptedException, ExecutionException, TimeoutException, Exception {
```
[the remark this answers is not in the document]
BrennanConroy: Yeah, for some reason "ctrl-." added all of them

PR #3032, opened 2018-09-28
Update Console Logger

mikaelm12:
Small thing I noticed in the ConsoleLogger implementation. The formatted string needs a newline


mikaelm12:
Derp. I saw a weird log message so I looked at the logger implementation and changed this without thinking

Issue #2976, opened 2018-09-18
Should we make HubException in the Java Client a Runtime Exception


mikaelm12:
Should we change HubException in the java client to derive from Runtime Exception. 
The main reason being 
>RuntimeException and its subclasses are unchecked exceptions. Unchecked exceptions do not need to be declared in a method or constructor's throws clause if they can be thrown by the execution of the method or constructor and propagate outside the method or constructor boundary.

[Runtime Exception Docs](https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html)

anurse:
We should get this figured out ASAP. Probably as part of the review with some other Java folks.

mikaelm12:
Looks like we wont be going this direction

PR #3043, opened 2018-10-01
Update HubException to be a RuntimeException

mikaelm12:
Making `HubException `derive from `RuntimeException `
Issue: https://github.com/aspnet/SignalR/issues/2976


mikaelm12:
Looks like we wont be going this direction

PR #3044, opened 2018-10-02
[Java] Clean up WebSockets Transport

mikaelm12:
We updated the Webosckets transport to use a string to internally represent the url that we use to create the underlying transport. There was some clean up that needed to happen.


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -198,11 +197,7 @@ public HubConnectionState getConnectionState() {
 
             logger.log(LogLevel.Debug, "Starting HubConnection");
             if (transport == null) {
-                try {
                     transport = new WebSocketTransport(url, headers, httpClient, logger);
```
BrennanConroy: fix indent

PR #3046, opened 2018-10-02
[Java] Add accessTokenProvider

BrennanConroy:


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -143,7 +149,15 @@ public HubConnection(String url, Transport transport, Logger logger, boolean ski
             }
 
             if (negotiateResponse.getAccessToken() != null) {
-                this.headers.put("Authorization", "Bearer " + negotiateResponse.getAccessToken());
+                this.accessTokenProvider = () -> CompletableFuture.completedFuture(negotiateResponse.getAccessToken());
+                String token = "";
+                try {
+                    // We know the future is already completed in this case
+                    // It's fine to call get() on it.
+                    token = this.accessTokenProvider.get().get();
```
davidfowl: I'm missing something, the first get is like .invoke?

anurse: Correct. `accessTokenProvider` is a `Supplier<CompletableFuture<String>>` so you need to call [`Supplier.get()`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html#get--) to invoke the method and then [`CompletableFuture.get()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get--) to resolve the future.

davidfowl: Couldn't they have called it invoke,wtf

BrennanConroy: `Supplier.get()`
`Consumer.accept(...)`
`Function.apply(...)`

mikaelm12: The wording for each of the functional interfaces makes sense based off of the param/return signatures for each. But still a little weird

BrennanConroy:

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -28,42 +29,47 @@
...
-    private String accessToken;
+    private Supplier<CompletableFuture<String>> accessTokenProvider;
     private Map<String, String> headers = new HashMap<>();
     private ConnectionState connectionState = null;
     private HttpClient httpClient;
 
     private static ArrayList<Class<?>> emptyArray = new ArrayList<>();
     private static int MAX_NEGOTIATE_ATTEMPTS = 100;
 
-    public HubConnection(String url, Transport transport, Logger logger, boolean skipNegotiate, HttpClient client) {
```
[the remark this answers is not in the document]
BrennanConroy: It's easy to add it back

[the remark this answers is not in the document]
mikaelm12: Unnecessarily removing it and then adding it back later seems like a waste. Should have at least discussed it more before just merging

PR #3051, opened 2018-10-02
[Java] Pass Url to transport start so we can reuse the transport

BrennanConroy:
I think this finishes client re-usability.

### DIFF
--- a/clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/WebSocketTransportTest.java
+++ b/clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/WebSocketTransportTest.java
@@ -14,8 +14,8 @@ import org.junit.jupiter.api.Test;
 class WebSocketTransportTest {
     @Test
     public void WebsocketThrowsIfItCantConnect() throws Exception {
-        Transport transport = new WebSocketTransport("http://www.notarealurl12345.fake", new HashMap<>(), new DefaultHttpClient(new NullLogger()), new NullLogger());
-        ExecutionException exception = assertThrows(ExecutionException.class, () -> transport.start().get(1, TimeUnit.SECONDS));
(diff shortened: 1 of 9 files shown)


clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/WebSocketTransportTest.java
```diff
@@ -14,8 +14,8 @@
 class WebSocketTransportTest {
     @Test
     public void WebsocketThrowsIfItCantConnect() throws Exception {
-        Transport transport = new WebSocketTransport("http://www.notarealurl12345.fake", new HashMap<>(), new DefaultHttpClient(new NullLogger()), new NullLogger());
-        ExecutionException exception = assertThrows(ExecutionException.class, () -> transport.start().get(1, TimeUnit.SECONDS));
+        Transport transport = new WebSocketTransport(new HashMap<>(), new DefaultHttpClient(new NullLogger()), new NullLogger());
+        ExecutionException exception = assertThrows(ExecutionException.class, () -> transport.start("http://www.notarealurl12345.fake").get(1, TimeUnit.SECONDS));
```
anurse: nit: this should always have just used `http://example.com/not-a-real-websocket` (or similar) since `example.com` is a reserved domain name for exactly this purpose. Not that I think it's likely for both a `.fake` TLD and `notarealurl12345.fake` to be registered, but...

BrennanConroy: @anurse Changing it makes the test flaky now?
```
com.microsoft.aspnet.signalr.WebSocketTransportTest > WebsocketThrowsIfItCantConnect() FAILED
2018-10-03T17:05:06.5667220Z       org.opentest4j.AssertionFailedError at WebSocketTransportTest.java:18
2018-10-03T17:05:06.5667400Z           Caused by: java.util.concurrent.TimeoutException at WebSocketTransportTest.java:18
```

Issue #3038, opened 2018-10-01
[Java] Transport needs to have a callback to the connection to stop the connection on failure


BrennanConroy:
If the transport fails for any reason the `HubConnection` will still be "connected" and all sends will fail.
We should be passing a callback to the transport so it can trigger a disconnect at the `HubConnection` layer if the transport is closed from failures etc.

See https://github.com/aspnet/SignalR/blob/release/2.2/clients/ts/signalr/src/HttpConnection.ts#L208

anurse:
Fixed in #3054

PR #3054, opened 2018-10-03
Transport stop triggers HubConnection stop

mikaelm12:
If there's an issue with the transport and it closes, the HubConnection should transition into the disconnected state. The stop/onClose logic in Websockets and HubConnection is now a lot closer to what we have in the TS client.
Issue:https://github.com/aspnet/SignalR/issues/3038

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -304,8 +304,6 @@ public class HubConnection {
         HubException hubException = null;
         hubConnectionStateLock.lock();
         try {
-            hubConnectionState = HubConnectionState.DISCONNECTED;
-
             // errorMessage gets passed in from the transport. An already existing stopError value
             // should take precedence.
             if (stopError != null) {
@@ -318,6 +316,7 @@ public class HubConnection {
             connectionState.cancelOutstandingInvocations(hubException);
             connectionState = null;
             logger.log(LogLevel.Information, "HubConnection stopped.");
+            hubConnectionState = HubConnectionState.DISCONNECTED;
         } finally {
             hubConnectionStateLock.unlock();
         }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -301,7 +301,7 @@ public class HubConnection {
     }
 
     private void stopConnection(String errorMessage) {
-        HubException hubException = null;
+        RuntimeException exception = null;
         hubConnectionStateLock.lock();
         try {
             // errorMessage gets passed in from the transport. An already existing stopError value
@@ -310,10 +310,10 @@ public class HubConnection {
                 errorMessage = stopError;
             }
             if (errorMessage != null) {
-                hubException = new HubException(errorMessage);
+                exception = new RuntimeException(errorMessage);
                 logger.log(LogLevel.Error, "HubConnection disconnected with an error %s.", errorMessage);
             }
-            connectionState.cancelOutstandingInvocations(hubException);
+            connectionState.cancelOutstandingInvocations(exception);
             connectionState = null;
             logger.log(LogLevel.Information, "HubConnection stopped.");
             hubConnectionState = HubConnectionState.DISCONNECTED;
@@ -324,7 +324,7 @@ public class HubConnection {
         // Do not run these callbacks inside the hubConnectionStateLock
         if (onClosedCallbackList != null) {
             for (Consumer<Exception> callback : onClosedCallbackList) {
-                callback.accept(hubException);
+                callback.accept(exception);
             }
         }
     }
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -195,7 +195,7 @@ public class HubConnection {
                     }
                 });
 
-        stopError = null;        
+        stopError = null;
         CompletableFuture<String> negotiate = null;
         if (!skipNegotiate) {
             negotiate = tokenFuture.thenCompose((v) -> startNegotiate(baseUrl, 0));
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -33,6 +33,7 @@ public class HubConnection {
     private Map<String, String> headers = new HashMap<>();
     private ConnectionState connectionState = null;
     private HttpClient httpClient;
+    private String stopError;
 
     private static ArrayList<Class<?>> emptyArray = new ArrayList<>();
     private static int MAX_NEGOTIATE_ATTEMPTS = 100;
@@ -194,6 +195,7 @@ public class HubConnection {
                     }
                 });
 
+        stopError = null;        
         CompletableFuture<String> negotiate = null;
         if (!skipNegotiate) {
             negotiate = tokenFuture.thenCompose((v) -> startNegotiate(baseUrl, 0));
@@ -208,6 +210,7 @@ public class HubConnection {
             }
 
             transport.setOnReceive(this.callback);
+            transport.setOnClose((message) -> stopConnection(message));
 
             try {
                 return transport.start(url).thenCompose((future) -> {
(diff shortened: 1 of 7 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -264,40 +266,16 @@ public HubConnectionState getConnectionState() {
                 return CompletableFuture.completedFuture(null);
             }
 
+            hubConnectionState = HubConnectionState.DISCONNECTED;
```
BrennanConroy: I know we do this here in the Typescript client, but I think we should remove it and just rely on the `stopConnection` callback setting this.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -307,6 +285,36 @@ public HubConnectionState getConnectionState() {
...
 
+    private void stopConnection(String errorMessage) {
+        HubException hubException = null;
+        hubConnectionStateLock.lock();
+        try {
+            hubConnectionState = HubConnectionState.DISCONNECTED;
+
+            // errorMessage gets passed in from the transport. An already existing stopError value
+            // should take precedence.
+            if (stopError != null) {
```
BrennanConroy: set `stopError` to null somewhere so if they reuse the connection and close it gracefully they don't get the error string

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -73,5 +80,6 @@ public void onReceive(String message) throws Exception {
     void onClose(int code, String reason) {
         logger.log(LogLevel.Information, "WebSocket connection stopping with " +
                 "code %d and reason '%s'.", code, reason);
+        onClose.accept(reason);
```
BrennanConroy: We should only pass a value here if `code != 1000`

BrennanConroy: It looks like if the WebSocket `onFailure` method is called then `onClosing` wont be called, so we might need to invoke the close callback on that path too.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -307,6 +285,36 @@ public HubConnectionState getConnectionState() {
         return stop(null);
     }
 
+    private void stopConnection(String errorMessage) {
+        HubException hubException = null;
+        hubConnectionStateLock.lock();
+        try {
+            hubConnectionState = HubConnectionState.DISCONNECTED;
```
BrennanConroy: move this to just above the `finally`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: Yeah, the logic we used in JavaScript is because of observed browser behavior. I'd rather this be consistent with .NET (which waits for the WebSocket to close) until we see a similar behavior that we need to work around.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
anurse: We should test that scenario

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
mikaelm12: This should happen earlier in the stopping logic rather than later

BrennanConroy: We should only set it once the connection is actually allowed to start again which would be right above the finally.

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -73,5 +85,12 @@ public void onReceive(String message) throws Exception {
     void onClose(int code, String reason) {
         logger.log(LogLevel.Information, "WebSocket connection stopping with " +
                 "code %d and reason '%s'.", code, reason);
+        if (code != 1000) {
+            onClose.accept(reason);
+        }
+        else {
+            onClose.accept(null);
+        }
+
```
BrennanConroy: nit: newline

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
```diff
@@ -45,7 +47,12 @@ private String formatUrl(String url) {
         logger.log(LogLevel.Debug, "Starting Websocket connection.");
         this.webSocketClient = client.createWebSocket(this.url, this.headers);
         this.webSocketClient.setOnReceive((message) -> onReceive(message));
-        this.webSocketClient.setOnClose((code, reason) -> onClose(code, reason));
+        this.webSocketClient.setOnClose((code, reason) -> {
+            if (onClose != null) {
```
BrennanConroy: This shouldn't be possible

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -29,6 +29,34 @@ public void checkHubConnectionState() throws Exception {
         assertEquals(HubConnectionState.DISCONNECTED, hubConnection.getConnectionState());
     }
 
+    @Test
+    public void transportCloseTriggersStopInHubConnection() throws Exception {
+        MockTransport mockTransport = new MockTransport();
+        HubConnection hubConnection = TestUtils.createHubConnection("http://example.com", mockTransport);
+        hubConnection.start();
```
BrennanConroy: Missing `.get()`s for all these

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -29,6 +29,34 @@ public void checkHubConnectionState() throws Exception {
...
+    }
+
+    @Test
+    public void transportCloseWithErrorTriggersStopInHubConnection() throws Exception {
+        MockTransport mockTransport = new MockTransport();
+        HubConnection hubConnection = TestUtils.createHubConnection("http://example.com", mockTransport);
+        String errorMessage = "Example transport error.";
+
+        hubConnection.onClosed((error) -> {
+            assertEquals(error.getMessage(), errorMessage);
```
BrennanConroy: Have to assert outside of the lambda otherwise we can't be sure this actually ran

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -29,6 +29,34 @@ public void checkHubConnectionState() throws Exception {
...
+        HubConnection hubConnection = TestUtils.createHubConnection("http://example.com", mockTransport);
+        hubConnection.start();
+        assertEquals(HubConnectionState.CONNECTED, hubConnection.getConnectionState());
+        mockTransport.stop();
+
+        assertEquals(HubConnectionState.DISCONNECTED, hubConnection.getConnectionState());
+    }
+
+    @Test
+    public void transportCloseWithErrorTriggersStopInHubConnection() throws Exception {
```
BrennanConroy: Can we write a test that has a transport failure during `hubConnection.start()` and make sure `hubConnection.start()` wont hang

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/WebSocketTransport.java
mikaelm12: For testing the transport without a HubConnection

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -319,6 +300,35 @@ public HubConnectionState getConnectionState() {
...
+        HubException hubException = null;
+        hubConnectionStateLock.lock();
+        try {
+            // errorMessage gets passed in from the transport. An already existing stopError value
+            // should take precedence.
+            if (stopError != null) {
+                errorMessage = stopError;
+            }
+            if (errorMessage != null) {
+                hubException = new HubException(errorMessage);
```
BrennanConroy: Per our offline discussions, `RuntimeException`

PR #3066, opened 2018-10-04
[WIP] Add LogMessageContent option

BrennanConroy:
TS Client has a `logMessageContent` option, we should add it to Java and .NET.

I want to solicit feedback on this especially for the .NET one, I only did it partially because it looks a little ugly.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.Log.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.Log.cs
@@ -70,11 +70,8 @@ namespace Microsoft.AspNetCore.Http.Connections.Client.Internal
             private static readonly Action<ILogger, Exception> _startedTransport =
                 LoggerMessage.Define(LogLevel.Debug, new EventId(19, "StartedTransport"), "Started transport.");
 
-            private static readonly Action<ILogger, WebSocketMessageType, int, bool, string, Exception> _messageReceivedString =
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.Log.cs
```diff
@@ -69,6 +70,12 @@ private static class Log
             private static readonly Action<ILogger, Exception> _startedTransport =
                 LoggerMessage.Define(LogLevel.Debug, new EventId(19, "StartedTransport"), "Started transport.");
 
+            private static readonly Action<ILogger, WebSocketMessageType, int, bool, string, Exception> _messageReceivedString =
+                LoggerMessage.Define<WebSocketMessageType, int, bool, string>(LogLevel.Debug, new EventId(20, "MessageReceivedString"), "Message received. Type: {MessageType}, size: {Count}, EndOfMessage: {EndOfMessage}. String data: {Message}.");
```
anurse: Couldn't this be one event with the "type" (text or binary) as just another parameter?

BrennanConroy: Yeah, I think when I first wrote it I was thinking I would print a byte[], but now they are both strings

muratg:
SignalR work has moved to aspnetcore repo, and this repo will be archived. If you're still interested in pursuing this PR, please move it over to aspnetcore. Thanks.

Issue #2110, opened 2018-04-20
Setting XMLHttpRequest "withCredentials" attribute


EvheniyHlushko:
After updating from "@aspnet/signalr-client" to "@aspnet/signalr" because "@aspnet/signalr-client" is deprecated.   

I have these errors:

` Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.`

`The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.`

How can I set `withCredentials` attribute?

anurse:
The error is occurring because we **already set** the `withCredentials` attribute. You need to adjust your CORS configuration to use specific Origin values, rather than a wildcard, because wildcards are not supported in CORS when using credentials.

blubiszirp:
Why do you set the `withCredentials` attribute? Is there any possibility to remove the attribute and use a wildcard?

davidfowl:
@blubiszirp because cookies don't flow without withCredentials and when running in environments that support stickyness via cookies, it required to function.

JanEggers:
im on the same team / problem as @blubiszirp 

we switched from

app.UseCors( builder => builder
                .AllowAnyHeader()
                .AllowAnyMethod()
                .AllowAnyOrigin()
            );

to 

app.UseCors( builder => builder
                .AllowAnyHeader()
                .AllowAnyMethod()
                .WithOrigins(corsOptions.Value.Origins)
            );

and now we are getting the error:
No 'Access-Control-Allow-Origin' header is present on the requested resource.

seems to be the same issue as #2095

gingters:
Now that's a bummer.

We plan to provide an open API with SignalR and a JS client lib that anybody can use on their website to connect to push notifications from our server. Therefore, it is impossible to know all possible origin values, as literally everybody should be able to use that on any website.

So, we must be able to use `Access-Control-Allow-Origin: *`.
If you need to support cookies, then at least make setting `withCredentials` configurable, so that we can turn it off.

An alternative would be to extend `.AllowAllOrigins()` configuration on aspnet core CORS to not send the wildcard, but reply with the value of the received origin header, so that this matches, but I strongly would prefer not sending `withCredentials` in the first place.

davidfowl:
[...]

PR #3067, opened 2018-10-04
Add options to disable Xhr requests using 'withCredentials'

FraserKillip:
Add options to disable Xhr requests using 'withCredentials'
 - Options class `IHttpClientOptions` added
 - Used only in XhrHttpClient

### DIFF
--- a/clients/ts/signalr/src/IHttpClientOptions.ts
+++ b/clients/ts/signalr/src/IHttpClientOptions.ts
@@ -1,5 +1,5 @@
 /** Options provided to {@link @aspnet/signalr.DefaultHttpClient} and {@link @aspnet/signalr.XhrHttpClient} to configure options for the HTTP-based transports. */
 export interface IHttpClientOptions {
     /** Disable withCredentials for the default {@link @aspnet/signalr.HttpClient} */
-    disableXhrWithCredentials?: boolean,
+    disableXhrWithCredentials?: boolean;
 }
--- a/clients/ts/signalr/src/IHttpClientOptions.ts
+++ b/clients/ts/signalr/src/IHttpClientOptions.ts
@@ -1,0 +1,5 @@
+/** Options provided to {@link @aspnet/signalr.DefaultHttpClient} and {@link @aspnet/signalr.XhrHttpClient} to configure options for the HTTP-based transports. */
+export interface IHttpClientOptions {
+    /** Disable withCredentials for the default {@link @aspnet/signalr.HttpClient} */
+    disableXhrWithCredentials?: boolean,
+}
--- a/clients/ts/signalr/src/XhrHttpClient.ts
+++ b/clients/ts/signalr/src/XhrHttpClient.ts
@@ -3,8 +3,8 @@
 
 import { AbortError, HttpError, TimeoutError } from "./Errors";
 import { HttpClient, HttpRequest, HttpResponse } from "./HttpClient";
-import { ILogger, LogLevel } from "./ILogger";
 import { IHttpClientOptions } from "./IHttpClientOptions";
+import { ILogger, LogLevel } from "./ILogger";
 
 export class XhrHttpClient extends HttpClient {
     private readonly logger: ILogger;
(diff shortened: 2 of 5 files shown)


FraserKillip:
I'm unsure what the rationale for having `withCredentials` hardcoded to `true` is, I initially tested it by providing my own `HttpClient` and it worked when I had `withCredentials` set to false.

Please let me know of any feedback you have.

anurse:
SignalR requires sticky sessions which is controlled via Cookies. Disabling `withCredentials` will break this functionality.

Why do you need to disable it? You should be able to set the `Access-Control-Allow-Credentials` on your server.

FraserKillip:
In order to allow credentials I can't have a wildcard origin. If `withCredentials` is required for the sticky sessions I can close this PR and work around it on my end or try and get my Cors to allow credentials. 

Thanks

anurse:
> In order to allow credentials I can't have a wildcard origin

What CORS implementation are you using? ASP.NET Core's built-in CORS implementation supports this by ending back the provided Origin rather than actually sending the wildcard.

anurse:
Btw, I'm not necessarily dismissing the change :). Just want to get an idea of the scenarios.

clients/ts/signalr/src/IHttpClientOptions.ts
```diff
@@ -0,0 +1,5 @@
+/** Options provided to {@link @aspnet/signalr.DefaultHttpClient} and {@link @aspnet/signalr.XhrHttpClient} to configure options for the HTTP-based transports. */
+export interface IHttpClientOptions {
+    /** Disable withCredentials for the default {@link @aspnet/signalr.HttpClient} */
+    disableXhrWithCredentials?: boolean,
```
anurse: Let's just call this `withCredentials` and treat undefined as `true`. Negative booleans are weird :)

clients/ts/signalr/src/XhrHttpClient.ts
```diff
@@ -31,7 +34,7 @@ export class XhrHttpClient extends HttpClient {
             const xhr = new XMLHttpRequest();
 
             xhr.open(request.method!, request.url!, true);
-            xhr.withCredentials = true;
+            xhr.withCredentials = !this.options.disableXhrWithCredentials;
```
anurse: This would be `typeof this.options.withCredentials === "undefined" || this.options.withCredentials` given my comment above.

clients/ts/signalr/src/IHttpClientOptions.ts
```diff
@@ -0,0 +1,5 @@
+/** Options provided to {@link @aspnet/signalr.DefaultHttpClient} and {@link @aspnet/signalr.XhrHttpClient} to configure options for the HTTP-based transports. */
+export interface IHttpClientOptions {
+    /** Disable withCredentials for the default {@link @aspnet/signalr.HttpClient} */
```
anurse: Let's expand a bit on what this means. Maybe something like

```
/** Specifies if credentials and cookies should be sent along with cross-domain requests. Only applies in browser environments. */
```

analogrelay (commented):
Looks pretty good. I have some comments about the variable pattern and comment text, but it seems good enough to take. I like that it's limited to people explicitly constructing the `DefaultHttpClient` since this is a very edge-case scenario (given that it breaks most sticky-session implementations).

JanEggers:
@FraserKillip did u try AllowCrentials in the server? I think this is just a dup of #2110

anurse:
Closing this PR as we haven't heard from you since our last set of feedback. If you're still interested in contributing this change, take a look at the feedback and reopen when you're able to address it and rebase on the latest version of the repo!

alexandrereyes:
[ASP.NET Core 2.2 doesn't allow AnyOrigin with credentials](https://docs.microsoft.com/pt-br/aspnet/core/migration/21-to-22?view=aspnetcore-2.2&tabs=visual-studio#update-cors-policy), so whats the workaround since there is no option to disable SignalR XMLHttpRequest withCredentials?

FraserKillip:
@alexandrereyes we had success with using [wildcard origins](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.cors.infrastructure.corspolicybuilder.setisoriginallowedtoallowwildcardsubdomains?view=aspnetcore-2.1)

alexandrereyes:
Tks @FraserKillip looks like this is a workaroud

```
app.UseCors(builder => builder
                .AllowAnyHeader()
                .AllowAnyMethod()
                .SetIsOriginAllowed((host) => true)
                .AllowCredentials()
            );
```

PR #3078, opened 2018-10-05
Java Client API Review pt2

mikaelm12:
Here are the API review changes from our meeting this week!

Thanks @JonathanGiles for the review!

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
@@ -3,28 +3,16 @@
 
 package com.microsoft.aspnet.signalr;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
 class CallbackMap {
-    private ConcurrentHashMap<String, List<InvocationHandler>> handlers = new ConcurrentHashMap<>();
+    private Map<String, List<InvocationHandler>> handlers = new ConcurrentHashMap<>();
 
-    public InvocationHandler put(String target, ActionBase action, ArrayList<Class<?>> classes) {
+    public InvocationHandler put(String target, ActionBase action, List<Class<?>> classes) {
         InvocationHandler handler = new InvocationHandler(action, Collections.unmodifiableList(classes));
-
-        handlers.computeIfPresent(target, (methodName, handlerList) -> {
-            handlerList.add(handler);
-            return handlerList;
-        });
-        handlers.computeIfAbsent(target, (ac) -> new ArrayList<>(Arrays.asList(handler)));
+        handlers.computeIfAbsent(target, (ac) -> new ArrayList<>()).add(handler);
         return handler;
-    }
-
-    public Boolean containsKey(String key) {
-        return handlers.containsKey(key);
     }
 
     public List<InvocationHandler> get(String key) {
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
@@ -9,9 +9,9 @@ import java.util.concurrent.ConcurrentHashMap;
 class CallbackMap {
     private Map<String, List<InvocationHandler>> handlers = new ConcurrentHashMap<>();
 
-    public InvocationHandler put(String target, ActionBase action, List<Class<?>> classes) {
-        InvocationHandler handler = new InvocationHandler(action, Collections.unmodifiableList(classes));
-        handlers.computeIfAbsent(target, (ac) -> new ArrayList<>()).add(handler);
+    public InvocationHandler put(String target, ActionBase action, Class<?>... classes) {
+        InvocationHandler handler = new InvocationHandler(action, classes);
+        handlers.computeIfAbsent(target, ac -> new ArrayList<>()).add(handler);
         return handler;
     }
 
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
@@ -11,7 +11,13 @@ class CallbackMap {
 
     public InvocationHandler put(String target, ActionBase action, Class<?>... classes) {
         InvocationHandler handler = new InvocationHandler(action, classes);
-        handlers.computeIfAbsent(target, ac -> new ArrayList<>()).add(handler);
+        handlers.compute(target, (key, value) -> {
+            if (value == null) {
+                value = new ArrayList<>();
+            }
+            value.add(handler);
+            return value;
+        });
         return handler;
     }
 
--- a/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
@@ -38,7 +38,7 @@ public class HubConnection {
     private HttpClient httpClient;
     private String stopError;
 
-    public HubConnection(String url, HttpConnectionOptions options) {
+    HubConnection(String url, HttpConnectionOptions options) {
         if (url == null || url.isEmpty()) {
             throw new IllegalArgumentException("A valid url is required.");
         }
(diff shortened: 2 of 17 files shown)


JonathanGiles:
This looks like most of my commit, so I'm happy! :-)

JonathanGiles:
Should the HubConnection ctor be made private as I proposed, or is that still TBD?

mikaelm12:
Ah yes. I'll change the ctor to be package private.

mikaelm12:
@dotnet-bot Windows Release Spanish Language x64 Build

mikaelm12:

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
```diff
@@ -3,30 +3,18 @@
...
-        InvocationHandler handler = new InvocationHandler(action, Collections.unmodifiableList(classes));
-
-        handlers.computeIfPresent(target, (methodName, handlerList) -> {
-            handlerList.add(handler);
-            return handlerList;
-        });
-        handlers.computeIfAbsent(target, (ac) -> new ArrayList<>(Arrays.asList(handler)));
+    public InvocationHandler put(String target, ActionBase action, Class<?>... classes) {
+        InvocationHandler handler = new InvocationHandler(action, classes);
+        handlers.computeIfAbsent(target, ac -> new ArrayList<>()).add(handler);
```
BrennanConroy: I don't think this is thread safe. You're calling add on a List

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -657,25 +605,20 @@ public Subscription on(String target, Action callback) {
...
-        classes.add(param7);
-        classes.add(param8);
-        InvocationHandler handler = handlers.put(target, action, classes);
-        logger.log(LogLevel.Trace, "Registering handler for client method: %s.", target);
+        return registerHandler(target, action, param1, param2, param3, param4, param5, param6, param7, param8);
+    }
+
+    private Subscription registerHandler(String target, ActionBase action, Class<?>... types) {
+        InvocationHandler handler = handlers.put(target, action, types);
+        logger.log(LogLevel.Debug, "Registering handler for client method: " + target);
```
BrennanConroy: Full stop

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
anurse: Do we support calling `.on` from multiple threads?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -179,10 +179,9 @@ public HubConnectionState getConnectionState() {
 
     /**
      * Starts a connection to the server.
-     *
-     * @throws Exception An error occurred while connecting.
+     * @return A completable future that represents starting.
```
anurse: nit: `A completable future that completes when the connection has been established`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -272,6 +267,8 @@ public HubConnectionState getConnectionState() {
 
     /**
      * Stops a connection to the server.
+     * @param errorMessage
+     * @return A completable future that represents stopping the {@link HubConnection}.
```
anurse: nit: `A completable future that completes when the connection has been stopped`

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -295,6 +292,7 @@ public HubConnectionState getConnectionState() {
 
     /**
      * Stops a connection to the server.
+     * @return A completable future that represents stopping the {@link HubConnection}.
```
anurse: nit: same message as above

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: nit: `errorMessage` should have a description, right?

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
anurse: .

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/CallbackMap.java
BrennanConroy: Technically yes, since we allow you to add or remove handlers at any time

anurse: I think we need to use [`compute`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#compute-K-java.util.function.BiFunction-) here. The `BiFunction` taken by compute gets the current value or `null` if there is no value, and it guarantees that the entire method is invoked atomically. So the code would be something like:

```java
handlers.compute((key, value) -> {
    if (value == null) {
        value = new ArrayList<>();
    }
    value.add(handler);
    return value;
});
```

That way we, in a single atomic invocation, initialize the list if there isn't one already and then add the item to the list. Since `compute` guarantees atomicity, other attempts to add to the list will be blocked while we initialize the list and then will see the updated list. At least that's how I understand the Javadocs.

mikaelm12:
>Is removing HubConnectionOptions coming in a different PR?

I was under the impression that we were keeping the HttpConnectionOptions? We discussed that the was some overlap but I thought ultimately we wanted to keep it. We also use it internally in the implementation of the HubConnection builder and it allows that implementation to be nicer.

anurse:
> I was under the impression that we were keeping the HttpConnectionOptions?

I was under the opposite impression :). I don't have a serious problem with either, but we need to have a clear decision and a single clean API here. The main negatives we discussed were that you have to create this object outside the fluent builder API and configure it. Whereas in .NET we use the `Action<HttpConnectionOptions>` pattern to make it clean.

We should **either**:

1. Remove it (or make it private if you still want to use it as the internal carrier for these options) and use `HubConnectionBuilder` entirely

2. Switch to take a `Consumer<HttpConnectionOptions>` and call that consumer, passing in an initialized options object. This would be more like the .NET API, and would mean you would do this:

```java
HubConnection connection = HubConnectionBuilder.create(url, (options) -> {
    options.addHeader("Foo", "Bar");
})
    .build();
```

or maybe this:

```java
HubConnection connection = HubConnectionBuilder.create(url)
    .withHttpOptions((options) -> {
        options.addHeader("Foo", "Bar");
    })
    .build();
```

anurse:
My thinking is that we either properly align with .NET (which uses `Action<HttpConnectionOptions>`) or we don't have the extra `HttpConnectionOptions` API in the way.

JonathanGiles:
In my opinion it shouldn't be necessary to have either `HttpConnectionOptions` or any allowance for it. It should just be part of the fluent interface exposed after the user creates the HubConnectionBuilder with the create method. I would suggest more consideration be given to the possibility that `HubConnectionBuilder.create(url)` return a subclass of the HubConnectionBuilder with additional API specific for creating http-specific hub connections.

mikaelm12:
Cool. I will make that change then. I'd like to do that in a different PR though since that will be a pretty large churn change (nearly every test we have on the hubconnection will be effected, along with where ever else we use HttpConnectionOptions internally).

This pr has  a lot of changes even with keeping the options :)

JonathanGiles:
FYI - my commit made all of these changes to the tests, so if you want you are free to copy/paste from there.

anurse:
I'm OK with doing this in a separate PR if it's easier.

mikaelm12:
Okay, I think this is good to go. Moving the removal of HttpConnectionOptions based on Jonathan's change to a follow up pr

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnection.java
```diff
@@ -272,6 +267,8 @@ public HubConnectionState getConnectionState() {
 
     /**
      * Stops a connection to the server.
+     * @param errorMessage An error message if the connected needs to be stopped because of an error.
```
anurse: This is private actually... doesn't technically need any Javadocs ;). It's fine to leave it, but I was about to nit pick the comment and realized it was private.

mikaelm12: Lol. Good point

Issue #3045, opened 2018-10-02
Feature Request - Add Invoke method to IClientProxy which can wait for a response from the client.


lianying:
Sometimes server side need wait for a response result from client invocation. I already tried to extend IClientProxy, but too many classes have to replace. DefaultHubDispatcher/HubConnectionContext... have many internal properties or classes.

anurse:
This is something we've talked about but it would be a significant work item. Also, it would *only* work when addressing a single client `Clients.Client(...)` since otherwise you'd have to wait for all clients to return data. I'm putting this in the backlog since I don't think we have an issue tracking this right now.

lianying:
Got it! Thanks for your reply! I’m trying to study this issue and I think it’s meaningful now.

davidfowl:
@anurse we do, this is basically a dupe of https://github.com/aspnet/SignalR/issues/903

anurse:
Agreed, but the other issue was a little unclear on that point. Clarified it.

Issue #903, opened 2017-09-18
Support Server -> Client ACKs for single clients


davidfowl:
Support server to clients acks so that reliable messaging can be implemented more easily. This would only be when using `Clients.Client()`. I think we should go back to the SendAsync (fire and forget) InvokeAsync (wait for ack) naming pattern. That's the one sticking point.

brettclutch:
Vote +1 since I was trying to implement exactly this and I was confused in the .net client the difference between connection.InvokeAsync() and connection.SendAsync(). Old SignalR docs don't shine any light and my google foo not returning anything but this reference.

davidfowl:
What exactly are you trying to do?

brettclutch:
Whoops on second visit maybe I misunderstood the thought here. 

In any case: I wanted to use the c# async/await construct to fire off a request message to the server from the .net client and "await' the server response. I was hoping I could use InvokeAsync to invoke a hub method which would have a return type once the server replies. 

* Dictionary of "PendingRequests" awaiting responses keyed by a GUID and paired to a TaskCompletionSource. 
* As SignalR.Client On<>() method is triggered by data from server I find the related GUID and return that response using the paired TaskCompletionSource return. 

Allows for IMHO nicer .net client code when trying to use SignalR in a Request/Response style. (Which I understand is just one style of SignalR usage and you want to stay agnostic/light) 

What the client code looks like being able to await a server response (much nicer, still don't like the cast at the end but still working on that) :)

` LoginRequest request = new LoginRequest();
            request.User = user;
            LoginResponse response = await Server.InvokeAsync(request) as LoginResponse;`

What the extension method looks like:
` public async Task<ServerResponse> InvokeAsync(ClientRequest request)
        {        
                var tcs = new TaskCompletionSource<ServerResponse>();
                request.Id = Guid.NewGuid().ToString();
                PendingResponse.Add(request.Id, tcs);
                await Connection.InvokeAsync(request);
                return await tcs.Task;
            }`

DOWN WITH REST :)

davidfowl:
@brettclutch 

> I was hoping I could use InvokeAsync to invoke a hub method which would have a return type once the server replies. But found that is not its use case and rolled my own version:

Can you clarify? This is from client to server or server to client? If it's client to server it's exactly the case.

If you are talking about the server awaiting a client response, then that's what this issue is about and it would only be possible when addressing a single client.

brettclutch:
This is client to server where client is awaiting a reply from the server.

[...]

PR #3086, opened 2018-10-08
Lay/server invoke

lianying:
Add server side invoke single client and wait for a result. #3045 

For client side: 
1、Make InvocatonHandler can return an object.
2、Send a CompetionMessage to server when an invocation hava non-null result.

For server side:
1、Add a state to HubConnectionContext which can store invocations.
2、Add CompetionMessage process to HubDispatcher.


davidfowl:
@lianying We don't have any plans to add this (especially not in 2.2). I'd recommend fleshing out the design on an issue first before jumping into a PR. Doing this is pretty invasive and needs to be planned and thought about in relation to other feature work. This also introduces breaking changes which we don't take lightly. 

As a result, I'm closing this PR. Lets discuss the implementation on https://github.com/aspnet/SignalR/issues/903

Issue #3064, opened 2018-10-04
[Java] Add Headers configurability


BrennanConroy:
In the .NET Client we expose a Headers collection in `HttpConnectionOptions` that we will send with every request. We currently don't have this exposed in Java.

PR #3090, opened 2018-10-08
[Java] Add ability to configure headers

mikaelm12:
Adding ability to configure headers in the Java client.
Fixes: https://github.com/aspnet/SignalR/issues/3064

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
@@ -4,6 +4,8 @@
 package com.microsoft.signalr;
 
 import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
 
 import io.reactivex.Single;
 
@@ -15,6 +17,7 @@ public class HttpHubConnectionBuilder {
     private boolean skipNegotiate;
     private Single<String> accessTokenProvider;
     private Duration handshakeResponseTimeout;
+    private Map<String, String> headers;
 
     HttpHubConnectionBuilder(String url) {
         this.url = url;
@@ -61,7 +64,20 @@ public class HttpHubConnectionBuilder {
         return this;
     }
 
+    public HttpHubConnectionBuilder withHeaders(Map<String, String> headers) {
+        this.headers = headers;
+        return this;
+    }
+
+    public HttpHubConnectionBuilder withHeader(String name, String value) {
+        if (headers == null) {
+            this.headers = new HashMap<>();
+        }
+        this.headers.put(name, value);
+        return this;
+    }
+
     public HubConnection build() {
-        return new HubConnection(url, transport, skipNegotiate, logger, httpClient, accessTokenProvider, handshakeResponseTimeout);
+        return new HubConnection(url, transport, skipNegotiate, logger, httpClient, accessTokenProvider, handshakeResponseTimeout, headers);
     }
 }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -72,7 +72,8 @@ public class HubConnection {
         this.tickRate = tickRate;
     }
 
-    HubConnection(String url, Transport transport, boolean skipNegotiate, Logger logger, HttpClient httpClient, Single<String> accessTokenProvider, Duration handshakeResponseTimeout) {
+    HubConnection(String url, Transport transport, boolean skipNegotiate, Logger logger, HttpClient httpClient,
+                  Single<String> accessTokenProvider, Duration handshakeResponseTimeout, Map<String, String> headers) {
         if (url == null || url.isEmpty()) {
             throw new IllegalArgumentException("A valid url is required.");
         }
@@ -104,6 +105,10 @@ public class HubConnection {
 
         if (handshakeResponseTimeout != null) {
             this.handshakeResponseTimeout = handshakeResponseTimeout;
+        }
+
+        if (headers != null){
+            this.headers.putAll(headers);
         }
 
         this.skipNegotiate = skipNegotiate;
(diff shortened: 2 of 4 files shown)


clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -43,6 +47,11 @@ public HubConnectionBuilder configureLogging(Logger logger) {
         return this;
     }
 
+    public HubConnectionBuilder withHeaders(Map<String, String> headers) {
```
mikaelm12: Should I add another API to add individual headers?

mikaelm12: `addHeader(String name, String value)`

anurse: Yeah, I think that would be good.

mikaelm12:
>Test?

Yeah, meant to add that a test was still needed in the description

clients/java/signalr/src/main/java/com/microsoft/aspnet/signalr/HubConnectionBuilder.java
```diff
@@ -43,6 +47,19 @@ public HubConnectionBuilder configureLogging(Logger logger) {
...
+        this.headers = headers;
+        return this;
+    }
+
+    public HubConnectionBuilder withHeader(String name, String value) {
```
anurse: Better name 👍

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -1053,6 +1055,64 @@ public void accessTokenProviderIsOverriddenFromRedirectNegotiate()
...
+                .build();
+
+        hubConnection.start().get(1000, TimeUnit.MILLISECONDS);
+        assertEquals(HubConnectionState.CONNECTED, hubConnection.getConnectionState());
+        hubConnection.stop();
+        assertEquals("ExampleValue", header.get());
+    }
+
+    @Test
+    public void headersAreSetAndSentThroughBuilder() throws Exception {
```
anurse: Can you have a test that calls `withHeader` twice with the same value? I think the best thing to do is to have the second value win (which is what `put` should do by default).

mikaelm12:
This needs to wait on https://github.com/aspnet/SignalR/issues/3097.

mikaelm12:
We're only supporting websockets for the time being

anurse:
> Does Java have equivalent header support for each transport?

Yes, the Java WebSocket library we are using does have header support, so we're good.

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -1053,6 +1055,93 @@ public void accessTokenProviderIsOverriddenFromRedirectNegotiate()
...
+    @Test
+    public void headersAreSetAndSentThroughOptions() throws Exception {
+        AtomicReference<String> header = new AtomicReference<>();
+        TestHttpClient client = new TestHttpClient()
+                .on("POST", "http://example.com/negotiate",
+                        (req) -> {
+                            header.set(req.getHeaders().get("ExampleHeader"));
+                            return CompletableFuture
+                                    .completedFuture(new HttpResponse(200, "", "{\"connectionId\":\"bVOiRPG8-6YiJ6d7ZcTOVQ\",\""
+                                            + "availableTransports\":[{\"transport\":\"WebSockets\",\"transferFormats\":[\"Text\",\"Binary\"]}]}"));
```
BrennanConroy: nit: Indenting is a bit aggressive here

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -1053,6 +1055,93 @@ public void accessTokenProviderIsOverriddenFromRedirectNegotiate()
...
+    @Test
+    public void headersAreSetAndSentThroughBuilder() throws Exception {
+        AtomicReference<String> header = new AtomicReference<>();
+        TestHttpClient client = new TestHttpClient()
+                .on("POST", "http://example.com/negotiate",
+                        (req) -> {
+                            header.set(req.getHeaders().get("ExampleHeader"));
+                            return CompletableFuture
+                                    .completedFuture(new HttpResponse(200, "", "{\"connectionId\":\"bVOiRPG8-6YiJ6d7ZcTOVQ\",\""
+                                            + "availableTransports\":[{\"transport\":\"WebSockets\",\"transferFormats\":[\"Text\",\"Binary\"]}]}"));
```
BrennanConroy: And here

clients/java/signalr/src/test/java/com/microsoft/aspnet/signalr/HubConnectionTest.java
```diff
@@ -1053,6 +1055,93 @@ public void accessTokenProviderIsOverriddenFromRedirectNegotiate()
...
+    @Test
+    public void sameHeaderSetTwiceGetsOverwritten() throws Exception {
+        AtomicReference<String> header = new AtomicReference<>();
+        TestHttpClient client = new TestHttpClient()
+                .on("POST", "http://example.com/negotiate",
+                        (req) -> {
+                            header.set(req.getHeaders().get("ExampleHeader"));
+                            return CompletableFuture
+                                    .completedFuture(new HttpResponse(200, "", "{\"connectionId\":\"bVOiRPG8-6YiJ6d7ZcTOVQ\",\""
+                                            + "availableTransports\":[{\"transport\":\"WebSockets\",\"transferFormats\":[\"Text\",\"Binary\"]}]}"));
```
BrennanConroy: And here

PR #3093, opened 2018-10-08
added workaround code for node.js-only users

bradygaster:
per request, changed incoming branch to release/2.1

### DIFF
--- a/clients/ts/signalr/README.md
+++ b/clients/ts/signalr/README.md
@@ -42,6 +42,9 @@ connection.start()
 ```JavaScript
 const signalR = require("@aspnet/signalr");
 
+XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest
+WebSocket = require("websocket").w3cwebsocket
+
 let connection = new signalR.HubConnectionBuilder()
     .withUrl("/chat")
     .build();


anurse:
@muratg Can you take this to shiproom? It's just a change to the README.md in our NPM package to include more detailed information about how to use the package from Node. This will update the information on our main NPM package page: https://www.npmjs.com/package/@aspnet/signalr

This is only needed for 2.1 because we have added native NodeJS support in 2.2+. There is **no** functional change to the code, it's purely a documentation change (it's just that the documentation lives inside our package in this case 😦 )

clients/ts/signalr/README.md
```diff
@@ -42,6 +42,9 @@ connection.start()
 ```JavaScript
 const signalR = require("@aspnet/signalr");
 
+XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest
```
BrennanConroy: nit: add semi-colons

muratg:
This does not meet the 2.1 LTS servicing bar. We can consider this again whenever SignalR has a patch in the future.

mikaelm12:
> We can consider this again whenever SignalR has a patch in the future.

Gonna close this in the mean time. We will revisit in the next patch. We have the poly-fills now so users on the latest bits won't run into this.

Issue #3091, opened 2018-10-08
Rename Java Client to com.microsoft.signalr:signalr


anurse:
With the release of the Azure SignalR service and it's server-less model, it's entirely possible for a SignalR application to have no ASP.NET Core in it at all. As a result, the name `aspnet` in the package seems odd. Let's rename **from** `com.microsoft.aspnet:signalr` **to** `com.microsoft.signalr:signalr`.

TODO: 
* [ ] Rename the JARs/packages/etc.
* [ ] Get approval for the groupId `com.microsoft.signalr` from Sonatype's OSS Maven Hosting
* [ ] Publish Preview 3 to the new name.

anurse:
https://issues.sonatype.org/browse/OSSRH-43203 filed to create the groupId

PR #3096, opened 2018-10-08
Change package name to com.microsoft.signalr

mikaelm12:
Changing the package name from `com.microsoft.aspnet.signalr` to `com.microsoft.signalr`
Fixes: #3091
//cc @JonathanGiles look at what we're doing!


anurse:
Don't forget to link the issue 😸

JonathanGiles:
Great work! My powers know no limits :-)

mikaelm12:
// cc @muratg just to make sure it's visible that we're changing the package name

anurse:
We've got the approval so as long as everything builds correctly we should be able to ship to this groupId.

anurse:
@mikaelm12 when this is merged, please post to aspnet/Announcements describing the upcoming change in preview3. Make sure to follow the pattern of creating an issue on aspnet/SignalR to discuss the change and locking the Announcements issue (to avoid spamming the people following that repo)

Issue #3097, opened 2018-10-08
[Java] Remove HttpConnectionOptions type 


mikaelm12:
We need to remove `HttpConnectionOptions `type and upgrade the `HubConnection `builder to replace the functionality.

PR #3101, opened 2018-10-09
Remove HttpConnectionOptions

mikaelm12:
Removing the `HttpConnectionOptions` type as the last part of the Java client API review. Now the builder should be able to fully handle all the `HubConnection` creation options.
Fixes: https://github.com/aspnet/SignalR/issues/3097


mikaelm12:
Actually, I still need to do some work on this to create an `HttpHubConnectionBuilder` type that derives from an abstract `HubConnectionBuilder ` base class.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnectionBuilder.java
```diff
@@ -35,25 +41,27 @@ public HubConnectionBuilder configureLogging(LogLevel logLevel) {
         return this;
     }
 
+    public HubConnectionBuilder withSkipNegotiate(boolean skipNegotiate) {
```
mikaelm12: `shouldSkipNegotiate` would be better here imo

Issue #3073, opened 2018-10-05
Switch to RxJava instead of CompletableFuture


anurse:
Since we're going to use RxJava for Streaming, we should just use it everywhere as our async library.

DHosseiny:
There is other good reason to change it. CompletableFuture requires API level 24 in android.

PR #3103, opened 2018-10-09
[Java] Make public API look like RxJava

BrennanConroy:
Part of https://github.com/aspnet/SignalR/issues/3073

### DIFF
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -33,6 +33,7 @@ dependencies {
     testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.3.1'
     implementation 'com.google.code.gson:gson:2.8.5'
     implementation 'com.squareup.okhttp3:okhttp:3.11.0'
+    implementation 'io.reactivex.rxjava2:rxjava:2.2.2'
 }
 
 spotless {
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -9,12 +9,14 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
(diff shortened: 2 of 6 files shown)


mikaelm12:
Summary of the conceptual changes here would be helpful

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -32,13 +34,13 @@
     private Logger logger;
     private List<Consumer<Exception>> onClosedCallbackList;
     private boolean skipNegotiate;
-    private Supplier<CompletableFuture<String>> accessTokenProvider;
+    private Supplier<Single<String>> accessTokenProvider;
```
anurse: We can just use `Single<String>` instead of `Supplier<Single<String>>` because each subscription to the Single re-invokes whatever the user provided as their implementation.

For example, I could implement the "renew token when expired" logic using pseudo-code like this:

```java
builder.withAccessTokenProvider(Single.fromCallable(() -> {
    if(accessToken == null || accessToken.isExpired()) {
        accessToken = renewAccessToken();
    }
    return accessToken;
});
```

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -181,18 +180,19 @@ public HubConnectionState getConnectionState() {
      * Starts a connection to the server.
      * @return A completable future that completes when the connection has been established.
```
anurse: Update these return docs to reference Completable instead of CompletableFuture

@dend If we want to link to external Javadocs (for third-party libraries) can we do that in this way? The `Completable` type here comes from RxJava. Can we make something like `{@link Completable}` generate a link to reactivex.io/RxJava/javadoc/io/reactivex/Completable.html when the docs are built?

clients/java/signalr/build.gradle
```diff
@@ -33,6 +33,7 @@ dependencies {
     testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.3.1'
     implementation 'com.google.code.gson:gson:2.8.5'
     implementation 'com.squareup.okhttp3:okhttp:3.11.0'
+    implementation 'io.reactivex.rxjava2:rxjava:2.2.2'
```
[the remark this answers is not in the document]
Eilon: Good to go.

PR #3112, opened 2018-10-10
[Java] Fix incorrect visibility and some minor cleanup

BrennanConroy:
`withHandshakeResponseTimeout` wasn't public and `TransferFormat` was public

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
@@ -8,7 +8,7 @@ import java.time.Duration;
 import io.reactivex.Single;
 
 public class HttpHubConnectionBuilder {
-    private String url;
+    private final String url;
     private Transport transport;
     private Logger logger;
     private HttpClient httpClient;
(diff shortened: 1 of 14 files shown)


clients/java/signalr/src/main/java/com/microsoft/signalr/HubException.java
```diff
@@ -4,6 +4,8 @@
 package com.microsoft.signalr;
 
 public class HubException extends Exception {
+    private static final long serialVersionUID = -572019264269821519L;
```
anurse: Why this number? Why not just `1`? AFAICT this just has to be a number that is incremented whenever we make a change that would affect serialization.

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -56,7 +56,7 @@ public HttpHubConnectionBuilder withLogger(Logger logger) {
         return this;
     }
 
-    HttpHubConnectionBuilder withHandshakeResponseTimeout(Duration timeout) {
+    public HttpHubConnectionBuilder withHandshakeResponseTimeout(Duration timeout) {
```
mikaelm12: 👍

PR #3114, opened 2018-10-11
Committing changes required to have SignalR depend on the centralized Microsoft BOM file for Java libraries

JonathanGiles:
For discussion with the Java client engineering team, @anurse @BrennanConroy @mikaelm12 @davidfowl. This is a proposal to externalise your dependency **version** management to an external POM. We are gradually rolling this out to all Java SDKs, and I am keen to hear your thoughts, feedback, concerns, or whatever else may be running through your mind 😄 

All versions that have been removed from your Gradle file are now represented in the Maven parent POM file, and as your dependencies have not yet been encountered (and were therefore not included up to now), I have taken the exact same version numbers as you use. There will therefore be no functional difference.

### DIFF
--- a/clients/java/signalr/build.gradle
+++ b/clients/java/signalr/build.gradle
@@ -25,15 +25,22 @@ sourceCompatibility = 1.8
 
 repositories {
     mavenCentral()
+    // add sonatype repository (temporary, due to java-8-parent being a snapshot)
+    maven {
+        url 'https://oss.sonatype.org/content/repositories/snapshots/'
+    }
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


anurse:
Just so I understand what's going on here. This is basically externalizing the **version numbers** of the dependencies we use so that we're aligned with the ones used by other Microsoft Java libraries? I think that sounds like a good idea.

What's the process we should follow if we need to update one of these dependencies? If it's urgent, can we override the version from `com.microsoft.maven:java-8-parent` with our own version locally?

JonathanGiles:
Thanks guys for your questions. I have answered these elsewhere for other teams and I realise I really should centralise the answers to these kinds of questions. I'll do that and then get back to you.

Issue #3070, opened 2018-10-05
Use slf4j logging library


mikaelm12:
The slf4j or Simple Logging Facade for Java seems to be the best pluggable logging option. We should update the client to depend on this.
Some docs: https://www.slf4j.org/manual.html

PR #3120, opened 2018-10-11
Logging with Slf4j

mikaelm12:
Removing our custom logging implementation in favor of using thew SLF4J APIs. Users will be able to take a dependency on a logging library of their choice, and slf4j will enable them to plug it in to the Java client. 
Fixes: https://github.com/aspnet/SignalR/issues/3070


mikaelm12:
cc @JonathanGiles for extra eyes on this 😄

clients/java/signalr/src/main/java/com/microsoft/signalr/DefaultHttpClient.java
```diff
@@ -24,10 +27,10 @@
 
 final class DefaultHttpClient extends HttpClient {
     private final OkHttpClient client;
-    private Logger logger;
+    private final Logger logger = LoggerFactory.getLogger(DefaultHttpClient.class);
```
JonathanGiles: This class no longer even needs a Logger instance, from what I can see.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -72,7 +74,10 @@ void setTickRate(Duration tickRate) {
         this.tickRate = tickRate;
     }
 
-    HubConnection(String url, Transport transport, boolean skipNegotiate, Logger logger, HttpClient httpClient,
+    private final Logger logger = LoggerFactory.getLogger(HubConnection.class);
```
JonathanGiles: It seems weird to put this field so far down in the class?

clients/java/signalr/src/main/java/com/microsoft/signalr/DefaultHttpClient.java
mikaelm12: Yes, I meant to remove that. Good catch

clients/java/signalr/build.gradle
```diff
@@ -41,6 +41,7 @@ dependencies {
     implementation 'com.google.code.gson:gson'
     implementation 'com.squareup.okhttp3:okhttp'
     implementation 'io.reactivex.rxjava2:rxjava'
+    implementation 'org.slf4j:slf4j-api'
```
mikaelm12: cc @Eilon for a new dependency

anurse: FYI, for Eilon, the version number here is: 1.7.25

Version numbers were centralized in a common "bill of materials" package across Microsoft in https://github.com/aspnet/SignalR/pull/3114

Eilon: Go for it.

anurse:
@mikaelm12 Can you file a bug on aspnet/Docs and assign it to yourself to add information about this error message and how to solve it to our Java Client doc before preview3 actually ships?

clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
```diff
@@ -19,17 +22,17 @@
...
     private Map<String, String> headers;
     private OkHttpClient client;
-    private Logger logger;
     private OnReceiveCallBack onReceive;
     private BiConsumer<Integer, String> onClose;
     private CompletableFuture<Void> startFuture = new CompletableFuture<>();
     private CompletableFuture<Void> closeFuture = new CompletableFuture<>();
 
-    public OkHttpWebSocketWrapper(String url, Map<String, String> headers, OkHttpClient client, Logger logger) {
+    private final Logger logger = LoggerFactory.getLogger(OkHttpWebSocketWrapper.class);
```
BrennanConroy: Not a fan of using this class as the "category" but I won't stop the PR because of it.

mikaelm12:
>Can you file a bug on aspnet/Docs and assign it to yourself to add information about this error message and how to solve it to our Java Client doc before preview3 actually ships?

Got it: https://github.com/aspnet/Docs/issues/8991

PR #3123, opened 2018-10-12
Add package-info file for docs

mikaelm12:
Adding a package-info file to get the high level class summaries on the docs page. 
Does this look right to you @dend ?

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/package-info.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/package-info.java
@@ -1,0 +1,4 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/package-info.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/package-info.java
@@ -1,4 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+/**
+ * This package contains the classes for SignalR Java client.
+ * SignalR Java client.
+ */
 package com.microsoft.signalr;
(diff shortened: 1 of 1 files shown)


mikaelm12:
cc @EdisonDa

clients/java/signalr/src/main/java/com/microsoft/signalr/package-info.java
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
```
anurse: How does this improve the current doc? Do we need to put additional Javadocs in here to do that? The [current doc](https://docs.microsoft.com/en-us/java/api/com.microsoft.aspnet.signalr?view=aspnet-signalr-java) already has a list of Classes/Interfaces so I think what we're looking for is the ability to add a description...

mikaelm12:
@EdisonDa Definitely, we are planning on doing a lot of docs work in  the coming week since we've locked down on preview 3 features. I just wanted to get this change in separately and run it by you to make sure that this is what you needed.

@anurse The docs you linked to has the list of classes but it should also have a short summary of those classes directly to the right. We need this `package-info` class for that to be generated. [Here's an example](https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.management.appservice?view=azure-java-stable)

mikaelm12:
Any comments here? Just want to be sure that this is what we need for the docs summary generation. Thanks!

anurse:
> Here's an example

Right, and they have a descriptive blurb for the package itself in [their package-info.java file](https://github.com/Azure/azure-libraries-for-java/blob/master/azure-mgmt-appservice/src/main/java/com/microsoft/azure/management/appservice/package-info.java). We should have the same

mikaelm12:
I just copied what they had in the second example. I don't think we need second line of the description though.

clients/java/signalr/src/main/java/com/microsoft/signalr/package-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+/**
+ * This package contains the classes for SignalR Java client.
+ * SignalR Java client.
```
anurse: Yeah, I don't think you need this.

```suggestion
```

mikaelm12:
@dend Does this look okay?

Issue #3121, opened 2018-10-11
Enable logging in Java client sample


mikaelm12:
After we move from our custom logging implementation to using SLF4J we should give guidance on how to enable logging by doing it in the chat sample.

mikaelm12:
Done here: https://github.com/aspnet/SignalR/pull/3134

PR #3134, opened 2018-10-15
Enable logging in chat sample

mikaelm12:
Issue: https://github.com/aspnet/SignalR/issues/3121
Updating the Chat sample to use java console logging through the SLF4J APIs. We have to take another dependency on the bridging jar between slf4j and java.util.logging


clients/java/signalr/build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
     testCompile 'org.junit.jupiter:junit-jupiter-params'
     testRuntime 'org.junit.jupiter:junit-jupiter-engine'
+    testCompile 'org.slf4j:slf4j-jdk14:1.7.25'
```
mikaelm12: cc @Eilon for the new dependency

Eilon: Good to go.

PR #3135, opened 2018-10-15
Check if HubConnection stopped normally in Chat sample

mikaelm12:
*This isn't a product issue*
The `OnClosed ` callbacks get an optional exception parameter which gets passed in when the `HubConnection ` stops with an error. If the `HubConnection ` stops normally the exception is null and the chat sample we have null refs as the application closes.


mikaelm12:

mikaelm12:

Issue #2982, opened 2018-09-19
Complete Java Docs


mikaelm12:
The current Java docs we have were a first pass for the main public types. We need to go through again and add any missing documentation before preview 3 since the API docs will be on the docs site.

anurse:
Just waiting on API docs team to update the docs themselves.

mikaelm12:
Docs are up now!
Main docs: https://docs.microsoft.com/en-us/aspnet/core/signalr/java-client?view=aspnetcore-2.2
Api Docs: https://docs.microsoft.com/en-us/java/api/com.microsoft.signalr?view=aspnet-signalr-java

PR #3138, opened 2018-10-15
More Java Docs

mikaelm12:
More Java Docs! Doing a another run of API docs work after the preview 3 changes.
Issue: https://github.com/aspnet/SignalR/issues/2982

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/Action.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/Action.java
@@ -3,6 +3,9 @@
 
 package com.microsoft.signalr;
 
+/**
+ * A callback that has no parameters.
+ */
 @FunctionalInterface
 public interface Action {
     void invoke();
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/Action.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/Action.java
@@ -4,7 +4,7 @@
 package com.microsoft.signalr;
 
 /**
- * A callback that has no parameters.
+ * A callback that takes no parameters.
  */
 @FunctionalInterface
 public interface Action {
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/Action4.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/Action4.java
@@ -3,6 +3,13 @@
 
 package com.microsoft.signalr;
 
+/**
+ * A callback that takes four parameters.
+ * @param <T1> The type of the one parameter to the callback.
+ * @param <T2> The type of the second parameter to the callback.
+ * @param <T3> The type of the third parameter to the callback.
+ * @param <T4> The type of the fourth parameter to the callback.
+ */
 @FunctionalInterface
 public interface Action4<T1, T2, T3, T4> {
     void invoke(T1 param1, T2 param2, T3 param3, T4 param4);
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/Action4.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/Action4.java
@@ -6,7 +6,7 @@ package com.microsoft.signalr;
 /**
  * A callback that takes four parameters.
  *
- * @param <T1> The type of the one parameter to the callback.
+ * @param <T1> The type of the first parameter to the callback.
  * @param <T2> The type of the second parameter to the callback.
  * @param <T3> The type of the third parameter to the callback.
  * @param <T4> The type of the fourth parameter to the callback.
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/Action4.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/Action4.java
@@ -5,6 +5,7 @@ package com.microsoft.signalr;
 
 /**
  * A callback that takes four parameters.
+ *
  * @param <T1> The type of the one parameter to the callback.
  * @param <T2> The type of the second parameter to the callback.
  * @param <T3> The type of the third parameter to the callback.
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
@@ -27,6 +27,7 @@ public class HttpHubConnectionBuilder {
 
     /**
      * Sets the transport to be used by the {@link HubConnection}.
+     *
      * @param transport
      * @return This instance of the HttpHubConnectionBuilder
      */
@@ -37,6 +38,7 @@ public class HttpHubConnectionBuilder {
 
     /**
      * Sets the HttpClient to be used by the {@link HubConnection}.
+     *
      * @param httpClient The HttpClient to be used by the {@link HubConnection}.
      * @return This instance of the HttpHubConnectionBuilder
      */
@@ -46,7 +48,9 @@ public class HttpHubConnectionBuilder {
     }
 
     /**
-     * Indicates to the {@link HubConnection}
+     * Indicates to the {@link HubConnection} that it should skip the negotiate process.
+     * Note: The Azure SignalR Service require the negotiate step.
+     *
      * @param skipNegotiate Boolean indicating if the {@link HubConnection} should skip the negotiate step.
      * @return This instance of the HttpHubConnectionBuilder
      */
@@ -56,7 +60,8 @@ public class HttpHubConnectionBuilder {
     }
 
     /**
-     * Sets the access token provider for
+     * Sets the access token provider for the {@link HubConnection}.
+     *
      * @param accessTokenProvider The Access Token Provider to be used by the {@link HubConnection}.
      * @return This instance of the HttpHubConnectionBuilder
      */
@@ -67,6 +72,7 @@ public class HttpHubConnectionBuilder {
 
     /**
      * Sets the duration the {@link HubConnection} should wait for a Handshake Response from the server.
+     *
      * @param timeout The duration that the {@link HubConnection} should wait for a Handshake Response from the server.
      * @return This instance of the HttpHubConnectionBuilder
      */
@@ -77,6 +83,7 @@ public class HttpHubConnectionBuilder {
 
     /**
      * Sets a collection of Headers for the {@link HubConnection} to send.
+     *
      * @param headers A Map representing the collection of Headers that the {@link HubConnection} should send
      * @return This instance of the HttpHubConnectionBuilder
      */
@@ -87,6 +94,7 @@ public class HttpHubConnectionBuilder {
 
     /**
      * Sets a single header for the {@link HubConnection} to send.
+     *
      * @param name The name of the header to set.
      * @param value The value
      * @return
@@ -100,6 +108,7 @@ public class HttpHubConnectionBuilder {
     }
 
     /**
+     * Builds a new instance of {@link HubConnection}.
      *
      * @return A new instance of {@link HubConnection}.
      */
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
@@ -9,6 +9,9 @@ import java.util.Map;
 
 import io.reactivex.Single;
 
+/**
+ * A builder for configuring {@link HubConnection} instances.
+ */
 public class HttpHubConnectionBuilder {
     private final String url;
     private Transport transport;
@@ -22,36 +25,72 @@ public class HttpHubConnectionBuilder {
         this.url = url;
     }
 
+    /**
+     * Sets the transport to be used by the {@link HubConnection}.
+     * @param transport
+     * @return This instance of the HttpHubConnectionBuilder
+     */
     HttpHubConnectionBuilder withTransport(Transport transport) {
         this.transport = transport;
         return this;
     }
 
+    /**
+     * Sets the HttpClient to be used by the {@link HubConnection}.
+     * @param httpClient The HttpClient to be used by the {@link HubConnection}.
+     * @return This instance of the HttpHubConnectionBuilder
+     */
     HttpHubConnectionBuilder withHttpClient(HttpClient httpClient) {
         this.httpClient = httpClient;
         return this;
     }
 
+    /**
+     * Indicates to the {@link HubConnection}
+     * @param skipNegotiate Boolean indicating if the {@link HubConnection} should skip the negotiate step.
+     * @return This instance of the HttpHubConnectionBuilder
+     */
     public HttpHubConnectionBuilder shouldSkipNegotiate(boolean skipNegotiate) {
         this.skipNegotiate = skipNegotiate;
         return this;
     }
 
+    /**
+     * Sets the access token provider for
+     * @param accessTokenProvider The Access Token Provider to be used by the {@link HubConnection}.
+     * @return This instance of the HttpHubConnectionBuilder
+     */
     public HttpHubConnectionBuilder withAccessTokenProvider(Single<String> accessTokenProvider) {
         this.accessTokenProvider = accessTokenProvider;
         return this;
     }
 
+    /**
+     * Sets the duration the {@link HubConnection} should wait for a Handshake Response from the server.
+     * @param timeout The duration that the {@link HubConnection} should wait for a Handshake Response from the server.
+     * @return This instance of the HttpHubConnectionBuilder
+     */
     public HttpHubConnectionBuilder withHandshakeResponseTimeout(Duration timeout) {
         this.handshakeResponseTimeout = timeout;
         return this;
     }
 
+    /**
+     * Sets a collection of Headers for the {@link HubConnection} to send.
+     * @param headers A Map representing the collection of Headers that the {@link HubConnection} should send
+     * @return This instance of the HttpHubConnectionBuilder
+     */
     public HttpHubConnectionBuilder withHeaders(Map<String, String> headers) {
         this.headers = headers;
         return this;
     }
 
+    /**
+     * Sets a single header for the {@link HubConnection} to send.
+     * @param name The name of the header to set.
+     * @param value The value
+     * @return
+     */
     public HttpHubConnectionBuilder withHeader(String name, String value) {
         if (headers == null) {
             this.headers = new HashMap<>();
@@ -60,6 +99,10 @@ public class HttpHubConnectionBuilder {
         return this;
     }
 
+    /**
+     *
+     * @return A new instance of {@link HubConnection}.
+     */
     public HubConnection build() {
         return new HubConnection(url, transport, skipNegotiate, httpClient, accessTokenProvider, handshakeResponseTimeout, headers);
     }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
@@ -49,7 +49,7 @@ public class HttpHubConnectionBuilder {
 
     /**
      * Indicates to the {@link HubConnection} that it should skip the negotiate process.
-     * Note: The Azure SignalR Service require the negotiate step.
+     * Note: This option only works with the Websockets transport and the Azure SignalR Service require the negotiate step.
      *
      * @param skipNegotiate Boolean indicating if the {@link HubConnection} should skip the negotiate step.
      * @return This instance of the HttpHubConnectionBuilder.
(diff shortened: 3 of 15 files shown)


clients/java/signalr/src/main/java/com/microsoft/signalr/Action.java
```diff
@@ -3,6 +3,9 @@
 
 package com.microsoft.signalr;
 
+/**
+ * A callback that has no parameters.
```
anurse: let's align the verb here and say `takes no parameters`

clients/java/signalr/src/main/java/com/microsoft/signalr/Action4.java
```diff
@@ -3,6 +3,14 @@
 
 package com.microsoft.signalr;
 
+/**
+ * A callback that takes four parameters.
+ *
+ * @param <T1> The type of the one parameter to the callback.
```
anurse: `one` -> `first`

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
         this.url = url;
     }
 
+    /**
+     * Sets the transport to be used by the {@link HubConnection}.
+     *
+     * @param transport The transport to be set.
```
anurse: The transport to ~~be set~~ **use**.

Aligns better with other docs

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
...
+     * @return This instance of the HttpHubConnectionBuilder
+     */
     public HttpHubConnectionBuilder withTransport(Transport transport) {
         this.transport = transport;
         return this;
     }
 
+    /**
+     * Sets the HttpClient to be used by the {@link HubConnection}.
```
anurse: Can we `@link` HttpClient here and in the `@param`?

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
...
+     */
     public HttpHubConnectionBuilder shouldSkipNegotiate(boolean skipNegotiate) {
         this.skipNegotiate = skipNegotiate;
         return this;
     }
 
+    /**
+     * Sets the access token provider for the {@link HubConnection}.
+     *
+     * @param accessTokenProvider The Access Token Provider to be used by the {@link HubConnection}.
```
anurse: Avoid Title Case Unless Dealing With Proper Nouns :trollface:

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -55,18 +55,38 @@
     private final Logger logger = LoggerFactory.getLogger(HubConnection.class);
 
 
+    /**
+     * Sets the server timeout interval for the connection.
+     *
+     * @param serverTimeout The timeout duration.
```
anurse: The **server** timeout duration

(so it agrees with `getServerTimeout`'s `@return`)

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
...
         this.headers = headers;
         return this;
     }
 
+    /**
+     * Sets a single header for the {@link HubConnection} to send.
+     *
+     * @param name The name of the header to set.
+     * @param value The value of the header to be set.
+     * @return
```
anurse: What does it return? What does it return!!??! I'm on the edge of my seat ;P

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -55,18 +55,38 @@
...
+     * @param keepAliveInterval The interval at which the connection should send keep alive messages.
+     */
     public void setKeepAliveInterval(Duration keepAliveInterval) {
         this.keepAliveInterval = keepAliveInterval;
     }
 
+    /**
+     * Gets the keep alive interval
+     * 
+     * @return The interval durations between keep alive messages.
```
anurse: The interval ~~durations~~ between keep alive messages.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -55,18 +55,38 @@
...
+     * Gets the server timeout duration.
+     *
+     * @return The server timeout duration.
+     */
     public Duration getServerTimeout() {
         return this.serverTimeout;
     }
 
+    /**
+     * Sets the connections Keep Alive Interval duration.
```
anurse: Sets the ~~connections~~ ~~K~~**k**eep ~~A~~**a**live ~~I~~**i**nterval.

(avoid Title Case and we already know this relates to the connection; plus that would require an apostrophe if we did use it: `connection's`)

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -416,6 +438,14 @@ public void send(String method, Object... args) {
         sendHubMessage(invocationMessage);
     }
 
+    /**
+     * Invokes a hub method on the server using the specified name and arguments.
```
anurse: Suggest: "Invokes a hub method on the server using the specified **method** name and arguments."

(To align with [.NET equivalent](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.signalr.client.hubconnectionextensions.invokeasync?view=aspnetcore-2.1#Microsoft_AspNetCore_SignalR_Client_HubConnectionExtensions_InvokeAsync_Microsoft_AspNetCore_SignalR_Client_HubConnection_System_String_System_Object_System_Object_System_Object_System_Object_System_Object_System_Object_System_Object_System_Object_System_Object_System_Object_System_Threading_CancellationToken_) docs, for consistency)

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -416,6 +438,14 @@ public void send(String method, Object... args) {
...
+     * @param returnType The expected return type.
```
anurse: Suggest: "The return type of the server method."

(Same as above, align with .NET docs)

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -416,6 +438,14 @@ public void send(String method, Object... args) {
...
+     * @param method The name of the server method to invoke.
+     * @param args The arguments used to invoke the server method.
+     * @param <T> The expected return type.
```
anurse: Suggest: "The return type of the server method."

(Same as above, align with .NET docs)

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -416,6 +438,14 @@ public void send(String method, Object... args) {
...
+     * @return A Single of the expected return type.
```
anurse: Suggest: "A Single that yields the return value when the invocation has completed"

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -478,6 +508,10 @@ public void remove(String name) {
         logger.trace("Removing handlers for client method: {}.", name);
     }
 
+    /**
+     * Registers callbacks to run when the connection closes.
```
anurse: Suggest: "Registers **a** callback~~s~~ to run when the connection **is** close**d**~~s~~."

The method only takes one callback at a time :)

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnectionBuilder.java
```diff
@@ -3,13 +3,25 @@
 
 package com.microsoft.signalr;
 
+/**
+ * A builder for configuring {@link HubConnection} instances.
+ */
 public abstract class HubConnectionBuilder {
+    /**
+     * Creates a new instance of {@link HttpHubConnectionBuilder}.
+     * @param url The end point url that the HubConnection will connect to
```
anurse: Suggest: The URL of the SignalR server to connect to.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubException.java
```diff
@@ -3,16 +3,32 @@
 
 package com.microsoft.signalr;
 
+/**
+ * The exception thrown from a hub when an error occurs.
```
anurse: Suggest: An exception thrown when the server fails to invoke a Hub method.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubException.java
```diff
@@ -3,16 +3,32 @@
...
+     * Initializes a new instance of the {@link HubException} class with a specified error message
+     * @param errorMessage The error message that explains the reason for the exception.
+     */
     public HubException(String errorMessage) {
         super(errorMessage);
     }
 
+    /**
+     * Initializes a new instance of the {@link HubException} class with a specified error message and a reference
+     * to the inner exception that is the cause of this exception.
```
anurse: Suggest: ...a reference to the ~~inner~~ exception that is the cause of this exception.

Sounds like "inner exception" is a .NET term, and "cause" is the preferred term in Java. No need to mix in .NET terms here. For example, see [Javadocs for `Exception`](https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html#Exception(java.lang.String,%20java.lang.Throwable))

anurse: Feel free to just copy the Javadocs off the official ones for Exception if you think that's more appropriate :). Up to you.

clients/java/signalr/src/main/java/com/microsoft/signalr/Subscription.java
```diff
@@ -5,17 +5,23 @@
 
 import java.util.List;
 
+/**
+ * A type that represents being subscribed to invocations to a specific target.
```
anurse: Suggest: Represents the registration of a handler for a client method.

"A type" is self-evident, since it's a class, and we don't use prefixes like that in other docs. "Represents being subscribed" feels a little awkward, and the type really represents the registration.

I kinda wonder if we should rename this type since a) it might conflict/confuse with [Rx's `Subscription`](http://reactivex.io/RxJava/javadoc/rx/Subscription.html) and b) we don't really use the term "subscribe" when you register the handler. It's really more like "defining" a method. We can talk about it post-preview3 (I know it's public API but it's minor and I don't think it's worth trying to squeeze it in; I'd rather just not rename it than try to do that)

clients/java/signalr/src/main/java/com/microsoft/signalr/Subscription.java
```diff
@@ -5,17 +5,23 @@
...
 
-    public Subscription(CallbackMap handlers, InvocationHandler handler, String target) {
+    Subscription(CallbackMap handlers, InvocationHandler handler, String target) {
         this.handlers = handlers;
         this.handler = handler;
         this.target = target;
     }
 
+    /**
+     * Unsubscribes from a target by removing all handlers associated with it.
```
anurse: It only unsubscribes the handler represented by this registration though right? So given:

```java
Subscription a = connection.on("foo", ...);
Subscription b = connection.on("foo", ...);
a.unsubscribe();
```

`b` is still subscribed right? The doc seems to imply this removes all handlers.

How about: "Removes the client method handler represented by this subscription" ?

clients/java/signalr/src/main/java/com/microsoft/signalr/Subscription.java
mikaelm12: Yeah, this was a weird one.

clients/java/signalr/src/main/java/com/microsoft/signalr/Subscription.java
mikaelm12: Yes, that's right. Mix this up with the remove method.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubException.java
anurse: Yuck. Don't like it there either :). We can fix it there later, but let's not perpetuate it unless we think it's clearer.

mikaelm12: Cool

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
...
         return this;
     }
 
+    /**
+     * Sets the {@link HttpClient} to be used by the {@link HubConnection}.
+     *
+     * @param httpClient The {@link HttpClient} to be used by the {@link HubConnection}.
+     * @return This instance of the HttpHubConnectionBuilder.
+     */
     public HttpHubConnectionBuilder withHttpClient(HttpClient httpClient) {
```
BrennanConroy: This shouldn't be public either!

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
...
         this.httpClient = httpClient;
         return this;
     }
 
+    /**
+     * Indicates to the {@link HubConnection} that it should skip the negotiate process.
+     * Note: The Azure SignalR Service require the negotiate step.
```
BrennanConroy: I know we only have WebSockets right now, but maybe we should note this only works with the WebSockets transport so we don't have to remember to update it later.

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
...
+     * @param timeout The duration that the {@link HubConnection} should wait for a Handshake Response from the server.
+     * @return This instance of the HttpHubConnectionBuilder.
+     */
     public HttpHubConnectionBuilder withHandshakeResponseTimeout(Duration timeout) {
         this.handshakeResponseTimeout = timeout;
         return this;
     }
 
+    /**
+     * Sets a collection of Headers for the {@link HubConnection} to send.
```
BrennanConroy: "to send with every Http request."

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -55,18 +55,38 @@
     private final Logger logger = LoggerFactory.getLogger(HubConnection.class);
 
 
+    /**
+     * Sets the server timeout interval for the connection.
```
BrennanConroy: Explain what it's timing out

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -55,18 +55,38 @@
...
+     * Sets the keep alive interval duration.
+     *
+     * @param keepAliveInterval The interval at which the connection should send keep alive messages.
+     */
     public void setKeepAliveInterval(Duration keepAliveInterval) {
         this.keepAliveInterval = keepAliveInterval;
     }
 
+    /**
+     * Gets the keep alive interval
```
BrennanConroy: full stop

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -416,6 +438,15 @@ public void send(String method, Object... args) {
...
     }
 
+    /**
+     * Invokes a hub method on the server using the specified method name and arguments.
+     *
+     * @param returnType The expected return type.
+     * @param method The name of the server method to invoke.
+     * @param args The arguments used to invoke the server method.
+     * @param <T> The expected return type.
+     * @return A Single that yields the return value when the invocation has completed
```
BrennanConroy: full-stop

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnectionBuilder.java
```diff
@@ -3,13 +3,25 @@
 
 package com.microsoft.signalr;
 
+/**
+ * A builder for configuring {@link HubConnection} instances.
+ */
 public abstract class HubConnectionBuilder {
+    /**
+     * Creates a new instance of {@link HttpHubConnectionBuilder}.
+     * @param url The URL of the SignalR server to connect to.
```
BrennanConroy: maybe mention hub "The URL of the SignalR hub to connect to"?

clients/java/signalr/src/main/java/com/microsoft/signalr/HttpHubConnectionBuilder.java
```diff
@@ -22,36 +25,80 @@
...
+     * @return This instance of the HttpHubConnectionBuilder.
+     */
     public HttpHubConnectionBuilder withHttpClient(HttpClient httpClient) {
         this.httpClient = httpClient;
         return this;
     }
 
+    /**
+     * Indicates to the {@link HubConnection} that it should skip the negotiate process.
+     * Note: This option only works with the Websockets transport and the Azure SignalR Service require the negotiate step.
```
anurse:
```suggestion
     * Note: This option only works with the Websockets transport and the Azure SignalR Service requires the negotiate step.
```

anurse: We only really have the one transport here, I wonder if this may confuse users who aren't able to find a way to configure the transport...

mikaelm12:
Let's get the last reviews on this so we can generate the new API docs.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
BrennanConroy: Not doing this?

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -478,6 +509,10 @@ public void remove(String name) {
         logger.trace("Removing handlers for client method: {}.", name);
     }
 
+    /**
+     * Registers a callback to run when the connection is closed.
```
BrennanConroy: nit: newline

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnectionBuilder.java
```diff
@@ -3,13 +3,25 @@
 
 package com.microsoft.signalr;
 
+/**
+ * A builder for configuring {@link HubConnection} instances.
+ */
 public abstract class HubConnectionBuilder {
+    /**
+     * Creates a new instance of {@link HttpHubConnectionBuilder}.
```
BrennanConroy: nit: newline

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnectionBuilder.java
```diff
@@ -3,13 +3,25 @@
...
+     */
     public static HttpHubConnectionBuilder create(String url) {
         if (url == null || url.isEmpty()) {
             throw new IllegalArgumentException("A valid url is required.");
         }
         return new HttpHubConnectionBuilder(url);
     }
 
+    /**
+     * Builds a new instance of {@link HubConnection}.
```
BrennanConroy: nit: newline

clients/java/signalr/src/main/java/com/microsoft/signalr/HubException.java
```diff
@@ -3,16 +3,32 @@
...
     private static final long serialVersionUID = -572019264269821519L;
 
+    /**
+     * Initializes a new instance of the {@link HubException} class.
+     */
     public HubException() {
     }
 
+    /**
+     * Initializes a new instance of the {@link HubException} class with a specified error message
```
BrennanConroy: nit: newline

clients/java/signalr/src/main/java/com/microsoft/signalr/HubException.java
```diff
@@ -3,16 +3,32 @@
...
+     * @param errorMessage The error message that explains the reason for the exception.
+     */
     public HubException(String errorMessage) {
         super(errorMessage);
     }
 
+    /**
+     * Initializes a new instance of the {@link HubException} class with a specified error message and a reference
+     * to the exception that is the cause of this exception.
```
BrennanConroy: nit: newline

PR #3148, opened 2018-10-18
[Java] Plumb RxJava through client

BrennanConroy:
Fixes https://github.com/aspnet/SignalR/issues/3073

~~TODO: Write a couple tests verifying the hot vs. cold observable behavior of things like `hubConnection.start` and `hubConnection.stop`~~
We have tests that don't subscribe to start and verify things have happened and similarly for stop. Also verified manually.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -208,8 +208,15 @@ public class HubConnection {
 
     private void timeoutHandshakeResponse(long timeout, TimeUnit unit) {
         ScheduledExecutorService scheduledThreadPool = Executors.newSingleThreadScheduledExecutor();
-        scheduledThreadPool.schedule(() -> handshakeResponseSubject.onError(
-                new TimeoutException("Timed out waiting for the server to respond to the handshake message.")), timeout, unit);
+        scheduledThreadPool.schedule(() -> {
+            // If onError is called on a completed subject the global error handler is called
+            // REVIEW: Should we lock?
+            if (!(handshakeResponseSubject.hasComplete() || handshakeResponseSubject.hasThrowable()))
+            {
+                handshakeResponseSubject.onError(
+                    new TimeoutException("Timed out waiting for the server to respond to the handshake message."));
+            }
+        }, timeout, unit);
     }
 
     private Single<NegotiateResponse> handleNegotiate(String url) {
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -22,6 +22,8 @@ import org.slf4j.LoggerFactory;
 
 import io.reactivex.Completable;
 import io.reactivex.Single;
+import io.reactivex.subjects.CompletableSubject;
+import io.reactivex.subjects.SingleSubject;
 
 public class HubConnection {
     private static final String RECORD_SEPARATOR = "\u001e";
@@ -49,7 +51,7 @@ public class HubConnection {
     private long keepAliveInterval = 15*1000;
     private long serverTimeout = 30*1000;
     private long tickRate = 1000;
-    private CompletableFuture<Void> handshakeResponseFuture;
+    private CompletableSubject handshakeResponseFuture;
     private long handshakeResponseTimeout = 15*1000;
     private final Logger logger = LoggerFactory.getLogger(HubConnection.class);
 
@@ -140,18 +142,18 @@ public class HubConnection {
                     handshakeResponse = HandshakeProtocol.parseHandshakeResponse(handshakeResponseString);
                 } catch (RuntimeException ex) {
                     RuntimeException exception = new RuntimeException("An invalid handshake response was received from the server.", ex);
-                    handshakeResponseFuture.completeExceptionally(exception);
+                    handshakeResponseFuture.onError(exception);
                     throw exception;
                 }
                 if (handshakeResponse.getHandshakeError() != null) {
                     String errorMessage = "Error in handshake " + handshakeResponse.getHandshakeError();
                     logger.error(errorMessage);
                     RuntimeException exception = new RuntimeException(errorMessage);
-                    handshakeResponseFuture.completeExceptionally(exception);
+                    handshakeResponseFuture.onError(exception);
                     throw exception;
                 }
                 handshakeReceived = true;
-                handshakeResponseFuture.complete(null);
+                handshakeResponseFuture.onComplete();
 
                 payload = payload.substring(handshakeLength);
                 // The payload only contained the handshake response so we can return.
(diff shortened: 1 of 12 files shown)


clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -187,15 +189,22 @@ void setTickRate(Duration tickRate) {
 
     private void timeoutHandshakeResponse(long timeout, TimeUnit unit) {
         ScheduledExecutorService scheduledThreadPool = Executors.newSingleThreadScheduledExecutor();
-        scheduledThreadPool.schedule(() -> handshakeResponseFuture.completeExceptionally(
-                new TimeoutException("Timed out waiting for the server to respond to the handshake message.")), timeout, unit);
+        scheduledThreadPool.schedule(() -> {
+            // If onError is called on a completed subject the global error handler is called
+            // REVIEW: Should we lock?
```
anurse: So the worst case scenario here is:

1. Handshake Response arrives *barely* in time
1. Subject is completed AFTER this code checks `hasComplete`
1. Immediately after being completed, the Subject is `onError`ed by the timeout
1. Start completes because the subject was completed
1. Global onError handler is fired with the timeout error.

Technically there's nothing that will "fail" the process. The handler is called spuriously but the actual action succeeds. If the race goes the other way (timeout trips before the subject is completed) then so be it (that race exists in .NET and we're fine with it there)

I'm OK with that rare race, at least until we get feedback about it happening in the wild. The locking would be a bit rough here since we'd need to lock around all completion actions.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -187,15 +189,22 @@ void setTickRate(Duration tickRate) {
...
+        }, timeout, unit);
     }
 
-    private CompletableFuture<NegotiateResponse> handleNegotiate(String url) {
+    private Single<NegotiateResponse> handleNegotiate(String url) {
         HttpRequest request = new HttpRequest();
         request.addHeaders(this.headers);
 
-        return httpClient.post(Negotiate.resolveNegotiateUrl(url), request).thenCompose((response) -> {
+        return httpClient.post(Negotiate.resolveNegotiateUrl(url), request).flatMap((response) -> {
```
anurse: If you're returning `Single.just` you don't need `flatMap`, you can just use `map` and return `negotiateResponse`. `flatMap` is used when you need to perform another **asynchronous** operation when the previous operation completes.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -236,25 +245,27 @@ public Completable start() {
             return Completable.complete();
         }
 
-        handshakeResponseFuture = new CompletableFuture<>();
+        handshakeResponseSubject = CompletableSubject.create();
         handshakeReceived = false;
-        CompletableFuture<Void> tokenFuture = new CompletableFuture<>();
+        CompletableSubject tokenCompletable = CompletableSubject.create();
         accessTokenProvider.subscribe(token -> {
```
anurse: This pattern is going to be a problem when we have transports other than WebSockets because we need to "re-run" the provider on each request in case the token needs to be updated...

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -300,18 +312,24 @@ public void run() {
...
-                });
-            });
-        }));
+
+                        return Completable.complete();
+                    }));
+                }));
+            }));
+        // subscribe makes this a "hot" completable so this runs immediately
+        }).subscribe(() -> start.onComplete(),
```
anurse: I think you can just use `.subscribeWith(start)` since `CompletableSubject` implements `CompletableObserver`

clients/java/signalr/src/main/java/com/microsoft/signalr/InvocationRequest.java
```diff
@@ -17,21 +17,21 @@
...
+        pendingCall.onError(ex);
     }
 
     public void cancel() {
-        pendingCall.cancel(false);
+        pendingCall.onError(new RuntimeException("Invocation was canceled."));
     }
 
-    public CompletableFuture<Object> getPendingCall() {
+    public SingleSubject<Object> getPendingCall() {
```
anurse: nit: make this return a `Single`. There's no need for the caller to get the subject.

clients/java/signalr/src/main/java/com/microsoft/signalr/WebSocketTransport.java
```diff
@@ -82,8 +83,8 @@ public void setOnClose(Consumer<String> onCloseCallback) {
     }
 
     @Override
-    public CompletableFuture<Void> stop() {
-        return webSocketClient.stop().whenComplete((i, j) -> logger.info("WebSocket connection stopped."));
+    public Completable stop() {
+        return webSocketClient.stop().doOnEvent(t -> logger.info("WebSocket connection stopped."));
```
anurse: Why is this `doOnEvent` vs `doOnComplete`?

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
BrennanConroy: Yeah, I was leaning towards not worrying about it.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
BrennanConroy: This should only trigger once since it's a `Single`. In the other transports we'll probably re-subscribe every time we want to call it.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
BrennanConroy: Oh neat!

clients/java/signalr/src/main/java/com/microsoft/signalr/WebSocketTransport.java
BrennanConroy: We still want to log even if there was a websocket error on close

anurse: Ah, so `onEvent` fires in either case. That's the part I was missing :).

clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
anurse: Right, but we'll need to flow it down to the transports rather than burning the token into the headers.

clients/java/signalr/src/main/java/com/microsoft/signalr/WebSocketTransport.java
```diff
@@ -4,12 +4,13 @@
 package com.microsoft.signalr;
 
 import java.util.Map;
-import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.reactivex.Completable;
```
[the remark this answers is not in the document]
BrennanConroy: Spotless didn't seem to care

BrennanConroy:
Ping

PR #3156, opened 2018-10-19
Adding support for micro-benchmarks on CI

sebastienros:


BrennanConroy:
Closing as we've upgraded to a newer version via the normal update deps.

PR #3160, opened 2018-10-19
[Java] Log invocation binding failures

BrennanConroy:
@mikaelm12 noticed this when doing some smoke testing with an Android app, we were seeing the exception locally because `e.printStackTrace();` writes to the console in some environments.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
@@ -10,7 +10,8 @@ enum HubMessageType {
     STREAM_INVOCATION(4),
     CANCEL_INVOCATION(5),
     PING(6),
-    CLOSE(7);
+    CLOSE(7),
+    INVOCATIONBINDINGFAILURE(-10);
 
     public int value;
     HubMessageType(int id) { this.value = id; }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
@@ -11,7 +11,7 @@ enum HubMessageType {
     CANCEL_INVOCATION(5),
     PING(6),
     CLOSE(7),
-    INVOCATIONBINDINGFAILURE(-1);
+    INVOCATION_BINDING_FAILURE(-1);
 
     public int value;
     HubMessageType(int id) { this.value = id; }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
@@ -11,7 +11,7 @@ enum HubMessageType {
     CANCEL_INVOCATION(5),
     PING(6),
     CLOSE(7),
-    INVOCATIONBINDINGFAILURE(-10);
+    INVOCATIONBINDINGFAILURE(-1);
 
     public int value;
     HubMessageType(int id) { this.value = id; }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/InvocationBindingFailureMessage.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/InvocationBindingFailureMessage.java
@@ -1,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
+
+class InvocationBindingFailureMessage extends HubMessage {
+    private final String invocationId;
+    private final String target;
+    private final Exception exception;
+
+    public InvocationBindingFailureMessage(String invocationId, String target, Exception exception) {
+        this.invocationId = invocationId;
+        this.target = target;
+        this.exception = exception;
+    }
+
+    public String getInvocationId() {
+        return invocationId;
+    }
+
+    public String getTarget() {
+        return target;
+    }
+
+    public Exception getException() {
+        return exception;
+    }
+
+    @Override
+    public HubMessageType getMessageType() {
+        return HubMessageType.INVOCATIONBINDINGFAILURE;
+    }
+}
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/InvocationBindingFailureMessage.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/InvocationBindingFailureMessage.java
@@ -28,6 +28,6 @@ class InvocationBindingFailureMessage extends HubMessage {
 
     @Override
     public HubMessageType getMessageType() {
-        return HubMessageType.INVOCATIONBINDINGFAILURE;
+        return HubMessageType.INVOCATION_BINDING_FAILURE;
     }
 }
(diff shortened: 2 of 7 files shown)


clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
```diff
@@ -10,7 +10,8 @@
     STREAM_INVOCATION(4),
     CANCEL_INVOCATION(5),
     PING(6),
-    CLOSE(7);
+    CLOSE(7),
+    INVOCATIONBINDINGFAILURE(-10);
```
mikaelm12: Why not -1?

clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
```diff
@@ -81,12 +81,7 @@ public void onOpen(WebSocket webSocket, Response response) {
 
         @Override
         public void onMessage(WebSocket webSocket, String message) {
-            try {
-                onReceive.invoke(message);
-            } catch (Exception e) {
-                // TODO Auto-generated catch block
-                e.printStackTrace();
-            }
+            onReceive.invoke(message);
```
mikaelm12: 👍

clients/java/signalr/src/test/java/com/microsoft/signalr/JsonHubProtocolTest.java
```diff
@@ -220,30 +220,51 @@ public void parseCompletionMessageWithOutOfOrderProperties() {
...
+    public void invocationBindingFailureWhileParsingTooManyArgumentsWithOutOfOrderProperties() {
         String stringifiedMessage = "{\"arguments\":[42, 24],\"type\":1,\"target\":\"test\"}\u001E";
         TestBinder binder = new TestBinder(new InvocationMessage(null, "test", new Object[] { 42 }));
 
-        RuntimeException exception = assertThrows(RuntimeException.class, () -> jsonHubProtocol.parseMessages(stringifiedMessage, binder));
-        assertEquals("Invocation provides 2 argument(s) but target expects 1.", exception.getMessage());
+        HubMessage[] messages = jsonHubProtocol.parseMessages(stringifiedMessage, binder);
+        assertEquals(1, messages.length);
+        assertEquals(InvocationBindingFailureMessage.class, messages[0].getClass());
+        assertEquals("Invocation provides 2 argument(s) but target expects 1.",
```
mikaelm12: We split the cast into a separate line in every other test. It's cleaner.

clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
mikaelm12: I don't think we should add a dedicated message for this.

mikaelm12: If anything a generic error message which wraps the actual error which we could reuse for other things that could come up.

anurse: We need a message type here, like in .NET, because we need to know the invocation ID and only the parser knows that. We use a pseudo-message to communicate the invocation ID and the binding error back to the caller of the parser.

anurse: I'd be OK with `INVOCATIONPARSINGFAILURE` I guess, but it still needs to be clearly an "invocation-related" error. .NET uses `InvocationBindingFailure` so I think I prefer consistency here (though that's just for our benefit rather than any customer-facing consistency)

clients/java/signalr/src/main/java/com/microsoft/signalr/JsonHubProtocol.java
```diff
@@ -83,8 +84,21 @@ public TransferFormat getTransferFormat() {
                             break;
                         case "arguments":
                             if (target != null) {
-                                List<Class<?>> types = binder.getParameterTypes(target);
-                                arguments = bindArguments(reader, types);
+                                try {
+                                    List<Class<?>> types = binder.getParameterTypes(target);
```
anurse: keep this line outside the try

Why? Because if `getParameterTypes` throws, the error recovery code will go haywire and start gobbling lots of JSON. Plus, we actually want an exception there to propagate out.

BrennanConroy: We put it inside the try in .NET too. The only error that should be thrown from `getParameterTypes` is "no callback registered"

anurse: Ok, if it's inside the try in .NET I guess we keep it in here too :)

anurse: Can you add a test for the case that throws that exception? I'm most worried about the JSON reader recovering properly. Make sure to put the invalid information **before** the Invocation ID in the test JSON payload so we can verify that the reader "recovers" and still is able to read the invocation ID

clients/java/signalr/src/main/java/com/microsoft/signalr/HubMessageType.java
```diff
@@ -10,7 +10,8 @@
     STREAM_INVOCATION(4),
     CANCEL_INVOCATION(5),
     PING(6),
-    CLOSE(7);
+    CLOSE(7),
+    INVOCATIONBINDINGFAILURE(-1);
```
mikaelm12: nit: `INVOCATION_BINDING_FAILURE`

clients/java/signalr/src/main/java/com/microsoft/signalr/InvocationBindingFailureMessage.java
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
+
+class InvocationBindingFailureMessage extends HubMessage {
+    private final String invocationId;
```
mikaelm12: In the other message types we include the HubMessage type as a property 
So something `private final int type = HubMessageType.INVOCATION_BINDING_FAILURE`

BrennanConroy: The reason we have that is so that gson writes the `type` field into the json. Since this type never goes through gson it doesn't need it.

BrennanConroy:
🆙 📅

clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -913,8 +913,8 @@ public void errorWhenReceivingInvokeWithIncorrectArgumentLength()  {
 
         hubConnection.start().timeout(1, TimeUnit.SECONDS).blockingAwait();
 
-        RuntimeException exception = assertThrows(RuntimeException.class, () -> mockTransport.receiveMessage("{\"type\":1,\"target\":\"Send\",\"arguments\":[]}" + RECORD_SEPARATOR));
-        assertEquals("Invocation provides 0 argument(s) but target expects 1.", exception.getMessage());
+        mockTransport.receiveMessage("{\"type\":1,\"target\":\"Send\",\"arguments\":[]}" + RECORD_SEPARATOR);
```
anurse: Filed https://github.com/aspnet/SignalR/issues/3225

BrennanConroy: Is this your version of :shipit:
:trollface:

PR #3170, opened 2018-10-22
Update deps

BrennanConroy:
Should fix the VSTS failures


BrennanConroy:
Fun, github has paused webhooks while they work on some issues...

anurse:
You need to manually retrigger the build too :)

Issue #3168, opened 2018-10-22
Non-browser client fails to create secure connection "1.1.0-preview3-35497"


Havunen:
Hi

There seems to be bug in preview version of npm `"@aspnet/signalr": "^1.1.0-preview3-35497"` when using SSL connection ( https / secure socket ). Our SignalR server is not accepting non secure connections and defining secure url does not use secure connection on non browser environment. 

Tested in
`Windows 10 + NodeJS  v10.11.0`

If you look into source code, it hard coded to use `"http"` client when using nodeJS. See here:

https://github.com/aspnet/SignalR/blob/a51eec7a1d76272721da7437a273b3c6c74c581a/clients/ts/signalr/src/DefaultHttpClient.ts#L10-L13

https://github.com/aspnet/SignalR/blob/a51eec7a1d76272721da7437a273b3c6c74c581a/clients/ts/signalr/src/NodeHttpClient.ts#L4

In NodeJS to open secure http connection it needs `https` client

Steps to reproduce:

index.js
```js
const signalR = require('@aspnet/signalr');

connection = global._conn = new signalR.HubConnectionBuilder().withUrl(
  'https://URL'
).build();

// Add connection listeners....

connection.start().catch(err => {console.error(err);});
```

Server responds http https redirection answer: `Error: Moved Permanently: 301 `

If I change source code to use `require("https");` instead of `require("http");`  it goes through.
But fails to Websocket connection `Error: Failed to start the transport 'WebSockets': null`

EDIT: This is probably non relevant
to fix that I changed:
https://github.com/aspnet/SignalR/blob/a51eec7a1d76272721da7437a273b3c6c74c581a/clients/ts/signalr/src/WebSocketTransport.ts#L46

`http => https` and `ws => wss`

Without secure connection all works as expected

anurse:
Good catch. We should import both modules in `NodeHttpClient` and select the appropriate one based on the URL.

anurse:
Your later change to the `url.replace` shouldn't actually be necessary since we intentional don't put the `:` in the regex, so if the URL is `https://foo`, the regex will replace `http` with `ws` and you'll end up with `wss://foo` (because the `s` from `https` was left behind). Is this not working for you? It works in our tests so I want to make sure that if there's an issue we update our tests.

anurse:
> If I change source code to use `require("https");` instead of `require("http");` it goes through.
> But fails to Websocket connection `Error: Failed to start the transport 'WebSockets': null`

[...]

PR #3171, opened 2018-10-22
Use Node https module for https requests

BrennanConroy:
Fixes https://github.com/aspnet/SignalR/issues/3168

I'll see if I can get a test certificate on our functional tests server so we can run https tests. But I want this in ASAP

### DIFF
--- a/clients/ts/signalr/src/NodeHttpClient.ts
+++ b/clients/ts/signalr/src/NodeHttpClient.ts
@@ -34,6 +34,9 @@ export class NodeHttpClient extends HttpClient {
                 port: url.port,
             };
 
+            // "any" is used here because require() can't be correctly resolved by the compiler
+            // when httpOrHttps is typeof http | typeof https. Change to http when editing to get
+            // intellisense.
             const httpOrHttps: any = url.protocol === "https" ? https : http;
 
             const req = httpOrHttps.request(options, (res: http.IncomingMessage) => {
--- a/clients/ts/signalr/src/NodeHttpClient.ts
+++ b/clients/ts/signalr/src/NodeHttpClient.ts
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 import * as http from "http";
+import * as https from "http";
 import { URL } from "url";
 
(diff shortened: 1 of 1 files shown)


clients/ts/signalr/src/NodeHttpClient.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 import * as http from "http";
+import * as https from "http";
```
campersau: I think this should be:
```suggestion
import * as https from "https";
```

BrennanConroy: Whoops, copy-pasta error, thanks

clients/ts/signalr/src/NodeHttpClient.ts
```diff
@@ -33,7 +34,15 @@ export class NodeHttpClient extends HttpClient {
                 port: url.port,
             };
 
-            const req = http.request(options, (res: http.IncomingMessage) => {
+            let httpOrHttps = http;
+            if (request.url) {
+                // poor man's "startsWith()"
+                if (request.url.lastIndexOf("https", 0) === 0) {
```
anurse: Check `url.protocol` instead. We already load the URL up into a `URL` so use it :).

```javascript
const httpOrHttps = url.protocol === "https:" ? https : http;
```

clients/ts/signalr/src/NodeHttpClient.ts
```diff
@@ -33,7 +34,9 @@ export class NodeHttpClient extends HttpClient {
                 port: url.port,
             };
 
-            const req = http.request(options, (res: http.IncomingMessage) => {
+            const httpOrHttps: any = url.protocol === "https" ? https : http;
```
BrennanConroy: `const httpOrHttps: typeof http | typeof https;` doesn't work I believe because it can't correctly resolve `.request` below since one returns `ClientRequest` and the other returns `http.ClientRequest` even though they are the same thing.

anurse: Stick that comment in the code, it'll last longer ;P

Issue #3164, opened 2018-10-19
Remove Functional Interfaces and java.util.Function dependencies


mikaelm12:
To target Android API Level 23 or lower we need to remove dependencies on `java.lang.FunctionalInterface` and `java.util.Function` which means we have to remove the functional interfaces and the references to `Consumer<T>`, `BiFunction<...>`, etc

anurse:
Closed in #3186

PR #3186, opened 2018-10-23
Remove Dependency on java.util.Function

mikaelm12:
Removing our uses of Consumer,BiConsumer etc and replacing them with our own interfaces. Moving towards being able to target Android API level 23.
The other half of issue: https://github.com/aspnet/SignalR/issues/3164

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
@@ -4,5 +4,5 @@
 package com.microsoft.signalr;
 
 public interface OnClosedCallback {
-    void invoke(Exception param1);
+    void invoke(Exception exception);
 }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 package com.microsoft.signalr;
 
 public interface OnClosedCallback {
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
@@ -1,0 +1,5 @@
+package com.microsoft.signalr;
+
+public interface OnClosedCallback {
+    void invoke(Exception param1);
+}
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
@@ -3,6 +3,9 @@
 
 package com.microsoft.signalr;
 
+/**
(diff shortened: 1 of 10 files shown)


clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
+
+public interface OnClosedCallback {
+    void invoke(Exception param1);
```
BrennanConroy: nit: "param1" -> "exception"

clients/java/signalr/src/main/java/com/microsoft/signalr/WebsocketWrapperOnClosed.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
+
+interface WebsocketWrapperOnClosed {
+    void invoke(Integer code, String string);
```
BrennanConroy: nit: "string" -> "reason"

clients/java/signalr/src/main/java/com/microsoft/signalr/WebsocketWrapperOnClosed.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
+
+interface WebsocketWrapperOnClosed {
```
anurse: Upper case `S` and there's no need to include `Wrapper` in this type even though it's used by the `WebSocketWrapper` type (it's still triggered because the WebSocket closed, so the name only needs to reflect that). And I think it should be a little more aligned with our other callback types with the `Callback` suffix:

```suggestion
interface WebSocketOnClosedCallback {
```

clients/java/signalr/src/test/java/com/microsoft/signalr/TestHttpClient.java
```diff
@@ -66,4 +65,8 @@ public TestHttpClient on(String method, String url, Function<HttpRequest, Single
     public WebSocketWrapper createWebSocket(String url, Map<String, String> headers) {
         throw new RuntimeException("WebSockets isn't supported in testing currently.");
     }
+
+    interface TestHttpClientHandler {
```
anurse: A slightly more descriptive name (since it is a function that handles requests) suggestion:

```suggestion
    interface TestHttpRequestHandler {
```

clients/java/signalr/src/main/java/com/microsoft/signalr/OnClosedCallback.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+package com.microsoft.signalr;
+
+public interface OnClosedCallback {
```
BrennanConroy: doc comments

PR #3194, opened 2018-10-24
Update npm readme with Node info

anurse:
I noticed that these READMEs are out of date now that we have "official" node support.


clients/ts/signalr/README.md
```diff
@@ -16,10 +16,7 @@ To use the client in a browser, copy `*.js` files from the `dist/browser` folder
 
 ### Node.js
 
-The following polyfills are required to use the client in Node.js applications:
```
mikaelm12: 😄

Issue #3192, opened 2018-10-24
Azure SignalR service and dotnet core non browser client


Havunen:
Hi,
I'm having an issue connecting non browser client to azure signalR service using dotnet core and nodejs client.

Client Setup
`NodeJS 10 + 1.1.0-preview3 ( with HTTPS work around )`

Server ( Azure )
```
- Microsoft.AspNetCore.All ( 2.1.5 )
- Microsoft.Azure.SignalR ( 1.0.2 )
```

## Local machine setup: this works
```C#
  services.AddCors(options => options.AddPolicy("CorsPolicy", builder =>
            {
          builder
             .AllowAnyMethod()
             .AllowAnyHeader()
             .AllowCredentials()
           .WithOrigins("http://localhost:3000");
}));
services.AddSignalR();
```

Configure...
```c#


            app.UseCors("CorsPolicy");
            app.UseSignalR((routes) =>
            {
                routes.MapHub<MonkeyHub>("/monkeyHub");
                routes.MapHub<GorillaHub>("/gorillaHub");
                routes.MapHub<MasterHub>("/masterHub");
            });
```

## Server ( Azure setup ) this fails

```c#
            services.AddCors(options => options.AddPolicy("CorsPolicy", builder =>
            {
                builder
                    .AllowAnyMethod()
                    .AllowAnyHeader()
                    .AllowCredentials()
                    .WithOrigins("http://localhost:3000");
            }));
services.AddSignalR().AddAzureSignalR(Configuration.GetConnectionString("SignalR"));
```

Configure...
```C#

            app.UseAzureSignalR((routes) =>
            {
                routes.MapHub<MonkeyHub>("/monkeyHub");
                routes.MapHub<GorillaHub>("/gorillaHub");
                routes.MapHub<MasterHub>("/masterHub");
            });
```

Here is log:
```
__COMPUTER_PATH__>node gorilla.js
Recreating connection
Connection starting...
ErrorEvent {
  target:
   WebSocket {
     _events:
      { open: [Function],
        error: [Function],
        message: [Function],
        close: [Function] },
     _eventsCount: 4,
     _maxListeners: undefined,
     readyState: 2,
     protocol: '',
     _binaryType: 'nodebuffer',
     _closeFrameReceived: false,
     _closeFrameSent: false,
     _closeMessage: '',
     _closeTimer: null,
     _closeCode: 1006,
     _extensions: {},
     _isServer: false,
     _receiver: null,
     _sender: null,
[...]

PR #3202, opened 2018-10-25
Enable cookie jar for Node client using request package

BrennanConroy:
Fixes https://github.com/aspnet/SignalR/issues/3192

### DIFF
--- a/clients/ts/FunctionalTests/Startup.cs
+++ b/clients/ts/FunctionalTests/Startup.cs
@@ -108,6 +108,7 @@ namespace FunctionalTests
                 if (context.Request.Path.Value.Contains("/negotiate"))
                 {
                     context.Response.Cookies.Append("testCookie", "testValue");
+                    context.Response.Cookies.Append("testCookie2", "testValue2");
                     context.Response.Cookies.Append("expiredCookie", "doesntmatter", new CookieOptions() { Expires = DateTimeOffset.Now.AddHours(-1) });
                 }
                 await next.Invoke();
--- a/clients/ts/FunctionalTests/Startup.cs
+++ b/clients/ts/FunctionalTests/Startup.cs
@@ -103,6 +103,16 @@ namespace FunctionalTests
                 routes.MapConnectionHandler<EchoConnectionHandler>("/echo");
             });
 
+            app.Use(async (context, next) =>
+            {
+                if (context.Request.Path.Value.Contains("/negotiate"))
+                {
+                    context.Response.Cookies.Append("testCookie", "testValue");
+                    context.Response.Cookies.Append("expiredCookie", "doesntmatter", new CookieOptions() { Expires = DateTimeOffset.Now.AddHours(-1) });
+                }
+                await next.Invoke();
+            });
+
             app.UseSignalR(routes =>
             {
                 routes.MapHub<TestHub>("/testhub");
--- a/clients/ts/signalr/package.json
+++ b/clients/ts/signalr/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aspnet/signalr",
-  "version": "1.1.0-preview1-t000",
+  "version": "1.1.0-rtm-t000",
   "description": "ASP.NET Core SignalR Client",
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm/index.js",
@@ -38,10 +38,12 @@
   "devDependencies": {
     "es6-promise": "^4.2.2",
     "@types/node": "^10.9.4",
-    "@types/eventsource": "^1.0.2"
+    "@types/eventsource": "^1.0.2",
+    "@types/request": "^2.47.1"
   },
   "dependencies": {
     "ws": "^6.0.0",
-    "eventsource": "^1.0.7"
+    "eventsource": "^1.0.7",
+    "request": "^2.88.0"
   }
 }
--- a/clients/ts/signalr/src/DefaultHttpClient.ts
+++ b/clients/ts/signalr/src/DefaultHttpClient.ts
@@ -45,4 +45,12 @@ export class DefaultHttpClient extends HttpClient {
 
         return this.httpClient.send(request);
     }
+
+    // @ts-ignore
+    private getCookies(url: string): string {
+        if ((this.httpClient as any).getCookies) {
+            return (this.httpClient as any).getCookies(url);
+        }
+        return "";
+    }
 }
--- a/clients/ts/signalr/src/DefaultHttpClient.ts
+++ b/clients/ts/signalr/src/DefaultHttpClient.ts
@@ -48,7 +48,7 @@ export class DefaultHttpClient extends HttpClient {
 
     public getCookies(url: string): string {
         if ((this.httpClient as any).getCookies) {
-            return (this.httpClient as any).getCookies(url);
+            return this.httpClient.getCookieString(url);
         }
         return "";
     }
--- a/clients/ts/signalr/src/DefaultHttpClient.ts
+++ b/clients/ts/signalr/src/DefaultHttpClient.ts
@@ -46,8 +46,7 @@ export class DefaultHttpClient extends HttpClient {
         return this.httpClient.send(request);
     }
 
-    // @ts-ignore
-    private getCookies(url: string): string {
+    public getCookies(url: string): string {
         if ((this.httpClient as any).getCookies) {
             return (this.httpClient as any).getCookies(url);
         }
--- a/clients/ts/signalr/src/DefaultHttpClient.ts
+++ b/clients/ts/signalr/src/DefaultHttpClient.ts
@@ -46,10 +46,7 @@ export class DefaultHttpClient extends HttpClient {
         return this.httpClient.send(request);
     }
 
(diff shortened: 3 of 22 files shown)


clients/ts/signalr/src/DefaultHttpClient.ts
```diff
@@ -45,4 +45,12 @@ export class DefaultHttpClient extends HttpClient {
 
         return this.httpClient.send(request);
     }
+
+    // @ts-ignore
+    private getCookies(url: string): string {
```
BrennanConroy: Have to do it this way so we don't make a breaking change by requiring this function on the base class `HttpClient`

anurse: Couldn't you define it as non-abstract and have it return an empty list?

clients/ts/FunctionalTests/Startup.cs
```diff
@@ -103,6 +103,16 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 routes.MapConnectionHandler<EchoConnectionHandler>("/echo");
             });
 
+            app.Use(async (context, next) =>
+            {
+                if (context.Request.Path.Value.Contains("/negotiate"))
+                {
+                    context.Response.Cookies.Append("testCookie", "testValue");
```
anurse: Can you put two valid cookies in? I'm paranoid about `getCookieString` only returning one cookie or something :)

clients/ts/signalr/package.json
```diff
@@ -38,10 +38,12 @@
   "devDependencies": {
     "es6-promise": "^4.2.2",
     "@types/node": "^10.9.4",
-    "@types/eventsource": "^1.0.2"
+    "@types/eventsource": "^1.0.2",
+    "@types/request": "^2.47.1"
```
anurse: @Eilon (this is just the typescript typings for `request` below)

clients/ts/signalr/package.json
```diff
@@ -38,10 +38,12 @@
...
   },
   "dependencies": {
     "ws": "^6.0.0",
-    "eventsource": "^1.0.7"
+    "eventsource": "^1.0.7",
+    "request": "^2.88.0"
```
anurse: @Eilon

clients/ts/signalr/package.json
Eilon: Good to go.

clients/ts/signalr/package.json
Eilon: Good to go.

clients/ts/signalr/src/DefaultHttpClient.ts
anurse: [You can override methods just fine in subclasses...](https://www.typescriptlang.org/play/index.html#src=abstract%20class%20Base%20%7B%0D%0A%20%20%20%20method2()%3A%20string%20%7B%20return%20%22Base%22%20%7D%0D%0A%7D%0D%0A%0D%0Aclass%20Derived%20extends%20Base%20%7B%0D%0A%20%20%20%20method2()%3A%20string%20%7B%0D%0A%20%20%20%20%20%20%20%20return%20%22Derived%22%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Aclass%20NonOverride%20extends%20Base%20%7B%0D%0A%7D%0D%0A%0D%0Aconsole.log(new%20NonOverride().method2())%3B%0D%0Aconsole.log(new%20Derived().method2())%3B) In TypeScript, everything is `virtual` ;)

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -56,7 +56,11 @@ export class ServerSentEventsTransport implements ITransport {
                 return;
             }
 
-            const eventSource = new this.eventSourceConstructor(url, { withCredentials: true });
+            let cookies: string = "";
+            if ((this.httpClient as any).getCookies) {
+                cookies = (this.httpClient as any).getCookies(url);
+            }
+            const eventSource = new this.eventSourceConstructor(url, { withCredentials: true, headers: { Cookie: cookies } } as EventSourceInit);
```
anurse: Can you explicitly avoid providing `headers` if we're in a browser (`typeof window !== "undefined"`)? I don't want it to weird out if browsers add some support for this but, for example, throw if the `Cookie` header is set manually (or something similar).

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -44,7 +47,22 @@ export class WebSocketTransport implements ITransport {
 
         return new Promise<void>((resolve, reject) => {
             url = url.replace(/^http/, "ws");
-            const webSocket = new this.webSocketConstructor(url);
+            let webSocket: WebSocket;
+            let cookies: string = "";
+            if ((this.httpClient as any).getCookies) {
+                cookies = (this.httpClient as any).getCookies(url);
+            }
+            if (cookies !== "") {
```
anurse: `if(cookies)` should be sufficient here.

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -44,7 +47,22 @@ export class WebSocketTransport implements ITransport {
...
+                webSocket = new this.webSocketConstructor(url, undefined, {
```
anurse: As with `EventSource`, can you only set the headers if we're not in the browser?

BrennanConroy:
:up: 📅

clients/ts/signalr/src/DefaultHttpClient.ts
```diff
@@ -45,4 +45,11 @@ export class DefaultHttpClient extends HttpClient {
 
         return this.httpClient.send(request);
     }
+
+    public getCookies(url: string): string {
```
anurse: I was thinking you'd add this to `HttpClient` as non-`abstract`...

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -141,4 +141,14 @@ export abstract class HttpClient {
...
      */
     public abstract send(request: HttpRequest): Promise<HttpResponse>;
+
+    /**
+     *
+     * @param url The URL that the cookies are valid for.
+     * @returns {string} A string containing all the key-value cookie pairs for the specified URL.
+     */
+    // @ts-ignore
+    public getCookies(url: string): string {
```
anurse:
```suggestion
    public getCookieString(url: string): string {
```

clients/ts/signalr/src/DefaultHttpClient.ts
```diff
@@ -45,4 +45,11 @@ export class DefaultHttpClient extends HttpClient {
...
+        if ((this.httpClient as any).getCookies) {
+            return (this.httpClient as any).getCookies(url);
```
anurse:
```suggestion
            return this.httpClient.getCookieString(url);
```

clients/ts/signalr/src/HttpClient.ts
```diff
@@ -141,4 +141,14 @@ export abstract class HttpClient {
      * @returns {Promise<HttpResponse>} A Promise that resolves with an HttpResponse describing the response, or rejects with an Error indicating a failure.
      */
     public abstract send(request: HttpRequest): Promise<HttpResponse>;
+
+    /**
```
anurse:
```suggestion
    /** Hello docs?
```

clients/ts/signalr/src/ServerSentEventsTransport.ts
```diff
@@ -56,7 +56,17 @@ export class ServerSentEventsTransport implements ITransport {
                 return;
             }
 
-            const eventSource = new this.eventSourceConstructor(url, { withCredentials: true });
+            let cookies: string = "";
+            if ((this.httpClient as any).getCookies) {
```
anurse: This can just use `this.httpClient.getCookieString()` safely now.

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -44,7 +47,28 @@ export class WebSocketTransport implements ITransport {
...
             url = url.replace(/^http/, "ws");
-            const webSocket = new this.webSocketConstructor(url);
+            let webSocket: WebSocket | undefined;
+            let cookies: string = "";
+            if ((this.httpClient as any).getCookies) {
+                cookies = (this.httpClient as any).getCookies(url);
+            }
+
+            if (typeof window === "undefined") {
+                if (cookies) {
```
anurse: roll this up into the previous condition `typeof window === "undefined" && cookies`

clients/ts/signalr/src/DefaultHttpClient.ts
```diff
@@ -45,4 +45,11 @@ export class DefaultHttpClient extends HttpClient {
 
         return this.httpClient.send(request);
     }
+
+    public getCookies(url: string): string {
+        if ((this.httpClient as any).getCookies) {
```
anurse: ... well now this will never work :). You can remove the `if` entirely :)

PR #3215, opened 2018-10-26
Remove hidden BiConsumer reference

mikaelm12:
Map's forEach method takes a BiConsumer. Removing it for Android API compatibility.


clients/java/signalr/src/main/java/com/microsoft/signalr/DefaultHttpClient.java
```diff
@@ -97,6 +98,10 @@ public void saveFromResponse(HttpUrl url, List<Cookie> cookies) {
 
         if (httpRequest.getHeaders() != null) {
             httpRequest.getHeaders().forEach(requestBuilder::addHeader);
+            Collection<String> keys = httpRequest.getHeaders().keySet();
+            for (String key: keys) {
```
BrennanConroy: nit: space after `key`

PR #3218, opened 2018-10-27
Fix NullPointerException when unboxing null to int.

DHosseiny:
Change int to Integer to allow use null as code used [here](https://github.com/aspnet/SignalR/blob/e1b602a7c5331e0c20efe4a087d73eb63441fd9c/clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java#L103). I think using null as code is legal. just fixed the bug.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/WebSocketTransport.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/WebSocketTransport.java
@@ -86,7 +86,7 @@ class WebSocketTransport implements Transport {
         return webSocketClient.stop().doOnEvent(t -> logger.info("WebSocket connection stopped."));
     }
 
-    void onClose(int code, String reason) {
+    void onClose(Integer code, String reason) {
         logger.info("WebSocket connection stopping with " +
                 "code {} and reason '{}'.", code, reason);
         if (code != 1000) {


davidfowl:
@DHosseiny is there a test you can write to show the before and after failure and pass?

anurse:
Seems like we should try to get this in for 2.2 RTM. @BrennanConroy can you whip up a test if @DHosseiny isn't able to provide one? We should look at getting this in ASAP.

BrennanConroy:
Closing this, as I've grabbed your commit and added another commit with a test here https://github.com/aspnet/SignalR/pull/3236

DHosseiny:
@davidfowl I tried to test it; but I need a tool to simulate connectivity status. [this test](https://github.com/aspnet/SignalR/blob/73f2f1998467e95c3dc6839268282d8b57c2d530/clients/java/signalr/src/test/java/com/microsoft/signalr/WebSocketTransportTest.java#L15) need such a tool too because is failing and passing based on connectivity status(failing when connected to network and passing when there is no connection). if you use a tool for WebSocketTransportTest I will use that tool for testing this issue.

anurse:
@DHosseiny no problem, we wrote the necessary test in https://github.com/aspnet/SignalR/pull/3236 and merged your change.

PR #3227, opened 2018-10-29
[Java] API level support down to 23

BrennanConroy:


clients/java/signalr/src/main/java/com/microsoft/signalr/DefaultHttpClient.java
```diff
@@ -110,7 +110,11 @@ public void saveFromResponse(HttpUrl url, List<Cookie> cookies) {
         client.newCall(request).enqueue(new Callback() {
             @Override
             public void onFailure(Call call, IOException e) {
-                responseSubject.onError(e.getCause());
+                Throwable cause = e.getCause();
```
BrennanConroy: This isn't related to API level, but was seeing a "nullref" in some cases here.

PR #3241, opened 2018-10-31
Remove SignalR.Redis package

BrennanConroy:
Fixes https://github.com/aspnet/SignalR/issues/3076


anurse:
To be clear, this is on `master` so it's for **3.0** (👍)

PR #3247, opened 2018-11-01
Clean up logging

BrennanConroy:
Mostly just adding full stops to places, couple other touch ups.

### DIFF
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
@@ -103,7 +103,7 @@ class OkHttpWebSocketWrapper extends WebSocketWrapper {
             // If the start task hasn't completed yet, then we need to complete it
             // exceptionally.
             if (!startSubject.hasComplete()) {
-                startSubject.onError(new RuntimeException("There was an error starting the Websockets transport."));
+                startSubject.onError(new RuntimeException("There was an error starting the WebSocket transport."));
             }
         }
     }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
@@ -103,7 +103,7 @@ class OkHttpWebSocketWrapper extends WebSocketWrapper {
             // If the start task hasn't completed yet, then we need to complete it
             // exceptionally.
             if (!startSubject.hasComplete()) {
-                startSubject.onError(new RuntimeException("The WebSockets transport was closed while start was in progress."));
+                startSubject.onError(new RuntimeException("There was an error starting the Websockets transport."));
             }
         }
     }
--- a/clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
+++ b/clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
@@ -93,7 +93,7 @@ class OkHttpWebSocketWrapper extends WebSocketWrapper {
 
         @Override
         public void onFailure(WebSocket webSocket, Throwable t, Response response) {
-            logger.error("Websocket closed from an error: {}.", t.getMessage());
+            logger.error("WebSocket closed from an error: {}.", t.getMessage());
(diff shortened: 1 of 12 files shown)


clients/java/signalr/src/main/java/com/microsoft/signalr/OkHttpWebSocketWrapper.java
```diff
@@ -103,9 +103,7 @@ private void checkStartFailure() {
             // If the start task hasn't completed yet, then we need to complete it
             // exceptionally.
             if (!startSubject.hasComplete()) {
-                String errorMessage = "There was an error starting the Websockets transport.";
-                logger.error("Websocket closed from an error: {}.", errorMessage);
-                startSubject.onError(new RuntimeException(errorMessage));
+                startSubject.onError(new RuntimeException("There was an error starting the Websockets transport."));
```
anurse: Casing and I prefer using the singular term (though we're inconsistent about that) since it is a transport that uses a WebSocket (not multiple WebSockets)

```suggestion
                startSubject.onError(new RuntimeException("There was an error starting the WebSocket transport."));
```

clients/java/signalr/src/test/java/com/microsoft/signalr/WebSocketTransportTest.java
```diff
@@ -19,7 +19,7 @@
     @Test
     public void WebsocketThrowsIfItCantConnect() {
         Transport transport = new WebSocketTransport(new HashMap<>(), new DefaultHttpClient());
-        RuntimeException exception = assertThrows(RuntimeException.class, () -> transport.start("http://www.example.com").blockingAwait(1, TimeUnit.SECONDS));
+        RuntimeException exception = assertThrows(RuntimeException.class, () -> transport.start("http://url.fake").blockingAwait(1, TimeUnit.SECONDS));
```
anurse:
```suggestion
        RuntimeException exception = assertThrows(RuntimeException.class, () -> transport.start("http://www.example.com").blockingAwait(1, TimeUnit.SECONDS));
```

`example.com` is specifically reserved [by RFC](https://tools.ietf.org/html/rfc2606#section-2). In theory, I could register `url.fake` (with the new gTLD registrations) and make this test break :). Is there a reason `example.com` was causing a problem?

clients/java/signalr/src/test/java/com/microsoft/signalr/WebSocketTransportTest.java
```diff
@@ -19,7 +19,7 @@
     @Test
     public void WebsocketThrowsIfItCantConnect() {
```
anurse: nit: while you're here

```suggestion
    public void WebSocketThrowsIfItCantConnect() {
```

clients/ts/signalr/src/WebSocketTransport.ts
```diff
@@ -127,7 +127,7 @@ export class WebSocketTransport implements ITransport {
         this.logger.log(LogLevel.Trace, "(WebSockets transport) socket closed.");
         if (this.onclose) {
             if (event && (event.wasClean === false || event.code !== 1000)) {
-                this.onclose(new Error(`Websocket closed with status code: ${event.code} (${event.reason})`));
+                this.onclose(new Error(`Websocket closed with status code: ${event.code} (${event.reason}).`));
```
anurse:
```suggestion
                this.onclose(new Error(`WebSocket closed with status code: ${event.code} (${event.reason}).`));
```

clients/ts/signalr/tests/WebSocketTransport.test.ts
```diff
@@ -129,7 +129,7 @@ describe("WebSocketTransport", () => {
             TestWebSocket.webSocket.onclose(message);
 
             expect(closeCalled).toBe(true);
-            expect(error!).toEqual(new Error("Websocket closed with status code: 1 (just cause)"));
+            expect(error!).toEqual(new Error("Websocket closed with status code: 1 (just cause)."));
```
anurse:
```suggestion
            expect(error!).toEqual(new Error("WebSocket closed with status code: 1 (just cause)."));
```

clients/java/signalr/src/test/java/com/microsoft/signalr/WebSocketTransportTest.java
BrennanConroy: Yes, it makes the test flaky

PR #3248, opened 2018-11-01
Test removal of Jenkins

anurse:
If this works I'll merge it. This is a test of the change from #3233


anurse:
Never mind, that doesn't work because the checks are associated with the commit, not the PR. Whatever :)

PR #3261, opened 2018-11-05
Wrap log in enabled check

BrennanConroy:
DotTrace showed this log took 1.5% of SignalR's runtime (not the apps runtime) with logging turned off.


anurse:
```
2018-11-05T19:56:57.3547890Z   com.microsoft.signalr.WebSocketTransportTest > WebSocketThrowsIfItCantConnect() FAILED
2018-11-05T19:56:57.3548230Z       org.opentest4j.AssertionFailedError at WebSocketTransportTest.java:22
```

https://dnceng.visualstudio.com/public/_build/results?buildId=39851&view=logs

... probably unrelated but 😦

BrennanConroy:
Yeah, I'm thinking we disable that test as its the only "functional" test we have, and it goes to an actual site instead of staying local.

anurse:
Sounds good. We should have proper functional tests with our own local servers anyway.

Issue #3256, opened 2018-11-05
Test failure: HubConnectionCanSendAndReceiveGroupMessages


aspnet-hello:
This issue was made automatically. If there is a problem contact ryanbrandenburg.

This test [failed](http://aspnetci/viewLog.html?buildId=588061&buildTypeId=Releases_22xPublic_UbuntuUniverse) with the following error:

```
System.AggregateException : One or more errors occurred. (No process is associated with this object.) (The following constructor parameters did not have matching fixture data: RedisServerFixture`1 serverFixture)
---- System.InvalidOperationException : No process is associated with this object.
---- The following constructor parameters did not have matching fixture data: RedisServerFixture`1 serverFixture
----- Inner Stack Trace #1 (System.InvalidOperationException) -----
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.GetWaitState()
   at System.Diagnostics.Process.WaitForExitCore(Int32 milliseconds)
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.RunProcessAndWait(String fileName, String arguments, String prefix, ILogger logger, TimeSpan timeout, String& output) in /_/test/Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests/Docker.cs:line 139
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.RunProcessAndThrowIfFailed(String fileName, String arguments, String prefix, ILogger logger, TimeSpan timeout) in /_/test/Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests/Docker.cs:line 120
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.Start(ILogger logger) in /_/test/Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests/Docker.cs:line 89
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.RedisServerFixture`1..ctor() in /_/test/Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests/RedisServerFixture.cs:line 36
----- Inner Stack Trace #2 (Xunit.Sdk.TestClassException) -----
```

Other tests within that build may have failed with a similar message, but they are not listed here. Check the link above for more info.

This test failed on 2.2.

CC muratg

muratg:
[...]

PR #3265, opened 2018-11-06
Retry redis docker

BrennanConroy:
Fixes https://github.com/aspnet/SignalR/issues/3256

cc @muratg Test only 2.2 change

### DIFF
--- a/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
+++ b/test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
@@ -70,6 +70,28 @@ namespace Microsoft.AspNetCore.SignalR.Redis.Tests
             return null;
         }
 
+        private void StartRedis(ILogger logger)
+        {
+            try
+            {
+                Run();
+            }
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Error starting redis docker container, retrying.");
+                Run();
+            }
+
+            void Run()
+            {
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.SignalR.Redis.Tests/Docker.cs
```diff
@@ -70,6 +70,28 @@ private static string GetDockerLocation()
...
 
+        private void StartRedis(ILogger logger)
+        {
+            try
+            {
+                Run();
+            }
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Error starting redis docker container, retrying.");
```
muratg: Wouldn't this be a possible infinite loop?

Do we want a counter and give up after N retries?

BrennanConroy: It only calls `Run` twice, no looping

anurse: Yeah, it's not recursive, it just looks kinda like it is because of the local func ;).

Add a delay? Just a second or two.

anurse: Yeah, then don't :(

anurse: Actually yeah, leave it in. It's OK here

muratg: Cool yeah re-reading I see it now. :)

PR #3272, opened 2018-11-07
Update RxJava dependency

mikaelm12:
RxJava 2.2.3 was released recently. Generally a good idea to keep our dependencies up to date.
[Release notes](https://github.com/ReactiveX/RxJava/releases/tag/v2.2.3)


anurse:
Should we consider this for 2.2? @muratg?

Filed #3237 for the AzDO failure, which appears to be an unrelated flaky test.

anurse:
Given that we don't have functional automation here, can you manually verify the updated version? We don't know what, if any, changes may have come in that could break our usage scenarios.

mikaelm12:
Yeah, I tested the chat sample to verify before making the PR because I had the same reservation as you.

anurse:
Meh, I'm not going to worry about 2.2. It's not like there's anything here we *need*. Let's just carry on and merge once the build is green.

PR #3276, opened 2018-11-07
Add docs blurb to Readme

anurse:
I realized there wasn't really a link to the docs on the README...


anurse:
FYI @muratg README-only change. I can do it in `master` if we're changing the default branch to that. I just want this to land in the default branch so it's on the "main" view for github.com/aspnet/SignalR

PR #3283, opened 2018-11-08
Add HTTPS tests for node/browser tests

### LINKED ISSUE #3279: Run functional tests against HTTPS
We should run our functional tests (browser, node, and maybe .NET?) against an HTTPS endpoint. Not all of them, but pick some that do a common scenario with a few requests and use them. All transports though.

### LINKED ISSUE #3297: .NET Core SignalR - close connection logic
Hello all,

I use SignalR v1.0.3 for ASP.NET Core 2.1 and I have question about close connection logic.
For example consider small example step by step:
1 - Client opened connection
2 - After that, server started produce items to client
3 - Over time client close connection

What happens to the messages that were in the queue for sending?
1 - They be lost 
2 - The connection will be closed as soon as the remaining messages are sent

BrennanConroy:
Fixes #3279

Not sure we need .NET test coverage so didn't add it for now.

@muratg FYI tests for 2.2


clients/ts/FunctionalTests/config.json
```diff
@@ -0,0 +1,13 @@
+{
+    "Kestrel": {
+        "Certificates": {
+            "Default": {
+                "Path": "testcert.pfx",
+                "Password": "testPassword"
+            }
+        },
+        "EndPointDefaults": {
+            "Protocols": "Http1"
```
halter73: This is now the default in 2.2 rtm. Is this not reflected yet in the 2.2.0-rtm-35661 packages?

anurse:
```
2018-11-08T21:15:15.6178226Z     ΓùÅ hubConnection ΓÇ║ using messagepack over WebSockets transport ΓÇ║ using https, can invoke server method and receive result
2018-11-08T21:15:15.6178317Z   
2018-11-08T21:15:15.6178368Z       socket hang up
```

😦

clients/ts/FunctionalTests/Program.cs
```diff
@@ -3,7 +3,9 @@
 
 using System;
 using System.IO;
+using System.Security.Cryptography.X509Certificates;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
```
halter73: Nit: The ConfigurationBuilder should no longer be needed now that config.json is removed.

PR #3292, opened 2018-11-10
Custom CORS-like middleware to allow wildcard origin in test app

BrennanConroy:
@muratg 2.2 Test only

@pranavkm In case you have any input.


clients/ts/FunctionalTests/Startup.cs
```diff
@@ -90,12 +90,36 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 
             app.UseFileServer();
 
-            app.UseCors(policyBuilder =>
+            // Custom CORS to allow any origin + credentials (which isn't allowed by the CORS spec)
```
pranavkm: 👍

Issue #3254, opened 2018-11-04
Angular 7 can't build with package @aspnet/signalr@next (1.1.0 preview3-35497)


x3r10n:
the cli can't resolve the http and https module.

```
ERROR in ./node_modules/@aspnet/signalr/dist/esm/NodeHttpClient.js
Module not found: Error: Can't resolve 'http' in '/Users/patrick/Workspaces/temp/NGSignalr/signalR/node_modules/@aspnet/signalr/dist/esm'
ERROR in ./node_modules/@aspnet/signalr/node_modules/eventsource/lib/eventsource.js
Module not found: Error: Can't resolve 'http' in '/Users/patrick/Workspaces/temp/NGSignalr/signalR/node_modules/@aspnet/signalr/node_modules/eventsource/lib'
ERROR in ./node_modules/@aspnet/signalr/node_modules/eventsource/lib/eventsource.js
Module not found: Error: Can't resolve 'https' in '/Users/patrick/Workspaces/temp/NGSignalr/signalR/node_modules/@aspnet/signalr/node_modules/eventsource/lib'

```

anurse:
@BrennanConroy Could be a WebPack thing. Angular shouldn't be using the `NodeHttpClient` at all...

BrennanConroy:
When importing the SignalR library use:
```ts
import * as signalR from '@aspnet/signalr/dist/browser/signalr.js';
```

@anurse We should talk and see if we can make it so Angular loads that file by default.

anurse:
@BrennanConroy it shouldn't have to, but we may need to change our config to make it work. Angular is using WebPack, which uses the Module version.

rcanpahali:
Any solution for this issue? I'm also facing with this problem...

`NodeInvocationException: Prerendering failed because of error: Error: Cannot find module './NodeHttpClient'
at Function.Module._resolveFilename (module.js:536:15)
at Function.Module._load (module.js:466:25)
at Module.require (module.js:579:17)
at require (internal/module.js:11:18)
at eval (webpack-internal:///./node_modules/@aspnet/signalr/dist/esm/DefaultHttpClient.js:26:28)
at Module../node_modules/@aspnet/signalr/dist/esm/DefaultHttpClient.js (C:\Users\can.pahali\source\WestCore.MvcSpa-Dev\ClientApp\dist\main-server.js:1923:1)
[...]

PR #3306, opened 2018-11-14
Fix dynamic require for Webpack

BrennanConroy:
Fixes https://github.com/aspnet/SignalR/issues/3254

@muratg We'd like to discuss the release timeline for this fix as it's a broken scenario when using Angular 6+.

### DIFF
--- a/clients/ts/signalr/src/DefaultHttpClient.ts
+++ b/clients/ts/signalr/src/DefaultHttpClient.ts
@@ -8,8 +8,10 @@ import { XhrHttpClient } from "./XhrHttpClient";
 
 let nodeHttpClientModule: any;
 if (typeof XMLHttpRequest === "undefined") {
-    // tslint:disable-next-line:no-var-requires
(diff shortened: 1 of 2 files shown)


BrennanConroy:
@anurse Couldn't extract this into a function as webpack needs to know the string value passed into `require` otherwise it complains.

anurse:
> it's a broken scenario when using Angular 6+

It's also a regression. 2.1 did not have this problem and would have worked from Angular 6.

muratg:
Also, should this go to master? Or does master not have this problem?

anurse:
Yes, this will go up to master as well.

vivmishra:
Approved for 2.2.1

natemcmaster:
Before this merges, we need to update branding on this repo. Please wait to merge until that is in.

BrennanConroy:
I'm assuming branding is done and I can push? @natemcmaster

PR #3311, opened 2018-11-16
Dispose WebSocket if StartAsync fails

### LINKED ISSUE #3127: Dispose the ClientWebSocket if WebSocketTransport.StartAsync throws
https://github.com/aspnet/SignalR/blob/facaa1387aae48933bacd2a45d6620fa1d5c9557/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs#L122

Today we only dispose in StopAsync. When the HttpConnection calls StartAsync on the transport and that fails, it nulls out the transport reference and bubbles the exception up. We need to make sure the WebSocket is properly disposed if StartAsync fails.

https://github.com/aspnet/SignalR/blob/facaa1387aae48933bacd2a45d6620fa1d5c9557/src/Microsoft.AspNetCore.Http.Connections.Client/HttpConnection.cs#L460-L473

BrennanConroy:
Fixes #3127 

Don't think we can test this since `_webSocket` is entirely internal and there is no way to check if it's disposed or not.

### DIFF
--- a/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
+++ b/src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
@@ -119,7 +119,15 @@ namespace Microsoft.AspNetCore.Http.Connections.Client.Internal
 
             Log.StartTransport(_logger, transferFormat, resolvedUrl);
 
-            await _webSocket.ConnectAsync(resolvedUrl, CancellationToken.None);
+            try
+            {
+                await _webSocket.ConnectAsync(resolvedUrl, CancellationToken.None);
+            }
+            catch
+            {
+                _webSocket.Dispose();
+                throw;
+            }
 
             Log.StartedTransport(_logger);
 


src/Microsoft.AspNetCore.Http.Connections.Client/Internal/WebSocketsTransport.cs
```diff
@@ -119,7 +119,15 @@ public async Task StartAsync(Uri url, TransferFormat transferFormat)
...
             Log.StartTransport(_logger, transferFormat, resolvedUrl);
 
-            await _webSocket.ConnectAsync(resolvedUrl, CancellationToken.None);
+            try
+            {
+                await _webSocket.ConnectAsync(resolvedUrl, CancellationToken.None);
+            }
+            catch
+            {
+                _webSocket.Dispose();
```
davidfowl: This doesn't throw right?

davidfowl: Actually dispose should never throw 😄

Issue #3316, opened 2018-11-17
[Java] Stream Item payload is an item, not a result


mikaelm12:
The stream item payload is an "item" as per the spec, not a "result". 
https://github.com/aspnet/SignalR/blob/3db604d9d25429b1a622bd018f6d59549a7a9f22/clients/java/signalr/src/main/java/com/microsoft/signalr/StreamItem.java#L9

This messes with serialization.

mikaelm12:
Done: 86787f1c2895dfcff62a569d00f26e5443576e7c

PR #3317, opened 2018-11-17
[Java] Change StreamItem to have items instead of results

mikaelm12:
The payload on `StreamItem` should be an "item" instead of a "result". This has serialization/deserialization implications when reading json payloads. 
https://github.com/aspnet/SignalR/issues/3316


clients/java/signalr/src/test/java/com/microsoft/signalr/sample/Chat.java
```diff
@@ -37,6 +39,10 @@ public static void main(String[] args) {
             // Scans the next token of the input as an int.
             message = reader.nextLine();
             hubConnection.send("Send", enteredName, message);
+            Observable<Integer> result = hubConnection.stream(Integer.class,"ObservableCounter", 10, 100);
```
BrennanConroy: Remove

PR #3326, opened 2018-11-20
Avoid zero-byte send in WebSockets

BrennanConroy:
The motivation behind this was a flaky test that started appearing in some PRs because of a race between the client closing the connection when it received all the bytes and the 0-byte write being sent.

Now we write one segment behind where we've read up to, to allow writing a non 0-byte end of message frame.

### DIFF
--- a/test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
+++ b/test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
@@ -403,31 +403,28 @@ namespace Microsoft.AspNetCore.Http.Connections.Tests
             using (var feature = new TestWebSocketConnectionFeature())
             {
                 var serverSocket = await feature.AcceptAsync();
-                using (var memory = MemoryPool<byte>.Shared.Rent(10))
-                {
-                    var sequence = ReadOnlySequenceFactory.CreateSegments(new byte[] { 1 }, new byte[] { 15 });
-                    Assert.False(sequence.IsSingleSegment);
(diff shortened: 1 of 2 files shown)


davidfowl:
> The motivation behind this was a flaky test that started appearing in some PRs because of a race between the client closing the connection when it received all the bytes and the 0-byte write being sent.

C# tests right? I believe its the only client that doesn't buffer the websocket frames.

anurse:
I'm fine with making more stuff `internal` so it can be accessed directly too, so you don't have to dive through layers of mocking.

anurse:
Actually yeah, it's not that hard. The main method (`SendAsync(this WebSocket ws, ROS<byte>...)`) is public on an internal type already. A little IVT here, a mock WebSocket there and you should be able to test it.

BrennanConroy:
:up: 📅 with test

test/Microsoft.AspNetCore.Http.Connections.Tests/WebSocketsTests.cs
```diff
@@ -396,5 +396,39 @@ public async Task SubProtocolSelectorIsUsedToSelectSubProtocol()
...
             }
         }
+
+        [Fact]
+        public async Task MultiSegmentSendWillNotSendEmptyEndOfMessageFrame()
+        {
+            using (var feature = new TestWebSocketConnectionFeature())
+            {
+                var serverSocket = await feature.AcceptAsync();
+                using (var memory = MemoryPool<byte>.Shared.Rent(10))
```
pakrym: You don't need this.

Issue #3329, opened 2018-11-21
[JAVA] Concurrency and hub connection state


ncthbrt:
The latest version of the Java client on the `/release/2.2` branch does not have measures in place to ensure that the hub connection state is synchronized across threads (with the naked variable being read without concurrency control), this leads to invoke calls failing even when the connection status is `Connected` on other threads. 

A possible fix might be to mark this variable as volatile (though can at times be a fragile solution), or switch to using the accessor at all times, along with some form of locking.

mikaelm12:
Hmm, interesting. We only update the actual `HubConnectionState` inside a lock in the client but I can do some investigation on this. Do you have any more information on a repro? 
Thanks!

ncthbrt:
Apologies, I'm not at a computer right now, but from the source it may be
because you're using a rentrant lock rather than a reentrant read write
lock, which as far as I understand means that reads much also be
synchronised.


I'm not a professional android developer however (work mainly in F#/C#),
soy understanding of Java threading semantics may be poor.

mikaelm12:
Hmm, well maybe would could transition to using a ReadWriteLock since it seems that we're not currently acquiring a lock on reads

ncthbrt:
If need be, could take a look at this on Monday, and maybe make a PR?

mikaelm12:
@ncthbrt I'll take this on. We are considering back porting this change as well.
Thanks for your help!

ncthbrt:
Of course. :)
Btw don't know if you saw the PR above.

mikaelm12:
This issue is being addressed in this PR https://github.com/aspnet/AspNetCore/pull/4400 and will be going into the 2.2.2 patch of the Java client. That issue is being tracked here:https://github.com/aspnet/AspNetCore/issues/4393. Thank you for the contribution @ncthbrt!

PR #3335, opened 2018-11-26
[JAVA] Added read locks for hub connection states

ncthbrt:
As mentioned in Issue #3329, reads of the connection state weren't synchronized. This PR adds a `ReentrantReadWriteLock` to `HubConnection.java`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,5 @@ ipch/
 *.opendb
 *.db
 coverage/
+
+clients/java/signalr/local\.properties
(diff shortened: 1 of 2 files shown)


mikaelm12 (commented):
Thanks for the pr! Looks pretty good. Our branches are currently locked but I'll keep you updated on that

mikaelm12:
Hi @ncthbrt. Thanks again for the PR. I think we found the source of the issue. It isn't a lack of locking on reads but rather not checking the connection state on invokes. We have a new PR here https://github.com/aspnet/AspNetCore/pull/4400. I'm going to go ahead and close this Pr but thanks again for your contribution! We were able to find this bug because of you 👍

